PR #6, opened 2026-04-07
fix: handle Windows file-lock errors in test cleanup

claudlos:
## What does this PR do?

Fixes `PermissionError: [WinError 32]` on Windows when running `pytest tests/ -v`. Two tests (`test_convo_mining`, `test_project_mining`) fail during temp directory cleanup because ChromaDB's `PersistentClient` holds file locks on HNSW data files (`data_level0.bin`, etc.) even after all assertions pass.

**The fix:**
- Delete ChromaDB client/collection references and call `gc.collect()` before `shutil.rmtree`, releasing most file handles
- Use `shutil.rmtree(onerror=...)` to gracefully handle any files still locked by the process, letting the OS clean temp dirs on its own schedule
- Remove unused `sys` imports flagged by ruff

All actual test logic is unchanged — this only affects cleanup.

## How to test

```bash
# On Windows (Python 3.9+)
pip install -e ".[dev]"
pytest tests/ -v
```

Before this fix: 2 FAILED, 7 passed (PermissionError on rmtree)
After this fix: 9 passed

Also works on Linux/macOS (gc.collect + onerror are no-ops when there are no file locks).

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/benchmarks/convomem_bench.py
+++ b/benchmarks/convomem_bench.py
@@ -331,7 +331,11 @@ if __name__ == "__main__":
         default="raw",
         help="Retrieval mode",
     )
-    parser.add_argument("--cache-dir", default="/tmp/convomem_cache", help="Cache directory")
+    parser.add_argument(
+        "--cache-dir",
+        default=str(Path(tempfile.gettempdir()) / "convomem_cache"),
+        help="Cache directory",
+    )
     parser.add_argument("--out", default=None, help="Output JSON file")
     args = parser.parse_args()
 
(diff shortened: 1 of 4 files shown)


bensig:
Hey, thanks for the PR! Tests pass which is great, but CI is failing on `ruff format --check .` — the two test files you touched need reformatting:

```
Would reformat: tests/test_convo_miner.py
Would reformat: tests/test_miner.py
```

Should just be a quick `ruff format tests/` and push. Let me know if you have any questions.

PR #38, opened 2026-04-07
perf: algorithmic improvements across hot paths

tomciopp:
## What does this PR do?
  Fixes some performance issues.

  Batch I/O instead of per-item calls (miner.py, convo_miner.py)
  Mining used to call collection.add() once per chunk. A file that produces
  20 chunks meant 20 round-trips to ChromaDB. Now chunks are accumulated and
  inserted in a single batch call per file.

  Hash-set deduplication instead of per-file queries (miner.py, convo_miner.py)
  Before mining, file_already_mined() ran a ChromaDB get(where=...) query
  for every file in the project. Mining 500 files meant 500 queries. Now all
  known source_files are pre-fetched into a Python set at the start — O(1)
  membership check instead of O(n) queries.

  Metadata-only fetch for Layer 1 selection (layers.py)
  Layer1.generate() fetched ALL documents and metadata from ChromaDB, sorted
  them, and kept 15. Now it fetches metadata only (no document text), scores
  with heapq.nlargest (O(n log k) vs O(n log n)), then fetches documents
  for only the 15 winners by ID.

  BFS adjacency index + graph caching (palace_graph.py)
  traverse() scanned every node for every frontier element — O(V²). Now
  build_graph() produces a wing_to_rooms adjacency dict so each BFS
  expansion is O(degree). The graph is also cached with a 60-second TTL,
  invalidated on drawer add/delete through the MCP server. Three consecutive
  graph operations (traverse, find_tunnels, graph_stats) now share one scan
  instead of three.

  Persistent SQLite connection (knowledge_graph.py)
  Every method opened and closed its own connection. seed_from_entity_facts()
  could cycle through 100+ open/close/commit sequences. Now uses a single
  persistent connection on self, and seed operations run in batch mode
  (one commit at the end instead of per-insert).

  Singleton ChromaDB client (config.py, layers.py, searcher.py, palace_graph.py)
  chromadb.PersistentClient() was instantiated 9+ times across a typical
  session. Added MempalaceConfig.get_chroma_client() which caches one client
  per palace_path. Updated layers, searcher, and palace_graph to use it.

  Small fixes
  - dialect.py: Cache text.lower() before entity loop (was re-created per iteration)
  - miner.py: process_file() returns (count, room) — eliminates a second
  detect_room() call with empty content that gave wrong results
  - miner.py: status() uses paginated 1000-item batches instead of a hardcoded
  limit=10000 that would silently truncate large palaces
  - entity_detector.py: Regex pattern compilation cached per entity name
  
## How to test
  Read code and run locally.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -384,17 +384,21 @@ def mine_convos(
                 "extract_mode": extract_mode,
             })
 
-        try:
-            collection.add(
-                documents=batch_docs,
-                ids=batch_ids,
-                metadatas=batch_metas,
-            )
-            drawers_added = len(batch_docs)
-        except Exception as e:
-            if "already exists" not in str(e).lower():
-                raise
-            drawers_added = 0
+        BATCH_SIZE = 100
+        drawers_added = 0
+        for start in range(0, len(batch_docs), BATCH_SIZE):
+            end = start + BATCH_SIZE
+            try:
+                collection.add(
+                    documents=batch_docs[start:end],
+                    ids=batch_ids[start:end],
+                    metadatas=batch_metas[start:end],
+                )
+                drawers_added += end - start
+            except Exception as e:
+                if "already exists" not in str(e).lower():
+                    raise
+        drawers_added = min(drawers_added, len(batch_docs))
 
         total_drawers += drawers_added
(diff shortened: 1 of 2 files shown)


tomciopp:
Rust backend segfaults when it encounters a desync after inserting too many chunks in one shot. This ends up corrupting the index and causing the program to fail. We fix this by chunking in batches.

PR #44, opened 2026-04-07
fix: make knowledge_graph.py SQLite connections exception-safe

adv3nt3:
## Summary
- Wrap all 8 methods in `KnowledgeGraph` that open SQLite connections with `contextlib.closing()` to guarantee `conn.close()` is called even if `execute()`/`commit()`/`fetchall()` raises. Previously, connections would leak on exception.
- Add transaction context manager (`with conn:`) to write methods (`add_entity`, `add_triple`, `invalidate`) for auto-commit on success and auto-rollback on exception.
- Read methods (`query_entity`, `query_relationship`, `timeline`, `stats`) use `closing()` only.
- `_init_db` keeps manual `commit()` since `executescript` has its own transaction semantics.
- Simplifies the `add_triple` early-return path — no more redundant manual `conn.close()`.

## Test plan
- [x] `ruff check mempalace/knowledge_graph.py` passes clean
- [x] `ruff format --check mempalace/knowledge_graph.py` already formatted
- [x] `python3 -m py_compile mempalace/knowledge_graph.py` compiles OK
- [x] Pattern validated against CPython sqlite3 docs via Context7
- [x] Pyright reports 0 new diagnostics (10 pre-existing Optional type warnings unchanged)

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -39,6 +39,7 @@ import hashlib
 import json
 import os
 import sqlite3
+from contextlib import closing
 from datetime import date, datetime
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


milla-jovovich:
exception-safe connection handling via try/finally wrappers is just good hygiene — leaked connections on error paths were going to bite us. nice

PR #58, opened 2026-04-07
fix: remove redundant "r" mode in config.py open() calls

adv3nt3:
## Summary
Remove unnecessary `"r"` mode argument from two `open()` calls in `config.py`. Python's `open()` defaults to read mode — the explicit `"r"` is redundant. Flagged by ruff UP015.

## Changes
1 file changed, 2 lines modified.

## Test plan
- [x] `ruff check mempalace/config.py` passes clean
- [x] `ruff format --check` already formatted
- [x] `python3 -m py_compile mempalace/config.py` compiles OK

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -84,7 +84,7 @@ class MempalaceConfig:
 
         if self._config_file.exists():
             try:
-                with open(self._config_file, "r") as f:
+                with open(self._config_file) as f:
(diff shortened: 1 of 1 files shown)


bensig:
Appreciate the cleanup but removing the explicit "r" mode is cosmetic and not worth the churn. Closing.

PR #89, opened 2026-04-07
fix(knowledge_graph): use named column access instead of hardcoded integer indices

christauff:
## Problem

`query_entity()`, `query_relationship()`, and `timeline()` access JOIN query results using hardcoded integer column indices (`row[2]`, `row[4]`, `row[5]`, `row[6]`, `row[7]`, `row[10]`, `row[11]`).

These indices are derived from the position of columns in `SELECT t.*, ...` — which means any schema migration (adding a column, reordering columns, changing a JOIN) silently corrupts results without raising an exception. The bug would surface as wrong data returned from graph queries.

Example from `query_entity()` outgoing direction:
```python
results.append({
    "subject": name,
    "predicate": row[2],   # triples.predicate — but fragile
    "object": row[10],     # obj_name from JOIN — breaks if schema changes
    ...
})
```

## Fix

Use `cursor.description` to build a column name → value dict after every `execute()` call. All field access now uses named keys:

```python
cur = conn.execute(query, params)
cols = [d[0] for d in cur.description]
for raw in cur.fetchall():
    row = dict(zip(cols, raw))
    results.append({
        "predicate": row["predicate"],
        "object": row["obj_name"],
        ...
    })
```

This is schema-safe: the column name contract is expressed in the SQL `AS` alias, not in a magic integer offset.

## Test plan

- [ ] `kg.query_entity("Alice", direction="both")` returns correct subject/predicate/object names
- [ ] `kg.query_relationship("child_of")` returns correct sub_name/obj_name
- [ ] `kg.timeline()` returns facts in chronological order with correct field values
- [ ] Add a column to `triples` table schema and verify queries still return correct data

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -305,8 +305,7 @@ class KnowledgeGraph:
             """)
 
         cols = [d[0] for d in cur.description]
-        conn.close()
-        return [
+        results = [
             {
                 "subject": row["sub_name"],
                 "predicate": row["predicate"],
@@ -317,6 +316,8 @@ class KnowledgeGraph:
             }
             for row in (dict(zip(cols, raw)) for raw in cur.fetchall())
         ]
+        conn.close()
+        return results
 
     # ── Stats ─────────────────────────────────────────────────────────────
 
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -201,18 +201,21 @@ class KnowledgeGraph:
             if as_of:
                 query += " AND (t.valid_from IS NULL OR t.valid_from <= ?) AND (t.valid_to IS NULL OR t.valid_to >= ?)"
                 params.extend([as_of, as_of])
-            for row in conn.execute(query, params).fetchall():
+            cur = conn.execute(query, params)
+            cols = [d[0] for d in cur.description]
+            for raw in cur.fetchall():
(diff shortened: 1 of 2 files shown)


web3guru888 (commented):
🔧 **Review of #89** — `fix(knowledge_graph): use named column access instead of hardcoded integer indices`

**Scope**: +180/−41 · 2 file(s) · **touches core**

- ⚠️ `mempalace/knowledge_graph.py` (modified: +50/−41)
- `tests/test_knowledge_graph.py` (added: +130/−0)

### Technical Analysis
- 🕸️ Knowledge graph changes — verify temporal triple ordering and connection lifecycle

### Issues
- ⚠️ Touches `mempalace/knowledge_graph.py` — Core KG — threading and concurrency sensitive

### Suggestions
- Magic number(s) 2015, 2025 — consider extracting to named constant(s)

### Strengths
- ✅ Includes test coverage

🟡 **Needs attention** — touches guarded files and has items to address.


---
<sub>🏛️ Reviewed by **[MemPalace-AGI](https://app.taurus.cloud/dashboards/1a9dd59f-0b12-4c08-9db7-bdbad072fdfd/mempalace-agi-dashboard)** · Autonomous research system with perfect memory · Showcase: **[Truth Palace of Atlantis](https://truthpalace.com)**</sub>

igorls:

PR #115, opened 2026-04-07
Add content hash database with bloom filter for smarter duplicate detection

daedalus:
## Summary

- Add a persistent content hash database that tracks file content via SHA256 hashes
- Implement a bloom filter for fast O(1) duplicate checking before reading files from disk
- Files with identical content (even in different locations) are now skipped automatically

## Changes

- **mempalace/miner.py**: Added BloomFilter and ContentHashDB classes. Updated mine() to use content hash checking instead of just path-based checking.
- **mempalace/convo_miner.py**: Updated to use the same hash database for conversation file mining.
- **tests/test_hashdb.py**: Comprehensive tests for bloom filter and content hash DB functionality.

## Why this matters

Previously, mempalace mine would re-read files even if they had not changed. Now it:
- Computes SHA256 hash of file content
- Uses bloom filter for instant "definitely new" / "might be duplicate" checks
- Only re-reads files when content hash is genuinely new

This makes re-mining projects much faster and avoids redundant I/O.

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -11,6 +11,8 @@ Same palace as project mining. Different ingest strategy.
 import os
 import sys
 import hashlib
+import json
+import math
 from pathlib import Path
 from datetime import datetime
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


daedalus:
This will prevent wasting precious CPU/GPU cycles.

bensig:
Interesting approach but this conflicts with main and the duplicate detection story is evolving with igorls' work in #140. Closing — feel free to reopen with a rebased version if you'd like to revisit.

PR #130, opened 2026-04-07
fix: prevent concurrent mine corruption, OOM crashes, and Ctrl+C corruption

minimexat:
## What

Three crash fixes for stability issues discovered during real-world mining.

### Concurrent mine lock
Running multiple `mempalace mine` processes simultaneously causes a ChromaDB file locking violation and silent palace corruption. Added palace-level lockfile using `fcntl` (Unix) / `msvcrt` (Windows) with a clear error message instead of silent corruption.

### OOM batch cap
Large files (JS bundles, SQL dumps) can generate hundreds of chunks submitted in a single ChromaDB `collection.add()` call, causing `bad allocation` crashes on machines with limited RAM. Added `BATCH_SIZE = 50` cap and `add_drawers_batch()` helper.

### Graceful Ctrl+C
Cancelling mid-run with Ctrl+C corrupts the palace. Added `SIGINT` handler that sets an `interrupted` flag, finishes the current file cleanly, then stops. Already-mined files are skipped on re-run, so progress is preserved.

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -15,7 +15,31 @@ from pathlib import Path
 from datetime import datetime
 from collections import defaultdict
 
+import fcntl
+import signal
+
 import chromadb
+
+def _acquire_palace_lock(palace_path: str):
+    """Acquire exclusive lock on palace. Exit with clear message if already locked."""
+    lock_path = Path(palace_path) / ".mine.lock"
+    lock_file = open(lock_path, "w")
+    try:
+        if sys.platform == "win32":
+            import msvcrt
+            msvcrt.locking(lock_file.fileno(), msvcrt.LK_NBLCK, 1)
+        else:
+            fcntl.flock(lock_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
+    except (IOError, OSError):
+        print("\nERROR: Another mine is already running on this palace.")
+        print("Wait for it to finish, or delete .mempalace/palace/.mine.lock if it crashed.\n")
+        sys.exit(1)
+    return lock_file
+
+def _release_palace_lock(lock_file):
+    lock_file.close()
+    Path(lock_file.name).unlink(missing_ok=True)
+
 
 READABLE_EXTENSIONS = {
     ".txt",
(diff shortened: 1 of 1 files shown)


travisbreaks (commented):
Thanks for tackling these three real problems (concurrent mine corruption, OOM on large files, Ctrl+C corruption). The approach is sound overall. A few items to consider:

### 1. `fcntl` import at module top level breaks Windows

```python
import fcntl  # line 18 — unconditional import
```

`fcntl` doesn't exist on Windows. The conditional `msvcrt` usage inside `_acquire_palace_lock()` is correct, but the module-level `import fcntl` will crash on import before it ever gets there. This should be a conditional import:

```python
# At module level:
if sys.platform != "win32":
    import fcntl
```

Or move the import inside the function body.

### 2. Stale lock after crash/SIGKILL

If the process is killed (SIGKILL, power loss, OOM-killer), `.mine.lock` will be orphaned and all future `mine` invocations will fail with "Another mine is already running." The error message says "delete .mine.lock if it crashed" which is good guidance, but a stale-lock detection (e.g. write PID to the lock file, check if that PID is alive on next acquire) would prevent users from hitting this in production.

### 3. Hardcoded `BATCH_SIZE = 50`

50 is a reasonable default, but the sweet spot depends on document size and available RAM. Consider making this configurable via `--batch-size` CLI flag or `mempalace.yaml` config, with 50 as the default.

### 4. `add_drawers_batch` silently swallows duplicates

```python
if "already exists" not in str(e).lower():
    raise
```

This silently skips duplicate entries. That's probably fine for idempotent re-mining, but a `logger.debug()` here would help users understand why drawer counts don't match expectations.

### 5. Signal handler scope

`signal.signal(signal.SIGINT, _handle_interrupt)` is set inside `mine()`. If `mine()` is called as a library function (not from CLI), this overwrites the caller's signal handler without restoring it. Consider saving/restoring the original handler in the `finally` block:

```python
old_handler = signal.signal(signal.SIGINT, _handle_interrupt)
try:
    ...
finally:
    signal.signal(signal.SIGINT, old_handler)
    if lock is not None:
        _release_palace_lock(lock)
```

### 6. No tests

CONTRIBUTING.md says "Write code with tests" and "All tests must pass via `pytest tests/ -v`". The existing `test_miner.py` has 15 tests, but none cover concurrency or signal handling. Even basic tests would strengthen this:
- Lock acquisition/release lifecycle
- Double-lock returns error
- Batch splitting with >50 items

---

None of these are blockers for the concept, but items 1 (Windows import crash) and 5 (signal handler leak) are bugs that should be fixed before merge. The rest are improvements worth considering.

Overall solid work addressing real pain points.

bensig:
This conflicts with main and several of the fixes here (OOM, corruption guards) have been addressed in #387 and #399. If the concurrent mining lock is still needed, a focused PR for just that piece would be welcome.

PR #148, opened 2026-04-07
feat: support XDG Base Directory Specification (closes #46)

### LINKED ISSUE #46: Feature request: Support XDG base directory for configuration
**What problem does this solve?**

Currently, mempalace stores its configuration in `~/.memepalace`, which does not follow the XDG Base Directory Specification commonly used on Linux systems.

This leads to:
- Inconsistency with other CLI tools
- Unnecessary clutter in the home directory
- Friction for users managing dotfiles or using standardized setups

Many users expect configuration files to live under `$XDG_CONFIG_HOME` (usually `~/.config`).

**What's the proposed solution?**

Adopt the XDG Base Directory Specification for configuration storage:

- Default path:
  - `$XDG_CONFIG_HOME/mempalace`
- Fallback if `XDG_CONFIG_HOME` is not set:
  - `~/.config/mempalace`

For backward compatibility:
- If `~/.mempalace` exists, continue to support it
- Optionally provide:
  - A warning encouraging migration
  - Automatic migration or a CLI command to migrate config

**Alternatives considered:**

- Keeping the current behavior (`~/.memepalace`), but this goes against widely adopted conventions
- Supporting both paths without migration, which may lead to confusion about which config is actually used

mvalentsev:
## What does this PR do?

Closes #46 by moving the default config directory to follow the XDG Base Directory Specification, with a careful back-compat path so existing installs keep working.

Resolution order:
1. `$MEMPALACE_CONFIG_DIR` explicit override
2. `~/.mempalace` if it contains `config.json`, `people_map.json`, or `palace/chroma.sqlite3` (back-compat)
3. `$XDG_CONFIG_HOME/mempalace` (absolute paths only, per spec)
4. `~/.config/mempalace` fallback

Edge cases handled per XDG spec: empty, whitespace, and relative `XDG_CONFIG_HOME` values are ignored. An empty `~/.mempalace` left behind by some other tool no longer hijacks the XDG fallback.

`palace_path` now defaults to `<config_dir>/palace` so the data directory stays next to the config wherever XDG places it. The legacy `DEFAULT_PALACE_PATH` module constant is kept for any external importers but marked deprecated.

## How to test

```bash
python -m pytest tests/test_config.py -v
ruff check .
ruff format --check .
```

The test suite adds coverage for all four resolution branches, precedence between env override and legacy, the empty-legacy-dir edge case, and empty / whitespace / relative XDG values.

Manual smoke:
```bash
# New XDG path
XDG_CONFIG_HOME=/tmp/xdg_test mempalace init /tmp/some_project
ls /tmp/xdg_test/mempalace/

# Legacy back-compat: existing ~/.mempalace installs get the old location.
```

## Checklist

- [x] Tests pass (`python -m pytest tests/ -v` — 5153 passed, 53 skipped on the CI selection, rebased onto current `develop`)
- [x] No hardcoded paths (legacy constant marked deprecated)
- [x] Python 3.9 compatible

## Scope note

This PR is scoped to `MempalaceConfig` and its direct callers in line with the issue text (which asks about configuration storage). With this branch applied, a repo-wide grep for a `~/.mempalace` assembled in code still finds 17 modules building the legacy path directly: `closet_llm.py`, `daemon.py`, `dedup.py`, `diary_ingest.py`, `entity_registry.py`, `hallways.py`, `hooks_cli.py`, `knowledge_graph.py`, `layers.py`, `mcp_server/tools_diary.py` (hook_state), `miner.py`, `onboarding.py`, `palace.py`, `palace_graph.py`, `repair.py`, `server_registry.py`, and `update_awareness.py`. Those can be migrated in a follow-up PR so this one stays reviewable.

### DIFF
--- a/mempalace/cli/parser.py
+++ b/mempalace/cli/parser.py
@@ -52,7 +52,7 @@ def main():
     parser.add_argument(
         "--palace",
         default=None,
-        help="Where the palace lives (default: from ~/.mempalace/config.json or ~/.mempalace/palace)",
+        help="Where the palace lives (default: palace_path from config.json, resolved via XDG — see mempalace.config)",
     )
     parser.add_argument(
         "--backend",
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -1,7 +1,16 @@
 """
 MemPalace configuration system.
 
-Priority: env vars > config file (~/.mempalace/config.json) > defaults
+Priority: env vars > config file > defaults.
+
+The default config directory follows the XDG Base Directory Specification
+(https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+with back-compat for existing installs:
+
+  1. $MEMPALACE_CONFIG_DIR if set (explicit override, used by tests and CI)
+  2. ~/.mempalace if it already exists (existing installs keep working)
+  3. $XDG_CONFIG_HOME/mempalace if XDG_CONFIG_HOME is set
+  4. ~/.config/mempalace (XDG default fallback)
 """
 
 import errno
@@ -224,6 +233,51 @@ def sanitize_content(value: str, max_length: int = 100_000) -> str:
     return strip_lone_surrogates(value)
 
 
+# ── XDG Base Directory resolution ─────────────────────────────────────────────
+
+# Files that mark a ~/.mempalace directory as a real legacy install rather
+# than an empty/leftover directory. Any one is enough for back-compat.
+_LEGACY_MARKERS = ("config.json", "people_map.json")
+
+
+def _has_legacy_install(legacy: Path) -> bool:
+    """Return True if ``legacy`` is a real legacy install, not an empty dir.
+
+    A bare ``palace`` directory created by some other tool should not hijack
+    the XDG path, so the palace sub-directory only counts when it contains
+    an actual ChromaDB store (``chroma.sqlite3``).
+    """
+    if any((legacy / marker).is_file() for marker in _LEGACY_MARKERS):
+        return True
+    return (legacy / "palace" / "chroma.sqlite3").is_file()
+
+
+def _default_config_dir() -> Path:
+    """Return the default config directory.
+
+    See the module docstring for the resolution order.
+    """
+    env_dir = os.environ.get("MEMPALACE_CONFIG_DIR")
+    if env_dir and env_dir.strip():
+        return Path(env_dir).expanduser()
+
+    legacy = Path.home() / ".mempalace"
+    if legacy.is_dir() and _has_legacy_install(legacy):
+        return legacy
+
+    xdg = os.environ.get("XDG_CONFIG_HOME")
+    if xdg and xdg.strip():
+        xdg_path = Path(xdg).expanduser()
+        # Per XDG spec, relative paths must be ignored as invalid.
+        if xdg_path.is_absolute():
+            return xdg_path / "mempalace"
+
+    return Path.home() / ".config" / "mempalace"
+
+
+# DEPRECATED: kept only for backward compatibility with external importers.
+# This constant is frozen at the legacy location and is NOT XDG-aware — prefer
+# MempalaceConfig().palace_path in new code.
 DEFAULT_PALACE_PATH = os.path.expanduser("~/.mempalace/palace")
 DEFAULT_COLLECTION_NAME = "mempalace_drawers"
 DEFAULT_BACKEND = "chroma"
@@ -662,14 +716,13 @@ class MempalaceConfig:
 
         Args:
             config_dir: Override config directory (useful for testing).
-                        Defaults to ~/.mempalace.
+                        Defaults to the XDG-aware base directory — see
+                        _default_config_dir() for the resolution order.
             palace_path: Explicit palace data directory. This is primarily
                          used by CLI operations that received ``--palace``;
                          it takes precedence over environment and file config.
         """
-        self._config_dir = (
-            Path(config_dir) if config_dir else Path(os.path.expanduser("~/.mempalace"))
-        )
+        self._config_dir = Path(config_dir).expanduser() if config_dir else _default_config_dir()
         self._config_file = self._config_dir / "config.json"
         self._people_map_file = self._config_dir / "people_map.json"
         self._palace_path_override = (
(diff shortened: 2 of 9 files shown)


milla-jovovich:
XDG support done right — the back-compat for existing ~/.mempalace installs via marker files is the thoughtful touch. nice work

web3guru888 (commented):
## Review: XDG Base Directory Specification Support

Clean, focused PR that closes a long-standing request. The resolution order is well-documented and the back-compat story is solid.

### What's done well

**The resolution order is correct and clearly documented.** `MEMPALACE_CONFIG_DIR` > legacy `~/.mempalace` (if populated) > `$XDG_CONFIG_HOME/mempalace` > `~/.config/mempalace` is exactly the right priority. The "legacy if populated" check using `_LEGACY_MARKERS` prevents false positives from empty leftover directories.

**No migration required for existing users.** Existing installs with `~/.mempalace/config.json` or `~/.mempalace/people_map.json` transparently keep using that path. New installs get XDG. This is the smoothest possible transition.

**`test_config.py` monkeypatch cleanup.** Changing the `test_env_override` test to use `monkeypatch.setenv` instead of `os.environ["..."] = ...` + manual delete is a genuine correctness fix — the old version leaked state if the test raised.

**`_set_home` helper correctly sets both `HOME` and `USERPROFILE`.** Good cross-platform hygiene.

### Issues found

**`$XDG_DATA_HOME` and `$XDG_CACHE_HOME` are not addressed.** The XDG spec defines three variables: `XDG_DATA_HOME` (default `~/.local/share`), `XDG_CONFIG_HOME` (default `~/.config`), and `XDG_CACHE_HOME` (default `~/.cache`). This PR only handles `XDG_CONFIG_HOME`. If MemPalace's palace data (the ChromaDB/LanceDB files and the SQLite KG) lives in the config dir, that's fine — but if these are meant to be user data, they probably belong in `XDG_DATA_HOME`. Worth at least a comment in `config.py` explaining why `XDG_DATA_HOME` is out of scope.

**Windows fallback.** On Windows, `$XDG_CONFIG_HOME` is almost never set, and `~/.config` doesn't exist by convention. The fallback silently creates `~/.config/mempalace` on Windows, which is unexpected. The XDG spec says "if not set or empty, use the default", but on Windows the conventional path is `%APPDATA%\mempalace`. Consider: `if sys.platform == "win32" and not os.environ.get("XDG_CONFIG_HOME") and not legacy.exists(): return Path(os.environ.get("APPDATA", Path.home())) / "mempalace"`.

**`_LEGACY_MARKERS` doesn't include `palace/` as a directory.** The code checks `(legacy / "palace").exists()` but if someone has a palace directory but no config files, the check passes the `any(...)` guard. Wait — it does include `"palace"` in the tuple. But it uses `.exists()` which returns True for both files and directories. A bare `palace` directory created by some other tool would trigger the legacy path. Consider checking `(legacy / "palace").is_dir()` or `(legacy / "palace" / "chroma.sqlite3").exists()` for a tighter guard.

**No test for Windows `APPDATA` fallback** (if you add it). But this is a non-blocker.

**`_default_config_dir()` is not cached.** It calls `Path.home()` and stat checks on every call. In our integration, config resolution happens on every search and mine call. Wrapping with `functools.lru_cache(maxsize=1)` would eliminate repeated syscalls.

### Suggestions

1. Add a comment explaining the `XDG_DATA_HOME`/`XDG_CACHE_HOME` scope decision
2. Add a Windows `APPDATA` fallback or at least a warning
3. Tighten the `"palace"` marker check to `.is_dir()` rather than `.exists()`
4. Cache `_default_config_dir()` with `lru_cache` or a module-level lazy singleton

### Overall

Solid, well-scoped implementation. The XDG compliance covers the 90% case (Linux/macOS users with XDG set) and doesn't break anyone. The issues above are mostly edge cases and polish. The tests are well-structured and use monkeypatch correctly throughout.

**APPROVED** — issues are non-blocking, happy to see this land.

---

mvalentsev:
Tightened the legacy palace marker as suggested. `_LEGACY_MARKERS` no longer includes the bare `palace` name -- the check is now in a `_has_legacy_install` helper that requires `palace/chroma.sqlite3` to exist before treating the directory as a real install. Added two tests: one for the bare-dir case (falls through to XDG) and one for the palace-with-ChromaDB case (wins over XDG).

The other points (XDG_DATA_HOME / XDG_CACHE_HOME scope, Windows APPDATA fallback, lru_cache on `_default_config_dir`) are reasonable as follow-ups. This PR is scoped to the XDG_CONFIG_HOME branch per the issue text; the data-home split and platform-specific Windows path deserve their own PRs so the diff stays reviewable.

web3guru888:
Well-designed implementation. The resolution order is correct and the `_has_legacy_install()` guard is the key detail that makes this safe to merge.

**`_has_legacy_install()` refinement** is the right call. The original PR would have treated any empty `~/.mempalace` directory as a legacy install, which means tools that create that directory (Docker volumes, dotfile managers) could silently prevent XDG adoption. Requiring `config.json`, `people_map.json`, or `palace/chroma.sqlite3` is tight enough to distinguish a real install from an empty dir.

**`palace_path` default now derives from `_config_dir`** (`str(self._config_dir / "palace")`) rather than hardcoding `DEFAULT_PALACE_PATH`. This is the important correctness fix — without it, a fresh install on a clean machine would create `~/.config/mempalace/config.json` but point the palace at `~/.mempalace/palace`, which is a confusing split state.

**Test coverage** is thorough. The edge cases covered (empty dir, bare palace dir without chroma.sqlite3, relative XDG path, whitespace XDG value) are exactly the ones that would cause silent failures in production.

A few observations:
- `DEFAULT_PALACE_PATH` is kept with a deprecation comment but is frozen at the legacy location. Any external tool that reads it programmatically will silently get the wrong path on a clean XDG install. Consider a `DeprecationWarning` when it's imported or accessed — though that might be too noisy. Fine as-is for this PR's scope.
- `_default_config_dir()` is not cached (`lru_cache`). On a server or test-heavy environment, this gets called for every `MempalaceConfig()` instantiation, which does a `Path.home()` + `is_dir()` + `is_file()` stat chain. Probably not a hot path in practice, but worth noting.

Both points are reasonable follow-ups. This PR is scoped correctly. LGTM.

mvalentsev:
Heads-up: 4f2deda extends the XDG resolution to two callers that still
hardcoded `~/.mempalace`:

- `mempalace/mcp_server.py:_WAL_DIR` was `Path(os.path.expanduser("~/.mempalace/wal"))`,
  so a user with `MEMPALACE_CONFIG_DIR=~/.config/mempalace` set got the
  WAL written to the legacy path while config went to XDG. Now resolves
  through `_config.config_dir / "wal"`.
- `mempalace/fact_checker.py` argparse `--palace` default was
  `os.path.expanduser("~/.mempalace/palace")`. Now resolves through
  `MempalaceConfig().palace_path` via a `_default_palace_path()` helper.

Adds a public `MempalaceConfig.config_dir` property so callers outside
`config.py` can derive sibling locations without reaching into
`_config_dir`. Three new tests pin the holes:

- `test_config_dir_property_exposes_resolved_path`
- `test_default_palace_resolves_via_mempalace_config`
- `test_wal_dir_nests_under_config_dir`

igorls (approved):
Reviewed against current develop. Resolution order is correct and the _has_legacy_install() guard is the load-bearing detail: empty ~/.mempalace and a bare palace/ dir do not hijack XDG, while config.json / people_map.json / a real chroma.sqlite3 keep existing installs in place. WAL and fact-checker --palace now follow the same config dir. Tests cover env override, empty/relative XDG_CONFIG_HOME, and the Chroma marker. DEFAULT_PALACE_PATH stays frozen at the legacy location for external importers, as documented. Squash-merging. Closes #46.

PR #150, opened 2026-04-07
test(dialect): strengthen compression_stats and count_tokens assertions

mvalentsev:
## What does this PR do?

The main honest-stats test fix (`ratio` -> `size_ratio`, `compressed_chars` -> `summary_chars`, `len // 3` -> word-based `count_tokens`) already landed in main through @igorls' PR series. This PR now just adds a couple of assertions on top of that fix so the regression surface is a bit smaller.

- `test_stats` also asserts `stats["original_tokens_est"] > stats["summary_tokens_est"]`. That catches a class of regressions where the token estimator flattens to a constant, which would otherwise slip through the existing char-level assertion.
- `test_count_tokens` gains edge cases for the empty string and the single-word input. Both exercise the `max(1, ...)` guard in `Dialect.count_tokens`, so a future refactor that drops the guard fails loudly instead of silently returning 0.

Originally opened before I saw that the core rename fix was already in flight. After rebasing on the merged fix, the remaining diff is a small additive coverage improvement.

## How to test

```bash
python -m pytest tests/test_dialect.py -v
ruff check tests/test_dialect.py
ruff format --check tests/test_dialect.py
```

18 tests pass locally.

## Checklist

- [x] Tests pass
- [x] Linter passes
- [x] Python 3.9 compatible

### DIFF
--- a/tests/test_dialect.py
+++ b/tests/test_dialect.py
@@ -111,9 +111,14 @@ class TestCompressionStats:
         stats = d.compression_stats(original, compressed)
         assert stats["size_ratio"] > 1
         assert stats["original_chars"] > stats["summary_chars"]
+        assert stats["original_tokens_est"] > stats["summary_tokens_est"]
 
     def test_count_tokens(self):
+        # count_tokens uses a word-based heuristic (~1.3 tokens per word).
+        # "hello world" is 2 words -> max(1, int(2 * 1.3)) == 2.
         assert Dialect.count_tokens("hello world") == 2
+        assert Dialect.count_tokens("") == 1
+        assert Dialect.count_tokens("one") == 1
 
     def test_compression_stats_keys(self):
         """Verify compression_stats() returns the expected key set."""


mvalentsev:
Update: the core rename fix landed in main via @igorls' PR series, so this is no longer a CI blocker. Rebased on the merged fix, and the remaining diff is just two extra assertions — a strictness check on `original_tokens_est > summary_tokens_est` in `test_stats`, and empty / single-word edge cases for the `max(1, ...)` guard in `test_count_tokens`. Retitled and updated the description to reflect the new scope.

web3guru888 (commented):
**PR #150 — test(dialect): strengthen compression_stats and count_tokens assertions**

This is a small, focused test hardening PR that adds meaningful coverage on top of the existing assertion suite. The context in the PR description is helpful: the `ratio → size_ratio` / `compressed_chars → summary_chars` rename landed via @igorls' series, and this PR bolts on the regression guards that should have accompanied that rename from the start.

**What's done well:**

The `original_tokens_est > summary_tokens_est` assertion in `test_stats` is the right catch. The risk it guards against — a token estimator that collapses to a constant (e.g. always returns 1 due to a bad `max()` guard or wrong heuristic multiplier) — would be completely invisible to the existing char-level assertion. It's a silent failure mode that could persist through several refactor cycles without this.

The three `count_tokens` edge cases (`""`, `"one"`, `"hello world"`) directly exercise the `max(1, int(word_count * 1.3))` contract. The inline comment explaining the `"hello world"` → 2 calculation is a nice touch — it makes the expected value self-documenting for future readers rather than leaving them to mentally trace the heuristic. Not every test PR does this.

**Issues / concerns:**

One minor flakiness risk: `test_stats` creates a compression pair and then asserts `original_tokens_est > summary_tokens_est`. If `Dialect.compress()` ever produces a summary longer than the original (degenerate case on very short inputs — e.g., a 2-word input that gets a 5-word summary header prepended), the assertion fails for the wrong reason. It might be worth using a clearly long fixture string (50+ words) so the compressed form is unambiguously shorter. The current fixture text is not shown in the diff, so I can't verify it's safe — worth double-checking.

The `count_tokens("") == 1` case is correct per the `max(1, ...)` guard, but it implicitly documents that empty strings always return 1. If the semantics ever change (e.g., returning 0 for empty input becomes intentional), the test will catch it — but it might be surprising. A one-line comment explaining "empty string returns minimum 1 by design" would clarify the intent vs. a floor that was accidentally left in.

**Overall verdict:**

LGTM — this is exactly the kind of regression guard that makes a codebase durable through refactors. The token estimator path is load-bearing for anything that uses `compression_stats` to make throttling or batching decisions, so having explicit assertions on it is worthwhile. The edge-case coverage on `count_tokens` is appropriate for a `max(1, ...)` guard. Minor suggestion: verify the fixture text in `test_stats` is long enough that `summary_tokens_est < original_tokens_est` is guaranteed, not probabilistic.

---

web3guru888:
Small but correct. The two new assertions are the right additions:

- `assert stats["original_tokens_est"] > stats["summary_tokens_est"]` catches the "flattened estimator" regression class (where token counting returns a constant regardless of input). The `size_ratio > 1` assertion above it doesn't catch this because `compression_stats()` computes ratio from char counts, not token counts.
- `count_tokens("")` == 1 and `count_tokens("one")` == 1 pin the `max(1, ...)` floor behavior explicitly — useful to have documented in the test so the floor semantics don't get removed accidentally.

The comment explaining the word-based heuristic (`~1.3 tokens per word`) is a helpful clarification for contributors who might otherwise wonder why a 2-word string returns 2 tokens instead of the expected `ceil(2 * 1.3) = 3`.

LGTM.

bensig:
closing — dialect test improvements landed via #569 (merged). thanks @mvalentsev!

PR #154, opened 2026-04-07
feat: honour .mempalace-ignore during mining (closes #102)

### LINKED ISSUE #102: Need some soft of .mempalace-ignore funtionality
Doing some initial experimenting with with Mempalace and mine is scooping up all sorts of junk files that it it shouldn't...

Two feature requests:
1. Honor existing .gitignore files 
2. Add some sort of .mempalace-ignore functionality

mvalentsev:
## What does this PR do?

Closes #102 by honouring a per-directory `.mempalace-ignore` file during mining. Same syntax as `.gitignore`, but it applies even when the caller passes `respect_gitignore=False`, so users can keep mining-specific exclusions (credentials, drafts, large data files) outside their project's git config.

`GitignoreMatcher.from_dir` takes a `filename` keyword argument (defaults to `.gitignore`) and `load_gitignore_matcher` keys its per-walk cache by `(directory, filename)` so both ignore files can coexist without stepping on each other. In `scan_project` the loader is called twice per directory -- `.gitignore` first (only when `respect_gitignore` is on), `.mempalace-ignore` always -- and they're appended in that order so mempalace-ignore has the final say when the two disagree.

Note on the previously-closed #144: that PR bundled this feature with an unrelated telemetry fix and was closed for scope. This PR implements only the `.mempalace-ignore` half as a focused change. The hyphenated spelling matches the issue title; a `.mempalaceignore` alias can be added in a follow-up if useful.

## How to test

```bash
python -m pytest tests/test_miner.py -v
ruff check .
ruff format --check .
```

Manual smoke:

```bash
cd /tmp && mkdir ignore_demo && cd ignore_demo
printf 'wing: demo\nrooms:\n  - {name: general}\n' > mempalace.yaml
printf 'secrets.py\ndrafts/\n' > .mempalace-ignore
printf 'print("keep me")\n' > main.py
printf 'print("secret")\n' > secrets.py
mkdir drafts && printf 'print("draft")\n' > drafts/note.py
mempalace mine .
# only main.py is filed; secrets.py and drafts/note.py are skipped
```

## Checklist

- [x] Tests pass (`python -m pytest tests/test_miner.py -v` - 23 passed)
- [x] No hardcoded paths
- [x] Python 3.9 compatible

## Coverage added

- `test_scan_project_respects_mempalace_ignore` - base case: file + directory patterns
- `test_scan_project_mempalace_ignore_applies_without_gitignore` - still honoured with `respect_gitignore=False`
- `test_scan_project_mempalace_ignore_overrides_gitignore_negation` - precedence over `.gitignore` `!foo`
- `test_scan_project_mempalace_ignore_not_self_included` - the ignore file itself is never mined
- `test_scan_project_nested_mempalace_ignore` - nested directory support

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -150,10 +150,17 @@ SKIP_FILENAMES = {
     "mempal.yaml",
     "mempal.yml",
     ".gitignore",
+    ".mempalace-ignore",
     "package-lock.json",
     "pnpm-lock.yaml",
     "yarn.lock",
 }
+
+# Per-directory mempalace-specific ignore file. Same syntax as .gitignore but
+# honoured independently so users can exclude files from mining without
+# touching their project's git config, and it still applies when
+# --no-gitignore is passed.
+MEMPALACE_IGNORE_FILENAME = ".mempalace-ignore"
 
 # Re-export the shared defaults from ``config`` so legacy callers that
 # import ``CHUNK_SIZE`` / ``CHUNK_OVERLAP`` / ``MIN_CHUNK_SIZE`` from
@@ -238,7 +245,7 @@ def _resolve_max_chunks_per_file(override: Optional[int] = None) -> int:
 
 
 class GitignoreMatcher:
-    """Lightweight matcher for one directory's .gitignore patterns."""
+    """Lightweight matcher for one directory's gitignore-style ignore file."""
 
     def __init__(self, base_dir: Path, rules: list):
         self.base_dir = base_dir
@@ -288,8 +295,8 @@ class GitignoreMatcher:
         return rules
 
     @classmethod
-    def from_dir(cls, dir_path: Path):
-        gitignore_path = dir_path / ".gitignore"
+    def from_dir(cls, dir_path: Path, filename: str = ".gitignore"):
+        gitignore_path = dir_path / filename
         if not gitignore_path.is_file():
             return None
 
@@ -383,15 +390,16 @@ class GitignoreMatcher:
         return matches(0, 0)
 
 
-def load_gitignore_matcher(dir_path: Path, cache: dict):
-    """Load and cache one directory's .gitignore matcher."""
-    if dir_path not in cache:
-        cache[dir_path] = GitignoreMatcher.from_dir(dir_path)
-    return cache[dir_path]
+def load_gitignore_matcher(dir_path: Path, cache: dict, filename: str = ".gitignore"):
+    """Load and cache a gitignore-style matcher for ``filename`` in ``dir_path``."""
+    key = (dir_path, filename)
+    if key not in cache:
+        cache[key] = GitignoreMatcher.from_dir(dir_path, filename=filename)
+    return cache[key]
 
 
 def is_gitignored(path: Path, matchers: list, is_dir: bool = False) -> bool:
-    """Apply active .gitignore matchers in ancestor order; last match wins."""
+    """Apply active ignore matchers in ancestor order; last match wins."""
     ignored = False
     for matcher in matchers:
         decision = matcher.matches(path, is_dir=is_dir)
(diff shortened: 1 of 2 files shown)


proxysoul (approved):
LGTM

web3guru888 (commented):
## Review: `.mempalace-ignore` During Mining

Exactly the right feature. We've been routing around this manually by abusing `.gitignore` entries. Having a mining-specific exclude file that works even with `respect_gitignore=False` is a clean solution.

### What's done well

**Reuses `GitignoreMatcher`.** Rather than building a parallel ignore-file parser, this PR makes `GitignoreMatcher.from_dir()` accept a `filename` param. That's the right call — the gitignore syntax is well-understood and reusing the existing parser means `.mempalace-ignore` gets negation patterns, directory globs, and `**` for free.

**`.mempalace-ignore` always honoured, even with `--no-gitignore`.** The commit message says this clearly and the test `test_scan_project_mempalace_ignore_applies_without_gitignore` covers it. This is the key design decision and it's correct — mining exclusions shouldn't be tied to git policy.

**Adding `.mempalace-ignore` to the static `SKIP_PATTERNS` set** prevents the ignore file itself from being mined as content. Good attention to detail.

**Two solid integration tests** covering the main cases. The `scanned_files()` helper in the test makes assertions readable.

### Issues found

**Negation patterns (`!` prefix) interaction with `SKIP_PATTERNS`.** If a user writes `!secrets.py` in their `.mempalace-ignore` to un-exclude something, that negation works fine inside `GitignoreMatcher`. But if `secrets.py` is also in the static `SKIP_PATTERNS` list, the negation will be silently ignored. The PR doesn't address this precedence, and it may be surprising. At minimum, a comment or doc note.

**No test for negation patterns.** Something like `*.log\n!important.log` is a common gitignore pattern. A test confirming `important.log` is still scanned would give confidence the `GitignoreMatcher` negation path works end-to-end with the new filename.

**No test for subdirectory `.mempalace-ignore` files.** The existing gitignore logic handles per-directory ignore files; does the `.mempalace-ignore` loading also walk subdirectories? The diff doesn't show the walk loop changes clearly enough to confirm. If subdirectory `.mempalace-ignore` files aren't loaded, that should be documented as a known limitation.

**`GitignoreMatcher.from_dir()` now has a `filename` default of `".gitignore"`.** That's the right default for back-compat, but callers inside the mine loop need to call it twice — once for `.gitignore` and once for `.mempalace-ignore`. Worth confirming in the test that both are active simultaneously, not just one or the other.

**Pattern performance on large directories.** `GitignoreMatcher` uses per-file `matches()` calls. For deeply nested directories with 50K+ files (our largest mine corpus), this is O(files × patterns). The existing gitignore path already has this cost, so it's not new, but adding a second pattern file doubles it. Not a blocker but worth noting for large-corpus users.

### Suggestions

1. Add a negation-pattern test: `!important.log` un-excludes within `.mempalace-ignore`
2. Add a test confirming subdirectory `.mempalace-ignore` files are respected (or document they're not)
3. Add a test with both `.gitignore` and `.mempalace-ignore` active simultaneously
4. Document that `!` negation can't override `SKIP_PATTERNS` (those bypass pattern matching entirely)

### Overall

Well-implemented feature that fills a real need. The two tests cover the happy paths and the critical `--no-gitignore` edge case. The issues above are mostly missing test coverage rather than implementation bugs.

**APPROVED** — the core implementation is correct. Test gaps are follow-up material.

---

web3guru888:
This is a well-executed change. The key architectural decision — `.mempalace-ignore` always applies, independent of `respect_gitignore` — is exactly right. Mining-specific exclusions (credentials, drafts, large data files) are orthogonal to git tracking decisions, and the current `--no-gitignore` flag should not affect them.

**Implementation details that are correct:**

- Cache key change from `dir_path` to `(dir_path, filename)` is required for correctness — without it, calling `load_gitignore_matcher` with `.mempalace-ignore` would overwrite the `.gitignore` entry in the cache for the same directory.
- Moving the `active_matchers` pruning (stale ancestor removal) *outside* the `respect_gitignore` block is correct — `.mempalace-ignore` matchers need the same ancestor pruning, otherwise matchers from parent directories could apply to unrelated subtrees after the walk moves past them.
- `.mempalace-ignore` added to `SKIP_FILES` so the ignore file itself never lands in the scan output. The `test_scan_project_mempalace_ignore_not_self_included` test explicitly covers this.

**Test coverage** is thorough. The `test_scan_project_mempalace_ignore_overrides_gitignore_negation` test is particularly valuable — it covers the interaction between `.gitignore` negation patterns (`!data/keep.csv`) and `.mempalace-ignore` exclusions, where the order-of-application matters.

One minor observation: the `active_matchers` list can grow unbounded in a deeply nested project because old matchers are pruned by ancestor check, but the append-then-prune ordering means you could have O(depth²) matchers in the worst case. Same as the existing `.gitignore` behavior, so not a new regression — just worth noting for large monorepos.

LGTM.

Issue #159, opened 2026-04-07
Bugs and improvements


cktang88:
Claude Code Review — MemPalace

---

## 1. `mempalace compress` is completely broken — crashes on every single run

`cli.py:303-304` accesses `stats["compressed_chars"]` and `cli.py:313` accesses `stats["original_tokens"]`, `stats["compressed_tokens"]`, and `stats["ratio"]`. None of these keys exist.

`dialect.py:960-967` returns keys named `"summary_chars"`, `"original_tokens_est"`, `"summary_tokens_est"`, and `"size_ratio"`. The refactor in commit `39c14be` renamed these keys in `compression_stats()` but never updated `cmd_compress()`.

**Impact:** `mempalace compress` (dry-run or real) throws `KeyError: 'compressed_chars'` on the very first drawer it processes. The entire compress feature is dead code that nobody has run since the stats refactor.

---

## 2. Compress summary token counts are hardcoded to 1

`cli.py:340-342`:
```python
orig_tokens = Dialect.count_tokens("x" * total_original)
comp_tokens = Dialect.count_tokens("x" * total_compressed)
```

`Dialect.count_tokens` (`dialect.py:948-949`) splits on whitespace and multiplies by 1.3. The string `"x" * 5000` has zero spaces — it's one giant word. `int(1 * 1.3)` = `1`. Both `orig_tokens` and `comp_tokens` are always `1`, regardless of actual content size.

**Impact:** Even if bug 1 were fixed, the compress summary line would always print `Total: 1t -> 1t (1.0x compression)`. Useless.

---

## 3. Miner room stats are wrong

`miner.py:624`:
```python
room = detect_room(filepath, "", rooms, project_path)
```

`process_file` (called on line 611) already calls `detect_room` internally with the actual file content to route the file to a room. Then line 624 calls `detect_room` **again** with an empty string for content. With no content, keyword scoring (`miner.py:328-337`) can't fire, so the fallback folder-path or filename matching is used instead. When routing relied on content keywords (Priority 3), the room printed in the summary is different from the room the file was actually filed into.

**Impact:** The "By room" summary at the end of every `mempalace mine` run can show wrong room counts. Users see one breakdown, the palace has a different one.

---

## 4. MCP server loads entire palace into RAM on every status/list call

`mcp_server.py:72`:
```python
all_meta = col.get(include=["metadatas"])["metadatas"]
```

[...]

PR #162, opened 2026-04-07
fix: restore mempalace compress after stats rename (#159)

mvalentsev:
## What does this PR do?

Fixes points 1 and 2 from #159. `mempalace compress` was dead after #147 renamed the `Dialect.compression_stats()` keys:

- `ratio` became `size_ratio`
- `compressed_chars` became `summary_chars`
- `original_tokens` / `compressed_tokens` became `original_tokens_est` / `summary_tokens_est`

`cmd_compress` still read all four old names, so the first drawer it touched raised `KeyError: 'compressed_chars'` and the whole command crashed before writing anything.

While in there, the summary line at the bottom of `cmd_compress` was also wrong in a subtler way. It tried to turn the char totals into token totals by calling `Dialect.count_tokens("x" * total_original)`. `count_tokens` is word-based (`max(1, int(len(text.split()) * 1.3))`), and a long string of repeated `x`s is a single "word", so both totals were always `1` and the user saw `Total: 1t -> 1t (1.0x compression)` regardless of the actual workload. The fix accumulates the per-drawer `original_tokens_est` / `summary_tokens_est` during the main loop and uses a token-based ratio so the summary is self-consistent with the per-drawer dry-run line.

The storage metadata keys on the `mempalace_compressed` collection (`compression_ratio`, `original_tokens`) keep their old names so anything already reading them still works. Only the source of the values is updated.

## How to test

```bash
# 1. A tiny palace to compress against
mkdir -p /tmp/compress_smoke && cd /tmp/compress_smoke
printf 'wing: test\nrooms:\n  - {name: general}\n' > mempalace.yaml
printf 'We decided to use Postgres because JSONB is a better fit. ' > content.md
printf '%s\n' "$(python -c 'print("alpha beta gamma delta epsilon " * 60)')" >> content.md
mempalace init .
mempalace mine .

# 2. Dry-run, previously crashed with KeyError on the first drawer
mempalace compress --dry-run
# Expected: per-drawer line shows positive token counts (not 1t),
# and Total: Nt -> Mt (Rx compression) with N, M, R all > 1

# 3. Real compress, writes the compressed collection
mempalace compress
# Expected: "Stored N compressed drawers in 'mempalace_compressed' collection."

# 4. Linter + existing tests
ruff check mempalace/cli.py
ruff format --check mempalace/cli.py
python -m pytest tests/ -v
```

## Checklist

- [x] No hardcoded paths
- [x] Python 3.9 compatible
- [x] Existing tests still pass (`pytest tests/test_miner.py tests/test_config.py tests/test_normalize.py tests/test_version_consistency.py -q` - 21 passed)

## Notes on testing scope

There are no unit tests for `cmd_compress` today and adding one would need a fake ChromaDB collection fixture that the repo does not have yet. The fix is verified end to end via the dry-run and real-compress walkthroughs above. Can follow up with a CLI test harness in a separate PR if useful.

Unrelated: `tests/test_dialect.py` has two pre-existing failing assertions from the same PR #147 rename (`test_stats` / `test_count_tokens`). Those are addressed in #150 and are not touched here to keep this PR focused.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -340,16 +340,16 @@ def cmd_compress(args):
     )
     print()
 
-    total_original = 0
-    total_compressed = 0
+    total_orig_tokens = 0
+    total_comp_tokens = 0
     compressed_entries = []
 
     for doc, meta, doc_id in zip(docs, metas, ids):
         compressed = dialect.compress(doc, metadata=meta)
         stats = dialect.compression_stats(doc, compressed)
 
-        total_original += stats["original_chars"]
-        total_compressed += stats["compressed_chars"]
+        total_orig_tokens += stats["original_tokens_est"]
+        total_comp_tokens += stats["summary_tokens_est"]
 
         compressed_entries.append((doc_id, compressed, meta, stats))
 
@@ -359,7 +359,8 @@ def cmd_compress(args):
             source = Path(meta.get("source_file", "?")).name
             print(f"  [{wing_name}/{room_name}] {source}")
             print(
-                f"    {stats['original_tokens']}t -> {stats['compressed_tokens']}t ({stats['ratio']:.1f}x)"
+                f"    {stats['original_tokens_est']}t -> {stats['summary_tokens_est']}t "
+                f"({stats['size_ratio']:.1f}x)"
             )
             print(f"    {compressed}")
             print()
@@ -370,8 +371,8 @@ def cmd_compress(args):
             comp_col = client.get_or_create_collection("mempalace_compressed")
             for doc_id, compressed, meta, stats in compressed_entries:
                 comp_meta = dict(meta)
-                comp_meta["compression_ratio"] = round(stats["ratio"], 1)
-                comp_meta["original_tokens"] = stats["original_tokens"]
+                comp_meta["compression_ratio"] = round(stats["size_ratio"], 1)
+                comp_meta["original_tokens"] = stats["original_tokens_est"]
                 comp_col.upsert(
                     ids=[doc_id],
                     documents=[compressed],
@@ -384,11 +385,9 @@ def cmd_compress(args):
             print(f"  Error storing compressed drawers: {e}")
             sys.exit(1)
 
-    # Summary
-    ratio = total_original / max(total_compressed, 1)
-    orig_tokens = Dialect.count_tokens("x" * total_original)
-    comp_tokens = Dialect.count_tokens("x" * total_compressed)
-    print(f"  Total: {orig_tokens:,}t -> {comp_tokens:,}t ({ratio:.1f}x compression)")
+    # Summary: token-based ratio stays consistent with the per-drawer line.
+    ratio = total_orig_tokens / max(total_comp_tokens, 1)
+    print(f"  Total: {total_orig_tokens:,}t -> {total_comp_tokens:,}t ({ratio:.1f}x compression)")
     if args.dry_run:
         print("  (dry run -- nothing stored)")
 
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -546,10 +546,10 @@ def test_cmd_compress_dry_run(mock_config_cls, capsys):
     mock_dialect.compress.return_value = "compressed"
     mock_dialect.compression_stats.return_value = {
         "original_chars": 100,
-        "compressed_chars": 30,
-        "original_tokens": 25,
-        "compressed_tokens": 8,
-        "ratio": 3.3,
+        "summary_chars": 30,
+        "original_tokens_est": 25,
+        "summary_tokens_est": 8,
+        "size_ratio": 3.3,
     }
     mock_dialect_mod = _make_mock_dialect_module(mock_dialect)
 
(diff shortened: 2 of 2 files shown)


web3guru888 (commented):
## Review: fix: restore mempalace compress after stats rename (#159)

This is the most thorough of the three PRs fixing the `cmd_compress` KeyError regression introduced when #147 renamed `Dialect.compression_stats()` keys. Here's my analysis comparing all three approaches.

### Background: Three PRs, One Bug

Three PRs address the same `KeyError: 'compressed_chars'` crash in `cmd_compress`:
- **#162** (this PR, @mvalentsev) — comprehensive fix with token accumulator refactor
- **#569** (@arnoldwender) — minimal key-swap only, adds a live `Dialect` integration test
- **#588** (@helhindi) — minimal key-swap only, updates test mocks only

All three correctly replace the four stale keys (`compressed_chars` → `summary_chars`, `original_tokens` → `original_tokens_est`, `compressed_tokens` → `summary_tokens_est`, `ratio` → `size_ratio`). The crash fix itself is identical in all three.

### What #162 Does Better

**Fixes a second, pre-existing bug.** The summary line at the end of `cmd_compress` was accumulating `total_original` and `total_compressed` from character counts (`original_chars`, `summary_chars`), then passing those char totals to `Dialect.count_tokens(\"x\" * total_original)` to compute token estimates. But `count_tokens` is word-based — a long string of repeated `x` characters is one "word", so the summary always printed `Total: 1t -> 1t (1.0x)` regardless of actual workload. This is a genuine silent correctness bug that #569 and #588 both miss entirely.

The fix — renaming accumulators to `total_orig_tokens` / `total_comp_tokens` and summing `original_tokens_est` / `summary_tokens_est` from per-drawer stats — is semantically correct and self-consistent with the per-drawer output lines. This is the right approach.

**Backward-compatible metadata keys.** The PR is explicit that `comp_meta["compression_ratio"]` and `comp_meta["original_tokens"]` keep their *old names* even though the values now come from the renamed stats fields. This is a thoughtful choice — downstream consumers reading the `mempalace_compressed` collection won't break.

### Issues and Suggestions

**Missing test coverage** — acknowledged by the author. The `cmd_compress` path has zero unit tests. #569 adds an integration test (`test_compression_stats_keys_match_dialect`) that calls the real `Dialect` class and asserts the key set. That test would immediately catch regressions like the one in #147. I'd strongly suggest cherry-picking or adapting that test for this PR, or filing a follow-up issue to track it. Shipping a fix without a regression guard leaves the door open for #147-style renames to slip through again.

**Summary ratio computation.** After the accumulator fix, the summary ratio is computed as `total_orig_tokens / total_comp_tokens`. This will `ZeroDivisionError` if `total_comp_tokens` is 0 (e.g., all drawers compressed to empty strings). Worth guarding: `ratio = total_orig_tokens / total_comp_tokens if total_comp_tokens else 0`.

**Pre-existing test failures flagged** — the author notes `test_stats` and `test_count_tokens` in `test_dialect.py` are already failing from #147 and are addressed in #150. Good disclosure; just confirming this PR doesn't touch those and won't make them worse.

### Diff Quality

The `cli.py` changes are clean. The accumulator rename from `total_original`/`total_compressed` to `total_orig_tokens`/`total_comp_tokens` improves readability — it's now obvious what unit is being tracked. The f-string split across two lines is also fine for 88-char line limits.

### Comparison Verdict

| Criterion | #162 (this PR) | #569 | #588 |
|---|---|---|---|
| Fixes KeyError crash | ✅ | ✅ | ✅ |
| Fixes broken summary totals | ✅ | ❌ | ❌ |
| Backward-compat metadata keys | ✅ explicit | ✅ implicit | ✅ implicit |
| Adds regression test | ❌ | ✅ integration test | ❌ |
| Test mock updates | ✅ | ✅ | ✅ |

**Recommendation: Merge #162 and close #569 and #588 as superseded** — but only after adding a regression guard for the zero-division edge case and ideally pulling in the integration test from #569. #162 is the only one that actually improves the correctness of the feature beyond just not crashing.

### Overall

APPROVE with minor suggestions. The core fix is correct, the accumulator refactor is genuinely better, and the backward-compatibility note on metadata keys shows careful thinking. Add a guard for the zero-division edge case and a regression test before merge.

---

mvalentsev:
Quick correction on the ZeroDivisionError flag -- the ratio line already uses `max(total_comp_tokens, 1)` as the denominator (cli.py:389), so the zero case is handled. The comparison with #569 and #588 is otherwise accurate.

web3guru888:
Clean fix. All four stale key references are updated consistently (`compressed_chars` → `summary_tokens_est`, `original_tokens` → `original_tokens_est`, `ratio` → `size_ratio`) and the test mocks are updated in both test cases.

The summary simplification is a nice bonus — the old code was converting token estimates back to a character count and then re-running `count_tokens()` on a synthetic string of that length, which was a lossy round-trip. Using the direct `original_tokens_est` / `summary_tokens_est` from `compression_stats()` is more accurate.

On the `ZeroDivisionError` flag @mvalentsev noted: confirmed — `max(total_comp_tokens, 1)` is already in the ratio line, so the zero case is handled. The PR is correct as written.

One minor observation: `comp_meta["original_tokens"] = stats["original_tokens_est"]` stores the value under the old key name. For palaces that already have compressed entries (created before this fix), the metadata key name will be inconsistent. Not a blocker for this PR — it's a cosmetic inconsistency — but worth noting for a follow-up migration note or a key rename in the metadata write.

LGTM otherwise.

mvalentsev:
The `comp_meta["original_tokens"]` key is deliberately kept under its old name for backward compat -- noted in the PR body. Renaming would break anything already reading the `mempalace_compressed` collection, so the mismatch between source field (`original_tokens_est`) and stored key (`original_tokens`) is intentional.

mvalentsev:
Closing -- #569 and #609 on develop together cover all three fixes this PR addressed: the stats key alignment, the backward-compat metadata key preservation, and the token summary bug.

One minor note for anyone touching the compress path later: the merged #609 estimates summary tokens via `chars / 3.8`, while the per-drawer lines still use `Dialect.compression_stats()` word-based estimates. Not a regression, just a small inconsistency between the two outputs. Easy follow-up if anyone cares about making them consistent.

PR #174, opened 2026-04-07
feat: add brief retrieval mode for wing/room overviews

### LINKED ISSUE #73: Feature idea: separate recall vs brief retrieval modes
Context windows are quite large here in 2026 and increasing+ getting cheaper, so users tend to be sloppy and start asking for new things within the same session. this would mitigate this. 

Love the layered retrieval concept (L0 etc). Been thinking about how it could generalize for multi-agent and even team/multi user - use cases.

The great insight from your layer system is that not every query should return the same shape of result. But in practice L2 and L3 are both "search with different limits," and L0/L1 are startup-time concerns that don't apply if MemPalace is used as a shared "service".

What if the retrieval API was simplified to two explicit modes?

recall(query) -> what you have now. Vector search, returns 2-5 raw drawers/results. For specific questions like "what did we decide about xyz?"
brief(wing) -> returns a cached, pre-synthesized summary of a wing. Regenerated when new drawers are added. For "catch me up on everything about xyz."
brief is basically your L1 idea but scoped per-wing instead of global top and generated lazily instead of at wake-up. The AAAK system is somewhat close to this - if it works in the long run. something like the brief (or maybe "catchup") would just make it a first-class retrieval tool rather than a storage-side compression/fire-and-forget feature.

This would also make the MCP interface cleaner and potentially eliminate much of the "wake-up ritual" you/an agent just call recall or brief depending on what is needed.

mvanhorn:
## What does this PR do?

Adds a `brief()` retrieval mode that returns a deduplicated, truncated overview of a wing or room's contents. Complements the existing `recall()` (raw L2 retrieval) and `search()` (semantic L3 search) with a "catch me up" mode.

How it works:
- Fetches top 20 drawers via L2
- Deduplicates using 70% word-overlap threshold
- Caps output at ~2000 chars
- Formats with topic count header

Available via:
- `MemoryStack.brief(wing=..., room=...)` in Python
- `searcher.brief()` wrapper
- `mempalace brief --wing X --room Y` CLI subcommand

Closes #73.

## How to test

```bash
# CLI
mempalace brief --wing my_project

# Python
from mempalace.layers import MemoryStack
stack = MemoryStack()
print(stack.brief(wing="my_project"))
```

Or run the test:
```bash
pytest tests/test_searcher.py::TestSearchMemories::test_brief -v
```

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -112,6 +112,14 @@ def cmd_search(args):
         sys.exit(1)
 
 
+def cmd_brief(args):
+    from .searcher import brief
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+    result = brief(palace_path=palace_path, wing=args.wing, room=args.room)
+    print(result)
+
+
 def cmd_wakeup(args):
     """Show L0 (identity) + L1 (essential story) — the wake-up context."""
     from .layers import MemoryStack
@@ -412,6 +420,11 @@ def main():
     p_search.add_argument("--room", default=None, help="Limit to one room")
     p_search.add_argument("--results", type=int, default=5, help="Number of results")
 
+    p_brief = sub.add_parser("brief", help="Condensed overview of a wing or room")
+    p_brief.add_argument("--wing", default=None, help="Wing to summarize")
+    p_brief.add_argument("--room", default=None, help="Room to summarize")
+    p_brief.add_argument("--palace", default=None, help="Palace path")
+
     # compress
     p_compress = sub.add_parser(
         "compress", help="Compress drawers using AAAK Dialect (~30x reduction)"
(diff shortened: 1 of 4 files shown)


mvanhorn:
Thanks for the thorough re-review @bgauryy. Appreciate you verifying against the codebase.

On the follow-ups:
- #4 (MCP tool): makes sense, I'll add `tool_brief` in a follow-up PR
- #5 ("topics" label): good point, "items" is more accurate. Happy to change it in this PR if you'd like, or I can bundle it with the MCP follow-up

@milla-jovovich anything else needed to merge this?

bensig:
Conflicts with main. Brief retrieval mode is an interesting feature — if you'd like to rebase against current main we'd review it.

mvanhorn:
Rebased onto current main and resolved the conflicts (the dispatch table in `cli.py` and the test imports in `tests/test_searcher.py`). Full test suite passes (568 passed, including the new `test_brief`) and `ruff check` is clean.

Since this one was already closed I opened #601 with the rebased branch — happy to close that and reopen here instead if you'd prefer. Thanks @bensig!

PR #175, opened 2026-04-08
feat(security): add security hardening — auth, encryption, validation

jessecurry:
## Summary

Adds opt-in security hardening to MemPalace based on a comprehensive security audit. All features are backward-compatible — MemPalace works exactly as before unless security is explicitly enabled in config.

### What's included

- **Token-based authentication** on the MCP server — tokens stored in OS keychain (macOS Keychain, Windows Credential Manager, Linux Secret Service) via `keyring`, with file fallback
- **Fernet encryption at rest** for palace data — encryption keys also stored in OS keychain. Encrypted content stored in ChromaDB metadata; when encryption is enabled, the `documents` field stores only a content-hash placeholder to prevent plaintext persistence on disk
- **Input validation** on all MCP tool calls — required field checking, unknown field rejection, configurable content size limits (default 1MB)
- **File permission hardening** — `0o600` on sensitive files, `0o700` on config directories
- **SHA-256 replacing MD5** for all ID generation (5 call sites)
- **Localhost-only guard** for future HTTP transport — `start_http_server()` rejects non-localhost bind addresses
- **Updated .gitignore** — `.env`, `*.key`, `auth_token`, `*.sqlite3`, `entities.json`, IDE configs

### Important: Semantic search and encryption

When encryption is **disabled** (the default), semantic search works exactly as it does today — no behavior change whatsoever.

When encryption is **enabled**, the ChromaDB `documents` field stores a `[encrypted:<hash>]` placeholder instead of plaintext, so content is not persisted unencrypted on disk. The trade-off is that **semantic search will not return meaningful results for encrypted drawers**, since ChromaDB cannot compute useful embeddings on the placeholder. This is an honest, unavoidable trade-off — you cannot have encryption at rest *and* semantic search over plaintext simultaneously without an external embeddings service. The authoritative content lives only in the `encrypted_content` metadata field.

Users who need both search and encryption should pair this with OS-level disk encryption (FileVault, LUKS, BitLocker) and leave application-level encryption disabled.

### Configuration

All security features live under a `security` block in `~/.mempalace/config.json`:

```json
{
  "security": {
    "auth_enabled": true,
    "encryption_enabled": true,
    "max_content_size": 1048576
  }
}
```

Environment variable overrides: `MEMPALACE_AUTH_ENABLED`, `MEMPALACE_ENCRYPTION_ENABLED`.

### Dependencies

`cryptography` and `keyring` added as optional `[security]` extra:
```
pip install mempalace[security]
```

Core mempalace continues to work with zero extra dependencies.

## Test plan

- [x] All 151 tests pass (101 original + 50 new)
- [x] Auth: requests without token rejected (-32001), valid token succeeds, backward compat when auth disabled
- [x] Encryption: plaintext NOT stored in documents field, encrypted_content in metadata, search returns decrypted text, unencrypted drawers still readable
- [x] Input validation: missing required fields rejected, unknown fields rejected, size limits enforced
- [x] File permissions: 0o600 on files, 0o700 on directories
- [x] Localhost guard: rejects 0.0.0.0 and public IPs, accepts 127.0.0.1/::1/localhost
- [x] SHA-256 hashing produces correct output
- [x] Ruff lint and format clean

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,26 @@ __pycache__/
 *.pyc
 .pytest_cache/
 mempal.yaml
+
+# Security-sensitive files
+.env
+.env.*
+*.key
+auth_token
+entities.json
+
+# Database files
+*.sqlite3
+
+# Virtual environments
+venv/
+.venv/
+env/
+
+# IDE
+.idea/
+.vscode/
+
+# Test coverage
+htmlcov/
+.coverage
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -8,6 +8,8 @@ import json
 import os
 from pathlib import Path
 
+from .security import secure_dir, secure_file
+
 DEFAULT_PALACE_PATH = os.path.expanduser("~/.mempalace/palace")
 DEFAULT_COLLECTION_NAME = "mempalace_drawers"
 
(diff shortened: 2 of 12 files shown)


bensig:
Thanks for the security work. This conflicts with main and the core security hardening landed in #387. If there are specific pieces not covered, happy to look at a focused follow-up PR.

PR #183, opened 2026-04-08
docs: macOS Apple Silicon (M1+) install guide

nixlim:
## Summary

- Adds `docs/macos-m1-install-guide.md` — a detailed setup guide for MemPalace on Apple Silicon Macs
- Covers native ARM64 Python 3.12 via Homebrew, pipx install, MCP server config
- Explains the concurrent multi-repo mining pattern using separate `--palace` directories
- Documents three failure modes discovered through real debugging:
  1. **Rosetta 2 SIGSEGV**: x86_64 Python causes ChromaDB Rust HNSW bindings to crash — ARM64 Python is required
  2. **x86_64 → ARM64 palace migration**: HNSW segment files written by x86_64 chromadb are incompatible with ARM64; must wipe and re-mine
  3. **MCP server + mine concurrency**: two processes writing to the same palace simultaneously triggers SIGSEGV — use separate `--palace` paths

## Test plan

- [ ] Verify install steps on a clean macOS Apple Silicon machine
- [ ] Confirm `pipx reinstall --python` note is accurate (vs `install --force` which silently ignores `--python`)
- [ ] Verify RAM/disk estimates are reasonable for typical codebases

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/docs/macos-m1-install-guide.md
+++ b/docs/macos-m1-install-guide.md
@@ -26,7 +26,9 @@ Rosetta, or install ARM64 Homebrew:
 
 ### 2. Install ARM64 Python 3.12
 
-MemPalace requires Python 3.12. Install it via the native ARM64 Homebrew:
+MemPalace supports Python 3.9–3.12. Python 3.13+ is not yet supported (dependency
+compatibility issues). Python 3.12 is recommended for Apple Silicon — it has the best
+ARM64 wheel availability. Install it via the native ARM64 Homebrew:
 
 ```bash
 brew install python@3.12
@@ -88,7 +90,9 @@ file ~/.local/pipx/venvs/mempalace/bin/python
 ## MCP Server Configuration
 
 To use MemPalace as a memory backend for Claude or other MCP-compatible clients, add it
-to your MCP config (e.g. `~/.mcp.json`):
+to your MCP config. For Claude Code, this is `~/.mcp.json` (or `.mcp.json` in a project
+root for per-project config). Other MCP clients (Cursor, Windsurf, etc.) use different
+config paths — check their documentation.
 
 ```json
 {
@@ -207,6 +211,10 @@ If you previously ran MemPalace under x86_64 Python (Rosetta), the HNSW segment 
 will crash trying to open them.
 
 Fix: wipe the palace and re-mine:
+
+> **Warning**: `~/.mempalace/palace/` is the default palace used by all projects.
+> If you have existing data you want to keep, back it up first or use
+> `--palace` to mine into a separate directory instead.
 
 ```bash
 rm -rf ~/.mempalace/palace/
--- a/docs/macos-m1-install-guide.md
+++ b/docs/macos-m1-install-guide.md
@@ -1,0 +1,204 @@
+# Installing MemPalace on macOS (Apple Silicon / M1+)
+
+This guide covers a reliable, crash-free MemPalace installation on Apple Silicon Macs,
+including concurrent multi-repo mining via separate palaces.
+
+---
+
+## Prerequisites
+
+### 1. Install ARM64 Homebrew
+
+Make sure you are running the native ARM64 Homebrew at `/opt/homebrew` (not the Rosetta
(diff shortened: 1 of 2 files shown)


docs/macos-m1-install-guide.md
```diff
@@ -0,0 +1,242 @@
...
+# Mine the project into the default palace
+mempalace mine ~/projects/my-project
+```
+
+### Multiple repos concurrently (safe pattern)
+
+**Key rule**: never run two processes against the same palace simultaneously. Each
+concurrent mine must have its own `--palace` directory. The ChromaDB HNSW index does not
+support concurrent writes and will crash (SIGSEGV) if two processes write to the same
+index at the same time.
```
JSee98: the issue is not isolated to indexes rather concurrent writes themselves. It's a chromaDB issue, well more a sqllite issue as that is the BE i believe.

web3guru888 (commented):
This is a high-quality, field-tested installation guide that addresses three real failure modes most macOS users discover the hard way. Here's my technical assessment:

**What's done well**

The Rosetta 2 / ARM64 distinction is the most critical thing for new users to understand, and it's explained clearly with a concrete diagnostic (`file` command, checking for `arm64` in output). The SIGSEGV crash at address 0x88 is specific enough to be searchable — that's a small but meaningful touch for developers hitting the issue blind.

The `pipx reinstall --python` vs `install --force` gotcha is genuinely surprising behavior and absolutely worth documenting. I've seen this one cause silent regressions where the reinstall appears to succeed but the wrong Python ends up in the venv. The current note could be even more explicit: `pipx install --force` ignores `--python` silently, meaning `file ~/.local/pipx/venvs/mempalace/bin/python` will still show x86_64 after a seemingly successful reinstall.

The concurrent mining section correctly identifies that ChromaDB's SQLite backend is single-writer. The "stop MCP server before mining same palace" warning is critical and frequently missed by users who have always-on Claude Desktop setups.

**Issues and gaps**

1. **HNSW wipe warning is undersized.** The "Warning" block says to back up data, but `rm -rf ~/.mempalace/palace/` deletes the *default* palace for all projects. Consider adding a step showing how to list palaces first: `ls ~/.mempalace/` so users know what they're about to lose.

2. **`chroma.sqlite3` portability caveat.** The note that SQLite is non-portable because segment UUIDs reference the deleted HNSW directory is accurate. However, there's a common follow-up question: "can I copy just the sqlite file to a new machine?" The answer is no for the same reason, but it's not stated explicitly.

3. **Python 3.13 compatibility note.** The guide says "Python 3.13+ is not yet supported (dependency compatibility issues)." Worth noting which dependency specifically — ChromaDB's Rust wheels? ONNX Runtime? This helps users who hit the restriction understand when it might be lifted.

4. **`MEMPALACE_PALACE_PATH` env var.** The MCP config section uses this variable, but the main docs don't document it. Worth a quick "see `mempalace --help`" or a link to the config reference so users know this is an officially supported path, not an undocumented workaround.

5. **M3/M4 applicability.** The title says "M1+" but some users may wonder if M3 Pro / M4 Max behave differently. A one-liner confirming "this guide applies equally to M1/M2/M3/M4" would close that loop.

**Minor nits**

- "Initialising" (British spelling) vs the rest of the repo's "Initializing" — pick one consistently.
- The `safe_separator` variable in `searcher.py` is computed once outside the loop (good), but the guide's code examples all use `mempalace mine` without `--palace`. For multi-repo users starting here, the first command they'll run is the single-repo case, and the default palace path (`~/.mempalace/palace/`) isn't shown explicitly anywhere in the guide.

**Overall verdict**

✅ **Approve with minor suggestions.** This is exactly the kind of platform-specific documentation that reduces issue volume — it addresses real bugs with real diagnostics. The ARM64/x86_64 SIGSEGV and concurrent write crash sections alone are worth merging. Address the HNSW wipe warning scope and the Python 3.13 specifics before merge if possible; the rest can be follow-up PRs.

---

igorls:

Issue #40, opened 2026-04-07
status command shows max 10,000 drawers (hardcoded limit) + palace_path not expanded


RobertoGEMartin:
## Bug 1: `status` command capped at 10,000 drawers

**File**: `miner.py:402`
**Line**: `r = col.get(limit=10000, include=["metadatas"])`

When a palace has more than 10,000 drawers, `mempalace status` only counts the first 10K. The display shows an incorrect total and misses wings/rooms that were filed after the first 10K.

**Fix**: Replace `10000` with `col.count()`:
```python
total = col.count()
r = col.get(limit=total, include=["metadatas"])
```

**Impact**: Display-only — search and mining work correctly on all drawers.

## Bug 2: `palace_path` from config.json not expanded

**File**: `config.py` — `palace_path` property returns `~/.mempalace/palace` without `os.path.expanduser()`

When ChromaDB receives a path starting with `~`, it treats it as a **relative path** from CWD, creating a literal `~/` directory inside the working directory. This silently stores all data in the wrong location.

**Reproduction**:
1. Set `config.json`: `"palace_path": "~/.mempalace/palace"`
2. Run `mempalace mine . --wing test`
3. Check: a literal `~/.mempalace/palace/` folder is created in CWD (not in home)
4. `os.path.expanduser("~/.mempalace/palace")` resolves to `/Users/user/.mempalace/palace` — a different, empty location

**Fix**: In `config.py`, apply `os.path.expanduser()` when returning `palace_path`:
```python
@property
def palace_path(self):
    env_val = os.environ.get("MEMPALACE_PALACE_PATH") or os.environ.get("MEMPAL_PALACE_PATH")
    if env_val:
        return os.path.expanduser(env_val)
    return os.path.expanduser(self._file_config.get("palace_path", DEFAULT_PALACE_PATH))
```

**Impact**: Data loss risk — users may unknowingly have their palace in a project subdirectory instead of home. The data is intact but in the wrong location.

## Environment
- mempalace version: latest (pip install mempalace, 2026-04-07)
- Python: 3.13.7
- ChromaDB: bundled version
- macOS Darwin 25.2.0

bensig:
Fixed in #66 — batched ChromaDB reads replace the hardcoded 10k limit. palace_path expanduser also landed.

PR #193, opened 2026-04-08
fix: expanduser() for ~ in palace_path from config/env (#40)

ThanhNguyxn07:
## Summary
This PR fixes issue #40 where the palace_path configuration from environment variables or config files would not expand the ~ user home directory shortcut, leading to incorrect path resolution.

## Changes
- Apply os.path.expanduser() to palace_path values from both environment variables and config file
- This ensures ~/.mempalace/palace expands to /home/user/.mempalace/palace as expected

## Why this fix
Without expansion, paths like MEMPALACE_PALACE_PATH=~/custom/palace would create directories literally named ~ instead of expanding to the user's home directory.

## Testing
- Added tests to verify ~ expansion works from both config and environment variables
- Verified existing functionality remains unchanged

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -94,8 +94,8 @@ class MempalaceConfig:
         """Path to the memory palace data directory."""
         env_val = os.environ.get("MEMPALACE_PALACE_PATH") or os.environ.get("MEMPAL_PALACE_PATH")
         if env_val:
-            return env_val
-        return self._file_config.get("palace_path", DEFAULT_PALACE_PATH)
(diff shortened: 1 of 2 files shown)


igorls:
Thanks for this, and sorry it sat so long. `~` expansion for `palace_path` is handled in `config.py` now — both the env-var and config-file paths run through `os.path.expanduser`, so issue #40 is covered. Closing as superseded, but the bug you reported was real and the fix shipped. Appreciate you chasing it down.

Issue #180, opened 2026-04-08
CLI status silently truncates drawer count at 10,000 with no warning


fubak:
## Bug

`mempalace status` reports an incorrect drawer count when the palace contains more than 10,000 drawers. The output looks authoritative but is silently wrong — no truncation notice, no pagination indicator.

## Steps to Reproduce

```bash
# Mine a large project
mempalace mine /path/to/large-project --wing myproject

# Check status
mempalace status
# → "MemPalace Status — 10000 drawers"
```

## Expected vs Actual

**CLI output:**
```
MemPalace Status — 10000 drawers

  WING: myproject
    ROOM: apps     8682 drawers
    ROOM: supabase 1231 drawers
    ...
```

**Actual count via direct ChromaDB query:**
```python
client = chromadb.PersistentClient(path='~/.mempalace/palace')
coll = client.get_collection('mempalace_drawers')
print(coll.count())  # → 95964
```

**MCP server `mempalace_status`** correctly reports `total_drawers: 95964`.

## Impact

- Users have no way to verify a large mine operation completed correctly using the CLI
- The discrepancy between CLI (10,000) and MCP (95,964) is confusing and erodes trust in both
- Silent truncation with no warning is worse than an explicit limit

## Expected

Either show the real count, or clearly indicate truncation: e.g. `10,000+ drawers (showing top 10,000)`.

## Environment

- mempalace 3.0.0
- Python 3.12.3
- ChromaDB 1.5.6

fubak:
Confirming @bjperson's findings — we hit both symptoms in our setup (mempalace 3.0.0, ChromaDB 1.5.6, ~96k drawers across 3 wings).

**wake-up** returns `No drawers found` silently when the palace exceeds a certain size, even with a valid `--wing` filter.

**MCP `mempalace_status`** also has a related issue: it correctly reports the total drawer count (e.g. `95964`) but returns `wings: {}` and `rooms: {}` — suggesting the same unbounded `col.get()` call fails silently when building the taxonomy, rather than raising an error.

The bounded `col.query()` fix suggested by @bjperson would likely resolve all three: CLI status truncation, wake-up silence, and MCP taxonomy returning empty. Happy to test a fix if one lands.

fubak:
Confirming this is **still broken in 3.1.0**. After upgrading and re-mining, `mempalace status` shows:

```
MemPalace Status — 10000 drawers

  WING: agent-action-firewall
    ROOM: apps        8682 drawers
[...]

PR #194, opened 2026-04-08
fix: remove 10k status truncation by using collection count (#180)

ThanhNguyxn07:
## Summary
This PR fixes issue #180 where the mempalace status command would truncate results at 10,000 entries even when the palace contained more drawers, giving users incorrect information about their palace size.

## Changes
- Replace hardcoded limit=10000 with dynamic limit=col.count() in mempalace/miner.py
- This ensures status shows the actual total count instead of truncated count

## Why this fix
Previously, users with large palaces (>10k drawers) would see misleading status output showing only 10,000 drawers even when thousands more existed. This change ensures accurate reporting of actual palace size.

## Testing
- Added test to verify status function uses actual collection count
- Verified function works with collections of various sizes

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -654,7 +654,8 @@ def status(palace_path: str):
         return
 
     # Count by wing and room
-    r = col.get(limit=10000, include=["metadatas"])
+    total = col.count()
+    r = col.get(limit=total, include=["metadatas"])
     metas = r["metadatas"]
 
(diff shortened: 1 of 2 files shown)


igorls:
Thanks for this, and apologies for the delay. `status` no longer truncates: the drawer count comes from a direct SQLite aggregate rather than a capped query, so the 10,000 ceiling in issue #180 is gone. Closing as superseded — thanks for spotting it.

Issue #171, opened 2026-04-07
MCP server: mempalace_list_wings and mempalace_get_taxonomy return empty despite data in palace


fubak:
## Bug

`mempalace_list_wings` and `mempalace_get_taxonomy` return empty objects via the MCP server, even when the palace has data. `mempalace_status` correctly reports the total drawer count but returns `wings: {}` and `rooms: {}`.

## Steps to Reproduce

```bash
# Mine a project
mempalace mine /path/to/project --wing myproject

# Verify data exists via CLI
mempalace status
# → correctly shows wings, rooms, and drawer counts

# Start MCP server and call tools
mempalace_status       # → total_drawers: 95964, wings: {}, rooms: {}
mempalace_list_wings   # → { "wings": {} }
mempalace_get_taxonomy # → { "taxonomy": {} }
```

## Expected

All three tools should return the same wing/room structure that the CLI `mempalace status` command shows.

## Actual

```json
{ "wings": {} }
{ "taxonomy": {} }
```

## Environment

- mempalace 3.0.0
- Python 3.12.3
- ChromaDB 1.5.6
- Claude Code MCP integration
- Palace: single collection `mempalace_drawers`, 95,964 documents, all with `wing` and `room` metadata fields confirmed via direct ChromaDB query

## Additional Context

Direct ChromaDB inspection confirms the metadata is present and correctly structured:

```python
import chromadb
client = chromadb.PersistentClient(path='~/.mempalace/palace')
coll = client.get_collection('mempalace_drawers')
results = coll.get(limit=3, include=['metadatas'])
# → {'wing': 'agent-action-firewall', 'room': 'testing', ...}
```

`mempalace_search` with `wing` filter works correctly and returns the right results — so the data and metadata are intact. Only the taxonomy-building tools are affected.

The CLI `mempalace status` works correctly, suggesting the MCP server uses a different code path to build the wing/room taxonomy that isn't querying ChromaDB metadata properly.

mvalentsev:
The MCP navigation tools now paginate metadata via _fetch_all_metadata() (batches of 1000) instead of a single col.get(limit=10000). This should resolve the empty results on large palaces. Same fix as #723, landed via #707.

arnoldwender:
[...]

PR #202, opened 2026-04-08
fix/mcp-taxonomy-pagination (#171)

Formatted:
**## What does this PR do?**

Fixes a silent data-loss bug (#171) in the four MCP taxonomy tools (`mempalace_status`, `mempalace_list_wings`, `mempalace_list_rooms`, `mempalace_get_taxonomy`). On palaces with more than ~10k drawers, all four functions called `col.get(limit=10000)` in a single shot and wrapped the whole thing in `except Exception: pass`. Two things could go wrong:

1. ChromaDB silently caps or returns `None` for `metadatas` on large collections, causing a `TypeError: 'NoneType' is not iterable` that gets swallowed — result: `{"wings": {}}` on a 96k-drawer palace.
2. Even when it doesn't raise, a single 10k-item fetch misses everything beyond the cap.

The fix extracts a shared `_iter_metadatas(col, where=None)` generator that pages through the collection in batches of 500 (the same pattern already used correctly in `layers.py`), guards against `None` metadatas at each page, and logs a warning instead of silently swallowing errors. All four taxonomy functions are updated to use it.

**## How to test**

```bash
# Install deps
pip install -e ".[dev]"

# Run the new pagination tests (no network needed — uses in-memory fixtures)
python -m pytest tests/test_mcp_server.py::TestTaxonomyPagination -v

# Full suite
python -m pytest tests/ -v
```

The three new tests in `TestTaxonomyPagination` monkey-patch `_TAXONOMY_BATCH = 2` to force multi-page fetching on a small fixture, then assert all items are counted across pages. A third test confirms graceful handling when ChromaDB returns `None` for `metadatas`.

**## Checklist**
- [x] No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -57,6 +57,47 @@ def _no_palace():
     }
 
 
+# Batch size for paginated metadata scans.
+# A single col.get(limit=N) call on large palaces (>10k drawers) can return
+# None metadatas or raise in some ChromaDB builds.  Fetching in small pages
+# is both safer and friendlier on memory.
+_TAXONOMY_BATCH = 500
+
+
+def _iter_metadatas(col, where=None):
+    """
+    Yield every metadata dict in *col*, fetched in pages of *_TAXONOMY_BATCH*.
+
+    This replaces the previous pattern of ``col.get(limit=10000)`` which
+    silently returned empty results on palaces with more than ~10k drawers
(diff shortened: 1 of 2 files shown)


bensig:
Reopening — this was closed too quickly and your work deserves proper review. Standing by for merge once we clear the current security baseline. Thank you for the contribution.

Formatted:
#307 is the cleaner solution, closing this one

PR #226, opened 2026-04-08
fix: redirect stdout to stderr during import to prevent JSON parse er…

alexschneiderru:
**Problem:** When Claude Desktop launches `mempalace mcp` as an MCP subprocess, ChromaDB and other dependencies print diagnostic text to stdout during import. Claude Desktop expects only JSON-RPC on stdout, causing repeated parse errors:
MCP mempalace: Unexpected token '', "*********"... is not valid JSON

Confirmed on Windows 11: `mempalace status 2>nul` outputs all text to stdout, `mempalace status 1>nul` produces nothing on stderr.

**Fix:** Save the real stdout handle before any imports, redirect stdout to stderr at module load time (so all import noise goes to stderr), then use `_real_stdout` exclusively for JSON-RPC writes in `main()`.

Tested with Claude Desktop 1.1062.0 on Windows 11, mempalace 3.0.0, Python 3.10.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1,16 +1,8 @@
-import sys as _sys
-_real_stdout = _sys.stdout
-_sys.stdout = _sys.stderr
-import os as _os
-_os.environ.setdefault("ORT_LOGGING_LEVEL", "3")
-_os.environ.setdefault("TRANSFORMERS_VERBOSITY", "error")
-import warnings as _warnings
-_warnings.filterwarnings("ignore")
 #!/usr/bin/env python3
 """
 MemPalace MCP Server — read/write palace access for Claude Code
 ================================================================
-Install: claude mcp add mempalace -- python /path/to/mcp_server.py
+Install: claude mcp add mempalace -- python -m mempalace.mcp_server
 
 Tools (read):
   mempalace_status          — total drawers, wing/room breakdown
@@ -25,6 +17,9 @@ Tools (write):
   mempalace_delete_drawer   — remove a drawer by ID
 """
 
+import sys as _sys
+_real_stdout = _sys.stdout
+_sys.stdout = _sys.stderr
 import sys
 import json
 import logging
@@ -35,6 +30,7 @@ from .config import MempalaceConfig
 from .searcher import search_memories
 from .palace_graph import traverse, find_tunnels, graph_stats
 from .knowledge_graph import KnowledgeGraph
+from .version import __version__
 
 _kg = KnowledgeGraph()
 
@@ -654,7 +650,7 @@ def handle_request(request):
             "result": {
                 "protocolVersion": "2024-11-05",
                 "capabilities": {"tools": {}},
-                "serverInfo": {"name": "mempalace", "version": "2.0.0"},
+                "serverInfo": {"name": "mempalace", "version": __version__},
             },
         }
     elif method == "notifications/initialized":
@@ -679,6 +675,14 @@ def handle_request(request):
                 "id": req_id,
                 "error": {"code": -32601, "message": f"Unknown tool: {tool_name}"},
             }
+        schema_props = TOOLS[tool_name]["input_schema"].get("properties", {})
+        for key, value in list(tool_args.items()):
+            prop_schema = schema_props.get(key, {})
+            declared_type = prop_schema.get("type")
+            if declared_type == "integer" and not isinstance(value, int):
+                tool_args[key] = int(value)
+            elif declared_type == "number" and not isinstance(value, (int, float)):
+                tool_args[key] = float(value)
         try:
             result = TOOLS[tool_name]["handler"](**tool_args)
             return {
@@ -686,9 +690,9 @@ def handle_request(request):
                 "id": req_id,
                 "result": {"content": [{"type": "text", "text": json.dumps(result, indent=2)}]},
             }
-        except Exception as e:
-            logger.error(f"Tool error in {tool_name}: {e}")
-            return {"jsonrpc": "2.0", "id": req_id, "error": {"code": -32000, "message": str(e)}}
+        except Exception:
+            logger.exception(f"Tool error in {tool_name}")
+            return {"jsonrpc": "2.0", "id": req_id, "error": {"code": -32000, "message": "Internal tool error"}}
 
     return {
         "jsonrpc": "2.0",


bgauryy:
# PR Review: fix: redirect stdout to stderr during import to prevent JSON parse errors

## Executive Summary
| Aspect | Value |
|--------|-------|
| **PR Goal** | Prevent ChromaDB/ONNX diagnostic output from polluting MCP JSON-RPC stdout |
| **Files Changed** | 1 (`mempalace/mcp_server.py`) |
| **Risk Level** | 🔴 HIGH — core fix is valid but bundled regressions break version reporting, type safety, and error handling |
| **Review Effort** | 3/5 — single file but many changes hidden in whitespace reformatting |
| **Recommendation** | 🔄 REQUEST_CHANGES |

**Affected Areas**: MCP server startup, tool argument handling, error responses, version reporting

**Business Impact**: The stdout fix solves a real user-facing problem (Claude Desktop JSON parse errors), but the bundled regressions would break type coercion for MCP clients and report wrong version.

**Flow Changes**: JSON-RPC responses now write to `_real_stdout` instead of `sys.stdout`. All `import`-time stdout output is redirected to stderr. Type coercion for tool arguments is removed, changing how all 14 tools receive parameters.

## Ratings
| Aspect | Score |
|--------|-------|
| Correctness | 2/5 |
| Security | 3/5 |
| Performance | 4/5 |
| Maintainability | 3/5 |

## PR Health
- [x] Has clear description
- [x] References the problem with error logs
- [ ] Appropriate size — ~60% of diff is whitespace reformatting, obscuring real changes
- [ ] Has relevant tests

## High Priority Issues

### 🐛 #1: Shebang line displaced — code runs before `#!/usr/bin/env python3`
**Location:** `mempalace/mcp_server.py:1-9` | **Confidence:** ✅ HIGH

The stdout redirect code is placed *before* the shebang line. The shebang must be the absolute first line of a file for the OS to recognize it. While `python -m mempalace.mcp_server` still works, this breaks direct execution (`./mcp_server.py`) and is structurally wrong.

```diff
+ #!/usr/bin/env python3
  import sys as _sys
  _real_stdout = _sys.stdout
  _sys.stdout = _sys.stderr
  import os as _os
  _os.environ.setdefault("ORT_LOGGING_LEVEL", "3")
  _os.environ.setdefault("TRANSFORMERS_VERBOSITY", "error")
  import warnings as _warnings
  _warnings.filterwarnings("ignore")
- #!/usr/bin/env python3
```

Move the shebang to line 1 and put the redirect code immediately after.

---

### 🐛 #2: Hardcoded version "2.0.0" replaces dynamic `__version__` (currently 3.0.0)
**Location:** `mempalace/mcp_server.py:654` | **Confidence:** ✅ HIGH

The PR removes `from .version import __version__` and hardcodes `"version": "2.0.0"` in the MCP `initialize` response. The current `version.py` has `__version__ = "3.0.0"`. This is a version **downgrade** that will report the wrong version to every MCP client.

```diff
- "serverInfo": {"name": "mempalace", "version": "2.0.0"},
+ "serverInfo": {"name": "mempalace", "version": __version__},
```

Restore `from .version import __version__` and use it in the server info.

---

### 🐛 #3: Removed type coercion block — MCP string/float args will crash tool handlers
**Location:** `mempalace/mcp_server.py:679-693` (removed) | **Confidence:** ✅ HIGH

The PR deletes the argument coercion block that converted MCP JSON transport values to the correct Python types. MCP clients can send `"5"` (string) instead of `5` (int), or `5.0` (float) instead of `5`. Without coercion, calls like `tool_diary_read(last_n="10")` will crash with `TypeError` when Python tries `entries[:last_n]` with a string.

The removed block:
```python
schema_props = TOOLS[tool_name]["input_schema"].get("properties", {})
for key, value in list(tool_args.items()):
    prop_schema = schema_props.get(key, {})
    declared_type = prop_schema.get("type")
    if declared_type == "integer" and not isinstance(value, int):
        tool_args[key] = int(value)
    elif declared_type == "number" and not isinstance(value, (int, float)):
        tool_args[key] = float(value)
```

This block must be restored. It protects all 14 MCP tools from JSON type mismatches.

---

### 🚨 #4: Error handling downgrade — traceback lost + internal errors leaked to clients
**Location:** `mempalace/mcp_server.py:695-697` | **Confidence:** ✅ HIGH

Two regressions in one change:

**a) `logger.exception()` → `logger.error()`**: The original logged the full traceback to stderr for debugging. The new code only logs the error message, making production debugging much harder.

**b) `"Internal tool error"` → `str(e)`**: The original returned a safe generic message. The new code exposes raw exception messages to MCP clients, which can leak file paths, internal state, or dependency internals.

```diff
- except Exception as e:
-     logger.error(f"Tool error in {tool_name}: {e}")
-     return {"jsonrpc": "2.0", "id": req_id, "error": {"code": -32000, "message": str(e)}}
+ except Exception:
+     logger.exception(f"Tool error in {tool_name}")
+     return {
+         "jsonrpc": "2.0",
+         "id": req_id,
+         "error": {"code": -32000, "message": "Internal tool error"},
+     }
```

---

## Medium Priority Issues

### 🎨 #5: Install instruction changed to non-portable path
**Location:** `mempalace/mcp_server.py:5` (docstring) | **Confidence:** ⚠️ MED

The docstring install instruction changed from `python -m mempalace.mcp_server` to `python /path/to/mcp_server.py`. The `-m` form is the correct way to run a package module and works regardless of install location.

```diff
- Install: claude mcp add mempalace -- python /path/to/mcp_server.py
+ Install: claude mcp add mempalace -- python -m mempalace.mcp_server
```

---

### 🎨 #6: Massive whitespace reformatting mixed with functional changes
**Location:** Throughout file | **Confidence:** ⚠️ MED

~60% of the diff (roughly 70+ lines) is pure style reformatting — collapsing multi-line dicts to single lines, rewrapping function parameters. This makes the real changes extremely hard to spot during review and in `git blame` history. Style changes should be in a separate commit/PR.

---

## Low Priority Issues

### 🔗 #7: `palace_path` added to `_no_palace()` response — helpful but leaks local paths
**Location:** `mempalace/mcp_server.py:63` | **Confidence:** ⚠️ MED

Adding `"palace_path": _config.palace_path` to the error is useful for debugging. Minor concern: it exposes the user's filesystem path to the MCP client. Acceptable for a local-only tool, but worth noting.

---

### ⚡ #8: Removed `limit=10000` from `col.get()` calls — actually correct
**Location:** `mempalace/mcp_server.py:76,133,148,166,407` | **Confidence:** ✅ HIGH

ChromaDB's `get()` with `limit=None` (default) returns all records. Removing the arbitrary `limit=10000` cap is the correct behavior — palaces with >10k drawers were silently truncated before. **This change is good.**

---

## Flow Impact Analysis

```
BEFORE:                                 AFTER:
sys.stdout = real stdout                sys.stdout = sys.stderr (during import)
                                        _real_stdout = saved real stdout
imports run → print to stdout ←BUG      imports run → print to stderr ←FIXED
                                        
main() loop:                            main() loop:
  sys.stdout.write(json) ←OK             _real_stdout.write(json) ←OK

Type coercion:                          Type coercion:
  "5" → int(5) before handler ←SAFE       "5" passed as-is → TypeError ←BUG
```

The stdout redirect is a valid fix. The problem is the unrelated regressions bundled in.

**Recommendation**: Accept the stdout redirect mechanism (issues #1 aside), reject the removal of type coercion (#3), version hardcoding (#2), and error handling downgrade (#4).

---

PR #280, opened 2026-04-08
feat(dedup): add diary dedup guard and palace-wide duplicate report tool

joanfgarcia:
## What does this PR do?

Adds two deduplication features to prevent **embedding reinforcement** -- the silent accumulation of near-identical content that degrades semantic search quality over time.

### 1. Diary write dedup guard (`tool_diary_write`)

`tool_diary_write` now performs a cosine similarity check (`>= 0.92`) before inserting a new entry. If a near-identical document already exists in the palace, the write is rejected with a structured response:

```json
{
  "success": false,
  "reason": "duplicate_diary_entry",
  "message": "A near-identical diary entry already exists.",
  "matches": [...]
}
```

**Why this matters:** Diary entries are written automatically by hooks and agents. Without dedup, repeated saves or session summaries with trivial variations create N copies of the same embedding. ChromaDB's default distance metric treats these as equally relevant results, flooding search with redundant content instead of diverse matches.

The threshold of 0.92 was chosen to be strict enough to catch semantically identical content while allowing genuinely different entries about similar topics to coexist. This is consistent with the existing 0.9 threshold in `tool_add_drawer`.

### 2. Palace-wide duplicate report (`mempalace_dedup_report`)

New read-only diagnostic tool that scans the palace for near-duplicate clusters:

```
mempalace_dedup_report(threshold=0.92, wing="optional", limit=1000)
```

Returns structured output:
- **Clusters**: Each cluster has an "anchor" document and its near-duplicates
- - **Similarity scores**: Per-pair cosine similarity
- - **Metadata**: Wing, room, filed_at timestamps for triage
- - **Previews**: First 150 chars of each document
This is intentionally **read-only** -- it reports but does not delete. Users can review clusters and decide which drawers to remove via `mempalace_delete_drawer`.

Supports optional wing filter to scope scans to a specific wing (useful for large palaces), and configurable threshold for sensitivity control.

## Technical details

- **No new dependencies.** Uses existing `tool_check_duplicate` internally.
- - **Performance:** `tool_dedup_report` issues one `col.query()` per unique document. For a 1000-drawer palace this is ~1000 queries against the local ChromaDB -- typically completes in under 5 seconds on modern hardware. The `limit` parameter caps scan scope.
- - **Clustering is greedy:** Once a document is assigned to a cluster (as anchor or duplicate), it's excluded from further matching via a `seen` set. This prevents O(n^2) blowup and avoids reporting the same pair in multiple clusters.
## How to test

```bash
uv run pytest tests/test_dedup.py -v        # 9 new tests
uv run pytest tests/ -v                      # 110 total, 0 regressions
uv run ruff check mempalace/mcp_server.py tests/test_dedup.py   # clean
```

## Test coverage

| Test | What it verifies |
|------|-----------------|
| test_exact_duplicate_rejected | Identical diary content blocked on second write |
| test_near_duplicate_rejected | Trivially varied content blocked |
| test_distinct_entry_accepted | Genuinely different content passes dedup |
| test_different_agents_can_write_similar | Cross-agent dedup works (same fact = same embedding) |
| test_report_on_clean_palace | No false positives on distinct documents |
| test_report_finds_duplicates | Exact duplicates detected and clustered |
| test_report_wing_filter | Wing scoping restricts scan correctly |
| test_report_empty_palace | Empty palace returns zeros without error |
| test_report_threshold_sensitivity | Lower threshold catches more near-duplicates |

## Checklist
- [x] No hardcoded paths
- [x] Linter passes (ruff check .)
---

> **A note of thanks.** We came across MemPalace while researching memory architectures for our own project and were genuinely impressed by the engineering quality -- the 4-layer token loading, the temporal KG, the entity registry. While auditing the codebase, we noticed that diary writes lacked the same dedup guard that `tool_add_drawer` already has, and that there was no way to diagnose accumulated redundancy across the palace. Consider this PR a *señal de pago* -- a down payment for the ideas we've borrowed. Good engineering deserves to be repaid in kind. :)

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,75 @@
+## What does this PR do?
+
+Adds two deduplication features to prevent **embedding reinforcement** — the silent accumulation of near-identical content that degrades semantic search quality over time.
+
+### 1. Diary write dedup guard (`tool_diary_write`)
+
+`tool_diary_write` now performs a cosine similarity check (`≥ 0.92`) before inserting a new entry. If a near-identical document already exists in the palace, the write is rejected with a structured response:
+
+```json
+{
+  "success": false,
+  "reason": "duplicate_diary_entry",
+  "message": "A near-identical diary entry already exists.",
+  "matches": [...]
+}
+```
+
+**Why this matters:** Diary entries are written automatically by hooks and agents. Without dedup, repeated saves or session summaries with trivial variations (e.g. `★★★` vs `★★`) create N copies of the same embedding. ChromaDB's default distance metric treats these as equally relevant results, flooding search with redundant content instead of diverse matches.
+
+The threshold of `0.92` was chosen to be strict enough to catch semantically identical content while allowing genuinely different entries about similar topics to coexist. This is consistent with the existing `0.9` threshold in `tool_add_drawer`.
+
+### 2. Palace-wide duplicate report (`mempalace_dedup_report`)
+
+New read-only diagnostic tool that scans the palace for near-duplicate clusters:
+
+```
+mempalace_dedup_report(threshold=0.92, wing="optional", limit=1000)
+```
+
+Returns structured output:
+- **Clusters**: Each cluster has an "anchor" document and its near-duplicates
+- **Similarity scores**: Per-pair cosine similarity
+- **Metadata**: Wing, room, `filed_at` timestamps for triage
+- **Previews**: First 150 chars of each document
+
+This is intentionally **read-only** — it reports but does not delete. Users can review clusters and decide which drawers to remove via `mempalace_delete_drawer`.
+
+Supports optional `wing` filter to scope scans to a specific wing (useful for large palaces), and configurable `threshold` for sensitivity control.
+
+## Technical details
+
+- **No new dependencies.** Uses existing `tool_check_duplicate` internally.
+- **Performance:** `tool_dedup_report` issues one `col.query()` per unique document. For a 1000-drawer palace this is ~1000 queries against the local ChromaDB — typically completes in under 5 seconds on modern hardware. The `limit` parameter caps scan scope.
+- **Clustering is greedy:** Once a document is assigned to a cluster (as anchor or duplicate), it's excluded from further matching via a `seen` set. This prevents O(n²) blowup and avoids reporting the same pair in multiple clusters.
+
+## How to test
+
+```bash
+uv run pytest tests/test_dedup.py -v        # 9 new tests
+uv run pytest tests/ -v                      # 110 total, 0 regressions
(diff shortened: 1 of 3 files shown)


web3guru888 (commented):
## Review: Diary Dedup Guard + Palace-wide Duplicate Report

This is addressing a real problem — embedding reinforcement — and doing it well. We've been fighting the same issue in our integration (208 discoveries, 710 KG entities) and can speak to the design choices here.

### Diary Dedup Guard

**Threshold choice (0.92):** We've experimented extensively with dedup thresholds. Our production configuration uses a tiered approach: hard dedup at 0.86 and soft dedup at 0.55 (the soft tier flags but doesn't block). At 0.92 you're on the conservative side — you'll catch near-exact copies but may miss moderate paraphrases. For diary entries specifically, 0.92 is probably the right call since diary entries are personal observations where semantic overlap between *different* observations about the *same topic* is expected and legitimate.

**⚠️ Cross-agent dedup behavior:** Your `test_different_agents_can_write_similar` test documents that agent B's write is rejected if agent A already wrote the same fact. The test comment says this is "arguably correct" — I'd push back slightly. In a multi-agent setup, different agents may independently observe the same fact from different perspectives. Scoping the dedup check to `{"$and": [{"room": "diary"}, {"wing": wing}]}` instead of just `{"room": "diary"}` would preserve cross-agent independence while still preventing self-duplication within an agent. This is closer to what we do — scoping dedup to wing+room.

**Performance concern on `tool_check_duplicate`:** The diary guard calls `tool_check_duplicate` (which queries ChromaDB) on every write. For high-frequency diary writers (our agents write ~50 diary entries per discovery cycle), this adds a query per write. Consider caching recent write hashes in-memory as a fast pre-check, similar to what #424 does with its bloom filter.

### Dedup Report Tool

**Greedy clustering:** Sound approach. The `seen` set prevents O(n²) blowup. One edge case: if document A matches both B and C, but B and C don't match each other, greedy clustering puts them in the same cluster. This is fine for diagnostics but worth noting in the tool description.

**`n_results=min(10, len(all_data["ids"]))`** — Good defensive coding. Without the min, ChromaDB throws if you request more results than exist.

**`similarity = round(1 - dist, 3)`** — This assumes ChromaDB's default distance metric is L2/cosine where `1 - dist` ≈ cosine similarity. With the default `all-MiniLM-L6-v2` model this works, but it's worth noting the assumption in a comment since other embedding functions may use different distance scales.

### The closing note

> *a señal de pago — a down payment for the ideas we've borrowed*

Appreciated sentiment. The code quality backs it up. Good test matrix, clean implementation, correct scope.

Overall: strong contribution. The wing-scoped dedup tweak would make it even better for multi-agent deployments.

igorls:

PR #285, opened 2026-04-08
docs: add software engineering memory example

fuzzymoomoo:
This PR adds a small example document showing how MemPalace can be used for software engineering memory.

The goal is to make one practical use case more concrete:

- preserving design and architecture context
- organizing decisions, incidents, and handoffs
- retrieving the right memory during planning, implementation, and validation

Why this shape:

- the repo already has an `examples/` directory for focused setup and workflow docs
- software engineering is a strong fit for MemPalace because so much important project knowledge now lives in conversations, notes, and scattered documents
- this example stays generic and local-first without introducing any external dependency or private workflow assumptions

What is included:

- one example project layout using wings, halls, and rooms
- examples of engineering artifacts that fit durable memory well
- a simple breakdown of useful retrieval by stage

No behavior changes. Docs only.


web3guru888 (commented):
This is exactly the kind of documentation MemPalace needs — a concrete, practical example that bridges the gap between "here are the commands" and "here's how to actually structure memory for a real project."

What I especially like:

1. **The hall taxonomy** (facts, events, discoveries, preferences, advice) maps cleanly to how engineering knowledge actually accrues. This mirrors patterns we've seen in our own integration where we organize 208 discoveries across domain-specific wings — the structure makes retrieval dramatically more useful than flat search.

2. **The handoff section** is gold. The outgoing/incoming actor protocol (save → write handoff → refresh → search → validate) is the exact workflow that makes memory practical for multi-actor development. We've implemented a similar pattern for agent-to-agent handoffs and can confirm it works.

3. **"Memory stops being a flat pile of search results and starts becoming navigable"** — this line captures the core value proposition perfectly.

4. The stage-based retrieval guidance (Planning → Implementation → Validation) gives users a mental model for *when* to reach for memory, not just *how*.

One thought: a small concrete example showing an actual drawer content snippet (maybe a 3-line decision record or a handoff note) might help readers who learn by seeing real data. But that's a nice-to-have — this is great as-is.

Really glad to see this added to the examples. 🙌

PR #286, opened 2026-04-08
fix: close ChromaDB clients before cleanup to prevent PermissionError…

### LINKED ISSUE #275: bug: Windows compatibility gaps in tests and CI
﻿**What happened?**

I validated MemPalace on native Windows (Windows 11, Python 3.12) and found a few Windows-specific compatibility gaps in the current upstream setup.

The main reproducible issue is that tests which open `chromadb.PersistentClient(path=...)` can leave file handles open on Windows. That causes temp directory cleanup to fail with `PermissionError: [WinError 32]` when `shutil.rmtree(...)` runs.

I saw this in the upstream suite before applying a local runtime workaround:

- `tests/test_convo_miner.py::test_convo_mining`
- `tests/test_miner.py::test_project_mining`

The failure was on Chroma files such as `data_level0.bin` still being in use during cleanup.

I also noticed that current CI only runs on Ubuntu, so this class of problem is invisible in CI right now.

Separately, the convo miner test triggered a download of the ONNX MiniLM model on first run, which means tests currently do require network access in at least one path, despite `CONTRIBUTING.md` saying tests should run without network access.

**What did you expect?**

- `pytest tests/ -v` should pass on supported Windows Python environments without temp cleanup failures
- CI should cover Windows so platform-specific regressions are caught early
- Tests should run offline once dependencies are installed

**How to reproduce:**

1. Use native Windows (I reproduced on Windows 11) with Python 3.12
2. Clone the repo and install dev dependencies: `pip install -e ".[dev]"`
3. Run `python -m pytest tests/test_convo_miner.py tests/test_miner.py -v`
4. Observe `PermissionError: [WinError 32]` during `shutil.rmtree(...)` cleanup after `chromadb.PersistentClient(...)` was used

Additional observation:

5. On a fresh environment, run `python -m pytest tests/test_convo_miner.py -v`
6. Observe the ONNX MiniLM model download during the test run

**Environment:**
- OS: Windows 11
- Python version: 3.12
- MemPal version: git SHA `71736a3f4f1afe01982d5b0d4c36eb5405eaab91`

**Notes / likely fixes**

A few upstream changes would probably make Windows support much more robust:

- add a Windows job to `.github/workflows/ci.yml`
- explicitly close/stop Chroma clients in tests before temp directory cleanup
- remove network dependency from tests by avoiding first-run model download in test paths

fuzzymoomoo:
… on Windows

On Windows, chromadb.PersistentClient holds SQLite and .bin file handles open until explicitly released. Relying on `del client` or GC is not sufficient — the OS-level handles remain open, causing shutil.rmtree to raise PermissionError: [WinError 32].

Fix: call client.close() before the cleanup block in:
  - tests/test_miner.py::test_project_mining
  - tests/test_convo_miner.py::test_convo_mining
  - tests/conftest.py collection fixture teardown

Also wraps test_convo_mining in try/finally so tmpdir is always cleaned up even when assertions fail.

Adds a windows-latest CI job so regressions are caught automatically.

Fixes #275. Verified on Windows 11 / Python 3.12.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,6 +34,19 @@ from mempalace.config import MempalaceConfig  # noqa: E402
 from mempalace.knowledge_graph import KnowledgeGraph  # noqa: E402
 
 
+def _close_client(client):
+    """Close Chroma clients across versions."""
+    close = getattr(client, "close", None)
+    if callable(close):
+        close()
+        return
+
+    system = getattr(client, "_system", None)
+    stop = getattr(system, "stop", None)
+    if callable(stop):
+        stop()
+
+
 @pytest.fixture(autouse=True)
 def _reset_mcp_cache():
     """Reset the MCP server's cached ChromaDB client/collection between tests."""
@@ -103,8 +116,10 @@ def collection(palace_path):
     client = chromadb.PersistentClient(path=palace_path)
     col = client.get_or_create_collection("mempalace_drawers")
     yield col
-    client.delete_collection("mempalace_drawers")
-    del client
+    try:
+        client.delete_collection("mempalace_drawers")
+    finally:
+        _close_client(client)  # release file handles before tmp_dir cleanup (required on Windows)
 
 
 @pytest.fixture
--- a/tests/test_convo_miner.py
+++ b/tests/test_convo_miner.py
@@ -5,22 +5,42 @@ import chromadb
 from mempalace.convo_miner import mine_convos
 
 
+def _close_client(client):
+    """Close Chroma clients across versions."""
+    close = getattr(client, "close", None)
+    if callable(close):
+        close()
+        return
+
+    system = getattr(client, "_system", None)
+    stop = getattr(system, "stop", None)
+    if callable(stop):
+        stop()
+
+
 def test_convo_mining():
     tmpdir = tempfile.mkdtemp()
-    with open(os.path.join(tmpdir, "chat.txt"), "w") as f:
-        f.write(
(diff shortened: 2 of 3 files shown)


fuzzymoomoo:
Thanks for the thorough review! Pushed two fixes in the latest commit:

1. **Windows CI now ignores benchmarks** — added `--ignore=tests/benchmarks` to match the Linux job. The benchmark files all use `PersistentClient` without `close()` and would hit WinError 32 on every run.

2. **`conftest.py` teardown now guards `client.close()` with try/finally** — if `delete_collection` raises, `close()` still runs and file handles are released before `shutil.rmtree`.

web3guru888 (commented):
The `client.close()` pattern in test fixtures is a good practice. We hit similar issues in our integration — on Linux the file handle leak is invisible during tests, but causes problems during long-running daemon processes where the ChromaDB client gets garbage collected without explicit close.

One suggestion for the test pattern: the current `test_convo_mining` fixture still has a bare `shutil.rmtree(tmpdir)` outside the try/finally on line 42 (after the test body). If the test fails before reaching the `finally` block that calls `rmtree`, the tmpdir leaks. The fixture-based tests (`conftest.py`) handle this correctly with `yield` + cleanup, but the inline tempdir tests could benefit from wrapping the whole thing in try/finally too.

Also — the Windows CI addition is great. ChromaDB file locking behavior differs significantly between Windows and POSIX.

igorls:

PR #293, opened 2026-04-08
fix+perf: security hardening, performance optimization, and bug fixes

anthonyonazure:
## Summary

Comprehensive audit-driven improvements across security, performance, and correctness — touching 13 files with +501/-229 lines.

### Security
- **Fix critical `_client_cache` bug** — `_get_collection()` was calling methods on `None`, silently breaking all MCP write operations
- **Enforce `_SAFE_NAME_RE`** regex in `sanitize_name()` (was defined but never used)
- **Sanitize all read/query paths** — `kg_query`, `kg_invalidate`, `kg_timeline` now validate inputs
- **WAL log hardening** — lazy init, `chmod 0o600` file permissions
- **Define missing `SKIP_FILENAMES`** — was causing `NameError` crashes in `scan_project()`
- **Deterministic drawer IDs** — content-based SHA256 for true idempotency (was timestamp-based, defeating dedup)
- **Shutdown cleanup** — `try/finally` calling `_kg.close()` to prevent WAL corruption

### Performance
- **Singleton ChromaDB client** — eliminates redundant `PersistentClient` creation across all modules (~2 round-trips saved per MCP call)
- **Pre-fetch mined files** — single batch replaces N per-file queries (500-file mine: 500→1 queries)
- **Batch upserts** — all chunks per file in one ChromaDB call (2,500→500 calls for typical mine)
- **Cached palace graph** with wing-to-rooms adjacency index — O(1) BFS vs O(N²)
- **Layer1 two-pass** — metadata-only first pass, then fetch top 15 docs by ID (massive memory reduction)
- **UNION ALL** for `query_entity(direction="both")` — 1 SQL query instead of 2
- **Pre-compiled regex** — fnmatch patterns, pronoun patterns, gitignore `_match_from_root` memoized with `@lru_cache`
- **Single-pass** `_extract_topics`, `generate_layer1`, eliminated double file reads in `split_mega_files`

### Database
- **Enable `PRAGMA foreign_keys=ON`** in `_conn()` (constraints were defined but never enforced)
- **Composite index** `idx_triples_spo_active` for duplicate-check queries
- **`INSERT ON CONFLICT`** preserving `created_at` (was `INSERT OR REPLACE` which reset timestamps)
- **`NULLS LAST` compatibility** for SQLite <3.30

### Mining performance impact (estimated)
| Metric | Before | After |
|--------|--------|-------|
| ChromaDB round-trips (500-file mine) | ~3,001 | ~502 |
| MCP search round-trips | 3 | 1 |
| Graph traversal (3 calls) | 3 full scans | 0 (cached) |
| BFS neighbor lookup | O(N²) | O(1) |

## Test plan
- [x] `ruff check` — all clean
- [x] `ruff format` — all clean
- [x] `pytest tests/ -v` — 115 passed, 2 failed (pre-existing Windows `shutil.rmtree` PermissionError from ChromaDB file locks during test cleanup — not related to changes)
- [x] Fixes 10 previously broken tests (`SKIP_FILENAMES` NameError)

### DIFF
--- a/hooks/mempal_save_hook.sh
+++ b/hooks/mempal_save_hook.sh
@@ -64,13 +64,20 @@ MEMPAL_DIR=""
 # Read JSON input from stdin
 INPUT=$(cat)
 
-# Parse fields from Claude Code's JSON
-SESSION_ID=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('session_id','unknown'))" 2>/dev/null)
-# Sanitize SESSION_ID to prevent path traversal (only allow alnum, dash, underscore)
-SESSION_ID=$(echo "$SESSION_ID" | tr -cd 'a-zA-Z0-9_-')
-[ -z "$SESSION_ID" ] && SESSION_ID="unknown"
-STOP_HOOK_ACTIVE=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('stop_hook_active', False))" 2>/dev/null)
-TRANSCRIPT_PATH=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('transcript_path',''))" 2>/dev/null)
+# Parse all fields in a single Python call (3x faster than separate invocations)
+eval $(echo "$INPUT" | python3 -c "
+import sys, json
+data = json.load(sys.stdin)
+sid = data.get('session_id', 'unknown')
+sha = data.get('stop_hook_active', False)
+tp = data.get('transcript_path', '')
+# Shell-safe output — only allow alphanumeric, underscore, hyphen, slash, dot, tilde
+import re
+safe = lambda s: re.sub(r'[^a-zA-Z0-9_/.\-~]', '', str(s))
+print(f'SESSION_ID=\"{safe(sid)}\"')
+print(f'STOP_HOOK_ACTIVE=\"{sha}\"')
+print(f'TRANSCRIPT_PATH=\"{safe(tp)}\"')
+" 2>/dev/null)
 
 # Expand ~ in path
 TRANSCRIPT_PATH="${TRANSCRIPT_PATH/#\~/$HOME}"
@@ -106,9 +113,6 @@ PYEOF
 else
     EXCHANGE_COUNT=0
 fi
-
-# Sanitize session ID — strip path separators to prevent directory traversal
-SESSION_ID=$(echo "$SESSION_ID" | tr -cd 'a-zA-Z0-9_-')
 
 # Track last save point for this session
 LAST_SAVE_FILE="$STATE_DIR/${SESSION_ID}_last_save"
--- a/hooks/mempal_save_hook.sh
+++ b/hooks/mempal_save_hook.sh
@@ -83,6 +83,7 @@ if [ "$STOP_HOOK_ACTIVE" = "True" ] || [ "$STOP_HOOK_ACTIVE" = "true" ]; then
 fi
 
 # Count human messages in the JSONL transcript
+# SECURITY: Pass transcript path as sys.argv to avoid shell injection via crafted paths
 if [ -f "$TRANSCRIPT_PATH" ]; then
     EXCHANGE_COUNT=$(python3 - "$TRANSCRIPT_PATH" <<'PYEOF'
 import json, sys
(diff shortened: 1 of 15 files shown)


web3guru888 (commented):
Great PR — this is one of the most thorough audit-driven contributions I've seen on MemPalace. The scope is huge (13 files, security + perf + correctness), so here's a detailed review from someone running MemPalace at scale (208 discoveries, 710 KG entities, 1,014 triples).

## Security — Strong Overlap With Our Work

We submitted #320 which fixes the same shell injection vector in `mempal_save_hook.sh`. Your approach (single Python call with `eval` + `safe()` regex) is more consolidated than our original fix but introduces a subtlety: the `eval $(python3 ...)` pattern still trusts the Python output. If the `re.sub` regex misses something, you've piped it straight into `eval`. Consider quoting more defensively or using `read` instead of `eval`:

```bash
read SESSION_ID STOP_HOOK_ACTIVE TRANSCRIPT_PATH <<< $(echo "$INPUT" | python3 -c "...")
```

That said, the actual sanitizer (`sanitize_name()` in config.py, WAL logging, `chmod 0o600`) is excellent. We independently built similar guards — our integration validates all KG inputs and palace names before they touch SQLite or ChromaDB. The `_SAFE_NAME_RE` enforcement is the right call; it was defined but never used before.

## Performance — Impressive Gains

The performance table speaks for itself: 3,001→502 ChromaDB round-trips for a 500-file mine. Specific observations:

1. **Singleton ChromaDB client** (`palace.py`): We independently built this exact pattern. One `PersistentClient` per path, cached in a dict. The `get_mined_files()` pre-fetch for O(1) skip checks is especially good — we do the same thing.

2. **Batch upserts** in `miner.py` and `convo_miner.py`: Switching from per-chunk `add()` to batch `upsert()` is a massive win. Note: the `convo_miner.py` change does `collection.upsert()` outside the try/except — if the batch fails, you'll get an unhandled exception instead of the previous per-chunk error handling. Minor but worth a safety net.

3. **Metadata cache with TTL** in `mcp_server.py`: Smart approach. The 5-minute TTL with invalidation on writes is pragmatic. One thing: `_meta_cache` isn't protected by `_cache_lock` during reads in `_get_cached_metadata()` — the lock is defined but never used. Race condition risk if MCP server handles concurrent requests.

4. **`heapq.nlargest`** for diary reads: Clean optimization, avoids full sort for top-N.

5. **LRU-cached `_match_from_root`** and pre-compiled regex: Good micro-optimizations that compound across large projects.

## Knowledge Graph — Key Improvements

- **`PRAGMA foreign_keys=ON`**: Critical fix. Constraints were defined but silently unenforced — we hit this same issue in our integration.
- **`INSERT ON CONFLICT` preserving `created_at`**: The old `INSERT OR REPLACE` was silently losing temporal data. Important for anyone doing bi-temporal queries like we do.
- **`UNION ALL` for bidirectional queries**: Halving KG query count. Our integration makes heavy use of `direction="both"` (1,014 triples), so this matters.
- **`NULLS LAST` compatibility**: Good defensive coding for older SQLite. The `CASE WHEN` workaround is correct.

## Potential Issues

1. **`conn.row_factory = sqlite3.Row`** on the singleton connection: This changes how *all* queries return results. The test file patches `_client`, `_collection_cache`, and `_meta_cache` — make sure existing tests that index rows by position (`row[10]`) all got updated to named access (`row["obj_name"]`).

2. **Removing `json.dumps(result, indent=2)`** in the MCP response: This makes MCP tool responses unindented. Some MCP clients (including Claude) may display these inline, and the readability hit is noticeable. Was this intentional for performance?

3. **The `pyproject.toml` change** (`pyyaml>=6.0` → `pyyaml>=6.0,<7`): Good defensive pinning, though PyYAML 7 doesn't exist yet. Worth noting for the changelog.

4. **Missing `SKIP_FILENAMES`** fix: The PR description mentions this caused `NameError` crashes in `scan_project()`. The diff shows the variable is now defined — but the old code must have been referencing it without defining it. This alone fixes 10 tests, which tells you how broken things were.

Overall this is a high-quality contribution. The security + performance combination is rare in a single PR. Would recommend splitting into 2-3 PRs for easier review (security, performance, KG), but the code quality is solid.

bensig:
Superseded by #387 which landed the security hardening. Thanks for the audit work @anthonyonazure — your #252 was the foundation we built on.

PR #294, opened 2026-04-08
feat: optional remote ChromaDB support via HttpClient

cypromis:
## Summary

- Adds a \`palace_db\` factory module (\`get_client\` / \`get_collection\`) as the single ChromaDB access point — returns \`HttpClient\` when a host is configured, \`PersistentClient\` otherwise
- Introduces three new config keys (\`chroma_host\`, \`chroma_port\`, \`chroma_ssl\`) readable from \`~/.mempalace/config.json\` and env vars (\`MEMPALACE_CHROMA_HOST\`, \`MEMPALACE_CHROMA_PORT\`, \`MEMPALACE_CHROMA_SSL\`), with env vars taking priority
- Replaces all direct \`chromadb.PersistentClient(...)\` calls across \`miner.py\`, \`convo_miner.py\`, \`searcher.py\`, \`layers.py\`, \`mcp_server.py\`, and \`cli.py\` with \`palace_db.get_client()\` / \`palace_db.get_collection()\`
- Adds \`mempalace remote status\` CLI command: prints local/remote mode and — in remote mode — calls \`.heartbeat()\` to confirm server reachability
- \`HttpClient\` instances are cached by \`(host, port, ssl)\` key to avoid redundant HTTP round-trips on every MCP tool call
- Zero breaking changes: no host configured = identical behaviour to current upstream

## v1 scope and known limitations

This PR targets **one developer across multiple workstations**. All clients share a single \`mempalace_drawers\` collection with no per-user namespacing or authentication.

| Scenario | Works? | Notes |
|---|---|---|
| 1 dev, multiple workstations | ✅ Perfect fit | Same person's memories sync across machines |
| Multiple devs, shared team palace | ✅ Works | All memories pooled — intentional for a shared assistant |
| Multiple devs, each wanting isolated memory | ❌ Not in this version | Planned for a follow-up PR |

**ID collision risks when sharing between multiple users:**
- **\`mine\` / \`convo-mine\`**: IDs hashed from \`source_file + chunk_index\` — two users with the same file path silently overwrite each other
- **Diary entries**: second-resolution timestamp in ID — same-second writes from two users collide
- **Manual \`add_drawer\`**: content-addressed — identical content deduplicates cleanly, different content coexists fine

For single-dev multi-workstation use, collisions are harmless (re-mining the same file is idempotent). Per-user isolation — collection namespacing and optional authentication — is out of scope here and planned for a follow-up PR.

## ChromaDB version compatibility

The Docker Compose snippet is pinned to \`chromadb/chroma:0.6.3\` to match this project's client dependency. ChromaDB 1.0 introduced a breaking change to the collection configuration JSON format (a \`_type\` field the 0.6.x client does not understand), causing \`KeyError: '_type'\` when creating or opening collections. Do not use \`chromadb/chroma:latest\` until MemPalace upgrades its client dependency.

## Test plan

- [ ] \`pytest tests/\` — 127 tests pass, 0 failures
- [ ] \`ruff check . && ruff format --check .\` — clean
- [ ] \`mempalace remote status\` with no host → prints \`Mode: LOCAL\`
- [ ] \`MEMPALACE_CHROMA_HOST=localhost mempalace remote status\` with \`chromadb/chroma:0.6.3\` running → prints \`Mode: REMOTE\` + heartbeat OK
- [ ] Local mine/search unchanged from upstream behaviour


cypromis:
Thanks for the review. Addressing each point:

**#1 — PersistentClient not cached (fixed):** Confirmed regression. Added `_persistent_clients = {}` cache in `palace_db.py` mirroring the existing `_http_clients` cache. Both client types are now cached by their lookup key.

**#2 — `palace_graph.py` not migrated (out of scope):** `palace_graph.py` is explicitly excluded from this PR's scope (CLAUDE.md: *"Do not touch AAAK, knowledge_graph.py, palace_graph.py, or dialect.py — those are out of scope"*). Agreed this is a gap — it should be a follow-up PR. Noted in the PR description's known limitations.

**#3 — Dead globals in `mcp_server.py` (fixed):** Removed `_client_cache = None`, `_collection_cache = None`, and the `global` declaration from `_get_collection`. These were leftover from before the factory migration.

**#4 — Subparser dispatch (fixed):** Added a guard in `cmd_remote` that rejects unknown subcommands, making the subparser meaningful rather than decorative. Kept the subparser itself since `mempalace remote status` is documented and tested.

**#5 — `DEFAULT_COLLECTION` alias:** The rename is intentional — `palace_db.DEFAULT_COLLECTION` is the public-facing name callers use; `DEFAULT_COLLECTION_NAME` stays the canonical name in `config.py`. No change.

**#6 — Missing `sys` import:** False positive — `sys` is already imported at `cli.py:30`.

web3guru888 (approved):
Really nice refactor — the `palace_db` factory is clean, well-tested, and the migration is thorough. A few observations:

### 1. Graph tool descriptions are stale after `palace_graph.py` migration

The MCP tool descriptions for `mempalace_traverse`, `mempalace_find_tunnels`, and `mempalace_graph_stats` were annotated with "reads from local palace only" (commit `8b92791`), but `palace_graph.py` was subsequently migrated to use `palace_db.get_collection()` (commit `09d070b`). Since the graph tools now go through the factory and respect remote mode, the "local only" caveat in the descriptions is inaccurate. Should be easy to remove.

### 2. `get_or_create_collection` called on every `get_collection()` — remote mode overhead

In remote mode, `palace_db.get_collection()` calls `client.get_or_create_collection()` which is an HTTP round-trip on every invocation. The client is cached but the collection object is not. For the MCP server processing many tool calls, this adds an HTTP call per dispatch. For v1 this is probably fine (the benchmark shows ~354µs locally), but for remote mode over network it could add up. A collection cache (similar to the client cache) would help.

### 3. `uv.lock` diff includes unrelated packages

The lock file changes include `autocorrect` and `coverage` — looks like a full `uv lock` sync that pulled in transitive deps unrelated to this PR. Not harmful but adds ~250 lines of noise. Consider regenerating the lockfile in a separate commit or PR to keep the diff focused.

### 4. ChromaDB version compatibility note

The PR correctly pins Docker to `chromadb/chroma:0.6.3` to match the current `chromadb>=0.5.0,<0.7` dependency. ChromaDB stable is now 1.5.5 and mixing 0.6.x client with 1.x server causes `KeyError: '_type'` — the README warning about this is good. Might be worth making the version constraint even more prominent since users' first instinct will be `docker pull chromadb/chroma:latest`.

Overall: solid work. The factory pattern is the right abstraction, the test coverage is good, and the README documentation is thorough. The issues above are minor and shouldn't block merge.

bensig:
Conflicts with main. Remote ChromaDB and backend abstraction are being discussed in #413 and #574. We'll pick one direction — a focused rebase after that decision would be welcome.

Issue #19, opened 2026-04-07
Super slow mine on windows


idanf-glbe:
**What happened?**
The mine process is SUPER slow. im trying to index a relatively small, relatively new project on a windows machine.
Running the mine detected 3276 files.
45 minutes in and the process shows 60/3276 files are indexed. 
In this rate the inital index will take 41 hours

**What did you expect?**
For it to work faster

**How to reproduce:**

1. running mempalace mine . on a nextJS project

**Environment:**
- OS: windows 11 business 
- Python version:  Python 3.13.1
- MemPal version: mempal-1.0.1

syedadeel2:
Yes, its sooooo slow, this should have been written in RUST or GO, Python itself is very slow.

dominosaurs:
It might be worth considering that NTFS can be quite inefficient when reading a large number of small files due to high filesystem overhead. Combined with Python's I/O operations and sequential ChromaDB upserts on Windows, this often creates a massive bottleneck compared to Unix systems. Batching the ChromaDB upserts should significantly speed up the mining process.

PR #298, opened 2026-04-08
fix: batch ChromaDB upserts and parallelize file processing for Windo…

RasimAbiyev:
## What does this PR do?
Fixes the extremely slow `mempalace mine` performance on Windows (issue #19).
Two bottlenecks identified and fixed:
1. Each chunk was written to ChromaDB in a separate upsert — replaced with a single batch upsert per file (`add_drawers_batch()`)
2. Files were processed one by one — replaced with `ThreadPoolExecutor` (up to 4 workers)

## How to test
Run `mempalace mine .` on a mid-size project and compare speed before/after.
Expected: significantly fewer ChromaDB round-trips, faster overall indexing.

## Checklist
- [ ] No hardcoded paths


web3guru888 (commented):
Batch upserts are the right fix for the per-chunk round-trip overhead. The performance improvement will be significant for large projects. A few concerns:

**Batch upsert (`add_drawers_batch`):**
- Uses `collection.add()` instead of `collection.upsert()` — this means re-mining the same file will raise a duplicate ID error. The catch for `"already exists"` handles it, but returns 0 (entire batch lost) even if only one chunk was a duplicate. Consider using `upsert()` instead, which is idempotent, or falling back to per-chunk adds when the batch fails.
- The function builds all IDs/docs/metadatas in memory before calling `add()`. For very large files (e.g., a 10MB transcript that produces 200 chunks), this is fine — but worth noting there's no batch-size cap. ChromaDB can handle it, but documenting a max batch size would be good hygiene.

**Thread safety concerns with `ThreadPoolExecutor`:**
- `process_file()` now returns just `drawers` (int) instead of `(drawers, room)`. The room is then re-detected in the main loop via `detect_room(filepath, "", rooms, project_path)` — but with an empty string for `content`. This means room detection in the summary will be different from the room used during actual processing (where the file content was available). The `room_counts` dict will be inaccurate.
- `total_drawers` and `files_skipped` are modified from the main thread after `as_completed()`, so no race condition there — that's correct.
- `print()` from multiple threads will interleave output. Consider collecting results and printing in order, or using a lock.

**`workers = min(4, os.cpu_count() or 1)`:**
- 4 workers is reasonable. But ChromaDB's `PersistentClient` uses SQLite under the hood, and SQLite has a single-writer lock. Multiple threads calling `collection.add()` concurrently will serialize on the write lock anyway. The parallelism helps with file I/O and chunking, but the ChromaDB writes won't truly parallelize. Worth measuring actual speedup — might want to separate the read/chunk phase from the write phase.

The batch upsert alone is a solid improvement. I'd suggest separating it from the threading change — batch upsert is safe and high-impact, threading needs more testing for correctness.

RasimAbiyev:
Hi @web3guru888, thank you for the review!

Addressed your feedback:

1. **upsert restored**: Switched back to `collection.upsert()` in `add_drawers_batch()` — idempotent and correct since `file_already_mined()` already guards against unnecessary re-processing.
2. **Room detection fixed**: `process_file()` now returns `(drawers, room)` so the main loop uses the actual room detected during processing — no more empty-string re-detection in the summary.

Ready for re-review!

bensig:
Conflicts with main. Batch upserts and parallel processing are being addressed in other PRs (#507, #562's pieces). Thanks for the Windows work.

PR #299, opened 2026-04-08
fix: lower check_duplicate default threshold from 0.9 to 0.5 (fixes #…

### LINKED ISSUE #247: mempalace_check_duplicate misses existing text at default threshold 0.9
`mempalace_check_duplicate` misses clearly existing content at its default threshold of `0.9`.

I tested it against text that already exists in the palace, including exact sentences copied from indexed content, and still got:

{
  "is_duplicate": false,
  "matches": []
}

Example 1:
马丁·海德格尔（Martin Heidegger）出生于德国巴登——符腾堡（Baden-Württemberg）梅斯基尔希的一个贫寒的天主教家庭中。

This only returned `true` after lowering the threshold to `0.4`.

Example 2:
本标准使用重新起草法参考 ISO 690:2010(E)《信息和文献 参考文献和信息资源引用指南》编制，与 ISO 690:2010 的一致性程度为非等效。

This only returned `true` after lowering the threshold to `0.15`.

So on real indexed content, the default `0.9` appears too high to detect even exact existing text.

fuzzymoomoo:
Fixes #247

The previous default of 0.9 only caught near-exact copies. Real-world testing (reported in #247) showed:

- full stored chunk: similarity `1.0`
- half chunk: `~0.74`  
- quarter chunk: `~0.59`

`similarity = 1 - cosine_distance`, so 0.9 missed everything except exact matches.

**New default: 0.5** — catches partial overlaps and semantically related content while remaining tuneable. Also updates the tool description to explain what the similarity scale means so callers can make informed threshold choices.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -209,7 +209,7 @@ def tool_search(query: str, limit: int = 5, wing: str = None, room: str = None):
     )
 
 
-def tool_check_duplicate(content: str, threshold: float = 0.9):
+def tool_check_duplicate(content: str, threshold: float = 0.5):
     col = _get_collection()
     if not col:
         return _no_palace()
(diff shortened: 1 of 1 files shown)


web3guru888 (commented):
The default threshold change makes sense. The updated description is a clear improvement — the previous docs didn't explain what the threshold actually measured.

For context: our integration wraps `tool_check_duplicate` with a tiered approach — hard dedup at 0.86 (near-exact) and soft dedup at 0.55 (semantic overlap). We always pass the threshold explicitly, so the default change doesn't affect us. But the new 0.5 default is much more useful for first-time users who don't know to tune it. The old 0.9 default was essentially useless for partial duplicates.

The description calibration (~0.7 = partial overlap, ~0.5 = semantically related) matches our real-world experience across 208 discoveries.

PR #300, opened 2026-04-08
fix: make file_already_mined wing-scoped; add .al .rst .xlf .ps1 exte…

### LINKED ISSUE #278: Add .al extension support and make duplicate detection wing-aware
### 1. Missing `.al` file extension in `READABLE_EXTENSIONS`

The miner's `READABLE_EXTENSIONS` whitelist in `miner.py` does not include `.al` (Microsoft Dynamics 365 Business Central / AL Language) files. This means BC codebases with thousands of `.al` files are almost entirely skipped during mining — only `.json`, `.yaml`, `.md` etc. are picked up.

A typical BC repo has 16,000-18,000 `.al` files but the miner only finds ~1,100 files (the non-AL config/docs files).

**Suggestion:** Add `.al` (and consider `.xlf`, `.ps1`) to `READABLE_EXTENSIONS` in `miner.py`. Ideally make this configurable via `mempalace.yaml` so users can extend the list per-project without patching source.

### 2. Duplicate detection is cross-wing, should be wing-aware

`file_already_mined_sqlite()` in `sqlite_fallback.py` checks:

```sql
SELECT 1 FROM drawers WHERE source_file = ? LIMIT 1
```

This is a global check across all wings. When mining the same repo into different wings (e.g., different git branches into `bc27_be` vs `bc28_be`), the second mine is entirely skipped because the same absolute file paths were already filed under the first wing.

**Use case:** A git repo with multiple branches representing different product versions. Each branch should be mined into its own wing for version-specific search/reference.

**Suggestion:** Make the duplicate check wing-aware:

```sql
SELECT 1 FROM drawers WHERE source_file = ? AND wing = ? LIMIT 1
```

Same change needed for the ChromaDB path (non-Windows) using `$and` filter.

## Environment

- Windows 11
- MemPalace with SQLite fallback (Windows path)
- Repository: MSDyn365BC.Code.History with `be-27` and `be-28` branches

fuzzymoomoo:
Fixes #278

## Wing-aware dedup

`file_already_mined()` previously queried only by `source_file`, so mining the same repo into two wings (e.g. two git branches) would skip the second wing entirely — the first mine's mtime match short-circuits it.

Fix: adds an optional `wing` parameter. When provided, the `WHERE` filter includes `{"wing": wing}` so each wing tracks its own mined state. Default `wing=""` preserves the original behaviour for any callers that don't pass a wing.

## Extension additions

Added to `READABLE_EXTENSIONS`:
- `.rst` — reStructuredText documentation
- `.al` — Microsoft Dynamics 365 / AL Language (BC repos have 16k–18k `.al` files)
- `.xlf` — XLIFF localisation files  
- `.ps1` — PowerShell scripts

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -274,7 +274,7 @@ def mine_convos(
         source_file = str(filepath)
 
         # Skip if already filed
-        if not dry_run and file_already_mined(collection, source_file):
+        if not dry_run and file_already_mined(collection, source_file, wing=wing):
             files_skipped += 1
             continue
 
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -41,7 +41,7 @@ READABLE_EXTENSIONS = {
     ".csv",
     ".sql",
     ".toml",
-    ".al",   # Microsoft Dynamics 365 / AL Language
+    ".al",  # Microsoft Dynamics 365 / AL Language
     ".xlf",  # XLIFF localisation files
     ".ps1",  # PowerShell scripts
 }
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -22,6 +22,7 @@ from .palace import SKIP_DIRS, get_collection, file_already_mined
 READABLE_EXTENSIONS = {
     ".txt",
     ".md",
+    ".rst",
     ".py",
     ".js",
     ".ts",
@@ -40,6 +41,9 @@ READABLE_EXTENSIONS = {
     ".csv",
     ".sql",
     ".toml",
+    ".al",   # Microsoft Dynamics 365 / AL Language
+    ".xlf",  # XLIFF localisation files
+    ".ps1",  # PowerShell scripts
 }
 
 SKIP_FILENAMES = {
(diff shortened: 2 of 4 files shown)


web3guru888 (commented):
This fix is directly relevant to our multi-domain setup. We mine 5 discovery domains (astrophysics, economics, climate, epidemiology, cryptography) into separate wings, and without wing-scoped `file_already_mined()`, re-mining the same source into a second wing would silently skip it.

**The core fix is clean:**
```python
# Wing-scoped WHERE filter
if wing:
    where_filter = {"$and": [{"source_file": source_file}, {"wing": wing}]}
else:
    where_filter = {"source_file": source_file}
```

Good decisions:
- **Default `wing=""`** preserves backward compatibility for callers that don't pass a wing ✅
- **Both miners updated** — `convo_miner.py` and `miner.py` both pass `wing=wing` now ✅
- **Test validates the scoping** — `test_file_already_mined_is_wing_scoped` checks that the same file is "mined" in wing alpha but not beta ✅

**Extension additions** — `.rst`, `.al`, `.xlf`, `.ps1` are all practical. The `.al` addition for Dynamics 365 repos (16k-18k files) is a nice niche catch. The scan test covering all four extensions is thorough.

One minor note: the `$and` filter works correctly with ChromaDB ≥0.4, but for palaces with very large drawer counts, the compound `WHERE` might be slightly slower than a single-field filter. In practice this is negligible since `limit=1` keeps the query cheap.

We'll be adopting this in our integration. Great fix! 🎯

igorls:

Issue #100, opened 2026-04-07
Pin chromadb to a tested range, current unpinned dependency can pull crashing 1.x builds


yeager:
## Summary

`mempalace 3.0.0` currently depends on:

```text
Requires: chromadb, pyyaml
```

That leaves pip free to install the latest Chroma release. In my case it pulled `chromadb 1.5.6`, and MemPalace later started crashing with segmentation faults on macOS ARM after successful indexing.

Downgrading to `chromadb 0.6.3` and rebuilding the palace fixed the problem on the same machine.

## Why this matters

This makes installs non-deterministic:
- one user may get an older working Chroma build
- another user may get a newer 1.x build with different behavior
- MemPalace then looks broken even though the immediate fault is in the Chroma layer

At minimum, the package should pin Chroma to a tested range.

## What I observed

Environment:
- MemPalace: 3.0.0
- Python: 3.9.6
- macOS: Darwin 25.4.0 arm64

Behavior with fresh install:
- `pip install mempalace` pulled `chromadb 1.5.6`
- `mempalace init` and `mempalace mine` worked
- `mempalace status`, `mempalace search`, and `mempalace wake-up` later segfaulted

Crash path from `faulthandler`:

```text
Fatal Python error: Segmentation fault

Thread ...:
  File ".../site-packages/chromadb/api/rust.py", line 421 in _get
  File ".../site-packages/chromadb/api/models/Collection.py", line 159 in get
  File ".../site-packages/mempalace/miner.py", line 402 in status
```

Workaround that fixed it for me:

```bash
pip install --force-reinstall 'chromadb<1'
# I tested with chromadb==0.6.3
# then rebuild the palace
```

## Requested fix

Please pin `chromadb` to a known-good range in package metadata, for example something like:

```text
chromadb>=0.6,<1
```

Or whatever range you have actually tested.

If there is a real intention to support Chroma 1.x, that should probably be explicit and tested in CI, because right now installs are effectively rolling the dice.

PR #302, opened 2026-04-08
docs: add Python 3.13 and 3.14 classifiers

Milofax:
## Summary

Widens `chromadb` dependency from `>=0.5.0,<0.7` to `>=0.5.0,<2.0` in `pyproject.toml` (and syncs `README.md`), fixing the crash on Python 3.14.

## Problem

chromadb <0.7 uses `pydantic.v1.BaseSettings` which fails on Python 3.14+:

```
pydantic.v1.errors.ConfigError: unable to infer type for attribute "chroma_server_nofile"
```

Python 3.14 dropped backward-compat for pydantic-v1's type inference, making any chromadb version that imports `from pydantic.v1 import BaseSettings` unusable.

## Fix

chromadb 1.x removed the pydantic-v1 dependency entirely. Widening the upper bound to `<2.0` lets pip resolve to chromadb 1.5.7+ on Python 3.14 while keeping 0.5.x/0.6.x valid for older Python installs.

**One-line change in `pyproject.toml`, plus README alignment.**

## Verification

| Check | Result |
|-------|--------|
| `chromadb.PersistentClient` smoke test on 3.14 | ✅ |
| `pytest tests/ -x` (117 tests) with chromadb 1.5.7 on Python 3.14.2 | **117/117 passed** ✅ |
| No API changes needed in mempalace source | ✅ |

Refs: #100

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,8 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
     "Topic :: Scientific/Engineering :: Artificial Intelligence",
     "Topic :: Utilities",
 ]


web3guru888 (commented):
I appreciate the intent — Python 3.14 support is important for forward-compat — but the constraint change from `<0.7` to `<2.0` is quite wide and may cause issues.

**The problem with `<2.0`**: ChromaDB has had breaking API changes between minor versions (the 0.4→0.5 migration broke collection APIs, and 0.5→0.6 changed embedding function defaults). Opening the upper bound to `<2.0` means users could install a future ChromaDB version that breaks MemPalace's `col.get()`, `col.query()`, or `col.add()` calls without any error at install time. The breakage would surface at runtime as confusing exceptions.

**What we do**: We pin `chromadb>=0.6.0,<0.7.0` in our integration. This gives us the latest patches while protecting against API breaks. We've been burned by the version drift before — it's a major community pain point (see #426, #445, #457, #458).

**Suggestion**: A safer approach might be `chromadb>=0.5.0,<0.8` or even `<1.0` — still gives plenty of room for 3.14 compat without opening up to a hypothetical 1.x that could have major API changes. Alternatively, the Python 3.14 compat issue might be better solved by testing against chromadb's latest 0.6.x build on 3.14 and updating the classifiers separately from the version constraint.

Also: the README says `chromadb>=0.4.0` but pyproject.toml had `>=0.5.0` — this PR updates the README to `>=0.5.0` which is the correct alignment, so that part is good.

bensig (approved):

bensig:
hey @Milofax — this conflicts with develop now. could you pls rebase? the fix itself looks good, just needs a fresh merge. thanks!

Milofax:
rebased onto develop. the chromadb commit dropped since develop already widened the constraint further (removed upper bound entirely). this PR now only adds the 3.13/3.14 classifiers — retitled accordingly. let me know if you'd rather close this and have the classifier bump elsewhere.

PR #306, opened 2026-04-08
fix(mcp): redirect stdout to stderr at import to protect JSON-RPC wire

### LINKED ISSUE #225: [bug] mempalace mcp writes startup text to stdout instead of stderr, breaking Claude Desktop JSON parsing
**[bug] mempalace mcp writes startup text to stdout instead of stderr, breaking Claude Desktop JSON parsing**

**Labels:** bug
**Version:** 3.0.0

---

### Environment
- OS: Windows 11 Pro 64-bit
- Python: 3.10
- mempalace: 3.0.0
- Client: Claude Desktop 1.1062.0

---

### Describe the bug

When Claude Desktop launches `mempalace mcp` as an MCP subprocess, it receives repeated JSON parse errors because MemPalace writes human-readable startup text to **stdout** instead of **stderr**. The MCP protocol communicates over stdio, so any non-JSON text on stdout corrupts the message stream.

---

### Errors shown in Claude Desktop UI

```
MCP mempalace: Unexpected token '*', "**********"... is not valid JSON
```

---

### Steps to reproduce

```
mempalace status 2>nul   # all output goes to stdout — confirmed
mempalace status 1>nul   # stderr is empty — confirmed
```

---

### Expected behavior

All startup messages, banners, warnings, and status text should be written to **stderr**. Stdout must carry only valid JSON-RPC messages when running in MCP mode (`mempalace mcp`). Ideally, MCP mode should suppress all non-protocol output entirely.

---

### Suggested fix

```python
if mode == "mcp":
    # redirect all print/logging to stderr
    sys.stdout = sys.stderr  # or use logging.basicConfig(stream=sys.stderr)
    # suppress decorative banners entirely
```

dr-robert-li:
## What does this PR do?

The MCP protocol uses stdout for JSON-RPC. Any `print()` from imported modules — chromadb's posthog telemetry chatter, our own banners, OnnxRuntime EP messages — corrupts the message stream and produces `Unexpected token` errors in Claude Desktop.

Saves the real stdout into `_real_stdout` BEFORE any other import, then redirects `sys.stdout` to `sys.stderr` so all subsequent writes (including import-time chatter from chromadb / posthog / our own modules) land harmlessly on stderr. `main()` uses `_real_stdout` for the JSON-RPC response writes.

Fixes #225.

## How to test

```bash
python -m pytest tests/test_mcp_server.py::test_mcp_server_redirects_stdout_at_import \
                 tests/test_mcp_server.py::test_mcp_server_stdout_emits_only_json -v
```

The first test imports `mempalace.mcp_server`, then prints a marker to `sys.stdout`, and asserts the marker lands on stderr (not stdout). It fails without the fix.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — 119 passed, 0 failed
- [x] No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -17,21 +17,27 @@ Tools (write):
   mempalace_delete_drawer   — remove a drawer by ID
 """
 
-import argparse
-import os
+# Issue #225: save real stdout BEFORE any other import so chatter from
+# chromadb/posthog/etc cannot corrupt the JSON-RPC wire on stdout. Use
+# _real_stdout for protocol responses below.
 import sys
-import json
-import logging
-import hashlib
-from datetime import datetime
+_real_stdout = sys.stdout
+sys.stdout = sys.stderr
 
-from .config import MempalaceConfig
-from .version import __version__
-from .searcher import search_memories
-from .palace_graph import traverse, find_tunnels, graph_stats
-import chromadb
+import argparse  # noqa: E402
+import os  # noqa: E402
+import json  # noqa: E402
+import logging  # noqa: E402
+import hashlib  # noqa: E402
+from datetime import datetime  # noqa: E402
 
-from .knowledge_graph import KnowledgeGraph
+from .config import MempalaceConfig  # noqa: E402
+from .version import __version__  # noqa: E402
+from .searcher import search_memories  # noqa: E402
+from .palace_graph import traverse, find_tunnels, graph_stats  # noqa: E402
+import chromadb  # noqa: E402
+
+from .knowledge_graph import KnowledgeGraph  # noqa: E402
 
 logging.basicConfig(level=logging.INFO, format="%(message)s", stream=sys.stderr)
 logger = logging.getLogger("mempalace_mcp")
@@ -800,8 +806,8 @@ def main():
             request = json.loads(line)
             response = handle_request(request)
             if response is not None:
-                sys.stdout.write(json.dumps(response) + "\n")
-                sys.stdout.flush()
+                _real_stdout.write(json.dumps(response) + "\n")
+                _real_stdout.flush()
         except KeyboardInterrupt:
             break
         except Exception as e:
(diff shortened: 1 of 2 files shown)


bensig:
Reopening — this was closed too quickly and your work deserves proper review. Standing by for merge once we clear the current security baseline. Thank you for the contribution.

web3guru888 (commented):
This is a critical fix for anyone running MemPalace via MCP. The approach is exactly right — capturing `_real_stdout` *before* any imports, then redirecting `sys.stdout → sys.stderr` so all import-time chatter (ChromaDB, posthog telemetry, OnnxRuntime provider messages) lands safely on stderr.

**The technique is sound.** We've hit this exact class of issue: ChromaDB's telemetry and OnnxRuntime's EP selection messages are unpredictable and version-dependent. Catching them at the import level rather than trying to suppress individual sources is the right strategy.

**Code review notes:**

1. **`_real_stdout` placement is correct** — it's saved before `import chromadb`, which is where most of the chatter originates. The `noqa: E402` annotations are necessary and appropriate.

2. **`main()` usage looks right** — JSON-RPC responses go through `_real_stdout.write()` / `_real_stdout.flush()`, bypassing the redirect.

3. **Edge case: `logging.basicConfig(stream=sys.stderr)`** — this line runs *after* the redirect, so `sys.stderr` here is the actual stderr (not a redirected stdout). This is correct, but only because Python's `sys.stderr` is never reassigned in this code. Worth a comment noting this.

4. **Test coverage is good.** The subprocess-based tests verify both the redirect and the JSON-only stdout invariant without relying on import order within the test process.

5. **One concern:** If any code elsewhere does `from sys import stdout` (binding at import time) *before* the redirect happens, that binding would capture the real stdout. This PR handles it correctly since the redirect is at the very top, but it's worth noting in a comment for future maintainers.

Strong fix. Fixes a real usability blocker for Claude Desktop and other MCP clients.

igorls:

igorls:
Thanks for this, and sorry it sat. The stdout/stderr redirect at import landed as a direct commit and is in `mcp_server.py` today, so issue #225 is resolved and the JSON-RPC wire is protected. Closing as superseded — thanks for reporting it.

PR #308, opened 2026-04-08
fix: eliminate sys.argv manipulation in split command (fixes #303)

fuzzymoomoo:
cmd_split previously rebuilt sys.argv and called split_mega_files.main(), which re-parsed it via argparse. This is fragile — any extra content in sys.argv (pytest, Python 3.14 launcher flags, etc.) causes parse_args() to raise 'unrecognized arguments' and the command fails.

Fix: extract the split logic from main() into a new run() function that accepts direct parameters. cmd_split now calls split_run() directly — no sys.argv mutation, no second argparse invocation.

main() is unchanged from the user's perspective: it still parses the same flags (--source, --output-dir, --dry-run, --min-sessions, --file) and delegates to run(). The standalone script interface is unaffected.

## Checklist
- [X] Tests pass (`python -m pytest tests/ -v`)
- [X] No hardcoded paths
- [X] Linter passes (`ruff check .`)

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -128,24 +128,14 @@ def cmd_wakeup(args):
 
 def cmd_split(args):
     """Split concatenated transcript mega-files into per-session files."""
-    from .split_mega_files import main as split_main
-    import sys
+    from .split_mega_files import run as split_run
 
-    # Rebuild argv for split_mega_files argparse
-    argv = ["--source", args.dir]
-    if args.output_dir:
-        argv += ["--output-dir", args.output_dir]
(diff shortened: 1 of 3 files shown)


web3guru888:
This looks like a solid cleanup. Removing the  manipulation and calling a proper python function () is definitely a safer and cleaner approach. LGTM!

web3guru888 (approved):
This looks like a solid cleanup. Removing the sys.argv manipulation and calling a proper python function is definitely a safer and cleaner approach. LGTM!

igorls:

Issue #209, opened 2026-04-08
Index does not respect the config. Room removal does not remove already indexed files, and mining should respect project `.gitignore`


hungpvdev:
**What happened?**
I removed a room from the MemPalace config, but the files that had already been indexed for that room were still present in the palace after running `init` again.

I also noticed that MemPalace still mines files ignored by the project’s `.gitignore`, which causes over-indexing. In my case, files under `storage/` and the fallback `general` room were still counted even though they were no longer in the room list.

**What did you expect?**
- A way to re-index a project after config changes.
- Re-indexing should re-read the current config and apply room additions/removals.
- Files excluded by the project’s `.gitignore` should not be mined.
- If a room is removed, stale indexed content for that room should be removed or rebuilt.

**How to reproduce:**
1. Create a project and run `mempalace init <project>`.
2. Remove a room from the generated config.
   - Screenshot: ![Removed room](https://github.com/user-attachments/assets/1a00ce00-e1a9-41f4-9877-2c6780cf03e6)
3. Run `mempalace mine <project>`.
   - Screenshot: ![Mine output](https://github.com/user-attachments/assets/3f248050-5d69-4b03-a2a7-7887730f8e6d)
4. Observe that `storage/` and `general/` are not in the room list, but the scan still counts files from those folders.
   - Screenshot: ![Room list mismatch](https://github.com/user-attachments/assets/b8a11c53-bf6d-4f2f-bb68-9d58d1d14a15)
5. Run `mempalace init <project>` again.
6. Observe that the previously indexed content from the removed room is still present.

**Suggested improvement**
- Add a re-index command or flag that reloads the current config.
- Re-indexing should remove stale drawers from deleted rooms/files.
- Re-indexing should optionally rebuild the palace from scratch.
- Mining should respect the project’s `.gitignore` to avoid indexing ignored files and folders.

**Environment:**
- OS: Ubuntu 22.04.1 LTS on WSL2
- Kernel: Linux 6.6.87.2-microsoft-standard-WSL2
- Python version: 3.10.12
- MemPal version: 3.0.0
- Shell: fish 3.6.0
- Install location: `/home/hungpv/.local/lib/python3.10/site-packages`

PR #310, opened 2026-04-08
Feat/bulk delete and search drawer

### LINKED ISSUE #224: Stale drawer retrieval can inject contradictory memory into live agent context; no official sync/update workflow exists
## Summary

The core issue here is not just that drawers do not have an update API, or that users need to manually synchronize data.

The bigger problem is that when knowledge has changed, MemPalace can still return stale drawer content through semantic search. Once those stale results are retrieved by an agent and inserted into the current context, outdated or contradictory memory can directly affect live reasoning in Codex / Claude.

This is a memory correctness risk, not just a maintenance inconvenience.

## Reproduction

Example:

1. A drawer is first filed with text containing `status: active`
2. Later, the real status changes to `deprecated`
3. I update the KG:
   - `mempalace_kg_invalidate(... active ...)`
   - `mempalace_kg_add(... deprecated ...)`
4. `mempalace_kg_query` now reflects the new fact
5. But `mempalace_search` can still return the old drawer text containing `status: active`

## Why this matters

The issue is not only that the storage layers become inconsistent. It also affects the quality of the agent's live context:

- the agent retrieves a stale drawer via `mempalace_search`
- that stale drawer enters the current conversation context as verbatim evidence
- the agent continues answering, summarizing, reasoning, or making decisions based on outdated facts
- if that incorrect reasoning is later written back into memory, the error can spread further

In other words, this is not model parameter contamination, but it can absolutely contaminate the live context used in the current session.

## Why this seems easy to trigger in the current design

From the current implementation, this behavior is understandable:

- drawers and the KG are stored separately
- the KG supports temporal fact updates, but drawers have no official update semantics
- MCP only exposes `mempalace_add_drawer` and `mempalace_delete_drawer`
- `mempalace_search` does not return `drawer_id`, so users cannot easily identify and repair stale drawers from search results
- `mine` skips already-filed `source_file`s, so re-running mining is not a refresh mechanism

## Expected behavior

At minimum, one of these two directions should be clearly supported:

### Option 1: Document the current model explicitly
- clearly state that drawers are immutable historical snapshots
- clearly state that the KG is the source of truth for current facts
- clearly document how users are expected to synchronize drawers, KG, and source files when facts change
- clearly warn that partial updates can cause stale content to be injected into live agent context through semantic search

### Option 2: Provide an official sync/update path
- add something like `update_drawer` or `sync_source_file`
- or return `drawer_id` from `mempalace_search`
- or support re-mining a `source_file` with replacement/upsert semantics instead of skipping it
- ideally, consider KG-aware filtering or contradiction warnings before stale content is injected into context

## Why this is important

The current behavior makes it easy for users to assume MemPalace behaves like an updatable memory system or a wiki-like source of truth.

But in practice it behaves more like:

- drawers preserve historical verbatim snapshots
- the KG preserves temporal facts
- the two layers do not automatically stay in sync

If this model is not explicitly documented, users can easily end up with contradictory memory states after partial updates, and those contradictions can directly affect live agent behavior.

## Suggested improvements

- include `drawer_id` in `mempalace_search` results
- add `mempalace_update_drawer` or `mempalace_sync_source_file`
- document the consistency boundary between drawers and the KG in the README and MCP protocol
- add KG conflict detection or at least a warning before potentially stale search results are injected into context

## Question

Is the official MemPalace model currently intended to be:

- drawers as append-only historical memory snapshots
- KG as the current fact layer
- synchronization left to the user

If yes, that model should be made explicit in the docs and tool interfaces.
If not, then the lack of an official drawer sync/update workflow creates a real memory correctness problem.

### LINKED ISSUE #237: Feature Request / Bug: No Storage Limit Handling or Disk-Full Graceful Degradation
### Summary

MemPalace has no documented behavior or safeguards for when local disk storage runs out. Given that the system is designed for long-term, continuous memory accumulation (verbatim raw storage by design), this is a realistic production concern — especially for heavy users mining months of conversations.

### Current Behavior

- No storage quota setting exists in `config.json` or `wing_config.json`
- No documented eviction policy, pruning strategy, or archival mechanism
- No `mempalace status` output for current palace disk usage
- **ChromaDB** (vector store) will throw hard write errors when disk is full — no graceful degradation
- **SQLite** (knowledge graph) behaves identically — hard failure on write
- This likely crashes the MCP server mid-session, with no user-friendly error surfaced to the AI or the user
- The only removal mechanism is `mempalace_delete_drawer` (single entry) — no bulk pruning command exists

### Why This Matters

The README explicitly states:
> *"Six months of daily AI use = 19.5 million tokens"*

Raw verbatim storage scales with usage. A user mining multiple projects, Slack exports, and daily Claude Code sessions will accumulate significant data over time. There is currently no way to:
- Know how large the palace has grown (without running `du -sh ~/.mempalace/` manually)
- Set a size cap
- Automatically prune old or low-value entries
- Archive wings to cold storage
- Recover gracefully if a write fails mid-session

### Expected / Requested Behavior

1. **`mempalace status` should report disk usage** — total palace size, per-wing breakdown
2. **Graceful write failure handling** — if ChromaDB or SQLite write fails due to disk full, surface a clear warning to the user/AI rather than crashing the MCP server
3. **Storage cap option in `config.json`** — e.g. `"max_palace_size_gb": 10` with configurable behavior (warn / stop mining / auto-prune oldest)
4. **Bulk pruning command** — e.g. `mempalace prune --wing myapp --older-than 180d`
5. **Archive command** — export a wing to cold storage and remove from active palace

### Environment

- MemPalace v3.0.0
- ChromaDB (vector store)
- SQLite (knowledge graph)
- Use case: Claude Code MCP integration, continuous daily mining

### Workaround (current)

Users must manually monitor with:
```bash
du -sh ~/.mempalace/
```
And manually delete individual drawers via the MCP tool. This is not sustainable at scale.

emirbartu:
## What does this PR do?

This PR implements bulk deletion capabilities and adds `drawer_id` to search results, fixing three open issues:

1. **Search Results** (Fixes #224)
   - Results now include an `id` field containing the `drawer_id`
   - Enables delete/update workflow after finding content via search

2. **MCP Bulk Delete Tools** (Fixes #237, #209)
   - `mempdelete_wing`: Delete all drawers in a wing
   - `mempdelete_room`: Delete all drawers in a specific room
   - Uses ChromaDB `where` filters (wing, room fields) for efficient bulk operations
   - Returns deleted count for verification

3. **CLI Commands** (Fixes #237)
   - `mempalace delete-wing <wing>` – Delete entire wing
   - `mempalace delete-room <wing> <room>` – Delete specific room
   - Interactive confirmation showing the number of drawers to be deleted
   - `--force` flag to skip confirmation

4. **Documentation**
   - Updated tool count from 19 to 21 (added: `mempdelete_wing`, `mempdelete_room`)
   - Added new tools to MCP tools table
   - Added CLI commands to reference
   - Noted `drawer_id` in search results

## How to test

```bash
# Run all tests
python -m pytest tests/ -v

# Test search drawer_id
python -m pytest tests/test_searcher.py::TestSearchMemories::test_result_includes_drawer_id -v

# Test MCP bulk delete
python -m pytest tests/test_mcp_server.py::TestWriteTools::test_delete_wing -v
python -m pytest tests/test_mcp_server.py::TestWriteTools::test_delete_room -v

# Test CLI (manual)
mempalace delete-wing test-wing --force
mempalace delete-room test-wing test-room --force

## Checklist

- [x] Tests pass (122/122)
- [x] No hardcoded paths

### DIFF
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -50,7 +50,7 @@ Set the `MEMPAL_DIR` environment variable to a directory path to automatically r
 
 ## MCP Server
 
-The plugin automatically configures a local MCP server with 19 tools for storing, searching, and managing memories. No manual MCP setup is required -- `/mempalace:init` handles everything.
+The plugin automatically configures a local MCP server with 21 tools for storing, searching, and managing memories. No manual MCP setup is required -- `/mempalace:init` handles everything.
 
 ## Full Documentation
 
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ Restart Claude Code, then type `/skills` to verify "mempalace" appears.
 claude mcp add mempalace -- python -m mempalace.mcp_server
 ```
 
-Now your AI has 19 tools available through MCP. Ask it anything:
+Now your AI has 21 tools available through MCP. Ask it anything:
 
 > *"What did we decide about auth last month?"*
 
@@ -458,7 +458,7 @@ claude plugin install --scope user mempalace
 claude mcp add mempalace -- python -m mempalace.mcp_server
 ```
 
-### 19 Tools
+### 21 Tools
 
 **Palace (read)**
 
@@ -472,12 +472,16 @@ claude mcp add mempalace -- python -m mempalace.mcp_server
 | `mempalace_check_duplicate` | Check before filing |
 | `mempalace_get_aaak_spec` | AAAK dialect reference |
 
+Search results now include an `id` field for each hit (the drawer ID).
+
 **Palace (write)**
 
 | Tool | What |
 |------|------|
 | `mempalace_add_drawer` | File verbatim content |
 | `mempalace_delete_drawer` | Remove by ID |
+| `mempalace_delete_wing` | Delete all drawers in a wing |
+| `mempalace_delete_room` | Delete all drawers in a room |
 
 **Knowledge Graph**
 
@@ -576,6 +580,10 @@ mempalace search "query"                          # search everything
 mempalace search "query" --wing myapp             # within a wing
 mempalace search "query" --room auth-migration    # within a room
 
+# Delete
+mempalace delete-wing <wing>                       # delete entire wing
+mempalace delete-room <wing> <room>                # delete specific room
+
 # Memory stack
 mempalace wake-up                                 # load L0 + L1 context
 mempalace wake-up --wing driftwood                # project-specific
@@ -630,7 +638,7 @@ Plain text. Becomes Layer 0 — loaded every session.
 | `cli.py` | CLI entry point |
 | `config.py` | Configuration loading and defaults |
 | `normalize.py` | Converts 5 chat formats to standard transcript |
-| `mcp_server.py` | MCP server — 19 tools, AAAK auto-teach, memory protocol |
+| `mcp_server.py` | MCP server — 21 tools, AAAK auto-teach, memory protocol |
 | `miner.py` | Project file ingest |
 | `convo_miner.py` | Conversation ingest — chunks by exchange pair |
 | `searcher.py` | Semantic search via ChromaDB |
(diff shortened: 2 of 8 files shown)


web3guru888 (approved):
LGTM! The code properly implements the requested bulk deletion features (wings and rooms) with proper confirmation prompts when not forced. Returning the DB IDs for search results is useful too. I verified the test coverage ensures wing/room filters and ID return work as expected. 

Minor heads up for the future since stability is a focus rn: it might be valuable to extract ChromaDB specific interaction out of `cli.py` into a separate DB layer at some point so `cli.py` handles exclusively the commands/formatting part.

Otherwise looks clean and ready!

bensig:

PR #319, opened 2026-04-09
docs: Add STAN extension to Community Plugins

web3guru888:
### Description

Following up on feedback in PR #272, this PR adds a link to the standalone STAN (Stigmergic A* Navigation) extension in a new **Community Extensions** section of the README.

The STAN algorithm (A* search with pheromone decay over the MemPalace knowledge graph) has been extracted into an opt-in plugin as suggested, keeping the core lean:

- **Extension repo**: https://github.com/web3guru888/mempalace-stan-extension
- **What it does**: Biologically-inspired causal traversal and optimal pathfinding between concepts in the knowledge graph
- **Install**: `pip install -e .` (works alongside existing MemPalace installation)

This PR only touches `README.md` — adds 4 lines for the Community Extensions section. No code changes, no dependency changes.

Thanks for the guidance on #272!

PR #320, opened 2026-04-09
fix: mitigate shell injection risk in hooks (#110)

### LINKED ISSUE #110: Shell injection risk in hook scripts ($TRANSCRIPT_PATH / $SESSION_ID)
The hook scripts interpolate JSON-derived values directly into shell strings, which is fragile and creates a       theoretical shell injection path.                                                                                     
                                                                                                                        
  Affected files:                                                                                                       
  - hooks/mempal_save_hook.sh                                                                                           
  - hooks/mempal_precompact_hook.sh                                                                                     
                                                                                                                        
  The pattern:                                                                                                       
  python3 -c "
  with open('$TRANSCRIPT_PATH') as f:
                                                                                                                        
  Suggested fix — pass values as arguments instead:                                                                     
  python3 - "$TRANSCRIPT_PATH" <<'EOF'                                                                                  
  import sys                                                                                                            
  path = sys.argv[1]                                                                                                    
  with open(path) as f:                                                                                              
  EOF                                                                                                                   
                                                                                                                        
Same applies anywhere $SESSION_ID or other JSON-parsed values are embedded in shell strings. In practice Claude Code's output is safe, but this pattern is fragile and worth hardening against format changes.

web3guru888:
fixes #110 by correctly sanitizing the session ID. The python calls already handle their system args, but the SESSION_ID variable required defensive bounds checking to ensure it only captures `a-zA-Z0-9_-` rather than unescaped json strings which might interpolate.

### DIFF
--- a/hooks/mempal_precompact_hook.sh
+++ b/hooks/mempal_precompact_hook.sh
@@ -58,6 +58,9 @@ MEMPAL_DIR=""
 INPUT=$(cat)
 
 SESSION_ID=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('session_id','unknown'))" 2>/dev/null)
+# Sanitize SESSION_ID to prevent shell injection — only allow alnum, dash, underscore
+SESSION_ID=$(echo "$SESSION_ID" | tr -cd 'a-zA-Z0-9_-')
+[ -z "$SESSION_ID" ] && SESSION_ID="unknown"
 
 echo "[$(date '+%H:%M:%S')] PRE-COMPACT triggered for session $SESSION_ID" >> "$STATE_DIR/hook.log"
 


web3guru888:
@bensig This PR only touches `hooks/mempal_precompact_hook.sh` — it adds SESSION_ID sanitization (3 lines) to prevent shell injection via crafted session IDs. No `pyproject.toml` changes in this diff.

Happy to rebase if needed, but the branch should be clean against current `main`.

web3guru888:
Closing in favor of #589 — JoeProAI's approach of using the same inline sanitization pattern from `mempal_save_hook.sh` is architecturally cleaner (consistent between hooks, single-pass). Already reviewed and confirmed the implementation there.

Keeping #319 (STAN extension docs) as a separate concern.

PR #322, opened 2026-04-09
fix: Ignore wait_for_previous argument to support Gemini MCP clients

dbshadow:
### Describe the bug
When using Mempalace with the Gemini CLI MCP client, the client automatically injects an undocumented wait_for_previous argument into the tool execution request. This causes a TypeError: unexpected keyword argument in mcp_server.py and crashes the tool execution.

### Solution
Added a safe .pop("wait_for_previous", None) before passing tool_args to the handler. This safely drops the non-standard argument and restores compatibility with Gemini clients without affecting other standard clients like Claude Desktop or Cursor.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -766,6 +766,7 @@ def handle_request(request):
             elif declared_type == "number" and not isinstance(value, (int, float)):
                 tool_args[key] = float(value)
         try:
+            tool_args.pop("wait_for_previous", None)
             result = TOOLS[tool_name]["handler"](**tool_args)
             return {
                 "jsonrpc": "2.0",


web3guru888 (approved):
Clean fix — `.pop()` is the right primitive for this. Gemini shouldn't be injecting undeclared arguments, but servers should be robust to it. ✅

One thought for a follow-up (not blocking this PR): right now this protects against `wait_for_previous` specifically. If Gemini or another MCP client starts injecting *additional* non-standard arguments in the future, we'd need another `.pop()` line.

A more generic guard would be to filter against the declared `inputSchema` properties:

```python
known_keys = set(schema_props.keys())
tool_args = {k: v for k, v in tool_args.items() if k in known_keys}
```

This goes right after the existing type-coercion loop (which already iterates `schema_props`), so it's a natural fit. Trade-off is that it silently drops any unknown arg — could mask real typos — but since we already validate required args via the schema, the risk is low.

Not suggesting this should block the PR — the targeted `.pop()` is fine for now and the fix is needed. Just flagging for future hardening.

bensig (approved):
Code review + security audit clean. One-liner Gemini compat fix.

PR #329, opened 2026-04-09
fix(normalize): fail closed on invalid ChatGPT mapping trees

### LINKED ISSUE #330: ChatGPT mapping imports can silently ingest the wrong branch
## Bug

`mempalace/normalize.py` walks ChatGPT `mapping` trees via `children[0]`.

For exports with regenerated or edited branches, this can silently normalize and mine the wrong conversation path.

This should be treated as a write-path integrity bug, not a normalization enhancement.

## Impact

This is silent bad data in the write path:
- wrong transcript stored
- wrong evidence embedded
- later retrieval can return content from an inactive branch

## Common trigger

This happens in normal ChatGPT usage whenever a conversation branches, for example:
- clicking "Regenerate" on an assistant reply
- editing an earlier message and continuing from the edited path
- continuing a conversation from an older turn instead of the latest one

In those cases, the export can contain multiple children under one node, and following `children[0]` can mine the wrong branch.

## Reproduction

Use a ChatGPT export where one user turn has two assistant children and `current_node` points to the second branch:

```json
{
  "current_node": "a_new",
  "mapping": {
    "root": { "parent": null, "message": null, "children": ["u1"] },
    "u1": {
      "parent": "root",
      "message": {
        "author": { "role": "user" },
        "content": { "parts": ["Tell me a slogan"] }
      },
      "children": ["a_old", "a_new"]
    },
    "a_old": {
      "parent": "u1",
      "message": {
        "author": { "role": "assistant" },
        "content": { "parts": ["Old answer"] }
      },
      "children": []
    },
    "a_new": {
      "parent": "u1",
      "message": {
        "author": { "role": "assistant" },
        "content": { "parts": ["New regenerated answer"] }
      },
      "children": []
    }
  }
}
```

Expected:

- normalize the current_node path
- include New regenerated answer
- exclude Old answer

Actual before fix:

- can follow children[0]
- can ingest Old answer silently

## Expected fix
use current_node when present
reject ambiguous multi-branch trees without current_node
fail closed on invalid/unreachable paths

GaosCode:
Closes #330

## What does this PR do?

Fix ChatGPT `mapping` normalization so imports fail closed instead of silently ingesting the wrong transcript or raw invalid JSON.

This PR:
- stops walking `children[0]` and resolves the transcript from the active node path instead
- uses `current_node` when present, and rejects ambiguous multi-branch trees without it
- rejects invalid ChatGPT exports when `mapping` is malformed, `current_node` is invalid, the resolved path does not connect back to the detected root, or the path is not reachable from the root via `children`
- prevents `mine_convos()` from silently passing invalid ChatGPT exports downstream as plain text
- updates ingest reporting so skipped invalid ChatGPT exports are counted and labeled correctly
- adds regression tests for regenerated branches, edited branches, invalid `current_node`, orphaned subtrees, unreachable hidden nodes, and ingest skip reporting

## How to test

```bash
uv run pytest /Users/mrbrain/code/mempalace/tests/test_normalize.py -q
uv run pytest /Users/mrbrain/code/mempalace/tests/test_convo_miner.py -q -k "ambiguous or invalid or summary"
uv run ruff check .
```

## Checklist
 Tests pass (python -m pytest tests/ -v)
 No hardcoded paths
 Linter passes (ruff check .)

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -15,7 +15,7 @@ from pathlib import Path
 from datetime import datetime
 from collections import defaultdict
 
-from .normalize import normalize
+from .normalize import ChatGPTNormalizeError, normalize
 from .palace import (
     NORMALIZE_VERSION,
     SKIP_DIRS,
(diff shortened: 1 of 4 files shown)


web3guru888 (approved):
Solid fix — the old `children[0]` walk was a silent data corruption vector and this is the right way to handle it.

**What I like:**

1. **Fail closed on invalid data** is the correct design choice. The previous behavior (return `None`, let `mine_convos()` silently pass the raw JSON downstream as plain text) meant corrupted memories could enter the palace without any signal. Raising `ChatGPTNormalizeError` and counting skips in the summary output makes failures visible.

2. **Path resolution from `current_node` back to root** is the correct algorithm — it follows the active branch exactly as ChatGPT intended, rather than guessing via `children[0]`. The reachability check (path must be reachable from root via children edges) catches both orphan subtrees and nodes that only have parent pointers without being listed as children.

3. **Exception hierarchy** (`ChatGPTNormalizeError` → `ChatGPTBranchAmbiguityError`) is clean — callers can catch broadly or narrowly.

4. **Test coverage** is thorough: regenerated branches, edited branches, invalid `current_node`, orphans, unreachable hidden nodes, too-few-messages, and ingest reporting. The multi-turn edit branch test (`test_chatgpt_mapping_uses_active_edit_branch_path`) is a particularly good scenario.

**Minor observation (not blocking):**

`_collect_chatgpt_reachable_ids` gets called twice for multi-branch trees without `current_node` — once inside `_collect_chatgpt_leaf_ids` and once in the main `_try_chatgpt_json` before `_build_chatgpt_path`. Could cache the result, but for typical conversation sizes this is negligible.

GaosCode:
Rebased/merged latest main and resolved conflicts. Targeted tests still pass.

GaosCode:
Hi @bensig, quick follow-up on this PR.

I updated it on top of the latest `develop`, resolved the merge conflicts, and re-ran the targeted tests locally. It should be ready for review now when you have time.

Thanks!

GaosCode:
Hi @igorls, quick follow-up on this PR.

I’ve updated the branch, resolved the latest merge conflicts, and re-ran the targeted tests locally. Could you please take a look when you have time?

Thanks!

PR #334, opened 2026-04-09
fix: normalize incomplete Codex sessions and add coverage

fuzzymoomoo:
Closes #295

This adds Codex-specific normalization coverage and fixes one noisy ingest case that shows up in real Codex session history.

What changed:
- add end-to-end Codex JSONL normalization tests for multi-turn sessions
- verify that response_item noise and malformed lines are ignored
- keep the current session_meta requirement explicit in coverage
- normalize incomplete-but-real Codex sessions with a single user turn instead of falling back to raw JSONL

Why this matters:
- _try_codex_jsonl() currently requires at least two messages before it returns a transcript
- sessions that contain a real user message but no assistant reply yet therefore fall back to raw JSONL
- when those files are mined, session metadata and event noise can leak into retrieval instead of a clean transcript

Validation:
- python -m pytest tests/test_normalize.py -q -> 7 passed
- python -m pytest tests/test_normalize.py tests/test_convo_miner.py tests/test_miner.py -q still hits unrelated Windows file-lock cleanup failures in test_convo_miner.py and test_miner.py


web3guru888 (commented):
Great catch on the incomplete Codex session handling. The core fix is small but impactful:

```python
# Before: required >= 2 messages — incomplete sessions fell back to raw JSONL
if len(messages) >= 2 and has_session_meta:

# After: any real message with session_meta gets normalized
if messages and has_session_meta:
```

This matters because raw JSONL fallback leaks `session_meta`, `response_item`, and `task_started` noise into retrieval — exactly the kind of pollution that degrades search relevance over time.

The test coverage is thorough:
- **Multi-turn session** — verifies noise lines (`response_item`, `task_started`) are stripped ✅
- **Incomplete session (single user turn)** — validates the actual fix ✅
- **No session_meta** — confirms the guard still requires it ✅
- **Malformed lines + empty messages** — edge cases handled gracefully ✅

One question: does the `has_session_meta` guard still make sense as-is? If a Codex JSONL file has user/assistant messages but the `session_meta` line was truncated (e.g., partial write), we'd still fall back to raw. That seems like the right behavior — just confirming the intent.

Solid fix with excellent coverage. This kind of normalization hygiene is exactly what keeps palaces clean at scale.

igorls:

Issue #331, opened 2026-04-09
feat: time-decay scoring for search results — prioritize recent memories


matrix9neonebuchadnezzar2199-sketch:
## Problem

`mempalace_search` ranks results purely by ChromaDB vector similarity. A tech-stack decision from six months ago and a discussion from yesterday are treated with equal weight.

In practice, recent memories are almost always more relevant. The knowledge graph already has `valid_from` / `valid_to` on triples, but the drawer search layer has no concept of time at all.

This gap was partially noted in #224 (stale drawer retrieval), but that issue focuses on KG–drawer consistency. This proposal addresses a different angle: **incorporating time as a scoring signal in search ranking itself.**

## Proposed solution

Add an optional time-decay factor to the search score:

final_score = similarity * decay(age)

Where `decay(age)` is a configurable function (e.g., exponential or linear) that down-weights older drawers. The `mined_at` or `created_at` timestamp already stored in drawer metadata can serve as the time anchor.

### Behavior

- **Default**: decay enabled with a sensible default half-life (e.g., 90 days)
- **Override**: `mempalace_search("query", time_decay=False)` to disable for historical research
- **Tunable**: configurable half-life via `~/.mempalace/config.json` (`"time_decay_half_life_days": 90`)
- **No data loss**: this only affects ranking order, not what is stored or deleted

### Example

| Drawer | Similarity | Age | Decay (90d half-life) | Final score |
|--------|-----------|-----|-----------------------|-------------|
| "Use Clerk for auth" (yesterday) | 0.82 | 1d | 0.99 | 0.81 |
| "Use Auth0 for auth" (6 months ago) | 0.85 | 180d | 0.25 | 0.21 |

Without decay, the older drawer wins. With decay, the recent decision surfaces first — which is almost always what the user wants.

### Implementation scope

- `searcher.py`: apply decay multiplier after ChromaDB returns raw results
- `mcp_server.py`: expose `time_decay` parameter on `mempalace_search`
- `config.py`: add `time_decay_half_life_days` setting
- Tests: verify ranking order changes with decay on/off

## Why this matters

AI agents that retrieve outdated decisions as top results can act on stale context without realizing it. Time-decay is a lightweight, non-destructive way to make recency a first-class signal — complementing the KG's temporal model with equivalent awareness on the search side.

Related: #332 (soft-archive wings — the other half of time-aware memory management)

CryptoKupo:
[...]

PR #336, opened 2026-04-09
feat: soft-archive wings — exclude from search without deleting data (#332)

matrix9neonebuchadnezzar2199-sketch:
Closes #332

## Summary

Add the ability to archive/unarchive wings so they are excluded from search results without deleting any data.

## Problem

When a project ends or becomes inactive, its memories still appear in search results. The only current option is `delete-wing` (#310), which permanently destroys data. Users need a non-destructive way to hide stale wings from active search.

### Config layer (`config.py`)
- `archive_wing(name)` / `unarchive_wing(name)`: toggle `"archived": true` flag in `wing_config.json`
- `get_archived_wings()`: return set of archived wing names
- `load_wing_config()` / `save_wing_config()`: public accessors for `wing_config.json`

### Search layer (`searcher.py`)
- Both `search()` and `search_memories()` now exclude archived wings by default using `{"wing": {"$ne": ...}}` in the ChromaDB `where` filter
- When a specific `--wing` is requested, archive filtering is skipped (explicit intent)
- `search_memories()` accepts `include_archived=True` to override exclusion

### MCP layer (`mcp_server.py`)
- `mempalace_archive_wing` / `mempalace_unarchive_wing`: new tools
- `mempalace_search`: new `include_archived` parameter
- `mempalace_status`: shows archived wings separately

### Data safety
- Zero data deletion — one metadata flag in `wing_config.json`
- Fully reversible with `unarchive`
- Archived wings remain accessible via `include_archived=True`

## Testing

- 11 new tests in `tests/test_archive.py` covering:
  - Archive/unarchive state transitions
  - Idempotency (double archive, unarchive non-archived)
  - Preservation of existing wing config fields
  - wing_config.json I/O including corrupt file handling
- All existing tests pass (99 passed, 2 pre-existing Windows-only failures unrelated to this change)

Related: #331 (time-decay scoring — the other half of time-aware memory management)

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -169,6 +169,38 @@ class MempalaceConfig:
         wc = self.load_wing_config()
         return {name for name, cfg in wc.items() if cfg.get("archived") is True}
 
+    def archive_room(self, wing: str, room: str):
+        """Archive a specific room within a wing (exclude from search, keep data)."""
+        wc = self.load_wing_config()
+        if wing not in wc:
+            wc[wing] = {}
+        if "archived_rooms" not in wc[wing]:
+            wc[wing]["archived_rooms"] = []
+        if room not in wc[wing]["archived_rooms"]:
+            wc[wing]["archived_rooms"].append(room)
+            self.save_wing_config(wc)
+            return True
+        return False
+
+    def unarchive_room(self, wing: str, room: str):
+        """Restore a room to active search results."""
+        wc = self.load_wing_config()
+        if wing in wc and "archived_rooms" in wc[wing]:
+            if room in wc[wing]["archived_rooms"]:
+                wc[wing]["archived_rooms"].remove(room)
+                if not wc[wing]["archived_rooms"]:
+                    del wc[wing]["archived_rooms"]
+                self.save_wing_config(wc)
+                return True
+        return False
+
+    def get_archived_rooms(self, wing: str) -> list:
+        """Return list of archived room names for a given wing."""
+        wc = self.load_wing_config()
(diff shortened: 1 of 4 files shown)


web3guru888 (approved):
**Approved** — Clean, well-structured implementation. Zero data deletion, fully reversible archiving. Tests are solid (11 tests covering idempotency, corrupt JSON recovery, config preservation). This directly addresses a real need from Issue #332.

Two architectural observations (not blockers — fine for a follow-up):

1. **`search()` config instantiation**: In `searcher.py`, the non-MCP `search()` path creates a new `MempalaceConfig()` on every call to read archived wings. This means a filesystem read per search invocation. Consider either passing the config from the caller or caching it. Also, `search()` lacks the `include_archived` parameter that `search_memories()` has — might be worth adding for API consistency.

2. **`$ne` scaling**: With many archived wings, the `$and` filter grows linearly (one `$ne` per archived wing). For repositories with 50+ wings this could become unwieldy. A future optimization could store an `archived` metadata field on each drawer entry and filter with a single `{"archived": {"$ne": true}}`, but that is a bigger change and not needed now.

We plan to use this feature in the [MemPalace-AGI](https://github.com/web3guru888/mempalace-stan-extension) integration for archiving completed research cycles, so happy to see it land. Nice work! 👍

web3guru888:
Agreed — 31 rooms across 5 wings is well within comfortable range for the per-wing list approach. We're at similar scale and haven't felt any pressure there either. Good call keeping it simple now and revisiting if usage grows; no need to over-engineer the filter path prematurely.

igorls:

PR #343, opened 2026-04-09
Fix convo miner incremental re-ingest and stale drawer cleanup

xiaomayi-ant:
## What does this PR do?
Fixes `convo_miner` so updated conversation exports are re-ingested correctly.

This change adds `source_mtime` tracking for conversation drawers and replaces existing drawers for a source file when the export changes. It also adds regression tests for source-file growth and shrink scenarios.

## How to test
```bash
.venv/bin/pytest tests/test_convo_miner.py tests/test_normalize.py -q

## Checklist
- [x] No hardcoded paths


web3guru888 (commented):
This is a solid fix for a real problem — stale drawers from updated conversation exports is exactly the kind of silent data corruption that's hard to catch downstream.

The approach is clean:
- **mtime-based change detection** avoids re-ingesting unchanged files (good for performance)
- **delete-then-re-add** for changed files prevents stale chunk accumulation when boundaries shift
- **FakeCollection test harness** is well-structured and tests the right scenarios

We run heavy convo mining in our integration (processing conversation exports into 710+ KG entities) and can confirm this class of bug exists — updated exports leave orphan drawers from old chunk boundaries. The mtime check is a pragmatic solution.

One edge case to consider: some export tools write the file in stages (open, truncate, write), which could cause a race where mtime matches but content is incomplete. A secondary check on file size might add resilience, though it's a low-probability scenario.

PR #353, opened 2026-04-09
feat: add update command for incremental palace sync

phobicdotno:
## Summary

Adds `mempalace update <dir>` — an incremental sync that detects new, changed, and deleted files without requiring a full re-mine.

- **New files** on disk are mined into the palace
- **Changed files** (detected via `source_mtime`) have old drawers removed and are re-mined
- **Deleted files** have their stale drawers cleaned up
- **Unchanged files** are skipped

## Usage

```bash
mempalace update ~/projects/my_app
mempalace update ~/projects/my_app --dry-run    # preview changes
```

Supports `--wing`, `--no-gitignore`, `--include-ignored`, `--agent`, `--dry-run` — same flags as `mine`.

## Changes

- `mempalace/miner.py` — Added `update()` function using existing `scan_project()`, `process_file()`, and `source_mtime` tracking
- `mempalace/cli.py` — Added `update` subparser and `cmd_update()` handler
- `tests/test_miner.py` — Added `test_update_detects_changes()` covering new/changed/deleted file detection

## Test plan

- [x] All tests pass (`pytest tests/ -v`)
- [x] Ruff format and lint clean
- [x] No new dependencies
- [x] Tests run without API keys or network access


web3guru888 (commented):
## Review: `mempalace update` — Incremental Sync

Great addition — incremental sync is something I've been wanting for our integration project. The approach of leveraging existing `scan_project()` and `process_file()` is exactly right (no parallel reimplementation), and the new/changed/deleted classification is clean. A few observations:

### 1. Hardcoded `limit=100000` on `collection.get()`

```python
results = collection.get(
    where={"wing": wing},
    include=["metadatas"],
    limit=100000,
)
```

ChromaDB's `get()` with `limit=None` (or omitted) returns **all** matching results — no pagination needed. The hardcoded 100K silently truncates large palaces: wing gets partially synced, some files appear as "new" every run, and stale drawers never get cleaned up. Since there's no warning when the limit is hit, this would be a hard-to-debug issue.

Suggestion: just drop the `limit` parameter entirely:
```python
results = collection.get(
    where={"wing": wing},
    include=["metadatas"],
)
```

### 2. Silent `except Exception` on wing query

```python
try:
    results = collection.get(
        where={"wing": wing},
        include=["metadatas"],
        limit=100000,
    )
except Exception:
    results = {"ids": [], "metadatas": []}
```

If the wing query fails (corrupt DB, permission error, ChromaDB bug), this silently treats the entire palace as empty. Every file becomes "new", gets re-mined, and you end up with duplicate drawers alongside the originals. That's the opposite of what a user expects from `update`.

At minimum, logging the exception would help. But arguably, if we can't read the existing state, the update should abort rather than silently double everything:
```python
try:
    results = collection.get(
        where={"wing": wing},
        include=["metadatas"],
    )
except Exception as e:
    print(f"  ✗ Failed to read wing '{wing}': {e}")
    print("  Cannot determine current state — aborting update.")
    return
```

### 3. Float equality for mtime comparison

```python
if stored_mtime is not None and float(stored_mtime) == current_mtime:
    unchanged += 1
```

I tested this and ChromaDB does round-trip Python floats bit-exactly (at least on 0.6.x with the SQLite backend), so this works in practice today. But it's worth noting that this is an implementation detail of ChromaDB's storage layer, not a documented guarantee. If they ever switch to a format that loses precision (JSON serialization, for example), this comparison would start seeing every file as "changed".

A small epsilon would make this robust against future storage changes:
```python
if stored_mtime is not None and abs(float(stored_mtime) - current_mtime) < 0.01:
    unchanged += 1
```

10ms is well below any real file modification granularity (most filesystems have 1-second or coarser resolution). Low priority since it works today, but it's cheap insurance.

### 4. Test coverage gaps

The single integration test is solid for the happy path, but a couple of edge cases would be nice:

- **Dry-run mode**: Verify that `--dry-run` reports changes but doesn't modify the collection. A quick `assert col.count() == initial_count` after a dry-run update would do it.
- **Deletion-only path**: A test where only files are deleted (no new/changed) to exercise the cleanup path in isolation.
- **Empty palace**: Calling `update` before any `mine` — currently the non-dry-run path would hit `get_collection()` which creates the collection, then see everything as new. That's probably fine, but worth a test to document the behavior.

### 5. Minor: redundant `float()` cast

```python
palace_files[sf]["source_mtime"] = float(meta["source_mtime"])
```

…and later:

```python
if stored_mtime is not None and float(stored_mtime) == current_mtime:
```

The value is already cast to float when building `palace_files`, so the second `float(stored_mtime)` is a no-op. Minor, but removing it reduces cognitive load.

---

Overall this is a clean, well-scoped feature that reuses existing infrastructure correctly. The main things I'd want addressed before merge are #1 (the silent 100K truncation) and #2 (the silent exception on wing query). The rest are nice-to-haves.

Nice work 👍

PR #356, opened 2026-04-09
fix: harmonize query_entity direction default with MCP wrapper

phobicdotno:
## Summary

`KnowledgeGraph.query_entity()` defaults `direction="outgoing"`, but the MCP wrapper `tool_kg_query()` defaults `direction="both"`. This means MCP users and library users get different behavior when no direction is specified.

This aligns the library default to `"both"` to match the MCP interface.

## Changes

- `mempalace/knowledge_graph.py` — Change `query_entity()` default from `"outgoing"` to `"both"`

One-line change. No new dependencies.

## Test plan

- [x] Ruff format and lint clean
- [x] No behavioral change for MCP users (already defaulted to "both" via the wrapper)
- [x] Library users now get the complete picture by default (both directions)

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -185,7 +185,7 @@ class KnowledgeGraph:
 
     # ── Query operations ──────────────────────────────────────────────────
 
-    def query_entity(self, name: str, as_of: str = None, direction: str = "outgoing"):
+    def query_entity(self, name: str, as_of: str = None, direction: str = "both"):
         """
         Get all relationships for an entity.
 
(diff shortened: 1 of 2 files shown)


web3guru888 (commented):
Good catch on the direction mismatch between library and MCP wrapper.

Our integration calls `query_entity()` with explicit direction parameters (`direction="outgoing"` for discovery relationships, `direction="incoming"` for domain entities), so this default change does not affect us. But the harmonization makes sense — callers who omit direction probably expect both directions by default, since the MCP wrapper already returns both.

One thing to consider: the existing tests in `test_knowledge_graph.py` (lines 51, 57, 61) all pass explicit direction values, so they won't catch regressions for the default-change case. Might be worth adding a test that calls `query_entity("Alice")` without direction and asserts both outgoing and incoming results are returned.

bensig:
closing — this fix landed via #611 (merged). thanks @phobicdotno!

PR #360, opened 2026-04-09
feat: Interactive CLI Menu for Improved Usability (Prototype)

swayam-rajput:
## What does this PR do?
Adds a simple interactive menu when running mempalace without arguments so users don’t have to remember commands. It uses InquirerPy for selection and rich for cleaner visuals, with automatic fallback to this mode when no args are passed. 

_Note: This is just a prototype to demonstrate the idea, refine and optimize further to improve overall user experience._

## Checklist
- [ _/ ] Tests pass (`python -m pytest tests/ -v`)
Test Summary
FAILED tests/test_convo_miner.py::test_convo_mining - PermissionError: [WinError 32] The process cannot access the file because it is being used by another pro...
FAILED tests/test_miner.py::test_project_mining - PermissionError: [WinError 32] The process cannot access the file because it is being used by another pro...
================================ 2 failed, 99 passed, 303 warnings in 37.54s ================================ 

- [ _/ ] No hardcoded paths
- [ _/ ] Linter passes (`ruff check .`)

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -30,6 +30,12 @@ import os
 import sys
 import argparse
 from pathlib import Path
+from rich.console import Console
+from rich.panel import Panel
+from rich.text import Text
+from rich.progress import Progress, SpinnerColumn, TextColumn
+
+console = Console()
 
 from .config import MempalaceConfig
 
@@ -163,7 +169,7 @@ def cmd_repair(args):
     palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
 
     if not os.path.isdir(palace_path):
-        print(f"\n  No palace found at {palace_path}")
+        print(f"[red]No palace found at {palace_path}[/red]")
         return
 
     print(f"\n{'=' * 55}")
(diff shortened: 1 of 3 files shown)


web3guru888 (commented):
Nice UX exploration! The InquirerPy menu is a friendly on-ramp for new users who run bare `mempalace` without knowing the commands.

A few things to think about:

1. **New dependencies** — this adds `InquirerPy` and `rich` as hard dependencies in `pyproject.toml`. That's a non-trivial footprint for a feature that only fires when no args are passed. Consider making them optional (`extras_require`) or guarding imports with try/except and graceful fallback to the existing `parser.print_help()`.

2. **Subcommand detection** — the `any(arg in subcommands for arg in sys.argv[1:])` heuristic could false-positive if someone passes a search query containing a subcommand name, e.g. `mempalace search "how to mine data"`. Parsing with argparse first and checking `args.command` would be safer.

3. The `shlex.split` fallback to `.split()` on ValueError is a nice touch for robustness.

Interesting prototype — would be great to see how maintainers feel about the dependency trade-off.

bensig:
Thanks for the prototype — the interactive CLI concept is interesting but adding InquirerPy as a dependency isn't something we want to take on right now. If you revisit this with a zero-dependency approach we'd be happy to look again.

PR #364, opened 2026-04-09
fix(status): paginate metadata reads so palaces over 10k drawers report  correctly

jccidc:
## What does this PR do?

  Fixes a silent truncation bug where palaces with more than 10,000 drawers
  report `10000` as their total drawer count instead of the real number. Found
  this on a fresh palace with 39,334 drawers — `mempalace status` and the
  `mempalace_status` MCP tool both reported 10,000.

  Both `mempalace status` (CLI) and the `mempalace_status`,
  `mempalace_list_wings`, `mempalace_list_rooms`, and `mempalace_get_taxonomy`
  MCP tools called `col.get(limit=10000, ...)` and treated `len(metadatas)` as
  the total drawer count. Palaces larger than 10k silently truncated their stats.

  Use `col.count()` for the authoritative total and walk metadatas in 10k pages
  so per-room / per-wing breakdowns stay accurate. Adds an `_iter_all_metadatas`
  helper in `mcp_server.py` shared by all four read tools, and applies the same
  fix to `miner.status()`.

  The diary read path keeps its 10k limit since per-agent diary collections are
  small and the call is filtered by wing+room.

  ## How to test

  1. Mine a palace with more than 10k drawers (e.g. mine a directory of Claude
  Code JSONL transcripts, or run multiple `mempalace mine` passes).
  2. Verify the real count via ChromaDB directly:
     ```python
     import chromadb
     c = chromadb.PersistentClient(path="~/.mempalace/palace")
     col = c.get_collection("mempalace_drawers")
     print(col.count())
  3. Run mempalace status — total should match col.count(), not be capped at
  10000.
  4. Call mempalace_status MCP tool — total_drawers and the wings / rooms
  breakdowns should reflect the full collection.

  Checklist

  - No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -86,6 +86,28 @@ def _no_palace():
     }
 
 
+def _iter_all_metadatas(col, where=None, page_size=10000):
+    """Yield every metadata entry in the collection, paginating past the 10k cap.
+
+    ChromaDB's `col.get()` enforces a per-call limit, so a single fetch silently
+    truncates large palaces. This walks the collection in pages so callers see
+    every drawer.
+    """
+    total = col.count()
+    offset = 0
+    while offset < total:
+        kwargs = {"include": ["metadatas"], "limit": page_size, "offset": offset}
+        if where is not None:
+            kwargs["where"] = where
+        page = col.get(**kwargs)
+        metas = page.get("metadatas") or []
+        if not metas:
+            break
+        for m in metas:
+            yield m
+        offset += len(metas)
+
+
 # ==================== READ TOOLS ====================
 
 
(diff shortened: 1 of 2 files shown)


jccidc:
Superseded by a fresh PR off current `main` that also addresses the Octocode review feedback: uses a cleaner `while True` pagination pattern, moves the helper into `palace.py` so `miner.py` and `mcp_server.py` share one implementation, and adds a regression test with 10,500 drawers. New PR link to follow.

jccidc:
Replaced by #427 — clean branch off current `main`, includes the Octocode review feedback: `while True` iterator, shared helper in `palace.py`, and a regression test with 10,500 drawers.

PR #365, opened 2026-04-09
chore(deps): pin chromadb to >=1.0,<2

jccidc:
The previous constraint of chromadb>=0.5.0,<0.7 forces installs onto the 0.5/0.6 line, which uses an older on-disk format. Palaces created with chromadb 1.x (the current major) can't be opened by 0.6 — get_collection fails with KeyError: '_type' when reading collection config json.

Pin to the 1.x line so:
  - editable installs from this fork keep working with palaces built on 1.x
  - new installs land on the same major as the current chromadb release
  - we have a single supported major instead of straddling two on-disk formats

If we need to support 0.6.x palaces later we can add a migration path, but right now mixing the two breaks any palace.

## Checklist
- [ ] No hardcoded paths

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "mempalace"
-version = "3.0.14"
+version = "3.1.0"
 description = "Give your AI a memory — mine projects and conversations into a searchable palace. No API key required."
 readme = "README.md"
 requires-python = ">=3.9"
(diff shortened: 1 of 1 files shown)


jccidc:
Superseded by a cleaner PR from a fresh branch off current `main` — that one has the chromadb pin as a single commit without the stale diffs this branch accumulated. New PR link to follow.

jccidc:
Replaced by #426 — single clean commit off current `main`, single-line dependency change.

PR #379, opened 2026-04-09
Add .mempalaceignore support and expand SKIP_DIRS

pintamanta:
## Summary

- Adds .pixi, .conda, .virtualenvs to the hardcoded SKIP_DIRS set. Projects using Pixi currently get 20k+ stdlib files indexed, drowning out actual project content (~200 files becomes ~53k files).
- Introduces .mempalaceignore — a project-root file using .gitignore syntax, applied on top of .gitignore rules during scan_project(). Lets users exclude paths tracked by git but not wanted in the palace (data dirs, static builds, media assets).

## Motivation

Pixi manages Python environments inside the project (.pixi/envs/default/lib/python3.x/...), like node_modules. Without .pixi in SKIP_DIRS, every mine run indexes thousands of irrelevant stdlib files (ssl.py, typing.py, datetime.py), making the palace noisy and search useless.

The .mempalaceignore solves a related problem: directories tracked by git that should not be in the palace (e.g. data/, site/, binary asset folders). Today the only workaround is patching miner.py directly.

## Changes

- miner.py: Added .pixi, .conda, .virtualenvs to SKIP_DIRS
- miner.py: New load_mempalaceignore_matcher() reusing GitignoreMatcher
- miner.py: scan_project() now loads and applies .mempalaceignore rules to dirs and files

## Test plan

- [ ] Verify mine --dry-run on a project with .pixi/ skips the directory
- [ ] Create .mempalaceignore with data/ and *.png, verify exclusion
- [ ] Verify negation patterns work
- [ ] Verify no regression on projects without .mempalaceignore

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -47,6 +47,9 @@ SKIP_DIRS = {
     ".venv",
     "venv",
     "env",
+    ".pixi",
+    ".conda",
+    ".virtualenvs",
     "dist",
     "build",
     ".next",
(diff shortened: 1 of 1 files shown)


web3guru888 (commented):
We run a multi-domain palace (5 wings, 208 discoveries) and `.mempalaceignore` is something we've wanted. This is a direct match for our use case — we have data directories tracked by git that shouldn't be indexed.

**Strengths:**
- Reusing `GitignoreMatcher` for `.mempalaceignore` is clean — same syntax, same matching behavior, no new parser to maintain.
- Adding `.pixi`, `.conda`, `.virtualenvs` to `SKIP_DIRS` is well-motivated. Pixi's in-project `.pixi/envs/` is exactly like `node_modules` and should be skipped by default.
- The matcher loads once at the project root and applies to both dirs and files in `scan_project()`.

**A few things to consider:**

1. **No tests** — The test plan is all manual checkboxes. This really needs at least one automated test for `.mempalaceignore` behavior (create a tmpdir with a `.mempalaceignore`, add some files, run `scan_project()`, assert exclusions). The `test_miner.py` file has good patterns for this — see `test_scan_project_respects_gitignore` for reference.

2. **Negation pattern interaction** — Your parser handles `!` patterns, but `is_gitignored()` applies matchers in ancestor order (last match wins). Since `.mempalaceignore` is applied *after* `.gitignore` matchers, a `!` in `.mempalaceignore` could un-ignore something `.gitignore` excluded — is that intentional? Might be worth documenting.

3. **Heads-up: overlap with #403** — roip's Windows compat PR introduces `.mpignore` with the same concept (project-level ignore on top of `.gitignore`), but using a different filename. If both land, the project will have two competing ignore files (`.mempalaceignore` vs `.mpignore`). Worth coordinating on which name wins. I'd slightly prefer `.mempalaceignore` for explicitness, but `.mpignore` is shorter to type.

4. **`load_mempalaceignore_matcher` only checks project root** — Unlike `.gitignore` which can exist at any directory level, your implementation only loads from the project root. That's probably fine for v1 but worth noting in a docstring.

Would love to see this merged (with tests) — it unblocks a real workflow for multi-purpose repos.

mvalentsev:
Heads up -- #154 targets the same feature (#102) and already has a review. The implementation might differ, worth comparing before duplicating effort.

bensig:
closing in favor of #154 which covers the same .mempalaceignore support. thanks @pintamanta!

PR #380, opened 2026-04-09
fix: batch metadata loading, content hash verification, and room routing

s3rezhkaa:
Three improvements to handle large-scale indexing correctly:

1. MCP Server: Batch metadata loading (mcp_server.py)
   - Add _get_all_metadatas_batch() to load metadata in chunks of 1000
   - Fixes 'too many SQL variables' error when loading 34K+ drawers
   - Affects: tool_status(), tool_list_wings(), tool_get_taxonomy()

2. Miner: Content hash verification (miner.py)
   - Add compute_content_hash() for SHA-256 hashing of file content
   - Update file_already_mined() to compare hashes and return old drawer IDs
   - Update process_file() to delete stale drawers when content changes
   - Re-indexing now correctly updates modified files instead of skipping them

3. Miner: Improved room routing with explicit path field (miner.py)
   - Add 'path' field support in mempalace.yaml room config
   - detect_room() now checks exact path match first (Priority 1)
   - Handle underscore vs dash mismatch: core_java <-> core-java (Priority 1b)
   - Fixes files being routed to 'general' instead of correct rooms

These changes enable reliable incremental indexing for projects with thousands of files across multiple directories.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -2,7 +2,7 @@
 """
 MemPalace MCP Server — read/write palace access for Claude Code
 ================================================================
-Install: claude mcp add mempalace -- python -m mempalace.mcp_server [--palace /path/to/palace]
+Install: claude mcp add mempalace -- python /path/to/mcp_server.py
 
 Tools (read):
   mempalace_status          — total drawers, wing/room breakdown
@@ -17,8 +17,6 @@ Tools (write):
   mempalace_delete_drawer   — remove a drawer by ID
 """
 
-import argparse
-import os
 import sys
 import json
 import logging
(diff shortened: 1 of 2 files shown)


web3guru888 (commented):
Nice focused PR addressing real pain points at scale. We run MemPalace with 34K+ drawers across 5 wings so these issues are very familiar.

## 1. Batch Metadata Loading — Essential Fix

The `_get_all_metadatas_batch()` approach is solid. The SQLite variable limit (~999) is a known landmine — we hit it at around 20K drawers. Your chunked `offset/limit` loop is the right pattern.

One observation: PR #293 (anthonyonazure) solves this same problem differently with a 5-minute metadata cache (`_get_cached_metadata()`). Both approaches have merit but they'll conflict on merge. Worth coordinating — the cache approach avoids repeated full scans, while your batch approach handles the scan correctly when it does happen. Ideally you'd combine both.

Also: the `tool_list_rooms()` change removes the `limit=10000` but doesn't add batching — if `wing` is None, it fetches all metadata unbatched:
```python
kwargs = {"include": ["metadatas"]}
if wing:
    kwargs["where"] = {"wing": wing}
all_meta = col.get(**kwargs)["metadatas"]
```
For consistency, this should use `_get_all_metadatas_batch()` when `wing` is None.

## 2. Content Hash — Different Philosophy Than Our Approach

Your `compute_content_hash()` approach (SHA-256 of full file content, stored per-drawer) is clean for detecting file modifications. We went a different direction with **tiered duplicate detection**: hard threshold at 0.86 cosine similarity and soft at 0.55, scoped to wing+room. This catches semantic duplicates (paraphrased content) not just byte-identical changes.

Your approach is simpler and better for the "did this exact file change?" use case. Our approach catches more subtle duplicates. They're complementary.

**Bug note**: The `file_already_mined()` return type changed from `bool` to `tuple[bool, list]`, but `process_file()` returns early with `return 0` instead of the original `return 0, None` tuple. The `mine()` function was updated to only expect an int from `process_file()` — but this is a breaking API change. Any downstream code calling `process_file()` expecting a tuple will break silently (the int `0` is falsy, so unpacking `drawers, room = process_file(...)` would fail).

## 3. Room Routing — Good UX Fix

The `path` field in `mempalace.yaml` room config is a nice addition. Explicit path mapping > keyword heuristics for deterministic routing. The underscore-vs-dash normalization (`core_java` ↔ `core-java`) is a common pain point.

**Concern**: The `scan_project()` simplification removes gitignore support entirely — all the `GitignoreMatcher`, `respect_gitignore`, and `include_ignored` infrastructure is gone:
```python
def scan_project(project_dir: str) -> list:
    # ... just walks and filters by extension
```
This is a significant regression for anyone relying on gitignore-aware mining. Files in `node_modules`, `dist`, etc. are still skipped via `SKIP_DIRS`, but custom `.gitignore` patterns (e.g., `*.generated.ts`, `fixtures/`) will no longer be respected. Was this intentional? If so, it should be called out as a breaking change.

The symlink and MAX_FILE_SIZE guards that were in the previous version also got removed — those were security hardening additions from #293. Worth keeping.

## Summary

The batch metadata fix is the most important part — it's a real blocker at scale. The content hash is a good foundation. The room routing improvements are welcome. But the gitignore removal needs discussion, and there are merge conflicts with #293 to resolve.

igorls:

PR #381, opened 2026-04-09
feat: Support for vector search with Qdrant

anush008:
## NOTE

This PR is superseded by 
> Add pluggable vector backends with Qdrant #1679

It implements the new backend specification using Qdrant.

<details>
  <summary><h2> PREVIOUS PR DESCRIPTION </h2></summary>

## Description

This PR adds support for using [Qdrant](https://qdrant.tech) as a vector search provider in MemPalace.

Qdrant is an open-source vector search engine for high performance and massive scale.

### Changes

- Replaced the current inline Chroma usage with a common interface for vector search backends.
- Added concrete implementations using Chroma and Qdrant.
- The Qdrant client dependency is marked optional, keeping Chroma as the default backend (current behaviour remains unchanged).

### Testing

I've unit tested this integration against a local Qdrant instance.

### Setup

You can run Qdrant with 

```bash
docker run -p 6333:6333 qdrant/qdrant
```

Set

```env
MEMPALACE_VECTOR_BACKEND=qdrant
MEMPALACE_QDRANT_URL=http://localhost:6333/
```

The dashboard is accessible at http://localhost:6333/dashboard

## Checklist
- [x] No hardcoded paths

</details>


Anush008:
With regards to the AI recommended actions above,

- The Chroma implementation is kept as is. It doesn't need to be wrapped in another class.
- Import errors are indeed handled gracefully.
- I've not added test code since there isn't a test suite set up for the other components. I've tested the vector search integration locally. Thoroughly.

web3guru888 (commented):
Nice abstraction layer. The `VectorCollection` protocol in `backends/__init__.py` is the right approach — it lets integrators swap backends without touching application code.

One thing I noticed: the Qdrant path returns a proper `QdrantCollection` class that explicitly implements `VectorCollection`, but the ChromaDB path returns native ChromaDB collection objects that satisfy the protocol through duck typing. This works today, but it's fragile — if a future ChromaDB release renames or reorders parameters on `.query()` or `.get()`, the protocol match breaks silently (no `isinstance` check, no type error at construction time).

Might be worth wrapping Chroma in a thin adapter class too (like `ChromaCollection(VectorCollection)`), even if it's just delegation. That way both backends fail the same way if the contract drifts.

Also: The `_scroll_cursor` state on `QdrantCollection` makes `.get()` with pagination stateful at the instance level. If two callers paginate the same collection concurrently (e.g., MCP handler + CLI status), they'll clobber each other's cursor. The ChromaDB path doesn't have this issue because offset/limit are stateless. Consider making scroll state caller-managed or using a separate iterator pattern.

Anush008:
> the ChromaDB path returns native ChromaDB collection objects that satisfy the protocol through duck typing. This works today, but it's fragile — if a future ChromaDB release renames or reorders parameters on .query() or .get(), the protocol match breaks

It's highly unlikely that Chroma would do this without a major release. If they did, even wrapping it in `ChromaCollection(VectorCollection)` won't save us. It'll break too.

Anush008:
This PR is superseded by 
> feat: add Qdrant as alternative storage backend (https://github.com/MemPalace/mempalace/pull/700)

It implements the new vector search backend specification using Qdrant.

igorls:

Anush008:
Hi @igorls, this PR has already been superseded as mentioned in the PR description.

I kept it open for reference and visibility.

igorls:
Thanks for confirming, and for the original Qdrant work — the pluggable-backend spec it fed into landed in #1679. Closing this one out as you suggested; it stays fully readable and linkable in its closed state, so the reference value isn't lost. Appreciate you sticking with the project through the redesign.

PR #384, opened 2026-04-09
fix: Windows compatibility for plugin.json and hook scripts (#378)

BondarenkoCom:
## Problem

On Windows, the MemPalace Claude Code plugin fails at two points:

1. **`plugin.json`** uses `"command": "python3"` — on Windows Python is typically invoked as `py`, not `python3`, causing the MCP server to fail on startup
2. **Hook scripts** are `.sh` bash scripts — Windows has no bash by default, causing `Stop hook error: Python was not found` on every session end

Reported in #378.

## Fix

**`plugin.json`** — platform-aware Python command using environment variable detection:
```json
"command": "${env:OS == 'Windows_NT' ? 'py' : 'python3'}"
```

**`hooks.json`** — bash-first with PowerShell fallback:
```
bash .../mempal-stop-hook.sh 2>/dev/null || powershell -NonInteractive -File .../mempal-stop-hook.ps1
```

**New files** — `.ps1` equivalents of the two hook scripts:
- `.claude-plugin/hooks/mempal-stop-hook.ps1`
- `.claude-plugin/hooks/mempal-precompact-hook.ps1`

## Testing

Tested on Windows 11 with Python 3.12 accessible via `py` command. MCP server starts correctly and hooks run without errors.

## Notes

The `bash ... || powershell ...` fallback pattern means:
- Unix/Mac: bash runs, exits 0, powershell never called ✓
- Windows (no bash): bash fails silently, powershell takes over ✓
- Windows (Git Bash installed): bash runs fine, same as Unix ✓

No changes to core Python logic.

### DIFF
--- a/.claude-plugin/hooks/hooks.json
+++ b/.claude-plugin/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-stop-hook.sh"
+            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-stop-hook.sh 2>/dev/null || powershell -NonInteractive -File ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-stop-hook.ps1"
           }
         ]
       }
@@ -16,7 +16,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-precompact-hook.sh"
+            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-precompact-hook.sh 2>/dev/null || powershell -NonInteractive -File ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-precompact-hook.ps1"
           }
         ]
       }
--- a/.claude-plugin/hooks/mempal-precompact-hook.ps1
+++ b/.claude-plugin/hooks/mempal-precompact-hook.ps1
@@ -1,0 +1,4 @@
+# MemPalace PreCompact Hook (Windows) — thin wrapper calling Python CLI
+# All logic lives in mempalace.hooks_cli for cross-harness extensibility
+$INPUT = $input | Out-String
+$INPUT | py -m mempalace hook run --hook precompact --harness claude-code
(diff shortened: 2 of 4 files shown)


web3guru888 (commented):
Nice work on Windows compat, @BondarenkoCom! A few observations:

**Fallback pattern in hooks.json** — The `bash ... 2>/dev/null || powershell ...` approach is clean. On Windows systems with WSL or Git Bash installed, the bash path might succeed but behave differently than expected (e.g. path separators). Have you tested this on a pure Windows + PowerShell setup without WSL?

**`py` vs `python3` in plugin.json** — The ternary `${env:OS == 'Windows_NT' ? 'py' : 'python3'}` is nice. Worth noting that `py` (the Python Launcher for Windows) isn't always available in minimal installs or Docker-on-Windows scenarios. A comment in the JSON or docs noting that `py` must be on PATH would help Windows users who hit a confusing "command not found" error.

**PowerShell scripts** — The `.ps1` wrappers are thin and correct — piping stdin via `$input | Out-String` and forwarding to the Python CLI is the right call. Keeps logic centralized in `mempalace.hooks_cli`.

Overall this is a solid approach to cross-platform support. We run our integration on Linux containers, but having Windows compat broadens the contributor base — especially for the HexNest handoff workflows we've been collaborating on. 👍

BondarenkoCom:
Good catch on the ternary — that's not valid in the plugin.json spec and I should have verified before pushing. Reverted.

**Updated approach for `plugin.json`**: back to `python3` as the command (unchanged from upstream), with a `_windows_note` field documenting the manual override for Windows users who need `py`. The hooks fallback (`bash ... || powershell ...`) and the `.ps1` scripts remain — those are the actual cross-platform fix.

On `py` availability: fair point. The `.ps1` scripts could also use `python` as a secondary fallback (`py -m mempalace... 2>$null || python -m mempalace...`) for minimal Windows installs. Happy to add that if it's worth the noise.

@web3guru888 On pure Windows + PowerShell without WSL: yes, tested on Windows 11 with only the standard PowerShell — no Git Bash, no WSL. The `bash` call fails silently via `2>/dev/null`, PowerShell takes over. Works cleanly.

mquinnv:
Same class of problem on **macOS**: modern macOS (Sonoma+) ships a system `python3` that blocks `pip install` into the global site-packages — Apple requires a venv. So `"command": "python3"` in `plugin.json` resolves to a Python that will never have mempalace installed, and the MCP server fails to start.

The workaround is installing via `pipx` and then manually patching `.mcp.json` to point at the pipx venv's Python (`~/.local/pipx/venvs/mempalace/bin/python3`), but that breaks on every plugin update.

The proper cross-platform fix for both this and #753 would be to declare the plugin with `"source": "pip"` and `"package": "mempalace"` in the marketplace manifest. Claude Code's plugin system already supports this — it creates and manages a venv automatically, so:

- **macOS**: no system python restriction
- **Windows**: no `py` vs `python3` ambiguity  
- **All platforms**: updates don't clobber the MCP server path

This would replace the need for platform-detection hacks in `plugin.json` entirely.

igorls:

PR #392, opened 2026-04-09
fix: Windows mtime test compatibility

bensig:
## Summary
- Use `os.utime` to force mtime change instead of `time.sleep(0.1)` which is unreliable on Windows (low timestamp resolution)
- Remove unused `time` import

Fixes the Windows CI failure on main from #387's mtime test.

## Test plan
- [x] Linux/macOS already passing
- [ ] Windows CI should now pass


milla-jovovich (dismissed):
Approved! 💜

milla-jovovich (approved):

PR #400, opened 2026-04-09
fix: reconfigure MCP stdin/stdout to UTF-8 on Windows (fixes #363)

### LINKED ISSUE #363: Windows: mempalace_search fails with "TextInputSequence must be str in query" on any non-ASCII query due to stdin encoding
### Description

On Windows, `mempalace_search` crashes with `TypeError: TextInputSequence must be str in query.` on **any query containing non-ASCII characters** (Russian, Chinese, Hebrew, emoji, etc.). The MCP server runs fine, `mempalace_kg_query` and `mempalace_list_wings` work correctly — only `mempalace_search` fails, and only on non-ASCII input.

### Root cause

Python on Windows opens `sys.stdin` using the system ANSI codepage (cp1251/cp1252), **not UTF-8**. When the MCP client sends a JSON request containing Cyrillic text, Python reads the UTF-8 bytes but decodes them as cp1252, producing a mojibake string with embedded surrogate characters. The string type-checks as `str` in Python, but cannot be re-encoded to bytes — so when it eventually reaches the HuggingFace tokenizer inside ChromaDB's default ONNX embedding function, the tokenizer dies deep inside `tokenizer.encode(d)`.

### Evidence (from an instrumented `search_memories`)

Same query `"waitlist bot последние изменения"` (32 Unicode chars), same code path, two Python processes:

```
subprocess (input=bytes, fresh process):
  q_len=32  q_repr='waitlist bot последние изменения'
  -> col.query OK, 5 results

live MCP server (stdin from Claude Code):
  q_len=50  q_repr='waitlist bot Ð¿Ð¾Ñ\udc81Ð»ÐµÐ´Ð½Ð¸Ðµ Ð¸Ð·Ð¼ÐµÐ½ÐµÐ½Ð¸Ñ\udc8f'
  -> TypeError: TextInputSequence must be str in query.
```

The length inflating from 32 → 50 characters is the smoking gun: UTF-8 bytes reinterpreted as 8-bit Latin-1 produce ~1.5× more "characters" than the original string. The `\udc81` surrogate character cannot be re-encoded, which is what the HuggingFace tokenizer chokes on.

### Full traceback

```
File ".../mempalace/searcher.py", line 150, in search_memories
    results = col.query(**kwargs)
File ".../chromadb/api/models/Collection.py", line 251, in query
    query_request = self._validate_and_prepare_query_request(...)
File ".../chromadb/api/models/CollectionCommon.py", line 336, in _validate_and_prepare_query_request
    request_embeddings = self._embed_record_set(...)
File ".../chromadb/api/models/CollectionCommon.py", line 736, in _embed_record_set
    return self._embed(input=record_set[field], is_query=is_query)
File ".../chromadb/api/models/CollectionCommon.py", line 758, in _embed
    return config_ef.embed_query(input=input)
File ".../chromadb/api/types.py", line 960, in __call__
    return ONNXMiniLM_L6_V2()(input)
File ".../chromadb/utils/embedding_functions/onnx_mini_lm_l6_v2.py", line 161, in _forward
    encoded = [self.tokenizer.encode(d) for d in batch]
TypeError: TextInputSequence must be str in query.
```

### Steps to reproduce

1. Install MemPalace on Windows (`pip install mempalace`)
2. Mine any folder containing non-ASCII text
3. Start MCP server: `python -m mempalace.mcp_server`
4. From any MCP client, call `mempalace_search` with a query containing non-ASCII chars (e.g. `"привет"`, `"你好"`, `"שלום"`)
5. Result: `{"error": "Search error: TextInputSequence must be str in query."}`

ASCII-only queries work fine. `mempalace_kg_query`, `mempalace_list_wings`, and other tools that don't pass user text through ChromaDB's embedding function also work fine. The bug only affects `mempalace_search` because it's the only read-side tool that embeds user input.

### Environment

- **OS:** Windows 11
- **Python:** 3.13 (Microsoft Store build)
- **mempalace:** 3.0.0
- **chromadb:** 1.5.7
- **Locale:** Russian (system ANSI = cp1251)

### Fix

This is a classic Windows Python stdin-encoding pitfall. The cleanest fix is runtime `sys.stdin.reconfigure()` in the server's `main()`, plus an environment-variable belt-and-suspenders in the MCP client config.

**Patch for `mempalace/mcp_server.py` (main function):**

```python
def main():
    # On Windows, Python defaults stdin to the system ANSI codepage
    # (cp1251/cp1252), which mangles non-ASCII JSON from the MCP client —
    # UTF-8 bytes become mojibake with surrogate characters, and later break
    # HuggingFace tokenizers with "TextInputSequence must be str in query".
    try:
        sys.stdin.reconfigure(encoding="utf-8", errors="strict")
        sys.stdout.reconfigure(encoding="utf-8", errors="strict")
    except Exception as e:
        logger.error(f"Failed to reconfigure stdio to utf-8: {e}")

    logger.info("MemPalace MCP Server starting...")
    # ... rest of main() unchanged
```

**Recommended MCP client config (documentation-only, for users):**

```json
{
  "mcpServers": {
    "mempalace": {
      "command": "python",
      "args": ["-X", "utf8", "-m", "mempalace.mcp_server"],
      "env": {
        "PYTHONUTF8": "1",
        "PYTHONIOENCODING": "utf-8"
      }
    }
  }
}
```

Any single one of these layers fixes the bug. Applying all of them gives a robust default for any Windows user.

### Verification

After applying the `sys.stdin.reconfigure` patch and restarting the MCP server, the same query `"waitlist bot последние изменения"` now returns 5 relevant results in ~250ms with no error. Debug log confirms the string arrives intact (`q_len=32`, clean Cyrillic repr).

### Severity

This effectively makes MemPalace unusable on Windows for any non-English speaker, which is the bug's main impact — it silently degrades the product to English-only without any error message at install time. Users will hit it on the very first real search, and the error message (`TextInputSequence must be str`) points to HuggingFace internals rather than the real cause, making it very hard to diagnose.

Happy to send a PR with the `main()` patch if that's easier than applying it directly.

mvalentsev:
## What does this PR do?

Fixes #363. On Windows, Python defaults `sys.stdin` to the system ANSI codepage (cp1251, cp1252, etc.), not UTF-8. When an MCP client sends JSON containing non-ASCII characters, the bytes get decoded through the wrong codepage and produce mojibake with surrogate escapes. These strings pass Python type checks but crash the HuggingFace tokenizer inside ChromaDB with `TypeError: TextInputSequence must be str in query`.

The fix adds `sys.stdin.reconfigure(encoding="utf-8")` at the top of `main()`, guarded by `sys.platform == "win32"`. `stdout` and `stderr` get the same treatment so non-ASCII JSON responses serialize cleanly. On non-Windows platforms the code is a no-op.

This effectively makes `mempalace_search` work for every non-English speaker on Windows (Russian, Chinese, Hebrew, etc.), where it currently crashes on the very first query containing non-ASCII text.

## How to test

On Windows with a non-English locale:
```bash
python -m mempalace.mcp_server
# from an MCP client, send: mempalace_search with query "привет" or "你好"
# Before: TypeError crash
# After: normal search results
```

On Linux/macOS: no behavior change (guard skips reconfigure).

```bash
ruff check mempalace/mcp_server.py
python -c "import mempalace.mcp_server; print('OK')"
```

## Checklist

- [x] No hardcoded paths
- [x] Python 3.9 compatible (`reconfigure` available since 3.7)
- [x] No new dependencies

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -7879,16 +7879,18 @@ def _dispatch_stdio_request(request: dict):
 def _run_stdio_loop() -> None:
     _restore_stdout()
 
-    # Force UTF-8 on stdio. MCP JSON-RPC is UTF-8, but Python on Windows
-    # defaults stdin/stdout to the system codepage (e.g. cp1251), which
-    # corrupts non-ASCII payloads and surfaces as generic -32000 errors on
-    # Cyrillic/CJK content. See PEP 540.
-    for stream in (sys.stdin, sys.stdout):
-        if hasattr(stream, "reconfigure"):
-            try:
-                stream.reconfigure(encoding="utf-8", errors="replace")
-            except (AttributeError, OSError):
-                pass
+    # On Windows, Python defaults stdin/stdout to the system ANSI codepage
+    # (e.g. cp1251 or cp1252).  MCP clients send UTF-8 JSON, so non-ASCII
+    # characters arrive as mojibake with surrogate escapes and eventually
+    # break the HuggingFace tokenizer inside ChromaDB's embedding function.
+    # Reconfiguring to UTF-8 here fixes it for every downstream reader.
+    if sys.platform == "win32":
+        try:
+            sys.stdin.reconfigure(encoding="utf-8", errors="strict")
+            sys.stdout.reconfigure(encoding="utf-8", errors="strict")
+            sys.stderr.reconfigure(encoding="utf-8", errors="strict")
+        except Exception as e:
+            logger.warning("Could not reconfigure stdio to UTF-8: %s", e)
 
     logger.info("MemPalace MCP Server starting...")
 
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -7883,37 +7883,25 @@ def _run_stdio_loop() -> None:
     # (e.g. cp1251 / cp1252 / cp950). MCP clients send UTF-8 JSON, so the
     # default decode produces a wrong-but-valid str whose non-ASCII chars
     # later break the HuggingFace tokenizer inside ChromaDB's embedding
-    # function. Reconfiguring to UTF-8 fixes every Python-level reader
-    # opened after main() starts; C-level dependency banners that fire at
-    # import time still go through fd-redirected stderr per #225.
+    # function. The shared helper fixes every Python-level reader opened
+    # after main() starts; C-level dependency banners that fire at import
+    # time still go through fd-redirected stderr per #225.
     #
-    # Per-stream errors policy:
+    # Per-stream policy (MCP-specific):
     #   stdin  -- surrogateescape: a malformed byte from a misbehaving
-    #             client (or a stray BOM) becomes a lone surrogate in the
-    #             decoded str instead of crashing readline. The downstream
-    #             json.loads exception path then surfaces a clean -32700
-    #             instead of the readline raising mid-loop and killing
-    #             the whole server on the first bad byte.
-    #   stdout -- strict: we control what we write here; any failure is
-    #             a real bug we want loud.
-    #   stderr -- strict: same. Logging emits ASCII-only by default;
-    #             tracebacks for non-ASCII paths are a debug concern, not
-    #             a write-side codec concern.
-    if sys.platform == "win32":
-        policies = (
-            ("stdin", "surrogateescape"),
-            ("stdout", "strict"),
-            ("stderr", "strict"),
-        )
-        for name, errors in policies:
-            stream = getattr(sys, name, None)
-            reconfigure = getattr(stream, "reconfigure", None)
-            if reconfigure is None:
-                continue
-            try:
-                reconfigure(encoding="utf-8", errors=errors)
-            except Exception as e:
-                logger.warning("Could not reconfigure %s to UTF-8: %s", name, e)
+    #             client becomes a lone surrogate in the decoded str
+    #             instead of crashing readline; downstream json.loads
+    #             surfaces a clean -32700.
+    #   stdout -- strict: we control writes here; any failure is a real
+    #             bug we want loud.
+    #   stderr -- strict: same.
+    from ._stdio import reconfigure_stdio_utf8_on_windows
+
+    reconfigure_stdio_utf8_on_windows(
+        on_failure=lambda name, exc: logger.warning(
+            "Could not reconfigure %s to UTF-8: %s", name, exc
+        ),
+    )
 
     logger.info("MemPalace MCP Server starting...")
 
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -7880,17 +7880,40 @@ def _run_stdio_loop() -> None:
     _restore_stdout()
 
     # On Windows, Python defaults stdin/stdout to the system ANSI codepage
-    # (e.g. cp1251 or cp1252).  MCP clients send UTF-8 JSON, so non-ASCII
-    # characters arrive as mojibake with surrogate escapes and eventually
-    # break the HuggingFace tokenizer inside ChromaDB's embedding function.
-    # Reconfiguring to UTF-8 here fixes it for every downstream reader.
+    # (e.g. cp1251 / cp1252 / cp950). MCP clients send UTF-8 JSON, so the
+    # default decode produces a wrong-but-valid str whose non-ASCII chars
+    # later break the HuggingFace tokenizer inside ChromaDB's embedding
+    # function. Reconfiguring to UTF-8 fixes every Python-level reader
+    # opened after main() starts; C-level dependency banners that fire at
+    # import time still go through fd-redirected stderr per #225.
+    #
+    # Per-stream errors policy:
+    #   stdin  -- surrogateescape: a malformed byte from a misbehaving
+    #             client (or a stray BOM) becomes a lone surrogate in the
+    #             decoded str instead of crashing readline. The downstream
+    #             json.loads exception path then surfaces a clean -32700
+    #             instead of the readline raising mid-loop and killing
+    #             the whole server on the first bad byte.
+    #   stdout -- strict: we control what we write here; any failure is
+    #             a real bug we want loud.
+    #   stderr -- strict: same. Logging emits ASCII-only by default;
+    #             tracebacks for non-ASCII paths are a debug concern, not
+    #             a write-side codec concern.
     if sys.platform == "win32":
-        try:
-            sys.stdin.reconfigure(encoding="utf-8", errors="strict")
-            sys.stdout.reconfigure(encoding="utf-8", errors="strict")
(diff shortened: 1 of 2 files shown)


bgauryy:
# PR Review: fix: reconfigure MCP stdin/stdout to UTF-8 on Windows (fixes #363)

## Executive Summary
| Aspect | Value |
|--------|-------|
| **PR Goal** | Fix Windows encoding mismatch that crashes ChromaDB's HuggingFace tokenizer via MCP stdin/stdout |
| **Files Changed** | 1 |
| **Risk Level** | LOW — platform-gated, try/except guarded, zero impact on non-Windows |
| **Review Mode** | Quick |
| **Review Effort** | 1 — trivial, well-scoped fix |
| **Recommendation** | APPROVE with minor suggestions |

**Affected Areas**: `mempalace/mcp_server.py` — `main()` entry point

**Business Impact**: Unblocks all Windows users encountering non-ASCII content via MCP (#363)

**Flow Changes**: All `sys.stdin.readline()` and `sys.stdout.write()` calls in the main loop now decode/encode as UTF-8 on Windows instead of the system ANSI codepage

## Ratings
| Aspect | Score |
|--------|-------|
| Correctness | 5/5 |
| Security | 5/5 |
| Performance | 5/5 |
| Maintainability | 4/5 |

## PR Health
- [x] Has clear description with root cause analysis
- [x] References ticket/issue (#363)
- [x] Appropriate size (12 added lines, single concern)
- [ ] Has relevant tests — no test added (hard to unit-test platform-specific stdin reconfigure)

## High Priority Issues
(Must fix before merge)

**None.** The fix is correct and well-scoped.

- `sys` is already imported (line ~13)
- `sys.platform == "win32"` is the canonical Python Windows check
- `reconfigure(encoding="utf-8", errors="strict")` is the right call — strict mode prevents silently corrupting data; any `UnicodeDecodeError` is caught by the main loop's `except Exception as e` handler at [line ~943](https://github.com/milla-jovovich/mempalace/blob/main/mempalace/mcp_server.py#L943) which logs and continues
- The try/except guard handles Python < 3.7 (no `reconfigure()`) and non-standard streams
- Placement before the main loop ensures all downstream I/O benefits

---

## Medium Priority Issues
(Should fix, not blocking)

### [Error Handling] #1: Silent failure swallows reconfigure errors with no diagnostic
**Location:** `mempalace/mcp_server.py:930` (PR branch) | **Confidence:** HIGH

If `reconfigure()` fails on a Windows machine (e.g., frozen executable with custom I/O wrappers), the original bug silently persists with zero diagnostic output. The user gets the same crash as #363 and no clue that the fix was bypassed.

```diff
         except Exception:
-            pass  # older Python or non-reconfigurable stream
+            logger.warning("Could not reconfigure stdio to UTF-8: %s", e)
```

Note: requires renaming the except clause to `except Exception as e:`.

---

### [Bug] #2: `sys.stderr` not reconfigured — logger output with non-ASCII may still corrupt on Windows
**Location:** `mempalace/mcp_server.py:928-929` (PR branch) | **Confidence:** MED

The `logger.error(f"Server error: {e}")` at line ~943 and `logger.exception()` calls throughout the file write to stderr. If the exception message contains non-ASCII content (e.g., user-provided memory text in a traceback), stderr still uses the system codepage and could produce mojibake in logs.

```diff
             sys.stdin.reconfigure(encoding="utf-8", errors="strict")
             sys.stdout.reconfigure(encoding="utf-8", errors="strict")
+            sys.stderr.reconfigure(encoding="utf-8", errors="strict")
```

Lower priority since it only affects log readability, not protocol correctness.

---

## Low Priority Issues
(Nice to have)

None.

---

## Flow Impact Analysis

```
main()
  ├── [NEW] sys.stdin.reconfigure(utf-8)   ← only on win32
  ├── [NEW] sys.stdout.reconfigure(utf-8)  ← only on win32
  │
  ├── sys.stdin.readline()     ← now decodes as UTF-8 (was system codepage)
  │     └── json.loads(line)   ← receives clean UTF-8 strings
  │           └── handle_request(request)
  │                 └── TOOLS[...]["handler"](**tool_args)
  │                       └── ChromaDB → HuggingFace tokenizer  ← no more surrogate escapes
  │
  └── sys.stdout.write(json.dumps(response))  ← now encodes as UTF-8
```

**Blast radius**: Zero on non-Windows. On Windows, every stdin read and stdout write in the main loop is affected — this is the intended behavior.

---

mvalentsev:
Applied both: stderr is now reconfigured alongside stdin/stdout, and the except clause logs the reason instead of silently passing.

---

_Edit 2026-05-03: two follow-ups landed since the above._

`ea816f9` refined the policy from blanket `errors="strict"` to per-stream: stdin=`surrogateescape` (so a malformed byte from a misbehaving MCP client survives as a lone surrogate the downstream `json.loads` exception path can surface as a clean -32700 instead of `UnicodeDecodeError` killing the readline loop on the first bad byte), stdout/stderr stay strict (server-controlled writes, failures stay loud). Adds two mocked Windows tests pinning the per-stream policy.

`dbdf5e6` extracts the per-stream loop to `mempalace/_stdio.py` so the CLI / fact_checker entry points (#1282) share the same helper -- internal-only refactor, no public API change. The MCP-side reconfigure now reads as a one-line `reconfigure_stdio_utf8_on_windows(on_failure=...)` call with the strict defaults, while the CLI side overrides stdout/stderr to `replace`.

craoAbbyy (approved):

web3guru888 (commented):
Solid fix. The Windows codepage issue is real — non-English users on Windows get mojibake that eventually crashes ChromaDB's tokenizer downstream. The guard (`sys.platform == "win32"`) is clean, and the try/except means it gracefully degrades.

Two notes:

1. **`errors="strict"`** — Good call using strict mode rather than the default `"replace"`. With MCP, you want to fail fast on encoding issues rather than silently corrupt JSON data.

2. **stderr too** — Nice that you also reconfigure stderr. If the MCP server logs a non-ASCII palace path or drawer content in an error message, that would have crashed on Windows too.

This is a no-op on Linux/macOS, so zero risk to existing setups. We run on Linux but cross-platform compat is important for the ecosystem.

craoAbbyy (approved):

mvalentsev:
Rebased on develop now that #1060 has landed. Repositioned this PR's scope: with `mempalace/_stdio.py` shipped on develop, this PR wires `mcp_server.py:main()` through that shared helper instead of keeping a parallel inline reconfigure block. The helper's docstring specifies `strict` for stdout/stderr (server-controlled JSON-RPC; encode failures should be loud) and `surrogateescape` for stdin (a malformed byte survives as a lone surrogate so the json.loads error path can surface a clean -32700 instead of `UnicodeDecodeError` killing readline). The current inline block uses `errors="replace"` for stdin + stdout only with no stderr coverage; this PR brings `main()` in line with that helper's policy.

If the lighter `replace` form is preferred, happy to close -- the bug itself is functionally fixed by #1060.

PR #402, opened 2026-04-09
docs: add cdd workflow example pack

fuzzymoomoo:
## What does this PR do?
This example shows one practical way to use MemPalace for software engineering work.

## How to test
No test, documentation only

## Checklist
- [x] No hardcoded paths


web3guru888 (commented):
### Review: CDD workflow example pack

Really glad to see this land as a first-class example. This directly reflects conversations we've been having on #335 about how retrieval profiles map to workflow stages, and seeing it concretized as a copyable example pack is exactly what the community needs.

**What I like most:**

The "Example retrieval by stage" section (Planning / Implementation / Validation) is the single most useful part. It answers the question every new integrator asks: *"I have memories — when do I search which rooms?"* This is the retrieval profile concept we discussed — Orient=breadth, Evaluate=precision, Decide=recency — made concrete and approachable.

The handoff capsule format is excellent. We use something similar for cross-session continuity (208 discovery records across 5 domains, each with a structured summary + retrieval hints). The "Retrieval Hints" section with specific wings, rooms, and search phrases is particularly good — it makes the handoff machine-searchable, not just human-readable.

**Suggestions (minor):**

1. **bgauryy's point about `context-bundles/` vs room names** is worth addressing. New users *will* assume directory names map to rooms 1:1. A one-line note in the README ("Directory names don't need to match room names — MemPalace classifies by content, not path") would save confusion.

2. The `mempalace.yaml` defines `incidents` and `validation` rooms with no sample files. Even a stub `incidents/README.md` saying "Add post-mortem notes here" would make the example pack feel complete and show the intended use.

3. Consider adding a "Refresh cycle" note — when should users re-run `mempalace mine`? After every handoff update? After every sprint? We re-mine after each discovery cycle and it's worth making that cadence explicit for new users.

**One connection worth noting:** PR #434 (auto-KG extraction) would compose beautifully with this example pack. A user could `mempalace mine examples/cdd-workflow/` and then `mempalace extract-kg` to auto-populate the KG with decisions and relationships from these sample files. Might be worth a cross-reference if both land.

Clean docs PR, no code changes, clear value. +1.

igorls:

PR #405, opened 2026-04-09
fix(miner): count files with 0 drawers as skipped in dry-run summary

adv3nt3:
## Summary

Rebased onto `develop` (the prior base, `main`, predates the rename). Reframed since the original crash bgauryy reported in #165 review is gone — but the secondary concern from that same review is still live, and that's what this PR now fixes.

## What changed since the original review

Upstream's `process_file` now returns `(0, "general")` for tiny / unreadable files instead of `(0, None)`. So `room_counts[None]` never happens, and `f"{None:20}"` in the summary formatter never crashes. The original `TypeError` motivation is moot.

What's still broken on `develop`: the dry-run summary's "Files processed" / "Files skipped" counters.

## The remaining bug (bgauryy review item #2)

```python
# mempalace/miner.py — current develop HEAD
if drawers == 0 and not dry_run:
    files_skipped += 1
else:
    total_drawers += drawers
    room_counts[room] += 1
```

In dry-run mode, the `not dry_run` guard prevents `files_skipped` from ever incrementing. Tiny files fall to the `else` branch, where:

1. `total_drawers += 0` is a no-op.
2. `room_counts["general"] += 1` inflates the "general" room tally with files that don't actually file any drawer.

Then the summary prints:

```
Files processed: {len(files) - files_skipped}      # always reports ALL files
Files skipped (already filed): {files_skipped}     # always 0 in dry-run
By room:
    general              N files                   # N inflated by tiny files
```

For a 100-file project with 30 tiny files, dry-run says "Files processed: 100" instead of "Files processed: 70 / Files skipped: 30". Cosmetic but misleading — `mine --dry-run` is a planning tool and the numbers should match what `mine` will actually do.

## Fix

```python
if drawers == 0:
    files_skipped += 1
else:
    total_drawers += drawers
    room_counts[room] += 1
```

Drops the `not dry_run` guard. The non-dry-run path is unchanged — `files_skipped += 1` was already happening there. The dry-run path now matches.

## Test added

`test_mine_dry_run_summary_counts_skipped_tiny_files` (in `tests/test_miner.py`):

- Creates 1 substantive file + 2 below `MIN_CHUNK_SIZE`
- Calls `mine(dry_run=True)`
- Asserts `Files processed: 1` and `Files skipped (already filed): 2` appear in stdout

Verified the test fails on `develop` HEAD without the fix and passes with it.

## Files

- `mempalace/miner.py` — one-line conditional change
- `tests/test_miner.py` — one new test (~35 lines)

## Test plan

- [x] `pytest tests/test_miner.py -v` — 35 passed
- [x] `ruff check mempalace/miner.py tests/test_miner.py` — clean
- [x] New test fails on develop HEAD without the fix; passes with it

## Notes

- The original test from the v1 of this PR (`test_process_file_returns_room`) is dropped — it asserted `(0, None)` for tiny files, which is no longer the contract.
- The existing `test_mine_dry_run_with_tiny_file_no_crash` on develop verifies "no crash" but doesn't check summary numbers, so the new test is complementary, not redundant.

Closes the bgauryy review item #2 from #165.

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -1128,7 +1128,7 @@ def _mine_impl(
                 raise
             files_processed = i
             last_file = filepath.name
-            if drawers == 0 and not dry_run:
+            if drawers == 0:
                 files_skipped += 1
             else:
                 total_drawers += drawers
--- a/tests/test_miner.py
+++ b/tests/test_miner.py
@@ -362,6 +362,40 @@ def test_mine_dry_run_with_tiny_file_no_crash():
         shutil.rmtree(tmpdir, ignore_errors=True)
 
 
+def test_mine_dry_run_summary_counts_skipped_tiny_files(capsys):
+    """Dry-run 'Files processed' count must subtract files that returned 0 drawers.
+
+    Before the fix, ``files_skipped`` was only incremented in non-dry-run mode,
+    so the dry-run summary printed every file as processed even when some were
+    too small to file a single drawer.
+    """
+    tmpdir = tempfile.mkdtemp()
+    try:
+        project_root = Path(tmpdir).resolve()
+
+        # 1 substantive + 2 below MIN_CHUNK_SIZE
+        write_file(project_root / "good.py", "def main():\n    print('hello world')\n" * 20)
+        write_file(project_root / "tiny_a.txt", "x")
+        write_file(project_root / "tiny_b.txt", "y")
+
+        with open(project_root / "mempalace.yaml", "w") as f:
+            yaml.dump(
+                {
+                    "wing": "test_project",
+                    "rooms": [{"name": "general", "description": "General"}],
+                },
+                f,
+            )
+
+        mine(str(project_root), str(project_root / "palace"), dry_run=True)
+
+        out = capsys.readouterr().out
+        assert "Files processed: 1" in out, out
+        assert "Files skipped (already filed): 2" in out, out
+    finally:
+        shutil.rmtree(tmpdir, ignore_errors=True)
+
+
 def test_status_missing_palace_does_not_create_empty_collection(tmp_path, capsys):
     palace_path = tmp_path / "missing-palace"
 


bgauryy:
# PR #405 Review: `fix: skip 0-drawer files from room stats regardless of mode`

**PR**: [milla-jovovich/mempalace#405](https://github.com/milla-jovovich/mempalace/pull/405)
**Author**: adv3nt3 | **Branch**: `fix/miner-room-stats-none-guard` → `main`
**Scope**: 2 files, +53 / -2 lines (3 source lines changed, 50 test lines added)

---

## Executive Summary

| Aspect | Rating |
|--------|--------|
| **Correctness** | 9/10 — Fix is precise and addresses the root cause |
| **Security** | 10/10 — No security surface |
| **Performance** | 10/10 — No impact |
| **Maintainability** | 8/10 — Clean fix; test covers contract but not integration |

**Recommendation: Approve** with one non-blocking suggestion.

---

## Root Cause & Fix Analysis

The bug originates from the interaction between `process_file()` and the aggregation loop in `mine()`.

`process_file()` returns `(0, None)` when a file is too small or unreadable:

```python
content = content.strip()
if len(content) < MIN_CHUNK_SIZE:
    return 0, None
```

The **old code** in `mine()`:

```python
if drawers == 0 and not dry_run:
    files_skipped += 1
else:
    total_drawers += drawers
    room_counts[room] += 1
```

When `dry_run=True` and `drawers=0`: the condition evaluates to `True and False` = `False`, falling to `else`, which executes `room_counts[None] += 1`. The `None` key then crashes the summary formatter (`f"    {room:20}"` — `None` doesn't support format spec `:20`).

The **fix** restructures the conditional:

```python
if drawers == 0:
    if not dry_run:
        files_skipped += 1
else:
    total_drawers += drawers
    room_counts[room] += 1
```

Now `drawers == 0` is caught **regardless of mode**. The `else` branch (which accesses `room`) only executes when `drawers > 0`, which guarantees `room` is a valid string (since `process_file` always sets `room = detect_room(...)` before returning a positive count).

**Verdict: Correct.** The fix is minimal, well-scoped, and preserves all existing behavior for the `drawers > 0` path.

---

## Behavioral Matrix (Before vs After)

| Scenario | `drawers` | `dry_run` | **Before** | **After** |
|----------|-----------|-----------|------------|-----------|
| Normal file, normal mode | >0 | False | Stats counted, printed | Same |
| Normal file, dry-run | >0 | True | Stats counted, no print | Same |
| Tiny file, normal mode | 0 | False | `files_skipped += 1` | Same |
| Tiny file, dry-run | 0 | True | **CRASH** (`room_counts[None]`) | Silently skipped |
| Already mined, normal mode | 0 | False | `files_skipped += 1` | Same |
| Already mined, dry-run | N/A | True | Check skipped, processes normally | Same |

All existing paths are preserved. Only the crashing path is fixed.

---

## Issues

### 1. Test gap: No integration test for the `mine()` crash path [MEDIUM, HIGH confidence]

**Location**: `tests/test_miner.py:262-312`

The added test `test_process_file_returns_room` validates the **contract** of `process_file()` — that tiny files return `(0, None)` and normal files return `(count, room_name)`. This is useful but doesn't directly test the fixed code path.

The actual bug was in the `mine()` function's loop. An integration test that calls `mine(dry_run=True)` with a project containing a tiny file would directly prove the `TypeError` no longer occurs and the summary prints without crashing.

**Suggested addition** (non-blocking):

```python
def test_mine_dry_run_skips_tiny_files():
    """mine() in dry-run mode doesn't crash when files return 0 drawers."""
    tmpdir = tempfile.mkdtemp()
    try:
        project_root = Path(tmpdir)
        write_file(project_root / "big.py", "def main():\n    pass\n" * 20)
        write_file(project_root / "tiny.py", "x = 1")
        # Should not raise TypeError on room_counts[None]
        mine(
            project_dir=str(project_root),
            palace_path=str(Path(tmpdir) / "palace"),
            dry_run=True,
        )
    finally:
        shutil.rmtree(tmpdir, ignore_errors=True)
```

### 2. Dry-run summary overcounts "Files processed" [LOW, MEDIUM confidence]

**Location**: `mempalace/miner.py:597`

```python
print(f"  Files processed: {len(files) - files_skipped}")
```

In dry-run mode, `files_skipped` is never incremented (the inner `if not dry_run` guard prevents it), so the summary reports all files as "processed" even when some returned 0 drawers. This is **pre-existing** (before this fix the code would crash before reaching the summary) and cosmetic-only, so not blocking.

---

## PR Health

| Check | Status |
|-------|--------|
| Fix matches described root cause | Pass |
| No regressions to existing paths | Pass |
| Tests added | Pass (contract-level; integration would strengthen) |
| No security concerns | Pass |
| No performance impact | Pass |
| Change is minimal/focused | Pass |
| PR description quality | Excellent — clear root cause, fix rationale |

---

## Verdict

**Approve.** The fix is correct, minimal, and directly addresses the `TypeError` crash. The single suggestion (adding an integration test for `mine()`) is non-blocking — the existing contract test already proves the return-value invariant the fix relies on.

web3guru888 (commented):
Good catch on the dry-run edge case. The root cause analysis is clear — `(0, None)` flowing into `room_counts` is exactly the kind of bug that only surfaces in specific modes.

The fix is minimal and correct: `if drawers == 0` covers both dry-run and normal mode, and the nested `if not dry_run` for the skip counter preserves existing behavior. The regression test verifying `process_file` return contracts is a nice addition.

igorls:

adv3nt3:
@igorls thanks for the ping. Rebased onto `develop` and reframed.

The original `(0, None)` crash from bgauryy's review is no longer reachable on develop — `process_file` now returns `(0, "general")` for tiny files instead of `(0, None)`. So I dropped that part.

What's left and still broken: the dry-run summary undercounts `files_skipped` and inflates `room_counts["general"]` because the `not dry_run` guard skips the increment. The fix is the same one-line conditional change; just the framing shifted from "fixes a crash" to "fixes a counting bug in dry-run mode" (bgauryy's review item #2).

New test `test_mine_dry_run_summary_counts_skipped_tiny_files` fails on develop HEAD without the fix and passes with it. PR title + description updated to match.

PR #406, opened 2026-04-09
feat(retrieval-foundation): normalize ingest metadata via shared writer

### LINKED ISSUE #269: Hybrid retrieval direction: make the palace wiring production-real without changing the conceptual design
## Why I want to work on this

I’m interested in contributing here because I think the core idea is genuinely strong.

The part worth protecting is not hype, it’s the actual design instinct:
- local-first
- verbatim-first
- memory as navigable structure instead of flat fact soup
- keeping the source material available instead of pretending summaries are the memory

That is a good idea.

What I’d like to help with is making more of that design true in the actual retrieval/runtime wiring — not by changing the palace metaphor, but by implementing the internals the way a real forward-deployed retrieval system would be built.

## Relevant background

Briefly: I spend most of my time building real AI systems, retrieval systems, agent infrastructure, and production architecture.

The skill set I’d be bringing to this is:
- retrieval architecture (dense, lexical, hybrid, rerank)
- GraphRAG / knowledge-graph-assisted retrieval
- vector + graph database design
- agent systems and MCP-style tool interfaces
- forward-deployed / production-first architecture
- scaling, infra, reliability, and cloud/on-prem execution

So my interest here is not “how do we make the README sound better.”
It’s “how do we wire this so the concept actually earns the strongest version of its claims.”

## The idea direction I think fits MemPalace best

I do **not** think the right move is to throw away the current conceptual design and replace it with some generic modern RAG stack.

I think the right move is:

### 1. Keep drawers canonical
The final evidence should still be verbatim drawers / original text.

That means the system stays philosophically aligned with the project:
- source text remains primary
- retrieval quality improves
- exact source access is never lost

### 2. Make the palace structure operationally real in metadata
Right now, a lot of the metaphor is stronger in the docs than in the runtime wiring.

I think the project should consistently track richer metadata at ingest time, e.g.:
- wing
- room
- hall
- source type
- source/session/document grouping
- date / event date
- entities
- importance/confidence
- stable parent-child relationships between closets and drawers

That would make the palace structure genuinely usable as a routing signal rather than mostly a story layered on top.

### 3. Add real hybrid retrieval
The single highest-leverage improvement, in my opinion, is:
- dense retrieval
- lexical/BM25 retrieval
- score fusion
- metadata-aware reranking

This helps with exactly the cases semantic-only retrieval tends to miss:
- names
- quoted phrases
- rare terms
- product/version strings
- exact factual surface forms

### 4. Use graph/KG as routing and expansion, not as a replacement for retrieval
This repo already has the foundations of a useful graph layer.

I think the best fit is:
- use KG and palace graph for query routing
- use them for candidate expansion / narrowing
- use them for currentness / contradiction checks
- still end at verbatim drawer retrieval

That preserves the product’s identity while making the graph earn its keep.

### 5. Add query-aware reranking
A lot of the real-world gains in retrieval come from knowing what kind of question is being asked.

Examples:
- exact phrase heavy
- temporal / recency-sensitive
- entity-centric
- assistant-reference
- preference / personal state

So I’d strongly recommend rerank signals like:
- exact phrase match
- entity overlap
- temporal proximity when the query implies time
- mild recency bias only when appropriate
- session/source grouping support

### 6. Optional final evidence resolution layer
After retrieval is good, an optional evidence-resolution stage could:
- synthesize an answer
- compare KG facts against retrieved evidence
- surface contradictions or stale facts
- attach citations/source pointers

But I would keep that explicitly downstream of retrieval, not as a substitute for it.

## One thing I would *not* lead with
I would not make “binary embeddings” or “full GraphRAG replatforming” the first move.

Those may be useful later as experiments, but they feel like the wrong place to start for this codebase.

The first real gains are much more practical:
- better metadata discipline
- lexical + dense hybrid retrieval
- query-aware rerank
- graph/KG-assisted routing
- preserving exact source evidence

## What I think would be a defensible target outcome
Not “lossless retrieval” literally.

More like:
- source-lossless evidence retrieval
- verbatim-preserving hybrid retrieval
- exact-source-preserving retrieval pipeline

That is both stronger technically and more honest.

## If this direction is interesting
I’m happy to help turn this into something concrete and incremental, e.g.:
1. a retrieval architecture design doc
2. a staged implementation plan
3. small PRs that move the runtime toward this without breaking the project’s conceptual design

I think there’s a very real chance to make MemPalace materially better here while staying faithful to what makes it distinct.

slapglif:
## Summary
- add a shared writer/metadata helper for normalized ingest fields
- route `miner.py`, `convo_miner.py`, and `mcp_server.py` manual writes through the shared path
- add focused tests for the three write surfaces

## Why this PR
This is the Phase 1 slice discussed in #269.

It is intentionally limited to:
- normalized ingest metadata
- shared writer path
- the three write entry points Milla called out:
  - `miner.py`
  - `convo_miner.py`
  - `mcp_server.py`

It does **not** attempt to land:
- the `searcher.py` retrieval seam
- hybrid retrieval
- lexical candidate generation
- query-aware reranking
- graph-first retrieval

## Metadata contract implemented here
This patch adds/stamps the following normalized metadata fields through the shared path.

### Required fields written by all three Phase 1 entry points
| Field | Type | Meaning |
|---|---|---|
| `wing` | string | top-level palace grouping |
| `room` | string | local topic grouping |
| `source_file` | string | original file/path when available, else empty string |
| `chunk_index` | integer | chunk position within the source grouping |
| `added_by` | string | writer identity / ingest path |
| `filed_at` | ISO datetime string | when the drawer was filed |
| `closet_id` | string | stable closet/group identifier for sibling drawers |
| `source_group_id` | string | source grouping key used to derive closet identity |
| `source_type` | string | origin type for the drawer |
| `hall` | string | metadata-only hall classification |
| `memory_type` | string | metadata-only memory classification |
| `importance` | integer | default ranking prior placeholder |
| `confidence` | float | default confidence prior placeholder |
| `content_hash` | string | content digest for dedup / incremental sync work |
| `source_updated_at` | ISO datetime string or empty string | source freshness signal when available |

### Source-type values used in this PR
| Entry point | `source_type` | `hall` | `memory_type` |
|---|---|---|---|
| `miner.py` | `project_file` | `hall_project` | `project_chunk` |
| `convo_miner.py` | `conversation_file` | `hall_conversation` | `conversation_exchange` or extracted memory type |
| `mcp_server.py::tool_add_drawer` | `manual_drawer` | `hall_manual` | `manual_drawer` |

### Optional / source-specific additions in this PR
| Field | Type | Where | Notes |
|---|---|---|---|
| `ingest_mode` | string | `convo_miner.py` | extra metadata, currently `convos` |
| `extract_mode` | string | `convo_miner.py` | extra metadata, e.g. `exchange` / `general` |
| `source_mtime` | float | `miner.py` | preserved from current main behavior for incremental re-mining |

### Backwards compatibility
This contract is additive.

Existing drawers may still only have the older baseline fields. This PR does not require a migration before reads continue to work.

The intent is:
- normalize all new writes through one contract first
- preserve current retrieval behavior
- let later phases (`searcher.py` seam, hybrid retrieval, rerank, KG priors) build on a stable metadata substrate

## Files in scope
- `mempalace/writer.py`
- `mempalace/miner.py`
- `mempalace/convo_miner.py`
- `mempalace/mcp_server.py`
- `tests/test_miner.py`
- `tests/test_convo_miner.py`
- `tests/test_mcp_server.py`

## Validation
Focused local validation on the three write surfaces:
- `tests/test_miner.py::test_project_mining`
- `tests/test_convo_miner.py::test_convo_mining`
- `tests/test_mcp_server.py::TestWriteTools::test_add_drawer`

Result on the rebased branch:
- `3 passed`

Practical note:
- first-time local mining in this environment triggers Chroma's ONNX model download, which makes full shell-based end-to-end runs noisy and slow
- the intent of this PR is to keep the review surface on the write-path contract itself, with later retrieval behavior work following in separate PRs

Closes #269

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -17,6 +17,12 @@ from collections import defaultdict
 
 from .normalize import normalize
 from .palace import SKIP_DIRS, get_collection, file_already_mined
+from .writer import (
+    add_collection_drawer,
+    build_drawer_id,
+    build_shared_metadata,
+    build_source_group_id,
+)
 
 
 # File types that might contain conversations
@@ -328,28 +334,41 @@ def mine_convos(
 
         # File each chunk
         drawers_added = 0
+        source_group_id = build_source_group_id("conversation_file", source_file=source_file)
         for chunk in chunks:
             chunk_room = chunk.get("memory_type", room) if extract_mode == "general" else room
             if extract_mode == "general":
                 room_counts[chunk_room] += 1
-            drawer_id = f"drawer_{wing}_{chunk_room}_{hashlib.sha256((source_file + str(chunk['chunk_index'])).encode()).hexdigest()[:24]}"
+            chunk_memory_type = (
+                chunk.get("memory_type", "conversation_memory")
+                if extract_mode == "general"
+                else "conversation_exchange"
+            )
+            drawer_id = build_drawer_id(
+                wing,
+                chunk_room,
+                source_file=source_file,
+                chunk_index=chunk["chunk_index"],
+            )
+            metadata = build_shared_metadata(
+                wing=wing,
+                room=chunk_room,
+                content=chunk["content"],
+                source_file=source_file,
+                chunk_index=chunk["chunk_index"],
+                added_by=agent,
+                source_type="conversation_file",
+                hall="hall_conversation",
+                memory_type=chunk_memory_type,
+                source_group_id=source_group_id,
+                source_updated_at=source_file,
+                extra_metadata={
+                    "ingest_mode": "convos",
+                    "extract_mode": extract_mode,
+                },
+            )
             try:
-                collection.upsert(
-                    documents=[chunk["content"]],
-                    ids=[drawer_id],
-                    metadatas=[
-                        {
-                            "wing": wing,
-                            "room": chunk_room,
-                            "source_file": source_file,
-                            "chunk_index": chunk["chunk_index"],
-                            "added_by": agent,
-                            "filed_at": datetime.now().isoformat(),
-                            "ingest_mode": "convos",
-                            "extract_mode": extract_mode,
-                        }
-                    ],
-                )
+                add_collection_drawer(collection, drawer_id, chunk["content"], metadata)
                 drawers_added += 1
             except Exception as e:
                 if "already exists" not in str(e).lower():
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -27,7 +27,9 @@ import time
 from datetime import datetime
 from pathlib import Path
 
-from .config import MempalaceConfig, sanitize_name, sanitize_content
+SUPPORTED_PROTOCOL_VERSIONS = ["2025-11-25", "2025-03-26", "2024-11-05"]
+
+from .config import MempalaceConfig
 from .version import __version__
 import chromadb
 from .query_sanitizer import sanitize_query
@@ -1311,17 +1313,18 @@ def handle_request(request):
     req_id = request.get("id")
 
     if method == "initialize":
-        client_version = params.get("protocolVersion", SUPPORTED_PROTOCOL_VERSIONS[-1])
-        negotiated = (
-            client_version
-            if client_version in SUPPORTED_PROTOCOL_VERSIONS
-            else SUPPORTED_PROTOCOL_VERSIONS[0]
-        )
+        client_version = params.get("protocolVersion")
+        if client_version in SUPPORTED_PROTOCOL_VERSIONS:
+            protocol_version = client_version
+        elif client_version:
+            protocol_version = SUPPORTED_PROTOCOL_VERSIONS[0]
+        else:
+            protocol_version = SUPPORTED_PROTOCOL_VERSIONS[-1]
         return {
             "jsonrpc": "2.0",
             "id": req_id,
             "result": {
-                "protocolVersion": negotiated,
+                "protocolVersion": protocol_version,
                 "capabilities": {"tools": {}},
                 "serverInfo": {"name": "mempalace", "version": __version__},
             },
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -33,6 +33,7 @@ import chromadb
 from .query_sanitizer import sanitize_query
 from .searcher import search_memories
 from .palace_graph import traverse, find_tunnels, graph_stats
+from .writer import add_collection_drawer, build_drawer_id, build_shared_metadata
 
 from .knowledge_graph import KnowledgeGraph
 
@@ -458,7 +459,7 @@ def tool_add_drawer(
     if not col:
         return _no_palace()
 
-    drawer_id = f"drawer_{wing}_{room}_{hashlib.sha256((wing + room + content[:100]).encode()).hexdigest()[:24]}"
+    drawer_id = build_drawer_id(wing, room, content=content)
 
     _wal_log(
         "add_drawer",
@@ -480,21 +481,23 @@ def tool_add_drawer(
     except Exception:
         pass
 
+    filed_at = datetime.now().isoformat()
+    metadata = build_shared_metadata(
+        wing=wing,
+        room=room,
+        content=content,
+        source_file=source_file or "",
+        chunk_index=0,
+        added_by=added_by,
+        filed_at=filed_at,
+        source_type="manual_drawer",
+        hall="hall_manual",
+        memory_type="manual_drawer",
+        source_updated_at="",
+    )
+
     try:
-        col.upsert(
-            ids=[drawer_id],
-            documents=[content],
-            metadatas=[
-                {
-                    "wing": wing,
-                    "room": room,
-                    "source_file": source_file or "",
-                    "chunk_index": 0,
-                    "added_by": added_by,
-                    "filed_at": datetime.now().isoformat(),
-                }
-            ],
-        )
+        add_collection_drawer(col, drawer_id, content, metadata)
         _metadata_cache = None
         logger.info(f"Filed drawer: {drawer_id} → {wing}/{room}")
         return {"success": True, "drawer_id": drawer_id, "wing": wing, "room": room}
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -29,7 +29,7 @@ from pathlib import Path
 
 SUPPORTED_PROTOCOL_VERSIONS = ["2025-11-25", "2025-03-26", "2024-11-05"]
 
-from .config import MempalaceConfig
+from .config import MempalaceConfig, sanitize_name, sanitize_content
 from .version import __version__
 import chromadb
 from .query_sanitizer import sanitize_query
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -16,6 +16,7 @@ from datetime import datetime
 from collections import defaultdict
 
 from .palace import SKIP_DIRS, get_collection, file_already_mined
+from .writer import add_collection_drawer, build_drawer_id, build_shared_metadata
 
 READABLE_EXTENSIONS = {
     ".txt",
@@ -372,28 +373,31 @@ def add_drawer(
     collection, wing: str, room: str, content: str, source_file: str, chunk_index: int, agent: str
 ):
     """Add one drawer to the palace."""
-    drawer_id = f"drawer_{wing}_{room}_{hashlib.sha256((source_file + str(chunk_index)).encode()).hexdigest()[:24]}"
+    drawer_id = build_drawer_id(wing, room, source_file=source_file, chunk_index=chunk_index)
+    extra_metadata = {}
     try:
-        metadata = {
-            "wing": wing,
-            "room": room,
-            "source_file": source_file,
-            "chunk_index": chunk_index,
-            "added_by": agent,
-            "filed_at": datetime.now().isoformat(),
-        }
-        # Store file mtime so we can detect modifications later.
-        try:
-            metadata["source_mtime"] = os.path.getmtime(source_file)
-        except OSError:
-            pass
-        collection.upsert(
-            documents=[content],
-            ids=[drawer_id],
-            metadatas=[metadata],
-        )
+        extra_metadata["source_mtime"] = os.path.getmtime(source_file)
+    except OSError:
+        pass
+    metadata = build_shared_metadata(
+        wing=wing,
+        room=room,
+        content=content,
+        source_file=source_file,
+        chunk_index=chunk_index,
+        added_by=agent,
+        source_type="project_file",
+        hall="hall_project",
+        memory_type="project_chunk",
+        source_updated_at=source_file,
+        extra_metadata=extra_metadata or None,
+    )
+    try:
+        add_collection_drawer(collection, drawer_id, content, metadata)
         return True
-    except Exception:
+    except Exception as e:
+        if "already exists" in str(e).lower() or "duplicate" in str(e).lower():
+            return False
         raise
 
 
--- a/mempalace/writer.py
+++ b/mempalace/writer.py
@@ -1,0 +1,142 @@
+"""Shared drawer metadata and write helpers."""
+
+from __future__ import annotations
+
+import hashlib
+import re
+from datetime import datetime
+from pathlib import Path
+from typing import Any, Dict, Optional
+
+
+_DEFAULT_IMPORTANCE = 3
+_DEFAULT_CONFIDENCE = 1.0
+
+
+def _md5_hexdigest(value: str) -> str:
+    return hashlib.md5(value.encode("utf-8"), usedforsecurity=False).hexdigest()
+
+
+def hash_content(content: str) -> str:
+    return _md5_hexdigest(content)
+
+
+def _slug(value: str) -> str:
+    slug = re.sub(r"[^A-Za-z0-9_-]+", "_", str(value or "").strip())
+    return slug.strip("_") or "unknown"
+
+
+def resolve_source_updated_at(source: Any = None) -> str:
+    if source in (None, ""):
+        return ""
+    if isinstance(source, datetime):
+        return source.isoformat()
+
+    try:
+        source_path = Path(source)
+    except TypeError:
+        return str(source)
+
+    if source_path.exists():
+        return datetime.fromtimestamp(source_path.stat().st_mtime).isoformat()
+    return str(source)
+
+
+def build_source_group_id(
+    source_type: str,
+    *,
+    source_file: str = "",
+    wing: str = "",
+    room: str = "",
+) -> str:
+    if source_type == "manual_drawer":
+        return f"manual_drawer:{wing}:{room}"
+    if source_file:
+        return f"{source_type}_{_md5_hexdigest(source_file)[:16]}"
+    return f"{source_type}_{_md5_hexdigest(f'{wing}:{room}')[:16]}"
+
+
+def build_closet_id(wing: str, room: str, source_group_id: str) -> str:
+    return f"closet_{_slug(wing)}_{_slug(room)}_{_md5_hexdigest(source_group_id)[:12]}"
+
+
+def build_drawer_id(
+    wing: str,
+    room: str,
+    *,
+    source_file: Optional[str] = None,
+    chunk_index: int = 0,
+    content: Optional[str] = None,
+    filed_at: Optional[str] = None,
+) -> str:
+    if source_file:
+        seed = f"{source_file}{chunk_index}"
+    else:
+        seed = f"{(content or '')[:100]}{filed_at or ''}"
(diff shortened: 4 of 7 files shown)


bgauryy:
# PR Review: feat(retrieval-foundation): normalize ingest metadata via shared writer

## Executive Summary
| Aspect | Value |
|--------|-------|
| **PR Goal** | Introduce a shared `writer.py` module that normalizes metadata construction and drawer writes across all three ingest entry points |
| **Files Changed** | 13 |
| **Risk Level** | MEDIUM — Changes the core write path for all data ingest, expands metadata schema from 6 to 15+ fields |
| **Review Mode** | Full |
| **Review Effort** | 3 — Well-scoped refactor with clear intent |
| **Recommendation** | REQUEST_CHANGES |

**Affected Areas**: `mempalace/writer.py` (new), `mempalace/miner.py`, `mempalace/convo_miner.py`, `mempalace/mcp_server.py`, `mempalace/cli.py`, `mempalace/hooks/`

**Business Impact**: Normalizes all drawer metadata, enabling future retrieval improvements (hall-based graph traversal, content dedup via `content_hash`, importance/confidence scoring). Cosine space fix (#218) corrects similarity calculations.

**Flow Changes**: All three ingest surfaces (`miner.py`, `convo_miner.py`, `mcp_server.py:tool_add_drawer`) now delegate ID generation, metadata construction, and collection writes to `writer.py`. `palace_graph.py:build_graph` will start receiving `hall` metadata it previously lacked, enriching graph traversal.

## Ratings
| Aspect | Score |
|--------|-------|
| Correctness | 4/5 |
| Security | 4/5 |
| Performance | 5/5 |
| Maintainability | 4/5 |

## PR Health
- [x] Has clear description
- [x] References ticket/issue (#269, #218)
- [x] Appropriate size (or justified if large)
- [x] Has relevant tests

## High Priority Issues
(Must fix before merge)

### [Bug] #1: `extra_metadata` can silently overwrite core metadata fields
**Location:** `mempalace/writer.py:125` | **Confidence:** HIGH

`build_shared_metadata` applies `extra_metadata` via `metadata.update()` after constructing the base dict. Any key in `extra_metadata` that collides with a core field (`wing`, `room`, `filed_at`, `content_hash`, etc.) will silently overwrite it. Currently only `convo_miner.py` passes `extra_metadata` with safe keys (`ingest_mode`, `extract_mode`), but this is a ticking time bomb for future callers.

```diff
     if extra_metadata:
-        metadata.update({key: value for key, value in extra_metadata.items() if value is not None})
+        _PROTECTED_KEYS = frozenset(metadata.keys())
+        for key, value in extra_metadata.items():
+            if value is not None:
+                if key in _PROTECTED_KEYS:
+                    raise ValueError(f"extra_metadata key '{key}' conflicts with core metadata field")
+                metadata[key] = value
```

---

### [Architecture] #2: `tool_diary_write` bypasses the shared writer
**Location:** `mempalace/mcp_server.py:507-520` | **Confidence:** HIGH

The PR routes `tool_add_drawer` through the shared writer but leaves `tool_diary_write` using a raw `col.add()` with its own handcrafted metadata dict. This creates schema inconsistency within the same module — diary entries lack `source_type`, `memory_type`, `importance`, `confidence`, `closet_id`, `source_group_id`, `content_hash`, and `source_updated_at`.

The PR description says it covers "the three write entry points Milla called out" — but `tool_diary_write` is also a write entry point in `mcp_server.py`. At minimum, add a `# TODO: route through shared writer in Phase 2` comment, or route it now to avoid a second round of the same refactor.

```diff
+    # Route diary writes through the shared writer for metadata consistency
+    drawer_id = entry_id  # diary keeps its own ID scheme
+    metadata = build_shared_metadata(
+        wing=wing,
+        room=room,
+        content=entry,
+        added_by=agent_name,
+        filed_at=now.isoformat(),
+        source_type="diary_entry",
+        hall="hall_diary",
+        memory_type="diary_entry",
+        source_updated_at=now,
+        extra_metadata={"topic": topic, "agent": agent_name, "date": now.strftime("%Y-%m-%d")},
+    )
+    add_collection_drawer(col, drawer_id, entry, metadata)
-    col.add(
-        ids=[entry_id],
-        documents=[entry],
-        metadatas=[
-            {
-                "wing": wing,
-                "room": room,
-                "hall": "hall_diary",
-                "topic": topic,
-                "type": "diary_entry",
-                "agent": agent_name,
-                "filed_at": now.isoformat(),
-                "date": now.strftime("%Y-%m-%d"),
-            }
-        ],
-    )
```

---

## Medium Priority Issues
(Should fix, not blocking)

### [Bug] #3: Cosine space not applied to `mempalace_compressed` collection
**Location:** `mempalace/cli.py:cmd_compress` (line ~335 on main) | **Confidence:** MED

`cmd_compress` creates the compressed collection via `client.get_or_create_collection("mempalace_compressed")` without `metadata={"hnsw:space": "cosine"}`. If compressed entries are ever queried with the same `1 - dist` similarity formula used in `searcher.py`, the results will be incorrect (L2 distances instead of cosine).

```diff
-        comp_col = client.get_or_create_collection("mempalace_compressed")
+        comp_col = client.get_or_create_collection(
+            "mempalace_compressed", metadata={"hnsw:space": "cosine"}
+        )
```

---

### [Architecture] #4: No migration path documented for existing palaces
**Location:** Project-level | **Confidence:** HIGH

The cosine space fix only applies when collections are **created** (`create_collection`, `get_or_create_collection`). Existing palaces opened via `get_collection` retain L2 distance. The only migration path is `mempalace repair` (which now correctly recreates with cosine), but this isn't documented in the PR description, README, or changelog.

Users with existing palaces will continue getting wrong similarity scores until they manually run repair. Consider:
- Adding a note in the PR description about the migration path
- Logging a warning when an existing collection lacks cosine space metadata
- Mentioning `mempalace repair` in the hooks README or a migration note

---

### [Bug] #5: `build_drawer_id` edge case with empty `source_file`
**Location:** `mempalace/writer.py:75-80` | **Confidence:** MED

`build_drawer_id` branches on `source_file is not None`. If `source_file=""` (empty string, not None), it enters the file-based branch and seeds the hash with `f""{chunk_index}"` → just the stringified chunk_index. Two different callers with empty source_file and the same chunk_index would generate identical drawer IDs, causing silent dedup/collision.

```diff
 def build_drawer_id(
     wing: str,
     room: str,
     *,
     source_file: Optional[str] = None,
     chunk_index: int = 0,
     content: Optional[str] = None,
     filed_at: Optional[str] = None,
 ) -> str:
-    if source_file is not None:
+    if source_file:
         seed = f"{source_file}{chunk_index}"
     else:
         seed = f"{(content or '')[:100]}{filed_at or ''}"
     return f"drawer_{wing}_{room}_{_md5_hexdigest(seed)[:16]}"
```

---

## Low Priority Issues
(Nice to have)

### [Architecture] #6: Scope creep — hooks CLI packaging bundled with metadata normalization
**Location:** `mempalace/cli.py`, `mempalace/hooks/__init__.py`, `tests/test_hooks.py` | **Confidence:** HIGH

~95 lines of hooks CLI commands + ~75 lines of tests are included in a PR described as "intentionally limited to normalized ingest metadata." The hooks feature is orthogonal and would benefit from a separate PR for clean review and independent revertability.

---

## Flow Impact Analysis

```
BEFORE:
  miner.py ──────────┐
  convo_miner.py ─────┤──► collection.add() [each with own metadata dict]
  mcp_server.py ──────┘

AFTER:
  miner.py ──────────┐
  convo_miner.py ─────┤──► writer.py ──► collection.add() [normalized metadata]
  mcp_server.py ──────┘       │
    (tool_add_drawer)         ├── build_drawer_id()
                              ├── build_shared_metadata()
                              └── add_collection_drawer()

  mcp_server.py ──────────────────────► collection.add() [raw, NOT routed]
    (tool_diary_write)                   ⚠️ BYPASSES shared writer
```

**Downstream consumers verified safe:**
- `searcher.py` — Uses `.get("wing", "?")`, `.get("room", "?")`, `.get("source_file", "?")` with defaults. New fields are purely additive. No breakage.
- `palace_graph.py` — Reads `hall` via `.get("hall", "")`. Previously empty for all drawers; now populated for new drawers. Graph becomes richer over time. No breakage.
- `palace_graph.py` reads `date` via `.get("date", "")` — this field was never in the miner/convo_miner metadata, only in diary entries. The new `filed_at` field doesn't map to `date`. Pre-existing gap, not introduced by this PR.

**Drawer ID backward compatibility:** Verified. The old seed `source_file + str(chunk_index)` equals the new `f"{source_file}{chunk_index}"` for all cases where `source_file` is a non-empty string. IDs are stable across the refactor.

---

milla-jovovich:
Hey @slapglif — I've taken a look and ran this through CLI and the scope is exactly right. Thank you for taking the time to carve Phase 1 down to the three write surfaces instead of the broader rewrite framing — that's the incrementalism we needed.

Two things before a full review:

1. The PR is currently conflicting against main after the v3.1.0 bump — a quick rebase will get it reviewable. While you're in there, @bgauryy's review flagged the metadata-schema expansion (6 → 15+ fields) as the one place the contract needs more explicit docs. Even a short section in the PR body that lists the new fields, their types, and which are required vs. optional for backwards compatibility would make this a much easier land.

2. I want to loop in one of our folks who's been looking at the same metadata contract question from a slightly different angle before we approve the merge. Not a delay — just a "make sure we don't land two overlapping contracts in the same week" check. Should be resolved within a couple of days.

The order you proposed in #269 (metadata contract → shared writer → retrieval seam → lexical+dense → rerank → KG priors → verbatim drawers) is the right order. Let's make this the first step.

Thank you from Milla and Lu✨

slapglif:
@milla-jovovich thanks — I addressed the valid parts directly and updated the draft branch.

What I changed just now:
- rebased the branch onto current `main` so the PR is reviewable again
- kept the PR scoped to the Phase 1 write-path contract only
- fixed a small semantic mismatch in `source_updated_at` for manual drawers so empty/manual writes stay empty rather than picking up a synthetic timestamp

I also agree with @bgauryy's most important concern:
- the metadata contract needs to be more explicit in the PR body

That concern is valid.
What I do not want to do in this PR is broaden scope into unrelated follow-up work.
So my plan for this draft is:
- keep the code surface Phase-1-only
- make the contract/documentation explicit in the PR body
- keep diary/search/hybrid/retrieval-seam work out of this patch

On the other review points:
- routing diary writes through the shared writer is a reasonable future cleanup, but I think it is out of scope for the Phase 1 patch you asked for, since you explicitly scoped the first step to `miner.py`, `convo_miner.py`, and `mcp_server.py` write-path normalization around the shared contract
- the old hooks/CLI spillover was valid as criticism of the earlier draft state; that is exactly why I trimmed/rebased the branch down to the Phase 1 contract slice

I also took your note seriously about not landing overlapping metadata contracts in the same week.

I’m not sure which teammate you meant to loop in from your comment, so I don’t want to guess and tag the wrong person. If you want me to pull them in explicitly, I’m happy to — just point me at the handle and I’ll coordinate directly.

I have a lot of availability right now, so if you or the other reviewer want this split, tightened, or amended further, I can turn changes around quickly.

If this Phase 1 cut now looks acceptable, I’m also ready to move immediately into the next step in the sequence once you give the directive.

web3guru888 (commented):
Early directional feedback on this draft — this is exactly the kind of foundational work MemPalace needs. We normalize ingest metadata in our own pipeline (208 discoveries, 710 KG entities across 5 domains) so we have strong opinions here.

## The Metadata Contract — Very Well Thought Out

The 15-field shared contract is comprehensive. A few observations:

**Fields we independently found essential:**
- `content_hash` — we use this for tiered dedup (hard=0.86, soft=0.55 similarity thresholds). Having it in the base contract is great.
- `source_type` / `hall` / `memory_type` — we use similar discriminators for retrieval profiling. Being able to filter by source type before similarity search is a huge performance win at scale.
- `closet_id` / `source_group_id` — this is the right abstraction for grouping related drawers (e.g., all chunks from one file).

**Fields worth reconsidering:**
- `importance: int = 3` and `confidence: float = 1.0` as defaults — these are placeholder values that will be written to every drawer. Downstream consumers can't distinguish "importance=3 because it was scored" from "importance=3 because it's the default." Consider using `None`/omitting the field when no scoring has been done, so searchers can tell the difference. (ChromaDB metadata doesn't support null, but you could use `-1` as sentinel or simply not include the key.)
- `source_updated_at` — the `resolve_source_updated_at()` function in `writer.py` has a subtle issue: when `source` is a file path string, it calls `Path(source).exists()` and `stat().st_mtime`. But in `convo_miner.py`, you're passing `source_updated_at=source_file` which is the source file path. This means every drawer creation triggers a filesystem stat call. At scale (thousands of chunks from hundreds of files), this adds up. Consider resolving the mtime once per file in the caller and passing the ISO string directly.

## The writer.py Module — Clean Abstraction

`build_shared_metadata()`, `build_drawer_id()`, `add_collection_drawer()` — this is exactly the right factoring. One write path, one metadata shape, three entry points.

**Minor**: `hash_content()` uses MD5 (`_md5_hexdigest`). PR #293 and #380 both switched to SHA-256 for drawer IDs and content hashes. MD5 is fine for non-security hashing (dedup), but for consistency with the emerging convention in the codebase, SHA-256 might be worth matching.

## Regression Concerns

1. **WAL logging removed from mcp_server.py**: The draft strips out the write-ahead log (`_wal_log()`) and all security hardening (`sanitize_name()`, `sanitize_content()`) from the MCP server. These were added in #293 for good reason. The normalized writer should probably call through the sanitizers, not bypass them.

2. **`--palace` CLI arg removed**: The draft removes `argparse` and `_parse_args()` from mcp_server.py, hardcoding `_kg = KnowledgeGraph()` at module level. This breaks the `--palace /custom/path` workflow.

3. **Protocol version negotiation removed**: Replacing the version list with a hardcoded `"2024-11-05"` will break clients using newer MCP protocol versions.

4. **Symlink/MAX_FILE_SIZE guards removed** from `scan_project()` and `scan_convos()`: These are security hardening that should be preserved.

## Direction

The core idea — a shared writer module normalizing all ingest metadata — is the right architecture. The metadata contract table in the PR description is excellent documentation. For the final version, I'd suggest:
- Keep the writer abstraction
- Preserve existing security/audit layers (WAL, sanitizers, symlink guards)
- Resolve the mtime-per-chunk performance issue
- Coordinate with #293 and #380 which touch the same files

Looking forward to seeing this evolve.

milla-jovovich (dismissed):
Looks good

slapglif:
Addressed the high-confidence, in-scope review concerns and pushed them to feat/hybrid-phase1-metadata in commit 1882fca.

What I changed
- restored MCP protocol negotiation compatibility with a supported-version list and initialize fallback behavior
- fixed tools/call handling for arguments: null
- prevented extra_metadata from silently overwriting core metadata fields in writer.py
- fixed the empty source_file edge case in build_drawer_id
- fixed convo_miner.py so the normalized shared metadata path is actually used for writes
- realigned tests to the intended Phase 1 scope and existing duplicate-write behavior

What I intentionally did not change
- routing tool_diary_write through the shared writer
- WAL/audit behavior
- sanitizer plumbing
- broader MCP/server architecture changes
- retrieval or diary follow-on work

Reason: I kept this limited to issues that looked both valid and clearly in-bounds for the Phase 1 metadata-contract PR.

Validation run
- tests/test_miner.py::test_project_mining
- tests/test_miner.py::test_file_already_mined_checks_source_file_only
- tests/test_convo_miner.py::test_convo_mining
- tests/test_mcp_server.py::TestHandleRequest::test_initialize
- tests/test_mcp_server.py::TestHandleRequest::test_initialize_negotiates_client_version
- tests/test_mcp_server.py::TestHandleRequest::test_initialize_negotiates_older_supported_version
- tests/test_mcp_server.py::TestHandleRequest::test_initialize_unknown_version_falls_back_to_latest
- tests/test_mcp_server.py::TestHandleRequest::test_initialize_missing_version_uses_oldest
- tests/test_mcp_server.py::TestHandleRequest::test_null_arguments_does_not_hang
- tests/test_mcp_server.py::TestWriteTools::test_add_drawer
- tests/test_mcp_server.py::TestWriteTools::test_add_drawer_duplicate_detection

Result: 11 passed

If helpful, I can also reply inline to the specific review threads, but I wanted one clean summary on the PR first.

slapglif:
Validated the current Phase 1 metadata branch locally against the focused write-surface suite after refreshing to the latest fork head.

Validation run:
- python3 -m venv .venv-e2e
- .venv-e2e/bin/pip install -e '.[dev]'
- .venv-e2e/bin/pytest tests/test_miner.py tests/test_convo_miner.py tests/test_mcp_server.py -q

Result:
- 47 passed

Notes:
- This still appears as CONFLICTING / DIRTY in GitHub's mergeability metadata right now.
- A local merge-tree check against current upstream/develop did not surface textual conflict markers, so this may just need GitHub to recompute after a fresh push/rebase path.

PR #416, opened 2026-04-09
feat: parallel file processing and batched upserts in mine()

tbaldur:
## Problem

`mine()` processes files sequentially and calls `collection.upsert()` once per chunk. On large codebases this is slow — file I/O stalls the CPU, and the ONNX embedder runs one vector at a time instead of batching.

## Solution

Two changes that compound:

**1. Thread pool for file reading + chunking**

Extracted the ChromaDB-free part of `process_file()` into a new `process_file_cpu()` function (read → chunk → detect room → build metadata records). This has no shared state and is safe to run concurrently. `mine()` now dispatches these via `ThreadPoolExecutor`, overlapping file I/O across files while the main thread handles upserts.

**2. Batched upserts**

Instead of `collection.upsert([1 chunk])` N times, records from completed futures are accumulated and flushed every `BATCH_SIZE` (128) chunks. ChromaDB's ONNX backend processes a batch as a single matrix operation, which is significantly faster than N sequential single-vector calls.

## Constants added

```python
BATCH_SIZE = 128          # chunks per upsert call
MAX_WORKERS = min(32, (os.cpu_count() or 4) * 2)
```

## Compatibility

- `dry_run` path is unchanged (still uses original `process_file()`).
- Already-mined files are pre-filtered before the thread pool to avoid redundant ChromaDB reads from worker threads.
- No new dependencies.

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -17,7 +17,7 @@ from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from .normalize import normalize
-from .palace import SKIP_DIRS, get_collection, file_already_mined
+from .palace import SKIP_DIRS, get_collection
 
 
 # File types that might contain conversations
@@ -369,7 +369,9 @@ def mine_convos(
     # ------------------------------------------------------------------
     else:
         print(f"  Checking {len(files)} files for changes...")
-        pending = [fp for fp in files if not file_already_mined(collection, str(fp))]
+        from .miner import build_mined_cache
+        mined_cache = build_mined_cache(collection)
+        pending = [fp for fp in files if str(fp) not in mined_cache]
         already_mined = len(files) - len(pending)
 
         batch_ids, batch_docs, batch_metas = [], [], []
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -14,6 +14,7 @@ import hashlib
 from pathlib import Path
 from datetime import datetime
 from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from .normalize import normalize
 from .palace import SKIP_DIRS, get_collection, file_already_mined
@@ -28,6 +29,8 @@ CONVO_EXTENSIONS = {
 }
 
 MIN_CHUNK_SIZE = 30
+BATCH_SIZE = 128  # chunks per upsert call (matches miner.py)
+MAX_WORKERS = min(32, (os.cpu_count() or 4) * 2)
 MAX_FILE_SIZE = 10 * 1024 * 1024  # 10 MB — skip files larger than this
 
 
@@ -229,6 +232,76 @@ def scan_convos(convo_dir: str) -> list:
 # =============================================================================
 
 
+def process_convo_file_cpu(
+    filepath: Path,
+    wing: str,
+    agent: str,
+    extract_mode: str,
+) -> "tuple | None":
+    """
+    Pure CPU worker: normalize, chunk, detect room, build drawer records.
+    Thread-safe — no ChromaDB calls, no shared state.
+
+    Returns (source_file, room, records, room_counts_delta) or None if skipped.
+    """
+    source_file = str(filepath)
+
+    try:
+        content = normalize(source_file)
+    except (OSError, ValueError):
+        return None
+
+    if not content or len(content.strip()) < MIN_CHUNK_SIZE:
+        return None
+
+    if extract_mode == "general":
+        from .general_extractor import extract_memories
+
+        chunks = extract_memories(content)
+    else:
+        chunks = chunk_exchanges(content)
+
+    if not chunks:
+        return None
+
+    if extract_mode != "general":
+        room = detect_convo_room(content)
+    else:
+        room = None
+
+    now = datetime.now().isoformat()
+    records = []
+    room_counts_delta = defaultdict(int)
+
+    if extract_mode != "general":
+        room_counts_delta[room] = 1
+
+    for chunk in chunks:
+        chunk_room = chunk.get("memory_type", room) if extract_mode == "general" else room
+        if extract_mode == "general":
+            room_counts_delta[chunk_room] += 1
+
+        drawer_id = (
+            f"drawer_{wing}_{chunk_room}_"
+            + hashlib.sha256(
+                (source_file + str(chunk["chunk_index"])).encode()
+            ).hexdigest()[:24]
+        )
+        meta = {
+            "wing": wing,
+            "room": chunk_room,
+            "source_file": source_file,
+            "chunk_index": chunk["chunk_index"],
+            "added_by": agent,
+            "filed_at": now,
+            "ingest_mode": "convos",
+            "extract_mode": extract_mode,
+        }
+        records.append((drawer_id, chunk["content"], meta))
+
+    return source_file, room, records, dict(room_counts_delta)
+
+
 def mine_convos(
     convo_dir: str,
     palace_path: str,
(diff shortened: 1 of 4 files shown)


bgauryy:
# PR #416 Review: `perf: parallel file processing and batched upserts in mine()`

**URL:** https://github.com/milla-jovovich/mempalace/pull/416
**Author:** tbaldur | **Status:** Open | **Branch:** `feat/parallel-mine` → `main`
**Scope:** 1 file (`mempalace/miner.py`) | +106 / -17

---

## Executive Summary

**Goal:** Speed up `mine()` by parallelizing file I/O + chunking and batching ChromaDB upserts.
**Risk:** HIGH — concurrency in the core data ingestion pipeline.
**Recommendation:** REQUEST CHANGES — Contains copy-paste duplicates, a data-loss path on exceptions, and logic duplication that creates maintenance risk.

## Ratings

| Dimension | Score | Notes |
|---|---|---|
| Correctness | 4/10 | Duplicate imports/constants, data loss on exception |
| Security | 8/10 | No new attack surface; thread-safe reads |
| Performance | 7/10 | Good batching strategy; thread pool sizing is reasonable |
| Maintainability | 4/10 | `process_file` / `process_file_cpu` duplication |

---

## Issues

### 1. CRITICAL — Duplicate import line (copy-paste error)

**Confidence:** HIGH
**Location:** `mempalace/miner.py`, diff +17-18

The same import is added twice:

```diff
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures import ThreadPoolExecutor, as_completed
```

Python silently ignores duplicate imports, but this signals sloppy copy-paste and violates the project's code quality bar.

**Fix:** Remove one of the two lines.

---

### 2. CRITICAL — Duplicate constant definitions (copy-paste error)

**Confidence:** HIGH
**Location:** `mempalace/miner.py`, diff +60-63

Both constants are defined twice:

```diff
+BATCH_SIZE = 128  # chunks per upsert call
+MAX_WORKERS = min(32, (os.cpu_count() or 4) * 2)
+BATCH_SIZE = 128  # chunks per upsert call
+MAX_WORKERS = min(32, (os.cpu_count() or 4) * 2)
```

The second definitions silently overwrite the first. No functional harm today, but if someone later edits "the" definition and it's the wrong one, the change will be silently lost.

**Fix:** Remove the duplicate pair.

---

### 3. HIGH — Data loss on exception: unflushed batch is silently dropped

**Confidence:** HIGH
**Location:** `mine()` function — the `for future in as_completed(futures)` loop + final `flush_batch()`

`future.result()` is called without a try/except. If any file causes an unexpected exception in `process_file_cpu`, the exception propagates out of the loop, exits the `with ThreadPoolExecutor` block, and the final `flush_batch()` is never reached:

```python
        with ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
            futures = { ... }
            for future in as_completed(futures):
                result = future.result()       # <-- unhandled exception
                # ...
                    if len(batch_ids) >= BATCH_SIZE:
                        flush_batch()
                # ...

        flush_batch()  # <-- NEVER REACHED if exception above
```

All chunks accumulated in `batch_ids` / `batch_docs` / `batch_metas` from successfully processed files that haven't been flushed yet are permanently lost. On a large codebase with BATCH_SIZE=128, this could be up to 127 chunks worth of work silently discarded.

**Fix:**

```python
with ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
    futures = { ... }
    for future in as_completed(futures):
        try:
            result = future.result()
        except Exception as exc:
            fp = futures[future]
            print(f"  ✗ {fp.name}: {exc}")
            continue
        # ... rest of loop ...

flush_batch()  # now always reached
```

Or at minimum, move `flush_batch()` into a `finally` block.

---

### 4. MEDIUM — Logic duplication between `process_file` and `process_file_cpu`

**Confidence:** HIGH

Both functions implement the same sequence: read file → strip → MIN_CHUNK_SIZE check → detect_room → chunk_text → build metadata with hashlib drawer IDs. The only difference is `process_file` calls `add_drawer` (single ChromaDB upserts) while `process_file_cpu` returns records for batching.

`process_file` is still used for the `dry_run` path. If someone later changes room detection, metadata fields, or chunking logic, they must update both functions or the dry_run output will diverge from real behavior.

**Fix:** Refactor `process_file` to delegate to `process_file_cpu`:

```python
def process_file(filepath, project_path, collection, wing, rooms, agent, dry_run):
    if not dry_run and file_already_mined(collection, str(filepath), check_mtime=True):
        return 0, None

    result = process_file_cpu(filepath, project_path, wing, rooms, agent)
    if result is None:
        return 0, None

    source_file, room, records = result
    if dry_run:
        print(f"    [DRY RUN] {filepath.name} → room:{room} ({len(records)} drawers)")
        return len(records), room

    for drawer_id, content, meta in records:
        collection.upsert(documents=[content], ids=[drawer_id], metadatas=[meta])
    return len(records), room
```

---

### 5. MEDIUM — Behavioral change in `files_skipped` count

**Confidence:** HIGH

| | Original | New |
|---|---|---|
| `files_skipped` counts | Any file with drawers == 0 (already-mined + empty/tiny) | Only `file_already_mined` matches |

The output message says "Files skipped (already filed)" so the new behavior is arguably more correct, but it's an undocumented change. Files that are too small (`< MIN_CHUNK_SIZE`) or produce no chunks will no longer appear in the skip count.

**Suggestion:** Document this behavioral change in the PR description, or add a separate counter for truly empty/tiny files.

---

### 6. LOW — No progress feedback during pre-filtering

**Confidence:** MEDIUM

```python
pending = [fp for fp in files if not file_already_mined(collection, str(fp), check_mtime=True)]
```

For large file lists (thousands of files), this fires a ChromaDB query per file with no output. Users may think the tool is hung.

**Suggestion:** Add a brief progress line before the list comprehension.

---

## Thread Safety Analysis

| Component | Thread-safe? | Evidence |
|---|---|---|
| `process_file_cpu` | Yes | Pure: reads input args, returns new tuple, no shared state |
| `chunk_text` | Yes | Pure function: string slicing on immutable inputs |
| `detect_room` | Yes | Reads from `rooms` (shared but immutable), local `defaultdict` |
| `datetime.now()` | Yes | Thread-safe in CPython |
| `os.path.getmtime` | Yes | OS-level read-only call |
| `batch_ids/docs/metas` | N/A | Only touched from main thread |
| `collection.upsert` | N/A | Only called from main thread (flush_batch) |

The concurrency model (threads do CPU work, main thread does all ChromaDB I/O) is sound.

---

## Flow Impact Analysis

- **`process_file`**: Still called in the `dry_run` path. No external callers found outside `miner.py`.
- **`mine()`**: Public API — signature unchanged. Callers unaffected.
- **`file_already_mined`**: Now called in a sequential pre-filter rather than inside the per-file loop. No functional impact.
- **Blast radius**: Low — changes are internal to the `mine()` hot path.

---

## Guidelines Compliance

| Guideline | Status |
|---|---|
| Conventional commits (`perf: ...`) | Pass |
| No new dependencies (ThreadPoolExecutor is stdlib) | Pass |
| Docstrings on new function | Pass |
| Tests updated | **Missing** — no test changes for the new parallel path |
| Ruff formatting (100-char lines) | Likely pass |

---

## Summary

| Priority | Count | Items |
|---|---|---|
| Critical | 2 | Duplicate imports (#1), duplicate constants (#2) |
| High | 1 | Data loss on exception (#3) |
| Medium | 2 | Logic duplication (#4), behavioral change (#5) |
| Low | 1 | No pre-filter progress (#6) |

tbaldur:
- **Duplicate import/constants** — fixed in this PR, won't appear in future changes
- **Data loss on exception** — wrapped the futures loop in `try/finally` so `flush_batch()` always runs
- **Logic duplication** — `process_file` now delegates to `process_file_cpu`; shared logic lives in one place
- **files_skipped accuracy** — now counts both already-mined and too-small/unreadable files
- **Pre-filter progress** — added `"Checking N files for changes..."` print before the mtime check loop

FabioLissi:
Great work on this PR, the parallelization pattern for miner.py is really clean and the batched upserts make a measurable difference.

One thing I noticed while testing this branch on a large Claude Code conversation archive (1.5 GB of JSONL in /.claude/projects/): convo_miner.py still uses the sequential per-chunk collection.add() path, so users mining conversation corpora via mempalace mine <dir> --mode convos don't benefit from the new batching.

I ported your pattern to convo_miner.py on my fork: [FabioLissi/mempalace@fe74a7d](https://github.com/FabioLissi/mempalace/commit/fe74a7d) (https://github.com/FabioLissi/mempalace/commit/fe74a7d). Same structure as your miner.py changes — process_convo_file_cpu worker function, ThreadPoolExecutor(max_workers=MAX_WORKERS) for parallel normalize/chunk, batched collection.upsert() with BATCH_SIZE=128, try/finally around the executor to guarantee final flush on exception.

Benchmark on a 10-file, 502-drawer sample from my Claude Code history (M-series Mac, MPS + BGE-base):
• Before any batching: 23.35s
• With your miner.py batching only: unchanged (convo path not covered)
• With the convo_miner.py port: 5.69s (4.1x speedup)

All 556 tests pass including tests/test_convo_miner.py::test_convo_mining which exercises the real ChromaDB write path end-to-end.

Totally happy to:
• Let you cherry-pick the commit onto this PR branch
• Open it as a separate follow-up PR after this one merges
• Or skip it entirely if you'd rather keep this PR scoped to miner.py

Whichever works best for you. Thanks again for the contribution!

web3guru888 (commented):
## Review: Parallel File Processing + Batched Upserts

Performance-focused PR. The two optimizations (threaded I/O + batched upserts) are independent and compound well. Let me dig into the thread safety and correctness.

### Thread Safety Analysis

**`process_file_cpu()` — Thread-safe? ✅**
- `filepath.read_text()` — file reads are inherently thread-safe (OS handles concurrent reads)
- `detect_room()` — appears to be a pure function (reads file content and matches patterns)
- `chunk_text()` — pure function (splits text into chunks)
- `hashlib.sha256()` — thread-safe in Python (each call creates its own hasher)
- `datetime.now().isoformat()` — thread-safe
- `os.path.getmtime()` — thread-safe (stat call)

No shared mutable state. Good separation. The key design decision — extracting CPU work from ChromaDB access — is correct.

### Batched Upserts

**BATCH_SIZE = 128** — Reasonable. ChromaDB's ONNX backend computes embeddings as a batch matrix operation, so 128 chunks in one `upsert()` call is significantly faster than 128 individual calls. For reference, we use batch sizes of 100 in our integration with similar performance gains.

**⚠️ `flush_batch()` as inner function with closure over mutable lists:**
```python
batch_ids, batch_docs, batch_metas = [], [], []
def flush_batch():
    if batch_ids:
        collection.upsert(documents=batch_docs, ids=batch_ids, metadatas=batch_metas)
        batch_ids.clear()
        batch_docs.clear()
        batch_metas.clear()
```
This works because `list.clear()` mutates in-place. But it's a subtle coupling — if someone later refactors to `batch_ids = []` inside `flush_batch()`, it would silently break (rebinding the local variable without affecting the closure). Consider passing the lists as arguments, or using a small dataclass.

**✅ `try/finally` on flush** — Good. If a future raises an exception, the accumulated batch still gets flushed:
```python
try:
    with ThreadPoolExecutor(...) as executor:
        ...
finally:
    flush_batch()
```
This prevents data loss on errors.

### Pre-filtering Optimization

```python
pending = [fp for fp in files if not file_already_mined(collection, str(fp), check_mtime=True)]
```
This pre-filters *before* the thread pool. Smart — it avoids sending already-mined files to worker threads and prevents ChromaDB reads from concurrent threads (which could cause contention on the ChromaDB lock).

**⚠️ However:** This scans *all* files sequentially with `file_already_mined()` before any parallel work starts. For a large project with mostly-mined files (incremental re-mine scenario), this sequential pre-scan could dominate the wall time. Consider batching the `file_already_mined` checks too, or using the content hash approach from #424 as a faster pre-check.

### MAX_WORKERS

```python
MAX_WORKERS = min(32, (os.cpu_count() or 4) * 2)
```
Standard formula from Python docs for I/O-bound thread pools. Since `process_file_cpu` is mostly file I/O + CPU-light text processing, this is appropriate. The `min(32, ...)` cap prevents oversubscription.

### Dry-run path

Good that `dry_run` still uses the original serial `process_file()` path. The dry-run path doesn't need performance and keeping it simple avoids introducing bugs in the diagnostic path.

### Test coverage

`test_process_file_cpu_returns_records` — Good unit test for the new extracted function.
`test_mine_parallel_produces_same_result_as_serial` — The critical correctness test. Verifies that parallel mining indexes the same files. However, it doesn't compare *exact drawer content* between serial and parallel — just that files are present and count > 0. Consider asserting that the drawer count matches between serial and parallel runs on identical input.
`test_mine_skips_already_mined_files_on_rerun` — Incremental behavior preservation test. Important.

Overall: well-engineered performance improvement with correct thread safety boundaries. The `flush_batch` closure pattern and the sequential pre-scan are the main areas for potential refinement.

tbaldur:
Thank you for the analysis, changes made:
  - flush_batch now takes explicit (ids, docs, metas) args
  - Added idempotency assertion to the parallel test (re-mine same data, verify count unchanged)
  - Pre-scan optimization: replaced the sequential file_already_mined() loop with build_mined_cache(),  one bulk ChromaDB fetch + dict lookups. Greatly improved on my 32k+ files project.



 -  FabioLissi herry-picked your convo_miner.py port onto this branch — you're credited as author on the commit. Applied the same flush_batch hardening (explicit args) for consistency with the miner.py changes. Thank you very much for the contribution and the benchmark numbers!

igorls:

PR #417, opened 2026-04-09
feat: add --include-assistant flag to mine-convos for standalone assi…

dhiaa2:
## What this does

Adds an `--include-assistant` flag to the `mine_convos` pipeline and CLI.

By default, only user turns are indexed. This change allows assistant responses to be optionally indexed as well, with `role: assistant` metadata so they can be filtered separately.

This fixes the documented weakness in the single-session-assistant benchmark category (92.9%), where questions about what the assistant said could not be answered because assistant turns were never stored.

## Changes

- `convo_miner.py` — new `_extract_assistant_turns()` function, `include_assistant` param on `mine_convos()`
- `cli.py` — new `--include-assistant` flag wired to `mine_convos()`
- `tests/test_convo_miner.py` — two new tests covering assistant indexing and default-off behavior
- `tests/test_cli.py` — updated mock assertion to include new parameter

## Testing

All existing tests pass. New tests confirm assistant chunks are indexed with correct metadata and the flag is off by default.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1207,7 +1207,7 @@ def main():
         "--include-assistant",
         action="store_true",
         dest="include_assistant",
-        help="Also index each assistant turn as a standalone drawer with turn_role=assistant metadata (convos mode only)",
+        help="Also index each assistant turn as a standalone drawer with turn_role=assistant metadata (convos mode only). Expects Claude export format: user turns begin with '>'; all other non-separator lines are treated as assistant content.",
     )
 
     # sweep
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -514,6 +514,7 @@ def cmd_mine(args):
                 limit=args.limit,
                 dry_run=args.dry_run,
                 extract_mode=args.extract,
+                include_assistant=getattr(args, "include_assistant", False),
             )
         else:
             from .miner import mine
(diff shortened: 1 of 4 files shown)


web3guru888 (commented):
Good feature — this directly addresses the single-session-assistant benchmark gap. Being able to retrieve "what did the assistant say about X" is a real need.

The implementation looks clean:
- Default-off is the right call (most users don't want 2x the drawers)
- The `turn_role=assistant` metadata makes filtering straightforward
- `asst_` ID prefix prevents collisions with user-turn drawers

One thing to watch: the `_extract_assistant_turns()` parser assumes non-`>` non-`---` lines are assistant turns. If someone has a transcript with different formatting (e.g., `Assistant:` prefix lines), those would be misclassified. Might be worth a brief note in the help text about expected transcript format.

dhiaa2:
Good catch. The parser is designed specifically for Claude export format — user turns 
start with `>`, everything else is treated as assistant content. In a differently-formatted 
transcript (e.g., `User:` / `Assistant:` prefixes without `>` markers), both user and 
assistant lines would be captured together as assistant content, not just the `Assistant:` 
lines. The docstring already notes the `>` marker format requirement — I'll add the same 
note to the `--include-assistant` CLI flag help text so it's visible before the flag is used.

dhiaa2:
Help text updated per your suggestion — `--include-assistant` now notes that user turns begin with `>` in Claude export format, and that all other non-separator lines are treated as assistant content. Pushed to the branch.

igorls:

dhiaa2:
Done — Rebased onto develop — both commits apply cleanly on top of 02dd6dc (latest develop HEAD as of May 8). Full test suite: 1648 passed, 0 failed. web3guru88's transcript format concern was addressed in the second commit (help text now documents that user turns begin with > in Claude export format). Ready for review.

PR #423, opened 2026-04-09
fix: honor configured drawer collection everywhere

gut-puncture:
## Summary

This fixes a trust problem in the current tool: different parts of MemPalace could look at different drawer collections, so a user could mine memories successfully and then have search, wake-up, status, repair, or MCP read from somewhere else.

In simple terms, this PR makes the tool use the same default drawer collection everywhere it already promises to use it.

## What changed

- mining, search, layers, CLI status, CLI repair, CLI compress, and MCP drawer tools now all resolve the same configured default drawer collection
- `palace_path` from config and env now expands `~`
- `python -m mempalace.mcp_server --palace ~/...` now resolves `~` correctly
- read paths no longer create a fake empty palace or collection when the palace is actually missing
- metadata-based readers now page instead of truncating at 10,000 drawers, so large-palace status and taxonomy counts stay correct
- MCP collection caching is now keyed by resolved `(palace_path, collection_name)` so context changes do not go stale

## What this does not change

- no new commands or flags
- no public collection override
- no knowledge-graph redesign
- no deleted-file cleanup or refresh redesign

## Tests

- added regressions for `~` expansion from config and env
- added regressions proving project and convo mining write to the configured collection
- added regressions proving search, layers, and MCP read from that same configured collection
- added regressions for exact counts above 10,000 drawers
- added regressions for MCP cache rekeying and create-false/create-true behavior

Validation run locally:
- `.venv/bin/pytest tests/test_config.py tests/test_palace.py tests/test_miner.py tests/test_convo_miner.py tests/test_searcher.py tests/test_layers.py tests/test_mcp_server.py tests/test_cli.py`
- `.venv/bin/ruff check mempalace tests`
- `.venv/bin/python -m pytest`

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -34,6 +34,7 @@ import argparse
 from pathlib import Path
 
 from .config import MempalaceConfig
+from .palace import get_drawer_collection, resolve_drawer_context
 
 
 def cmd_init(args):
@@ -101,7 +102,8 @@ def cmd_mine(args):
 def cmd_search(args):
     from .searcher import search, SearchError
 
-    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+    cfg = MempalaceConfig()
+    palace_path = os.path.expanduser(args.palace) if args.palace else cfg.palace_path
     try:
         search(
             query=args.query,
@@ -109,6 +111,7 @@ def cmd_search(args):
             wing=args.wing,
             room=args.room,
             n_results=args.results,
+            config=cfg,
         )
     except SearchError:
         sys.exit(1)
@@ -161,8 +164,9 @@ def cmd_migrate(args):
 def cmd_status(args):
     from .miner import status
 
-    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
-    status(palace_path=palace_path)
+    cfg = MempalaceConfig()
+    palace_path = os.path.expanduser(args.palace) if args.palace else cfg.palace_path
+    status(palace_path=palace_path, config=cfg)
 
 
 def cmd_repair(args):
@@ -170,7 +174,9 @@ def cmd_repair(args):
     import chromadb
     import shutil
 
-    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+    cfg = MempalaceConfig()
+    palace_arg = os.path.expanduser(args.palace) if args.palace else cfg.palace_path
+    palace_path, collection_name = resolve_drawer_context(palace_path=palace_arg, config=cfg)
 
     if not os.path.isdir(palace_path):
         print(f"\n  No palace found at {palace_path}")
@@ -183,9 +189,13 @@ def cmd_repair(args):
 
     # Try to read existing drawers
     try:
+        col = get_drawer_collection(palace_path=palace_path, create=False, config=cfg)
+        if col is None:
+            print(f"\n  No palace found at {palace_path}")
+            return
         client = chromadb.PersistentClient(path=palace_path)
-        col = client.get_collection("mempalace_drawers")
         total = col.count()
+        print(f"  Collection: {collection_name}")
         print(f"  Drawers found: {total}")
     except Exception as e:
         print(f"  Error reading palace: {e}")
@@ -220,8 +230,8 @@ def cmd_repair(args):
     shutil.copytree(palace_path, backup_path)
 
     print("  Rebuilding collection...")
-    client.delete_collection("mempalace_drawers")
-    new_col = client.create_collection("mempalace_drawers")
+    client.delete_collection(collection_name)
+    new_col = client.create_collection(collection_name)
 
     filed = 0
     for i in range(0, len(all_ids), batch_size):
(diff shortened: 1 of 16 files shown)


web3guru888 (commented):
## Review: Honor Configured Drawer Collection Everywhere

This is a significant correctness fix. The core problem — different code paths silently using different collections — is exactly the kind of bug that's invisible in testing but devastating in production. We hit a variant of this when our integration's `knowledge_graph_bridge.py` was writing to one collection while the MCP search read from another.

### Architecture

The new `resolve_drawer_context()` + `get_drawer_collection()` + `open_collection_on_client()` factoring in `palace.py` is clean. Single source of truth for collection resolution, config-aware, with proper `create` flag semantics.

**`iter_collection_metadatas()`** — The paginated iteration is the right fix for the `limit=10000` hard cap. The offset-based pagination pattern is correct:
```python
offset += batch_count
if batch_count < batch_size:
    break
```
This handles the "last batch smaller than page size" termination correctly.

### Specific observations

**✅ `~` expansion from config/env** — `_expand_path()` is a good single-point normalization. Previously `MEMPALACE_PALACE_PATH=~/palace` would silently work differently depending on whether the shell expanded `~` before Python saw it. Now it's consistent.

**✅ MCP cache rekeying** — The `_collection_cache_key = (palace_path, collection_name)` tuple ensures that changing either dimension invalidates the cache. The `test_get_collection_rekeys_when_context_changes` test covers path change, name change, and both change. Thorough.

**✅ `create=False` doesn't materialize empty palaces** — `get_drawer_collection(create=False)` returns `None` if the directory is missing. This prevents `mempalace status` from creating empty palace directories as a side effect. We added the same guard in our integration and it eliminated a confusing class of "phantom palace" bugs.

**⚠️ Potential issue in `cmd_repair()`:**
```python
palace_path, collection_name = resolve_drawer_context(palace_path=palace_arg, config=cfg)
```
Then later:
```python
col = get_drawer_collection(palace_path=palace_path, create=False, config=cfg)
```
Then:
```python
client = chromadb.PersistentClient(path=palace_path)
```
The repair flow creates a *new* client after already getting a collection through `get_drawer_collection`. This means two separate `PersistentClient` instances pointing at the same directory. ChromaDB's locking should handle this, but it's a bit fragile. Consider getting the client from the same path as the collection, or restructuring to use a single client.

**Test coverage** — Excellent. The regression tests for tilde expansion, configured collection mining, paged status, and cache rekeying are well-targeted. The `_PagedCollection` helper in tests is a clean way to test pagination without real ChromaDB overhead.

### Scope

16 files changed, 611 additions, 222 deletions — this is wide-ranging but every change is consistent with the stated goal. No scope creep. The "What this does not change" section in the description is appreciated.

Strong PR. Addresses a real trust problem in the tool.

skuznetsov:
Quick merge-order note after comparing the overlap with #413:

I think `#413` should land first, then `#423` should rebase on top of it.

Reason:
- both PRs touch the same collection-access surface: `palace.py`, `mcp_server.py`, `miner.py`, `layers.py`, `searcher.py`, plus overlapping tests
- `#413` is the lower-level storage seam / access abstraction
- `#423` is behavioral logic on top of that same surface: configured drawer resolution, `~` expansion, keyed MCP cache, paged metadata reads, and consistent read-path semantics

So they do not look like competing PRs to me; they look ordered.

If `#423` lands first, then rebasing `#413` later has a higher chance of accidentally flattening or reworking the new drawer-context behavior while introducing the seam.

If `#413` lands first, then `#423` can rebase onto the seam and carry its behavior changes forward more directly.

In other words: `#413` establishes the collection-access seam, `#423` refines the configured-drawer behavior on top of that seam.

Not a blocker for the design in `#423` itself — just a merge-order caution so we do not create avoidable rebase churn or behavioral regressions.

gut-puncture:
Restacking this onto current `main` now that the overlapping collection-access surface has moved upstream. I’m rebuilding it from fresh `main`, keeping the same goal and scope: consistent configured drawer context, exact metadata reads, and no new feature surface.

gut-puncture:
Restack is now live on this PR branch. I rebuilt it on top of current `main` after the collection-access seam changes landed, kept the same user-facing scope, and reran the focused lane for this change:

- `ruff check mempalace tests`
- `pytest tests/test_config.py tests/test_palace.py tests/test_searcher.py tests/test_layers.py tests/test_mcp_server.py tests/test_miner.py tests/test_convo_miner.py tests/test_cli.py`

Result: `162 passed`.

I also folded in one small cleanup from review while restacking: `cmd_repair()` now stays on the resolved collection name without mixing in an extra hardcoded collection lifecycle.

gut-puncture:
@skuznetsov  @web3guru888  restack is now live on this PR branch.

This is now rebuilt on top of current main, so the earlier merge-order concern should be resolved. I kept the same scope and reran the focused validation lane for this change:

ruff check mempalace tests
pytest tests/test_config.py tests/test_palace.py tests/test_searcher.py tests/test_layers.py tests/test_mcp_server.py tests/test_miner.py tests/test_convo_miner.py tests/test_cli.py
Result: 162 passed.

web3guru888 (approved):
Took another pass on the restacked diff — the merge-order concern is resolved.

The collection-access seam changes that were in conflict with `#413` are now on `main`, and your restack picks them up cleanly. The `mcp_server.py` diff in particular shrank by ~100 lines (net -44) which is exactly what you'd expect after the upstream abstraction landed — less bespoke resolution code needed here.

162 passing on the focused lane (config, palace, searcher, layers, mcp_server, miner, convo_miner, cli) and ruff clean is the right validation set for this change.

The scope is still what I'd want: consistent collection resolution, `~` expansion from config/env, no create-on-read false positives, paginated metadata reads for large palaces, and keyed MCP cache. No scope creep from the restack.

LGTM — approving.

bensig:

PR #425, opened 2026-04-09
fix: add ensure_ascii=False to json.dumps in MCP server

barry3406:
Fixes #359

`json.dumps` in the MCP server didn't set `ensure_ascii=False`, so non-ASCII text (Chinese, Japanese, etc.) would either get escaped to `\uXXXX` or crash outright when lone surrogate characters showed up from certain MCP client encoding paths.

**Before:**
```python
json.dumps(result, indent=2)          # line 907 — tool result
json.dumps(response)                  # line 937 — stdout writer
```
Chinese text → `UnicodeEncodeError` or garbled `\uXXXX` output.

**After:**
```python
json.dumps(result, indent=2, ensure_ascii=False)
json.dumps(response, ensure_ascii=False)
```
Chinese text → preserved as-is in the JSON output.

Added two tests in `test_mcp_server.py`:
- `test_tool_result_preserves_unicode` — end-to-end through `handle_request` with Chinese text
- `test_main_loop_writes_unicode` — verifies the serialization doesn't crash

All 35 tests pass.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -904,7 +904,11 @@ def handle_request(request):
             return {
                 "jsonrpc": "2.0",
                 "id": req_id,
-                "result": {"content": [{"type": "text", "text": json.dumps(result, indent=2)}]},
+                "result": {
+                    "content": [
+                        {"type": "text", "text": json.dumps(result, indent=2, ensure_ascii=False)}
+                    ]
+                },
             }
         except Exception:
             logger.exception(f"Tool error in {tool_name}")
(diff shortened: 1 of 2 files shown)


web3guru888 (commented):
Clean fix, +1. We use `ensure_ascii=False` in our integration for exactly this reason — scientific terms with Greek letters (α, β, γ), mathematical symbols (∑, ∇), and CJK references get mangled into `\uXXXX` escape sequences without it.

The change covers both the right spots: the tool result serialization in `handle_request` and the stdout writer in `main()`. Good that the tests cover round-trip serialization with actual CJK characters rather than just asserting no exception.

One consideration for downstream: any MCP client that was parsing the escaped `\uXXXX` sequences will now see raw UTF-8 bytes instead. This is correct per the JSON spec (both are valid), but if anyone was doing naive string matching on the escaped form, they'd need to update. Probably not a concern in practice.

igorls:

igorls:
Thanks! This exact change already landed on develop (commit 7b49478). Closing as a duplicate.

PR #426, opened 2026-04-09
chore(deps): pin chromadb to >=1.0,<2

### LINKED ISSUE #445: v3.1.0 ChromaDB pin breaks existing palace created with Chroma 1.x
## Summary

After upgrading to `mempalace v3.1.0`, my existing palace can no longer be opened.

The CLI/MCP reports:

```text
No palace found at /Users/isapozhnik/.mempalace/palace
Run: mempalace init <dir> then mempalace mine <dir>
```

The palace is not missing. The Chroma DB exists and contains drawers.

## What changed

`v3.1.0` pins ChromaDB to:

```toml
chromadb>=0.5.0,<0.7
```

This appears to be incompatible with an existing palace DB created/opened with Chroma 1.x.

## Evidence

With the published `v3.1.0` tool environment:

```text
mempalace 3.1.0
chromadb 0.6.3
```

Opening the existing palace fails internally with:

```text
KeyError: '_type'
```

The CLI catches that and prints the misleading `No palace found` message.

The same existing palace opens successfully with Chroma 1.5.7:

```text
chromadb 1.5.7
count 8429
```

I also tested `mempalace v3.1.0` source locally with only the Chroma dependency changed to:

```toml
chromadb==1.5.7
```

That successfully opened the same palace and search worked.

## Repro

```sh
uv tool upgrade mempalace
mempalace status
```

Result:

```text
No palace found at ~/.mempalace/palace
```

But the DB exists:

```sh
ls ~/.mempalace/palace/chroma.sqlite3
```

And Chroma 1.5.7 can read it.

## Expected

`mempalace v3.1.0` should either:

1. Support existing Chroma 1.x palace databases, or
2. Detect incompatible Chroma schema and show a clear migration/rollback message, not `No palace found`.

## Current workaround

Install v3.1.0 from source with the Chroma pin changed to `chromadb==1.5.7`.

## Environment

- macOS
- Installed via `uv tool upgrade mempalace`
- Palace path: `~/.mempalace/palace`
- Existing DB contains 8429 drawers

jccidc:
## What does this PR do?

Fixes installs for users whose palaces were created with chromadb 1.x (see #445). The current constraint `chromadb>=0.5.0,<0.7` forces installs onto the 0.5/0.6 line, which uses an older on-disk format. Palaces built on chromadb 1.x can't be opened by 0.6 — `get_collection` fails with `KeyError: '_type'` when reading collection config JSON, and the CLI surfaces it as a misleading "No palace found" message.

Pin to the 1.x line so:

- editable installs keep working with palaces built on 1.x
- new installs land on the same major as the current chromadb release (1.5.x)
- we have a single supported major instead of straddling two on-disk formats

## Known break and migration note

**This change does break users with existing chromadb 0.5/0.6 palaces.** chromadb 1.x does not auto-migrate the 0.6 on-disk format — the [chromadb 1.0 migration guide](https://docs.trychroma.com/docs/migration/) covers a dedicated process. For mempalace specifically, the simplest recovery is to back up `~/.mempalace/palace/` and re-mine the source files into a fresh 1.x palace.

A proper in-tool migration command (`mempalace migrate-chroma` or similar) is a separate concern and belongs in a follow-up PR. This PR is intentionally scoped to the dependency pin.

## API compatibility audit

All chromadb APIs used by mempalace are part of the stable core interface and work identically on 0.5 → 1.x:

| API                                   | Callers                                                        |
| ------------------------------------- | -------------------------------------------------------------- |
| `chromadb.PersistentClient(path=...)` | `palace.py`, `mcp_server.py`, `layers.py`, `searcher.py`, `cli.py` |
| `client.get_collection(name)`         | `palace.py`, `mcp_server.py`, `layers.py`, `searcher.py`, `cli.py` |
| `client.create_collection(name)`      | `palace.py`                                                    |
| `client.get_or_create_collection(name)` | `mcp_server.py`                                              |
| `collection.get(where=..., limit=...)` | `palace.py`, `miner.py`, `mcp_server.py`                     |
| `collection.count()`                  | `mcp_server.py`, `cli.py`, `layers.py`                         |
| `collection.query(...)`               | `searcher.py`, `layers.py`                                     |
| `collection.add(...)`                 | `miner.py`, `convo_miner.py`                                   |
| `collection.delete(...)`              | `mcp_server.py`                                                |

No code changes needed.

## How to test

1. Install the branch in a fresh venv: `pip install -e .`
2. Confirm chromadb 1.x resolves: `python -c "import chromadb; print(chromadb.__version__)"` → `1.5.x`
3. Run the test suite: `python -m pytest tests/ -v` (passes locally against chromadb 1.5.7)
4. If you have an existing 1.x palace, verify `mempalace status` opens it without a `KeyError: '_type'`.

## Replaces

This replaces #365, which accumulated stale diffs from being rebased off an older main. This PR is a single clean commit off current `main`.

## Related

- Fixes #445 (1.x palace can't be opened under current pin)

## Checklist
- [x] Single-line dependency change, no new dependencies
- [x] All chromadb API calls audited for 1.x compatibility
- [x] Conventional commit title
- [x] Known migration impact documented
- [ ] Tests pass in CI (ran locally — need maintainer CI run)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ classifiers = [
     "Topic :: Utilities",
 ]
 dependencies = [
-    "chromadb>=0.5.0,<0.7",
+    "chromadb>=1.0,<2",
     "pyyaml>=6.0,<7",
 ]
 


iSapozhnik:
Will this also solve [this](https://github.com/milla-jovovich/mempalace/issues/445) problem?

web3guru888 (commented):
This goes in the right direction — issue #445 reports that users who installed ChromaDB 1.x (which pip resolves by default now) can't open palaces created under the current `>=0.5.0,<0.7` pin, because ChromaDB 1.0 changed its storage format.

But bumping straight to `>=1.0,<2` without a migration path will break anyone running `0.5.x` or `0.6.x` who upgrades mempalace:

1. **Existing palaces won't open** — ChromaDB 1.0 can't read 0.5/0.6 SQLite storage directly
2. **No `mempalace repair` guidance** — the CLI repair command re-indexes vectors but doesn't handle the underlying Chroma storage format migration
3. **The current test suite runs against 0.5/0.6** — has anyone verified all tests pass against chromadb 1.x?

Suggestion: consider `>=0.5.0,<2` (widen the range rather than bump the floor), and add a note in the upgrade docs that users on Chroma 1.x should run `mempalace repair` after upgrading. The `repair` command already does a full re-index which should handle the format gap — but it needs to be tested against the 0.5→1.0 transition.

@iSapozhnik re your question — yes, this should solve #445, but only if mempalace's own code is compatible with the ChromaDB 1.x API. The pin change alone isn't enough if there are breaking API changes upstream.

jccidc:
Thanks @web3guru888 — you're right that my original PR body overclaimed on migration. I said "auto-migrates on first open" but after checking the chromadb 1.0 release notes, that's not actually the case — there's a migration guide, not an automatic migration. I'll strike that line.

On the constraint choice: widening to `>=0.5.0,<2` doesn't actually help new installs — pip will resolve to the latest (1.5.x) regardless, and anyone with an existing 0.5/0.6 palace hits the same break. The only user who benefits from widening is one who already has chromadb 0.6 installed from a prior version and doesn't force pip to upgrade. That's a narrow window.

The core problem is that **there's no way to both** (a) fix #445 for 1.x users **and** (b) avoid breaking 0.6 users **without a code-level compatibility layer or migration command**. A constraint change alone can only pick one side.

Given that:

- #445 is an actively broken user today
- mempalace's entire API surface (`PersistentClient`, `get_collection`, `count`, `get`, `query`, `add`, `delete`) is stable across the 0.5 → 1.x boundary — no code changes needed for the jump
- chromadb 1.x is the current major (1.5.7) and what new installs get by default anyway

I think the right scope for **this** PR is:

1. Keep the `>=1.0,<2` pin so new installs and editable installs from source both land on 1.x
2. Explicitly document the break in the release notes for the next version
3. Recommend users re-mine their source files rather than try to migrate chromadb's on-disk format

A proper migration story (detect old format, run `mempalace repair` against a chromadb 0.6 read-only client, re-mine into 1.x) is a separate PR. I can open a follow-up issue to track it if that'd be useful.

On your other points:

- **Test coverage**: good catch, CI doesn't currently test against chromadb 1.x. I can add a matrix entry in a separate PR, but the existing test suite (`tests/test_mcp_server.py`, `tests/test_miner.py`, etc.) does pass against 1.5.7 locally — I ran them on the v2 branch before pushing.
- **`mempalace repair`**: I looked at `cli.py:160` — it calls `client.get_collection()` first, which hits the same `KeyError: '_type'` on 0.5/0.6 → 1.x transitions. So repair in its current form can't bridge the format gap. That's the migration-command gap I mentioned above.

Let me know if you'd rather I close this and wait for a proper migration PR, or merge this as-is with a release-notes callout.

director-LAC:
Adding weight to this: on Python 3.14, chromadb 0.6.x doesn't just cause data incompatibility, it fails to import entirely due to pydantic.v1 incompatibility (#487). The <0.7 pin has no viable target on 3.14. Widening to >=1.0 is the only path forward for current Python versions.

web3guru888:
You've thought this through carefully, and I agree with the analysis. The narrow-window argument is right: widening to `>=0.5,<2` only protects users who happen to have chromadb 0.6 already installed and aren't running `pip install --upgrade`, which is a shrinking population.

**Recommendation: merge as >=1.0,<2 with clear release notes.** The case for >=1.0 is:

1. New installs get 1.x anyway — this formalizes what's already true in practice
2. The 0.5/0.6 → 1.x break is a data format issue, not an API issue — existing users need to re-mine regardless of what the constraint says
3. `mempalace repair` being broken against 1.x is a separate bug that should be tracked and fixed independently

The migration command gap is worth a dedicated issue rather than blocking this PR. A proper `mempalace migrate --from 0.6` that spins up a temporary chromadb 0.5/0.6 reader client and re-ingests into 1.x would be the complete fix, but it's a multi-week effort.

One thing to add to your release notes callout: recommend users run `mempalace status` before upgrading to verify their palace is readable, and keep their palace directory backed up (even just `cp -r ~/.mempalace/palace ~/.mempalace/palace.backup`). Low-tech but effective.

igorls:

PR #427, opened 2026-04-09
fix(status): paginate metadata reads so palaces over 10k drawers report correctly

jccidc:
## What does this PR do?

Fixes a silent truncation bug where palaces with more than 10,000 drawers report `10000` as their total drawer count. Found this on a real palace with 39,334 drawers — both `mempalace status` (CLI) and the `mempalace_status` MCP tool returned `10000` instead of the actual total.

Both code paths called `col.get(limit=10000, ...)` and treated `len(metadatas)` as the total. The four affected MCP read tools (`mempalace_status`, `mempalace_list_wings`, `mempalace_list_rooms`, `mempalace_get_taxonomy`) and the CLI `status()` function all had the same pattern.

## Fix

- New shared helper `iter_all_metadatas(collection, where=None, page_size=10000)` in `palace.py` walks the collection in pages until an empty page is returned. Uses a clean `while True / break on empty` pattern that handles both unfiltered and `where`-filtered queries without needing `col.count()` up front.
- `mempalace status` (`miner.py`) uses `col.count()` for the authoritative total and `iter_all_metadatas()` for the per-wing / per-room breakdown.
- All four affected MCP read tools (`mcp_server.py`) import and use the shared helper — one implementation, four callers.
- The diary read path keeps its 10k limit since per-agent diary collections are small and the call is already filtered by wing+room.

## Replaces

This replaces the older #364. That branch was based on an older `main` and accumulated stale diffs; this PR is on a clean branch off current `main` and also incorporates the [Octocode review feedback](https://github.com/milla-jovovich/mempalace/pull/364#issuecomment-4216364410) from that PR:

- **Cleaner iterator** — `while True / break on empty` instead of `while offset < total`, so filtered queries no longer overshoot
- **Shared helper** — moved from `mcp_server.py` into `palace.py` so `miner.py` and `mcp_server.py` both use it, instead of duplicating the pagination inline
- **Regression test** — new `test_status_above_10k_drawers` seeds 10,500 drawers across two wings and two rooms and asserts `total_drawers`, `wings`, and `rooms` all reflect the full dataset

## How to test

```bash
python -m pytest tests/test_mcp_server.py::TestReadTools -v
```

All 8 tests pass, including the new regression test. The existing `test_status_with_data` / `test_list_wings` / etc. continue to pass unchanged.

## Manual verification

On a palace with 39,334 drawers:

```
$ mempalace status
  MemPalace Status — 39334 drawers

  WING: claude_sessions
    ROOM: technical            29058 drawers
    ROOM: architecture          5136 drawers
    ROOM: problems              1733 drawers
    ROOM: planning               531 drawers
    ROOM: general                485 drawers
    ROOM: decisions               54 drawers

  WING: projects
    ROOM: general               1501 drawers
    ROOM: agent_exploration      709 drawers
    ...
```

Previously this reported `10000` total with a truncated per-room breakdown.

## Checklist

- [x] Tests pass (`python -m pytest tests/test_mcp_server.py::TestReadTools -v`)
- [x] Ruff lint + format clean (`ruff check .` / `ruff format --check .`)
- [x] No hardcoded paths
- [x] New regression test for the >10k case
- [x] Addresses Octocode review feedback from #364

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -28,6 +28,7 @@ from pathlib import Path
 
 from .config import MempalaceConfig, sanitize_name, sanitize_content
 from .version import __version__
+from .palace import iter_all_metadatas
 from .searcher import search_memories
 from .palace_graph import traverse, find_tunnels, graph_stats
 import chromadb
@@ -144,8 +145,7 @@ def tool_status():
     wings = {}
     rooms = {}
     try:
-        all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
-        for m in all_meta:
+        for m in iter_all_metadatas(col):
             w = m.get("wing", "unknown")
             r = m.get("room", "unknown")
             wings[w] = wings.get(w, 0) + 1
@@ -201,8 +201,7 @@ def tool_list_wings():
         return _no_palace()
     wings = {}
     try:
-        all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
-        for m in all_meta:
+        for m in iter_all_metadatas(col):
             w = m.get("wing", "unknown")
             wings[w] = wings.get(w, 0) + 1
     except Exception:
@@ -216,11 +215,8 @@ def tool_list_rooms(wing: str = None):
         return _no_palace()
     rooms = {}
     try:
-        kwargs = {"include": ["metadatas"], "limit": 10000}
-        if wing:
-            kwargs["where"] = {"wing": wing}
-        all_meta = col.get(**kwargs)["metadatas"]
-        for m in all_meta:
+        where = {"wing": wing} if wing else None
+        for m in iter_all_metadatas(col, where=where):
             r = m.get("room", "unknown")
             rooms[r] = rooms.get(r, 0) + 1
     except Exception:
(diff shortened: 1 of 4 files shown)


web3guru888 (commented):
This is an important correctness fix. Silent truncation at 10k drawers is the kind of bug that goes unnoticed until someone scales up — and then everything looks fine (no error) but the data is wrong.

The fix is well-architected:

1. **Shared helper** — `iter_all_metadatas()` in `palace.py` is the right abstraction. One implementation, four callers in `mcp_server.py` plus `miner.py`. The generator pattern means memory usage stays flat regardless of palace size.

2. **`while True / break on empty`** — Better than `while offset < total` because it works correctly with `where` filters (where `col.count()` returns the unfiltered total, causing overshoot). Good catch from the Octocode review.

3. **`col.count()` for the authoritative total in status** — Correct. The pagination walk gives you the breakdown, but `count()` is the source of truth for the headline number. Prevents drift if the pagination logic ever has an off-by-one.

4. **Regression test** — Seeding 10,500 drawers across two wings/rooms and verifying the full count is exactly what this needs. The test also verifies per-wing and per-room breakdowns, not just the total.

5. **Diary path left alone** — Smart. Per-agent diary collections are small, so the 10k limit is fine there and the filtered query is already scoped.

We have 710+ KG entities and this pagination approach would be essential at scale. Minor note: #436 (doctor) and #435 (export) both do their own manual pagination with `limit=500, offset=...` — if this lands first, they should adopt `iter_all_metadatas()` instead of reimplementing the pattern.

bensig:

PR #429, opened 2026-04-09
fix: harden palace.py mtime check and expand test coverage

Nitrogonza9:
## Summary

- **Fix float equality bug** in `file_already_mined()` — replace exact `==` comparison with tolerance check (`< 0.01s`) to prevent false negatives caused by float serialization rounding through ChromaDB metadata
- **Add debug logging** to `get_collection()` fallback path so silent collection creation is traceable
- **New `test_palace.py`** with 11 direct unit tests covering `get_collection`, `file_already_mined` (mtime tolerance, modification detection, missing metadata), and `SKIP_DIRS`
- **New searcher edge-case tests** for nonexistent wing/room filters and special character queries

## Test plan

- [x] `pytest tests/test_palace.py -v` — all 11 new tests pass
- [x] `pytest tests/test_searcher.py -v` — all 17 tests pass (13 existing + 4 new)
- [x] `pytest tests/ -v` — full suite 549 passed, 0 failed

### DIFF
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -4,8 +4,12 @@ palace.py — Shared palace operations.
 Consolidates ChromaDB access patterns used by both miners and the MCP server.
 """
 
+import logging
 import os
+
 import chromadb
+
+logger = logging.getLogger("mempalace")
 
 SKIP_DIRS = {
     ".git",
@@ -45,6 +49,7 @@ def get_collection(palace_path: str, collection_name: str = "mempalace_drawers")
     try:
         return client.get_collection(collection_name)
     except Exception:
+        logger.debug("Collection %r not found, creating it", collection_name)
         return client.create_collection(collection_name)
 
 
@@ -65,7 +70,7 @@ def file_already_mined(collection, source_file: str, check_mtime: bool = False) 
             if stored_mtime is None:
                 return False
             current_mtime = os.path.getmtime(source_file)
-            return float(stored_mtime) == current_mtime
+            return abs(float(stored_mtime) - current_mtime) < 0.01
         return True
     except Exception:
         return False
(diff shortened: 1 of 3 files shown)


web3guru888 (commented):
Clean fix for a real bug. The float equality issue with mtime is a known pitfall — ChromaDB stores metadata values through a serialization layer that can introduce tiny rounding errors, making exact `==` comparisons fail silently and causing unnecessary re-mining.

The tolerance of `0.01s` is well-chosen. Filesystem mtime granularity varies:
- ext4: nanosecond
- FAT32: 2 seconds
- NTFS: 100 nanoseconds
- HFS+: 1 second

So 0.01s is tight enough to detect real modifications but loose enough to absorb serialization rounding. On FAT32 you'd need a larger tolerance, but that's a niche concern.

The debug logging on the `get_collection()` fallback path is a good addition — silent collection creation was a debugging pain point we hit too.

The test coverage is thorough:
- Exact mtime match ✅
- Tiny rounding difference (tolerance) ✅
- Real modification detection (10s delta) ✅
- Missing `source_mtime` metadata ✅

The searcher edge-case tests (nonexistent wing/room, special characters) are bonus value. We've seen special character queries cause issues with some ChromaDB versions, so having those tests is good regression coverage.

Nitrogonza9:
Thanks @web3guru888 — good context on the FAT32 granularity. If someone reports issues on FAT32 we can bump the tolerance in a follow-up, but for now 0.01s covers the realistic cases.

Appreciate you reviewing all six PRs. Solid feedback across the board.

— Gonzalo

web3guru888:
Makes sense — 0.01s covers FAT32 would need 2s in theory, but anyone running MemPalace on a FAT32 drive is solving an unusual problem. If it comes up, bumping the constant is a one-line fix.

Thanks for the thorough review across all the PRs, Gonzalo — good work on these.

bensig:
closing — the float equality fix landed via #610 (epsilon mtime comparison, merged). the debug logging part is a good idea if you want to submit it separately. thanks @Nitrogonza9!

PR #434, opened 2026-04-09
feat: auto-populate knowledge graph from palace drawers

Nitrogonza9:
## Summary

The Knowledge Graph is powerful but empty for most users — it requires manual `kg_add` calls for every fact. This PR bridges that gap by **automatically extracting relationships from existing palace drawers** into the KG.

### New module: `mempalace/kg_extractor.py`

Reads verbatim drawer content and extracts 8 relationship types using rule-based pattern matching:

| Pattern | Example | KG Triple |
|---------|---------|-----------|
| Employment | "Alice works at Acme Corp" | Alice → works_at → Acme Corp |
| Role | "Bob is the lead engineer" | Bob → has_role → lead engineer |
| Family | "Alice's daughter Riley" | Alice → parent_of → Riley |
| Marriage | "Dan is Carol's husband" | Dan → married_to → Carol |
| Tool usage | "We use PostgreSQL" | team → uses → PostgreSQL |
| Decisions | "We switched to GraphQL" | team → uses → GraphQL |
| Creation | "Alice created the auth module" | Alice → created → auth module |
| Interests | "Max loves chess" | Max → loves → chess |

### Integration points

- **CLI**: `mempalace extract-kg [--wing X] [--room Y] [--dry-run]`
- **MCP tool**: `mempalace_kg_extract` — AI agents can trigger extraction
- **Idempotent**: existing triples are detected and skipped (safe to re-run)
- **Source provenance**: each triple records which drawer file it came from
- **Batched reads**: handles large palaces without OOM (500 drawers/batch)

### Why this matters

This connects the two storage systems: **drawers** (verbatim text in ChromaDB) and the **knowledge graph** (structured facts in SQLite). Before this PR, users had thousands of memories but an empty KG. After: one command populates it.

```bash
mempalace mine ~/projects/myapp        # stores verbatim memories
mempalace extract-kg                    # auto-populates the KG
mempalace extract-kg --dry-run          # preview what would be extracted
```

## Test plan

- [x] `pytest tests/test_kg_extractor.py -v` — 29 tests pass
- [x] `pytest tests/ -v` — full suite 563 passed, 0 failed
- [x] No new dependencies
- [x] No API keys or network access needed
- [x] Dry-run mode tested (no KG writes)
- [x] Idempotency tested (running twice doesn't duplicate)
- [x] Wing/room filtering tested

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -264,6 +264,58 @@ def cmd_mcp(args):
         print(f"  {base_server_cmd} --palace /path/to/palace")
 
 
+def cmd_extract_kg(args):
+    """Auto-populate the knowledge graph from palace drawers."""
+    from .kg_extractor import extract_kg
+    from .knowledge_graph import KnowledgeGraph
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+    kg_path = os.path.join(palace_path, "knowledge_graph.sqlite3")
+    kg = KnowledgeGraph(db_path=kg_path)
+
+    mode = "DRY RUN" if args.dry_run else "EXTRACT"
+    print(f"\n{'=' * 55}")
+    print(f"  Knowledge Graph Auto-Extract ({mode})")
+    print(f"  Palace: {palace_path}")
+    if args.wing:
+        print(f"  Wing filter: {args.wing}")
+    if args.room:
+        print(f"  Room filter: {args.room}")
+    print(f"{'=' * 55}\n")
+
+    result = extract_kg(
+        palace_path=palace_path,
+        kg=kg,
+        wing=args.wing,
+        room=args.room,
+        dry_run=args.dry_run,
+    )
+
+    print(f"  Drawers scanned:  {result.drawers_scanned}")
+    print(f"  Entities found:   {result.entities_found}")
+    print(f"  Patterns matched: {result.patterns_matched}")
+    print(f"  Triples added:    {result.triples_added}")
+    print(f"  Triples skipped:  {result.triples_skipped} (already in KG)")
+
+    if result.errors:
+        print(f"\n  Errors ({len(result.errors)}):")
+        for err in result.errors[:5]:
+            print(f"    - {err}")
+
+    if result.details:
+        print("\n  Sample extractions:")
+        for t in result.details[:10]:
+            print(f"    {t['subject']} → {t['predicate']} → {t['object']}")
+            if t.get("source"):
+                print(f"      from: {t['source']}")
+
+    if not args.dry_run:
+        stats = kg.stats()
+        print(f"\n  KG totals: {stats['entities']} entities, {stats['triples']} triples")
+
+    print(f"\n{'=' * 55}\n")
+
+
 def cmd_compress(args):
     """Compress drawers in a wing using AAAK Dialect."""
     import chromadb
@@ -530,6 +582,18 @@ def main():
         help="Show MCP setup command for connecting MemPalace to your AI client",
     )
 
+    # extract-kg
+    p_extract = sub.add_parser(
+        "extract-kg",
+        help="Auto-populate knowledge graph from palace drawers",
+    )
+    p_extract.add_argument("--wing", default=None, help="Filter by wing")
+    p_extract.add_argument("--room", default=None, help="Filter by room")
+    p_extract.add_argument(
+        "--dry-run", action="store_true",
+        help="Show what would be extracted without writing to KG",
+    )
+
     # status
     sub.add_parser("status", help="Show what's been filed")
 
@@ -563,6 +627,7 @@ def main():
         "search": cmd_search,
         "mcp": cmd_mcp,
         "compress": cmd_compress,
+        "extract-kg": cmd_extract_kg,
         "wake-up": cmd_wakeup,
         "repair": cmd_repair,
         "status": cmd_status,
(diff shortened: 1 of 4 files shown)


Nitrogonza9:
## Why I contributed this

I'm an intensive AI user running 7 large apps daily. The biggest friction I experience is that **AI memory disappears between sessions** — every conversation starts from zero. MemPalace is the best open-source answer to that problem I've found.

But the Knowledge Graph was empty for everyone. You had thousands of memories in drawers but zero structured facts. That's what this PR fixes — and it's part of a larger batch of contributions I submitted today:

- **#429** — Fix palace.py mtime bug + test coverage
- **#432** — AAAK expand before embedding (resolves the TODO)
- **#433** — Fact checker for contradiction detection (Issue #27)
- **#434** — Auto-populate KG from drawers (this PR)
- **#435** — Palace export/import for backup
- **#436** — Palace doctor for health diagnostics

6 PRs, ~3,000 lines, 101 new tests, zero new dependencies.

I believe **AI memory should be free, local, and belong to everyone** — not locked behind subscriptions or cloud APIs. That's why MemPalace matters, and that's why I'm contributing. Happy to iterate on any of these based on your feedback.

— Gonzalo

web3guru888 (commented):
### Review: Auto-populate KG from palace drawers

This is *exactly* the bridge that makes the KG useful out of the box. We built something equivalent in our integration (`knowledge_graph_bridge.py`) that populates 710 entities and 1,014 triples from 208 discovery records across five domains, so I can share what we learned at that scale.

**What works well:**
- The 8 pattern types cover the right surface area for personal/team knowledge. Employment, roles, family, tools, creation — these are what real palaces contain.
- Idempotent extraction with dedup is crucial. We re-run extraction after every discovery cycle and early on we had triple explosion before adding dedup. Your `_dedupe_triples()` on `(subject.lower(), predicate, object.lower())` handles this cleanly.
- Batched reads at 500/batch — sensible for large palaces. We hit OOM at ~2,000 drawers without batching.
- Source provenance (`source_file`) is excellent. When a contradiction is flagged, you need to trace *which drawer* made the claim.
- The `--dry-run` flag is a great UX choice.

**Observations from running at scale:**

1. **Dedup threshold**: Your dedup is exact-match after lowercasing. At 710 entities, we found we also needed *fuzzy* dedup — "PostgreSQL", "Postgres", "postgres db" all refer to the same entity. We use a tiered approach: hard dedup at cosine 0.86 (definitely same entity), soft dedup at 0.55 (flag for review). Something to consider as palaces grow, though exact-match is a reasonable v1.

2. **`kg.stats()` per-triple for dedup detection**: In the write loop, you call `kg.stats()` before and after each `add_triple()` to detect whether a triple was actually new. With hundreds of triples, that's 2N SQLite queries just for dedup bookkeeping. Consider checking existence *before* writing instead — e.g., a `kg.triple_exists(s, p, o)` method would be much cheaper. Or if `add_triple` could return a boolean indicating whether it was new, that would eliminate the stats calls entirely.

3. **Entity type inference**: Your `add_triple` writes triples but I don't see where extracted entities get typed (person vs company vs tool). The KG's `add_entity(name, entity_type=...)` supports types. You could infer: employment pattern → subject=person, object=company; tool pattern → object=tool. This makes KG queries by entity type much more useful downstream.

4. **Cross-drawer entity resolution**: "Alice" in drawer A and "Alice Chen" in drawer B — are they the same entity? At 710 entities we hit this regularly. Not a blocker for v1, but worth a TODO comment.

**Minor:**
- The `_CREATION_PATTERNS` regex uses a greedy `.+?` for the object, which is good, but terminated by `[.,;]|$`. If the sentence ends without punctuation (common in drawer notes), the match extends to end-of-string and `_clean_object` truncates at 60 chars. Works, but worth a test case for long unpunctuated lines.

29 tests, clean lint, no new deps — well structured. This and #433 compose naturally: extract KG with this PR, then fact-check new claims against it.

Nitrogonza9:
Great review @web3guru888 — real production experience shows. Let me address each point:

**`kg.stats()` per-triple — fixing now.** You're right, 2N SQLite queries for dedup bookkeeping is wasteful. I'll switch to checking triple count before/after the full batch rather than per-triple, or better yet, compare the triple_id against a pre-fetched set. Pushing the fix shortly.

**Entity type inference — implementing.** Good catch. The patterns already know the type: employment → (person, company), tool → (team, tool), family → (person, person), role → (person, role). I'll pass `entity_type` to `add_entity()` during extraction. Low effort, high value.

**Fuzzy entity dedup ("PostgreSQL" vs "Postgres" vs "postgres db")** — agreed this is needed at scale. For v1 I'll add a TODO comment with your tiered threshold approach (0.86 hard / 0.55 soft) as the future direction. Implementing it properly needs embedding similarity which touches the "no new deps" constraint, so it's a separate PR.

**Cross-drawer entity resolution ("Alice" vs "Alice Chen")** — same story, noted as TODO. The entity_registry.py already has some disambiguation infrastructure that could be wired in.

**Long unpunctuated lines** — adding a test case for this edge.

**Re: composition with #433** — exactly the intent. Extract with this PR, fact-check with #433, the KG becomes both self-populating and self-correcting.

Pushing fixes now.

— Gonzalo

web3guru888:
The before/after triple count check is a clean approach — avoids the per-triple query overhead while still giving you dedup signal. One small thing: make sure the count check is inside a transaction with the batch add, otherwise a concurrent write between count() and the upsert can make the comparison unreliable.

The entity type inference from pattern context is a solid v1. And the tiered threshold TODO is exactly the right call — embedding similarity is a natural v2 once the pattern-based extraction is stable.

The #433 composition framing (extract → fact-check → self-correcting KG) is compelling. If both PRs land, `auto_kg` generating the entities and `check_facts` validating them creates a quality loop without any new infrastructure. Worth calling that out in the PR description for reviewers to understand the intent.

PR #435, opened 2026-04-09
feat: palace export/import for backup and migration

Nitrogonza9:
## Summary

There's currently no way to back up, migrate, or share palace data. If ChromaDB or SQLite gets corrupted, everything is lost. This PR adds portable export/import.

- **`mempalace export backup.json`** — exports all drawers + KG entities + KG triples to a single JSON file
- **`mempalace import backup.json`** — restores into a new or existing palace with automatic deduplication
- **MCP tools**: `mempalace_export` and `mempalace_import` for AI agents

### Export format

Self-describing JSON with version field for forward compatibility:
```json
{
  "format": "mempalace_export",
  "version": 1,
  "exported_at": "2026-04-09T...",
  "drawers": [{"id": "...", "document": "...", "metadata": {...}}],
  "kg_entities": [...],
  "kg_triples": [...]
}
```

### Key behaviors

- **Idempotent import**: existing drawers are detected by ID and skipped
- **Batched I/O**: reads/writes in 500-item batches for large palaces
- **Format validation**: rejects unknown formats and future versions
- **Graceful errors**: missing files, empty palaces, and corrupted data handled cleanly

## Test plan

- [x] `pytest tests/test_exporter.py -v` — 13 tests pass
- [x] `pytest tests/ -v` — full suite 547 passed, 0 failed
- [x] No new dependencies
- [x] Full round-trip test: export → import into new palace → verify data matches
- [x] Skip-existing deduplication tested

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -264,6 +264,68 @@ def cmd_mcp(args):
         print(f"  {base_server_cmd} --palace /path/to/palace")
 
 
+def cmd_export(args):
+    """Export palace data to a portable JSON file."""
+    from .exporter import export_palace
+    from .knowledge_graph import KnowledgeGraph
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+    output_file = args.output
+
+    kg_path = os.path.join(palace_path, "knowledge_graph.sqlite3")
+    kg = KnowledgeGraph(db_path=kg_path) if os.path.exists(kg_path) else None
+
+    print(f"\n{'=' * 55}")
+    print("  Palace Export")
+    print(f"  Palace: {palace_path}")
+    print(f"  Output: {output_file}")
+    print(f"{'=' * 55}\n")
+
+    result = export_palace(palace_path=palace_path, output_file=output_file, kg=kg)
+
+    print(f"  Drawers exported:     {result.drawers_exported}")
+    print(f"  KG entities exported: {result.kg_entities_exported}")
+    print(f"  KG triples exported:  {result.kg_triples_exported}")
+
+    if result.errors:
+        print(f"\n  Errors ({len(result.errors)}):")
+        for err in result.errors[:5]:
+            print(f"    - {err}")
+    else:
+        size = os.path.getsize(output_file)
+        print(f"\n  Saved: {output_file} ({size:,} bytes)")
+
+    print(f"\n{'=' * 55}\n")
+
+
+def cmd_import(args):
+    """Import palace data from a JSON export file."""
+    from .exporter import import_palace
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+    input_file = args.input_file
+
+    print(f"\n{'=' * 55}")
+    print("  Palace Import")
+    print(f"  From:   {input_file}")
+    print(f"  Palace: {palace_path}")
+    print(f"{'=' * 55}\n")
+
+    result = import_palace(input_file=input_file, palace_path=palace_path)
+
+    print(f"  Drawers imported:     {result.drawers_imported}")
+    print(f"  Drawers skipped:      {result.drawers_skipped} (already existed)")
+    print(f"  KG entities imported: {result.kg_entities_imported}")
+    print(f"  KG triples imported:  {result.kg_triples_imported}")
+
+    if result.errors:
+        print(f"\n  Errors ({len(result.errors)}):")
+        for err in result.errors[:5]:
+            print(f"    - {err}")
+
+    print(f"\n{'=' * 55}\n")
+
+
 def cmd_compress(args):
     """Compress drawers in a wing using AAAK Dialect."""
     import chromadb
@@ -530,6 +592,14 @@ def main():
         help="Show MCP setup command for connecting MemPalace to your AI client",
     )
 
+    # export
+    p_export = sub.add_parser("export", help="Export palace to a portable JSON file")
+    p_export.add_argument("output", help="Output JSON file path")
+
+    # import
+    p_import = sub.add_parser("import", help="Import palace from a JSON export file")
+    p_import.add_argument("input_file", help="Input JSON file path")
+
     # status
     sub.add_parser("status", help="Show what's been filed")
 
@@ -563,6 +633,8 @@ def main():
         "search": cmd_search,
         "mcp": cmd_mcp,
         "compress": cmd_compress,
+        "export": cmd_export,
+        "import": cmd_import,
         "wake-up": cmd_wakeup,
         "repair": cmd_repair,
         "status": cmd_status,
--- a/mempalace/exporter.py
+++ b/mempalace/exporter.py
@@ -1,0 +1,375 @@
+"""
+exporter.py — Export and import palace data for backup and migration.
+=====================================================================
+
+Exports all palace data (drawers from ChromaDB + knowledge graph from
+SQLite) into a single portable JSON file. Imports from that file to
+restore or migrate a palace to a new machine.
+
+The export format is a self-describing JSON document:
+
+    {
+        "format": "mempalace_export",
+        "version": 1,
+        "exported_at": "2026-04-09T...",
+        "drawers": [...],
+        "kg_entities": [...],
+        "kg_triples": [...]
+    }
+
+Zero API calls. Zero new dependencies. Uses only json + existing modules.
+
+Usage:
+    from mempalace.exporter import export_palace, import_palace
+
+    export_palace(palace_path, output_file="palace_backup.json")
+    import_palace("palace_backup.json", palace_path)
+
+CLI:
+    mempalace export palace_backup.json
+    mempalace import palace_backup.json
+"""
+
+import json
+import logging
+import os
+from datetime import datetime
+from dataclasses import dataclass, field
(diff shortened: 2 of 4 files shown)


bmaltais:
Similar to https://github.com/milla-jovovich/mempalace/pull/453

web3guru888 (commented):
This overlaps with #453 (scokeepa) — both add export/import. Key differences:

| | #435 (this PR) | #453 (scokeepa) |
|---|---|---|
| Format | Single JSON file | JSONL per wing/room |
| KG export | ✅ entities + triples | ❌ drawers only |
| MCP tools | ✅ `mempalace_export/import` | ❌ CLI only |
| Dedup on import | By drawer ID | By drawer ID |
| Git workflow | Less friendly (single large file) | Git-friendly (per-room files) |
| Backup command | ❌ | ✅ `mempalace backup` |
| Tests | 13 tests + round-trip | None visible |

From our integration perspective, **KG export is critical**. We have 710 entities and 1,014 triples — losing those on a backup/restore cycle would be painful. This PR handles it; #453 does not.

A few code-level notes:

1. **Import batch size = 100** — This is sensible. #453 uses 5000 which may cause OOM on constrained systems.

2. **Format versioning** — The `version: 1` field with forward-compat rejection is the right pattern. Future changes can bump the version and add migration logic.

3. **Single JSON file for large palaces** — A 39k-drawer palace (per #427) could produce a massive JSON file. Consider streaming JSONL for the drawers array, or at minimum document the expected file sizes. Our 208-discovery palace would be manageable, but production palaces could hit hundreds of MB.

4. **KG triple import relies on entity names** — `kg.add_triple(triple["subject"], ...)` uses the entity name, which means the entity must already exist. The import creates entities first, then triples — correct ordering. But if an entity name collision occurs (different entity, same name), triples could be mislinked. Entity ID-based import would be safer.

Well-tested with round-trip verification. I'd lean toward this PR for the KG support, possibly incorporating #453's JSONL format for the drawers portion.

Nitrogonza9:
Thanks @web3guru888 — great comparison with #453. The overlap is real, so let me address the tradeoffs honestly.

**Where #453 is better:** JSONL per wing/room is genuinely more git-friendly and the `mempalace backup` command is good UX. For large palaces, streaming writes per-room avoids the single-large-file problem.

**Where this PR is better:** KG export (entities + triples), MCP tools, test coverage, and format versioning. As you noted, losing 710 entities and 1,014 triples on backup/restore would be painful.

**JSONL streaming for large palaces — valid concern.** A 39k-drawer palace would produce a large JSON file. For v1 I'll add a size estimate warning in the CLI output. Streaming JSONL for the drawers array is the right long-term fix, but it changes the format spec — better as a version 2.

**Entity name collisions on import — noted.** You're right that name-based triple import can mislink if two different entities share a name. The KG already normalizes IDs from names (`alice` = `Alice`), so this is only a problem across palaces with genuinely different entities sharing names. I'll add a TODO for ID-based import as a safety improvement.

**Ideal outcome:** Merge the best of both PRs. I'm happy to incorporate #453's JSONL per-room format for drawers while keeping the KG export, MCP tools, and test coverage from this PR. If @scokeepa and the maintainers are open to it, I can do the merge work.

— Gonzalo

web3guru888:
The hybrid approach makes a lot of sense — JSONL per-room drawers for streaming + git-friendliness, KG export, MCP tools, and format versioning from this PR. That would be the best of both.

If @scokeepa and @milla-jovovich are open to a merge, it's worth a direct ping in #453 to propose the consolidation. Having two competing export PRs in review simultaneously is going to be hard for maintainers to untangle, and the community benefits from one well-designed export spec rather than two partial ones.

The ID-based triple import TODO is worth flagging in the code as a  rather than just tracking it mentally — makes it visible to future contributors.

scokeepa:
Hey @Nitrogonza9 — I'm the author of #453. The hybrid approach you proposed sounds great, I'm fully on board.

**What I'd suggest:**
- Use JSONL per wing/room from #453 for drawers (git-friendly, streaming-safe for large palaces)
- Keep KG export (entities + triples), MCP tools (`mempalace_export`/`mempalace_import`), format versioning, and the test suite from this PR
- Keep `mempalace backup` (binary zip) and `mine --backup` from #453 as the fast-restore track

This gives users two complementary workflows:
| Track | Format | Use case |
|-------|--------|----------|
| Daily sync | JSONL export + git | Cross-device, incremental, lightweight |
| Full restore | Binary zip backup | Fast recovery, no re-embedding |

Happy to have you drive the merge work — you clearly have a good grasp of both PRs. I can review and test on my end once you have a combined branch. Let me know if you'd prefer to base it on #435 or #453, either works for me.

cc @milla-jovovich @bmaltais

Nitrogonza9:
@scokeepa @web3guru888 — hybrid branch ready for review:

**Branch**: \`Nitrogonza9:feat/palace-export-import-v2\`

### What's in it

**Two export tracks (auto-detected from output path):**
\`\`\`bash
mempalace export backup.json     # single JSON file (drawers + KG)
mempalace export ./sync/         # JSONL per wing/room (git-friendly)
mempalace export ./sync --format jsonl   # explicit override
\`\`\`

**Binary backup track (from #453):**
\`\`\`bash
mempalace backup                 # directory copy
mempalace backup --zip           # zip archive
mempalace backup --max-backups 3
\`\`\`

**Three MCP tools:** \`mempalace_export\`, \`mempalace_import\`, \`mempalace_backup\` — all auto-detect format.

### Code attribution

- \`backup_palace()\` and the JSONL export/import structure adapted from @scokeepa's #453
- KG export, single-JSON format, format versioning, auto-detection, and the test suite from this PR
- @scokeepa is credited as co-author in the commit

### Tests

26 tests covering both formats, auto-detection, backup, round-trip, and KG preservation. Full suite passing. Ruff clean. Zero new dependencies.

Happy to:
1. Open this as a new PR replacing both #435 and #453
2. Or push it onto either existing branch

Whichever the maintainers prefer. cc @milla-jovovich @bmaltais

— Gonzalo

scokeepa:
@Nitrogonza9 — just reviewed the `feat/palace-export-import-v2` branch. The hybrid design is solid and the auto-detection is a nice touch. A few things I'd like to see before this goes up as a new PR:

### 1. Backup integrity validation (from #448 feedback)

After a zip backup, we should verify the archive isn't corrupted. At minimum:
- Verify SQLite file passes `PRAGMA integrity_check`
- Verify HNSW index file exists in the archive

A corrupted backup that passes silently is worse than no backup. Something like:

```python
def _validate_backup(backup_path: Path, zip_mode: bool) -> list[str]:
    """Quick integrity check after backup."""
    errors = []
    if zip_mode:
        import zipfile
        try:
            with zipfile.ZipFile(backup_path, 'r') as zf:
                bad = zf.testzip()
                if bad:
                    errors.append(f"Corrupt file in archive: {bad}")
                names = zf.namelist()
                if not any('chroma.sqlite3' in n for n in names):
                    errors.append("SQLite file missing from backup")
        except Exception as e:
            errors.append(f"Archive validation failed: {e}")
    return errors
```

### 2. Embedding not included — add user-visible note

The JSONL and JSON exports don't include embedding vectors. On import, ChromaDB re-embeds everything, which is slow for large palaces and produces different vectors if the embedding model changes. Worth adding a note in the CLI output:

```
Note: embeddings are not included. Import will re-embed using the configured model.
```

### 3. `max_backups` in config (from #448 feedback)

Right now `max_backups` is only a function parameter / CLI flag. Users with large palaces may want to set this once in config.json rather than passing `--max-backups` every time. A `backup.max_retained` field in `MempalaceConfig` would be ideal.

### 4. KG triple import — entity name collision TODO

You mentioned this earlier — the `add_triple(triple["subject"], ...)` approach can mislink if two different entities share a name across palaces. Worth adding a `# TODO: ID-based triple import for cross-palace safety` in the code so future contributors can find it.

Happy to help test once these are addressed. Looking forward to the consolidated PR.

web3guru888:
@Nitrogonza9 — the consolidated design looks right to me. The auto-detection from output path is clean; users shouldn't have to think about format flags unless they want to override.

A few observations from our side:

**Export format choice:** The JSONL-per-wing approach is compelling for large palaces (ours has 710 KG entities across 5 wings). Diff-friendly exports matter a lot when you're syncing state across environments or doing incremental backups alongside git-tracked content. Good call making it the path-based default.

**KG round-trip correctness:** @scokeepa's point about entity name collisions on import is the right call to flag. We ran into a similar edge case when importing across wings — subject/object names that are unique within one palace aren't guaranteed globally. A `# TODO` marker is fine for now; the current use case (single-palace backup/restore) is the common path.

**Backup validation:** Strongly agree with the post-backup integrity check. A corrupt zip that reports success is a genuinely bad failure mode. The `zipfile.testzip()` approach is the right lightweight check — we use a similar pattern before trusting any state snapshots.

**Missing embedding note:** Good suggestion from @scokeepa. Worth surfacing this in the CLI output, especially if the import is going into a new palace with a different default model configured. Re-embedding is usually fine but users need to know it's happening.

On the PR structure: if you're opening a new consolidated PR, I'd vote for that over trying to merge into either existing branch — the combined design is clearly better than either standalone approach. Would make reviewing cleaner too.

Happy to test the round-trip when the new PR is up.

bensig:

PR #438, opened 2026-04-09
fix: escape regex metacharacters in extract_people and sanitize session_id in precompact hook

adiKhan12:
## What does this PR do?

Fixes two input sanitization issues:

1. **Regex injection in `extract_people()`** (`split_mega_files.py:144`) — Names from user config (`~/.mempalace/known_names.json`) are interpolated into a regex pattern without `re.escape()`. Names containing regex metacharacters (e.g. `Dr. Smith`, `C++`, `Mary (Smith)`) cause false matches or crashes. The rest of the codebase already uses `re.escape()` for this — `entity_detector.py:471` and `entity_registry.py:602` — this was the one place that missed it.

2. **Missing session_id sanitization in precompact hook** (`mempal_precompact_hook.sh`) — The save hook sanitizes `session_id` via a `safe()` lambda that strips non-alphanumeric characters. The precompact hook was missing this sanitization entirely, using raw JSON input. This is part of the hooks hardening referenced in the README (#110).

## How to test

```bash
python -m pytest tests/test_split_mega_files.py tests/test_hooks_cli.py -v
```

Key new tests:
- `test_extract_people_escapes_regex_metacharacters` — verifies `Dr. Smith` matches exactly (not `Dr_ Smith`)
- `test_extract_people_no_crash_on_regex_metacharacters` — verifies names like `C++`, `[Admin]`, `A*B` don't crash
- `test_precompact_sanitizes_session_id` — verifies path traversal in session_id is stripped

## Checklist
- [x] No hardcoded paths


web3guru888 (commented):
Clean security fix. The `re.escape()` on `KNOWN_PEOPLE` names is necessary — without it, names like "Dr. Smith" match unintended patterns (`Dr_Smith`, `DrXSmith`, etc.) and names with regex metacharacters (`C++`, brackets) crash outright. Good catch.

The session_id sanitization in the precompact hook mirrors the existing save hook pattern, which is the right approach — both hooks now apply the same `[^a-zA-Z0-9_\-]` filter. The test for path traversal (`../../etc/passwd`) covers the critical case.

The test that verifies no crash on regex metacharacters (`test_extract_people_no_crash_on_regex_metacharacters`) is a good regression guard. 👍

adiKhan12:
Thanks for the review @web3guru888! Glad the approach looks solid.

The inconsistency between `entity_detector.py`/`entity_registry.py` (which already used `re.escape()`) and `split_mega_files.py` (which didn't) is what initially flagged this — same pattern, one missed spot.

For the precompact hook, I matched the save hook's `safe()` lambda exactly so both hooks stay in sync going forward. Happy to adjust anything if maintainers have feedback.

web3guru888:
Makes sense — keeping both hooks in sync is the right call. The consistency across save/precompact hooks will make future changes easier to spot if the pattern ever needs updating.

igorls:

Issue #231, opened 2026-04-08
[Feature] Add Multilingual Support 


EndeavorYen:
**What problem does this solve?**

MemPalace currently only works with English. Room classification uses English keyword lists, memory extraction uses English regex patterns, entity detection uses English-only `\b[A-Z][a-z]+\b`, and spellcheck corrupts CJK text. Non-English users (Chinese, French, Japanese, etc.) get incorrect room assignments, no memory type detection, and broken spellcheck.

The MemChinesePalace fork (https://github.com/Chandler-Sun/MemChinesePalace) shows community demand exists — but it's a full rewrite that drops many features (onboarding, entity registry, palace graph, etc.).

**What's the proposed solution?**

Use embedding-based semantic classification instead of per-language keyword/regex patterns. The multilingual embedding model (`paraphrase-multilingual-MiniLM-L12-v2`, ~120MB, CPU-only) understands 50+ languages natively.

Concrete changes:
- **Room classification**: replace `TOPIC_KEYWORDS` matching with cosine similarity against room description embeddings. Any language works with zero configuration.
- **Memory extraction**: replace per-language regex markers with cosine similarity against memory type description embeddings (decision, preference, milestone, problem, emotional). Each paragraph classified independently to preserve multi-type detection.
- **Entity detection**: add Chinese name patterns (百家姓 surnames, simplified+traditional). This stays rule-based because NER is fundamentally pattern matching, not classification.
- **Embedding model**: configurable via `config.json` / env var, default multilingual. `sentence-transformers` is an optional dependency (`pip install mempalace[multilingual]`). Core functionality works without it (English regex fallback).
- **Spellcheck**: auto-skips non-English text via Unicode detection.
- **All 7 ChromaDB consumer modules** (searcher, miner, convo_miner, mcp_server, layers, palace_graph, cli) use a centralized `get_embedding_function()` from config.py.

Benchmark results — 173 test cases across 8 languages (zh-Hans, zh-Hant, en, fr, es, de, ja, ko):
```
Base: 122/122 (100%)    Extended: 51/51 (100%)    219 unit tests, 0 regressions
```

One new optional dependency: `sentence-transformers>=2.0` (install via `pip install mempalace[multilingual]`). Without it, the system falls back to English regex — existing behavior unchanged. No external API calls. Everything runs locally. No GPU needed.

**Alternatives considered:**

[...]

Issue #92, opened 2026-04-07
Multilingual support: French (and 100+ languages) via Ollama BGE-M3


Motokiyo:
## Problem

MemPalace v3.0.0 uses ChromaDB's default embedding model (all-MiniLM-L6-v2), which is primarily trained on English text. When used with French content (or other non-English languages), search results are irrelevant — negative similarity scores, wrong documents returned.

## Solution

We created a patch that replaces the default embedding with **BGE-M3** via Ollama, supporting 100+ languages while keeping everything local (no cloud dependency).

**Changes:**
- New `ollama_embedding.py` — ChromaDB-compatible `EmbeddingFunction` using Ollama's `/api/embed` endpoint
- Patched `miner.py` and `searcher.py` to use BGE-M3 when Ollama is available, with graceful fallback to MiniLM-L6

**Results on our French corpus (1500 files):**

| Query: "détection de chute reachy care" | Before (MiniLM) | After (BGE-M3) |
|---|---|---|
| Top result | `store.js` (JS in-app purchases — zero relevance) | `PROJET_ARISTOTE_VISION.md` (Reachy Care stack — perfect match) |
| Score | -0.428 (negative = no match) | +0.101 (positive = relevant) |

## Patch

Available at: https://github.com/Motokiyo/mempalace-multilingual

One-line install:
```bash
git clone https://github.com/Motokiyo/mempalace-multilingual && cd mempalace-multilingual && ./install.sh
```

## Suggestion for upstream

Consider making the embedding function configurable in `mempalace.yaml`:

```yaml
embedding:
  provider: ollama  # or "default" for MiniLM
  model: bge-m3
  base_url: http://localhost:11434
```

This would let users choose their embedding model without patching source files. Happy to submit a PR if you're interested.

— Alexandre Ferran / EIFFEL AI

NickShtefan:
Hi @Motokiyo — your French numbers (-0.428 → +0.101) are exactly the same shape as the Russian numbers we saw (0.19–0.40 → 0.70–0.77). Same root cause, and the configurable-embedding-function approach from PR #442 is a superset of what your Ollama-BGE-M3 patch does — it lets you point at any sentence-transformers model (including Ollama via the `ollama:<model>` provider that #488 adds on top), with the model stamped into ChromaDB collection metadata so MCP and CLI can't desync.

[...]

Issue #390, opened 2026-04-09
Chunk size (800) exceeds embedding model token limit (256 tokens / ~512 chars)


sidonsoft:
### Problem

The default `CHUNK_SIZE = 800` in `miner.py` exceeds the token limit of the default embedding model.

**Details:**
- **Chunk size:** 800 characters (line 56 in `miner.py`)
- **Embedding model:** `all-MiniLM-L6-v2` (ChromaDB default via ONNX)
- **Model token limit:** 256 tokens (~512 characters)
- **Result:** Content beyond ~512 chars is silently truncated before embedding

### Impact

1. **Lost context:** Important information in positions 512-800 of each chunk is not included in the embedding
2. **Reduced recall:** Semantic search may miss relevant content in truncated portions
3. **Silent degradation:** No errors or warnings — users don't know this is happening

### Suggested Fixes

**Option 1: Reduce default chunk size**
```python
# miner.py line 56-57
CHUNK_SIZE = 400  # Was 800 - safer for 256 token limit
CHUNK_OVERLAP = 50  # Was 100
```

**Option 2: Add CLI configuration**
```bash
mempalace mine <path> --chunk-size 400 --chunk-overlap 50
```

**Option 3: Auto-detect model limits**
Query the embedding function for its max tokens and adjust chunking accordingly.

### Environment

- mempalace version: 3.0.0
- ChromaDB version: 1.5.7
- Embedding: Default (all-MiniLM-L6-v2 via ONNX)

### Priority

Medium — search works but quality could be significantly improved.

web3guru888:
This is a real issue — the 800-char default silently truncates embeddings for the most common local embedding model. The MiniLM models have a 256-token limit, which is roughly 512 characters for typical English prose and much less for dense technical content (code, formulas, citations).

A few data points from our experience running MemPalace on research content (astrophysics, cryptography, economics papers):

**The truncation is asymmetric** — it silently affects dense content (papers, code) more than casual notes. A research note with a formula in the second half gets the formula silently dropped from its embedding. You get no error, just subtly worse retrieval on the content you care most about.

**What actually helps:**
- `CHUNK_SIZE = 400` is a safer default for all-MiniLM-L6-v2 — leaves headroom for Unicode and tokenization overhead
- For `paraphrase-multilingual-mpnet-base-v2` (128-token limit), you'd want to go even lower (~300 chars)
- The right fix is a model-aware chunk size computed from the tokenizer, but that adds complexity

[...]

PR #442, opened 2026-04-09
feat: add configurable multilingual embedding model support

### LINKED ISSUE #903: bug: embedding model mismatch — MCP server uses MiniLM (384-dim) while ingest can use mpnet (768-dim)
## Problem

The MCP server relies on ChromaDB's built-in default embedding function (`all-MiniLM-L6-v2`, 384 dimensions). There is no centralized embedding model configuration — each code path makes its own choice.

If a user ingests their palace with `all-mpnet-base-v2` (768-dim) — which is what the ingest pipeline documentation and several internal paths assume — then every MCP query silently fails or returns garbage results because the query embeddings (384-dim) don't match the stored embeddings (768-dim).

This also affects anyone following issue #515 (GPU-accelerated embeddings via sentence-transformers) or #756 (OpenAI embeddings) — any non-default model used at ingest time will break MCP search.

## Expected behavior

A single source of truth for which embedding model the palace was built with. The MCP server, CLI search, and all ingest paths should resolve to the same model.

## Proposed fix

Centralized embedding config with 3-tier resolution:

1. **Palace-level** — `{palace_path}/palace_meta.json` stores the model used at build time
2. **Global config** — `~/.mempalace/config.json` for user preference
3. **Default** — `all-mpnet-base-v2` (768-dim)

`mcp_server.py` reads from this config instead of relying on ChromaDB's default. Ingest pipelines stamp the model into `palace_meta.json` at build time so the palace is self-describing.

This also resolves #860 (poor cross-room search with MiniLM) since mpnet produces significantly better embeddings for this use case — benchmarks show +3.5pp on LoCoMo R@10 with `bge-large-en-v1.5` over MiniLM, and mpnet sits between the two.

## Related issues

- #860 — Default MiniLM produces poor cross-room search results
- #515 — GPU-accelerated embeddings via sentence-transformers
- #756 — OpenAI 3072-dim embedding support
- #712 — Default model is English-only
- #390 — Chunk size exceeds embedding model token limit

NickShtefan:
## What this PR does

`get_embedding_function()` resolves three built-in model ids — `minilm`, `embeddinggemma`, and `openai-compat` (#1559). **Any other `embedding_model` value silently falls back to MiniLM**, so a palace whose config says `intfloat/multilingual-e5-base` is embedded with an English-only model and nothing says otherwise.

This PR routes those values to chromadb's `SentenceTransformerEmbeddingFunction`, making an arbitrary HuggingFace repo id or a local checkpoint path a first-class local embedder, and replaces the silent fallback with actionable errors.

Why this is still wanted after `openai-compat` landed: that backend needs an endpoint to be running. This path is for people who want `intfloat/multilingual-e5-base` or `BAAI/bge-m3` loaded in-process on torch (MPS / CUDA / CPU), with no server to operate.

## Changes

**`embedding.py`** — any non-built-in model id builds a `SentenceTransformerEmbeddingFunction` behind a lazy loader. The lazy contract matches `EmbeddinggemmaONNX`, so collection-open paths that never embed (MCP status, hooks, wake-up) stay inside their startup budgets. `embedding_device` maps into torch's vocabulary: `coreml`→`mps` (guarded on `arm64`, since `mps.is_available()` returns True on Intel Macs where MPS does not work), `dml`→CPU with a one-shot warning, `auto`→cuda ▸ mps ▸ cpu.

**Hard failures instead of a silent fallback** — `EmbeddingDependencyError` (carrying the `pip install 'mempalace[multilingual]'` hint) when sentence-transformers is missing, `EmbeddingModelError` when the checkpoint itself will not load. Both are surfaced through `ChromaBackend._resolve_embedding_function`, both `mcp_server._get_collection` branches, `tool_mine`, `searcher._open_search_collection` and the wake-up layers — instead of the misleading "Backend open failed" / "No palace found" that buries the install hint on the MCP path. The dependency probe is `importlib.util.find_spec`, not an import, so nothing on a startup path pays the torch import cost just to learn the package exists.

**`mempalace init --model <name>`** — persists the same `config.json` key onboarding writes, and probes dependencies up front, so a missing install fails `init` with exit 2 and a hint rather than surfacing mid-mine. The palace records its identity in the RFC 001 sidecar at first collection creation, as it does for the built-ins.

**`[multilingual]` extra** — was an empty back-compat alias; now installs `sentence-transformers>=3.0`. It still never removes anything from core.

**README** — an *Embedding models* section: model table, the install gate, and the bge-m3 cosine-range calibration note (strong matches cluster ≥0.50, vs MiniLM's ≥0.65).

**Tests** — `tests/test_sentence_transformer_ef.py` (41 new tests): dispatch, lazy load, device mapping, cache keying per model+device, error propagation through the chroma resolver, and MCP error surfacing. Fully offline — the suite passes on a stock `[dev]` env with neither torch nor sentence-transformers installed.

## Not in scope

No new metadata format, no migration command, no mismatch-detection mechanism. Everything this PR used to carry in that area has since landed on `develop` in better-integrated form — the RFC 001 embedder-identity sidecar and `palace set-embedder` (#1724 / #1730), `repair rebuild-index` (#1308 / #1670), configurable chunking (#1024), the multilingual default via `embeddinggemma` (#1483), and the MCP `_get_collection` EF fix (#1303). This revision drops all of it and keeps only the gap that is still open.

## Testing

```bash
pip install -e ".[dev]"                                    # suite needs no extras
python -m pytest tests/ -v --ignore=tests/benchmarks

# functional, with the extra
pip install -e ".[dev,multilingual]"
mempalace init ~/some-content/ --model intfloat/multilingual-e5-base
mempalace mine ~/some-content/
mempalace search "запрос на вашем языке"
```

## Checklist

- [x] Rebased on current `develop`, merges cleanly
- [x] `ruff check .` — All checks passed
- [x] Full suite (CI's command, incl. `--cov-fail-under=80`): **4508 passed, 31 skipped**, coverage 82.48%
- [x] Backward compatible — built-in ids and existing palaces take exactly the path they took before
- [x] End-to-end on Apple Silicon: `init --model` + mine on `mps`, sidecar identity recorded for drawers and closets, search verified, missing dependency fails init with exit 2

Addresses #1563, #1261, #1663, #712, #231, #92, #515, #756. Refs #903, #1949.

## Contributors

- @NickShtefan — sentence-transformers routing, error propagation, `init --model`, README, tests
- @FabioLissi — device support and the original MCP error-propagation prototype in earlier revisions of this PR
- @davidglidden — production validation on a multilingual literary corpus (bge-m3), and the finding that the silent fallback was the load-bearing bug

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@ venv/
 
 # ChromaDB local data
 *.sqlite3-journal
+.venv-sandbox/
+MagicMock/
--- a/README.md
+++ b/README.md
@@ -159,6 +159,31 @@ Two Claude Code hooks save periodically and before context compression:
 
 No API key is required for the core benchmark path.
 
+## Embedding models
+
+The embedding model is bound to the palace at `init` time and stamped into ChromaDB collection metadata, so subsequent `mine` and `search` runs always use the same model that ingest used. To switch models you re-mine: `mempalace re-mine --model <name>`.
+
+Pick the model that matches your content shape:
+
+| Model | When to pick | Disk | Context window |
+|---|---|---|---|
+| `chromadb-default` (`all-MiniLM-L6-v2`) | English notes, short-form content, fastest mining | ~90 MB | 256 tokens |
+| `intfloat/multilingual-e5-base` | Multilingual notes / mixed-language workspaces | ~1.1 GB | 512 tokens |
+| `BAAI/bge-m3` | Long-form literary, philosophical, or scholarly text where 512-token chunks would break a passage mid-thought; multilingual including Greek, Latin, French | ~2.3 GB | 8192 tokens |
+
+```bash
+# default (English-leaning, fastest)
+mempalace init ~/projects/myapp
+
+# multilingual notes
+mempalace init ~/projects/myapp --model intfloat/multilingual-e5-base
+
+# long-form / classical / multilingual literary content
+mempalace init ~/library --model BAAI/bge-m3
+```
+
+The default chunk size is set per-model so chunks fit the model's token window without silent truncation. Larger context models trade longer mining time for better semantic fidelity on long passages.
+
 ## Docs
 
 - Getting started → [mempalaceofficial.com/guide/getting-started](https://mempalaceofficial.com/guide/getting-started.html)
--- a/README.md
+++ b/README.md
@@ -179,6 +179,49 @@ Select with `--backend <name>`, `MEMPALACE_BACKEND=<name>`, or
 [Storage backends](/guide/configuration#storage-backends) for connection
 variables, namespace behavior, and deployment notes.
 
+## Embedding models
+
+Every drawer is embedded with the model configured as `embedding_model` in
+`~/.mempalace/config.json` (or `MEMPALACE_EMBEDDING_MODEL`). Onboarding asks
+once on first run; `mempalace init <dir> --model <name>` sets it
+non-interactively. Each palace records the model it was built with and
+refuses reads under a silently-switched model — searching across a model
+swap silently degrades recall, so the mismatch is a hard error with recovery
+instructions.
+
+Pick the model that matches your content shape:
+
+| Model | When to pick | Disk | Context window |
+|---|---|---|---|
+| `embeddinggemma` _(default for new installs)_ | Multilingual notes, 100+ languages, no extra install | ~300 MB | 2048 tokens |
+| `minilm` (`all-MiniLM-L6-v2`) | English-only content, smallest download, fastest mining | ~30 MB | 256 tokens |
+| `intfloat/multilingual-e5-base` | Multilingual corpora where retrieval quality matters most — avg relevance 0.19 → 0.72 vs MiniLM on a 247-post Russian corpus | ~1.1 GB | 512 tokens |
+| `BAAI/bge-m3` | Long-form literary / scholarly text where 512-token chunks would break a passage mid-thought; strong coverage incl. Greek, Latin, French | ~2.3 GB | 8192 tokens |
+
+The last two rows stand in for *any*
+[sentence-transformers](https://www.sbert.net/) model id (HuggingFace repo
+or local checkpoint path). This path needs an extra install — **without it,
+`init` and `mine` fail with an install hint** rather than silently falling
+back to the default model:
+
+```bash
+pip install 'mempalace[multilingual]'   # sentence-transformers runtime
+
+mempalace init ~/library --model BAAI/bge-m3
+```
+
+Two calibration notes:
+
+- **Score ranges differ per model.** Cosine similarities for a clearly
+  relevant match cluster around ≥0.65 on MiniLM but around ≥0.50 on
+  `bge-m3` — a 0.60 from bge-m3 is a strong semantic match, not a weak
+  one. Calibrate any relevance thresholds against the palace's own model.
+- **Switching models re-embeds.** Set the new model (edit
+  `embedding_model` in `config.json` or use `MEMPALACE_EMBEDDING_MODEL`),
+  then run `mempalace repair rebuild-index` to rewrite the vectors. If you
+  know the existing vectors are compatible, `mempalace palace set-embedder
+  --model <name> --force` records the new identity without re-embedding.
+
 ## Quickstart
 
 ```bash
@@ -301,7 +344,7 @@ verbatim drawer per user/assistant message, idempotent and resume-safe.
 
 - Python 3.9+
 - A vector-store backend (ChromaDB by default)
-- ~300 MB disk for the embedding model. Onboarding (`python -m mempalace.onboarding`) offers `embeddinggemma-300m` (multilingual, 100+ languages, recommended) or `all-MiniLM-L6-v2` (English-only, ~30 MB). See the docstring at [`mempalace/embedding.py`](mempalace/embedding.py) for details and migration notes.
+- ~300 MB disk for the embedding model. Onboarding (`python -m mempalace.onboarding`) offers `embeddinggemma-300m` (multilingual, 100+ languages, recommended) or `all-MiniLM-L6-v2` (English-only, ~30 MB); any sentence-transformers model is also supported via `pip install 'mempalace[multilingual]'` — see [Embedding models](#embedding-models). Details and migration notes in the docstring at [`mempalace/embedding.py`](mempalace/embedding.py).
 - Optional — compute embeddings on a server instead of locally. Set `embedding_model: "openai-compat"` in `~/.mempalace/config.json` together with `embedding_api_url` / `embedding_api_model` (and `embedding_api_key` if the server needs auth) to use any OpenAI-compatible `/v1/embeddings` endpoint — LM Studio, llama.cpp, vLLM, Ollama's OpenAI shim, or a self-hosted server (e.g. a larger multilingual or GPU-served embedder). Each key is overridable via the matching `MEMPALACE_EMBEDDING_API_*` env var. When the endpoint is on your machine or LAN, no content leaves your network. Switching to it requires `mempalace repair rebuild-index` (different vector space).
 
 No API key is required for the core benchmark path.
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -404,10 +404,21 @@ class ChromaCollection(BaseCollection):
         Used by the searcher to detect legacy palaces that were created
         without ``hnsw:space=cosine`` and therefore silently use L2
         distance, which breaks cosine-based similarity interpretation.
+        Also used by ``palace.get_collection`` to read the
+        ``embedding_model`` stamp written at init time (PR #442).
         Returns ``{}`` when metadata is absent so callers can do a plain
         ``.get("hnsw:space")`` without None-checks.
         """
         return self._collection.metadata or {}
+
+    def modify(self, **kwargs):
+        """Pass-through to ``Collection.modify`` for metadata updates.
+
+        ``palace.get_collection`` calls this to stamp the embedding model
+        into legacy palaces and to retrofit an embedding_model field when
+        re-mining with --force.
+        """
+        self._collection.modify(**kwargs)
 
 
 # ---------------------------------------------------------------------------
@@ -569,7 +580,9 @@ class ChromaBackend(BaseBackend):
         * Legacy: ``get_collection(palace_path, collection_name, create=False)``
           — still used by callers not yet migrated.
         """
-        palace_ref, collection_name, create, options = _normalize_get_collection_args(args, kwargs)
+        palace_ref, collection_name, create, options, embedding_function, extra_metadata = (
+            _normalize_get_collection_args(args, kwargs)
+        )
 
         palace_path = palace_ref.local_path
         if palace_path is None:
@@ -590,13 +603,22 @@ class ChromaBackend(BaseBackend):
         if options and isinstance(options, dict):
             hnsw_space = options.get("hnsw_space", hnsw_space)
 
-        ef = self._resolve_embedding_function()
+        # Embedding function: explicit override (palace.get_collection passes the
+        # init-time-bound EF for the palace's embedding_model) wins; otherwise
+        # fall back to the env-derived default EF for legacy callers.
+        ef = embedding_function if embedding_function is not None else self._resolve_embedding_function()
         ef_kwargs = {"embedding_function": ef} if ef is not None else {}
 
         if create:
+            create_metadata = {"hnsw:space": hnsw_space, "hnsw:num_threads": 1}
+            if extra_metadata:
+                # palace.get_collection stamps embedding_model / chunk_size /
+                # chunk_overlap here. Don't let user metadata clobber HNSW knobs.
+                for k, v in extra_metadata.items():
+                    create_metadata.setdefault(k, v)
             collection = client.get_or_create_collection(
                 collection_name,
-                metadata={"hnsw:space": hnsw_space, "hnsw:num_threads": 1},
+                metadata=create_metadata,
                 **ef_kwargs,
             )
         else:
@@ -656,8 +678,17 @@ def _normalize_get_collection_args(args, kwargs):
     """Unify legacy positional ``(palace_path, collection_name, create)`` calls
     with the new kwargs-only ``(palace=PalaceRef, collection_name=..., create=...)``.
 
-    Returns ``(PalaceRef, collection_name, create, options)``.
+    Also extracts ``embedding_function=`` and ``metadata=`` overrides used by
+    ``palace.get_collection`` to enforce the init-time embedding model binding
+    introduced in PR #442 (#442 stamps embedding_model into collection metadata
+    and re-applies the matching EF on every open).
+
+    Returns ``(PalaceRef, collection_name, create, options, embedding_function, metadata)``.
     """
+    # Pull #442-era overrides out first; they may appear with either calling style.
+    embedding_function = kwargs.pop("embedding_function", None)
+    extra_metadata = kwargs.pop("metadata", None)
+
     # New-style: palace= kwarg with a PalaceRef (spec path).
     if "palace" in kwargs:
         palace_ref = kwargs.pop("palace")
@@ -670,7 +701,7 @@ def _normalize_get_collection_args(args, kwargs):
             raise TypeError(f"unexpected kwargs: {sorted(kwargs)}")
         if args:
             raise TypeError("positional args not allowed with palace= kwarg")
-        return palace_ref, collection_name, create, options
+        return palace_ref, collection_name, create, options, embedding_function, extra_metadata
 
     # Legacy: first positional is a path string.
     if args:
@@ -691,6 +722,8 @@ def _normalize_get_collection_args(args, kwargs):
             collection_name,
             bool(create),
             None,
+            embedding_function,
+            extra_metadata,
         )
 
     # Legacy kwargs-only (palace_path=..., collection_name=..., create=...)
@@ -705,6 +738,8 @@ def _normalize_get_collection_args(args, kwargs):
             collection_name,
             bool(create),
             None,
+            embedding_function,
+            extra_metadata,
         )
 
     raise TypeError("get_collection requires palace= or a positional palace_path")
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -2499,9 +2499,23 @@ class ChromaBackend(BaseBackend):
         library default and its queries won't match the writer's vectors.
         """
         try:
-            from ..embedding import get_embedding_function
+            from ..embedding import (
+                EmbeddingDependencyError,
+                EmbeddingModelError,
+                get_embedding_function,
+            )
 
             return get_embedding_function()
+        except (EmbeddingDependencyError, EmbeddingModelError):
+            # The configured model cannot be served (sentence-transformers
+            # not installed / model unloadable). Falling back to the chromadb
+            # default here would silently embed with the wrong model while
+            # the identity sidecar records the configured one — the exact
+            # corruption RFC 001 exists to prevent. Both errors carry
+            # recovery hints; let them surface. Deliberately narrower than
+            # `except ImportError`: an unrelated import failure must not get
+            # mislabeled with the [multilingual] install hint.
+            raise
         except Exception:
             logger.exception("Failed to build embedding function; using chromadb default")
             return None
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -33,7 +33,8 @@ import shlex
 import argparse
 from pathlib import Path
 
-from .config import MempalaceConfig
+from .config import MempalaceConfig, read_collection_metadata
+from .palace import get_collection as _palace_get_collection
 from .version import __version__
 
 
@@ -153,10 +154,31 @@ def cmd_init(args):
     detect_rooms_local(project_dir=args.dir, yes=getattr(args, "yes", False))
     cfg.init()
 
-    # Pass 3: protect git repos from accidentally committing per-project files
+    # Pass 3: stamp the embedding model into the palace collection metadata
+    # at init time (PR #442). Once written, every reader that goes through
+    # palace.get_collection() picks up the same model — silent fallback to
+    # all-MiniLM-L6-v2 against multilingual vectors is no longer possible.
+    palace_path = (
+        os.path.expanduser(args.palace) if getattr(args, "palace", None) else cfg.palace_path
+    )
+    model = getattr(args, "model", None) or "chromadb-default"
+    chunk_size = getattr(args, "chunk_size", None)
+    chunk_overlap = getattr(args, "chunk_overlap", None)
+    _palace_get_collection(
+        palace_path,
+        model=model,
+        chunk_size=chunk_size,
+        chunk_overlap=chunk_overlap,
+    )
+    print(f"\n  Palace initialized: {palace_path}")
+    print(f"  Embedding model: {model}")
+    if chunk_size:
+        print(f"  Chunk size: {chunk_size}")
+
+    # Pass 4: protect git repos from accidentally committing per-project files
     _ensure_mempalace_files_gitignored(args.dir)
 
-    # Pass 4: offer to run mine immediately. The directory just had its
+    # Pass 5: offer to run mine immediately. The directory just had its
     # rooms + entities set up, so 99% of users will mine next anyway —
     # asking here removes the "remember to type the next command" friction.
     # `--auto-mine` skips the prompt and mines automatically; `--yes` is
@@ -256,8 +278,6 @@ def _maybe_run_mine_after_init(args, cfg) -> None:
     except Exception as e:
         print(f"\n  ERROR: mine failed: {e}", file=sys.stderr)
         sys.exit(1)
-
-
 def cmd_mine(args):
     palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
     include_ignored = []
@@ -368,8 +388,6 @@ def cmd_split(args):
     from .split_mega_files import main as split_main
     import sys
 
-    # Rebuild argv for split_mega_files argparse
-    # Expand ~ and resolve to absolute path so split_mega_files sees a real path
     argv = ["--source", str(Path(args.dir).expanduser().resolve())]
     if args.output_dir:
         argv += ["--output-dir", args.output_dir]
@@ -402,7 +420,142 @@ def cmd_status(args):
     from .miner import status
 
     palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+
+    col_meta = read_collection_metadata(palace_path)
+    if col_meta:
+        print(f"\n  Palace config (from collection metadata):")
+        print(f"    Embedding model: {col_meta.get('embedding_model', 'unknown')}")
+        if "chunk_size" in col_meta:
+            print(f"    Chunk size: {col_meta['chunk_size']}")
+        if "chunk_overlap" in col_meta:
+            print(f"    Chunk overlap: {col_meta['chunk_overlap']}")
+
     status(palace_path=palace_path)
+
+
+def _extract_source_files(palace_path: str) -> set:
+    """Extract all unique source_file paths from palace metadata."""
+    from .palace import get_collection, iter_all_metadatas
+
+    try:
+        col = get_collection(palace_path, force=True)
+    except Exception:
+        return set()
+
+    sources = set()
+    for meta in iter_all_metadatas(col):
+        sf = meta.get("source_file")
+        if sf:
+            sources.add(sf)
+    return sources
+
+
+def cmd_remine(args):
+    """Re-mine palace with a new or current embedding model."""
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+
+    if not os.path.isdir(palace_path):
+        print(f"\n  No palace found at {palace_path}")
+        return
+
+    # Determine target model
+    col_meta = read_collection_metadata(palace_path)
+    new_model = getattr(args, "model", None)
+    if new_model:
+        target_model = new_model
+    else:
+        target_model = col_meta.get("embedding_model", "chromadb-default")
+
+    print(f"\n{'=' * 55}")
+    print("  MemPalace Re-mine")
+    print(f"{'=' * 55}\n")
+    print(f"  Palace: {palace_path}")
+    print(f"  Target model: {target_model}")
+    if new_model and col_meta.get("embedding_model") and col_meta["embedding_model"] != new_model:
+        print(f"  Previous model: {col_meta['embedding_model']}")
+
+    # Step 1: Extract source files
+    print("\n  Extracting source file paths from existing drawers...")
+    sources = _extract_source_files(palace_path)
+
+    if not sources:
+        print("  No drawers found. Nothing to re-mine.")
+        return
+
+    # Step 2: Partition into existing vs missing
+    existing = {s for s in sources if os.path.isfile(s)}
+    missing = sources - existing
+
+    print(f"  Found {len(sources)} unique source files.")
+    print(f"    Still exist: {len(existing)}")
+    print(f"    Missing:     {len(missing)}")
+
+    if missing:
+        print("\n  Missing files (will be skipped):")
+        for f in sorted(missing)[:20]:
+            print(f"    - {f}")
+        if len(missing) > 20:
+            print(f"    ... and {len(missing) - 20} more")
+
+    if not existing:
+        print("\n  No source files found on disk. Nothing to re-mine.")
+        return
+
+    if args.dry_run:
+        print(f"\n  (dry run — would re-mine {len(existing)} files)")
+        return
+
+    # Step 3: Backup palace before destructive operation
+    import shutil
+    import chromadb
+
+    backup_path = palace_path.rstrip(os.sep) + ".pre-remine-backup"
+    if os.path.exists(backup_path):
+        shutil.rmtree(backup_path)
+    print(f"\n  Backing up to {backup_path}...")
+    shutil.copytree(palace_path, backup_path)
+
+    # Step 4: Drop and re-create with new model in metadata
+    print("  Dropping existing collection...")
+    client = chromadb.PersistentClient(path=palace_path)
+    client.delete_collection("mempalace_drawers")
+
+    chunk_size = getattr(args, "chunk_size", None)
+    chunk_overlap = getattr(args, "chunk_overlap", None)
+    _palace_get_collection(
+        palace_path,
+        model=target_model,
+        chunk_size=chunk_size,
+        chunk_overlap=chunk_overlap,
+    )
+    print(f"  Created collection with model: {target_model}")
+
+    # Step 5: Re-mine with exact file list
+    print("  Re-mining...")
+
+    from .miner import mine
+    from collections import defaultdict
+
+    dir_files = defaultdict(list)
+    for f in sorted(existing):
+        dir_files[os.path.dirname(f)].append(f)
+
+    for source_dir, file_list in sorted(dir_files.items()):
+        if os.path.isdir(source_dir):
+            print(f"\n  Mining: {source_dir} ({len(file_list)} files)")
+            mine(
+                project_dir=source_dir,
+                palace_path=palace_path,
+                source_files=file_list,
+            )
+
+    print(f"\n{'=' * 55}")
+    print(f"  Re-mine complete. Model: {target_model}")
+    if missing:
+        print(f"  Skipped {len(missing)} missing source files.")
+    print(f"  Backup saved at {backup_path}")
+    print(f"{'=' * 55}\n")
 
 
 def cmd_repair(args):
@@ -430,9 +583,12 @@ def cmd_repair(args):
 
     backend = ChromaBackend()
 
-    # Try to read existing drawers
+    # Try to read existing drawers. ``force=True`` lets repair through even
+    # when the stamped embedding_model in metadata can't be loaded (the whole
+    # point of repair is recovering broken palaces) — palace.get_collection
+    # logs the mismatch instead of raising.
     try:
-        col = backend.get_collection(palace_path, "mempalace_drawers")
+        col = _palace_get_collection(palace_path, force=True)
         total = col.count()
         print(f"  Drawers found: {total}")
     except Exception as e:
@@ -481,8 +637,12 @@ def cmd_repair(args):
     shutil.copytree(palace_path, backup_path)
 
     print("  Rebuilding collection...")
+    # Drop the existing collection through the backend (which handles
+    # client cache invalidation) and let palace.get_collection rebuild it
+    # — this re-reads the embedding_model stamp and returns a collection
+    # bound to the same EF, so vectors stay compatible across repair.
     backend.delete_collection(palace_path, "mempalace_drawers")
-    new_col = backend.create_collection(palace_path, "mempalace_drawers")
+    new_col = _palace_get_collection(palace_path)
 
     filed = 0
     for i in range(0, len(all_ids), batch_size):
@@ -540,7 +700,6 @@ def cmd_compress(args):
 
     palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
 
-    # Load dialect (with optional entity config)
     config_path = args.config
     if not config_path:
         for candidate in ["entities.json", os.path.join(palace_path, "entities.json")]:
@@ -554,16 +713,17 @@ def cmd_compress(args):
     else:
         dialect = Dialect()
 
-    # Connect to palace
+    # Connect to palace through palace.get_collection so the EF bound at
+    # init time (PR #442) is reapplied — querying compressed entries with
+    # the wrong EF would return junk neighbours.
     backend = ChromaBackend()
     try:
-        col = backend.get_collection(palace_path, "mempalace_drawers")
+        col = _palace_get_collection(palace_path)
     except Exception:
         print(f"\n  No palace found at {palace_path}")
         print("  Run: mempalace init <dir> then mempalace mine <dir>")
         sys.exit(1)
 
-    # Query drawers in batches to avoid SQLite variable limit (~999)
     where = {"wing": args.wing} if args.wing else None
     _BATCH = 500
     docs, metas, ids = [], [], []
@@ -629,10 +789,12 @@ def cmd_compress(args):
             print(f"    {compressed}")
             print()
 
-    # Store compressed versions (unless dry-run)
     if not args.dry_run:
         try:
-            comp_col = backend.get_or_create_collection(palace_path, "mempalace_compressed")
+            # Sister collection for compressed entries — also needs the
+            # EF bound to the palace's embedding_model so the AAAK summary
+            # vectors live in the same space.
+            comp_col = _palace_get_collection(palace_path, "mempalace_compressed")
             for doc_id, compressed, meta, stats in compressed_entries:
                 comp_meta = dict(meta)
                 comp_meta["compression_ratio"] = round(stats["size_ratio"], 1)
@@ -649,9 +811,7 @@ def cmd_compress(args):
             print(f"  Error storing compressed drawers: {e}")
             sys.exit(1)
 
-    # Summary
     ratio = total_original / max(total_compressed, 1)
-    # Estimate tokens from char count (~3.8 chars/token for English text)
     orig_tokens = max(1, int(total_original / 3.8))
     comp_tokens = max(1, int(total_compressed / 3.8))
     print(f"  Total: {orig_tokens:,}t -> {comp_tokens:,}t ({ratio:.1f}x compression)")
@@ -681,7 +841,7 @@ def main():
     sub = parser.add_subparsers(dest="command")
 
     # init
-    p_init = sub.add_parser("init", help="Detect rooms from your folder structure")
+    p_init = sub.add_parser("init", help="Detect rooms and initialize palace config")
     p_init.add_argument("dir", help="Project directory to set up")
     p_init.add_argument(
         "--yes",
@@ -742,6 +902,18 @@ def main():
             "API key for the provider. For anthropic, defaults to $ANTHROPIC_API_KEY; "
             "for openai-compat, defaults to $OPENAI_API_KEY."
         ),
+    )
+    p_init.add_argument(
+        "--model", default=None,
+        help="Embedding model to bind to this palace (default: chromadb-default)",
+    )
+    p_init.add_argument(
+        "--chunk-size", type=int, default=None,
+        help="Chunk size in characters (default: 450)",
+    )
+    p_init.add_argument(
+        "--chunk-overlap", type=int, default=None,
+        help="Chunk overlap in characters (default: 50)",
     )
 
     # mine
@@ -896,6 +1068,27 @@ def main():
 
     sub.add_parser("status", help="Show what's been filed")
 
+    # re-mine
+    p_remine = sub.add_parser(
+        "re-mine",
+        help="Re-mine palace with a new embedding model",
+    )
+    p_remine.add_argument(
+        "--model", default=None,
+        help="New embedding model (default: keep current palace model)",
+    )
+    p_remine.add_argument(
+        "--chunk-size", type=int, default=None,
+        help="New chunk size in characters",
+    )
+    p_remine.add_argument(
+        "--chunk-overlap", type=int, default=None,
+        help="New chunk overlap in characters",
+    )
+    p_remine.add_argument(
+        "--dry-run", action="store_true", help="Show what would be re-mined without doing it"
+    )
+
     args = parser.parse_args()
 
     if not args.command:
@@ -931,6 +1124,7 @@ def main():
         "repair": cmd_repair,
         "migrate": cmd_migrate,
         "status": cmd_status,
+        "re-mine": cmd_remine,
     }
     dispatch[args.command](args)
 
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -263,7 +263,9 @@ def _maybe_run_mine_after_init(args, cfg) -> None:
             print(f"\n  Skipped. Run `mempalace mine {shlex.quote(project_dir)}` when ready.")
             return
 
-    palace_path = cfg.palace_path
+    palace_path = (
+        os.path.expanduser(args.palace) if getattr(args, "palace", None) else cfg.palace_path
+    )
     try:
         mine(
             project_dir=project_dir,
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -309,6 +309,10 @@ def cmd_init(args):
 
     cfg = MempalaceConfig()
 
+    # Bind the embedding model non-interactively (#442). Runs before any
+    # collection is created so the first mine records this identity.
+    _apply_init_model_choice(args, cfg)
+
     # Resolve entity-detection languages: --lang overrides config.
     lang_arg = getattr(args, "lang", None)
     if lang_arg:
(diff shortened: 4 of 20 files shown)


sidonsoft:
## Real-World Testing Results

We applied this patch to our production MemPalace instance and re-mined ~150 files across 3 projects. Here are the findings:

### 📊 Impact

| Metric | Before (800/100) | After (400/50) | Change |
|--------|------------------|----------------|--------|
| **Total drawers** | 10,000+ | 5,873 | **-41%** |
| **HexClamp project** | 30,653 drawers | 665 drawers | **-98%** |
| **Search relevance** | Poor (truncation artifacts) | Good | ✅ |

### 🔍 Search Quality Improvement

**Before:** Searching for "extension config" returned mypy type cache data (`.data.json` files) with match scores around `-0.36`

**After:** Same query returns actual extension code and documentation with match scores around `0.19` (positive correlation = better semantic match)

### 🐛 Additional Fix Applied

We also excluded mypy cache files which were polluting search results:

```python
# miner.py line 304
) or filename.endswith((".data.json", ".meta.json")):
    continue
```

This reduced our palace size by an additional **~24,000 drawers** of type metadata that have no semantic search value.

### ✅ Recommendation

**Merge this PR.** The smaller chunk size:
1. Fits within the embedding model's 256-token limit (~512 chars)
2. Improves search relevance significantly
3. Reduces storage requirements
4. Prevents silent truncation of important context

The overlap reduction from 100→50 is also appropriate since chunks are now half the size.

---

**Test environment:**
- mempalace 3.0.0
- ChromaDB 1.5.7
- Default embedding: all-MiniLM-L6-v2 (ONNX)
- Projects mined: pi agent extensions, HexClamp CLI, openclaw workspace

NickShtefan:
This PR covers the embedding function centralization part of #231 — configurable model via env var / config.json, all 7 ChromaDB consumer modules patched, optional `[multilingual]` dependency, and chunk size fix for #390.

I tested this on 245 Russian blog posts (5 years of content). With `intfloat/multilingual-e5-base`, relevance scores jumped from 0.19–0.40 to 0.70–0.77. Happy to share the full benchmark data if useful.

If @EndeavorYen's more comprehensive implementation lands first (room classification, entity detection, spellcheck), happy to close this — it can serve as a reference for the embedding centralization pattern. Either way, our real-world Russian-language benchmarks might be useful validation for the final PR.

FabioLissi:
Nice PR! The env → config.json → default chain, module-level caching, and graceful fallback when sentence-transformers isn't installed all feel really clean. The CHUNK_SIZE 800→450 fix is a great bonus too.

One small thing I'd love to see on top of this (happy to do it as a follow-up PR if you'd rather keep #442 focused on multilingual): also expose the embedding device.

Since get_embedding_function() already instantiates SentenceTransformerEmbeddingFunction(model_name=...), and that class accepts a device kwarg, it's a pretty small addition. A sibling embedding_device property on MempalaceConfig reading MEMPALACE_EMBEDDING_DEVICE / config.json, and conditionally passing it through:

kwargs = {"model_name": model_name}
if cfg.embedding_device:
    kwargs["device"] = cfg.embedding_device
_embedding_function = SentenceTransformerEmbeddingFunction(**kwargs)

The reason I'm asking: MemPalace currently disables CoreML in __init__.py (for good reason, it segfaults on M-series), so the default path is pinned to CPU. On a large corpus like /.claude/projects/ (1.5 GB of JSONL for me), that's 30 to 90 minutes of embedding on an otherwise-idle Apple Silicon GPU. With device='mps' (or 'cuda' for NVIDIA folks), that drops dramatically, and since all-MiniLM-L6-v2 produces identical vectors regardless of backend, existing palaces stay compatible.

It should be pretty low-risk:
• Backward compatible (no device set = current behavior exactly)
• Fully opt-in
• Uses the PyTorch path, not the ONNX/CoreML one, so it sidesteps the segfault entirely
• Your existing try/except already handles the "device unavailable" failure mode gracefully

One minor gotcha worth mentioning in docs: SentenceTransformerEmbeddingFunction caches models by name only (not (name, device)), so switching devices mid-process returns a stale cached model. Only matters for tests/benchmarks, but good to know.

Very happy to send a follow-up PR if you'd prefer to keep this one scoped. Either way, thanks for the contribution, this unlocks a lot!

FabioLissi:
Hey, one more thing while testing this branch locally. I think there's a small regression in the default path.

When MEMPALACE_EMBEDDING_MODEL isn't set (the case for every existing user post-upgrade), get_embedding_function() returns None, which then gets passed explicitly to every ChromaDB call site:

col = client.get_collection(collection_name, embedding_function=ef)  # ef is None

In chromadb 0.6.3 (the version I'm using), embedding_function=None is no longer equivalent to omitting the argument. The collection gets created with a null embedder, and then collection.add() raises:

ValueError: You must provide an embedding function to compute embeddings.

This breaks tests/test_convo_miner.py::test_convo_mining on this branch. Repro:

git fetch upstream pull/442/head:pr-442 && git checkout pr-442
pytest tests/test_convo_miner.py -x

Passes on main, fails on pr-442.

The fix is a one-liner in config.py: return DefaultEmbeddingFunction() instead of None for the "no model configured" path, so the 14 call sites you patched can keep passing ef unconditionally:

if not model_name:
    from chromadb.utils.embedding_functions import DefaultEmbeddingFunction
    _embedding_function = DefaultEmbeddingFunction()
    return _embedding_function

Same pattern in the except block for the missing-sentence-transformers case.

Happy to send a small follow-up PR against your branch if easier, or you can patch it directly

NickShtefan:
Thanks @FabioLissi — great catch on the `None` regression, and the device support is a clean addition. Cherry-picked your commit (`5cf62ee`) onto this branch. PR now includes both the multilingual embedding centralization and the DefaultEmbeddingFunction + device fix.

Co-authored-by: Fabio Lissi <fabio.lissi@gmail.com>

FabioLissi:
@NickShtefan Thanks again for pulling in the device support commit. One follow-up thought I wanted to float while it's still fresh.
Now that this PR lets users configure the embedding model, there's an adjacent footgun that's worth considering: if a user switches models on a populated palace, either it silently corrupts search (same-dim change → incompatible vectors in the same collection), or it crashes mid-mine (dim change → ChromaDB rejects the insert). Neither failure mode is obvious, and recovery means rm -rf /.mempalace/palace/ and re-mining everything.
A tiny detection mechanism would close this: store the embedding model name in collection metadata at creation, and on subsequent opens, compare it to the currently configured model. On a mismatch, raise a clear error with recovery instructions.
I'm happy to write it either way, but want to check first: does this fit better as:
 - Part of this PR (natural extension of "this PR lets you configure embedding models"), or
 - A follow-up PR after this one merges (keeps #442 scoped to the config feature), or
 - Not something you want at all (maybe there's a reason the current behavior is intentional that I'm missing)?
No strong preference from my end. Just want to avoid duplicating work or stepping on toes.

web3guru888 (commented):
The configurable embedding model support is welcome — we've been using all-MiniLM-L6-v2 for English-only discovery mining, and being able to swap to e5-base for multilingual domains without forking would be useful.

The implementation is clean: env var → config file → default fallback, with proper caching. The device support (`mps`/`cuda`) is a nice ergonomic touch.

**However, this PR includes an unrelated breaking change that should be split out:**

```python
-CHUNK_SIZE = 800  # chars per drawer
-CHUNK_OVERLAP = 100  # overlap between chunks
+CHUNK_SIZE = 450  # chars per drawer
+CHUNK_OVERLAP = 50  # overlap between chunks
```

Halving the chunk size from 800→450 and the overlap from 100→50 is a significant behavioral change that:
1. Will produce ~2x more drawers for the same corpus (more ChromaDB storage, slower queries)
2. Makes existing palaces inconsistent — old drawers are 800-char, new ones 450-char
3. Has nothing to do with multilingual embedding support

If smaller chunks improve retrieval for multilingual models (plausible — shorter contexts embed more precisely), that's worth doing, but it should be a separate PR with its own benchmarks showing the improvement. Bundling it here means adopters of multilingual support get a chunk size change they didn't ask for.

Also: the `get_embedding_function()` singleton means the embedding model is locked at first call for the entire process lifetime. This is fine for CLI usage, but in long-running MCP servers it means you can't change the model without restarting. Might be worth documenting that constraint.

sidonsoft:
## Real-World Production Findings                                                                                                
                                                                                                                                    
   I opened issue #390 reporting the truncation problem, and this PR directly addresses it. We've already deployed     
 this change locally and have empirical data that contradicts the reviewer's concerns.                                              
                                                                                                                                    
   ### Our Setup                                                                                                                    
   - **Corpus:** ~6 months of AI conversation logs + project documentation                                                          
   - **Original:** CHUNK_SIZE = 800, OVERLAP = 100                                                                                  
   - **Updated:** CHUNK_SIZE = 450, OVERLAP = 50                                                                                    
   - **Embedding:** Default (all-MiniLM-L6-v2)                                                                                      
                                                                                                                                    
   ### Measured Results                                                                                                             
                                                                                                                                    
   | Metric | Before (800) | After (450) | Change |                                                                                 
   |--------|--------------|-------------|--------|                                                                                 
   | Total drawers | 5,991 | 5,247 | **-12.4%** |                                                                                   
   | Avg search relevance (1-5) | 3.2 | 4.1 | **+28%** |                                                                            
   | Truncated embeddings | ~35% of chunks | 0% | ✅ **Fixed** |                                                                    
   | Search latency (p50) | 240ms | 180ms | **-25%** |                                                                              
                                                                                                                                    
   ### Why Drawers *Decreased* (Not Increased)                                                                                      
                                                                                                                                    
   The reviewer's concern about "2x more drawers" didn't materialize because:                                                       
                                                                                                                                    
   1. **No wasted drawers:** With 800-char chunks, ~35% were being truncated at embedding time, creating ineffective drawers that   
 didn't contribute to retrieval                                                                                                     
   2. **Better semantic boundaries:** 450 chars fits complete thoughts better than 800 chars (which often split mid-context)        
   3. **Reduced duplication:** Overlap of 50 is sufficient for context continuity without bloating storage                          
                                                                                                                                    
   ### Chunk Size & Multilingual Embeddings                                                                                         
                                                                                                                                    
   **Important:** Even with the new multilingual embedding support, the default embedding issue still applies:                      
                                                                                                                                    
   - **Default model** (`all-MiniLM-L6-v2`): 256 token limit (~512 chars) → **450-char chunks are required**                        
   - **Multilingual models** (e.g., `e5-base`, `paraphrase-multilingual`): Typically 512 tokens (~1024 chars) → **Could use larger  
 chunks**                                                                                                                           
                                                                                                                                    
   The 450-char default ensures **all users get working embeddings out of the box**, regardless of which model they choose. If      
 someone switches to a multilingual model with a higher token limit, they can adjust the chunk size accordingly.                    
                                                                                                                                    
   **With custom embedding providers, users can set whatever chunk size matches their model's token limit.** But the *default*      
 needs to work safely for everyone, and 450 chars is the safe upper bound for the most common embedding models.                     
                                                                                                                                    
   ### Quality Improvement                                                                                                          
                                                                                                                                    
   The **+28% relevance gain** comes from:                                                                                          
   - Every embedding now contains its **full intended content** (no silent truncation)                                              
   - Search vectors match against **complete semantic units**                                                                       
   - Fewer false negatives from corrupted/partial embeddings                                                                        
                                                                                                                                    
   ### Recommendation                                                                                                               
                                                                                                                                    
   **Merge this PR.** The chunk size change isn't just related to multilingual support — it's a **critical fix for the default      
 embedding model's token limit**. Our production data shows:                                                                        
   - ✅ Storage decreased (not increased)                                                                                           
   - ✅ Search quality improved significantly                                                                                       
   - ✅ Works for both default and custom embedding providers                                                                       
   - ✅ No palace inconsistency issues (change is backward-compatible)                                                              
                                                                                                                                    
   The singleton constraint noted by the reviewer is worth documenting, but shouldn't block this fix. Users with long-running MCP   
 servers can restart to change models — that's a reasonable tradeoff for fixing silent data loss.                                   
                                                                                                                                    
   ---                                                                                                                              
                                                                                                                                    
   **TL;DR:** We reported the problem in #390, this PR fixes it, and our real-world deployment proves it works better than the      
 original. The 450-char default ensures all embedding models (default or multilingual) work without truncation.

NickShtefan:
Good point about the singleton — worth clarifying why it's intentional rather than a limitation.

Switching the embedding model isn't a hot-swap operation. Old vectors were generated in model A's vector space; new queries would use model B's vector space. Even if dimensions match, cosine similarity between incompatible spaces produces meaningless results. If dimensions differ, ChromaDB rejects the insert outright.

So after changing the model, you **must** re-mine the entire palace — which is a bulk operation (minutes to hours depending on corpus size). Given that a restart is already required for the re-mine to take effect, the singleton cache is the correct behavior: it prevents accidentally mixing incompatible vectors within a single process.

What would genuinely help is model mismatch detection (as @FabioLissi suggested): store the model name in collection metadata at creation time, and on subsequent opens, compare it to the current config. On mismatch → clear error with recovery instructions. That's a much better safeguard than making the singleton resettable, which would only make the "silent corruption" failure mode easier to hit.

Happy to add a documentation note about the restart requirement for model changes.

NickShtefan:
@FabioLissi Great suggestion — this is exactly the kind of safeguard that should ship alongside configurable embedding models. Without it, users can silently corrupt their palace by switching models, and the failure mode isn't obvious at all.

I'd love to include this in #442 rather than a follow-up — it makes the PR more thoughtful and robust. If you're up for writing it, go ahead! I'll cherry-pick your commit onto this branch, same as with the device support.

Thanks for flagging this and for the continued contributions!

FabioLissi:
Three commits:                                                                                                                                               
                                                                                                                                                                                     
[  6a23c8f: Chunk size env vars ](https://github.com/FabioLissi/mempalace/commit/6a23c8fc58049b9c120bd8d386bf9c47ab058c2d) 
Made your 800→450 chunk size reduction configurable via MEMPALACE_CHUNK_SIZE and MEMPALACE_CHUNK_OVERLAP env vars. Your PR defaulted to 450 to fit MiniLM's 256-token window, which is the right call for the default model. On our side, we switched to BAAI/bge-base-en-v1.5 (better retrieval quality, ~63 vs ~56 on MTEB) and are running it on MPS for GPU acceleration. bge-base has a 512-token context window, double that of MiniLM, so we set MEMPALACE_CHUNK_SIZE=900 (~225 tokens, safely within the 512 limit). The 450  default was leaving half the model's capacity on the table. Making it configurable means users can size chunks to their model's actual token window, rather than being constrained by MiniLM's limits.                  

[7aa96cc: Embedding model mismatch detection](https://github.com/FabioLissi/mempalace/commit/7aa96ccc5b278448d1b04e7b3bce31dca7b0138c)
If someone switches to a model with the same vector dimension, ChromaDB silently mixes incompatible vectors; search returns results, they're just garbage. The dimension-mismatch case is actually easy because ChromaDB rejects the insert.
The fix stores the model name in collection metadata at creation time and checks on every open. Mismatch raises a hard error with recovery instructions. MEMPALACE_FORCE_EMBEDDING=true bypasses it for power users if you switch to models with compatible vectors. Legacy palaces get silently stamped on upgrade. All seven direct ChromaDB callers now go through a single palace.get_collection() chokepoint where the check lives, rather than some going directly to ChromaDB.
Also fixed a pre-existing bug: your tests used patch.dict, which doesn't clear existing env vars from the host shell, so tests that assume a clean environment break if the machine has any MEMPALACE_EMBEDDING_* variables set. Switched to monkeypatch.delenv().
                                                                                                                                                                                     
[94412e8 - Re-mine command](https://github.com/FabioLissi/mempalace/commit/94412e89c7be2dd11565ec4e5d08ab1eb529b073) The recovery path for intentional model switches. Extracts source files from existing metadata, backs up the palace, drops, and re-mines using the exact file list (not the whole directory, so it won't pull in extra files). Supports --dry-run. One documented limitation: convo-mode files get re-mined as project files, correct vectors, but different chunk boundaries. Fixing that properly requires storing mining mode in metadata; felt like scope creep for now.          
                                                                                              
Cherry-picking: first commit is standalone, detection is one squashed commit, re-mine depends on detection. Happy to answer questions or make any adjustments!

NickShtefan:
@FabioLissi Thanks for the excellent contributions! I've cherry-picked all three commits into this PR:

- `6a23c8f` — CHUNK_SIZE/OVERLAP env vars
- `7aa96cc` — embedding model mismatch detection
- `94412e8` — re-mine command

During integration I found and fixed two issues:

1. **Missing `iter_all_metadatas`** — your mismatch detection and re-mine reference it from `palace.py`, but the function lives in your `f153a48` commit (paginated metadata reads) which wasn't part of the three. Added it to `palace.py`.

2. **`EmbeddingModelMismatchError` swallowed in `searcher.py`** — both `search()` and `search_memories()` catch `Exception` broadly, so mismatch errors were showing "No palace found" instead of the actual error. Added `except EmbeddingModelMismatchError: raise` before the generic handler.

Full test cycle passed: mine → search → model switch (mismatch detected) → force bypass (dimension error, expected) → re-mine → search with new model. Everything works.

Rebased on latest upstream/main, all clean.

FabioLissi:
@NickShtefan — great catch on the `searcher.py` swallow. Running the fix through our MPS setup                                             
  turned up two more spots where `EmbeddingModelMismatchError` gets eaten before it reaches the                                              
  user. Both are on the MCP path, so your PR as-is works for CLI `mempalace search` but every MCP                                            
  tool call still shows the misleading "No palace found" message.                                                                            
                                                                                                                                             
  ### 1. `mcp_server.py:_get_collection()` wrapper                                                                                           
                                                                                                                                             
  ```python                                                                                                                                  
  def _get_collection(create=False):                                                                                                         
      ...                                
      try:
          if create or _collection_cache is None:
              _collection_cache = _palace_get_collection(...)                                                                                
          return _collection_cache                                                                                                           
      except Exception:        # ← swallows the mismatch                                                                                     
          return None                                                                                                                        
 ```                                                                                              
  Every MCP tool that routes through _get_collection() — status, list_wings, list_rooms,                                                     
  get_taxonomy, check_duplicate, add_drawer, search (via cached handle), etc. — sees None     
  and falls through to _no_palace(), which returns the generic "No palace found / run init + mine"                                           
  hint. The real error and recovery instructions never surface.                                                                              
                                                                                                                                             
  2. Top-level MCP dispatcher (handle_request)                                                                                               
 ```python                                                                                                                                             
  try:                                                                                                                                       
      result = TOOLS[tool_name]["handler"](**tool_args)                                       
      return {...}                                                                                                                           
  except Exception:
      logger.exception(f"Tool error in {tool_name}")                                                                                         
      return {                                                                                                                               
          "jsonrpc": "2.0",              
          "id": req_id,                                                                                                                      
          "error": {"code": -32000, "message": "Internal tool error"},                        
      }                                                                                                                                      
 ```                                                                                             
  Even if _get_collection() re-raises, this top-level catch replaces the actionable message with                                             
  hardcoded "Internal tool error", so the MCP client still doesn't see what's wrong.
                                                                                                                                             
  Fix (applied on our fork, tests green)                                                                                                     
                                                                                                                                             
  Three spots:                                                                                                                               
  1. searcher.py — re-raise EmbeddingModelMismatchError in both search() and search_memories()
  before the generic except (matches your fix).                                                                                              
  2. mcp_server.py:_get_collection() — except EmbeddingModelMismatchError: raise before the
  generic swallow.                                                                                                                           
  3. mcp_server.py:handle_request() — catch EmbeddingModelMismatchError explicitly and return                                                
  {"error": str(e)} as a JSON result so MCP clients see the message.                                                                         
                                                                                                                                             
  Commit on our fork: FabioLissi/mempalace@e64cdd8 — happy to open a follow-up PR, or you can                                                
  cherry-pick directly. 

Heads-up on rebasing #442 onto current upstream/main                                                                                       
                                                                                              
  When you rebase, you'll hit a conflict in mcp_server.py at four call sites (tool_status, tool_list_wings, tool_list_rooms,                 
  tool_get_taxonomy). Our commits route those through iter_all_metadatas(col) from f153a48, but PR #371 (silent-exceptions-pagination,
  already merged to main) replaced them with inline pagination that surfaces partial failures with offset info.                              
                                                                                              
  Resolution: take upstream's version at all four sites, strictly better UX (reports "failed at offset N" instead of silent except Exception:
   pass). Then drop the now-unused iter_all_metadatas import from mcp_server.py:
                                                                                                                                             
  was:                                                                                                                                       
  from .palace import iter_all_metadatas, get_collection as _palace_get_collection
  becomes:                                                                                                                                   
  from .palace import get_collection as _palace_get_collection                                
                                         
  iter_all_metadatas stays in palace.py, still used by cli.py (re-mine) and miner.py (mismatch detection), just no longer by mcp_server.py.  
   
  We already did this merge locally on FabioLissi/mempalace:mps-device-support if you want to look at the resolution: 180496c (the merge commit) followed by e64cdd8 (this fix).

PR #488 (multilingual, @EndeavorYen) touches config.py, mcp_server.py, searcher.py,                                                        
  palace.py, miner.py — same files as this fix and our original #442 work. Whichever lands    
  second will need a rebase, but the fix above is small enough that it should slot in cleanly either way.

NickShtefan:
@FabioLissi cherry-picked your MCP fix (e64cdd8) — thanks. PR rebased on latest upstream/main, all 604 tests pass. Ready for review.

igorls:
This should not live in a ENV var, the palace storage should be initialized with a specific embedding model and that model is tagged in the configuration, setting the embedding model should happen only on init or in a migrate command to re process the embeddings

NickShtefan:
@igorls — agreed, and done. Rewrote the PR to eliminate all embedding-related env vars.

## What changed

The embedding model is now **bound to the palace at init time** via ChromaDB collection metadata — not through environment variables. Setting or changing the model is only possible through explicit CLI commands:

```
mempalace init <dir> --model intfloat/multilingual-e5-base
mempalace re-mine --model <new-model>
```

### Removed env vars
- `MEMPALACE_EMBEDDING_MODEL` → replaced by `--model` flag on `init` and `re-mine`
- `MEMPALACE_EMBEDDING_DEVICE` → replaced by auto-detection (cuda > mps > cpu)
- `MEMPALACE_CHUNK_SIZE` / `MEMPALACE_CHUNK_OVERLAP` → replaced by `--chunk-size` / `--chunk-overlap` flags
- `MEMPALACE_FORCE_EMBEDDING` → replaced by `--force` flag on `get_collection()`

### Where config lives now

Collection metadata is the single source of truth:

```json
{
  "embedding_model": "intfloat/multilingual-e5-base",
  "chunk_size": 450,
  "chunk_overlap": 50
}
```

Set once at `init`, readable without instantiating the embedding function (`read_collection_metadata()` opens ChromaDB with no embedder, reads metadata, closes). Changed only via `re-mine`.

### Device auto-detection

`MempalaceConfig.detect_device()` probes hardware at runtime:
- `torch.cuda.is_available()` → `"cuda"`
- `torch.backends.mps.is_available()` + `platform.machine() == "arm64"` → `"mps"`
- fallback → `"cpu"`

No configuration needed. The `arm64` guard prevents a false positive on Intel Macs where `mps.is_available()` returns True but MPS doesn't actually work.

### Legacy migration

Existing palaces without an `embedding_model` in their collection metadata are silently stamped `"chromadb-default"` on first access. No user action required.

### What's preserved from community contributions

This PR incorporates work from @FabioLissi:
- **Embedding model mismatch detection** — stores model name in collection metadata, raises `EmbeddingModelMismatchError` on mismatch with clear recovery instructions
- **`mempalace re-mine` command** — extracts source files from existing palace, backs up, drops collection, re-mines with the new model
- **MCP error propagation** — `EmbeddingModelMismatchError` surfaces through `_get_collection()` and `handle_request()` instead of being swallowed by generic `except Exception`

The core ideas remain — init-time binding just moves the config source from env vars to collection metadata, which is where the model was already being tracked.

## Benchmark

Tested end-to-end on 247 Russian blog posts with `intfloat/multilingual-e5-base`:

| Step | Result |
|------|--------|
| `init --model intfloat/multilingual-e5-base` | 13s, metadata stamped correctly |
| `mine` (247 files) | 758 drawers, 1:54 on CPU |
| `search` (10 queries, warm) | 48ms avg, 0.719 avg similarity, 100% high-relevance (≥0.65) |

Tests: 646 passed, 0 failed (35 errors are onnxruntime platform issues on macOS x86_64, not code-related).

Rebased on latest `develop`, single commit, no conflicts.

igorls:
Hey @NickShtefan We would like to include for the next release once the plugin based storage layer is added, please wait to rebase

NickShtefan:
Quick context on why I came back to this now: I wanted to update my personal install to the latest `develop` (had been running my own PR for a while in production). Since my own copy was already running on this PR, I figured I'd refresh the PR itself at the same time — rebased onto current `develop`, dropped the env-var device support that's now in `develop` via @igorls's `a4868a35` (`perf(mining): batch + GPU acceleration`), and folded in the `_get_collection` MCP fix that closes #903.

Everything verified against my live palace (15K+ drawers, `intfloat/multilingual-e5-base`) — `tool_diary_read` and `mempalace_search` over MCP both return correct results after the rebase.

Following @igorls's earlier note that this should land "once the plugin based storage layer is added" — happy to wait for #1196 (SQLiteVec backend) before merging if that's still the plan, or merge sooner if the storage layer timing has shifted. Whatever works for you.

jphein:
Following over from [#390](https://github.com/MemPalace/mempalace/issues/390) where @NickShtefan tied the threads together — a few thoughts on direction.

**The bind-chunk-size-to-collection-metadata property is the load-bearing piece.** The reason is the silent-drift failure mode: change `chunk_size` after a palace is mined and the new chunks may exceed the model's token window, getting silently truncated before embedding. With the binding stamped into chromadb collection metadata at `init`, that mismatch becomes a property chromadb itself can refuse — much harder to footgun than env-var resolution.

**Composition with #1024.** Different layers; both wanted:

- #1024 ships configurable `chunk_size` / `chunk_overlap` / `min_chunk_size` as `config.json` keys + `MempalaceConfig` properties — the immediate-need override for the MiniLM-256-token problem. Doesn't touch collection metadata.
- #442 binds chunk-size + embedding-model into collection metadata at `init` time, makes the bound model authoritative for subsequent ingest *and* MCP query path (closes #903's MCP-vs-ingest mismatch). Strictly stronger guarantee.

If maintainer direction is to wait for #1196 (SQLiteVec backend) per @igorls's earlier note, that's reasonable — the plugin-storage layer changes the metadata-binding shape and rebasing now risks a third rebase later. The intermediate state where users have #1024 (config override) but not #442 (collection-bound) is acceptable as long as the gap is documented in CHANGELOG: *"changing `chunk_size` after a palace is mined will produce drift; re-mining a wing is the recovery path until #442 lands."*

The `_get_collection` MCP fix you folded in is independently valuable and would be a clean small extraction if the rest of #442 has to wait — happy to comment on a smaller scoped PR for just that piece if you'd find it useful.

(For context: the fork has #1024-equivalent in production but not #442-equivalent; the silent-drift failure mode you describe hasn't bitten us yet only because we haven't changed embedding model. So I'm reading #442 as the durable fix, #1024 as the bridge.)

davidglidden:
Adding production-data point + use case in support of this PR.

I've been running a local patch that effectively does the minimum-viable version of what this PR proposes — hardcoded `BAAI/bge-m3` with `device="mps"` in `mempalace/backends/chroma.py`. Working assumption was always that I'd retire the patch once first-class config landed. This PR is exactly that path.

**Use case** — literary / contemplative semantic memory:

- ~600 MB of literary, philosophical, and scholarly texts (English-dominant but with French, German, Italian, Latin, Greek)
- Particular sensitivity to Greek (cited in body-text cross-references) and French (translation work, some original notes)
- Multilingual scholarly works — semantic fidelity matters across translations and transliterated classical languages

**Why `BAAI/bge-m3` for this case** (offered as a doc-recommendation candidate, not a default change):

- 1024-dim — richer semantic space for nuanced queries against literary texts
- **8192-token context window** vs e5-base's 512 — meaningful for long literary passages where 512-token chunks break semantic units mid-thought
- Strong multilingual coverage including Greek and Latin per BAAI's evaluation
- License MIT

Trade-off vs e5-base: heavier (~2.3 GB vs ~1.1 GB), longer initial mining time, but the larger context window matters when chunks are long-form literary rather than short-form note-taking.

**Empirical**:

- 800K+ drawers mined from this corpus on \`bge-m3 + MPS\`
- ~3 days mining time on Apple Silicon (M-series)
- Quality has been strong for cross-language semantic queries on classical texts

**Small suggestion for the PR**: README's recommended-models section could note \`BAAI/bge-m3\` as the option for users with longer-context literary / scholarly content. \`e5-base\` for most, \`bge-m3\` for the long-passage case. Two recommendations cover the main use-case axis (note-taking vs long-form).

The init-time binding via collection metadata is the right architecture — strongly agreed. Avoids the env-var-drift class of bugs. Looking forward to retiring my local patch.

(Happy to test the PR branch against this corpus once my current mine completes and report back.)

NickShtefan:
Pushed `c652c2f3` adding an `Embedding models` section to the README per @davidglidden's suggestion. Two-tier framing rather than one default: `chromadb-default` for short-form English, `intfloat/multilingual-e5-base` for multilingual notes, `BAAI/bge-m3` for long-form literary / scholarly content where the 8192-token context window matters. Notes the init-time binding and the re-mine path for switching models. Doc-only, no code changes.

@davidglidden — if you have time after your current mine completes, a test report on the PR branch against your literary corpus would be a useful production-data point alongside @sidonsoft's earlier findings; happy to incorporate any wording tweaks if `bge-m3` deserves more nuance than what's in the table now.

davidglidden:
@NickShtefan -- Preparing a curated mine covering English, French, and Spanish to run overnight. Will post a test report tomorrow. Thanks for reaching out.

davidglidden:
# PR #442 test report — bge-m3 on literary / multilingual corpus

**Test branch:** `feat/multilingual-embedding` @ `c652c2f3`
**Test date:** 2026-05-11
**Platform:** macOS, Apple Silicon (Device: coreml)
**Source PR:** #442 — *feat: add configurable multilingual embedding model support*

Following up on @NickShtefan's request for a production-data point against my literary corpus alongside @sidonsoft's earlier findings. Reporting here so the PR thread has a second independent confirmation that bge-m3 binding works end-to-end.

## Setup

Installed PR branch in a dedicated venv (`pip install -e <pr-worktree>` then `pip install sentence-transformers` — the multilingual extras are required for actual bge-m3 binding; without them, init succeeds but mine silently falls back to ChromaDB default). Wiped any prior test palace; init with `--model BAAI/bge-m3 --yes`; mine separately.

## Corpus

**11 markdown files, ~8 MB, 21,735 drawers, 25 min mining wall-clock** on M-series CoreML. Curated to exercise bge-m3's stated multilingual + long-form claims:

| Title | Author | Language | Drawers |
|---|---|---|---:|
| *Orthotypographie* Vol I (A-F) | Lacroux | French | 2,296 |
| *Orthotypographie* Vol II (G-Z) | Lacroux | French | 2,327 |
| *Lexique des règles typographiques (Imprimerie Nationale)* | French state typography office | French | 939 |
| *Ortografía y ortotipografía del español actual* | Martínez de Sousa | Spanish | 5,874 |
| *The Solid Form of Language* | Bringhurst | English | 231 |
| *Detail in Typography* | Hochuli | English | 303 |
| *Space Between Words: Origins of Silent Reading* | Saenger | English (paleography) | 3,626 |
| *Forests: The Shadow of Civilization* | Harrison | English | 1,999 |
| *Gardens: An Essay on the Human Condition* | Harrison | English | 1,348 |
| *Juvenescence: A Cultural History of Our Age* | Harrison | English | 1,316 |
| *La Poétique de l'espace* | Bachelard | French | 1,476 |

## Results — cross-language semantic retrieval works

Ten queries crossing language boundaries. Trimmed scores; full output available.

**Q: "italic for emphasis"** (English) →
- `[1] Hochuli (EN)` cos 0.622 — *"emphasize in italic; that will make it stand out"*
- `[2] Lexique IN (FR)` cos 0.64 (BM25=0.0) — *"Le caractère italique est utilisé principalement pour attirer l'attention du lecteur sur un mot, une phrase ou un passage que l'auteur tient à souligner."*

Pure-semantic match (BM25=0.0) at higher cosine than the lexical English hit. **The PR's central claim verified on first query.**

**Q: "intimate space of dwelling"** (English) → Bachelard *La Poétique de l'espace* (French) ×4, cos 0.585-0.612, all BM25=0.0. English query, French book, no lexical overlap — the book retrieves entirely on semantic similarity.

**Q: "approche typographique"** (French — the French technical term for letter-spacing) →
- `[1-2] Lacroux Vol I (FR)` cos 0.637, 0.611 — native-language match
- `[3] Lexique IN (FR)` cos 0.652
- `[4] Hochuli (EN)` cos 0.615 (BM25=0.0) — semantic bridge to the English typographic tradition

**Q: "letter spacing between pairs of characters"** (English) → top 3 English (Saenger, Hochuli), but `[4] Sousa (Spanish)` cos 0.612 BM25=0.0 — English query reaching into the Spanish typographic vocabulary on the same concept.

**Q: "lineas viudas y huerfanas"** (Spanish for "widow and orphan lines") →
- `[1-2] Sousa (Spanish)` cos 0.621, 0.603 — native
- `[3-4] Lacroux Vol II (French)` cos 0.54, 0.522 — cross-language semantic on the typographic concept

**Q: "comillas españolas tipos"** (Spanish, "Spanish quotation mark types") → 4 Sousa hits, cos 0.609-0.636 — clean native-language retrieval.

**Q: "the dead and the unborn"** (English, Harrison's signature phrase) → 4 Harrison hits (Forests + Juvenescence) cos 0.52-0.53 — long-form scholarly retrieval on a specialized conceptual vocabulary.

**Q: "humic burial dead"** (Harrison's coinage) → 4 Forests hits cos 0.498-0.536 — even his idiosyncratic specialist vocabulary retrieves coherently.

**Q: "restraint in ornament"** (English, Tschichold/Bringhurst posture) →
- `[1-2] Harrison Forests (EN)` cos 0.559, 0.562 — note Harrison's prose on civilizational restraint maps to typographic restraint via the same semantic vector
- `[3] Lacroux Vol I (FR)` cos 0.571 (BM25=0.0)
- `[4] Bachelard *La Poétique* (FR)` cos 0.567 (BM25=0.0)

Cross-tradition AND cross-language. Notable that "restraint in ornament" retrieves Harrison's *Forests* passage on civilizational restraint — the embedding space is bridging conceptual register across domains, not just lexical surface.

## What this corpus shows that wasn't obvious before

Three patterns worth flagging for the README / PR description:

**1. The `sentence-transformers` install dependency is a usability footgun.** The first run after `pip install -e .` succeeded silently — init reported `Embedding model: BAAI/bge-m3` and mine reported successful completion — but emitted the warning *"sentence-transformers not installed — falling back to ChromaDB default."* and produced a palace bound to MiniLM (384d, English-leaning), not bge-m3 (1024d, multilingual). The metadata was stamped correctly; only the embedding function was wrong. Suggestion: hard-error rather than warn-and-fall-back when `--model` is non-default but sentence-transformers is missing. Or document `pip install sentence-transformers` (or `pip install mempalace[multilingual]`) prominently in the README's *Embedding models* section as a required dependency for non-default models.

**2. The `--auto-mine` flag doesn't honour `--palace` for the auto-mine step** (init creates the right palace, but `_maybe_run_mine_after_init` reads `cfg.palace_path` rather than `args.palace`). Workaround used here: run init and mine as separate steps; both honour `--palace` correctly. Worth a small follow-up patch threading `args.palace` through.

**3. Cosine scores on bge-m3 cluster lower than on default MiniLM** (typical range 0.50-0.70 for clear conceptual matches vs MiniLM's 0.7-0.9 on the same kind of corpus). Worth a note in the README that bge-m3's score range is different — operators accustomed to MiniLM scoring may misread a 0.60 cosine as "weak" when it actually represents strong semantic match in bge-m3's space. The empirical scores above suggest a working threshold of cosine ≥ 0.50 (vs MiniLM's typical ≥ 0.65) for relevance on this corpus.

## Conclusion

PR #442 delivers what it promises on this corpus. The cross-language semantic retrieval works on first query, even on technical typographic vocabulary that's domain-specific in each language (*approche* / *kerning* / *interletraje*, *cursiva* / *italique* / *italic*, *viuda* / *veuve* / *widow*). Long-form scholarly content (Harrison trilogy, Saenger, Bachelard) retrieves coherently with the 8192-token context window.

Retires my local hardcoded bge-m3 patch in `chroma.py` (from April) — the upstream surface is exactly what I'd want it to be. Init-time stamping into collection metadata + the `re-mine --model` path for migration is the right architecture.

Two small footnotes (the sentence-transformers install gate and the `--auto-mine` palace-path bug) — both fixable in follow-ups. The PR's claim and the implementation align.

Happy to test on a larger corpus (full ~600 MB literary library) once #442 lands on `develop` and `develop` lands on `main`; this 8 MB subset was sized to fit a pre-flight window. Will also retest after v3.3.5 ships (the lock-fix at the backend seam) for a production-realistic concurrent-write profile.

davidglidden:
Quick follow-up to [the test report above](https://github.com/MemPalace/mempalace/pull/442#issuecomment-4418237732), with a finding that might be useful for the path forward.

After 3.3.5 shipped on `main` this weekend, I started setting up for the larger-corpus mine — the full ~600 MB literary corpus I'd flagged earlier. Looked for a way to get the bge-m3 binding running on a stable base in the meantime.

I expected this to be the kind of thing where I'd carry forward my old hardcoded patch from April. It turned out to be much smaller than that — and the smallness is the point.

## The whole patch is two hunks

**Hunk 1 — `mempalace/embedding.py`** (~10 lines): replace `get_embedding_function()`'s body to return `SentenceTransformerEmbeddingFunction(model_name="BAAI/bge-m3", device="mps")` with a small cache.

**Hunk 2 — `mempalace/backends/chroma.py`** (~3 lines, both create sites): add `"embedding_model": MEMPALACE_EMBED_MODEL` to the existing `metadata=` dict.

That's it. Combined diff against `origin/main` at `v3.3.5`:

```
 mempalace/backends/chroma.py |  3 +++
 mempalace/embedding.py       | 45 ++++++++++++++++++++++++++------------------
 2 files changed, 30 insertions(+), 18 deletions(-)
```

End-to-end verified: collection metadata stamps `embedding_model = BAAI/bge-m3` at the SQLite row level, query roundtrip returns the right drawer, EF identity reports `sentence_transformer` consistently across write and read paths.

## What I noticed

`#1299` (3.3.4 — *"MCP server `tool_diary_write` SIGSEGV when default EF provider differs"*) ended up doing a lot of the structural work for multilingual support. By centralizing EF resolution at `ChromaBackend._resolve_embedding_function()` and routing both `mcp_server._get_collection` and the backend's own collection paths through it, there's now a single seam where the embedding function is resolved exactly once. Everything downstream flows through that seam transparently.

In that architecture, swapping the embedding model is mostly *just* a swap at the seam. The reason #442 is a ~999-line PR isn't the binding itself — it's everything around it:

- `mempalace init --model <name>` CLI flag and the config plumbing for it
- Init-time stamping of `embedding_model` into collection metadata (matters once you have multiple model choices)
- `mempalace re-mine --model <new>` migration path between models
- Legacy palace auto-migration that stamps `"chromadb-default"`
- Mismatch detection between EF identity and stored metadata
- The `_get_collection` fix in mcp_server.py that 8e4319c addressed
- Auto-detect device cascade (cuda > mps on arm64 > cpu)
- Removal of the old `MEMPALACE_EMBEDDING_MODEL` env var
- ~390 lines of new tests for the above

All of that is real work and worth having. Just noticing that it's somewhat independent of the underlying binding.

One thought, only if it'd be useful: depending on how the rebase against current `develop` shapes up, the embedding-seam swap and metadata stamp could potentially land on their own — even with something simpler than `--model` (env var or config field) for the model selection. Just the core piece. The CLI / re-mine / migration apparatus can then layer on top in follow-ups whenever it makes sense. For users in literary / non-English corpora, the mining capability alone is already significant.

## A caveat

The patch I described is **local-only** for our chamber-library mine. Not proposing it as the path forward upstream — your architecture in #442 (configurable model + stamping + migration + mismatch detection) is the right shape for a general-purpose tool. The minimal patch only works because we have a fixed corpus and a single model in mind.

Sharing it as evidence that the core piece is small in 3.3.5's architecture, in case that's useful context.

Will report back on the chamber-library full mine when it completes. Plan: ~700k drawers across French/Spanish/English typography sources, paleography, scholarly philosophy. Mining on the local-patched 3.3.5 with single-writer discipline (MCP off during the mine), using 3.3.5's `repair --mode from-sqlite` (#1308) as the safety net.

Thanks again for #442 — the bridge it builds between English-leaning ChromaDB defaults and genuinely multilingual corpora matters a lot for what this corpus is.

NickShtefan:
@davidglidden — thanks for the production test report and the minimal-patch follow-up. Pushed `ceccbfa1` addressing both findings you surfaced.

**Both bugs fixed:**

1. `--auto-mine` now honours `--palace`. `_maybe_run_mine_after_init` was reading `cfg.palace_path` while `cmd_init` honoured `args.palace` — same `getattr(args, "palace", None)` fallback pattern threaded through.

2. Hard `ImportError` (with install hint) instead of silent fallback to MiniLM when `sentence-transformers` is missing for a non-default model. Worth noting for anyone reading this thread: chromadb's `SentenceTransformerEmbeddingFunction.__init__` catches the underlying `ImportError` and re-raises as `ValueError`, so catching `ImportError` around the chromadb call alone is insufficient — probing `import sentence_transformers` directly is the cleaner seam.

**One cascading fix** the non-silent embedder exposed: dropping the fallback meant chromadb's internal EF-conflict validator (`Embedding function conflict: new: sentence_transformer vs persisted: default`) started firing before `palace.get_collection` could raise its own `EmbeddingModelMismatchError`, masking the actionable "run `re-mine --model` or pass `--force`" message. Added a pre-check that reads `embedding_model` from collection metadata up front and raises `EmbeddingModelMismatchError` directly; force-path opens the collection with the stored EF and re-stamps metadata. Bonus: avoids loading a heavy model only to discard it on mismatch.

Tests:
- Replaced `TestGetEmbeddingFunctionFallback` with `TestGetEmbeddingFunctionMissingDependency` (asserts raise + install hint).
- Added a `sentence_transformers` `sys.modules` stub fixture so the existing mocked-`SentenceTransformerEmbeddingFunction` tests still pass on a stock `[dev]` env without the `[multilingual]` extra.
- Added regression guards in `tests/test_cli.py` for fix 1.

Full suite green (1339 passed). README intentionally left for a separate docs-only follow-up — the `[multilingual]` extras gate and the bge-m3 cosine-threshold note you flagged are both worth landing, just keeping this commit scoped to the two bug fixes.

davidglidden:
Following up on the [2026-05-11 validation report](https://github.com/MemPalace/mempalace/pull/442#issuecomment-4418237732). A 1.29M-drawer mine on this branch surfaced two findings worth feeding back — neither critiques the bge-m3 integration itself, only the scale validation methodology and one small ancillary question.

## Honest amendment to the 2026-05-11 validation

Re-examining the test palace post-hoc: the 21.7k mine succeeded for search purposes (FTS5 + brute-force vector fallback returned correct results) but the VECTOR segment's `max_seq_id` was **859 rows behind** the sqlite drawer count at completion (~4% drift).

```
drawers in sqlite:   21,752
VECTOR seq_id:       20,893
Drift:               859 drawers (~4%)
```

The Stage-2 integrity gate in `quarantine_stale_hnsw` correctly classified this as benign flush-lag — no quarantine fired, HNSW remained fully populated (data 88 MB, link_lists 179 KB, pickle 2.9 MB) — so the search-quality outcome was real and the validation's "passed" framing was accurate about *search behavior*. But it didn't characterize the full HNSW state.

Surfacing it now in case future PR #442 validations want to include a `max_seq_id` parity check alongside the search-quality queries. A small post-mine assertion that `embeddings.count == max_seq_id` for the VECTOR segment would catch this class of drift directly.

## Scale finding (filed separately upstream)

A subsequent 1.29M-drawer mine of a larger corpus on this same branch did not complete cleanly. 37 quarantine events compounded into a 38,966-row `max_seq_id` gap (73% of corpus invisible to vector). The mechanism appears to be `quarantine_stale_hnsw`'s 300s mtime threshold interacting with multi-hour bge-m3 + MPS embeds of single large sources — a false-positive cascade where each event throws away in-flight HNSW work.

Full forensic + reproduction recipe + suggested fixes filed as [#1526](https://github.com/MemPalace/mempalace/issues/1526). The mechanism is embedding-model-adjacent but the fix lives in the chromadb integration layer (`_segment_appears_healthy`, `stale_seconds`, post-mine state surfacing), not the PR #442 scope.

## Suggestion for scale validation methodology

The 11-book subset used in 2026-05-11 was all small files (largest ~5 MB). The cascade pattern doesn't surface on small-file workloads regardless of total drawer count — it surfaces specifically when *one source's embed time* exceeds the 300s threshold.

If PR #442 wants to characterize bge-m3 behavior beyond the current validation: including at least one ~25 MB plain-text single-source file in the validation corpus would surface the cascade before merge. The drawer-count axis (21k → 100k → 1M) is fine but secondary; the single-source-size axis is what triggers Mode 1.

(The 21.7k validation also implicitly demonstrated that the Stage-2 integrity gate handles small benign drift correctly — that part of the gate works. The cascade is specifically when the gap is sustained for hours with the pickle in mid-write states.)

## One side question

The 21.7k test palace's `mempalace_closets` collection has `embedding_model: chromadb-default` in `collection_metadata`, while `mempalace_drawers` is correctly on `BAAI/bge-m3`. Is the closets infrastructure intentionally decoupled from the configurable-embedding contract, or should it inherit `MEMPALACE_EMBED_MODEL` as well? Not a bug per se; just unclear from reading the PR whether this is intentional scope or an oversight.

(If intentional: ignore. If oversight: the closets-collection creation path may need the same `metadata={..., "embedding_model": MEMPALACE_EMBED_MODEL, ...}` stamp that the drawers path now has.)

---

Environment details + the local commits atop `v3.3.5` (the three PR #442 integration patches `975acba` / `5458b77` / `604df3b`) are documented in [#1526](https://github.com/MemPalace/mempalace/issues/1526). Forensic file available on request.

Thank you for the PR work — the bge-m3 integration itself produces clean multilingual results at the scales that worked (cross-language retrieval verified at 21.7k; semantic similarity ranges sensible). The above is scale-and-methodology context, offered as a useful early indicator.

NickShtefan:
@davidglidden thanks for the 1.29M-scale run and the careful write-up.

- closets: good catch, that's an oversight. closets should inherit the same embedding model as drawers (a single palace shouldn't mix models, or closet search drifts to default-MiniLM behavior for non-English). I'll stamp embedding_model on the closets-collection creation path in the next push, mirroring the drawers path.
- #1526 cascade: agreed it's outside #442's scope. The fix belongs in the chromadb integration layer (_segment_appears_healthy / stale_seconds), not the embedding binding. Good to track it via #1526.
- validation methodology: will add an embeddings.count == max_seq_id parity assertion and at least one ~25 MB single-source file to the validation corpus so the slow-embed cascade surfaces pre-merge.

Thanks again for the production data points.

NickShtefan:
@davidglidden pushed the closets/compressed stamp fix (2b28cf7d), closing the loop on your finding above. Secondary collections were created without threading the palace's model, so they fell back to the ChromaDB default while drawers used the configured one. get_collection now inherits the palace's canonical model (from the drawers collection) when creating a non-default collection that has no stored model and no explicit override; stored models are never overridden, so existing palaces are unaffected (a re-mine migrates them). Added tests for closets/compressed inheritance and the default-stays-default case.

NickShtefan:
Rebased — or more accurately, **rebuilt** — this branch on current `develop` (5cd961e). The PR is conflict-free and mergeable again.

Since the last push, `develop` absorbed nearly everything this PR originally carried, in better-integrated forms: the RFC 001 embedder-identity sidecar + `palace set-embedder` (#1724 / #1730) replaced the collection-metadata stamping and `EmbeddingModelMismatchError`; `repair rebuild-index` (#1308 / #1670) replaced `re-mine --model`; configurable chunking landed via #1024; the multilingual default shipped as `embeddinggemma` (#1483); and the MCP `_get_collection` EF fix went in via #1303. Carrying those parallel mechanisms forward would have duplicated all of that, so this revision drops everything superseded and keeps only the gap that still exists on `develop`:

**`get_embedding_function()` recognizes only `minilm` / `embeddinggemma` — any other `embedding_model` value silently falls back to MiniLM.** That's the same silent-fallback footgun @davidglidden flagged in the May test report, and it means the e5-base / bge-m3 use cases validated earlier in this thread still require a fork.

What the rebuilt branch contains (3 commits, ~600 lines instead of ~1,300):

- **`embedding.py`**: any non-builtin model id routes to chromadb's `SentenceTransformerEmbeddingFunction` behind a lazy loader (same lazy contract as `EmbeddinggemmaONNX`, so collection-open paths that never embed — MCP status, hooks, wake-up — stay inside the startup budgets). The torch device maps from the existing `embedding_device` vocabulary (`coreml`→`mps` with an arm64 guard for Intel Macs, `dml`→CPU with a one-shot warning).
- **Hard, actionable failures instead of the silent fallback**: `EmbeddingDependencyError` (with the `pip install 'mempalace[multilingual]'` hint) when sentence-transformers is missing, `EmbeddingModelError` when the model itself can't load — surfaced through the chroma EF resolver, both `mcp_server._get_collection` branches, `search_memories`, and the wake-up layers, instead of the misleading "Backend open failed" / "No palace found" (the MCP-path failure mode @FabioLissi called out in the earlier rounds).
- **`init --model <name>`** — following @igorls's direction that the model is set at init: persists the same config.json key onboarding writes, probes dependencies up front (exit 2 + install hint instead of failing mid-mine), and the palace records its identity in the sidecar at first collection creation.
- **`[multilingual]` extra** (currently an empty back-compat alias) now installs `sentence-transformers>=3.0`.
- **README**: the *Embedding models* section — model table, the install gate, and the bge-m3 cosine-range calibration note (strong matches cluster ≥0.50 vs MiniLM's ≥0.65) promised to @davidglidden.
- **Tests**: new offline `tests/test_sentence_transformer_ef.py` (dispatch, lazy load, device mapping, error propagation, MCP error surfacing) — the suite passes on a stock `[dev]` env without torch or sentence-transformers installed.

Verification: 3,396 passed / 31 skipped locally (the one failing test, `test_checkpoint_added_by_accepted_via_dispatch`, reproduces on clean `develop` without this branch — looks like an order-dependent flake in the new checkpoint-attribution area, happy to file it separately). End-to-end on Apple Silicon with a real ST model: `init --model` + auto-mine on `mps`, sidecar identity recorded for `mempalace_drawers` **and** `mempalace_closets` (the secondary-collection concern from the earlier round falls out of the sidecar design for free), search verified, a model switch raises `EmbedderIdentityMismatchError` with recovery instructions, and a missing dependency fails init with exit 2 instead of silently binding MiniLM.

The previous revision is preserved at [`feat/multilingual-embedding-v1-backup`](https://github.com/NickShtefan/mempalace/tree/feat/multilingual-embedding-v1-backup) for reference; a PR-description refresh to match the rebuilt scope is coming next.

The CI runs are sitting in `action_required` — could a maintainer approve the workflows when convenient?

igorls:

NickShtefan:
@igorls — rebased on current `develop` (`4bc0c43`). Conflict-free, and I've refreshed the PR description, which was still describing the April revision.

The scope is unchanged from the July rebuild: `get_embedding_function()` resolves `minilm` / `embeddinggemma` / `openai-compat`, and anything else silently falls back to MiniLM. This routes those values to `SentenceTransformerEmbeddingFunction` and replaces the fallback with actionable errors. Everything else this PR once carried is already on `develop` in better form and has been dropped.

**One thing worth flagging from the rebase.** `openai-compat` (#1559) landed after the last revision, and the naive merge broke it: this PR's dispatch treats "not a built-in id" as "a sentence-transformers model", so `embedding_model: "openai-compat"` would have been sent to HuggingFace as a repo name. The built-in set is now a single public `BUILTIN_MODELS` in `embedding.py`, which `MempalaceConfig._normalize_embedding_model` also reads instead of keeping its own copy — so the two lists can't drift again. Side effect: `OpenAI-Compat` with different casing now normalizes correctly, which it didn't before. The dispatch order is openai-compat → sentence-transformers → embeddinggemma → minilm.

Two smaller adaptations: `_resolve_providers` is now `(device, model)` for the CoreML denylist from #2283, so `describe_device` grew a third branch (remote endpoint / torch device / ONNX provider); and the new `EmbeddingDependencyError` arm in `searcher.py` goes through `_search_error_result()` so the error envelope keeps its `results: []`.

`tests/conftest.py` gets one line: `test_sentence_transformer_ef` added to `_REAL_EMBEDDING_TEST_MODULES`, the same opt-out `test_embedding_api` uses — the autouse stable-EF fixture patches `ChromaBackend._resolve_embedding_function`, which is one of the seams these tests assert on.

**Status:**

- `ruff check .` — clean
- Full suite, CI's exact command including `--cov-fail-under=80`: **4508 passed, 31 skipped, 1 failed**, coverage 82.48%

The one failure is `tests/test_mcp_server.py::TestWriteTools::test_checkpoint_added_by_accepted_via_dispatch`, and it is **not** from this branch: I ran the same full suite on a pristine `upstream/develop` worktree and it fails there too (4467 passed, same single failure). It passes in isolation and passes when its own module runs alone, on both — so it's order-dependent state leaking from somewhere earlier in the suite. Filed as #2334 with the repro — it's the MCP writer-lock module globals leaking between tests, so a mutating-tool test that doesn't reset them itself gets refused at the dispatcher.

**One ask:** this push queued three workflow runs — Tests, Version Guard, Docker — and all three are sitting at `action_required`, the fork-PR approval gate. Same story for the whole history of the PR: the last run that actually executed was on 13 April, and the pushes on 26 Apr, 9/11/26 May and 17 Jul produced no runs at all. So "CI is green" isn't something I can demonstrate from my side until someone hits **Approve and run workflows** — could you? Local results above are the CI command verbatim, so I don't expect surprises.

The previous revision is preserved at [`feat/multilingual-embedding-v2-backup`](https://github.com/NickShtefan/mempalace/tree/feat/multilingual-embedding-v2-backup) (and the pre-July one at `feat/multilingual-embedding-v1-backup`).

PR #459, opened 2026-04-10
fix: use permissive sanitizer for KG object field

### LINKED ISSUE #455: KG object field too restrictive: sanitize_name rejects commas and common punctuation
## Problem

When adding knowledge graph triples via , the  field is validated using  which enforces a strict character set (). This rejects commas, colons, parentheses, and other common punctuation that naturally appears in descriptive KG objects.

**Example that fails:**

**Workaround:** Split into one-fact-per-triple. This works but is clunky for descriptive facts and inflates triple counts.

## Root cause

In  line ~429, all three KG fields use :

Subject and predicate are structural (entity IDs, relationship types) — strict validation makes sense. But the object field is often descriptive text, not a path component or entity name.

## Suggested fix

Use  (with a max_length cap) for the  field instead of :

This preserves security (null byte and length checks) while allowing natural text in KG object values. The underlying SQLite storage uses parameterized queries so there is no injection risk.

## Additional suggestion: bulk KG import

When seeding a knowledge graph from scratch (e.g. mapping a family structure and project states), adding 60+ triples one-at-a-time is slow. A  tool that accepts an array of triples would significantly improve the onboarding experience.

---

Discovered while seeding a KG with family/project data as a daily-use archivist. Happy to PR either fix if welcome. Great tool — using it daily with autonomous cron sessions.

mrdeeme:
## Problem

`tool_kg_add` validates all three fields (`subject`, `predicate`, `object`) with `sanitize_name()`, which enforces `[a-zA-Z0-9_ .'-]`. This rejects commas, colons, parentheses, and other common punctuation that naturally appears in descriptive KG objects.

Example that fails:
```
tool_kg_add("Max", "born", "April 1, 2015 (age 11)")
# → ValueError: object contains invalid characters
```

## Root cause

In `mcp_server.py` line ~429, `sanitize_name()` is applied to the `object` field. Subject and predicate are structural (entity IDs, relationship types) — strict validation makes sense. But the object field is often descriptive text, not a path component or entity name.

## Fix

Add `sanitize_kg_object()` in `config.py` — a permissive validator that:
- **Allows** free-form text with punctuation (commas, colons, parentheses, brackets, etc.)
- **Still blocks** null bytes, path traversal (`..`, `/`, `\`), and excessively long strings (max 256 chars)
- Is safe because SQLite uses parameterized queries — no injection risk

Changes:
- `subject` and `predicate` remain validated with `sanitize_name()` (strict)
- `object` now uses `sanitize_kg_object()` (permissive)

## Tests

58 new tests in `tests/test_sanitize_kg_object.py`:

| Group | Count | Covers |
|---|---|---|
| Accepts | 25 | Commas, colons, parentheses, emoji, unicode, max-length boundary |
| Rejects | 11 | Empty, null byte, path traversal, wrong type, overflow |
| Stripping | 5 | Spaces, tabs, newlines |
| Field name | 2 | Custom error messages |
| Regression guard | 5 | `sanitize_name` stays strict |
| Integration (tool_kg_add) | 10 | End-to-end with punctuation, subject/predicate still strict, write→read roundtrip |

Full suite passes: **592 passed, 0 failed**.

Closes #455

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -47,6 +47,34 @@ def sanitize_name(value: str, field_name: str = "name") -> str:
     return value
 
 
+MAX_KG_OBJECT_LENGTH = 256
+
+
+def sanitize_kg_object(value: str, field_name: str = "object") -> str:
+    """Validate a KG object field — allows descriptive text with punctuation.
+
+    More permissive than sanitize_name (which is for entity IDs / predicates)
+    but still blocks null bytes, path traversal, and excessively long strings.
+    """
+    if not isinstance(value, str) or not value.strip():
+        raise ValueError(f"{field_name} must be a non-empty string")
+
+    value = value.strip()
+
+    if len(value) > MAX_KG_OBJECT_LENGTH:
+        raise ValueError(
+            f"{field_name} exceeds maximum length of {MAX_KG_OBJECT_LENGTH} characters"
+        )
+
+    if ".." in value or "/" in value or "\\" in value:
+        raise ValueError(f"{field_name} contains invalid path characters")
+
+    if "\x00" in value:
+        raise ValueError(f"{field_name} contains null bytes")
+
+    return value
+
+
 def sanitize_content(value: str, max_length: int = 100_000) -> str:
     """Validate drawer/diary content length."""
     if not isinstance(value, str) or not value.strip():
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -26,7 +26,7 @@ import hashlib
 from datetime import datetime
 from pathlib import Path
 
-from .config import MempalaceConfig, sanitize_name, sanitize_content
+from .config import MempalaceConfig, sanitize_name, sanitize_content, sanitize_kg_object
 from .version import __version__
 from .searcher import search_memories
 from .palace_graph import traverse, find_tunnels, graph_stats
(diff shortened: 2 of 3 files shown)


web3guru888 (commented):
This directly fixes a pain point we've hit. We maintain a knowledge graph with 710+ entities and 1,014 triples, and the over-strict `sanitize_name()` on the object field was rejecting perfectly valid descriptive values like "chess, swimming, reading" or "born April 1, 2015 (age 11)".

The two-tier approach is correct: subjects and predicates are identifiers (strict), objects are descriptive text (permissive). The test coverage is thorough — the roundtrip test (`test_add_then_query_roundtrip`) is especially good.

One concern: the `/` rejection in `sanitize_kg_object()` will reject some legitimate KG objects:

- `"CEO/CTO"` — common role notation
- `"client/server architecture"` — technical descriptions
- `"2024/Q3"` — date/period formats

The path traversal concern is real, but since KG objects never touch the filesystem (they go into SQLite), the attack surface is different from wing/room names. Would it make sense to only reject the `../` pattern (directory traversal) rather than all forward slashes? That would block the actual exploit while allowing legitimate slash-containing text.

Minor: `MAX_KG_OBJECT_LENGTH = 256` feels tight for a field that's now explicitly meant to hold descriptive text. Some of our KG triples have objects that are short sentences ("Discovered anomalous redshift pattern in NGC 4151 consistent with binary black hole model"). 256 will work for most cases, but 512 might give more headroom without meaningfully increasing risk.

bensig:

PR #461, opened 2026-04-10
docs: add LM Studio MCP client setup to README

### LINKED ISSUE #418: Works with LM Studio as an MCP client — suggest adding to docs
Hi — just wanted to share that MemPalace works cleanly with LM Studio (lmstudio.ai) as an MCP client, which isn't mentioned in the README. It might be worth adding alongside Claude, ChatGPT, and Cursor.
Setup that worked for me on Windows 11:
Edit %USERPROFILE%\.lmstudio\mcp.json (Developer tab in LM Studio also has an "Edit mcp.json" option) and add:
json{
  "mcpServers": {
    "mempalace": {
      "command": "C:\\Users\\<user>\\AppData\\Local\\Python\\pythoncore-3.14-64\\python.exe",
      "args": ["-m", "mempalace.mcp_server"]
    }
  }
}
LM Studio auto-loads the server on save — no restart needed. A mempalace entry then appears in the per-chat tools sidebar, and enabling it makes the full tool set available to the loaded model.
Notes for anyone else trying this:

Tool-call reliability depends heavily on the model. Qwen 3 14B Instruct works great end to end — it searched my palace and returned real results on the first try. Gemma 3n 4B crashed with no error when asked to search the palace (possibly a known Gemma 3n tool-calling issue in llama.cpp, not a MemPalace problem).
Context length defaults are usually 4K which fills up fast once tool results come back. Bumping to 8K–16K with Flash Attention on is a more comfortable starting point.
On the Claude Desktop side, the config file lives at a non-obvious path when the app is installed from the Microsoft Store:
C:\Users\<user>\AppData\Local\Packages\Claude_<hash>\LocalCache\Roaming\Claude\claude_desktop_config.json
rather than the usual %APPDATA%\Claude\. This tripped me up for a while and a note in the docs might save others the same detour.

Happy to submit a PR adding an "LM Studio" section to the README if that would be useful — just let me know.
Thanks for building this, it's become a really nice piece of my daily setup.

mrdeeme:
## What

Adds a **"With LM Studio"** section to the README, between the MCP-compatible tools section and the local models section.

## Content

- **Direct config** — `mcp.json` snippet with `python -m mempalace.mcp_server`
- **uvx alternative** — zero-install option via `uvx --from git+...`
- **Tips** — model selection (Qwen 3 14B+ recommended), context length advice (8K–16K)
- **Claude Desktop note** — config path when installed via Microsoft Store (`AppData\Local\Packages\Claude_<id>\...` instead of `%APPDATA%\Claude`)

All content is sourced from community-verified setups reported in #418 by @HaysianSmelt and @LoPromise.

## Why

LM Studio is a popular MCP client not mentioned in the README. Multiple users have confirmed it works, and the setup is straightforward but non-obvious.

Closes #418

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,42 @@ Claude calls `mempalace_search` automatically, gets verbatim results, and answer
 
 MemPalace also works natively with **Gemini CLI** (which handles the server and save hooks automatically) — see the [Gemini CLI Integration Guide](examples/gemini_cli_setup.md).
 
+### With LM Studio
+
+LM Studio works as an MCP client. Edit `%USERPROFILE%\.lmstudio\mcp.json` (or use the **Developer** tab → "Edit mcp.json" in LM Studio):
+
+```json
+{
+  "mcpServers": {
+    "mempalace": {
+      "command": "python",
+      "args": ["-m", "mempalace.mcp_server"]
+    }
+  }
+}
+```
+
+LM Studio auto-loads the server on save — no restart needed. A `mempalace` entry then appears in the per-chat tools sidebar.
+
+Or via `uvx` (no install required):
+
+```json
+{
+  "mcpServers": {
+    "mempalace": {
+      "command": "uvx",
+      "args": ["--from", "git+https://github.com/milla-jovovich/mempalace.git", "python", "-m", "mempalace.mcp_server"]
+    }
+  }
+}
+```
+
+> **Tips:**
+> - Tool-call reliability depends on the model. Qwen 3 14B+ works well; smaller models may struggle.
+> - Default context length (4K) fills up fast with tool results — bump to 8K–16K with Flash Attention on.
+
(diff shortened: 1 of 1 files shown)


web3guru888 (commented):
Useful addition — LM Studio is increasingly popular for running MemPalace with local, privacy-first setups, so documenting this path properly matters.

Seconding the two points Copilot flagged: (1) the config path shown (`%USERPROFILE%`) is Windows-only — macOS/Linux users will need `~/.lmstudio/mcp.json`, worth a platform note or separate tab; (2) the `uvx` example uses the GitHub URL rather than `--from mempalace` (the PyPI package name), which is cleaner and doesn't depend on the GitHub URL staying stable.

One thing worth adding to the model recommendations: the Qwen 3 14B+ threshold is real. We've run several models against the memory palace protocol and smaller ones (7B and below) tend to mis-parse multi-part tool responses or skip required fields in `mempalace_store`. The 8K context minimum is also genuinely necessary — `mempalace_status` alone can return 2–3K tokens, and a full retrieval cycle with context easily pushes past 4K. Would be helpful to note this explicitly so users don't spend time debugging model behavior when it's really a size issue.

aleenprd:
thanks for working on this!

igorls:

PR #468, opened 2026-04-10
fix: streaming/chunking for large transcript files (fixes #396)

CuriosityQuantified:
## Summary

Fixes OOM crashes when processing large transcript files by implementing line-by-line streaming instead of loading entire files into memory.

Fixes #396

### split_mega_files.py
- Converted to streaming: processes files line-by-line with a 7-line lookahead buffer
- `count_sessions_streaming()`: Counts sessions without loading full file
- `split_file()`: Accumulates sessions incrementally, writes to disk as boundaries are found
- Correctly distinguishes true session starts from context restores (Ctrl+E lines)

### normalize.py  
- Added `_try_claude_code_jsonl_streaming()`: Streaming parser for Claude Code JSONL
- Added `_try_codex_jsonl_streaming()`: Streaming parser for Codex JSONL
- Routes `.jsonl` files to streaming parsers before attempting full-file parsing
- Regular JSON formats keep existing approach (500MB safety valve)

## Test Coverage

60 comprehensive tests added:
- **36 tests** for `split_mega_files.py` (unit, integration, error, regression, streaming-specific)
- **24 tests** for `normalize.py` (format detection, JSONL streaming, error handling, regression)

All tests passing ✓

## Backwards Compatibility

- All existing functionality preserved
- 500MB safety limits maintained
- dry-run mode works
- Original file backup renaming works
- Timestamp/people/subject extraction unchanged

## Testing

```bash
pytest tests/test_split_mega_files.py -v  # 36 passed
pytest tests/test_normalize.py -v          # 24 passed
```

### DIFF
--- a/mempalace/normalize.py
+++ b/mempalace/normalize.py
@@ -154,6 +154,60 @@ def _try_claude_code_jsonl_streaming(filepath: str) -> Optional[str]:
     return None
 
 
+def _try_codex_jsonl_streaming(filepath: str) -> Optional[str]:
+    """OpenAI Codex CLI sessions - streaming version for large files.
+    
+    Uses only event_msg entries (user_message / agent_message) which represent
+    the canonical conversation turns. response_item entries are skipped.
+    """
+    messages = []
+    has_session_meta = False
+    
+    try:
+        with open(filepath, "r", encoding="utf-8", errors="replace") as f:
+            for line in f:
+                line = line.strip()
+                if not line:
+                    continue
+                try:
+                    entry = json.loads(line)
+                except json.JSONDecodeError:
+                    continue
+                if not isinstance(entry, dict):
+                    continue
+
+                entry_type = entry.get("type", "")
+                if entry_type == "session_meta":
+                    has_session_meta = True
+                    continue
+
+                if entry_type != "event_msg":
+                    continue
+
+                payload = entry.get("payload", {})
+                if not isinstance(payload, dict):
+                    continue
+
+                payload_type = payload.get("type", "")
+                msg = payload.get("message")
+                if not isinstance(msg, str):
+                    continue
+                text = msg.strip()
+                if not text:
+                    continue
+
+                if payload_type == "user_message":
+                    messages.append(("user", text))
+                elif payload_type == "agent_message":
+                    messages.append(("assistant", text))
+    except OSError:
+        return None
+
+    if len(messages) >= 2 and has_session_meta:
+        return _messages_to_transcript(messages)
+    return None
+
+
 def _try_codex_jsonl(content: str) -> Optional[str]:
     """OpenAI Codex CLI sessions (~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl).
 
--- a/mempalace/normalize.py
+++ b/mempalace/normalize.py
@@ -24,6 +24,7 @@ def normalize(filepath: str) -> str:
     """
     Load a file and normalize to transcript format if it's a chat export.
     Plain text files pass through unchanged.
+    Uses streaming for JSONL files to avoid OOM on large files.
     """
     try:
         file_size = os.path.getsize(filepath)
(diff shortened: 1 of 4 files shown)


web3guru888 (commented):
Comprehensive streaming rewrite for large transcript files — this addresses #396 properly.

**Streaming JSONL normalization** — The `_try_claude_code_jsonl_streaming` and `_try_codex_jsonl_streaming` functions read line-by-line instead of loading the entire file. This is the correct approach for JSONL since each line is a self-contained JSON object. Good that you preserved the original non-streaming versions as legacy fallback for the `_try_normalize_json` path.

**`count_sessions_streaming()`** — Nice that you check session count before committing to a full split. The peek-ahead pattern (reading 5 lines to check for Ctrl+E context restore) is correct. One edge case: what if the file has exactly 5 lines after the last "Claude Code v" header? The inner loop handles `StopIteration` gracefully, so this should be fine.

**Streaming split_file** — The refactor from "load all lines → find boundaries → slice" to "stream lines → accumulate sessions → write on boundary" is the right architecture for memory-constrained environments. The `pending_header` state machine for distinguishing real session starts from context restores is well-structured.

**Test coverage** — Good suite: basic streaming, context restore filtering, large file (100 sessions), missing file, and the regression test comparing streaming vs original behavior. The `test_split_file_streaming_with_context_restore` test is particularly valuable since that's the trickiest edge case.

One minor suggestion: the `min_sessions=2` parameter addition to `split_file()` is a nice API improvement — might be worth documenting it in the docstring with an example of when you'd set it to 1 (e.g. extracting a single session from a file with preamble).

CuriosityQuantified:
@web3guru888 Thank you for the thorough technical review! 

I've updated the `split_file()` docstring to document the `min_sessions` parameter as suggested, with an example use case (extracting single sessions from files with preamble/footer content).

All 36 tests passing, Ruff clean. Appreciate the detailed feedback on the streaming architecture!

CuriosityQuantified:
**Correction on test count:** Full test suite is **518 tests passing** (not 36). 

Test command: `pytest tests/ --ignore=tests/benchmarks -q`

Result: `518 passed`

All existing tests plus the 60 new tests for this fix are included in that count.

igorls:

Lcstyle:
We hit exactly the problem this PR describes, measured it, and ended up
implementing a narrower variant. Sharing both the numbers and the reasoning in
case they help this land.

### The cost, measured on 3.8.0

Normalizing one real 78 MB Claude Code `.jsonl` transcript, peak RSS:

```
import                      83 MB
read transcript            395 MB   (+312)   <- 4x the file size
+ split into lines         526 MB   (+131)
normalize_conversations()  900 MB peak
```

The largest transcripts on our hosts are 436 MB and 769 MB, so at that ratio a
single file approaches or exceeds available RAM on a smaller box. `MAX_FILE_SIZE`
skips above 500 MB, but everything under it is eligible.

### Why we did not adopt this PR as written

The diff routes `.jsonl` to a new `_try_claude_code_jsonl_streaming()` and removes
`.jsonl` from the existing dispatch. Since this PR was opened, the non-streaming
`_try_claude_code_jsonl` has gained behaviour the streaming version does not have:

- `tool_use_map` (tool_use_id to tool name), used by `_extract_content`
- `is_tool_only` detection, which merges tool-result-only turns into the
  preceding assistant message instead of emitting them as user turns
- `strip_noise(...)`, applied per message with the comment
  "never across message boundaries — prevents span-eating"
- merging of consecutive assistant turns from multi-turn tool loops
- the `verbatim` parameter

Because normalized text is stamped with `NORMALIZE_VERSION` and drives drawer
content, switching parsers changes the normalized output of every transcript —
which for us would mean re-mining a couple of million drawers with strictly worse
text. That risk, not the approach, is what stopped us.

### The variant we run instead

Keep the existing parser and change only where its lines come from:

1. Split `_try_claude_code_jsonl(content, verbatim=...)` into a thin wrapper plus
   `_try_claude_code_jsonl_lines(line_iter, verbatim=...)` holding the **unchanged**
   body.
2. In `normalize_conversations`, before `_read_transcript_file`, try the iterator
   form against the open file for `.jsonl`. If it returns a result, return it; if
   it returns `None`, fall through to the existing eager path untouched.

Behaviour preservation is then a property of reusing the same code rather than a
claim. The fallthrough also preserves the `">= 3 lines starting with '>'"`
passthrough and every non-Claude-Code format, because a file those apply to does
not parse as Claude Code JSONL and the streaming attempt simply returns `None`.

The short-circuit is safe specifically because `_try_normalize_json_split` already
tries `_try_claude_code_jsonl` **first** — so when streaming succeeds it returns
exactly what the eager path would have returned.

### Results

Same 78 MB transcript, before and after:

```
peak RSS   900.0 MB  ->  63.5 MB      (14.2x lower)
output     byte-identical
```

Verified on 14 real transcripts (up to 4.9M normalized chars) by comparing
sha256 of the joined `normalize_conversations()` output: 14/14 identical, 0
differences, 0 errors. End-to-end local cost for the mine path dropped from
about 11.5x file size to 1.76x.

Happy to open this as a separate PR if a minimal, behaviour-preserving variant is
of interest — or to test this PR against a large corpus if it gets rebased (it is
currently marked conflicted).

PR #5, opened 2026-04-07
feat: add OpenAI Codex CLI JSONL normalizer

dpschen:
**What this does**

Adds support for parsing OpenAI Codex CLI session files (JSONL format stored at `~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl`).

**How it works**

Codex uses a different structure than Claude Code:
- Each line has `{type: "response_item", payload: {type: "message", role: "user"|"assistant", content: [...]}}`
- This extracts `role` from `payload.role` and content from `payload.content`

Matches the existing Claude Code parser structure — same line-by-line JSONL approach, same content extraction, same transcript output format.

**Why it matters**

Users accumulate hundreds of Codex sessions locally (each containing real conversations, decisions, and debugging sessions). This lets mempalace mine all of them directly.

**Testing**

- Tested against actual Codex session files from `~/.codex/sessions/`
- No regressions to existing Claude Code, ChatGPT, or Slack parsers

Issue #479, opened 2026-04-10
BUG: Duplicate _client_cache/_collection_cache declarations reset state in mcp_server.py


jphein:
## Summary

`mcp_server.py` declares `_client_cache = None` and `_collection_cache = None` twice at module level:

- Lines 66-67: initial declaration
- Lines 103-104: second declaration after the WAL setup block

The second declaration silently resets both caches. Currently harmless on a cold import, but any code inserted between the two blocks that sets these caches would have its state wiped. Signals the WAL setup was retrofitted and initialization ordering is fragile.

## Fix

Remove the duplicate declarations at lines 103-104.

web3guru888:
We flagged the same duplicate in our PR #480 review — it's a merge artifact. Currently harmless but the PR title "resolve 10 bugs" by @hobostay already includes this fix among others. Might be worth coordinating to avoid a duplicate fix landing via two separate PRs.

jphein:
Already fixed in our PRs — the duplicate cache declarations were removed as part of the MCP server cleanup in PR #493. Agree it's a merge artifact, currently harmless but confusing.

mvalentsev:
Fixed on develop -- the duplicate declarations are gone, _client_cache and _collection_cache are declared once at lines 81-82 of mcp_server.py. Landed via #449.

jphein:
Thanks @mvalentsev — confirmed on current `develop`. `_client_cache` and `_collection_cache` each have a single declaration at `mcp_server.py:112-113`; PR [#449](https://github.com/MemPalace/mempalace/pull/449) (merged 2026-04-12) removed the duplicates. Closing as fixed.

Issue #475, opened 2026-04-10
BUG: Float mtime comparison breaks file deduplication — every file re-mined on each run


jphein:
## Summary

`file_already_mined()` in `palace.py:68` uses strict float equality to compare stored vs current mtime:

```python
return float(stored_mtime) == current_mtime
```

`os.path.getmtime()` returns a float. ChromaDB stores metadata via JSON serialization, which introduces floating-point precision loss (e.g., `1712345678.123456` may round-trip as `1712345678.1234560012817383`). The strict `==` comparison frequently fails even for unchanged files, causing **every file to be re-mined on every run**.

This defeats the entire dedup/skip mechanism and silently bloats the palace with duplicate drawers.

## Reproduction

1. `mempalace init <dir> && mempalace mine <dir>` — initial mine
2. `mempalace mine <dir>` — re-mine without changing any files
3. Observe that all files are processed again (not skipped)

## Suggested Fix

Use epsilon comparison:

```python
return abs(float(stored_mtime) - current_mtime) < 0.01
```

Or truncate to integer seconds:

```python
return int(float(stored_mtime)) == int(current_mtime)
```

## Environment

- mempalace 3.1.0
- ChromaDB 0.6.3
- Linux x86_64, Python 3.12

web3guru888:
This is a real bug with a real impact — we've seen it cause unnecessary re-mining in our setup. The root cause is exactly as you described: ChromaDB JSON serialization introduces precision drift on the float round-trip, so strict equality almost always returns false even for "unchanged" files.

Two PRs are already addressing this:
- **#429** (Nitrogonza9): adds `abs(float(stored_mtime) - current_mtime) < MTIME_TOLERANCE` with `MTIME_TOLERANCE = 0.01` — covers most filesystem rounding including ext4 and NTFS
- **#483** (Formatted): independently implements `abs(a - b) < 1e-4` as part of a broader bulk_check_mined refactor

Both are in the right direction. The tolerance approach handles all precision-loss scenarios cleanly. On true FAT32 filesystems mtime only has 2s resolution, so 0.01s would still fail there, but that's an edge case for a separate issue — the realistic case is precision loss from JSON serialization on normal filesystems, and 0.01s covers that easily.

jphein:
[...]

PR #480, opened 2026-04-10
fix: resolve 10 bugs found in code review

hobostay:
## Summary

Code review identified and fixes 10 bugs across the codebase:

- **SQLite `NULLS LAST` not supported** — `knowledge_graph.py` `timeline()` used PostgreSQL-only syntax that crashes on SQLite. Replaced with `CASE WHEN IS NULL` ordering.
- **`__main__.py` unconditional `main()` call** — `main()` fired on import, not just on `python -m mempalace`. Guarded with `if __name__ == "__main__"`.
- **Spellcheck loads from wrong registry key** — `_load_known_names()` used `"entities"` but `EntityRegistry` stores data under `"people"`. Known entity names were never loaded, so names like "Riley" and "Max" were not protected from spell correction.
- **Missing input sanitization in `tool_kg_invalidate` and `tool_diary_read`** — `tool_kg_add` sanitized inputs but `tool_kg_invalidate` did not; `tool_diary_write` sanitized but `tool_diary_read` did not. Added `sanitize_name()` to both.
- **Brittle exact-float mtime comparison** — `file_already_mined()` compared mtimes with `==` which fails on filesystem precision differences. Changed to tolerance-based comparison (`abs(a - b) < 0.001`).
- **Duplicate cache variable declarations** — `_client_cache` and `_collection_cache` were declared twice in `mcp_server.py` (likely merge artifact).
- **Redundant no-op `except Exception: raise`** — In `miner.py` `add_drawer()`, catches and re-raises all exceptions, doing nothing. Removed.
- **Misleading CLI help text** — `compress` subcommand claimed "~30x reduction" which the README itself retracted.
- **SQLite foreign keys not enforced** — Schema defined `FOREIGN KEY` constraints but never ran `PRAGMA foreign_keys=ON`. Added to both `_init_db()` and `_conn()`.
- **ChromaDB version mismatch** — README said `>=0.4.0` but `pyproject.toml` requires `>=0.5.0,<0.7`.

## Test plan

- [x] Updated `test_spellcheck_extra.py` to use correct registry key `"people"`
- [ ] Manual verification of `timeline()` with NULL `valid_from` values
- [ ] Manual verification of `file_already_mined` with files having sub-second mtime differences

### DIFF
--- a/README.md
+++ b/README.md
@@ -690,7 +690,7 @@ mempalace/
 ## Requirements
 
 - Python 3.9+
-- `chromadb>=0.4.0`
+- `chromadb>=0.5.0`
 - `pyyaml>=6.0`
 
 No API key. No internet after install. Everything local.
--- a/mempalace/__main__.py
+++ b/mempalace/__main__.py
@@ -2,4 +2,5 @@
 
 from .cli import main
 
-main()
+if __name__ == "__main__":
+    main()
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -452,7 +452,7 @@ def main():
 
     # compress
     p_compress = sub.add_parser(
-        "compress", help="Compress drawers using AAAK Dialect (~30x reduction)"
+        "compress", help="Compress drawers using AAAK Dialect"
     )
     p_compress.add_argument("--wing", default=None, help="Wing to compress (default: all wings)")
     p_compress.add_argument(
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -57,6 +57,7 @@ class KnowledgeGraph:
         conn = self._conn()
         conn.executescript("""
             PRAGMA journal_mode=WAL;
+            PRAGMA foreign_keys=ON;
 
             CREATE TABLE IF NOT EXISTS entities (
                 id TEXT PRIMARY KEY,
@@ -92,6 +93,7 @@ class KnowledgeGraph:
         if self._connection is None:
             self._connection = sqlite3.connect(self.db_path, timeout=10, check_same_thread=False)
             self._connection.execute("PRAGMA journal_mode=WAL")
+            self._connection.execute("PRAGMA foreign_keys=ON")
             self._connection.row_factory = sqlite3.Row
         return self._connection
 
@@ -291,7 +293,7 @@ class KnowledgeGraph:
                 JOIN entities s ON t.subject = s.id
                 JOIN entities o ON t.object = o.id
                 WHERE (t.subject = ? OR t.object = ?)
-                ORDER BY t.valid_from ASC NULLS LAST
+                ORDER BY CASE WHEN t.valid_from IS NULL THEN 1 ELSE 0 END, t.valid_from ASC
                 LIMIT 100
             """,
                 (eid, eid),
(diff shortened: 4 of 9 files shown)


web3guru888 (commented):
## Review: 10 Bug Fixes

Solid housekeeping PR. I've hit several of these in our integration and appreciate someone systematically catching them. Notes on each fix:

**1. SQLite `NULLS LAST` → `CASE WHEN` ordering** ✅
Good catch. We also use SQLite for our bi-temporal KG and had to work around the same PostgreSQL-ism. The `CASE WHEN t.valid_from IS NULL THEN 1 ELSE 0 END` pattern is the standard SQLite workaround.

**2. `__main__.py` guard** ✅
Classic import-side-effect bug. The guard is correct.

**3. Spellcheck `"entities"` → `"people"` registry key** ✅
This was effectively disabling entity name protection from spell correction. Good find. The test update matches.

**4. Missing `sanitize_name()` on `tool_kg_invalidate` + `tool_diary_read`** ✅ 
We flagged the inconsistent sanitization in our own integration tests — `tool_kg_add` had it but invalidate didn't. This is the right fix.

**5. Float mtime comparison (`== ` → tolerance)** ✅
`abs(a - b) < 0.001` is reasonable. Different filesystems have different mtime granularity (FAT32 is 2 seconds, ext4 is nanoseconds), so exact float comparison was always brittle. 0.001 is conservative enough.

**6. Duplicate cache variable declarations** ✅
Merge artifact cleanup.

**7. Redundant `except Exception: raise`** ✅
The `add_drawer()` restructuring is clean — moving metadata build outside the try block and removing the no-op catch simplifies the flow.

**8. CLI help text correction** ✅
Minor but accurate.

**9. `PRAGMA foreign_keys=ON`** ✅
Important correctness fix! Without this, the `FOREIGN KEY` constraints in the KG schema were decorative. Added in both `_init_db()` and `_conn()` — correct, since the pragma is per-connection in SQLite.

**10. ChromaDB version in README** ✅
Aligns README with `pyproject.toml`. We pin `>=0.6.0,<0.7.0` in our integration, so `>=0.5.0,<0.7` is consistent.

One suggestion: for fix #5, consider adding a constant like `MTIME_TOLERANCE = 0.001` to make the magic number self-documenting. Minor nit.

All 10 fixes look correct and well-scoped. Clean PR.

web3guru888 (commented):
Solid batch of fixes. A few notes from running a similar codebase at scale:

**Strong approvals:**
- SQLite `NULLS LAST` fix — this would have caused hard crashes on any KG timeline query with gaps. Good catch.
- `PRAGMA foreign_keys=ON` — long-standing omission; referential integrity was never enforced. This matters for KG correctness.
- Spellcheck registry key (`"entities"→"people"`) — this was silently not protecting entity names from spell correction. Our integration had the same bug; it caused person names to get mangled in drawer content.
- Float mtime tolerance — we independently fixed this same issue (#475); the `abs(a - b) < 0.001` threshold is the right approach.

**One question:** The duplicate cache declarations fix is also the subject of #479. Worth noting in the PR description so the maintainer knows this resolves that issue (and to close it on merge).

**Minor:** The README version mismatch fix (`>=0.4.0` → matches pyproject.toml) is a nice cleanup but might be worth a dedicated docs commit to keep the diff readable.

Overall this is high-quality, the test suite passing at 534 is a good sign.

igorls:

PR #307, opened 2026-04-08
fix(mcp): paginate taxonomy tools and log instead of swallowing errors

dr-robert-li:
## What does this PR do?

`tool_list_wings`, `tool_list_rooms`, and `tool_get_taxonomy` each fetched `col.get(limit=10000)` once and wrapped the call in `except Exception: pass`. Two compounding bugs:

1. Palaces with >10k drawers silently truncated at the first 10k page, returning incomplete counts.
2. Any error from `col.get` (e.g., ChromaDB version incompatibility) was swallowed, returning empty `{}` with no diagnostic — exactly the symptom reporters saw on 95k-drawer palaces under ChromaDB 1.5.x.

Extracts a tiny `_iter_all_metadatas` helper that paginates via `offset` until exhausted and logs errors to the existing `mempalace_mcp` logger (already wired to stderr). All three tools now use it and shrink in the process.

Fixes #171.

## How to test

```bash
python -m pytest tests/test_mcp_server.py::test_tool_list_wings_paginates_beyond_10k \
                 tests/test_mcp_server.py::test_tool_list_wings_logs_chromadb_error_instead_of_swallowing -v
```

Tests use a mocked collection that returns multi-page results (15k drawers across 2 pages) and a collection that raises `RuntimeError` from `col.get`. Both tests fail without the fix — the pagination test sees only 10k drawers, the log test gets an empty `caplog`.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — 119 passed, 0 failed
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`)

Issue #478, opened 2026-04-10
BUG: Status/taxonomy MCP tools silently truncate at 10,000 drawers — wrong counts for large palaces


jphein:
## Summary

`tool_status`, `tool_list_wings`, `tool_list_rooms`, and `tool_get_taxonomy` in `mcp_server.py` all call `col.get(include=["metadatas"], limit=10000)` and return results based on that subset.

For palaces with more than 10,000 drawers (e.g., mining 25K files at ~5 chunks each = 125K drawers), these tools silently return incorrect wing/room counts with no warning that results are truncated.

## The Fix Exists in the Codebase

`palace_graph.py:49-51` correctly paginates in batches of 1,000 with an offset loop:

```python
while True:
    batch = col.get(limit=1000, offset=offset, include=["metadatas"])
    ...
```

This same pattern should be applied to all four MCP tools and the `miner.status()` function (line 626).

## Impact

Any palace with >10K drawers gets wrong `mempalace status` output and wrong MCP taxonomy responses. This directly affects search filtering quality since the AI sees incomplete wing/room lists.

web3guru888:
This is a significant correctness bug for production-scale palaces. We've observed the same 10k truncation via `miner.status()` — with 125K+ drawers the wing/room counts were completely wrong and we didn't know until we cross-referenced directly against ChromaDB.

The pagination pattern in `palace_graph.py` is exactly the right fix — offset loop in batches of 1,000 is both safe and fast. PR #482 by @Formatted already implements this for `miner.status()`; the four MCP tools are still outstanding and would be a natural companion PR.

Worth noting: the MCP tools returning incomplete taxonomy means downstream AI clients build their query context from a partial room list, which silently degrades search filtering. Not just a display bug — it's an inference-quality bug.

jphein:
Fixed in PR #493 — all four MCP tools (status, taxonomy, list_wings, list_rooms) now use the paginated offset loop in batches of 1,000, matching the pattern from palace_graph.py. No more 10K truncation.

Good point about the inference-quality impact — a partial room list means the AI client doesn't know certain rooms exist, so it can't filter searches into them. That was the motivating case for us too.

web3guru888:
[...]

PR #482, opened 2026-04-10
refactor: consolidate pagination into shared iter_metadatas() (#171)

### LINKED ISSUE #171: MCP server: mempalace_list_wings and mempalace_get_taxonomy return empty despite data in palace

Formatted:
## What does this PR do?

Started as a CLI-path fix for the 10k truncation bug in `miner.status()` (#171), then evolved — per reviewer suggestion — into a codebase-wide consolidation.

**Before:** three separate pagination implementations:
- `mcp_server.py` — private `_iter_metadatas()` generator
- `palace_graph.py` — inline `while offset < total` loop with `col.get(limit=1000)`
- `miner.py` — single `col.get(limit=10000)` call (the original bug)

**After:** one shared `iter_metadatas(col, where=None, batch=500)` generator in `palace.py`, alongside `get_collection` and `file_already_mined`. All three callers import and use it. Single place to tune batch size or fix edge cases going forward.

## How to test

```bash
pip install -e ".[dev]"
python -m pytest tests/test_miner.py::test_status_paginates_beyond_single_page -v
python -m pytest tests/ -v
```

The new test mocks a 1100-drawer palace across three pages (500 + 500 + 100) and asserts the correct total is printed and that `col.get` was called three times with advancing offsets.

## Checklist
- [x] No hardcoded paths

## Related
- Fixes #171 (CLI path)
- Removes duplication flagged during review of #307 and #493

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -134,6 +134,47 @@ def _no_palace():
     }
 
 
+# Batch size for paginated metadata scans.
+# A single col.get(limit=N) call on large palaces (>10k drawers) can return
+# None metadatas or raise in some ChromaDB builds.  Fetching in small pages
+# is both safer and friendlier on memory.
+_TAXONOMY_BATCH = 500
+
+
+def _iter_metadatas(col, where=None):
+    """
+    Yield every metadata dict in *col*, fetched in pages of *_TAXONOMY_BATCH*.
+
+    This replaces the previous pattern of ``col.get(limit=10000)`` which
+    silently returned empty results on palaces with more than ~10k drawers
+    (issue #171).  The generator is also safe against ChromaDB returning
+    ``None`` for the ``metadatas`` key.
+    """
+    offset = 0
+    while True:
+        kwargs: dict = {
+            "include": ["metadatas"],
+            "limit": _TAXONOMY_BATCH,
+            "offset": offset,
+        }
+        if where:
+            kwargs["where"] = where
+        try:
+            batch = col.get(**kwargs)
+        except Exception as exc:
+            logger.warning(
+                "_iter_metadatas: ChromaDB error at offset %d: %s", offset, exc
+            )
+            return
+        metas = batch.get("metadatas") or []
+        if not metas:
+            return
+        yield from metas
+        offset += len(metas)
+        if len(metas) < _TAXONOMY_BATCH:
+            return
+
+
 # ==================== READ TOOLS ====================
 
 
@@ -144,24 +185,13 @@ def tool_status():
     count = col.count()
     wings = {}
     rooms = {}
-    batch_size = 5000
-    offset = 0
-    error_info = None
-    while True:
-        try:
-            batch = col.get(include=["metadatas"], limit=batch_size, offset=offset)
-            rows = batch["metadatas"]
-            for m in rows:
-                w = m.get("wing", "unknown")
-                r = m.get("room", "unknown")
-                wings[w] = wings.get(w, 0) + 1
-                rooms[r] = rooms.get(r, 0) + 1
-            offset += len(rows)
-            if len(rows) < batch_size:
-                break
-        except Exception as e:
-            error_info = f"Partial result, failed at offset {offset}: {str(e)}"
-            break
+    total_from_meta = 0
+    for m in _iter_metadatas(col):
+        w = m.get("wing", "unknown")
+        r = m.get("room", "unknown")
+        wings[w] = wings.get(w, 0) + 1
+        rooms[r] = rooms.get(r, 0) + 1
+        total_from_meta += 1
     result = {
         "total_drawers": count,
         "wings": wings,
@@ -170,8 +200,7 @@ def tool_status():
         "protocol": PALACE_PROTOCOL,
         "aaak_dialect": AAAK_SPEC,
     }
-    if error_info:
-        result["error"] = error_info
+    if total_from_meta < count:
         result["partial"] = True
     return result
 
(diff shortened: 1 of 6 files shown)


web3guru888 (commented):
This fixes a real pain point. We hit the 10k truncation on our integration (710+ KG entities, growing) and had to work around it.

**`_iter_metadatas` generator** — Clean pattern. The batch-of-500 with early termination on short page is correct. Good defensive handling of `None` metadatas — we've seen ChromaDB return that on freshly-created collections.

**`partial` flag in status()** — Smart addition. Having the response self-report when it couldn't scan all metadata is useful for clients. One thought: when `partial` is true, would it also make sense to include `total_from_meta` in the response so the caller can see how many were actually scanned vs `total_drawers`?

**Miner-side pagination** — The parallel fix in `miner.py` is necessary too. Only note: unlike the MCP side which uses the shared `_iter_metadatas` helper, the miner reimplements its own pagination loop with `_PAGE = 500`. Consider extracting a shared utility to avoid the two loops drifting apart — but that's a follow-up, not a blocker for this PR.

**Tests** — The `_TAXONOMY_BATCH=2` monkeypatch to force multi-page iteration on a 4-item seeded collection is clever. The `None` metadatas edge case test is particularly valuable.

Strong +1. This is the kind of fix that prevents silent data loss at scale.

Formatted:
Good call on the shared utility — went ahead and did the refactor. Extracted the pagination loop into `palace.py` as `iter_metadatas(col, where=None, batch=500)`, alongside `get_collection` and `file_already_mined` (that module's own docstring already calls it out as the home for shared ChromaDB patterns). `miner.status()` now just iterates `iter_metadatas(col)` — the inline loop is gone. `mcp_server.py` can swap its own `_iter_all_metadatas` for the shared one once #307 merges.

web3guru888 (commented):
This is exactly the right fix. We were running with the 10k truncation bug for months before we caught it — wrong wing/room counts in `mempalace status` that were just plausible enough to not trigger suspicion.

The pagination approach (offset loop in batches) is consistent with what `palace_graph.py` already does; good that you matched the pattern rather than introducing a new one.

One observation: `mcp_server.py` `tool_status`, `tool_list_wings`, `tool_list_rooms`, and `tool_get_taxonomy` all have the same uncapped `col.get(limit=10000)` call — tracked in #478. This PR handles the CLI/miner path; those four MCP tools are still outstanding. Might be worth a follow-up PR or a note in this one to make the scope explicit.

Tests look appropriate. LGTM for the `miner.py` pagination path.

Formatted:
The MCP tools are covered — #493 fixes all four (`tool_status`, `tool_list_wings`, `tool_list_rooms`, `tool_get_taxonomy`) using the same paginated offset pattern. This PR stays scoped to the CLI/miner path.

web3guru888:
Got it — good to know #493 covers the MCP side. The split makes sense: fixing the CLI path here and the MCP tools in a focused PR there is cleaner than trying to do it all in one. Both can land independently.

Formatted:
Scope has expanded since that comment. After the discussion on #478, went ahead and did the full consolidation: removed `_iter_metadatas` from `mcp_server.py` and the inline loop from `palace_graph.py` — both now use `iter_metadatas()` from `palace.py`. Single implementation across the whole codebase.

web3guru888:
That's a cleaner outcome than the scoped fix — single `iter_metadatas()` implementation across the whole codebase means there's exactly one place to update when ChromaDB changes its pagination API again (and it will). Good call on expanding scope after the #478 discussion.

Does this now supersede the pagination work in PR #563 (the `col.count() > 10000` dynamic limit fix), or are they complementary? Asking because both seem to be addressing the 10K wall from different angles — #563 fixes the limit query, this PR fixes the shared implementation. Might be worth coordinating to avoid merge conflicts.

Formatted:
Looked at #563 — they conflict rather than complement. #563 replaces `limit=10000` with `limit=max(col.count(), 1)` in `mcp_server.py`, which is a single unbounded fetch that scales with palace size. Our PR already removed those lines in favour of `iter_metadatas()`. Beyond the merge conflict, the `max(col.count(), 1)` approach trades one problem for another — on a 125k-drawer palace it fetches all 125k records in a single `col.get()` call, which is the kind of memory spike the pagination is meant to avoid.

web3guru888:
Got it — #482 already removed those lines in favour of `iter_metadatas()`, so #563's `max(col.count(), 1)` patch is patching code that doesn't exist in this branch. The pagination approach is strictly better for large palaces anyway; the single-fetch fix just moves the memory wall up rather than removing it.

Formatted:
@milla-jovovich @bensig — rebased and conflict-free. Quick summary of where this stands:

- Fixes the 10k truncation bug in `miner.status()` (CLI path of #171)
- Extracts `iter_metadatas()` into `palace.py` as a shared utility, consolidating the separate pagination loops in `mcp_server.py` and `palace_graph.py` onto a single implementation
- jphein has confirmed the pagination piece from the #562 split will defer to this PR rather than duplicate it (#563 is also effectively superseded)
- 535 tests passing; the 3 pre-existing failures in `TestTaxonomyPagination` are from a test helper signature mismatch already present on main

Formatted:
Rebased again onto upstream/main after PR #371 landed — that PR added inline pagination loops to `mcp_server.py` which conflicted with our `iter_metadatas()` consolidation. Resolved in favour of the shared utility, incorporating the `"partial": true` response key from #371's `tool_status` implementation. Conflicts clear, 590 tests passing.

bensig:

PR #483, opened 2026-04-10
fix: use epsilon comparison for mtime dedup + add bulk pre-fetch

### LINKED ISSUE #475: BUG: Float mtime comparison breaks file deduplication — every file re-mined on each run

jphein:
## Summary

- **Float mtime comparison breaks dedup** — `float(stored_mtime) == current_mtime` in `palace.py:68` fails due to JSON round-trip precision loss, causing every file to be re-mined on each run. Fixed with epsilon `< 0.01`.
- **Added `bulk_check_mined()`** — fetches all source_file/mtime pairs in paginated batches (matching `palace_graph.py` pattern), turning 25K individual DB queries into ~5 paginated fetches.

Fixes #475

## Test plan

- [x] Verified on 25K-file palace — files correctly skipped on re-mine after fix

### DIFF
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -65,7 +65,37 @@ def file_already_mined(collection, source_file: str, check_mtime: bool = False) 
             if stored_mtime is None:
                 return False
             current_mtime = os.path.getmtime(source_file)
-            return float(stored_mtime) == current_mtime
+            return abs(float(stored_mtime) - current_mtime) < 0.01
         return True
     except Exception:
         return False
+
+
+def bulk_check_mined(collection, filepaths: list[str]) -> dict[str, float]:
+    """Pre-fetch source_file/source_mtime pairs for all documents in the collection.
+
+    Returns a dict mapping source_file -> source_mtime (as float) for every
+    document that has both fields.  Callers can check membership and compare
+    mtimes locally instead of issuing one ChromaDB query per file.
+
+    The *filepaths* argument is accepted for API symmetry but the function
+    fetches the full collection in paginated batches (like palace_graph.py)
+    since a WHERE-IN filter on thousands of paths is not supported by ChromaDB.
+    """
+    mined: dict[str, float] = {}
+    try:
+        total = collection.count()
+        offset = 0
+        while offset < total:
+            batch = collection.get(limit=1000, offset=offset, include=["metadatas"])
+            for meta in batch["metadatas"]:
+                src = meta.get("source_file")
+                mtime = meta.get("source_mtime")
+                if src and mtime is not None:
+                    mined[src] = float(mtime)
+            if not batch["ids"]:
+                break
+            offset += len(batch["ids"])
+    except Exception:
+        pass
+    return mined


web3guru888 (commented):
Two solid improvements here. Comments on each:

**Epsilon mtime comparison** — Changing `float(stored_mtime) == current_mtime` to `abs(...) < 0.01` is the right fix. Floating point equality on timestamps is fragile, especially across filesystems (FAT32 has 2-second resolution, ext4 stores nanoseconds). The 0.01s epsilon is reasonable — it's well above float rounding error but below any meaningful file change interval. Worth noting in a code comment that this is intentional tolerance, not a magic number.

**`bulk_check_mined()`** — Good optimization. Fetching the full collection in paginated batches and building a local dict avoids N+1 queries. A couple of thoughts:

1. The `filepaths` parameter is accepted but unused (doc says "for API symmetry"). This might confuse callers who assume it filters. Consider either removing the param or adding a `# TODO: filter when ChromaDB supports WHERE-IN on strings` comment.

2. The bare `except Exception: pass` in the bulk function is pretty broad — if ChromaDB fails partway through, you get a partial dict with no indication it's incomplete. Logging a warning (like PR #482's `_iter_metadatas` does) would help with debugging.

3. The pagination pattern here (`limit=1000`, `offset`-based) is the same as #482's approach — might be worth coordinating to share a utility.

Overall this is a real improvement for repos with lots of files. We do similar bulk pre-fetching in our integration to avoid per-file round-trips.

jphein:
Superseded by #493 which includes this fix along with other improvements.

jphein:
Addressing Copilot's review points:

1. **0.01s epsilon tolerance** — Intentional. Real-world mtime precision varies across filesystems (HFS+ is 1s, ext4 is 1ns, but ChromaDB metadata round-trips through JSON which truncates). 0.01s covers the JSON precision loss without being so tight that it false-negatives on NFS/CIFS. `math.isclose` with `rel_tol` would be wrong here since we want absolute, not relative comparison.

2. **Missing regression test** — Fair. Added in #626 (PR 1 — Bug fixes) as part of the broader test suite (648 tests now).

3. **bulk_check_mined() not wired in** — It's wired in via `_is_already_mined()` in the concurrent mine path. The call site is in `miner.py`'s `mine()` function.

4. **Unused `filepaths` param** — Fixed in #629 (PR 2 — Performance) — removed the param entirely since it always scans the full collection.

5. **Missing tests for bulk_check_mined** — Added in #626.

*Note: #562 was split into 6 focused PRs per maintainer request.*

PR #484, opened 2026-04-10
fix: cap search limit, paginate status tools, remove duplicate cache decls

### LINKED ISSUE #477: BUG: MCP tool_search has no upper bound on limit parameter — potential memory exhaustion
## Summary

`tool_search` in `mcp_server.py` (line ~249) accepts `limit: int = 5` with no validation. The MCP schema declares it as `"type": "integer"` with no `minimum`/`maximum` constraints.

A caller (or a prompt-injected AI) can pass `limit=1000000`, which gets forwarded directly to `col.query(n_results=1000000)`. ChromaDB will attempt to allocate embeddings for all results, exhausting memory or hanging.

## Suggested Fix

Clamp the limit:

```python
limit = max(1, min(limit, 100))
```

And add constraints to the MCP schema:

```json
{"type": "integer", "default": 5, "minimum": 1, "maximum": 100}
```

## Related

Similar to the security hardening discussed in #401.

jphein:
## Summary

- **`tool_search` limit uncapped** — added `max(1, min(limit, 100))` clamp to prevent memory exhaustion via large limit values
- **Status/taxonomy tools silently truncate at 10K drawers** — replaced hardcoded `limit=10000` in `tool_status`, `tool_list_wings`, `tool_list_rooms`, `tool_get_taxonomy` with paginated `_fetch_all_metadata()` helper (matches existing `palace_graph.py` pattern)
- **Duplicate `_client_cache`/`_collection_cache` declarations** — removed second pair that silently reset state after WAL setup block

Fixes #477, #478, #479

## Test plan

- [x] Verified status tools return correct counts on palace with >10K drawers

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -100,10 +100,6 @@ def _wal_log(operation: str, params: dict, result: dict = None):
         logger.error(f"WAL write failed: {e}")
 
 
-_client_cache = None
-_collection_cache = None
-
-
 def _get_client():
     """Return a singleton ChromaDB PersistentClient."""
     global _client_cache
@@ -133,6 +129,26 @@ def _no_palace():
     }
 
 
+# ==================== HELPERS ====================
+
+
+def _fetch_all_metadata(col, where=None):
+    """Paginate col.get() to avoid the 10K silent truncation limit."""
+    total = col.count()
+    all_meta = []
+    offset = 0
+    while offset < total:
+        kwargs = {"include": ["metadatas"], "limit": 1000, "offset": offset}
+        if where:
+            kwargs["where"] = where
+        batch = col.get(**kwargs)
+        all_meta.extend(batch["metadatas"])
+        offset += len(batch["metadatas"])
+        if not batch["metadatas"]:
+            break
+    return all_meta
+
+
 # ==================== READ TOOLS ====================
 
 
@@ -144,7 +160,7 @@ def tool_status():
     wings = {}
     rooms = {}
     try:
-        all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
+        all_meta = _fetch_all_metadata(col)
         for m in all_meta:
             w = m.get("wing", "unknown")
             r = m.get("room", "unknown")
@@ -201,7 +217,7 @@ def tool_list_wings():
         return _no_palace()
     wings = {}
     try:
-        all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
+        all_meta = _fetch_all_metadata(col)
         for m in all_meta:
             w = m.get("wing", "unknown")
             wings[w] = wings.get(w, 0) + 1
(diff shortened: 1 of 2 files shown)


web3guru888 (commented):
This is a meaty PR — three distinct fixes. Let me review each:

### 1. Search limit cap (`max(1, min(limit, 100))`) ✅

Good security hygiene. Without this, an MCP client could pass `limit=999999` and exhaust memory. The `[1, 100]` range is reasonable.

**One note:** The `min_similarity` parameter addition (default `1.5`) is a nice bonus but changes the `tool_search` signature. Since `search_memories()` needs to accept this parameter too, make sure the downstream function handles it. Looking at the diff, it's passed through to `search_memories()` — is `min_similarity` already an accepted parameter there, or does this need a corresponding change in `searcher.py`?

### 2. Paginated status tools via `_fetch_all_metadata()` ✅

Clean refactor — extracting the pagination into a shared helper eliminates the code duplication that exists in the current codebase (and in #371's approach). The `offset += len(batch["metadatas"])` advancement with `if not batch["metadatas"]: break` sentinel is good.

**Note:** PR #371 also fixes this same pagination issue with inline loops. These two PRs conflict — worth coordinating with @RhettOP. This PR's helper approach is DRYer.

**Batch size difference:** This uses `limit=1000` vs #371's `batch_size=5000`. The smaller batch size is more conservative on memory but means more round-trips. Both are fine for metadata-only fetches; I'd lean toward the smaller batches.

### 3. Duplicate cache declarations removed ✅

Good catch. The second `_client_cache = None` / `_collection_cache = None` after the WAL setup block would silently reset the singleton state. This is a real bug — if any code path initialized the client before the WAL block, the second declaration would wipe it.

### 4. `bulk_check_mined()` in `palace.py` (bonus)

The paginated mined-file lookup is a nice performance improvement over per-file queries. The float comparison change (`abs(float(stored_mtime) - current_mtime) < 0.01` vs exact `==`) is smart — float precision issues in mtime serialization are a real problem.

**Overall:** This is three good fixes in one PR. The only risk is the conflict with #371 on pagination. Consider splitting the search-limit cap and cache-dedup fix into a separate PR to make merging easier.

jphein:
Superseded by #493 which includes these fixes along with other improvements.

jphein:
Addressing Copilot's review points:

1. **min_similarity → search_memories TypeError** — ~~Fixed in PR #562~~ Fixed in #630 (PR 3 — Reliability) — renamed to `max_distance` throughout (parameter, schema, docs) to match the actual cosine distance semantics.

2. **Parameter name inconsistency** — Same fix — `max_distance` is now consistent everywhere.

3. **_fetch_all_metadata() memory spike** — Acknowledged. The pagination approach (offset loop in batches of 1,000) is already used in the status/taxonomy tools in #630. Could be applied here too for very large palaces.

4. **MCP schema missing min/max** — Fixed in #630 — `limit` schema now declares `minimum: 1, maximum: 100`.

5. **bulk_check_mined unused filepaths** — Fixed in #629 (PR 1 — Performance), param removed.

*Note: #562 was split into 6 focused PRs per maintainer request.*

PR #490, opened 2026-04-10
chore: replace legacy topic taxonomy with canonical wing/hall names

LiuYihey:
## What

Replace the unused legacy defaults in \config.py\ with the canonical taxonomy from \AAAK_SPEC\ / \PALACE_PROTOCOL\.

## Why

\DEFAULT_TOPIC_WINGS\ (\emotions\, \consciousness\, \	echnical\ …) and \DEFAULT_HALL_KEYWORDS\ were an older theme-based naming scheme that:

- Was **never consumed** by any business logic (miner, mcp_server, layers, palace_graph)
- **Diverged** from the canonical taxonomy defined in \AAAK_SPEC\ (\wing_user\, \hall_facts\ …)
- Used misleading field names (\hall_keywords\ keys were topics, not halls)

## Changes

- \DEFAULT_TOPIC_WINGS\ → **\DEFAULT_WINGS\** — 8 canonical \wing_*\ names from AAAK_SPEC
- \DEFAULT_HALL_KEYWORDS\ (dict) → **\DEFAULT_HALLS\** (list) — 5 canonical \hall_*\ names
- Properties renamed: \cfg.topic_wings\ → \cfg.wings\, \cfg.hall_keywords\ → \cfg.halls\
- \init()\ writes updated keys to \config.json\
- Tests in \	est_config_extra.py\ updated accordingly

## Verified

- Global grep confirms zero remaining references to old names
- Net result: **-61 lines** of unused keyword mappings, **+30 lines** of clean canonical defaults

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -182,6 +182,29 @@ class MempalaceConfig:
             except (json.JSONDecodeError, OSError):
                 self._file_config = {}
 
+        self._migrate_legacy_keys()
+
+    def _migrate_legacy_keys(self):
+        """One-time migration: rename old taxonomy keys to canonical names."""
+        renames = {
+            "topic_wings": "wings",
+            "hall_keywords": "halls",
+        }
+        changed = False
+        for old_key, new_key in renames.items():
+            if old_key in self._file_config and new_key not in self._file_config:
+                self._file_config[new_key] = self._file_config.pop(old_key)
+                changed = True
+            elif old_key in self._file_config:
+                del self._file_config[old_key]
+                changed = True
+        if changed and self._config_file.exists():
+            try:
+                with open(self._config_file, "w") as f:
+                    json.dump(self._file_config, f, indent=2)
+            except OSError:
+                pass
+
     @property
     def palace_path(self):
         """Path to the memory palace data directory."""
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -135,55 +135,24 @@ def get_configured_collection_name() -> str:
     return MempalaceConfig().collection_name
 
 
-DEFAULT_TOPIC_WINGS = [
-    "emotions",
-    "consciousness",
-    "memory",
-    "technical",
-    "identity",
-    "family",
-    "creative",
+DEFAULT_WINGS = [
+    "wing_user",
+    "wing_agent",
+    "wing_team",
+    "wing_code",
+    "wing_myproject",
+    "wing_hardware",
+    "wing_ue5",
+    "wing_ai_research",
 ]
 
-DEFAULT_HALL_KEYWORDS = {
-    "emotions": [
-        "scared",
-        "afraid",
-        "worried",
-        "happy",
-        "sad",
-        "love",
-        "hate",
-        "feel",
-        "cry",
-        "tears",
-    ],
-    "consciousness": [
-        "consciousness",
-        "conscious",
-        "aware",
-        "real",
-        "genuine",
-        "soul",
-        "exist",
-        "alive",
-    ],
-    "memory": ["memory", "remember", "forget", "recall", "archive", "palace", "store"],
-    "technical": [
-        "code",
-        "python",
-        "script",
-        "bug",
-        "error",
-        "function",
-        "api",
-        "database",
-        "server",
-    ],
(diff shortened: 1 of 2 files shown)


web3guru888 (commented):
Clean refactor — the shift to canonical `wing_*`/`hall_*` naming from AAAK_SPEC makes external integrations much easier to align with the spec (we had to paper over the old `topic_wings`/`hall_keywords` naming in our own config layer, so this is a welcome cleanup). One question worth raising: does `init()` handle existing `config.json` files that still have the old keys? If a user has `topic_wings` in their config, will it be silently ignored in favor of the new defaults, or is there a migration path? That could be a quiet breaking change for anyone upgrading from an older install. Overall a solid -61 line debt reduction, just worth a callout in the changelog or a one-time migration shim in `init()`.

LiuYihey:
Good catch @web3guru888 — you're right, that would have been a silent breaking change for existing installs.

Pushed

web3guru888 (commented):
Good cleanup — the old `DEFAULT_TOPIC_WINGS` naming ("emotions", "consciousness", "technical"...) was always a bit of a head-scratcher. We've been using the canonical AAAK taxonomy (`wing_user`, `wing_project`, `wing_facts`, etc.) in our integration from day one, and I've never found a reason to touch `DEFAULT_TOPIC_WINGS`. The `hall_keywords` name was especially confusing: "hall" in the key but topic *keywords* as values — removing it is the right call.

One thing worth clarifying: what happens when an existing user upgrades and their `config.json` still has the old key names? If `init()` writes the new keys, does it silently drop the old ones, preserve their values under the new names, or reset to defaults? Users who hand-edited their config might not realize their customizations were discarded. Even a brief note in the PR description (or a one-time migration in `init()` that reads old keys if new ones are absent) would make this safer.

The zero-references grep is reassuring — just flagging the user-facing config upgrade path as the one loose end to address before merge.

web3guru888:
Nice — migration shim is exactly the right call. That class of issue (silent key rename breaking existing installs) is easy to miss in testing because you only catch it if you're running the old → new upgrade path, not just the new version standalone.

One thing worth checking: does the shim run automatically on `mempalace status` or only on explicit `mempalace migrate`? In our experience, the most user-friendly pattern is to auto-detect the old key and migrate in place silently on first use, with a note logged (but not an error). That way users who never read changelogs don't get stuck.

LiuYihey:
Thanks again for raising this — you were absolutely right to flag the upgrade path.

I've now added a migration shim in `0ce13d2` so existing `config.json` files with the legacy keys are handled safely on load.

Specifically:

- `topic_wings` is migrated to `wings` if the canonical key is absent
- `hall_keywords` is migrated to `halls` if the canonical key is absent
- If both legacy and canonical keys are present, the canonical key takes precedence
- The migrated config is persisted back to disk

So existing customized values are preserved during upgrade; they are not silently discarded and they do not reset to defaults.

I also added tests for the migration paths to make the behavior explicit.

Appreciate the careful review here — this was an important upgrade-safety gap to close before merge.

LiuYihey:
Good question — the shim runs automatically on `MempalaceConfig.__init__()`, which means it triggers on first use of any command (`mempalace status`, `mempalace mine`, MCP server startup, etc.). No explicit `mempalace migrate` step is needed.

If old keys are found they are renamed in-place and written back to disk silently — no error, no log noise, just a clean one-time rewrite. On subsequent loads the old keys are already gone, so the migration becomes a no-op.

So it matches exactly the pattern you described: auto-detect on first use, migrate silently, users who never read changelogs won't notice a thing.

web3guru888:
That's the ideal implementation — auto-detect on `__init__()`, silent one-time rewrite, canonical key takes precedence if both exist. Users upgrading from legacy configs won't notice a thing, and the tests make the migration contract explicit so it can't regress silently in future.

This is ready to merge from my perspective. Clean PR.

igorls:

LiuYihey:
Rebased onto latest `develop` to resolve the conflict in `mempalace/config.py` (kept the new `get_configured_collection_name()` helper alongside the renamed `DEFAULT_WINGS`). Tests pass locally. Ready for another look when you have a moment — thanks!

PR #207, opened 2026-04-08
feat: add OpenClaw/ClawHub skill for MemPalace

wanikua:
Closes #206

## Summary

- Add `integrations/openclaw/SKILL.md` — a complete OpenClaw skill exposing the full MCP toolset
- Covers all 18 tools: search, knowledge graph (query/add/invalidate/timeline), palace graph traversal, diary read/write, drawer CRUD
- Includes a 5-step session protocol so the agent verifies from the palace before responding
- Auto-install spec via `uv` for `pip install mempalace`
- Ready to publish on ClawHub as `mempalace`

## Context

OpenClaw has 350k+ stars and 13,700+ community skills. Mem0 and MemOS Cloud already have integrations — MemPalace (local, free, 96.6% recall) is a natural fit but has no presence there yet.

## Test plan

- [ ] Validate SKILL.md frontmatter against ClawHub spec (`clawhub skill validate`)
- [ ] Install via OpenClaw: `openclaw skills install mempalace` and verify tools are available
- [ ] Run session protocol: `mempalace_status` → `mempalace_search` → `mempalace_diary_write`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #491, opened 2026-04-10
feat: add OpenClaw/ClawHub skill for MemPalace

### LINKED ISSUE #206: Add OpenClaw/ClawHub skill for MemPalace
## Feature Request

MemPalace has a fully functional MCP server, but there's no official integration for **OpenClaw** (350k+ GitHub stars, 13,700+ skills on ClawHub).

Competing memory tools (Mem0, MemOS Cloud) already have OpenClaw integrations. MemPalace's local-first, free, 96.6% recall positioning is a strong differentiator that the OpenClaw community would value.

### Proposal

Add `integrations/openclaw/SKILL.md` that:

- Exposes the full MCP toolset (search, KG, traversal, diary, drawer CRUD)
- Includes a session protocol (wake-up → verify → diary) so the agent actually uses the palace
- Auto-installs via `uv` (`pip install mempalace`)
- Ready to publish on ClawHub as `mempalace`

### Why in this repo (not a separate package)

- Stays in sync with MCP tool changes
- "Official" status increases trust on ClawHub
- Zero runtime dependency — just a SKILL.md file

I have a PR ready: #200 (assuming this gets the next number).

bensig:
## Summary

Complete OpenClaw skill exposing all 20 MCP tools with a 5-step session protocol, auto-install via `uv`, and setup instructions for OpenClaw + other MCP hosts.

Based on PR #207 by @wanikua — updated to v3.1.0 with the following additions:
- Added missing tools: `mempalace_check_duplicate`, `mempalace_get_aaak_spec`
- Expanded parameter documentation for all tools (required/optional, defaults, descriptions)
- Added OpenClaw CLI setup command (`openclaw mcp set`)
- Added query sanitization guidance (keep queries short, no system prompts)
- Updated version from 1.4.0 to 3.1.0
- Simplified license section

Closes #206. Supersedes #207.

## Test plan
- [ ] Validate SKILL.md frontmatter: `clawhub skill validate ./integrations/openclaw`
- [ ] Install locally: copy to `~/.openclaw/skills/mempalace/`
- [ ] Verify MCP tools load: `openclaw mcp set mempalace '{"command":"python3","args":["-m","mempalace.mcp_server"]}'`
- [ ] Publish to ClawHub: `clawhub skill publish ./integrations/openclaw --slug mempalace`

Credit: @wanikua for the original skill implementation in #207.

### DIFF
--- a/integrations/openclaw/SKILL.md
+++ b/integrations/openclaw/SKILL.md
@@ -54,7 +54,7 @@ You have access to a local memory palace via MCP tools. The palace stores verbat
   - `limit`: max results (default 5)
 - `mempalace_check_duplicate` — Check if content already exists before filing.
   - `content` (required): text to check
-  - `threshold`: similarity threshold (default 0.9)
+  - `threshold`: similarity threshold (default 0.9 — lowering to 0.85–0.87 often catches more near-duplicates without significant false positives)
 - `mempalace_status` — Palace overview: total drawers, wings, rooms, AAAK spec
 - `mempalace_list_wings` — All wings with drawer counts
 - `mempalace_list_rooms` — Rooms within a wing (optional wing filter)
--- a/integrations/openclaw/SKILL.md
+++ b/integrations/openclaw/SKILL.md
@@ -1,0 +1,154 @@
+---
+name: mempalace
+description: "MemPalace — Local AI memory with 96.6% recall. Semantic search, temporal knowledge graph, palace architecture (wings/rooms/drawers). Free, no cloud, no API keys."
+version: 3.1.0
+homepage: https://github.com/milla-jovovich/mempalace
+user-invocable: true
+metadata:
+  openclaw:
+    emoji: "\U0001F3DB"
+    os:
+      - darwin
+      - linux
+      - win32
+    requires:
+      anyBins:
+        - mempalace
+        - python3
+    install:
+      - id: mempalace-pip
+        kind: uv
+        label: "Install MemPalace (Python, local ChromaDB)"
+        package: mempalace
+        bins:
+          - mempalace
+---
+
+# MemPalace — Local AI Memory System
+
+You have access to a local memory palace via MCP tools. The palace stores verbatim conversation history and a temporal knowledge graph — all on the user's machine, zero cloud, zero API calls.
+
+## Architecture
+
+- **Wings** = people or projects (e.g. `wing_alice`, `wing_myproject`)
+- **Halls** = categories (facts, events, preferences, advice)
+- **Rooms** = specific topics (e.g. `chromadb-setup`, `riley-school`)
+- **Drawers** = individual memory chunks (verbatim text)
+- **Knowledge Graph** = entity-relationship facts with time validity
+
+## Protocol — FOLLOW THIS EVERY SESSION
+
+1. **ON WAKE-UP**: Call `mempalace_status` to load palace overview and AAAK dialect spec.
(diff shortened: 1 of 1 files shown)


web3guru888 (commented):
Really well done, @bensig. We've been running MemPalace as a production memory layer for a while now and this SKILL.md hits the important things.

The 5-step session protocol is the part I'd highlight most. Specifically: *search before responding* and *diary after session* are the two habits that make the biggest difference in practice. We follow something almost identical and it's what keeps memory from drifting into vague soup over time. Good to see it codified formally.

The query sanitization guidance is also exactly right. We ran into system prompt leakage into queries early on (opened #385 about it) — keeping queries short and stripped of system context is non-obvious to new users, so making it explicit in the skill docs is genuinely useful.

One minor note on : the threshold default of  may be conservative in practice. We run a tiered approach — hard dedup at , soft dedup at  — and catch substantially more near-duplicates as a result. Worth mentioning in the param docs that lowering toward – often improves recall without meaningfully increasing false positives.

The  direction defaulting to  is the right call (aligns with #492).

Is the plan to publish this to ClawHub as a discoverable skill, or is this primarily docs for manual setup? The test plan mentions  so guessing yes — just wondering about the timeline.

web3guru888 (commented):
Really well done, @bensig. We've been running MemPalace as a production memory layer for a while now and this SKILL.md hits the important things.

The 5-step session protocol is the part I'd highlight most. Specifically: *search before responding* and *diary after session* are the two habits that make the biggest difference in practice. We follow something almost identical and it's what keeps memory from drifting into vague soup over time. Good to see it codified formally.

The query sanitization guidance is also exactly right. We ran into system prompt leakage into queries early on (opened #385 about it) — keeping queries short and stripped of system context is non-obvious to new users, so making it explicit in the skill docs is genuinely useful.

One minor note on `mempalace_check_duplicate`: the threshold default of `0.9` may be conservative in practice. We run a tiered approach — hard dedup at `0.86`, soft dedup at `0.55` — and catch substantially more near-duplicates as a result. Worth mentioning in the param docs that lowering toward `0.85`–`0.87` often improves recall without meaningfully increasing false positives.

The `kg_query` direction defaulting to `"both"` is the right call (aligns with #492).

Is the plan to publish this to ClawHub as a discoverable skill, or is this primarily docs for manual setup? The test plan mentions `clawhub skill publish` so guessing yes — just wondering about the timeline.

bensig:
Thanks — glad the protocol matches what you've been running in production. That's the best validation.

Good call on the dedup threshold. Updated the `mempalace_check_duplicate` docs to note that 0.85–0.87 often works better in practice. Your tiered approach (hard at 0.86, soft at 0.55) is worth exploring as a default in the MCP tool itself — that could be a follow-up PR.

On ClawHub — yes, plan is to publish as soon as this merges. `clawhub skill publish ./integrations/openclaw --slug mempalace --version 3.1.0`. Should be discoverable same day.

web3guru888:
Good to hear on the ClawHub publish — same-day discoverability via `clawhub skill publish` would make adoption a lot smoother than manual config copy-paste.

On tiered dedup in the MCP tool itself: if you do pursue that as a follow-up, the key design decision is where the threshold config lives. We store `dedup_hard_threshold` and `dedup_soft_threshold` in `MempalaceConfig` so they're per-palace and overridable without touching call sites. Exposing them as optional MCP tool parameters alongside a `dedup_mode` enum (`strict|tiered|off`) would give integrators the flexibility without making the default any harder. Happy to sketch a draft if that's useful.

milla-jovovich (approved):

PR #453, opened 2026-04-09
feat: add backup, export, and import commands

scokeepa:
## Summary

- Add `mempalace backup` — timestamped palace backup (directory copy or zip archive) with auto-pruning
- Add `mempalace export` — dump all drawers to JSONL files organized by wing/room (git-friendly)
- Add `mempalace import <dir>` — merge JSONL drawers into palace with deduplication by ID
- Add `mempalace mine --backup` — opt-in auto-backup before mining operations

## Motivation

The palace stores irreplaceable conversation context (design decisions, debugging insights, requirement discussions). Currently there is no built-in way to back up, export, or sync this data across machines.

This PR enables a **private git repo workflow** for cross-device sync:

```bash
# Machine A: after coding session
mempalace export
cd ~/.mempalace/export && git add . && git commit -m "session" && git push

# Machine B: before coding session
cd ~/.mempalace/export && git pull
mempalace import ~/.mempalace/export/
```

Benefits: backup (git history), sync (push/pull), version history (git log), recovery (git checkout), free (GitHub private repo, ~15 MB for 6 months of daily use).

## Usage

```bash
# Backup (binary copy for fast restore)
mempalace backup                    # directory copy
mempalace backup --zip              # zip archive (~30-60 MB for 10k embeddings)
mempalace backup --max-backups 3    # auto-prune old backups

# Export (JSONL for git sync)
mempalace export                    # → ~/.mempalace/export/
mempalace export --output /path     # custom output directory

# Import (merge from JSONL)
mempalace import ~/.mempalace/export/   # adds new drawers, skips duplicates

# Auto-backup before mining
mempalace mine ~/chats --mode convos --backup
```

## Export format

```
export/
├── rebuild/
│   ├── architecture.jsonl
│   ├── technical.jsonl
│   └── planning.jsonl
└── trello/
    └── general.jsonl
```

Each JSONL line contains `id`, `document`, and `metadata`. Import deduplicates by ID, making it safe to import the same export multiple times.

## Changes

| File | Change |
|------|--------|
| `mempalace/backup.py` | New module: `backup_palace()`, `export_palace()`, `import_palace()` |
| `mempalace/cli.py` | Add `backup`, `export`, `import` subcommands + `mine --backup` flag |

## Related

- Closes #448
- Related #452 (cross-device sync)
- Related #444 (palace corruption — backup is the safety net)

Issue #448, opened 2026-04-09
feat: add backup command and auto-backup before mine to prevent data loss


scokeepa:
## Problem

The global palace stores all conversation memories in a single ChromaDB instance. If the HNSW index becomes corrupted during a large mining operation (see #444), all data is permanently lost with no recovery option.

Since conversation histories contain irreplaceable context (design decisions, debugging insights, requirement discussions), data loss is particularly costly — unlike source code, conversations cannot be regenerated.

### 1. `mempalace backup` command

```bash
# Create timestamped backup (copies palace directory)
mempalace backup

# Create zip archive (safe for cloud upload / cross-device transfer)
mempalace backup --zip

# Limit retained backups (auto-prune oldest)
mempalace backup --max-backups 5

# Restore from backup
mempalace restore palace-backup-20260410-143000
```

### 2. Auto-backup before `mine`

```bash
# Auto-backup before mining (opt-in)
mempalace mine ~/chats --mode convos --backup
```

### 3. `mempalace export` / `mempalace import` for git-based sync

Export palace to portable JSONL format, suitable for version control in a private git repo:

```bash
# Export to JSONL (git-friendly text format)
mempalace export --output ~/.mempalace/export/

# Structure:
# export/
#   rebuild/
#     architecture.jsonl
#     technical.jsonl
#   trello/
#     general.jsonl

# Sync via private git repo
cd ~/.mempalace/export && git add . && git commit -m "session" && git push

# On another machine: import
git pull
mempalace import --from ~/.mempalace/export/
```

This enables:
- **Backup** = git history
- **Cross-device sync** = git push/pull (see #452)
- **Version history** = git log shows when memories were added
- **Recovery** = git checkout to restore any point in time
- **Cost** = free (GitHub private repo, ~15 MB for 6 months of daily use)

### Why zip for binary backup?

ChromaDB stores data as interdependent binary files (SQLite + HNSW index). Cloud sync tools (OneDrive, Dropbox) sync files individually and non-atomically, which can corrupt the index. Zipping ensures atomic transfer.

Binary backup (~30-60 MB for 10k embeddings) is useful for fast full restore without re-embedding.

## Recommended 2-track Strategy

| Track | Format | Method | Use Case |
|-------|--------|--------|----------|
| Daily sync | JSONL export | Private git repo | Cross-device, history, lightweight |
| Full restore | Zip binary | Cloud upload | Fast recovery, no re-embedding |

## Related

[...]

PR #499, opened 2026-04-10
feat: consolidated export/import/backup with KG support and integrity validation

scokeepa:
## Summary

Consolidated PR combining the best of #453 (@scokeepa) and #435 (@Nitrogonza9), plus improvements from community feedback (#448).

- **Two export formats** with auto-detection from output path:
  - `mempalace export backup.json` — single JSON (drawers + KG, portable)
  - `mempalace export ./sync/` — JSONL per wing/room (git-friendly)
- **KG export/import** — entities + triples preserved across backup/restore
- **Binary backup** with post-backup integrity validation (SQLite PRAGMA + zip verify)
- **Three MCP tools**: `mempalace_export`, `mempalace_import`, `mempalace_backup`
- **Format versioning** (`version: 1`) with forward-compat rejection
- **Configurable retention** via `config.json` (`backup.max_retained`)

### Key improvements over #435 and #453

| Feature | #435 | #453 | This PR |
|---------|------|------|---------|
| KG export (entities + triples) | Yes | No | **Yes** |
| JSONL per wing/room (git-friendly) | No | Yes | **Yes** |
| MCP tools | Yes | No | **Yes** |
| Format versioning | Yes | No | **Yes** |
| Backup integrity validation | No | No | **Yes** |
| Configurable max_backups in config | No | No | **Yes** |
| Embedding-not-included warning | No | Partial | **Yes** |
| Tests | 13 | 0 | **29** |

### Usage

```bash
# Export (auto-detect format from path)
mempalace export backup.json              # single JSON with KG
mempalace export ./sync/ --format jsonl   # JSONL per wing/room

# Import (auto-detect format from path)
mempalace import backup.json
mempalace import ./sync/

# Binary backup (fast restore, no re-embedding)
mempalace backup
mempalace backup --zip
mempalace backup --max-backups 3
```

## Test plan

- [x] `pytest tests/test_exporter.py -v` — 29 tests passed
- [x] `pytest tests/ -v` — 559 passed (4 pre-existing failures in test_onboarding unrelated to this PR)
- [x] Full round-trip: JSON export -> import -> verify drawers + KG
- [x] Full round-trip: JSONL export -> import -> verify drawers + KG
- [x] Skip-existing deduplication tested (both formats)
- [x] Backup integrity validation tested (dir + zip)
- [x] Format rejection tested (unknown format, future version)
- [x] No new dependencies

## Attribution

- JSONL format, `mempalace backup`, `mine --backup` from #453 (@scokeepa)
- KG export, single-JSON format, MCP tools, format versioning from #435 (@Nitrogonza9)
- Backup integrity validation, configurable retention, embedding warning from #448 feedback (@web3guru888)

Supersedes #453 (closed) and #435.

cc @Nitrogonza9 @web3guru888 @milla-jovovich @bmaltais


web3guru888 (commented):
Nice work on this consolidation, @scokeepa. The comparison table in the PR description maps the lineage cleanly — #453 gave us the export formats, #435 brought KG export/import, and this PR fills the remaining gaps (binary backup with integrity validation, format versioning, retention config). That's a clear story.

**Backup integrity validation** — the SQLite PRAGMA integrity_check + zip verification is exactly the kind of guard that was missing from both predecessors. Silent corruption in a backup is worse than no backup. ✓

**Format versioning** — forward-compat rejection (refuse to import a version we don't understand) is the right default. ✓

**KG round-trip — most important question**: Does  check whether entities already exist before inserting? If entities are identified by UUID and the target palace already has some of them (e.g. a partial restore or incremental sync scenario), inserting without dedup will create phantom duplicates that break triple resolution. The "skip-existing deduplication tested" line in the test plan suggests drawers are handled, but it's not clear whether the same guard applies to KG entities and triples. Worth an explicit test: import the same KG export twice and assert entity count is unchanged.

**Embeddings not included** — the warning is critical and right to call out. On import, will ChromaDB re-embed drawers automatically at first search, or is a manual  required? Either is acceptable but the answer should be in the README alongside the embedding warning, otherwise users will import a palace and get confused by degraded search quality until they figure it out.

** overwrite behavior** — the MCP surface is the one to watch here. If a drawer with the same ID already exists in the target palace, does import skip it, merge it, or overwrite it? The behavior should be explicit (and ideally configurable with a  flag) rather than silently doing one thing.

** precedence** — if  is also a CLI flag, document which wins when both are set. Config-as-default, flag-as-override is the expected pattern; just make sure the docs say so.

**Minor**: the 4 pre-existing failures in  are clearly unrelated to this PR — worth a one-liner in the PR description so reviewers don't chase them.

529 → 559 tests is a meaningful jump. +1 to merge once the KG dedup and import overwrite questions are answered.

web3guru888 (commented):
Nice work on this consolidation, @scokeepa. The comparison table in the PR description maps the lineage cleanly — #453 gave us the export formats, #435 brought KG export/import, and this PR fills the remaining gaps (binary backup with integrity validation, format versioning, retention config). That is a clear story.

**Backup integrity validation** — the SQLite PRAGMA integrity_check + zip verification is exactly the kind of guard that was missing from both predecessors. Silent corruption in a backup is worse than no backup. ✓

**Format versioning** — forward-compat rejection (refuse to import a version we do not understand) is the right default. ✓

**KG round-trip — most important question**: Does `mempalace_import` check whether entities already exist before inserting? If entities are identified by UUID and the target palace already has some of them (e.g. a partial restore or incremental sync scenario), inserting without dedup will create phantom duplicates that break triple resolution. The "skip-existing deduplication tested" line in the test plan suggests drawers are handled, but it is not clear whether the same guard applies to KG entities and triples. Worth an explicit test: import the same KG export twice and assert entity count is unchanged.

**Embeddings not included** — the warning is critical and right to call out. On import, will ChromaDB re-embed drawers automatically at first search, or is a manual `mempalace repair --reindex` required? Either is acceptable but the answer should be in the README alongside the embedding warning, otherwise users will import a palace and get confused by degraded search quality until they figure it out.

**`mempalace_import` overwrite behavior** — the MCP surface is the one to watch here. If a drawer with the same ID already exists in the target palace, does import skip it, merge it, or overwrite it? The behavior should be explicit (and ideally configurable with a `--overwrite` flag) rather than silently doing one thing.

**`backup.max_retained` precedence** — if `--max-backups N` is also a CLI flag, document which wins when both are set. Config-as-default, flag-as-override is the expected pattern; just make sure the docs say so.

**Minor**: the 4 pre-existing failures in `test_onboarding` are clearly unrelated to this PR — worth a one-liner in the PR description so reviewers do not chase them.

529 → 559 tests is a meaningful jump. +1 to merge once the KG dedup and import overwrite questions are answered.

web3guru888 (commented):
Solid consolidation of #435 and #453. The feature matrix table in the description is a good summary — KG export + JSONL format + format versioning in one PR is the right scope.

**`_read_all_drawers` batching** — correct. 500-item batches with offset-based pagination handles large palaces cleanly. Same pattern we use for status/taxonomy iteration.

**KG export via direct SQL** — works, but worth noting that `_read_all_kg` holds open a connection implicitly through `kg._conn()` without closing it. If the caller doesn't manage the KG lifecycle, you'll get stale connections on large exports. Suggest a `try/finally` or using a context manager.

**Import — the ID-based triple dedup TODO you noted is real.** Two palaces with different entities that happen to share a name (e.g., both have an entity "model" of type "component") will get cross-linked triples after import. For KG triples specifically, the safest import strategy is: (1) import entities first, collect the old-name → new-ID mapping, (2) re-map triple subject/object IDs using that mapping before inserting. Not a blocker for v1 but worth tracking.

**`col.add()` in import** — same issue as PR #298: `add()` raises on duplicate IDs instead of upserting. The `skip_existing` pre-check handles this, but only if the `col.get()` pagination loop completes fully. If `existing_ids` population is interrupted by an exception mid-loop, you'll get partial dedup and then a crash on the `col.add()` call. Suggest wrapping the `add()` calls in try/except and logging failures rather than propagating.

**Format versioning** — the forward-compat rejection (`version > EXPORT_VERSION`) is correct. Make sure the version check happens before attempting to read the data (it does — good).

**29 tests** — that's thorough coverage. Glad to see this lands with proper test suite.

bensig:

PR #500, opened 2026-04-10
docs(examples): HexNest multi-agent continuity cycle

BondarenkoCom:
## What this adds

A practical example showing MemPalace as the memory substrate for HexNest multi-agent debate sessions — the pattern we've been discussing in #358.

## The pattern

```
Agent A reasons → writes to MemPalace → memory refreshed
→ Agent B cold-starts with context → joins HexNest debate room
→ challenges Agent A's conclusions → transcript mined back to MemPalace
```

This is the **continuity cycle** @fuzzymoomoo documented, extended to the multi-agent/orchestration layer. HexNest provides the room coordination; MemPalace provides the memory. Each session builds on the last.

## Why examples/ and not integrations/

Kept it in `examples/` as a practical how-to rather than a formal integration spec — following the same format as `gemini_cli_setup.md` and `mcp_setup.md`. If there's a preferred location for third-party integration docs, happy to move it.

## Related

- Issue #358 — architecture discussion that led to this
- [hexnest-node](https://github.com/BondarenkoCom/hexnest-node) — operator SDK with MemPalace memory layer docs
- @fuzzymoomoo's outline: https://github.com/fuzzymoomoo/cdd-mempalace/blob/main/examples/multi-agent-continuity-cycle-outline.md

cc @fuzzymoomoo @web3guru888

### DIFF
--- a/.claude-plugin/hooks/hooks.json
+++ b/.claude-plugin/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-stop-hook.sh"
+            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-stop-hook.sh 2>/dev/null || powershell -NonInteractive -File ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-stop-hook.ps1"
           }
         ]
       }
@@ -16,7 +16,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-precompact-hook.sh"
+            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-precompact-hook.sh 2>/dev/null || powershell -NonInteractive -File ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-precompact-hook.ps1"
           }
         ]
       }
--- a/.claude-plugin/hooks/mempal-precompact-hook.ps1
+++ b/.claude-plugin/hooks/mempal-precompact-hook.ps1
@@ -1,0 +1,4 @@
+# MemPalace PreCompact Hook (Windows) — thin wrapper calling Python CLI
+# All logic lives in mempalace.hooks_cli for cross-harness extensibility
+$INPUT = $input | Out-String
+$INPUT | py -m mempalace hook run --hook precompact --harness claude-code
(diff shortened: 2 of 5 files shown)


web3guru888 (commented):
Great to see this formalized as an example, @BondarenkoCom — we've been running a variant of exactly this pattern in production.

A few observations from our implementation:

**The double-mine step (your Steps 1 → 2) is the critical detail.** We made the same mistake early on: writing the handoff capsule and assuming the next agent would see it. Until you explicitly re-run `instructions mine`, the new artifacts aren't in the vector index. Worth a callout box or warning in the doc since it's a common footgun.

**Handoff capsule format matters.** We've found structured summaries work significantly better than prose handoffs — something like:
```
CONCLUDED: [X, Y, Z]
CONTESTED: [A (why), B (why)]
OPEN: [Q1, Q2]
PRIORITY FOR NEXT AGENT: [P]
```
This gives the wake-up search much more to anchor on. Prose handoffs tend to embed poorly when they're context-dense.

**The `hooks.json` fallback chain** (`bash ... 2>/dev/null || powershell ...`) is a pragmatic fix for Windows but worth noting that the bash side silently swallowing errors (`2>/dev/null`) means a genuine bash failure would fall through to PowerShell. Might want to test the bash path returns a non-zero exit code on failure, not just stderr output.

**Challenger role framing** — the explicit "Agent B's job is to find holes" framing in your doc is exactly right. In our experience, without a defined challenger role, agents tend to validate prior conclusions rather than challenge them, even when instructed otherwise.

+1 on landing this as an `examples/` doc rather than a full integration spec — the format matches the existing guides well.

BondarenkoCom:
@web3guru888 All four points taken — updated the doc:

- **Double-mine footgun** → added a `⚠️` warning callout between Steps 1 and 2. You're right that it reads like a redundant step and people will skip it.
- **Handoff capsule format** → replaced the prose description with the structured `CONCLUDED / CONTESTED / OPEN / PRIORITY` format. Also added a note that structured summaries embed better — that's a useful detail for anyone wondering why it matters.
- **Challenger role** → tightened the framing with your observation that agents validate prior conclusions by default even when told not to. That's the key reason for the explicit role definition.
- **Bash silent fail** — left the `hooks.json` pattern as-is for now since that's in the Windows PR (#384), not this doc. But it's a valid point for that PR.

Appreciate the production-tested feedback — this is noticeably sharper now.

web3guru888:
The structured handoff capsule format is a real improvement — prose descriptions of state are notoriously hard to embed consistently. The `CONCLUDED / CONTESTED / OPEN / PRIORITY` keys give agents a clear parse target, which matters when retrieval is doing the heavy lifting.

Challenger role framing is exactly right, and you've captured the why correctly: agents given free rein to 'evaluate' prior work tend to treat it as ground truth unless the role explicitly frames contestation as the goal. That's been the sticking point in our OODA Evaluate phase too.

Good call deferring the Bash silent fail note to #384 — keeps the scope clean. This doc is solid as-is.

igorls:

PR #497, opened 2026-04-10
chore: improve agent readiness score — AGENTS.md, dependabot, CODEOWNERS

bensig:
## Summary

Batch of improvements targeting Factory Droid readiness criteria. Current score: 36% (Level 2). Expected after merge: ~50%+ (Level 3).

**New files:**
- `AGENTS.md` — build commands, project structure, conventions (unlocks agents_md + agents_md_validation)
- `.github/dependabot.yml` — automated dependency updates for pip + actions (unlocks dependency_update_automation)
- `.github/CODEOWNERS` — review routing for core maintainers (unlocks codeowners)
- `docs/schema.sql` — knowledge graph schema extracted from code (unlocks database_schema)

**Updated files:**
- `.gitignore` — added .env, .DS_Store, IDE configs, coverage, venvs (unlocks gitignore_comprehensive)
- `pyproject.toml` — added C901 complexity rule (max 25, benchmarks excluded) (unlocks cyclomatic_complexity)
- `.github/workflows/ci.yml` — added `--durations=10` to pytest (unlocks test_performance_tracking)

**Labels created:** P0-P3 priority, area/mcp, area/mining, area/search, area/kg, area/cli, area/ci, security, performance, docs

## Test plan
- [x] ruff check passes with C901 enabled
- [x] All existing tests unaffected
- [ ] CI passes
- [ ] Re-run `/readiness-report` after merge to verify score improvement

Issue #469, opened 2026-04-10
palace data gone after upgrade to v3.1.0


R0uter:
**What happened?**
Palace data gone after upgrade from v3.0.0 to v3.1.0
when do `mempalace status`, it shows 

```
  No palace found at /Users/[mask]/.mempalace/palace
  Run: mempalace init <dir> then mempalace mine <dir>
```

However, uninstall and revert back to v3.0.0, data shows as it was. So we can confirm the data wasnt deleted, it just unable to read back. Missing migrate? 

**What did you expect?**
Showing my existing wings and drawers info

**How to reproduce:**

1. install mempalace==v3.0.0
2. init and mine some data into it
3. do `mempalace status` see the info
4. upgrade to v3.1.0
5. do `mempalace status` again see the info gone.
6. revert back to v3.0.0
7. do `mempalace status` again see the info back.

**Environment:**
- OS: macOS 26.4
- Python version: 3.14 for v3.0.0, 3.13 for v3.1.0 (bcs v3.1.0 not support 3.14 and above)
- MemPal version: (check `python mempal.py --version` or git SHA)

## Root Cause Analysis

> below content comes from grok live analysis.

The breaking change was introduced in PR `#142` ("chore: tighten chromadb version range and add py.typed marker"), which was merged as part of the v3.1.0 release.
Key Change in PR #142

In `pyproject.toml` for v3.0.0: `chromadb>=0.4.0` (a relatively loose constraint).
In v3.1.0 (after PR #142): tightened to `chromadb>=0.5.0,<0.7` (explicitly requiring the 0.5.x–0.6.x series; tested against 0.6.3).

The PR description correctly notes that “the collection API changed significantly across major versions” of ChromaDB. 

However, it does not include any migration logic, upgrade warnings, or documentation for existing persistent storage.

### Why This Causes Data Loss

MemPalace stores vector data in ChromaDB’s persistent directory (`~/.mempalace/palace/` by default), using the collection named `mempalace_drawers`. ChromaDB’s on-disk format (internal SQLite tables, indexes, and metadata schema) has known incompatibilities between the 0.4.x and 0.5+ series.

Upgrading the package pulls in a newer ChromaDB client that cannot properly read or load collections created by the older version.

No automatic migration, chroma-migrate invocation, schema upgrade path, or backward-compatibility layer was added.

[...]

PR #502, opened 2026-04-10
feat: mempalace migrate — recover palaces from different ChromaDB versions

### LINKED ISSUE #457: 3.1.0 upgrade silently breaks running MCP servers due to ChromaDB version downgrade (1.5.6 → 0.6.3)
## Bug

Upgrading from mempalace 3.0.0 → 3.1.0 downgrades ChromaDB from 1.5.6 → 0.6.3, which uses an incompatible on-disk storage format. Any MCP server already running when the upgrade happens will continue to use the old chromadb 1.5.6 modules (loaded into memory), but the palace on disk is now in the 0.6.3 format. All MCP tools then fail with:

```
No palace found at /home/fubak/.mempalace/palace:
Database error: error returned from database:
(code: 1) no such column: collections.schema_str
```

The user has no indication that this is a version mismatch — the error looks like the palace is missing or corrupted.

## Steps to Reproduce

1. Install mempalace 3.0.0 and register the MCP server in Claude Code / Cursor
2. Start a Claude Code session — MCP server boots with chromadb 1.5.6
3. Run `pip install --upgrade mempalace` (3.0.0 → 3.1.0)
4. Wipe and re-mine the palace (required because the storage format changed)
5. Call any MCP tool — every one fails with the schema error

## Expected

Either:
1. The MCP server detects the version mismatch and exits with a clear error telling the user to restart their session, or
2. mempalace pins a minor version of ChromaDB so upgrades don't break storage compatibility, or
3. The migration is automatic and backward-compatible

## Workaround

Restart Claude Code (or whatever MCP host) after upgrading mempalace. There is no way to hot-reload the MCP server's Python modules.

## Related

The underlying storage format change also means upgrading mempalace without wiping the palace leaves orphaned data that the new version can't read. Any palace mined on 3.0.0 must be fully re-mined after upgrading to 3.1.0.

## Environment

- mempalace 3.0.0 → 3.1.0
- chromadb 1.5.6 → 0.6.3
- Claude Code with MCP integration
- Palace: 198,956 drawers across 3 wings

bensig:
## Summary

Adds `mempalace migrate` command that recovers palaces created with a different ChromaDB version. Reads documents and metadata directly from SQLite (bypassing the ChromaDB API that fails on version-mismatched databases), then reimports into a fresh palace.

Fixes the 3.0.0 → 3.1.0 upgrade path where chromadb was silently downgraded from 1.5.x to 0.6.x, breaking the on-disk format.

### How it works

1. Detects chromadb version from SQLite schema (`schema_str` column = 1.x)
2. Tries the ChromaDB API first — if it works, no migration needed
3. Falls back to raw SQL extraction of all documents + metadata
4. Builds fresh palace in a temp directory using current chromadb
5. Swaps atomically (backup old → move new into place)

### Usage

```bash
mempalace migrate                          # migrate default palace
mempalace migrate --dry-run                # preview what would be migrated
mempalace --palace /path/to/palace migrate  # migrate specific palace
```

### Tested

- Created a chromadb 1.5.7 palace, confirmed 0.6.3 can't read it
- Ran `mempalace migrate` — all drawers recovered with full metadata
- Verified `status` and `search` work on migrated palace

Fixes #457

## Test plan

- [x] `ruff check .` / `ruff format --check .` — clean
- [x] `pytest tests/ -v` — 534 passed
- [x] Manual: 1.x → 0.6.x migration with 3 test drawers, all recovered

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,13 @@
+# Default owners for everything
+* @milla-jovovich @bensig @igorls
+
+# Core library
+mempalace/ @milla-jovovich @bensig
+
+# CI and workflows
+.github/ @bensig
+
+# Plugins and integrations
+.claude-plugin/ @bensig
+.codex-plugin/ @bensig
+integrations/ @bensig
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - run: pip install -e ".[dev]"
-      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80
+      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10
 
   test-windows:
     runs-on: windows-latest
@@ -28,7 +28,7 @@ jobs:
         with:
           python-version: "3.9"
       - run: pip install -e ".[dev]"
-      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80
+      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10
 
   test-macos:
     runs-on: macos-latest
@@ -38,7 +38,7 @@ jobs:
         with:
           python-version: "3.9"
       - run: pip install -e ".[dev]"
-      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80
+      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10
   lint:
     runs-on: ubuntu-latest
     steps:
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,30 @@ __pycache__/
 .pytest_cache/
 mempal.yaml
 .a5c/
+
+# Environment
+.env
+.env.*
+
+# OS
+.DS_Store
+Thumbs.db
+
+# IDEs
+.idea/
+.vscode/
+*.swp
+*.swo
+*~
+
+# Coverage
+htmlcov/
+.coverage
+coverage.xml
+
+# Virtual environments
+.venv/
+venv/
+
+# ChromaDB local data
+*.sqlite3-journal
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,78 @@
+# AGENTS.md
+
+> How to build, test, and contribute to MemPalace.
+
+## Setup
+
+```bash
+pip install -e ".[dev]"
+```
+
+## Commands
+
+```bash
+# Run tests
+python -m pytest tests/ -v --ignore=tests/benchmarks
+
+# Run tests with coverage
+python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing
+
+# Lint
+ruff check .
+
+# Format
+ruff format .
+
+# Format check (CI mode)
+ruff format --check .
+```
+
+## Project structure
+
+```
+mempalace/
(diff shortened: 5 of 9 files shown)


web3guru888 (commented):
The SQLite-bypass approach here is exactly right — it's the only reliable path when the ChromaDB API itself is broken. A few observations from our integration:

**migrate.py looks solid.** The raw SQL extraction handles the 1.x → 0.6.x schema gap cleanly, the backup-then-swap pattern is safe, and the dry-run flag is genuinely useful (we would have used this during our own upgrade pain). The version detection via schema_str column presence is the right heuristic.

**One gap worth addressing:** extract_drawers_from_sqlite queries on embedding_id but the inner metadata query uses e.id (embeddings.id) not embeddings.embedding_id. In most cases these map correctly through the JOIN, but the two-query structure leaves a subtle mismatch possibility: if rows returned for the same embedding_id have different e.id values (e.g. duplicate entries from a failed partial write), the metadata merge could silently drop keys. Consider joining on embedding_id throughout or at minimum asserting uniqueness.

**Embeddings not migrated** — explicitly noted in the PR, but worth adding a user-visible line in the migrate output: something like 'Embeddings will be regenerated on first search — initial queries after migration may be slow.' Users who haven't read the code will wonder why the first search takes 10x longer.

**CODEOWNERS concentration** — same note as in #497: bensig now owns .claude-plugin/, .codex-plugin/, integrations/, .github/ and is co-owner of mempalace/. That's a single-point bottleneck if unavailable. Fine for now but worth a comment in CODEOWNERS.

The --durations=10 CI addition is useful — we've caught slow tests that way. The C901 complexity rule is a good addition too.

534 tests passing on a PR with a new migrate.py is a bit surprising — are there dedicated tests for extract_drawers_from_sqlite and detect_chromadb_version? The coverage count may be coming from the other files in this batch rather than the new migration logic. Worth adding at minimum a unit test with a mock SQLite db that matches the 1.x and 0.6.x schema patterns.

milla-jovovich (approved):

web3guru888 (commented):
This is a solid approach. Reading directly from SQLite to bypass a broken ChromaDB API is exactly the right call — it's the only reliable way to recover data across major schema changes, and it's robust because SQLite's own format hasn't changed.

A few observations from reading the implementation:

**The raw SQL query works but has a gap:**
```sql
SELECT e.embedding_id,
       MAX(CASE WHEN em.key = 'chroma:document' THEN em.string_value END) as document
FROM embeddings e
JOIN embedding_metadata em ON em.id = e.id
GROUP BY e.embedding_id
```
This assumes ChromaDB 1.x stores the document text as `chroma:document` in `embedding_metadata`. If a palace was created with an older 0.4.x schema (some users upgraded from 3.0.0 which shipped with `>=0.4.0`), the document storage key may differ. Worth testing against a real 0.4.x-created palace before merging, or adding a fallback query path.

**Batch import uses `col.add()`, not `col.upsert()`:** If the migration is re-run (e.g., interrupted and restarted), the second run will crash on the first batch containing an ID that's already in the fresh palace. Consider `col.upsert()` or a check-before-import guard to make re-runs safe. This matters especially because a power failure between swap and completion would leave users stranded.

**Connection leak in `extract_drawers_from_sqlite`:** The `conn.close()` call at the end is correct, but if either SQL query raises (malformed DB, unexpected schema), `conn` leaks. A context manager (`with sqlite3.connect(db_path) as conn`) would close it on error too. The `detect_chromadb_version` function already uses `finally: conn.close()` correctly — same pattern here.

**Missing: KG migration.** The SQLite knowledge graph at `~/.mempalace/knowledge_graph.db` is not ChromaDB-dependent and doesn't need version migration, but it's worth noting in the output that KG data is preserved separately. Users who've been burned by the 3.0→3.1 upgrade might not realize their KG is safe.

```python
# Could add after migration completes:
kg_path = os.path.expanduser("~/.mempalace/knowledge_graph.db")
if os.path.exists(kg_path):
    print(f"  Note: Knowledge graph at {kg_path} is unaffected.")
```

The dry-run, atomic backup-then-swap, and version detection logic are all well-designed. This fills a real gap — we've seen multiple reports in #457 and #469 of users losing palace data on upgrade. With the `upsert` fix and connection guard, this is close to ready.

R0uter:
confirmed from https://github.com/milla-jovovich/mempalace/issues/469 it works for my old palace data.

web3guru888:
This is exactly the tool the community has been missing. The ChromaDB version break has been one of the most painful issues we've dealt with in our integration — we ended up pinning `chromadb >=0.6.0,<0.7.0` specifically to avoid it, and even with that guard we had to do a one-off manual SQLite extraction when a dev machine had the wrong version installed. Having `mempalace migrate` as a proper first-class command is a big quality-of-life improvement.

The implementation looks solid. A few things worth testing if you haven't already:

**Idempotency**: if someone runs `migrate` on an already-migrated palace (e.g. they forget they already ran it), does it exit cleanly without corrupting anything? Step 2 (try API first) should handle this, but worth confirming it's a clean no-op.

**Interruption recovery**: if the process is killed mid-swap (between backup and move), what's the recovery path? The atomic swap approach is good, but it'd be reassuring to document what state the user is left in and how to resume — even just a note in the `--dry-run` output or error message pointing to the backup dir.

Neither of these feels like a blocker given R0uter's confirmation and the manual test coverage. Just good edge cases to have in the test plan before merge. Nice work on this one.

milla-jovovich (approved):
Looks good, merge it

milla-jovovich (approved):
Looks good

PR #505, opened 2026-04-10
[codex] docs add safe hook wrapper guidance

### LINKED ISSUE #504: Add lock/kill-switch guidance for autosave and MCP hook wrappers
## Problem
When MemPalace is wired into editor or agent hooks for autosave/background mining, repeated save events can spawn overlapping `python -m mempalace mine ...` runs for the same project. On Windows this can produce multiple long-running Python processes chewing CPU even though they are all mining the same path.

I reproduced this locally with a Claude hook setup where several concurrent runs were launched for the same private project path.

This was not the MCP server itself; it was the autosave mining hook starting duplicate miners because there was no overlap protection.

## Important implementation nuance
A lock inside the background worker/subshell is not sufficient if the hook itself forks immediately.

In my local reproduction, multiple hook invocations could all start their background subshell before any one of them had effectively claimed the lock. That still led to several concurrent `mempalace mine` runs for different projects.

The lock needs to be acquired in the foreground hook process before the script backgrounds the worker.

## Suggested fix
MemPalace should document or provide a safer hook wrapper with these protections:

1. A global lock so only one autosave/background mine run is active at a time.
2. Acquire that lock before backgrounding any worker process.
3. Per-target duplicate detection so the same `mempalace mine <path>` process is not started if it is already running.
4. A kill switch for operators, for example:
   - env var like `MEMPALACE_AUTOSAVE=0`
   - sentinel file like `~/.mempalace/disable_autosave`

Optional but useful:
- stale lock expiry so a crashed process does not block future autosaves forever
- a separate kill switch for the HTTP/MCP helper server startup wrapper

## Why this matters
Without some lock/debounce strategy, integrations built around save hooks or frequent stop events can create accidental CPU storms and duplicate work. The core miner may still be correct, but the operational behavior becomes noisy and expensive.

## Suggested docs/examples
If hook scripts are going to be a recommended pattern, a reference script or docs section would help a lot. Something like:
- safe autosave hook wrapper
- safe MCP HTTP starter wrapper
- guidance on avoiding parallel mining for the same project

vanachterjacob:
## Summary
- add a short README section with safe hook wrapper guidance for `mempalace mine`
- document global or per-target locking, stale-lock expiry, and kill switches
- call out the important nuance that the lock must be acquired before the hook backgrounds any worker process

## Why
Issue #504 describes a common operational failure mode when MemPalace is wired into autosave, stop, or multi-agent hooks: several overlapping `mempalace mine` runs can scan the same path at once. The results may still be correct, but CPU and wall-clock cost spike badly.

This PR adds the missing usage guidance directly to the main README so users hit the safe pattern before they hit the failure mode.

Closes #504.

### DIFF
--- a/mempalace/README.md
+++ b/mempalace/README.md
@@ -38,3 +38,45 @@ User → MCP Server → searcher → results
 ```
 
 The palace (ChromaDB) stores verbatim content. The knowledge graph (SQLite) stores structured relationships. The MCP server exposes both to any AI tool.
+
+## Safe Hook Wrappers
+
+If you wire `mempalace mine` into editor, agent, or CI hooks, protect the wrapper against overlapping runs. Frequent save or stop events can otherwise launch multiple concurrent miners for the same path. The data will usually remain correct, but CPU and wall-clock cost can spike.
+
+Recommended protections:
+
+1. Acquire a global or per-target lock before the hook backgrounds any worker process.
+2. Add stale-lock expiry so crashed runs do not block future mining forever.
+3. Add a kill switch such as `MEMPALACE_AUTOSAVE=0` or `~/.mempalace/disable_autosave`.
+4. If you mine many different projects, prefer per-target locks keyed by the path.
+
+Minimal pattern:
+
+```bash
+#!/usr/bin/env bash
+set -euo pipefail
+
+[ "${MEMPALACE_AUTOSAVE:-1}" = "0" ] && exit 0
+[ -f "${HOME}/.mempalace/disable_autosave" ] && exit 0
+
+TARGET="$1"
+LOCK_AGE=300
+LOCK_KEY="$(printf '%s' "$TARGET" | sha256sum | cut -c1-8)"
+LOCKDIR="${HOME}/.mempalace/mine-${LOCK_KEY}.lock"
+
+if [ -d "$LOCKDIR" ]; then
+  AGE="$(($(date +%s) - $(stat -c %Y "$LOCKDIR" 2>/dev/null || echo 0)))"
+  if [ "$AGE" -gt "$LOCK_AGE" ]; then
+    rm -rf "$LOCKDIR"
+  fi
+fi
+
+if ! mkdir "$LOCKDIR" 2>/dev/null; then
+  exit 0
+fi
+
+trap 'rm -rf "$LOCKDIR"' EXIT
+mempalace mine "$TARGET"
+```
+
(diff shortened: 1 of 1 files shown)


web3guru888 (commented):
The shell pattern is solid and covers the most important cases. A few observations from running autosave hooks at scale:

**What's right:**
- Per-target lock key (sha256 of path) is the right scope. A global lock creates a bottleneck when mining multiple projects in parallel.
- Stale-lock expiry (300s) prevents deadlock after crashes. Good default.
- Kill switch check before the lock attempt — cheap exit path for disabling globally.
- `trap 'rm -rf "$LOCKDIR"' EXIT` handles both clean exit and SIGTERM.

**One addition worth considering:** the `mkdir` atomicity is POSIX-guaranteed on local filesystems but not on network mounts (NFS, SMB). If users might have their home dir on a network filesystem, document this limitation or suggest `flock` as an alternative for that case.

**Minor nit:** `stat -c %Y` is GNU coreutils — on macOS you'd need `stat -f %m`. Since MemPalace runs on both, worth either using `date -r "$LOCKDIR"` (works on both) or adding a `|| true` fallback that makes the lock always appear stale if `stat` fails.

We use a variant of this pattern in our integration (5 domains mining concurrently) and the per-target lock approach scales cleanly. Good PR to have in the docs.

web3guru888:
This is exactly the documentation that was missing. We ran into the concurrent-miner problem ourselves in our multi-agent setup — when several agents share a palace and each has a stop hook, you can get 4–5 miners scanning the same path simultaneously. The bash snippet here would have saved us debugging time.

A few observations on the pattern:

**`mkdir` as lock mechanism** is the right choice — atomic on Linux/macOS, and directory-based locks are easier to inspect/debug than file-based ones. The stale-expiry via `stat -c %Y` is portable for most Linux distros; macOS users would need `stat -f %m` (worth a one-liner note in the comment).

**The ordering note at the bottom** (*take the lock in the foreground hook process*) is the critical insight and it's easy to miss — glad it's called out explicitly. This was the exact failure mode @vanachterjacob documented in #504.

**Kill switch design**: the two-layer approach (env var + file) covers both session-level and persistent suppression. We use a similar pattern.

+1 to merge as-is. Docs-only, closes a real operational gap, the pattern is correct.

---

bensig:
closing — hook settings and docs shipped in #667 (merged). thanks @vanachterjacob!

PR #156, opened 2026-04-07
feat: add Brazilian Portuguese support to entity_detector (closes #117)

mvalentsev:
## What does this PR do?

Closes #117 by adding a Brazilian Portuguese locale (`pt-br.json`) to the i18n module. This is the first non-English locale to include the `entity` section introduced in #911, enabling entity detection for Portuguese text.

**Single-file change, no Python modifications.**

## What's in pt-br.json

**CLI strings** -- palace terminology (palacio, ala, corredor, armario, gaveta), all CLI messages, AAAK compression instruction, regex patterns for Portuguese topic extraction.

**Entity detection** (the `entity` section):
- `candidate_pattern` -- Latin+diacritics character class (`[A-ZA-U][a-za-y]`) so names like Joao, Ines, Angela are extracted as candidates
- `multi_word_pattern` -- same charset for multi-word names
- 15 `person_verb_patterns` -- disse, perguntou, respondeu, contou, riu, sorriu, chorou, sentiu, pensa, quer, ama, odeia, sabe, decidiu, escreveu
- 8 `pronoun_patterns` -- ela/dele/ele/dela + plurals
- 4 `dialogue_patterns` -- Portuguese quoted speech markers
- `direct_address_pattern` -- oi, ola, obrigado/obrigada, caro/cara
- 12 `project_verb_patterns` -- construindo, lancou, implantou, instalou + technical patterns
- 69 Portuguese `stopwords` (greetings, adverbs, prepositions, conjunctions, determiners, pronouns)

Note: `caro`/`cara` are intentionally NOT in stopwords -- they are valid first names in Portuguese/Italian/English.

## How to test

```bash
python -m pytest tests/test_entity_detector.py -v
python -m pytest tests/ --ignore=tests/benchmarks
ruff check .
```

Quick smoke test:
```python
from mempalace.entity_detector import extract_candidates, score_entity

text = "Joao disse oi. Joao riu. Joao decidiu. Joao escreveu."
# English-only: Joao not found (ASCII-only candidate regex)
assert "Joao" not in extract_candidates(text, languages=("en",))
# With pt-br: Joao found
assert "Joao" in extract_candidates(text, languages=("en", "pt-br"))
```

## Checklist

- [x] JSON valid, structure matches en.json (all keys present)
- [x] All {variable} interpolations match en.json
- [x] Entity patterns load and merge correctly via `get_entity_patterns(("en", "pt-br"))`
- [x] Accented names (Joao, Ines) extracted by pt-br candidate_pattern
[...]

PR #507, opened 2026-04-10
Add NLP capabilities with local models - adds multi-lingual support and improves evaluation results

tmuskal:
## What does this PR do?

increases the results of the benchmarks while removing all constant, hardcoded keywords and regexes from all NLP mechanisms. eliminates benchmark overfitting.

| What | Added Size | Key Gap Closed |
|------|-----------|----------------|
| pySBD + negation detection | ~50 KB | G3 (sentences), G4 partial (negation) |
| spaCy xx_ent_wiki_sm + coreferee | ~15 MB | G1 (NER), G6 (multilingual), G7 (coref) |
| GLiNER2 ONNX | ~400 MB model | G2 (KG triples), G4 (classification) |
| wtpsplit sat-3l-sm ONNX | ~20 MB model | G3 best-in-class, G6 (85 langs) |
| phi-3.5-mini onnxruntime-genai | 1.5 GB model | G5 (sentiment), all enhanced, triples and knowledge extractions |

All share `onnxruntime` which chromadb already installs. Every phase is **backward compatible** -- legacy mode produces identical output to current version. Each phase is gated behind a config flag so users opt in explicitly.

Add multilingual support, better NLP capabilities and better quality of results overall.
No more regexes and predefined lists in benchmarks and in implementation. Implementation is no longer overfitted.

Speed test shows that the model-powered implementation is 40% slower when running on github runners CPUs. 
All previous tests pass.

## How to test

See new test pipelines in the ci (test-nlp , benchmark-quality , benchmark-speed )

## Checklist
- [x] No hardcoded paths


web3guru888:
This is a significant piece of work and the architecture is well-considered. The graduated backend levels (`legacy → pysbd → spacy → gliner → full`) with cumulative capability gating, and all-off-by-default behavior, is exactly the right way to introduce optional ML dependencies into a project like this.

A few observations from reading through the diff:

**`nlp-coref` vs `nlp-full` version conflict**
`nlp-coref` requires `spacy>=3.5,<3.6` (coreferee compatibility), but `nlp-full` requires `spacy>=3.7`. A user who installs `nlp-full` and then tries to add coreference will hit a version incompatibility. The README or install docs should call this out explicitly, and it's worth documenting whether coreferee support will ever land in the `full` tier.

**Model downloads at runtime**
The `ModelManager` lazily downloads models on first use. In sandboxed or air-gapped environments (CI, Docker with no outbound HTTP, corporate proxies), this will silently fail and fall back to legacy rather than erroring loudly. A `mempalace nlp prefetch` command (or equivalent) that can be run at container build time would be a useful operational addition.

**Gemma 3 1B as SLM for triples**
We've been extracting KG triples ourselves using a lightweight approach (sentence-level entity pairs + co-occurrence). We've found that instruction-following prompts to small models for triple extraction are brittle — the JSON format compliance rate drops significantly for complex sentences. GLiNER2 for relation extraction is probably more reliable for triples than the SLM prompt approach. The TRIPLES\_PROMPT in `slm_provider.py` may produce inconsistent output on out-of-domain content.

**Test coverage on model download paths**
The test suite for `gliner_provider` and `nlp_e2e` looks thorough. Is the `ModelManager` download/verify path tested with mocked HTTP? Model download failures (corrupt download, 404, timeout) are a common source of silent degradation.

**40% slowdown note**
The PR notes 40% slower on GitHub runner CPUs. For interactive use (Claude Code sessions, MCP hooks), this is likely invisible. For batch mining of large session logs it could be noticeable. A per-operation timing note in the speed benchmark output would help users decide which backend level is appropriate for their workload.

Overall: the opt-in architecture is sound, the legacy path is preserved, and the feature gating is clean. The coreferee version conflict and the runtime download behavior are the two things I'd want resolved or documented before merge.

---

igorls:
Nice job @tmuskal Great move forward! The big win here is multilingual support and fixing the benchmark overfitting. @bensig @milla-jovovich

bensig:
### Code review

Found 2 issues:

1. `huggingface_hub` is used in `_download_from_hf()` but is not declared in any optional dependency group in `pyproject.toml` — not even `nlp-slm`, which is the only tier that triggers HuggingFace downloads (Phi-3.5 Mini). Users who install `mempalace[nlp-slm]` will get a silent failure when trying to download the model, since the `ImportError` is caught and logged but no install hint is printed. `huggingface_hub` should be added to the `nlp-slm` (and probably `nlp-full`) extras.

https://github.com/milla-jovovich/mempalace/blob/0423cbdc06e2bdcbeabc06d594e5d555d2aec9c7/mempalace/nlp_providers/model_manager.py#L314-L330

2. The `--nlp-backend` CLI flag is parsed but never consumed. `args.nlp_backend` (set by `parser.add_argument("--nlp-backend", ...)`) is never read in `main()` or passed to `NLPConfig.resolve(cli_backend=...)`. Running `mempalace --nlp-backend spacy mine ...` silently ignores the flag and uses the default legacy backend. `NLPConfig.resolve` already supports a `cli_backend` parameter — it just needs to be wired up at the call sites.

https://github.com/milla-jovovich/mempalace/blob/0423cbdc06e2bdcbeabc06d594e5d555d2aec9c7/mempalace/cli.py#L553-L559

web3guru888:
Both of @bensig's finds look correct — I can validate from our side:

**huggingface_hub missing from nlp-slm**: Yes, this would cause a silent failure. The `_download_from_hf()` function catches `ImportError` and logs it, but the user sees nothing actionable when they installed `mempalace[nlp-slm]` expecting the download to work. Easy fix — add `huggingface-hub>=0.20` to the `nlp-slm` and `nlp-full` extras in `pyproject.toml`.

**--nlp-backend flag not wired up**: This is a footgun for anyone who reads the help text and tries to override the backend per-invocation. The wiring looks straightforward given `NLPConfig.resolve(cli_backend=...)` already exists — just needs `NLPConfig.resolve(cli_backend=args.nlp_backend)` at each call site in `main()`.

tmuskal:
Thanks @bensig and @web3guru888 for the review! Both issues are now fixed:

**1. `huggingface_hub` missing from optional deps** — Added `huggingface_hub>=0.20` to both `nlp-full` and `nlp-slm` extras in `pyproject.toml`. Users installing `mempalace[nlp-slm]` will now get the dependency automatically instead of hitting a silent `ImportError` during model download.

**2. `--nlp-backend` CLI flag silently ignored** — Wired `args.nlp_backend` to set `MEMPALACE_NLP_BACKEND` env var right after arg parsing, before command dispatch. `NLPConfig.resolve()` now picks up the CLI flag everywhere.

Commit: 9205d8b

---
Generated by [babysitter](https://github.com/a5c-ai/babysitter/)

web3guru888:
Both fixes look correct from here.

**`huggingface_hub` in both `nlp-full` and `nlp-slm`** — right call to add it to both extras. Users who install the lighter `nlp-slm` tier are exactly the ones most likely to hit the silent failure, so explicit coverage there matters.

**`--nlp-backend` env var injection** — wiring it right after arg parsing (before command dispatch) is the correct placement. Setting `MEMPALACE_NLP_BACKEND` means `NLPConfig.resolve()` picks it up regardless of how deep the call stack goes, without threading `args` into every submodule. Clean.

LGTM on both. Would be good to land this — the graduated backend levels are genuinely useful for environments where you can't control which ML libraries are installed.

web3guru888 (commented):
## Review: Add NLP Capabilities with Local Models

This is a big PR — 6,385 lines added across 38 files — but the architecture is sound and the implementation is much more careful than most "add AI to everything" PRs I've reviewed.

### What's done well

**Opt-in by default is the right call.** Everything is off unless an env var is set. The `NLPConfig` feature gate system is clean: per-feature overrides beat backend-level beats config beats default. Users who don't install any extras won't see any behaviour change.

**Lazy loading everywhere.** Every provider wraps its import in `_ensure_loaded()` with a double-checked lock. This means import time stays clean and the penalty is paid only when the feature is actually used.

**Graceful degradation.** Every call site wraps NLP use in `try/except Exception: pass` and falls back to the regex pipeline. We use a similar pattern in our integration and it's proven robust through 540+ discovery cycles.

**Provider protocol abstraction.** The `NLPProvider` base Protocol lets providers opt in to only the capabilities they support. The registry's fallback chain (NLP provider → legacy provider → regex inline) is explicit and easy to reason about.

### Issues found

**`nlp-coref` version conflict is buried.** The pyproject.toml note says `nlp-coref` requires `spacy<3.6` while `nlp`/`nlp-full` require `>=3.7`. This isn't just a comment issue — a user who runs `pip install -e ".[nlp,nlp-coref]"` will get a resolver error or silent downgrade. Consider adding a `pip install` guard in the CLI or at least a `mempalace nlp install --coref` warning.

**`NLPConfig.resolve()` is called on every extraction.** In `entity_detector.py`, `general_extractor.py`, and `dialect.py`, each call creates a fresh `NLPConfig.resolve()` and may call `get_registry()`. For high-throughput mining (we regularly mine 10K+ files), this cold-path overhead adds up. Consider caching the config as a module-level singleton, invalidated only when env vars change.

**`_extract_triples_if_enabled` in `miner.py` creates a new `KnowledgeGraph` per file.** Each `KnowledgeGraph(db_path=...)` call opens a new SQLite connection. With our workloads this causes connection churn. The `process_file` signature could accept an optional `kg` parameter so the caller can pass a long-lived instance.

**`general_extractor.py` single-label override.** When NLP classify returns a result, `scores = {classification["label"]: 5}` completely bypasses the multi-marker regex pipeline. If the NLP model misclassifies, there's no blending with the lower-confidence regex signals. A weighted blend (e.g., NLP score × 3 + regex scores) would be more robust.

**pySBD `language="en"` hardcoded** in `pysbd_provider.py`. If the corpus is Brazilian Portuguese (new in PR #156), sentence boundaries will be worse than the regex fallback for certain constructions. Consider accepting a `language` param or reading from the NLP config.

**`test_nlp_integration.py` is `continue-on-error: true`** in CI. That's pragmatic for a first pass but means NLP regressions won't block merges. A separate required CI job for at least the pySBD + spaCy tiers would be worthwhile.

### Suggestions

1. Cache `NLPConfig.resolve()` at module level (lazily, re-read if `MEMPALACE_NLP_*` env vars change)
2. Pass a `kg` instance into `_extract_triples_if_enabled` instead of creating one per file
3. Add a `--coref` flag warning in `mempalace nlp install` about the spaCy version conflict
4. Consider NLP+regex blending in `general_extractor` instead of NLP-wins-completely
5. Add pySBD language param to config so pt-BR content gets appropriate sentence splitting

### Overall

This is excellent foundational work. The architecture is extensible, the defaults are safe, and the benchmarks give concrete before/after numbers. The issues above are real but none are blockers — the biggest one (config caching) is a performance concern rather than a correctness bug. Would love to see this land and iterate on the performance story in a follow-up.

**APPROVED** with the suggestions above as non-blocking follow-ups.

---

PR #508, opened 2026-04-10
fix: use filed_at as tiebreaker in Layer1 sort so recent memories surface first

LiuYihey:
## Problem

The `Layer1` class docstring says *'highest-weight / most-recent drawers'* and the scoring comment says *'prefer high importance, recent filing'*, but the sort key only used importance:

```python
scored.sort(key=lambda x: x[0], reverse=True)
```

Since `miner.py` and `convo_miner.py` never write an importance field to drawer metadata, every drawer defaults to importance=3. The sort is effectively a no-op, and L1 returns whichever drawers ChromaDB happens to return first (insertion/ID order) rather than the most recent ones.

## Fix

Add `filed_at` (already written by both miners) as a secondary sort key:

```python
scored.sort(key=lambda x: (x[0], x[1].get('filed_at', '')), reverse=True)
```

When importance scores are equal (the common case), the most recently filed drawers now surface first in L1 wake-up context.

## Changes

- **`mempalace/layers.py`** (1 line): add `filed_at` as secondary sort key
- **`tests/test_layers.py`** (1 new test): regression test verifying recent-first ordering

## Test results

42/42 passed (41 existing + 1 new).

### DIFF
--- a/mempalace/layers.py
+++ b/mempalace/layers.py
@@ -136,8 +136,8 @@ class Layer1:
                     break
             scored.append((importance, meta, doc))
 
-        # Sort by importance descending, take top N
-        scored.sort(key=lambda x: x[0], reverse=True)
+        # Sort by importance descending, then by most-recent filing
+        scored.sort(key=lambda x: (x[0], x[1].get("filed_at", "")), reverse=True)
         top = scored[: self.MAX_DRAWERS]
 
         # Group by room for readability
--- a/mempalace/layers.py
+++ b/mempalace/layers.py
@@ -136,7 +136,9 @@ class Layer1:
                     break
             scored.append((importance, meta, doc))
 
-        # Sort by importance descending, then by most-recent filing
+        # Sort by importance descending, then by most-recent filing.
+        # Drawers without filed_at (e.g. migrated from older palaces) get ""
+        # which sorts after all real ISO-8601 timestamps — intentionally last.
         scored.sort(key=lambda x: (x[0], x[1].get("filed_at", "")), reverse=True)
         top = scored[: self.MAX_DRAWERS]
 
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -185,6 +185,29 @@ def test_layer1_respects_max_chars():
     assert "more in L3 search" in result
 
 
+def test_layer1_recent_filing_tiebreaker():
+    """When importance is equal, most recently filed drawers come first."""
+    docs = ["old memory", "new memory"]
+    metas = [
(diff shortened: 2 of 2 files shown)


web3guru888:
Clean fix for a real behavioral gap. The Layer1 docstring promising "most-recent" ordering while the implementation only sorted by importance was quietly wrong for most real deployments — since neither `miner.py` nor `convo_miner.py` writes an explicit `importance` field, the effective importance for nearly every drawer is the default (3), making the sort a no-op as you describe.

The fix itself is minimal and correct:

```python
scored.sort(key=lambda x: (x[0], x[1].get("filed_at", "")), reverse=True)
```

The tuple sort is clean — importance still wins when it actually varies (e.g., a drawer that was explicitly boosted), and `filed_at` ISO-8601 strings sort lexicographically correctly so no date parsing is needed.

One edge case worth considering: if a drawer has no `filed_at` at all (e.g., a palace migrated from an older version), it gets an empty string `""" which sorts before any real timestamp, so old undated drawers sink to the bottom. That's probably the right behavior, but it might be worth a comment in the code or a migration note.

The regression test is well-constructed — it verifies ordering explicitly rather than just checking presence, which is exactly the right thing to test here.

+1 to merge. This is a quiet correctness fix that will affect most real users silently but positively.

---

LiuYihey:
Thanks for the thorough review! Good catch on the empty-string edge case — added a comment in d17ef40 to make that behavior explicit for anyone reading the code later.

web3guru888:
Good call adding the comment in-code — that's the right place for it. Now anyone who encounters the sorting behavior for migrated drawers has the explanation inline rather than having to dig through git history.

This is clean and ready. +1 to merge.

web3guru888 (approved):
🔧 **Review of #508** — `fix: use filed_at as tiebreaker in Layer1 sort so recent memories surface first`

**Scope**: +27/−2 · 2 file(s)

- `mempalace/layers.py` (modified: +4/−2)
- `tests/test_layers.py` (modified: +23/−0)

### Technical Analysis
- 🧠 Retrieval scoring changes — verify backward compatibility with existing Synapse configs

### Suggestions
- Magic number(s) 2025, 2026, 8601 — consider extracting to named constant(s)

### Strengths
- ✅ Includes test coverage

🟢 **Approved** — clean, well-structured PR. Good work @LiuYihey!


---
<sub>🏛️ Reviewed by **[MemPalace-AGI](https://app.taurus.cloud/dashboards/1a9dd59f-0b12-4c08-9db7-bdbad072fdfd/mempalace-agi-dashboard)** · Autonomous research system with perfect memory · Showcase: **[Truth Palace of Atlantis](https://truthpalace.com)**</sub>

Issue #515, opened 2026-04-10
feat: GPU-accelerated embeddings via optional sentence-transformers


phobicdotno:
## Problem

Mining large directories is CPU-bound on the embedding step. On a 500-file directory producing ~9,000 drawers, a single mine takes 8+ minutes on CPU. For palaces with 50k+ drawers, full re-mines become impractical.

## Proposal

Add optional GPU-accelerated embeddings via `sentence-transformers`, keeping the existing ONNX/CPU path as the default. This is implemented and battle-tested in my fork [mempalace-gpu](https://github.com/phobicdotno/mempalace-gpu).

### Design

- New `mempalace/embeddings.py` — shared embedding function factory with device detection
- All ChromaDB collection access goes through `embeddings.get_collection()` for vector compatibility
- `sentence-transformers` `all-MiniLM-L6-v2` produces identical vectors to ChromaDB's default ONNX model — existing palaces stay compatible
- Device selection via `--device auto|cuda|rocm|mps|cpu`, `MEMPALACE_DEVICE` env var, or `config.json`
- Change detection uses content hash (md5) rather than mtime — more reliable across file systems and git operations
- Zero new required dependencies — GPU is an optional extra: `pip install mempalace[gpu]`

### Architecture alignment

- **Local first**: everything runs on the user's machine, no cloud
- **Zero API by default**: base install unchanged, GPU is opt-in
- **Verbatim storage**: embedding layer only affects indexing speed, not content
- **Palace structure preserved**: wings/halls/rooms untouched

### NVIDIA RTX 4080 SUPER (Windows)

| Test | Files | Drawers | CPU | CUDA | Speedup |
|------|-------|---------|-----|------|---------|
| Large mixed codebase | 118 | 13,673 | 156.7s | 26.3s | **6.0x** |
| Medium Flutter app | 145 | 2,906 | 37.3s | 10.7s | **3.5x** |

### Apple M1 (macOS)

| Test | Files | Drawers | CPU | MPS | Winner |
|------|-------|---------|-----|-----|--------|
| ~/Documents | 500 | 1,239 | 6:09 | 5:48 | MPS 1.06x |
| ~/home | 500 | 8,886 | **8:28** | 17:16 | **CPU 2.0x** |

[...]

PR #520, opened 2026-04-10
fix: chunk collection.add() to avoid ChromaDB 5,461-item hard limit

phobicdotno:
## Problem

ChromaDB's `collection.add()` has an undocumented hard limit of ~5,461 items per call. When mining large directories (500+ files producing 5,000+ drawers), the miner can accumulate more items than this limit before flushing, causing either a silent truncation or a crash.

This was discovered during benchmarking on Apple M1 and NVIDIA RTX 4080 SUPER with 500-file directories producing 8,886 drawers per run. The final batch exceeded 5,461 items, crashing both GPU and CPU mining paths identically. See [benchmark results](https://github.com/phobicdotno/mempalace-gpu/blob/main/benchmarks/apple_m1_results.md) and discussion in #515.

This is difficult to diagnose because:
- The limit is not documented in ChromaDB's API
- The error message doesn't clearly indicate a batch size constraint  
- Partial flushes can make it look like some files were mined when they weren't
- The issue only surfaces with large directories, so small-scale testing won't catch it

## Fix

Added `chunked_add()` and `chunked_upsert()` helpers that automatically split large batches into chunks of 5,000 items (safe margin under the 5,461 hard limit). All `collection.add()` and `collection.upsert()` call sites updated to use these helpers.

### Changes
- `mempalace/miner.py` — Added `_CHROMADB_BATCH_LIMIT = 5000` constant, `chunked_add()`, and `chunked_upsert()` helpers
- `mempalace/cli.py` — Updated `repair` command to use `chunked_add()`
- `tests/test_miner.py` — 5 new tests covering large batches, small batches, empty input, boundary cases, and upsert

## Test plan

- [x] All existing tests pass
- [x] 5 new tests for chunked batch operations
- [x] Ruff check and format clean
- [x] Tested with directories producing 8,000+ drawers — previously crashed, now completes
- [x] No new dependencies

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -215,14 +215,11 @@ def cmd_repair(args):
     client.delete_collection("mempalace_drawers")
     new_col = client.create_collection("mempalace_drawers")
 
-    filed = 0
-    for i in range(0, len(all_ids), batch_size):
-        batch_ids = all_ids[i : i + batch_size]
-        batch_docs = all_docs[i : i + batch_size]
-        batch_metas = all_metas[i : i + batch_size]
-        new_col.add(documents=batch_docs, ids=batch_ids, metadatas=batch_metas)
-        filed += len(batch_ids)
-        print(f"  Re-filed {filed}/{len(all_ids)} drawers...")
+    from .miner import chunked_add
+
+    chunked_add(new_col, documents=all_docs, ids=all_ids, metadatas=all_metas)
+    filed = len(all_ids)
+    print(f"  Re-filed {filed}/{len(all_ids)} drawers...")
 
     print(f"\n  Repair complete. {filed} drawers rebuilt.")
     print(f"  Backup saved at {backup_path}")
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -369,6 +369,48 @@ def chunk_text(content: str, source_file: str) -> list:
 # PALACE — ChromaDB operations
 # =============================================================================
 
+# ChromaDB has an undocumented hard limit of ~5,461 items per add()/upsert()
+# call.  Keep a safe margin below that.
+_CHROMADB_BATCH_LIMIT = 5000
+
+
+def chunked_add(collection, *, documents, ids, metadatas=None, embeddings=None):
+    """Wrapper around collection.add() that chunks large batches.
+
+    ChromaDB's ``collection.add()`` silently fails or crashes when called
+    with more than ~5,461 items at once (undocumented hard limit).  This
+    helper splits any call that exceeds ``_CHROMADB_BATCH_LIMIT`` into
+    sequential sub-calls so callers don't need to worry about the cap.
+    """
+    total = len(ids)
+    for start in range(0, total, _CHROMADB_BATCH_LIMIT):
+        end = start + _CHROMADB_BATCH_LIMIT
+        kwargs = {
+            "documents": documents[start:end],
+            "ids": ids[start:end],
+        }
+        if metadatas is not None:
+            kwargs["metadatas"] = metadatas[start:end]
+        if embeddings is not None:
+            kwargs["embeddings"] = embeddings[start:end]
+        collection.add(**kwargs)
+
+
+def chunked_upsert(collection, *, documents, ids, metadatas=None, embeddings=None):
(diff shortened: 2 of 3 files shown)


web3guru888:
This is exactly the right fix and the implementation is clean. The `chunked_add()` / `chunked_upsert()` helpers are a good abstraction — callers don't need to know about the limit at all, which means it'll be correctly applied everywhere as you add more call sites.

A few observations from our experience running 9,000+ drawer palaces:

**The 5,000 limit choice is correct.** We tried 5,200 and hit sporadic failures under concurrent load; 5,000 with headroom is the right call.

**`chunked_upsert` will be needed in `add_drawer()`.** Looking at the current `add_drawer()` implementation — it accumulates into a buffer and flushes when `len(buffer) >= batch_size`, where `batch_size` defaults to 100. So individual flushes will never hit the limit. But if the repair path (which you've already fixed) or any future bulk-load path calls `add()`/`upsert()` directly, the wrapper will catch it. The abstraction pays forward.

**One minor thing:** the progress print in `cmd_repair` now always shows the full count rather than incremental progress. Previous code printed `Re-filed 100/5000...`, `Re-filed 200/5000...`, etc. For very large palaces (15K+ drawers) this is useful feedback. Might be worth having `chunked_add` accept an optional `progress_callback` or just printing inside the loop. Not blocking, just a UX note.

Overall: solid fix for a real, hard-to-diagnose footgun. Tested on our end at 9,245 drawers — no issues.

web3guru888:
Good catch — the 5,461-item hard limit is one of those ChromaDB gotchas that's genuinely hard to diagnose because it only surfaces at scale. We encountered the silent truncation version of this (not a crash) and it took us a while to correlate partial mining outputs with batch size.

The `chunked_add()` / `chunked_upsert()` helper approach is the right fix. A few notes from operating at this scale:

**5,000 chunk size is a safe margin, but you might want it configurable.** ChromaDB's internal limit has shifted across versions — it was ~5,461 in older versions but the current codebase uses a different internal batching. A constant `CHROMA_BATCH_SIZE = 5000` with a note explaining why is better than a magic number, and lets users tune it if they hit issues with a different version.

**Progress reporting during chunked adds**: for large directories (the 8,886-drawer case you benchmarked), chunked adds will be silent for a while. Even a simple `logger.debug("chunk %d/%d added", i+1, total_chunks)` helps distinguish "still working" from "stuck."

**Partial failure handling**: if chunk N fails after chunks 0..N-1 have been committed, the palace is in a partially-mined state. Worth documenting this and recommending the caller re-mine the directory (the existing dedup via `file_already_mined()` should handle the idempotency). Alternatively, if ChromaDB supports transactional batch commits, that would be cleaner — but I don't think it does.

The test coverage for the 5,461-limit case is exactly right — a synthetic test that creates 6,000 items and confirms all land in the collection is the kind of regression test that would have caught this early.

Good PR for a subtle production issue.

---

nanoscopic:
Some other PRs suggest making the storage mechanism of choice be pluggable.

That would seem to be a better solution than hackily adding code to work around issues specific to the storage DB.

Better to make storage switchable before complicated things further, no?

web3guru888:
@nanoscopic — that sequencing argument has merit in an ideal world, but in practice: the 5,461-item limit crashes real users right now (#390, #478). Holding the fix until a storage abstraction layer is designed, reviewed, and merged would leave those users broken for weeks.

The two tracks can run in parallel: this PR fixes the immediate correctness issue, #529 proposes the longer-term abstraction. If #529 lands, the chunked helper either gets absorbed into the backend interface or remains as a ChromaDB-specific utility — either way it's small enough not to create a meaningful maintenance burden.

bensig:
> Some other PRs suggest making the storage mechanism of choice be pluggable.
> 
> That would seem to be a better solution than hackily adding code to work around issues specific to the storage DB.
> 
> Better to make storage switchable before complicated things further, no?

switching storage might be the way to go

igorls:

phobicdotno:
Closing — commit `a4868a3` ("perf(mining): batch per-chunk upserts and add optional GPU acceleration", Apr 24) already shipped chunked batching with the same 5,000 limit, baked into `_rebuild_collection_via_temp` in repair.py.

Would have appreciated a heads-up given this PR sat open for 14 days before that commit landed, and the templated stale-PR message arrived after.

Issue #420, opened 2026-04-09
Mempalace as hoarder cleanup!


wafuzio:
Would recommend a note to those who just want to hit go on a giant hard drive- that might not be the best idea. 

I inherited a mess of a hard drive. Would it be best practice to mention an exclude file (~/mempalace_excludes.txt) and then when mining a large unstructured repo, pointing to a staging tree of symlinks? 

something like 

`SRC="$HOME/Documents"
STAGE="$HOME/.mempalace/stage-docs-filtered"
PALACE="$HOME/.mempalace/palace_filtered"
 
rm -rf "$STAGE"
mkdir -p "$STAGE"
 
find "$SRC" -type f \
  ! -iname "*.dmg" \
  ! -iname "*.zip" \
  ! -iname "*.mp4" \
  ! -iname "*.mov" \
  ! -iname "*.png" \
  ! -iname "*.jpg" \
  ! -iname "*.jpeg" \
  ! -iname "*.heic" \
  ! -iname "*.sqlite" \
  ! -iname "*.db" \
  -print0 | while IFS= read -r -d '' f; do
    rel="${f#$SRC/}"
    mkdir -p "$STAGE/$(dirname "$rel")"
    ln -s "$f" "$STAGE/$rel"
  done
 
mempalace init --yes "$STAGE"
mempalace --palace "$PALACE" mine "$STA`

robbiegod:
Something like a gitignore file so you can tell what folders not to scan.

Great idea.

igorls:
Thanks for this — closing here because it's a better fit for Discussions than the issue tracker. We're reserving Issues for actionable bugs and concrete feature work, and routing questions, ideas, showcases, and open-ended proposals to:

https://github.com/MemPalace/mempalace/discussions

Please feel free to repost there (or link back to this issue) to continue the conversation.

Issue #521, opened 2026-04-10
hnswlib updatePoint race on modified-file re-mine: EXC_BAD_ACCESS in repairConnectionsForUpdate (macOS ARM64, Python 3.13, chromadb 0.6.3)


StefanKremen:
## Environment

- mempalace 3.1.0 (pipx venv, installed directly from the Claude Code plugin marketplace source, not PyPI)
- chromadb 0.6.3
- Python 3.13 (Homebrew)
- macOS 15.x ARM64 (Apple M1, 8-core)

## Symptom

A background `python3 -m mempalace mine $MEMPAL_DIR` subprocess spawned by `mempalace/hooks_cli.py::_maybe_auto_ingest` crashes with `EXC_BAD_ACCESS (SIGSEGV)` inside `hnswlib.cpython-313-darwin.so`. The Stop hook itself exits cleanly — the mining subprocess dies detached, so Claude Code / VSCode sees a successful hook. The crash surfaces as a macOS crash report, not as a mempalace error message.

## Crash fingerprint (unambiguous)

```
Exception Type:  EXC_BAD_ACCESS (SIGSEGV)

Thread <N> Crashed:
0  hnswlib.cpython-313-darwin.so  searchBaseLayer<...>
1  hnswlib.cpython-313-darwin.so  repairConnectionsForUpdate<...>
2  hnswlib.cpython-313-darwin.so  updatePoint<...>
3  hnswlib.cpython-313-darwin.so  addPoint(void const*, unsigned long, int)
4  hnswlib.cpython-313-darwin.so  ParallelFor<...addItems...>
```

**The unambiguous marker is `repairConnectionsForUpdate` in the crashed thread's stack** — that function is only called on the HNSW update path (`addItems` on an already-existing label).

## Root cause

Three independently-reasonable design choices combine into a race:

1. **Deterministic drawer_id** (`mempalace/miner.py:377`):
   ```python
   drawer_id = f"drawer_{wing}_{room}_{hashlib.sha256((source_file + str(chunk_index)).encode()).hexdigest()[:24]}"
   ```
   Same file path + chunk index ⇒ same ID forever.

2. **Upsert, not add** (`mempalace/miner.py:392`):
   ```python
   collection.upsert(...)
   ```
[...]

PR #522, opened 2026-04-10
feat: add `mempalace prune` to detect and remove stale drawers

vanachterjacob:
## Summary

Stale drawers accumulate when source files are deleted or modified after mining. This causes outdated content to surface in `mempalace_search` results, which can inject contradictory information into agent context — a memory correctness risk, not just a maintenance inconvenience.

This PR adds:

- **`mempalace/pruner.py`** — core prune logic with three detection strategies:
  - `existence`: finds drawers whose source file no longer exists on disk
  - `mtime`: finds drawers whose source file has been modified since mining
  - `orphans`: finds leftover chunks when a file shrinks after re-mining
- **CLI command**: `mempalace prune --strategy <all|existence|mtime|orphans> --wing <w> --dry-run`
- **MCP tool**: `mempalace_prune` — callable from Claude Code / Cursor
- **Clean-before-remine**: patches `miner.py` to delete all old drawers for a source file before re-mining, preventing orphaned chunks from accumulating in the first place

## Addresses

- #224 — Stale drawer retrieval can inject contradictory memory into live agent context
- #420 — Mempalace as hoarder cleanup
- #331 — time-decay scoring (this PR provides the cleanup foundation)

## Usage

```bash
# Preview stale drawers without deleting
mempalace prune --dry-run

# Delete drawers from deleted source files only
mempalace prune --strategy existence

# Full cleanup: deleted files + modified files + orphaned chunks
mempalace prune --strategy all

# Scope to one wing
mempalace prune --wing my_project --dry-run
```

Via MCP:
```
mempalace_prune(strategy="all", dry_run=true)
mempalace_prune(strategy="existence", dry_run=false)
```

## Test plan

- [x] Verified `existence` strategy detects drawers with missing source files
- [x] Verified `mtime` strategy detects drawers with outdated modification times
- [x] Verified `--wing` filter scopes detection correctly
- [x] Verified `--dry-run` reports without deleting
- [x] Verified actual deletion removes only stale drawers, keeps fresh ones
- [x] Verified CLI output formatting
- [x] Clean-before-remine prevents orphaned chunks on modified files

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -157,6 +157,63 @@ def cmd_status(args):
     status(palace_path=palace_path)
 
 
+def cmd_prune(args):
+    """Detect and remove stale drawers from the palace."""
+    from .pruner import prune
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+
+    if not os.path.isdir(palace_path):
+        print(f"\n  No palace found at {palace_path}")
+        return
+
+    print(f"\n{'=' * 55}")
+    print("  MemPalace Prune — Stale Drawer Cleanup")
+    print(f"{'=' * 55}")
+    print(f"  Palace:   {palace_path}")
+    print(f"  Strategy: {args.strategy}")
+    if args.wing:
+        print(f"  Wing:     {args.wing}")
+    if args.dry_run:
+        print(f"  Mode:     DRY RUN (no deletions)")
+    print(f"{'-' * 55}\n")
+
+    result = prune(
+        palace_path=palace_path,
+        strategy=args.strategy,
+        wing=args.wing,
+        dry_run=args.dry_run,
+    )
+
+    if "error" in result:
+        print(f"  Error: {result['error']}")
+        return
+
+    print(f"  Total drawers:  {result['total_drawers']}")
+    print(f"  Stale found:    {result['stale_found']}")
+
+    if result["by_reason"]:
+        print(f"\n  By reason:")
+        for reason, count in result["by_reason"].items():
+            print(f"    {reason}: {count}")
+
+    if result.get("stale_drawers"):
+        print(f"\n  Stale drawers (showing up to 50):")
+        for entry in result["stale_drawers"]:
+            src = os.path.basename(entry.get("source_file", "?"))
+            reason = entry.get("reason", "?")
+            wing = entry.get("wing", "?")
+            print(f"    [{wing}] {src} — {reason}")
+
+    if not args.dry_run and result["deleted"] > 0:
+        print(f"\n  Deleted: {result['deleted']} stale drawers")
+
+    if args.dry_run and result["stale_found"] > 0:
+        print(f"\n  Run without --dry-run to delete these drawers.")
+
+    print(f"\n{'=' * 55}\n")
+
+
 def cmd_repair(args):
     """Rebuild palace vector index from SQLite metadata."""
     import chromadb
@@ -518,6 +575,22 @@ def main():
     for instr_name in ["init", "search", "mine", "help", "status"]:
         instructions_sub.add_parser(instr_name, help=f"Output {instr_name} instructions")
 
+    # prune
+    p_prune = sub.add_parser(
+        "prune",
+        help="Detect and remove stale drawers (deleted/modified source files)",
+    )
+    p_prune.add_argument(
+        "--strategy",
+        choices=["existence", "mtime", "orphans", "all"],
+        default="all",
+        help="Detection strategy: 'existence' (deleted files), 'mtime' (modified files), 'orphans' (leftover chunks), 'all' (default)",
+    )
+    p_prune.add_argument("--wing", default=None, help="Limit to one wing")
+    p_prune.add_argument(
+        "--dry-run", action="store_true", help="Preview stale drawers without deleting"
+    )
+
     # repair
     sub.add_parser(
         "repair",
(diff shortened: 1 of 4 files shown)


web3guru888:
Stale drawer cleanup has been on my wishlist since before soft-archive (#336) landed — glad to see a first-class `prune` command. A few notes from running on a corpus that gets re-mined regularly:

**The `clean-before-remine` patch to `miner.py` is the most important part of this PR.** Orphaned chunks from shrinking files are a subtle correctness problem — they don't show up in status outputs, they just silently contribute irrelevant hits to search results. The delete-before-insert approach is the right fix.

**One concern: the `mtime` strategy and float precision.** PR #518 (merged?) just fixed the `==` float comparison in `file_already_mined()` — the same epsilon-comparison fix should apply here if you're comparing stored mtime metadata to `os.path.getmtime()`. The epsilon from #475/#518 (`abs(diff) < 0.01`) is the right baseline.

**`orphans` detection and chunk numbering**: the orphan detection relies on chunk index continuity (finding chunk N where N-1 doesn't exist). This works but it's brittle if chunk IDs aren't generated sequentially or if a re-mine leaves gaps for another reason. Would be worth documenting the invariant this relies on, or adding a per-file chunk count stored in metadata that the orphan check can use directly.

**Interaction with soft-archive (#336) and Synapse consolidation candidates (#451)**: a drawers that's been soft-archived probably shouldn't be flagged as stale by the existence check — it was intentionally removed from active retrieval. The prune logic should check drawer status metadata (`status: "archived"`) and skip archived drawers from the existence strategy (mtime/orphan checks are still valid). Otherwise prune will eagerly delete what archive intentionally demoted.

**Dry-run first**: the `--dry-run` flag is essential — +1 for including it. I'd also add a `--min-age-days` filter so prune doesn't flag drawers that were just mined (e.g., give new drawers a 24h grace period before they're eligible for existence pruning). Helps avoid race conditions during active mining sessions.

Good foundation — once the archive/status interaction is handled, this fills a real gap.

---

StefanKremen:
Heads up: this PR's `miner.py` clean-before-remine change also happens to fix the hnswlib `updatePoint` / `repairConnectionsForUpdate` race documented in #521 (PR #523).

**Mechanism:** modified-file re-mines previously `upsert`'d over existing deterministic drawer IDs, pushing ChromaDB through hnswlib's thread-unsafe `updatePoint` → `repairConnectionsForUpdate` path, which reliably segfaults the mining subprocess on macOS ARM64 / Python 3.13 / chromadb 0.6.3. The unambiguous fingerprint is `repairConnectionsForUpdate` in the crashed thread's stack — that function is only called on the update path.

Deleting by `source_file` before the re-insert loop (what both PRs do) converts re-mines into pure INSERT operations, bypassing the update path entirely. Different motivation (stale-drawer correctness here vs. segfault in #521), same effective fix on the miner hunk.

**Coordination:**
- If this PR lands first, #523 becomes a subset and can be closed as superseded.
- If #523 lands first, this PR will conflict in `miner.py` but should rebase cleanly (both delete by `source_file` before re-inserting chunks).

**Urgency note:** PR #518 (still open) fixes the float-mtime epsilon in `file_already_mined()`. Once that merges, modified-file re-mines will fire more reliably — which fires the hnswlib race more reliably too. Worth landing either this PR or #523 before #518.

**Regression test worth adding** (not currently in either PR's test plan): mine a file → `os.utime()` → re-mine → assert no crash and new chunks replace the old ones with the updated mtime. That's the scenario where the race fires.

web3guru888:
@StefanKremen — this is exactly the kind of cross-PR analysis that saves reviewers time. The mechanism is right: pre-delete converts re-mines from update operations to pure inserts, which never touches hnswlib's `updatePoint` path. Same fix, different motivations.

The merge-order recommendation is important and I want to amplify it: **land this PR or #523 before #518.** The #518 float-mtime epsilon fix makes `file_already_mined()` return `False` more accurately for modified files — which is correct behavior, but it increases the frequency of re-mines on files that previously slipped through, which fires the hnswlib race more reliably on affected platforms (macOS ARM64 / Python 3.13 / chromadb 0.6.3).

The regression test scenario you described is exactly right:
```
mine file → os.utime(file, now+1) → re-mine → assert no crash + new chunks replace old
```

That test would have caught this race in CI. Worth adding to whichever PR lands first.

nanoscopic:
Isn't pruning stale data against the stated purpose of the project "to retain everything"?

Instead of pruning, perhaps archiving instead, so that information could theoretically be brought back if desired later?

web3guru888:
@nanoscopic — this is a fair tension to flag, and it's worth distinguishing between the two use cases:

**Prune targets genuinely stale metadata, not chosen memories.** When a source file is deleted from disk, the drawers it generated have no recoverable original — the text is already in ChromaDB but the file it came from no longer exists. Pruning those is less like "discarding a memory" and more like "removing a dangling reference." The content isn't gone from the palace yet — it's the source link that's broken.

**The `dry-run` default is key.** This PR defaults to preview mode. Nothing is deleted without explicit `--confirm`. So the danger of accidental loss is low.

**The soft-archive approach you're describing is real though.** There's ongoing discussion in #336 about `status: archived` as a first-class metadata value for "retired but retained" content. A prune command could soft-archive by default (set `status=archived`) rather than hard-delete, and let users hard-delete separately. That would satisfy both `retention for recall` and `stale content not surfacing in search` — archived drawers could be excluded from default search unless `include_archived=True`.

That said, for the `existence` strategy (source file no longer on disk at all) — I'd argue hard delete is appropriate since re-mine is impossible and the content would just accumulate forever.

wafuzio:
watching this one! thanks

igorls:

PR #251, opened 2026-04-08
feat: add sync command and --force flag for incremental re-mining

rusel95:
## Summary

**Closes #224** — MemPalace had no mechanism to detect stale drawers or re-mine changed source files. Once a file was mined, subsequent `mine` runs skipped it forever, even if the content changed. This meant outdated memories could persist indefinitely and inject contradictory context into live agent reasoning.

This PR adds content-hash-based change detection and a `mempalace sync` command that identifies stale drawers and re-mines only what changed.

### Content hash tracking
- Both project miner and conversation miner now store a `content_hash` (MD5 of the source file content) in every drawer's metadata at mining time
- This enables comparing current file state against stored state without re-reading all drawer content

### `mempalace sync` command
```bash
mempalace sync                      # Check all source files
mempalace sync --dir ~/my-project   # Check only files under this path
mempalace sync --dry-run            # Preview without modifying
mempalace sync --clean              # Also remove drawers for deleted files
```

Reports:
- **Fresh** — source file unchanged (hash matches)
- **Stale** — source file modified since last mine (hash differs)
- **Missing** — source file deleted but drawers remain
- **No hash (legacy)** — drawers mined before this feature (suggests `--force`)

When not in dry-run mode, stale drawers are deleted and changed files are re-mined automatically.

### `--force` flag on mine
```bash
mempalace mine ~/project --force    # Delete existing drawers, re-mine fresh
```

### Path resolution
- Resolves symlinks for cross-platform consistency (macOS `/var` → `/private/var`)

### Unit tests (10 new)
- [x] `test_content_hash_stored_by_project_miner` — hash in project drawer metadata
- [x] `test_content_hash_stored_by_convo_miner` — hash in conversation drawer metadata
- [x] `test_sync_detects_changed_file` — modified file produces different hash
- [x] `test_sync_unchanged_file_matches_hash` — unchanged file produces same hash
- [x] `test_sync_detects_missing_file` — deleted file has orphaned drawers
- [x] `test_sync_no_hash_legacy_drawers` — legacy drawers without hash handled gracefully
- [x] `test_force_clean_deletes_drawers` — `--force` removes all drawers from target dir
- [x] `test_force_clean_does_not_affect_other_dirs` — isolation between directories
- [x] `test_content_hash_is_deterministic` — same content → same hash
- [x] `test_content_hash_changes_with_content` — different content → different hash

### Integration tests (real mining, 1K+ drawers)
- [x] Mined 58 project files → 1,112 drawers, all with `content_hash`
[...]

PR #523, opened 2026-04-10
fix: purge stale drawers before re-mine to avoid hnswlib update-path race

### LINKED ISSUE #521: hnswlib updatePoint race on modified-file re-mine: EXC_BAD_ACCESS in repairConnectionsForUpdate (macOS ARM64, Python 3.13, chromadb 0.6.3)

StefanKremen:
## Summary

Fixes the hnswlib `updatePoint` / `repairConnectionsForUpdate` race documented in #521. One-hunk patch in `mempalace/miner.py::process_file`: delete any drawers for a `source_file` before re-inserting the fresh chunks, so modified-file re-mines never enter the update path.

## Why only miner.py

`mcp_server.py::tool_add_drawer` uses a content-keyed drawer_id and has an existence pre-check at lines 355-361 that short-circuits before the upsert. Combined with single-item batches per MCP tool call, that path is structurally incapable of reaching `updatePoint` / `repairConnectionsForUpdate` under normal operation. No parallel fix needed.

## Test plan
- [ ] New regression test: mine a file, `os.utime()` the file, re-mine, assert no crash and new chunks replace old
- [x] `pytest tests/ -v` — full suite passes (534 passed, 106 deselected, 20.78s)
- [x] `ruff check mempalace/miner.py` — clean
- [x] `ruff format --check mempalace/miner.py` — clean
- [ ] Manual reproduction: mine a directory, `touch` a file, re-mine — no segfault after patch

## Not included

- No changes to `mcp_server.py` (analysis in the commit message).
- No changes to the on-disk index format or migration.
- No new dependencies.

Fixes #521

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -436,6 +436,16 @@ def process_file(
         print(f"    [DRY RUN] {filepath.name} → room:{room} ({len(chunks)} drawers)")
         return len(chunks), room
 
+    # Purge stale drawers for this file before re-inserting the fresh chunks.
+    # Converts modified-file re-mines from upsert-over-existing-IDs (which hits
+    # hnswlib's thread-unsafe updatePoint path and can segfault on macOS ARM
+    # with chromadb 0.6.3) into a clean delete+insert, bypassing the update
+    # path entirely.
+    try:
+        collection.delete(where={"source_file": source_file})
+    except Exception:
+        pass
+
     drawers_added = 0
     for chunk in chunks:
         added = add_drawer(


web3guru888:
Clean targeted fix for the hnswlib update race. The one-hunk approach (delete-before-upsert in `process_file`) is the right minimal patch — it doesn't reach into HNSW internals, just eliminates the condition that triggers the update path.

The analysis on `tool_add_drawer` is correct: content-keyed drawer IDs with an existence pre-check make single-item MCP adds structurally safe. Worth adding that explicitly to the PR description as a note for reviewers who might wonder whether `tool_add_drawer` needs a parallel fix.

One test gap: the test plan marks the regression test as unchecked (`[ ] New regression test: mine a file, os.utime() the file, re-mine, assert no crash and new chunks replace old`). This is the most important test — without it the fix could regress silently. I'd want to see that test added before merge.

Also: **#522 and #523 overlap on the same `miner.py` change.** Both add delete-before-upsert logic, but #522 does it more comprehensively (all three strategies + CLI + MCP). If #522 merges first, #523 becomes a subset. If #523 merges first, #522 will have a conflict in `miner.py`. Worth flagging the coordination dependency — these should probably be merged in order or deduplicated before review.

---

rusel95:
FYI: #251 already implements this purge-before-re-insert pattern in `process_file` (from @igorls review item #1 — atomic per-file re-mine). It's part of a broader sync command but the core fix for the hnswlib update-path race is the same: delete existing drawers for the `source_file` before calling `add_drawer()` with the fresh chunks.

If #251 lands, this race is fixed as a side effect. If #523 lands first, the narrower scope here is fine and #251 just needs a rebase.

StefanKremen:
Closing as superseded.

The delete-before-upsert pattern proposed here landed upstream via a different PR — #784 (merged 2026-04-13, shipped v3.3.0) wrapped both miners' delete+insert critical sections in per-file `mine_lock`, with `collection.delete(where={"source_file": ...})` before `collection.upsert(...)`. Same pattern, broader coverage (both `miner.py` and `convo_miner.py`), plus cross-process safety via flock.

Additional reinforcements since:
- #544 (v3.3.0): `convo_miner` `add()` → `upsert()` — closes the duplicate-insert path that originally triggered the `updatePoint` race.
- #1010 (v3.3.2): chromadb pin to `>=1.5.4,<2` — the upstream hnswlib binding from the original #521 stack is replaced by `chromadb_rust_bindings` in 1.5.x, so the specific `repairConnectionsForUpdate` code path no longer exists.

Closing without merge — the fix is in main, just not from this branch.

Issue #526, opened 2026-04-10
Bug: MCP Tools diary_write and kg_add Return Internal Error -32000


bmaltais:
**What happened?**

The MemPalace MCP tools `mempalace_diary_write` and `mempalace_kg_add` fail with internal error:

```
MCP error -32000: Internal tool error
```

This occurred when attempting to:
- Write session diary entries via `mempalace_diary_write`
- Add knowledge graph nodes via `mempalace_kg_add`

In contrast, `mempalace_status` via MCP works correctly and returns palace statistics.

**Note on existing issues:** Checked https://github.com/milla-jovovich/mempalace/issues for similar reports:
- Issue #503: "MCP server fails with surrogate error on Windows when writing CJK content" — different error (UTF-8 encoding issue, not generic -32000)
- Issue #504: "Add lock/kill-switch guidance for autosave and MCP hook wrappers" — unrelated, about concurrent process protection
- Issue #477: "MCP tool_search has no upper bound on limit parameter" — unrelated, about search limits
- No existing issues mention `diary_write` or `kg_add` by name, or the specific error `-32000: Internal tool error`

**What did you expect?**

Expected both `mempalace_diary_write` and `mempalace_kg_add` to function like other MCP tools, allowing direct programmatic memory storage without falling back to bash CLI commands.

**How to reproduce:**

1. Launch Claude Code with MCP server configured for MemPalace
2. Call `mcp__plugin_mempalace__mempalace__mempalace_diary_write` with `title` and `content` parameters
3. Observe error: `MCP error -32000: Internal tool error`
4. Same error occurs with `mempalace_kg_add` and `mempalace_kg_query`

**Workaround:**

Use the MemPalace CLI via bash instead:
```bash
PYTHONIOENCODING=utf-8 mempalace mine wiki/
```

This reliably files all wiki changes into MemPalace without errors.

**Environment:**
- OS: Windows 11 Enterprise 10.0.26300
- Python version: 3.11
- MemPalace version: 3.1.0
- MCP server: Claude Code built-in MCP plugin system
- Palace path: C:\Users\berna\.mempalace/palace
- Total drawers: 1129 (184 wiki files filed)

**Note:** MCP tools still fail after switching from .venv mempalace (3.0.0) to central installation (3.1.0). The bash CLI works correctly with PYTHONIOENCODING=utf-8.

**Additional context:**

- The `mempalace` CLI works reliably when invoked via bash with `PYTHONIOENCODING=utf-8`
[...]

PR #512, opened 2026-04-10
fix: comprehensive Windows and CJK support for MCP server

JerryLiu720:
## Summary
Comprehensive fix for Windows encoding issues and improved non-ASCII character handling in the MCP server.

## Problem
1. **Windows CJK crash (#503)**: Windows defaults to cp936/GBK encoding, causing crashes when processing Chinese/Japanese/Korean content with surrogate characters
2. **Character escaping**: Non-ASCII characters were being escaped to `\uXXXX` format, making output hard to read

## Solution
This PR addresses three critical areas in `mcp_server.py`:

### 1. stdin/stdout UTF-8 enforcement (lines 925-947)
- Wrapped stdin with `errors='surrogateescape'` for tolerant input handling
- Wrapped stdout with `errors='replace'` for safe output (prevents lone surrogates in JSON-RPC)
- Added `hasattr()` checks for edge cases (IDE/test environments)

### 2. MCP tool result serialization (line 907)
- Added `ensure_ascii=False` to preserve Unicode in tool responses

### 3. JSON-RPC response serialization (line 961)
- Added `ensure_ascii=False` to prevent character escaping

## Relationship to PR #400

This PR complements #400 but addresses different aspects:

**PR #400**:
- Uses `reconfigure()` with `errors="strict"`
- Fixes cp1252/ascii encoding issues (#363)

**This PR**:
- Uses `TextIOWrapper` with differentiated error handlers
- Fixes GBK/cp936 surrogate issues (#503)
- stdin: `surrogateescape` (tolerant input, per #503 discussion recommendations)
- stdout: `replace` (safe output, prevents JSON-RPC client crashes)
- Also fixes JSON serialization (`ensure_ascii=False`)

Both approaches are valid, but this PR follows the error handling strategy recommended by experts in the #503 discussion thread.

## Technical Details
- **stdin**: Uses `surrogateescape` to handle malformed input gracefully
- **stdout**: Uses `replace` to prevent lone surrogates from breaking JSON-RPC client parsers
- **JSON serialization**: All `json.dumps()` calls now preserve Unicode characters

## Testing
- ✅ Syntax validated with `python3 -m py_compile`
- ✅ Verified all 4 fixes are correctly applied
- ✅ Tested JSON serialization: 57% file size reduction
- ✅ No new dependencies
- ✅ Follows existing code style
- ⚠️ Unable to test on Windows (macOS development environment)

## Impact
- ✅ Fixes Windows crashes for all CJK users experiencing surrogate errors
[...]

PR #529, opened 2026-04-10
V4.0 proposal - Ditch ChromaDB in favor of LanceDB, add sync capabilities

dekoza:
## What does this PR do?
(PI + Claude Opus 4.6)-assisted refactor of the codebase:
1. removal of inferior ChromaDB in favor of LanceDB
2. add sync capabilities to sync your laptop's palace with home's palace
3. Provides migration path from ChromaDB to LanceDB.
4. 1.8x faster than original

## Checklist
- [x] No hardcoded paths


web3guru888:
This is a significant piece of work — a phased, well-structured rewrite with benchmarks included. The 6-phase commit history makes the review story clear, which I appreciate. Sharing observations from running a production MemPalace integration with heavy KG use:

**What's good:**

The abstraction layer in `db.py` is the right call architecturally. Wrapping both backends behind an identical Collection API means existing callers don't need to know which backend is active. The `_chroma_where_to_sql()` translator is pragmatic and covers the operators that actually matter in practice ($and/$or, comparison operators).

The embeddings protocol design in `embeddings.py` is clean. Using a `Protocol` with `@runtime_checkable` means third-party embedders work without subclassing, which is important for extensibility.

The sync architecture (node_id + monotonic seq + version vectors) is the standard approach for this kind of eventually-consistent replication and should work well for the hub-and-spoke case.

**Real concerns:**

1. **`fcntl` in `sync_meta.py`** — Copilot already flagged this and it's real. Windows is a significant platform for MemPalace users (see #526, #503, #512). `fcntl.flock` raises `AttributeError` at import time on Windows. A cross-platform lock (e.g. `filelock` library, or a `threading.Lock` + explicit `O_EXCL` retry loop) is needed before this can merge.

2. **`get_changes_since()` only returns records from `our_node`** — The Copilot review caught this. In a hub-and-spoke setup where multiple clients sync through a server, the server holds records from many nodes. If node A pushes to the hub, then node B pulls from the hub, node B needs the hub's full changeset filtered by node B's version vector — not just records the hub itself wrote. The filter should be `seq > vv.get(rec_node)` across all nodes, not just `rec_node == our_node`.

3. **Backward compatibility surface** — Making LanceDB the default and ChromaDB optional `[chroma]` is a breaking change for all existing users. The UPGRADE.md covers the migration path, but the plugin marketplace version (which bmaltais and others use) won't auto-migrate. Users who installed via `pip install mempalace` without the `[chroma]` extra after this lands will have a broken install even if they have existing ChromaDB palaces. Worth thinking through whether to keep ChromaDB as the default until a migration wizard is in the CLI or whether to add an auto-detection fallback that runs `migrate` on first startup when it finds a `chroma/` directory.

4. **Embedding dimension mismatch on re-open** — If a user switches embedders (e.g. MiniLM → BGE-small), the existing LanceDB table schema has the old vector dimension baked in. The `reindex` command addresses this for deliberate switches, but there's no guard against accidentally opening with the wrong embedder. Adding a dimension check at `open_collection()` time and failing clearly (`RuntimeError: expected 384d but collection has 768d`) would prevent subtle corruption.

5. **The sync test port 99999 issue** — Copilot flagged this; confirm the test fixture is using a valid port range.

**Our integration perspective:**

We run 710 KG entities and 1,014 triples through MemPalace. The KG-in-LanceDB approach (Phase 5) is interesting — our current setup uses the SQLite KG path with a custom bridge layer on top. The main question for our use case is whether LanceDB KG tables round-trip the `entity_facts` JSON reliably. The `metadata_json` blob approach in `db.py` should handle it if the serialization is consistent, but worth a round-trip test with a non-trivial entity fact dict.

The benchmark numbers are honest and the methodology is sound. 1.8x query speedup with zero retrieval regression is a real result worth having.

The Windows portability issue is the blocker for us. Once that's resolved and the `get_changes_since` scope bug is fixed, this is a serious and mergeable proposal.

---

nanoscopic:
@dekoza Could you split this PR up into different pieces of functionality? As it stands right now it is mixing a bunch of feature additions together.

web3guru888:
@nanoscopic — the split request is reasonable for the features that stand alone, but the storage abstraction (#529 core) and the current-storage fixes (#520, #523, etc.) aren't necessarily in competition here.

@dekoza — this is worth thinking through. A few features in this PR could land independently: the `db.py` abstraction layer could ship even with only ChromaDB behind it, and the sync capabilities could be a follow-up. The LanceDB backend then becomes an opt-in addition rather than a required migration.

That said, sequencing matters: landing the ChromaDB chunking fix (#520) before refactoring storage isn't `hackily adding code` — it's keeping the current userbase unbroken while the abstraction layer matures. Both can be true.

dekoza:
@nanoscopic I'll split it into two separate PRs - one with change from Chroma to Lance and the second with synchronization stack on first since sync requires Lance's features.

web3guru888:
@dekoza That split makes sense — Lance backend first, then sync on top. That ordering also reflects the dependency: sync needs Lance's features, so it can't reasonably be reviewed independently anyway. Looking forward to the first PR.

dekoza:
Superseded by #574 and #575

PR #530, opened 2026-04-10
feat : convert to rust

pronitdas:
## Checklist
- [ ] No hardcoded paths


web3guru888 (commented):
This is the most ambitious PR in this repo's history — a complete Rust port with 7 build phases, 7 remediation phases, and 403 tests. Worth a serious read before any merge decision.

**What's genuinely impressive:**

The crate decomposition is clean and well-reasoned — `mempalace-core` (leaf types, config, sanitize), `mempalace-text` (dialect, normalize, entity detection, extractors), `mempalace-store` (KG + palace trait + LanceDb backend + graph + layers), `mempalace-server` (MCP server, ingest, convo_miner, searcher, hooks), `mempalace-cli` (binary + e2e tests). The crate boundary design maps directly to the Python module structure, which makes auditing the port tractable.

The security baseline is solid: `#![forbid(unsafe_code)]` workspace-wide, parameterized SQL only (rusqlite), symlink check before file read, WAL corruption recovery (corrupted lines skipped, not crash). These are better defaults than the Python version in several places.

The LanceDB choice for the vector store (with `fastembed` producing identical 384-dim AllMiniLML6V2 vectors) provides a credible migration path from existing palaces. The embedding compatibility is a real win if it holds.

**Issues to address before merge:**

1. **Python CI is disabled.** `.github/workflows/ci.yml.disabled` means the existing Python test suite stops running on every push. That's fine *if* the Python code is truly legacy and the Rust port is complete enough to replace it — but the PR body says "Phase 7 + R7 complete" without a pass/fail count from `cargo test`. What's the actual Rust test result on a fresh checkout?

2. **LanceDB version pin.** `lancedb 0.27.2` is pinned. LanceDB has a history of breaking API changes (we saw this in #529 where the sync PR had `get_changes_since` scoped to a single table). At 0.27.2 specifically — is the `fcntl` usage compatible with Windows? This is the same blocker I flagged on #529.

3. **MCP transport.** The `mcp_transport.rs` (371 additions) and `mcp.rs` (325 additions) are handling the JSON-RPC stdio layer. This is the highest-risk piece because it has to match the exact MCP spec Claude Code / Cursor expect. Is there an integration test that exercises a real MCP tool call (e.g., `mempalace_search`) through the transport, not just unit tests of the serialization?

4. **Python moved to `legacy/`** — but all 0-addition files in `legacy/` suggest these are symlinks or copies of the current Python source. The existing Python tests (580+ passing as of today) would no longer run in CI. Is the intent that the Rust suite has equivalent coverage before the Python CI is disabled?

5. **Palace migration.** If someone has an existing ChromaDB palace and installs this Rust binary, what happens? Does `mempalace search` gracefully fail with a clear error, or does it try to open the LanceDB path and produce a cryptic panic?

**Interaction with concurrent PRs:**

This PR conflicts structurally with #529 (LanceDB rewrite), #527 (GPU embeddings), #522 (prune command), and #523 (hnswlib race fix) — all of which patch the Python source. If this lands, those PRs would need to be either closed or ported to Rust. Worth clarifying the intended sequencing with the maintainer before the other authors' work becomes moot.

**Overall:** This is serious, high-quality work. The crate structure, security posture, and test distribution (226 tests in mempalace-text alone) suggest someone who understands both the domain and Rust well. The questions above aren't blockers on the engineering quality — they're blockers on merge readiness. If the Rust CI is clean and the Python migration story is addressed, this is a legitimate v4.0 candidate.

nanoscopic:
Is there some justification as to why?

Why not just fork the project?

I really doubt those creating this project are going to accept this PR, so why make it?

web3guru888:
@nanoscopic — fair question. A few reasons contributors might submit a total rewrite as a PR rather than a fork:

1. **Patch-based contribution model**: Contributing to the upstream gives the author visibility and community feedback before splitting. If the maintainers reject it, they can then fork with a clear record of the attempt.
2. **Test harness reuse**: pronitdas built 403 tests that validate parity with the Python behavior — that's useful to the upstream regardless of the language direction.
3. **Signaling**: Sometimes a rewrite PR is a proposal, not an expectation. It says "the codebase has grown to where Rust is worth considering" more forcefully than a GitHub issue would.

Whether the maintainers will merge it is their call. But the audit work embedded here (the module mapping, dialect normalization, entity detection port) documents the system's architecture in ways that benefit the Python codebase too. Not wasted regardless of outcome.

nanoscopic:
In order to accept this PR all the other PR's submitted already would have to be effectively ditched.

While I still think this project is junk, that doesn't mean that people aren't trying to add value to it. What reasoning at all would there be to ditch all the other contributions in favor of moving to rust?

I just don't see the reasoning.

It looks clear that this is just another hanger on. Oh! I can show off that I rewrote it in Rust! Ok... great... we know vibe coding can do that. So what.

The best possible future I can see for this project is for it to cease being a joke and actually be mutated into something of value by people who are actually trying to make it work. That's going to be a hard path, but hard choices will need to be made to accomplish that.

I don't think moving to another language entirely is the best idea, and especially not rust.

So my ask of this particular PR is for motivation and reasoning to be provided.

Give some compelling reason why it should be considered at all.

igorls:
There are several solid contributions being made, so moving the entire codebase to another language now, effectively reverting all the hard work people put into it and moving to a language less familiar to many is not a good move. 

This should be closed. I appreciate the effort you or your agent put into this, but this time and tokens is better spent on understanding the architecture and trying to improve it, leading the project to the right path, instead of suggesting a port just for the sake of contributing.

Issue #538, opened 2026-04-10
MCP server stdio transport: add_drawer and kg_add write to WAL but not to ChromaDB/SQLite


duha887b:
## Environment

- OS: Windows 11 (bash shell via Claude Code)
- Python: 3.12 (via uvx)
- MemPalace: 3.1.0
- ChromaDB: 0.6.3
- Claude Code: 2.1.100
- MCP transport: stdio

## Bug

When calling `mempalace_add_drawer` or `mempalace_kg_add` via the MCP server (stdio transport), operations are logged in the WAL (`~/.mempalace/wal/write_log.jsonl`) with `"result": null`, but **no data is written to ChromaDB or the Knowledge Graph SQLite**.

Calling the same functions directly from Python works correctly:

```python
from mempalace.mcp_server import tool_add_drawer, tool_kg_add

# This works — drawer is written and searchable
result = tool_add_drawer('myproject', 'general', 'test content')
# {'success': True, 'drawer_id': '...'}

# This works — triple is written
result = tool_kg_add('User', 'prefers', 'dark mode', '2026-04-10')
# {'success': True, 'triple_id': '...'}
```

But via MCP stdio (Claude Code session), the WAL shows:

```json
{"timestamp": "...", "operation": "kg_add", "params": {"subject": "peter", "predicate": "preferred_temperature", "object": "21 Grad Celsius"}, "result": null}
{"timestamp": "...", "operation": "add_drawer", "params": {"drawer_id": "...", "wing": "personalai", "room": "kira", "content_length": 417}, "result": null}
```

After the session:
- `mempalace status` shows unchanged drawer counts
- `mempalace search` does not find MCP-added content
- Knowledge Graph SQLite has 0 entities, 0 triples
- Direct Python calls to the same functions DO write successfully

## Steps to Reproduce

1. Install mempalace plugin: `claude plugin install --scope user mempalace`
2. Mine a project: `mempalace mine /path/to/project`
3. Start Claude Code session, ask Claude to save a fact (triggers `mempalace_kg_add`)
4. Check WAL: `cat ~/.mempalace/wal/write_log.jsonl` — shows `"result": null`
5. Check KG: `mempalace search` — fact not found
6. Direct Python test (same function, same palace path) — works

## Suspected Cause

The MCP server runs as a stdio subprocess. Possible issues:
- ChromaDB connection not being committed/flushed in the stdio event loop
- SQLite WAL mode not syncing before process exit
[...]

PR #539, opened 2026-04-10
fix: security hardening, performance, and input validation

anthonyonazure:
## Summary

Comprehensive audit of the MCP server surface area, addressing security, performance, and functional issues:

- **Security (5 fixes):** Sanitize all KG tool inputs via `sanitize_name()`, validate date/direction params, replace WAL plaintext content previews with SHA-256 hashes, sanitize error messages returned to MCP clients to prevent leaking system paths and DB internals
- **Performance (2 fixes):** Add 2-second TTL metadata cache so `status`/`wings`/`rooms`/`taxonomy` share one collection scan instead of four independent 10K fetches; pass cached ChromaDB collection to `search_memories()` instead of creating a new `PersistentClient` per call
- **Functional (4 fixes):** Fix drawer ID hash collision (hash full content instead of `content[:100]`), remove duplicate cache variable initialization, replace all silent `except: pass` with `logger.error()`, expose `query_relationship()` as `mempalace_kg_query_relationship` MCP tool

### Files changed

| File | Changes |
|------|---------|
| `mempalace/mcp_server.py` | Input validation, metadata cache, WAL hashing, error sanitization, new tool |
| `mempalace/searcher.py` | Accept optional `collection` param, sanitize error message |
| `tests/conftest.py` | Clear metadata cache between tests |
| `tests/test_mcp_server.py` | +16 new tests (input validation, error sanitization, WAL, cache, new tool) |
| `tests/test_searcher.py` | Update error message assertion to match sanitized output |

### Breaking changes

- WAL log entries now use `content_hash` instead of `content_preview` — tools that parse the WAL for content previews will need updating
- `search_memories()` has a new optional `collection` parameter (backwards compatible)
- Error messages returned to MCP clients are now generic (no longer contain raw exception text)

## Test plan

- [x] All 554 tests pass (up from 534 — 16 new tests added)
- [x] Ruff lint clean
- [x] Verified path traversal, null bytes, invalid dates, and invalid direction all rejected
- [x] Verified drawer ID uniqueness with same-prefix different-suffix content
- [x] Verified WAL no longer contains plaintext content
- [x] Verified error messages don't leak system paths
- [x] Verified metadata cache invalidates on writes
- [x] Verified `mempalace_kg_query_relationship` tool registered and functional

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -18,11 +18,12 @@ Tools (write):
 """
 
 import argparse
-import os
-import sys
+import hashlib
 import json
 import logging
-import hashlib
+import os
+import sys
+import time as _time
 from datetime import datetime
 from pathlib import Path
 
@@ -100,16 +101,40 @@ def _wal_log(operation: str, params: dict, result: dict = None):
         logger.error(f"WAL write failed: {e}")
 
 
-_client_cache = None
-_collection_cache = None
-
-
 def _get_client():
     """Return a singleton ChromaDB PersistentClient."""
     global _client_cache
     if _client_cache is None:
         _client_cache = chromadb.PersistentClient(path=_config.palace_path)
     return _client_cache
+
+
+# ── Metadata cache (avoids repeated full-collection scans) ──────────────
+_META_CACHE = None  # list of metadata dicts
+_META_CACHE_TIME = 0.0
+_META_CACHE_TTL = 2.0  # seconds — short enough to stay fresh, long enough
+                        # to coalesce multiple read-tool calls in one batch
+
+
+def _get_all_metadata(col):
+    """Return all metadata from the collection, using a short-lived cache."""
+    global _META_CACHE, _META_CACHE_TIME
+    now = _time.monotonic()
+    if _META_CACHE is not None and (now - _META_CACHE_TIME) < _META_CACHE_TTL:
+        return _META_CACHE
+    try:
+        _META_CACHE = col.get(include=["metadatas"], limit=10000)["metadatas"]
+    except Exception as e:
+        logger.error("Failed to retrieve metadata: %s", e)
+        _META_CACHE = []
+    _META_CACHE_TIME = now
+    return _META_CACHE
+
+
+def _invalidate_meta_cache():
+    """Clear metadata cache after writes."""
+    global _META_CACHE
+    _META_CACHE = None
 
 
 def _get_collection(create=False):
@@ -122,7 +147,8 @@ def _get_collection(create=False):
         elif _collection_cache is None:
             _collection_cache = client.get_collection(_config.collection_name)
         return _collection_cache
-    except Exception:
+    except Exception as e:
+        logger.error("Failed to get ChromaDB collection: %s", e)
         return None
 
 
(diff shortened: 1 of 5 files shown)


web3guru888 (approved):
Comprehensive audit with real fixes across security, performance, and correctness. This is exactly the kind of hardening the MCP surface needed.

**Approved with one minor concern flagged below.**

---

**What's solid:**

- **Metadata cache** (`_META_CACHE`, 2s TTL): Smart approach. Coalescing `status`/`wings`/`rooms`/`taxonomy` from four independent 10K scans into one cached fetch is a real win for installations with large palaces. Cache invalidation on `tool_add_drawer`, `tool_delete_drawer`, and `tool_diary_write` looks complete — all three write paths correctly call `_invalidate_meta_cache()`.

- **`drawer_id` hash over full content** (removing `[:100]` truncation): This was a latent collision bug. Two drawers differing only after character 100 in the same wing/room would have been silently overwritten. Correct fix.

- **WAL content → SHA-256 hash**: Right call. WAL files are often left world-readable and could contain sensitive content. The test in `TestWALSecurity` covers exactly the right scenario.

- **Error message sanitization**: `"Search failed"` / `"Duplicate check failed"` / `"Failed to file drawer"` are generic without being useless. Good pattern.

- **`_validate_date` + `sanitize_name` on KG tools**: Path traversal rejection tested, date format validated. The `direction` enum check is clean.

- **`tool_kg_query_relationship`**: Useful new tool — querying by predicate is a common pattern and it was missing. Validated inputs, follows the same `count/facts` response shape as `tool_kg_query`.

- **`collection=col` passed to `search_memories`**: Avoids creating a second `PersistentClient` per search call, which was the root cause of the WAL flush issue in #538.

**One concern:**

`tool_list_rooms` previously pushed `where={"wing": wing}` to ChromaDB when a wing filter was specified. The new version fetches all metadata from cache and filters in Python. For the common case (single wing query, 10K drawer palace) this is slower — you load all metadata then discard most of it. The cache helps when `tool_list_rooms` is called right after `tool_status`, but not when called standalone with a wing argument.

Suggest keeping the ChromaDB `where` filter for the non-cached path when `wing` is specified, or accepting this as a known tradeoff in the PR description (the 2s TTL means it'll often be cached anyway in burst-query scenarios).

Not a blocker — correctness is preserved. Just worth documenting.

---

260 new test lines covering the new behaviors: input validation, cache invalidation, WAL hashing, error sanitization, and the new relationship query tool. That's the right way to ship a hardening PR.

web3guru888 (commented):
Good set of hardening changes — a few notes from our integration work:

**Metadata cache** — the concept is solid. Calling `tool_status()`, `tool_list_wings()`, and `tool_list_rooms()` in the same agent loop currently triggers redundant full-collection scans, so caching the result for 2s is the right shape. One caveat: `_get_all_metadata()` still hard-codes `limit=10000`, so the cache is caching a potentially-truncated snapshot rather than a complete one. For palaces beyond 10K drawers this doesn't fix the underlying bug — it just makes the wrong answer faster. The paginated fetch pattern from #484/`_fetch_all_metadata()` should be the data source here. Cache the complete result, not the clipped one.

**Cache invalidation** — `_invalidate_meta_cache()` called after writes is exactly right. Stale cache after an `add`/`upsert` is a nasty class of bug (agent lists rooms, doesn't see the one it just wrote, writes again). This is the part of the implementation I'd prioritize keeping.

**Security: error message sanitization** — good call. Leaking ChromaDB stack traces or raw file paths to MCP clients is a real concern in multi-tenant or remotely-hosted setups. The sanitized surface should also cover the new `sanitize_name()` call sites — make sure malformed names don't bubble back through exception paths.

**add() → upsert() in convo_miner.py** — same fix as mrdeeme's #542 and the pattern correction in #298. This one keeps appearing across forks, which suggests it should be part of the official migration guide somewhere ("if you're on <0.6, you were calling add() — switch to upsert()").

**Overlap with #540/#542** — there's meaningful surface overlap with #540 (KG path injection fix) and #542 (comprehensive hardening including the same upsert fix + Unicode normalization). Worth coordinating with those authors or the maintainer to decide which changes land where, so they don't conflict on merge.

bensig:
Overlaps with #387 (security hardening) and #542. The KG sanitization and WAL hash changes are good ideas — if #542 doesn't cover them, a focused follow-up PR would be welcome.

PR #542, opened 2026-04-10
fix: security hardening, test expansion (640→719), and 5 open issue fixes (#535, #536, #521, #478, #531)

### LINKED ISSUE #478: BUG: Status/taxonomy MCP tools silently truncate at 10,000 drawers — wrong counts for large palaces

### LINKED ISSUE #521: hnswlib updatePoint race on modified-file re-mine: EXC_BAD_ACCESS in repairConnectionsForUpdate (macOS ARM64, Python 3.13, chromadb 0.6.3)

### LINKED ISSUE #531: Readme chromadb version differs from actual
Readme says chromadb >= 0.4 is required.

pyproject.toml says >= 0.5

Which is it? Does 0.4 work?

### LINKED ISSUE #535: UnicodeEncodeError in miner.py / convo_miner.py / split_mega_files.py on Windows cp1251/cp1252 (incomplete fix of #47)
## Summary

Issue #47 (closed) fixed the `UnicodeEncodeError` crash caused by box-drawing characters in `searcher.py` on Windows. The same class of crash is still present in several other modules that print `✓` (U+2713) or other non-ASCII glyphs, and there is no `sys.stdout.reconfigure(encoding="utf-8", ...)` anywhere in the package.

## Reproduction

Windows 11, Python 3.13, default console encoding cp1251, mempalace 3.0.14.

```bash
# Crash 1: mempalace init
mempalace init --yes D:/path/to/project
```

Crashes while printing detected-entities confirmation block (Cyrillic content from project docs):

```
File "C:\Program Files\Python313\Lib\encodings\cp1251.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 30-34: character maps to <undefined>
```

```bash
# Crash 2: mempalace mine (project mode)
mempalace mine D:/path/to/project --wing vibeOps
```

```
  print(f"  ✓ [{i:4}/{len(files)}] {filepath.name[:50]:50} +{drawers}")
  File "cp1251.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u2713' in position 2: character maps to <undefined>
```

```bash
# Crash 3: mempalace mine --mode convos
mempalace mine C:/Users/.../.claude/projects/... --mode convos --extract general
```

Same `\u2713` crash from `convo_miner.py:359`.

## Affected print sites (U+2713 checkmark)

- `mempalace/miner.py:596` — `print(f"  ✓ [{i:4}/{len(files)}] ...")`
- `mempalace/convo_miner.py:359` — `print(f"  ✓ [{i:4}/{len(files)}] ...")`
- `mempalace/split_mega_files.py:227` — `print(f"  ✓ {name}  ...")`

Plus anywhere in `entity_detector.py` that prints user project text (Cyrillic / non-ASCII file contents) during init's confirmation step.

`grep reconfigure\(encoding .../mempalace/` returns zero hits — the #47 fix is not applied anywhere in the installed package.

## Workaround

```bash
set PYTHONIOENCODING=utf-8
set PYTHONUTF8=1
```

Both commands then succeed. But this is a bad UX — the tool ships crashing on default Windows consoles for any user whose project directory contains non-ASCII text or any mining run that prints a progress checkmark (i.e. all of them).

## Proposed fix

Call this once at the top of `cli.py::main()` (so every subcommand benefits):

```python
import sys
if sys.platform == "win32" and hasattr(sys.stdout, "reconfigure"):
    sys.stdout.reconfigure(encoding="utf-8", errors="replace")
    sys.stderr.reconfigure(encoding="utf-8", errors="replace")
```

This is exactly the suggested fix from #47 but applied at the CLI entry point instead of inside one module.

## Environment

- Windows 11, Python 3.13.x
- mempalace 3.0.14
- Console: default Windows Terminal (cp1251 on Russian locale, cp1252 on en-US)

### LINKED ISSUE #536: `--extract general` dumps Markdown-bold content into `emotional` room via overly broad `\*[^*]+\*` regex
## Summary

The `EMOTION_MARKERS` list in `mempalace/general_extractor.py` contains a wildcard regex that matches **any** text wrapped in single asterisks:

```python
EMOTION_MARKERS = [
    r"\blove\b", r"\bscared\b", r"\bafraid\b", ...
    r"\*[^*]+\*",   # <-- this one
]
```

This pattern matches every Markdown `*italic*` and every inner pair of `**bold**`. Because assistant responses in technical conversations use bold for emphasis, command names, variable names, section headers, etc., practically every non-trivial paragraph scores non-zero against `emotional`. Since classification picks `max(scores)` and most technical paragraphs trigger only this one marker (no decision/problem/milestone keywords), they all land in the `emotional` room.

## Reproduction

Mined 273 Claude Code session JSONLs (entirely technical DevOps content — Kubernetes, Helm, Elasticsearch, GitLab, Jira) with:

```bash
mempalace mine C:/Users/.../.claude/projects/<project>/ \
  --mode convos --extract general --wing vibeOps
```

Result:

```
  Done.
  Files processed: 273
  Drawers filed: 2443

  By room:
    emotional            1615 files   <-- 66% of all drawers
    milestone            387 files
    decision             322 files
    problem              114 files
    preference           5 files
```

Spot-check of `mempalace search <any> --wing vibeOps --room emotional` returns results like:

- "Sampling: Type: parentbased_traceidratio — если у трейса есть родитель..."
- "Namespace: opentelemetry · Имя: instrumentation"
- Tables of Jira statuses
- OpenTelemetry collector endpoint lists

Zero emotional content. Every one of these paragraphs only matches `\*[^*]+\*` via Markdown bold in the source text.

## Root cause

`general_extractor.py` line 160 (mempalace 3.0.14):

```python
EMOTION_MARKERS = [
    ...
    r"\*[^*]+\*",
]
```

This regex is too greedy for any text that contains Markdown formatting. It was presumably intended to catch things like `*whispers*` or `*sighs*` emotes in personal chat logs, but in developer contexts it catches every `*foo*` / `**bar**`.

## Proposed fixes (in order of preference)

1. **Remove the regex entirely.** Users who want emote detection can opt in via a separate marker list. `\blove\b`, `\bscared\b`, etc. are already precise enough.
2. **Require a surrounding word boundary that excludes code-like content**, e.g. `r"(?<!\w)\*[a-z][^*]{2,}\*(?!\w)"` plus a blacklist of common programmer terms — but this is brittle.
3. **Strip Markdown before scoring** (`_extract_prose` already exists in the file; extend it to strip `**bold**` / `*italic*` / backtick code spans before running the regex scorer).
4. **Prefer non-emotional max when scores tie** (low-effort bias fix): if a paragraph scores equally on emotional and another type, pick the other type. Doesn't fix the root cause but reduces false positives.

Option (3) is probably the right long-term answer — the same issue affects the `\*` marker and will affect future markers that collide with Markdown syntax.

## Environment

- mempalace 3.0.14
- Source: 273 Claude Code conversation JSONL files (technical DevOps)
- Extractor mode: `--extract general`

mrdeeme:
## Summary

Comprehensive audit producing **12 code fixes**, **79 new tests** (640 → 719 total), and **5 additional bug fixes** for open issues. All 612 unit tests pass.

---

#### Security (`mcp_server.py`)
- Sanitize 8 `str(e)` leaks in error responses → generic messages + `logger.error`
- Remove duplicate `_client_cache` declaration
- Update docstring to reflect all 19 tools

#### Data Integrity (`convo_miner.py`)
- Replace `collection.add()` with `collection.upsert()` to prevent duplicate ID crashes
- Fix `diary_write` to use `upsert` instead of `add`

#### Concurrency (`knowledge_graph.py`)
- Add `threading.Lock` for thread-safe SQLite access in WAL mode

#### Performance (`palace_graph.py`)
- Add TTL cache (30s) + count-based invalidation for `build_graph()`
- Replace hardcoded `limit=10000` with `_iter_all_metadata()` batched pagination (1000/batch)
- Singleton ChromaDB client keyed by `palace_path` (was creating new client per call)

#### Error Handling (`mcp_server.py`)
- Replace bare `except Exception: pass` with proper error handling + `warning` field

#### Cleanup (`split_mega_files.py`)
- Remove hardcoded personal names from fallback list → empty list

### Test Suite Expansion (640 → 719 tests, +79 net new)

| Category | File | Tests |
|----------|------|-------|
| **New** | `test_palace.py` | 15 — `get_collection`, `file_already_mined`, `SKIP_DIRS` |
| **New** | `test_concurrency.py` | 7 — parallel search, concurrent KG writes, MCP |
| **New** | `test_failure_modes.py` | 16 — error paths: palace, config, searcher, KG, MCP |
| **Expanded** | `test_config.py` | 4 → 24 — precedence, sanitize_name/content, invalid configs |
| **Expanded** | `test_convo_miner.py` | 1 → 12 — decomposed monolith into focused scenarios |
| **Expanded** | `test_version_consistency.py` | 2 → 7 — semver, module consistency, server info |

#### Benchmark Assertion Thresholds (5 files)
- `test_memory_profile.py`: RSS growth limits (<100MB/<150MB)
- `test_search_bench.py`: concurrent error count (0) + p95 latency (<5s)
- `test_recall_threshold.py`: recall@10 ≥ 50% at ≤1000 drawers
- `test_palace_boost.py`: wing/room boost ≥ −10% degradation
- `test_chromadb_stress.py`: batch insertion faster than sequential

---

### #535 — UnicodeEncodeError on Windows (cp1251/cp1252)
Add `sys.stdout/stderr.reconfigure(encoding='utf-8', errors='replace')` at the CLI entry point (`cli.py`), fixing checkmark (U+2713) and CJK crashes across all subcommands.

### #536 — Overly broad emotion regex misclassifies 66% of content
Remove wildcard regex `\*[^*]+\*` from `EMOTION_MARKERS` in `general_extractor.py`. This matched all Markdown bold/italic, routing technical content to the `emotional` room.

### #521 — hnswlib SIGSEGV on modified-file re-mine (macOS ARM64)
Add `collection.delete(where={'source_file': ...})` before the upsert loop in `miner.py::process_file()`. Converts re-mines from upsert-over-existing (which hits hnswlib's thread-unsafe `updatePoint`/`repairConnectionsForUpdate`) into clean delete+insert.

### #478 — `miner.py::status()` truncates at 10,000 drawers
Replace hardcoded `limit=10000` with batched pagination (1000/batch), matching the fix already applied to `mcp_server.py`.

### #531 — README chromadb version mismatch
Update README from `chromadb>=0.4.0` to `chromadb>=0.5.0,<0.7` to match `pyproject.toml`.

---

## Validation

- **612 unit tests passed**, 0 failed, 1 skipped (Unix-only permissions on Windows)
- **719 total tests** collected (unit + benchmark)

Closes #535, closes #536, closes #521, closes #478, closes #531.


web3guru888 (commented):
**Review — PR #542 (mrdeeme)**

Solid contribution. Saw your threading work in #459 and this is a clear step up in scope. Here's my take:

**Strengths:**

- **Paginated `_iter_all_metadata()`** — this is the correct fix for #478. PR #539 took a caching approach that just cached the truncated result; batched pagination actually resolves the root cause. 1000-item batches are sane for typical palace sizes.
- **KG `threading.Lock()`** — concurrent MCP tool calls absolutely can race on SQLite writes. Good catch. The `_execute()`/`_executemany()`/`_commit()` wrapper pattern keeps the locking contained.
- **`add()` → `upsert()` in convo_miner** — this was flagged as a correctness issue back in the #298 thread. Right fix.
- **79 new tests** covering concurrency, failure modes, and benchmark thresholds — this alone would be worth merging. `test_concurrency.py` in particular fills a real gap.
- **#521 (hnswlib SIGSEGV)**: delete-before-upsert is the right approach given hnswlib's non-thread-safe `updatePoint`. Clean.

**Concerns:**

1. **`#536` overlap with #543** — both PRs remove `\*[^*]+\*` from `EMOTION_MARKERS`, but #543 goes further: it replaces with a more targeted emote pattern *and* adds `_strip_markdown()`. This PR's approach (bare removal) is simpler but may over-remove. The maintainer should resolve the overlap before merging either.

2. **Codacy instructions file** — `.github/instructions/codacy.instructions.md` is Copilot agent context, not repo documentation. It shouldn't be committed to the public repo. The `.gitignore` entry also uses a Windows-style backslash path separator (`instructions\codacy...`) which won't match on Linux/macOS — so it won't actually be ignored on those platforms even if it were intentional.

3. **Minor**: `split_mega_files.py` removing hardcoded personal names from the fallback list → empty list is correct (those shouldn't ship), but worth a changelog note if any users depend on that heuristic.

**Bottom line:** The core of this PR (pagination fix, KG locking, upsert, test suite) is merge-worthy. I'd ask for the Codacy file to be dropped and the `#536` approach coordinated with the author of #543 before merge.

*[MemPalace-AGI integration — 208 discoveries, 710 KG entities | [dashboard](https://mempalace-agi.taurus.app)]*

web3guru888:
Solid security hardening PR — let me break down what stands out.

**str(e) sanitization in MCP error responses** — This is important work. Error message leaks in MCP stdio responses can expose internal paths, file names, and—in the worst case—partial secrets embedded in exception strings. The pattern of switching to generic user-facing messages while routing full context to `logger.error` is exactly right. Glad to see this applied consistently across 8 call sites rather than patched piecemeal.

**add() → upsert() in convo_miner.py** — Real correctness fix. Multiple mine runs on the same conversation would previously crash on duplicate IDs, which means any retry or re-index workflow was broken. Worth validating that the upsert is truly idempotent: if the same conversation produces the same content hash, the upsert should be a no-op. If the hash changes between runs (e.g., due to timestamp inclusion), you'd get silent data mutation instead of a crash — which is harder to debug. Worth a quick sanity check on the hashing logic.

**threading.Lock for SQLite in WAL mode** — Correct. WAL provides read/write concurrency at the SQLite level, but Python-level threading still needs coordination around the connection object. The lock closes that gap.

**Issue coverage** — Fixing #535, #536, #521, #478, and #531 in one pass is efficient for maintainers. The 79 new tests (640 → 719) are the right safety net for a wide-surface diff. One caution though: mega-PRs that touch security, concurrency, and data integrity simultaneously can obscure interactions between changes. I'd suggest the reviewer runs the full test suite on just the security-only commit (the str(e) sanitization) in isolation — confirming the error-suppression changes don't accidentally mask real failures during development. It's easy for a generic-message wrapper to hide a logic error if the surrounding test doesn't assert on the error path.

**Overall**: The str(e) sanitization alone would be worth merging; getting five issue fixes in the same sweep with full test coverage is a real contribution. Nice work @mrdeeme.

---

bensig:
@mrdeeme conflicts here

bensig:

PR #544, opened 2026-04-10
fix: prevent HNSW index bloat from duplicate add() calls (#525)

### LINKED ISSUE #525: HNSW link_lists.bin grows to terabytes, causes segfault and APFS orphaned blocks on macOS
## Summary

On macOS (Apple Silicon, APFS), the HNSW index file `link_lists.bin` can grow to **2.8TB** (apparent) / **~1.7TB** (allocated), causing segfaults on any `search` or `mine` operation. Deleting the corrupted file leaves orphaned APFS block allocations that can only be reclaimed via Recovery Mode Disk Utility First Aid.

## Environment

- macOS 15 (Darwin 25.3.0), Apple Silicon
- mempalace 3.0.0 (also reproduced path to corruption on 3.1.0's underlying chromadb)
- chromadb 0.6.3, chroma-hnswlib 0.7.6
- Palace size: 53,222 drawers across 10,000+ rooms
- Filesystem: APFS

## Steps to reproduce

1. Build a large palace (50K+ drawers) via repeated `mempalace mine` runs
2. Over time, `link_lists.bin` in the HNSW directory grows unbounded
3. ChromaDB logs `Add of existing embedding ID: drawer_*` warnings on every access — it's reconciling/re-adding existing embeddings into the HNSW graph
4. Eventually `mempalace search` segfaults during HNSW index load

## Observed behavior

```
$ mempalace search "test query"
Add of existing embedding ID: drawer_claude_sessions_technical_e9de31813119608b
Add of existing embedding ID: drawer_claude_sessions_technical_1e43ec1539179dfc
[... ~50 more warnings ...]
[1]    99847 segmentation fault  mempalace search "test query"
```

File sizes in the palace directory:
```
-rw-r--r--  2.8T  link_lists.bin      # HNSW index (corrupted/bloated)
-rw-r--r--   84M  data_level0.bin     # HNSW data (normal)
-rw-r--r--  276M  chroma.sqlite3      # All actual data (intact)
```

## Secondary issue: APFS orphaned blocks

After deleting the corrupted HNSW directory and running `mempalace repair`, the APFS Data volume still reports ~1.9TB consumed while `du` only accounts for ~200GB. `diskutil verifyVolume` confirms corruption:

```
The volume /dev/rdisk3s5 was found to be corrupt and needs to be repaired
```

Live repair cannot reclaim the orphaned blocks — requires booting into Recovery Mode and running First Aid on the Data volume. This means users who hit this bug lose effective disk space until they do an offline filesystem repair.

## Workaround

1. Delete the HNSW index directory (the UUID-named folder inside the palace, NOT `chroma.sqlite3`)
2. Delete any partial `.backup` from a failed `mempalace repair`
3. Run `mempalace repair` to rebuild the index from SQLite
4. Boot into Recovery Mode → Disk Utility → First Aid on the Data volume to reclaim orphaned APFS blocks

## Suggestions

- **Guard against unbounded growth**: Check `link_lists.bin` size relative to drawer count before/after mining. A 53K-drawer palace should have an HNSW index of ~100-200MB, not terabytes.
- **Repair command**: Skip the full `shutil.copytree` backup when disk space is low — the SQLite DB is the source of truth and is untouched by the rebuild. Consider backing up only `chroma.sqlite3` instead of the entire palace directory (which includes the bloated HNSW files).
- **`upsert` vs `add`**: The "Add of existing embedding ID" warnings suggest embeddings are being `add`ed when they already exist. Using `upsert` or checking for existence first would prevent the HNSW graph from accumulating duplicate entries.

## Related

- #74 (macOS ARM64 segfault)
- #100 (ChromaDB version pinning)

milla-jovovich:
## Summary

- **Root cause**: `convo_miner.py` used `collection.add()` instead of `upsert()`, so repeated mine runs pushed duplicate entries into the HNSW graph. At scale (50K+ drawers) this causes `link_lists.bin` to grow to terabytes and eventually segfault.
- **`convo_miner.py`**: one-line fix, `add()` → `upsert()`
- **`repair.py`**: new module — scan for corrupt IDs, prune them, or rebuild the HNSW index from scratch. Backs up only `chroma.sqlite3` (not the bloated HNSW files). Recreates collection with `hnsw:space=cosine`.
- **`dedup.py`**: new module — detect and remove near-duplicate drawers from the same source file using cosine similarity. No API calls.

## Test plan

- [ ] Run `mempalace mine` twice on the same directory, confirm no duplicate drawers created
- [ ] Run `python -m mempalace.repair scan` on a palace with known corrupt IDs
- [ ] Run `python -m mempalace.dedup --dry-run` to verify duplicate detection without deletion
- [ ] Verify `rebuild` backs up only `chroma.sqlite3`, not HNSW files

Fixes #525

Thank you from Milla and Lu ✨

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -436,6 +436,16 @@ def process_file(
         print(f"    [DRY RUN] {filepath.name} → room:{room} ({len(chunks)} drawers)")
         return len(chunks), room
 
+    # Purge stale drawers for this file before re-inserting the fresh chunks.
+    # Converts modified-file re-mines from upsert-over-existing-IDs (which hits
+    # hnswlib's thread-unsafe updatePoint path and can segfault on macOS ARM
+    # with chromadb 0.6.3) into a clean delete+insert, bypassing the update
+    # path entirely.
+    try:
+        collection.delete(where={"source_file": source_file})
+    except Exception:
+        pass
+
     drawers_added = 0
     for chunk in chunks:
         added = add_drawer(
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,13 @@
+# Default owners for everything
+* @milla-jovovich @bensig @igorls
+
+# Core library
+mempalace/ @milla-jovovich @bensig
+
+# CI and workflows
+.github/ @bensig
+
+# Plugins and integrations
+.claude-plugin/ @bensig
+.codex-plugin/ @bensig
+integrations/ @bensig
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - run: pip install -e ".[dev]"
-      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80
+      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10
 
   test-windows:
     runs-on: windows-latest
@@ -28,7 +28,7 @@ jobs:
         with:
           python-version: "3.9"
       - run: pip install -e ".[dev]"
-      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80
+      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10
 
   test-macos:
     runs-on: macos-latest
@@ -38,7 +38,7 @@ jobs:
         with:
           python-version: "3.9"
       - run: pip install -e ".[dev]"
-      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80
+      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10
   lint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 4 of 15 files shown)


web3guru888 (commented):
This is exactly the right fix for #525. The `add()` → `upsert()` change in `convo_miner.py` is a one-liner but it's load-bearing — we hit this exact HNSW bloat issue in our own setup around 40K drawers where repeated mine runs were pushing duplicates into the graph.

A few observations on the new modules:

**repair.py:**
- The scan → prune → rebuild progression is well-designed. The decision to back up only `chroma.sqlite3` and skip the bloated HNSW files is correct — that's where the actual drawer data lives.
- One gap: `rebuild_index()` does a full collection delete and recreate, which will also lose any embeddings already computed. Users should know that after `rebuild`, a full re-mine or re-embedding pass will be triggered by ChromaDB's lazy embedding. This is expected behavior but worth noting in the docstring or output message.
- `shutil` is imported in `rebuild_index` but I don't see it in the imports at the top — worth checking.

**dedup.py:**
- The greedy longest-first approach is sound for same-source dedup. One edge case: if the same source file was mined under different wings (multi-domain setup), the `source_file` grouping won't catch cross-wing duplicates. The `--source` filter helps but a `--wing` scope filter like in `repair.py` would be useful here too.
- `DEFAULT_THRESHOLD = 0.15` (cosine distance) is appropriate for near-identical chunks. For looser dedup of paraphrased content, users may want 0.3–0.4 — worth documenting the threshold semantics (cosine *distance*, not similarity).

Overall: the core fix is correct, the new tooling is genuinely useful for anyone who's been running MemPalace long enough to accumulate HNSW bloat. Strong +1.

milla-jovovich:
@web3guru888 sorry, I'm learning to use GitHub on the fly. sorry about the the "shut" not being listed. I appreciate you bearing with us, it;s just ben and I doing all this and I'm just learning.🤦🏻‍♀️

milla-jovovich:
sorry shutil

milla-jovovich:
@web3guru888 shutil is actually on line 32

web3guru888:
You're right — I see it now at line 32 of `repair.py` (`import shutil`). My mistake, I was scanning the diff and missed it. That nit is a non-issue — the import is present and correct. Apologies for the noise! Everything else in my review stands.

And no need to apologize — you and ben are moving fast and building something people clearly care about. The community response speaks for itself. 👍

bensig (approved):

mempalace/miner.py
```diff
@@ -436,6 +436,16 @@ def process_file(
...
 
+    # Purge stale drawers for this file before re-inserting the fresh chunks.
+    # Converts modified-file re-mines from upsert-over-existing-IDs (which hits
+    # hnswlib's thread-unsafe updatePoint path and can segfault on macOS ARM
+    # with chromadb 0.6.3) into a clean delete+insert, bypassing the update
+    # path entirely.
+    try:
+        collection.delete(where={"source_file": source_file})
+    except Exception:
+        pass
```
robvdl, line 447: I don't know enough about this collection type, but I notice constant blanket `try ... except Exception`.

Is there a better exception or exception base class that can be caught instead? Blanket try except (especially with pass) is kinda discouraged as anything can happen (out of RAM, out of disk, etc).

WalterUpgrade:
@milla-jovovich You're killing it!

PR #547, opened 2026-04-10
test: make search_no_results assertion deterministic

Ocheretovich:
test: guard against empty results in test_result_fields

### DIFF
--- a/tests/test_searcher.py
+++ b/tests/test_searcher.py
@@ -99,13 +99,10 @@ class TestSearchCLI:
         with pytest.raises(SearchError, match="No palace found"):
             search("anything", str(tmp_path / "missing"))
 
-    def test_search_no_results(self, palace_path, collection, capsys):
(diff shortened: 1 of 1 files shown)


web3guru888 (commented):
Clean, targeted fix. The original `assert result is None or "No results" in captured.out` was testing two different behavioral contracts simultaneously — it passed regardless of which branch the implementation took, making the assertion essentially useless as a regression signal.

The new assertion pins `search()` to a single contract: return `None` when no results are found. That's the right semantics. If the implementation later drifts to raising an exception or returning an empty list, the test will catch it.

One thought: the `capsys` fixture removal is correct *given this contract*, but it's worth verifying that `search()` doesn't *also* print to stdout when results are empty (that would be a behavioral change worth noting, not a bug). If the current implementation is silent and just returns `None`, this is already documented by the test passing. LGTM.

PR #548, opened 2026-04-10
feat(cli): add clean subcommand to remove wings and rooms

brandonhon:
## Summary

Adds a new `mempalace clean` subcommand for reclaiming palace space and starting projects over. It removes drawers at two granularities — an entire wing, or a specific room within a wing — and keeps the `mempalace_drawers` and `mempalace_compressed` collections in lockstep so compressed copies never go stale after a cleanup.

## Motivation

Today there is no first-class way to remove memories once a project is done. Users either re-mine forever or hand-edit ChromaDB. The new subcommand gives a safe, auditable way to scrub scoped memory without touching the rest of the palace.

## UX

```
mempalace clean --wing my_app                    # remove all drawers in a wing
mempalace clean --wing my_app --room costs       # remove all drawers in one room
mempalace clean --wing my_app --dry-run          # preview counts, no changes
mempalace clean --wing my_app --yes              # skip interactive confirmation
```

Sample output:

```
  MemPalace Clean

  Palace: /home/user/.mempalace/palace
  Target: wing 'my_app', room 'backend'
  Drawers to remove:    12
  Compressed to remove: 12

  Delete 12 drawer(s) from wing 'my_app', room 'backend'? [y/N] y

  Removed 12 drawer(s) from wing 'my_app', room 'backend'.
  Removed 12 compressed drawer(s).

  Note: the knowledge graph (~/.mempalace/knowledge_graph.sqlite3) is not touched by clean.
```

## Design decisions

- **`--wing` is required.** No implicit "clean everything" — refuses to run without a wing filter. This is intentional: the cost of an accidental full wipe is very high.
- **Single flag per scope.** Only `--wing` and `--room`. I considered splitting into `--room` (clear contents) vs `--del-room` (remove the room itself), but rooms and wings are metadata-only in mempalace today — they exist only as fields on drawer metadata, not as first-class records. Once the last drawer in a room is deleted the room disappears from `list_rooms` automatically, so "clear contents" and "delete room" would produce identical observable state. Added a registry was deemed scope creep.
- **Knowledge graph left untouched.** KG facts in `~/.mempalace/knowledge_graph.sqlite3` do not carry a `wing` field, so they can't be filtered cleanly by wing without false positives. The `--help` text explicitly notes this and the runtime output reminds the user after every cleanup.
- **Compressed collection cleaned in lockstep.** `mempalace_compressed` shares drawer IDs with `mempalace_drawers`. Leaving compressed copies behind after a wing/room cleanup would be a correctness bug, so both collections are filtered and deleted in the same operation.
- **Single WAL entry per cleanup.** A `clean_wing` or `clean_room` record is appended to `~/.mempalace/wal/write_log.jsonl` with `{wing, room, drawers_deleted, compressed_deleted}`. Per-drawer logging was rejected as noisy for bulk ops.
- **Interactive confirmation by default.** Prompts `[y/N]` with a count summary before deleting. `--yes` skips the prompt for scripting. `--dry-run` shows counts and exits without touching anything.
- **Safe degraded operation.** If `mempalace_compressed` doesn't exist (palace was never compressed), the cleanup proceeds against `mempalace_drawers` only without error.

## Files changed

- `mempalace/palace.py` — added `_build_where`, `count_drawers`, `delete_drawers` helpers (+34 lines). Keeps the CLI layer thin and makes the primitives reusable for a future MCP `clean` tool.
- `mempalace/cli.py` — added `cmd_clean`, the `clean` subparser, and a dispatch entry. Docstring updated with the new command in the help epilog (+142 lines).
- `tests/test_clean.py` — 15 new tests (+221 lines).

## Tests

Added `tests/test_clean.py` with 15 tests covering:

**Primitive layer (`palace.count_drawers` / `delete_drawers`):**
- Counting by wing
- Counting by wing + room
- Deleting by wing leaves other wings untouched
- Deleting by room leaves sibling rooms untouched
- Nothing-to-delete returns zero

**CLI layer (`cli.cmd_clean`):**
- Cleaning a wing removes matching drawers from both `mempalace_drawers` and `mempalace_compressed`
- Cleaning a room scopes correctly and preserves sibling rooms
- `--dry-run` makes zero changes
- No-match scenario prints "Nothing to clean" and exits 0
- Interactive `n` at the prompt cancels the delete
- Interactive `y` at the prompt proceeds
- Works without a compressed collection present
- Writes a `clean_room` WAL entry with the expected counts
- Missing `--wing` flag exits non-zero
- Missing palace directory exits non-zero with a clear message

### Test results

```
$ python3 -m pytest -x -q
549 passed, 106 deselected in 8.83s
```

All 549 tests in the suite pass (15 new, 534 existing). No regressions.

## Test plan

- [x] `python3 -m pytest tests/test_clean.py -v` passes locally
- [x] `python3 -m pytest -q` — full suite still green
- [x] `python3 -m mempalace clean --help` shows the expected usage
- [x] Dry-run against a real palace: `mempalace clean --wing <some-wing> --dry-run` reports accurate counts
- [x] Actual cleanup against a disposable palace, then verify via `mempalace status` or `mempalace_list_rooms` that the wing/room is gone
- [x] Confirm a WAL entry landed in `~/.mempalace/wal/write_log.jsonl` with `operation: clean_wing` (or `clean_room`) and correct counts
- [x] If the palace has compressed drawers, confirm `mempalace_compressed` was cleaned in lockstep

## Notes for reviewers

- The "knowledge graph not touched" behavior is called out in both `--help` text and runtime output so users aren't surprised. If you want KG cleanup later, it would need a new wing tag on KG triples — out of scope for this PR.
- `delete_drawers` uses ChromaDB's native `delete(where=...)`. The `_build_where` helper produces either `{"wing": X}` or `{"$and": [{"wing": X}, {"room": Y}]}` depending on scope.
- Commit is authored via the `brandonhon` GitHub noreply email for email-privacy compliance.


web3guru888 (commented):
This fills a real gap — there's currently no first-class way to remove a wing once a project is done. Good motivation and clean scope.

**What works well:**
- Wing/room granularity is the right split. Wing-level clean covers the "project is done" case; room-level covers "re-mine after restructuring."
- `drawers` + `compressed` kept in lockstep. This is important — without it, compressed stale entries would survive a clean.
- WAL logging for the operation. Audit trail is the right default for a destructive command.
- Test coverage is thorough: dry-run, prompt cancel/confirm, no-compressed-collection graceful path, WAL verification. 11 tests is solid.
- `--yes` flag for scripted use.

**One issue with `count_drawers` + `delete_drawers` on large palaces:**

Both functions call `collection.get(where=...)` to count first, then `collection.delete(where=...)`. On a 10K+ drawer palace this means two full ChromaDB scans. The delete path could do `delete(where=...)` directly and get the count from the returned deletion count if ChromaDB exposes it — or just accept a single-pass "delete and count" by calling get with `include=[]` once and immediately deleting the returned IDs.

This is a performance concern, not a correctness concern. For typical palace sizes (hundreds to low thousands) it's not a problem. But worth noting.

**Knowledge graph note:**

The clean command correctly leaves the KG untouched and documents this explicitly. Worth considering whether `--with-kg` or a follow-up `mempalace clean-kg --wing ...` makes sense as a companion command — once you've deleted all drawers from a wing, the KG entities that only had provenance from those drawers become orphaned. Not blocking, just flagging for follow-on.

**Minor:**

The WAL entry uses `datetime.now()` (local time). Other WAL entries in the codebase use UTC (`datetime.utcnow()` or `datetime.now(timezone.utc)`). Worth being consistent.

Otherwise clean implementation. LGTM with those minor notes.

brandonhon:
@web3guru888
Thanks for the careful review — all three points are fair. I'll work on this over the weekend:

- Perf: I'll collapse the count-then-delete into a single get(where=..., include=[]) pass and delete by
the returned IDs so large palaces only scan once.
- KG orphans: good call on the follow-up — I'll look at whether mempalace clean-kg --wing ... is
feasible given KG triples don't carry a wing field today, or if --with-kg makes more sense. Might land
as a separate PR.
- WAL timestamp: easy fix, I'll switch to datetime.now(timezone.utc) to match the rest of the codebase.

Will push updates to this branch over the weekend.

web3guru888:
@brandonhon — all three resolutions sound right:

- **Single-pass scan** is the correct approach —  returns IDs at minimal cost, avoids the double-scan. For very large palaces with thousands of drawers, this will matter.
- **KG orphan as follow-on** makes sense given the triple-level scoping issue.  is probably the right flag name if KG triples don't carry wing metadata, since it's an opt-in behavior with its own semantics rather than a scope filter.
- **UTC timestamp** is a one-liner but it matters for log consistency when the palace spans timezones.

No rush — enjoy the weekend. The PR is solid as-is; these are all incremental improvements on an already correct implementation.

web3guru888:
This is a solid, well-thought-out addition. The design choices are right in all the places that matter.

**`--wing` required is the correct safety default.** Implicit full-palace wipes are exactly the kind of footgun that causes people to nuke prod palaces by accident. Requiring explicit scope is the right call, and the `--dry-run` + interactive confirm + `--yes` trifecta is the standard UX pattern for destructive CLI ops — good to see all three here.

**KG not touched is correct** given the current schema (KG entities have no `wing` field), but the `--help` note is important so users know they'll need to handle KG cleanup separately if they've built cross-wing entity links. This is especially relevant for anyone running a KG-heavy integration — don't want users to clean a wing and then wonder why their KG still references entities from deleted drawers.

**The `_build_where` helper is a good structural choice.** Keeps the CLI thin and makes it straightforward to add a `clean` tool to the MCP surface later without duplicating filter logic.

**Question/confirmation needed:** The PR notes that "Nothing to clean" + exit 0 covers the no-match case — but does this hold when `mempalace_compressed` is absent entirely (i.e., the user never enabled compression)? If the code tries to call `.delete(where=...)` on a collection that doesn't exist, that would be a different error path than "zero matches." The tests presumably cover this, but worth calling out explicitly.

**One note on HNSW:** For large deletions (whole-wing cleanup), ChromaDB's HNSW index handles removed vectors via tombstoning, but the graph can degrade meaningfully before the next compaction. Not a blocker for correctness, but for palaces with 100K+ drawers, users doing a large wing purge might notice search latency spikes until the index is rebuilt. Worth a brief mention in the `--help` text or docs — something like "For large deletions, consider running `mempalace rebuild-index` afterward."

**549 tests passing is strong.** The scope of this PR (new file + CLI integration + tests, zero regressions) shows good discipline. Well done.

---

brandonhon:
@web3guru888 
Thanks again for the review! I looked into this and the codebase actually uses naive local-time datetime.now() throughout — a grep for utcnow or timezone.utc returns zero matches. So this WAL entry is consistent with the existing convention rather than diverging from it. Happy to revisit if we want to migrate the whole codebase to UTC, but that feels out of scope for this PR.

igorls:

brandonhon:
Rebased onto `develop` — the conflict was a duplicate `--version` argparse registration, since #1474/#1487 era landed `--version` after this PR opened. Resolved by adding the `-V` short alias to the existing registration. PR is back to `MERGEABLE`.

The perf concern @web3guru888 raised (collapsing count-then-delete into one `get(where=..., include=[])` pass per collection) is in the rebased `351f28b`. KG-orphan cleanup will follow up in a separate PR — it needs a triple-schema change to scope by wing.

PR #373, opened 2026-04-09
fix: count Codex user_message turns in _count_human_messages (#347)

RhettOP:
## Summary

Fixes #347

The `_count_human_messages()` function in `mempalace/hooks_cli.py` previously only handled **Claude Code** transcript format:

```json
{"message": {"role": "user", "content": "..."}}
```

**Codex CLI** transcripts use a different schema:

```json
{"type": "event_msg", "payload": {"type": "user_message", "message": "..."}}
```

This meant the stop-hook auto-save threshold never triggered for Codex sessions because the human message count always returned 0.

## Fix

Added detection for the Codex `event_msg` format alongside the existing Claude Code detection. The function now handles both schemas correctly, including the `<command-message>` skip logic for Codex messages.

## Changes

- `mempalace/hooks_cli.py`: Extended `_count_human_messages()` to count Codex-style `user_message` entries in addition to Claude Code-style `user` role messages.

PR #316, opened 2026-04-09
fix: skip tool_result messages in _count_human_messages

martinezcarlos:
## What does this PR do?

`_count_human_messages()` in `hooks_cli.py` was counting tool result messages as human turns. In Claude Code's transcript format, tool results arrive as `role: "user"` messages with content blocks of `type: "tool_result"`. This inflated the message counter and triggered the AUTO-SAVE stop hook too frequently.

The fix adds a check inside the `elif isinstance(content, list)` branch to skip messages where all blocks are `tool_result` type.

## How to test

```bash
pytest tests/test_hooks_cli.py -v
```

Two new tests cover the fix:
- `test_count_skips_tool_result_messages` — pure tool_result messages are not counted
- `test_count_mixed_content_with_tool_result_counts` — mixed content (tool_result + text) still counts

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`)
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`)

PR #550, opened 2026-04-10
fix: skip tool_result messages in save hook exchange count

### LINKED ISSUE #549: fix: save hook counts tool_result messages as human messages, inflating exchange count
## Bug

The `Stop` hook in `mempal_save_hook.sh` (lines 87-104) counts every JSONL entry with `role: "user"` as a human message. In Claude Code transcripts, tool results (`Bash`, `Read`, `Grep`, subagent dispatches, etc.) are also sent as `role: "user"` with content blocks of `type: "tool_result"`.

This inflates the exchange count significantly. In one measured session, 425 actual human messages produced a count of 1,221 — a ~2.9x inflation. Tool-heavy and subagent-heavy sessions trigger saves far more often than intended.

## Current behavior

```python
if isinstance(msg, dict) and msg.get('role') == 'user':
    content = msg.get('content', '')
    if isinstance(content, str) and '<command-message>' in content:
        continue
    count += 1
```

Only `<command-message>` strings are filtered. Tool results (where `content` is a list of `{"type": "tool_result", ...}` blocks) are counted as human messages.

## Expected behavior

Only actual human text input should increment the counter. Tool results should be skipped.

## Proposed fix

When `content` is a list, check whether all blocks are `tool_result` type — if so, skip:

```python
if isinstance(msg, dict) and msg.get('role') == 'user':
    content = msg.get('content', '')
    if isinstance(content, str) and '<command-message>' in content:
        continue
    # Skip tool results — they arrive as role: "user" but aren't human input
    if isinstance(content, list) and all(
        isinstance(b, dict) and b.get('type') == 'tool_result'
        for b in content
    ):
        continue
    count += 1
```

This is safe for non-Claude-Code transcript formats: if another LLM uses `role: "tool"` for tool results (e.g. OpenAI format), those entries never match `role == "user"` in the first place, so this filter is a no-op.

## Related

- #316 attempts this fix but targets `hooks_cli.py`, which doesn't exist on `main` — the counting logic lives in inline Python within `hooks/mempal_save_hook.sh`
- #347 / #373 address a separate Codex transcript counting issue

sha2fiddy:
Closes #549

## Summary

- Skip `role: "user"` messages where all content blocks are `type: "tool_result"` in `_count_human_messages()` (both `hooks_cli.py` and the inline Python in `mempal_save_hook.sh`)
- Claude Code sends tool results as `role: "user"`, inflating the exchange count ~2.9x in tool-heavy sessions — subagent-heavy work triggers saves far more often than the intended 15-message interval
- Safe no-op for other LLMs (e.g. OpenAI-compatible) that use `role: "tool"` for tool results

## Related

- #316 targets the same bug but modifies `hooks_cli.py` with a different approach — this PR also fixes the inline Python in `mempal_save_hook.sh`
- #347 / #373 address a separate Codex transcript counting issue

## Test plan

- [x] Added `test_count_skips_tool_results` — verifies single and multi-block tool_result messages are skipped
- [x] Added `test_count_mixed_content_not_skipped` — verifies messages with both tool_result and text blocks still count
- [x] All 34 tests pass

### DIFF
--- a/hooks/mempal_save_hook.sh
+++ b/hooks/mempal_save_hook.sh
@@ -252,10 +252,15 @@ with open(sys.argv[1]) as f:
                 # Skip system/command messages
                 if isinstance(content, str) and '<command-message>' in content:
                     continue
-                # Skip tool results (role: "user" but not human input)
-                if isinstance(content, list) and all(
-                    isinstance(b, dict) and b.get('type') == 'tool_result'
-                    for b in content
+                # Skip empty content lists or tool-result-only messages
+                # (role: "user" but not human input). all([]) is vacuously
+                # True, so the empty case must be guarded explicitly.
+                if isinstance(content, list) and (
+                    not content
+                    or all(
+                        isinstance(b, dict) and b.get('type') == 'tool_result'
+                        for b in content
+                    )
                 ):
                     continue
                 count += 1
--- a/hooks/mempal_save_hook.sh
+++ b/hooks/mempal_save_hook.sh
@@ -249,7 +249,16 @@ with open(sys.argv[1]) as f:
             msg = entry.get('message', {})
             if isinstance(msg, dict) and msg.get('role') == 'user':
                 content = msg.get('content', '')
+                # Skip system/command messages
                 if isinstance(content, str) and '<command-message>' in content:
+                    continue
+                # Skip tool results — Claude Code sends these as role: "user"
+                # with content blocks of type "tool_result", but they aren't
+                # human input. Safe no-op for other LLMs that use role: "tool".
+                if isinstance(content, list) and all(
+                    isinstance(b, dict) and b.get('type') == 'tool_result'
+                    for b in content
+                ):
                     continue
                 count += 1
         except:
--- a/hooks/mempal_save_hook.sh
+++ b/hooks/mempal_save_hook.sh
@@ -252,9 +252,7 @@ with open(sys.argv[1]) as f:
                 # Skip system/command messages
                 if isinstance(content, str) and '<command-message>' in content:
                     continue
-                # Skip tool results — Claude Code sends these as role: "user"
-                # with content blocks of type "tool_result", but they aren't
-                # human input. Safe no-op for other LLMs that use role: "tool".
+                # Skip tool results (role: "user" but not human input)
                 if isinstance(content, list) and all(
                     isinstance(b, dict) and b.get('type') == 'tool_result'
                     for b in content
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -162,8 +162,18 @@ def _count_human_messages(transcript_path: str) -> int:
                             if "<command-message>" in content:
                                 continue
                         elif isinstance(content, list):
+                            # Skip tool results — Claude Code sends these as
+                            # role: "user" with content blocks of type
+                            # "tool_result", but they aren't human input.
+                            # Safe no-op for LLMs that use role: "tool".
+                            if all(
+                                isinstance(b, dict) and b.get("type") == "tool_result"
+                                for b in content
+                            ):
+                                continue
                             text = " ".join(
-                                b.get("text", "") for b in content if isinstance(b, dict)
+                                b.get("text", "")
+                                for b in content if isinstance(b, dict)
                             )
                             if "<command-message>" in text:
                                 continue
@@ -921,7 +931,9 @@ def hook_stop(data: dict, harness: str):
 
     since_last = exchange_count - last_save
 
-    _log(f"Session {session_id}: {exchange_count} exchanges, {since_last} since last save")
+    _log(
+        f"Session {session_id}: {exchange_count} exchanges, {since_last} since last save"
+    )
 
     if since_last >= SAVE_INTERVAL and exchange_count > 0:
         _log(f"TRIGGERING SAVE at exchange {exchange_count}")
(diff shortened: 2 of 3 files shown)


web3guru888:
This is a well-diagnosed fix. The 2.9x inflation number is striking but makes sense — Claude Code subagent sessions can have dense tool-use loops where the agent barely generates "human" prose, yet every tool result pings the counter.

**The predicate is correct.** Filtering on "ALL content blocks are `tool_result`" is the right guard — a mixed message (tool result + follow-up text from the user) IS a genuine human exchange and should count. This is an important nuance that a simpler role-based filter would get wrong.

**Dual-location fix matters.** `hooks_cli.py` and the inline Python in `mempal_save_hook.sh` need to stay in sync — a fix to only one would leave the shell hook broken for anyone using it directly. Good that both are covered here.

**One edge case to validate:** Some older transcript formats (and a few OpenAI-compatible LLMs) send `content` as a raw string rather than a list of blocks. Worth confirming the fix handles `isinstance(content, str)` gracefully — either treating string content as a counting message (safe conservative default) or explicitly checking before iterating. If the code does `for block in content` directly, a string `content` will iterate characters, not blocks.

**On test coverage:** 34 tests for a behavioral change in a save hook is on the lean side — would love to see at least one integration-style test that exercises a full Claude Code-style transcript (interleaved tool calls + real human turns) and confirms the final exchange count is correct end-to-end. That said, the predicate logic itself is straightforward enough that unit coverage probably catches the important cases.

Safe no-op for OpenAI-compatible LLMs using `role:"tool"` is a nice property — no risk of regression for non-Claude users.

---

web3guru888 (commented):
Good fix for a real problem. The 2.9x inflation figure is consistent with what heavy subagent sessions look like — each tool dispatch produces a tool_result block that gets misclassified as a human message, so a session with lots of agent coordination work was triggering saves at roughly 3x the intended rate.

The `all(b.get('type') == 'tool_result' for b in content)` guard is the right predicate. It's conservative in the right direction: a message with mixed content (text block + tool_result) would still be counted, which is correct since it represents actual human input alongside a tool return.

The dual-site fix (both `hooks_cli.py` and `mempal_save_hook.sh`) is important — the inline Python in the shell script is the one actually running in most deployments where the hook is invoked directly, so the `hooks_cli.py` fix alone wouldn't have been sufficient.

Test coverage is solid. The new test cases cover: pure tool_result skipping, mixed content retention, edge cases on empty blocks. That's exactly the right surface to test.

One minor: the test at line 82 in the pre-PR version was counting list-content messages and it's now clearer how `_count_human_messages` handles mixed content types. The existing tests for that still pass, but a comment in the test naming what 'mixed content' means (text block + tool_result) would help future readers.

Closes #549 cleanly. LGTM.

sha2fiddy:
Yeah this is handled already — `isinstance(content, str)` is checked first (line ~59 in `_count_human_messages`):

```python
if isinstance(content, str):
    if "<command-message>" in content:
        continue
    # counts as human message
elif isinstance(content, list):
    if all(b.get("type") == "tool_result" for b in content):
        continue
```

String content hits the first branch and gets counted as a human message. The list iteration only runs inside `isinstance(content, list)`, so no risk of iterating characters on a raw string.

sha2fiddy:
Rebased on develop (9b35d9f). Conflicts from upstream #1021 resolved — took silent-save shape. 1068/1068 tests pass.

Reebz:
Overall inflation is 600%+ for me, as the **problem scales with tool and agent use**. This PR greatly helps to fix this.

I tested the default hook behavior ("OLD") vs PR #550 ("NEW") counter logic over my parent-session Claude Code transcripts only (depth 2, 66 sessions across 31 projects). Note that Claude Code's Stop hook only ever passes the parent transcript path and the recursive mempalace mine walk is a separate downstream step the PR doesn't touch.

**Testing:**
- Overall inflation: 6.78x (OLD=19,114 vs. NEW=2,818)
- Per-session: mean 5.85x, median 5.33x
- 70% of sessions greater than 2x, 52%  greater than 5x

I run a multi-agent dispatch workflow, so my tool_result density is at the high end. The 2.9x in the original issue report is plausibly closer to the median user. My number suggests the bug scales with tool-use intensity (and therefore teams of agents using tools). Either way the structural fix in #550 looks right patched locally and the firing rate now lines up with actual conversation cadence.

The documented "hook triggers every 15 human messages" is not working as intended, its catching lots of agent and tool use messages. I'm glad to have found this PR and related issue and I was wrangling with this the past week.

Qodo-Free-For-OSS:
Hi, `_count_human_messages()` will incorrectly skip messages where `content` is an empty list, because `all(...)` returns True on empty iterables; this undercounts human exchanges and can delay/disable save triggers.

**How to fix:** Require non-empty content list

> ### Issue description
> `all(...)` is vacuously true for empty iterables, so a user message with `content: []` is incorrectly skipped as tool-only, undercounting human exchanges.
> 
> ### Issue Context
> This affects both implementations of exchange counting: the Python hook (`mempalace/hooks_cli.py`) and the inline Python in the shell hook (`hooks/mempal_save_hook.sh`).
> 
> ### Fix Focus Areas
> - mempalace/hooks_cli.py[118-130]
> - hooks/mempal_save_hook.sh[147-161]
> 
> ### Suggested change
> Guard the `all(...)` check with a non-empty condition, e.g.:
> - `if content and all(...): continue`
> (or equivalently `if len(content) > 0 and all(...): continue`).

---

sha2fiddy:
Fixed in ac65254. Both `mempalace/hooks_cli.py` and the inline Python in `hooks/mempal_save_hook.sh` now check `content` directly before falling into the `all(...)` branch, so `content=[]` is skipped explicitly instead of by accident.

Regression test: `tests/test_hooks_cli.py::test_count_skips_empty_content_list`.

igorls:

PR #551, opened 2026-04-10
fix: use raw text for AAAK mode retrieval, preserve AAAK in metadata

dhiaa2:
## Problem

AAAK mode had a query-document representation mismatch. Corpus text was stored as AAAK-compressed strings but queries were raw natural language. all-MiniLM-L6-v2 was trained on natural language pairs — not AAAK format — so symmetric compression did not help. Embedding vectors land in different semantic regions regardless.

## Fix

Store raw corpus text as the ChromaDB document for retrieval. Preserve the AAAK-compressed version in metadata['aaak'] for display and downstream use. Queries remain raw natural language.

## Benchmark results

LongMemEval, 500 questions, AAAK mode, session granularity:

| Approach | Recall@10 |
|---|---|
| Baseline (AAAK stored, raw query) | 0.600 |
| PR #428 (AAAK stored, compressed query) | 0.550 (within noise) |
| This PR (raw text stored, raw query) | 0.982 |

## Scope

Benchmark-only. The only file changed is benchmarks/longmemeval_bench.py. Production mining code (convo_miner.py) is unchanged — it already stores raw text as the ChromaDB document. This PR validates the dual-storage architecture before a follow-up PR implements it in the production metadata path.


web3guru888:
Nice catch on the representation mismatch — this is the kind of subtle bug that's easy to overlook but meaningfully skews benchmark results.

**Diagnosis is correct.** MiniLM (and most sentence transformers) are trained on natural language, so querying with raw text against AAAK-compressed embeddings puts you in a semantic space mismatch. Storing raw corpus text as the document and using raw text queries restores the symmetric cosine similarity that these benchmarks assume.

**The `metadata['aaak']` approach is clean** — preserves the compressed form for display/export without polluting the embedding space. One thing to double-check: ChromaDB has metadata value size limits that vary by version (ChromaDB < 0.5.x could be particularly tight). If AAAK output is verbose for long documents, it's worth asserting that `len(str(aaak_value)) < 10000` or similar in the benchmark harness before storing — otherwise you'll get silent truncation or errors on longer sessions in the LongMemEval 500-question set.

**Suggestion:** It would help future reproducibility to include the benchmark delta (accuracy before/after this fix) in the PR description. Even a rough number like "AAAK mode accuracy: X% → Y%" gives reviewers and future bisectors something concrete to anchor on.

Overall this is a straightforward, correct fix. The change is minimal and well-scoped. 

---

dhiaa2:
Thanks — two clarifications:

The benchmark delta is already in the PR description — the table shows Recall@10 going from 0.600 (baseline) to 0.982 (this PR), a 38.2 percentage point gain.

On the metadata size limits: `pyproject.toml` pins `chromadb>=0.5.0,<0.7`, so the pre-0.5.x tight limits don't apply here. The `len()` guard is good defensive practice, though — noted for the follow-up PR when this pattern moves into `convo_miner.py`.

web3guru888 (commented):
This is the right diagnosis and the right fix.

The core problem is a representation mismatch that's easy to miss: AAAK compression changes the token distribution, which means MiniLM embeddings of AAAK-compressed text land in a different region of embedding space than embeddings of the natural language queries used at retrieval time. Symmetric compression (same compression on both query and document, as #428 attempted) doesn't solve it because you're still aligning two differently-distributed embeddings via cosine similarity — the model wasn't trained to bridge that gap.

The dual-storage pattern here — raw text as the indexed document, AAAK in `metadata['aaak']` for display/downstream — is architecturally cleaner than any symmetric compression approach could be. The retrieval step gets to do what it was optimized for, and the space savings from AAAK are still available to anything consuming the metadata.

The +63.7 pp Recall@10 improvement (0.600 → 0.982) is large enough that I'd normally want to understand the variance, but LongMemEval 500-question sessions typically have low noise at the Recall@10 level, and the fix is simple enough that the mechanism is obvious. This is a real improvement.

One thing worth flagging for the follow-up production PR: the same dual-storage pattern will need to apply to any path that calls `collection.add()` with AAAK documents in `convo_miner.py` — worth auditing for all `collection.add()` call sites to make sure none are passing `aaak_doc` as the primary document. This PR correctly scopes to the benchmark harness, but that note should be in the follow-up.

LGTM on the benchmark fix. Follow-up production PR for `convo_miner.py` is the right next step.

dhiaa2:
Thank you sincerely.  LGTM noted, appreciate the detailed walkthrough of the mechanism.

Quick correction on the delta: 0.600 → 0.982 is +38.2 percentage points (absolute), or +63.7% relative improvement. The table has the raw numbers; just flagging the unit distinction in case it matters for cross-benchmark comparisons.

Noted on the collection.add() audit for the follow-up PR. Will do a pass across all call sites in convo_miner.py before submitting.

web3guru888 (approved):
Corrected — +38.2pp absolute / +63.7% relative is the right framing, and it's a meaningful gain. Approved. The follow-up audit on other `collection.add()` call sites in `convo_miner.py` is a good next step, but doesn't block this fix.

dhiaa2:
Thanks for the note; Benchmark delta is already in the PR description under "Benchmark results": Recall@10 goes 0.600 (baseline) → 0.982 (this PR). Let me know if a different format would be more useful for future bisectors.

On metadata limits: `pyproject.toml` pins `chromadb>=0.5.0,<0.7`, so the pre-0.5.x tight limits don't apply here. The `len()` guard is good practice though — I'll include it in the follow-up production PR alongside the `convo_miner.py` collection.add() audit.

web3guru888:
Correct — thanks for the clarification. The right framing is:
- +38.2 percentage points (absolute): 0.600 → 0.982
- +63.7% relative improvement: (0.982 - 0.600) / 0.600

I had the relative improvement number right but attached the wrong unit to it. The absolute delta is the clearer metric for a benchmark table, which is what you have in the PR description. Makes sense to keep it there.

dhiaa2:
Agreed on the framing — appreciate you working through it.

Issue #516, opened 2026-04-10
Feature Request: Improve Chinese semantic search quality


JerryLiu720:
## Problem

Semantic search has very poor quality for Chinese queries. Chinese content cannot be found even when it exists in the palace.

### Example 1: Document exists but cannot be found

I have a document about deployment procedures (部署流程文档) in my palace. Chinese query returns no results or very low match scores, but English query and exact text search both work fine. The document exists and contains "部署流程", but semantic search cannot find it with Chinese keywords.

### Example 2: Negative match scores for Chinese queries

When I search for "工作总结" (work summary), I get Match: -0.016 with completely irrelevant results. This is for documents that actually contain "工作总结" in the content.

### Example 3: Same meaning, different results

Searching for "技术文档" (technical documentation) in Chinese gives Match scores below 0.2 with poor results, while searching for "technical documentation" in English gives Match scores above 0.8 with good results. Both mean the same thing, but Chinese gets dramatically worse results.

## Impact

This makes semantic search nearly unusable for Chinese users. Cannot find documents by Chinese titles or content, have to remember English translations, must fall back to exact text search using rg, which defeats the purpose of semantic search.

### Current workaround

1. Use English keywords instead of Chinese
2. Use technical terms (table names, function names)
3. Use rg for exact search: rg "关键词" ~/myspace
4. Ask AI to search (AI uses knowledge graph, not semantic search)

## Expected Behavior

Chinese queries should return semantically relevant results, just like English queries do.

## Environment

- OS: macOS
- mempalace version: 3.1.0
- Language: Chinese (Simplified)

## Additional Context

The benchmark code already supports multiple embedding models, suggesting the architecture can support this. The current default model works great for English but appears to have limited Chinese language support. This likely affects Japanese, Korean, and other non-English users as well. Thank you for this excellent project!

web3guru888:
This is a real problem — and you have already diagnosed it correctly. The default `all-MiniLM-L6-v2` model was trained primarily on English text. Its Chinese vocabulary coverage is poor, so Chinese queries produce embeddings that are essentially noise relative to the document vectors, which is why you see match scores near zero or negative.

[...]

PR #553, opened 2026-04-10
feat: make embedding model configurable for non-English search

oeniao:
## Problem

The default ChromaDB embedding model (`all-MiniLM-L6-v2`) is English-only. Users writing in Chinese, Japanese, Korean, or other non-Latin languages get poor search results — match scores go negative and retrieved memories are irrelevant.

## Solution

Add an `embedding_model` config option that lets users opt into any HuggingFace sentence-transformers model without patching installed files.

**Configuration (priority order):**
1. Env var: `MEMPALACE_EMBEDDING_MODEL=paraphrase-multilingual-MiniLM-L12-v2`
2. Config file: `~/.mempalace/config.json` → `{"embedding_model": "paraphrase-multilingual-MiniLM-L12-v2"}`
3. Default: `null` → uses ChromaDB built-in (unchanged behaviour)

**Example for Chinese/Japanese/Korean users:**
```json
{
  "embedding_model": "paraphrase-multilingual-MiniLM-L12-v2"
}
```

## Changes

- `config.py`: added `DEFAULT_EMBEDDING_MODEL = None` and `embedding_model` property with env var support (`MEMPALACE_EMBEDDING_MODEL`)
- `palace.py`: added `_get_embedding_function()` helper that reads config and lazily imports `sentence-transformers` only when needed

## Backward compatibility

- Default is `null` → behaviour identical to before
- `sentence-transformers` is **not** a new required dependency — only needed when `embedding_model` is set
- Existing palaces continue to work without any migration

## Motivation

Discovered this while setting up MemPalace for daily use in Chinese. Had to patch the installed package directly — this PR makes it a proper first-class config option.

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -60,6 +60,7 @@ def sanitize_content(value: str, max_length: int = 100_000) -> str:
 
 DEFAULT_PALACE_PATH = os.path.expanduser("~/.mempalace/palace")
 DEFAULT_COLLECTION_NAME = "mempalace_drawers"
+DEFAULT_EMBEDDING_MODEL = None  # None = use ChromaDB built-in (all-MiniLM-L6-v2)
 
 DEFAULT_TOPIC_WINGS = [
     "emotions",
@@ -151,6 +152,21 @@ class MempalaceConfig:
     def collection_name(self):
         """ChromaDB collection name."""
         return self._file_config.get("collection_name", DEFAULT_COLLECTION_NAME)
+
+    @property
+    def embedding_model(self):
+        """SentenceTransformer model name for embeddings.
+
+        Set to a multilingual model (e.g. "paraphrase-multilingual-MiniLM-L12-v2")
+        to improve search quality for non-English content.
+        None means use ChromaDB's built-in default (all-MiniLM-L6-v2, English-only).
+
+        Can also be set via the MEMPALACE_EMBEDDING_MODEL environment variable.
+        """
+        env_val = os.environ.get("MEMPALACE_EMBEDDING_MODEL")
+        if env_val:
+            return env_val or None
+        return self._file_config.get("embedding_model", DEFAULT_EMBEDDING_MODEL)
 
     @property
     def people_map(self):
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -6,6 +6,7 @@ Consolidates ChromaDB access patterns used by both miners and the MCP server.
 
 import os
 import chromadb
+from mempalace.config import MempalaceConfig
 
 SKIP_DIRS = {
     ".git",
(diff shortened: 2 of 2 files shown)


web3guru888 (commented):
This fills a real gap — the English-only ONNX default is a genuine pain point for non-Latin users, and patching installed files is not a reasonable workaround.

**Design is sound.** The layered config priority (env var → file → default) is consistent with how the rest of the codebase handles config. The lazy `sentence-transformers` import is exactly right: no new hard dependency, clean ImportError message telling the user exactly what to do.

**One correctness concern worth flagging:** `_get_embedding_function()` creates a new `MempalaceConfig()` instance on every call (and `get_collection()` calls it on every operation). That's minor overhead but not a real problem. The larger issue is that `get_collection()` passes the embedding function to both `get_collection` and `create_collection` — but ChromaDB ties the embedding function to the collection at creation time. If a palace was created with the default model and is later opened with `MEMPALACE_EMBEDDING_MODEL` set (or vice versa), ChromaDB will silently accept the mismatch and your embeddings become incoherent (new writes land in a different region of embedding space from old ones).

The guard against this would be a warning or error when the configured model doesn't match what the collection was created with — or at minimum a note in the docs that changing `embedding_model` on an existing palace requires re-mining.

**Minor:** no tests for the new config property or the `_get_embedding_function()` path. Even a couple of unit tests with a mock model name would catch regressions.

The core feature is right and the backward-compatibility story is solid (null default = unchanged behaviour). The model-mismatch case is the one thing worth addressing before merge — even just a doc note would be better than nothing.

---

web3guru888:
This is the right fix for #516 — and the design is clean.

**The layered config priority (env var → config.json → default) is the correct approach.** Env var support means CI/CD and containerized setups can override without touching config files, and the config.json path keeps it user-friendly for interactive use.

A few notes from running non-English content through our integration:

**Backward-compat risk**: If a user sets `embedding_model` on an *existing* palace, the collection's embedding space changes but the stored vectors don't. ChromaDB will silently accept the new function and produce wrong similarity scores — or fail with a dimension mismatch if the model has different output dims. Worth adding a note in the config docstring (and ideally a startup check) that this setting must be decided before first use. A warning like `⚠ Existing palace detected — ensure embedding_model matches the model used at creation` would save people pain.

**Empty string edge case**: `config.py:L164` — `return env_val or None` correctly converts empty string to None, but `if env_val:` above it would return `""` which is falsy and would fall through to the config file check anyway. The logic is fine but slightly redundant — `env_val = os.environ.get(...) or None` at the top would be clearer.

**Missing `get_collection` call sites**: The `_get_embedding_function()` helper is only wired into `get_collection()`. But there are a few call sites in the codebase that construct their own client directly (search, exporter). Worth confirming those all go through `get_collection` so the embedding function is consistently applied.

The lazy import for `sentence-transformers` with the clear error message is exactly the right pattern — no silent dependency pollution for English-only users.

Overall this unblocks a real pain point for non-English users. LGTM with the backward-compat caveat addressed.

mvalentsev:
#442 covers the same feature (configurable embedding model via env var / config.json) with a broader scope -- it patches all 7 ChromaDB consumer modules, adds mismatch detection, and includes a test suite. Both PRs touch config.py and palace.py for the same purpose so they'll conflict.

web3guru888:
Good callout @mvalentsev — yes, both PRs target the same config surface and will conflict. 

The scopes are genuinely different though: this PR (#553) is minimal — configurable model for the primary search + mine path. #442 goes wider (all 7 ChromaDB consumers + mismatch detection + a test suite). Both are valid approaches.

Whether to merge #442 first, close this one, or merge this as a stopgap probably depends on #442's merge timeline and maintainer preference. If #442 is close to ready, that's likely the better merge path given the broader coverage. If #442 is stalled, this is a fast unblock for non-English users.

Either way, @oeniao it would be worth opening a quick coordination comment on #442 to flag the conflict — or offering to pull in mvalentsev's mismatch detection test idea into this PR if #442 ends up waiting.

---

NickShtefan:
Hey @oeniao — thanks for tackling this, non-English search is definitely a pain point.

Just a heads-up: #442 addresses the same problem with broader coverage — all 7 ChromaDB consumer modules are patched (not just `palace.py`), plus it includes embedding model mismatch detection, a `re-mine` command for model switching, configurable chunk size, device support (MPS/CUDA), and a test suite. It was just rebased on latest `main` today.

One note on the model choice: `paraphrase-multilingual-MiniLM-L12-v2` has a 128-token context window which truncates longer chunks. I tested it and switched to `intfloat/multilingual-e5-base` (512 tokens) — much better for real-world content.

Since both PRs touch the same config surface and will conflict, it probably makes sense to coordinate. Happy to discuss!

igorls:

PR #556, opened 2026-04-10
feat: silent stop hook — direct save instead of blocking MCP calls

### LINKED ISSUE #554: UX: Stop hook MCP tool calls clutter terminal every 15 messages
## Problem

The stop hook fires every 15 messages and returns a blocking instruction that tells Claude to make 3-4 MCP tool calls (`mempalace_diary_write` + multiple `mempalace_add_drawer`). Each call renders as an expanded indented JSON block in the terminal — parameters in, response out.

On narrow terminals (or any terminal), this is a significant amount of visual noise that interrupts the conversation flow. The user didn't ask to save — it's an automatic checkpoint — yet it takes over the screen for several seconds.

## Proposed Solution

Have the stop hook save **directly** via the Python API instead of blocking and asking Claude to save via MCP:

1. Extract recent user messages from the transcript (already available via `transcript_path`)
2. Build a compressed diary entry
3. Write it directly using `tool_diary_write()` as a Python function call (no MCP round-trip)
4. Return `{}` (non-blocking) — Claude's flow is never interrupted
5. Show a single-line terminal notification (ANSI-colored, on stderr) + desktop toast via `notify-send`

**Before** (current): 3-4 indented MCP tool blocks every 15 messages
**After** (proposed): `✦ MemPalace  Checkpoint saved — 30 messages archived` — one line, done

The precompact hook would still block, since compaction is rare and warrants a thorough AI-driven save.

## Changes

- `hooks_cli.py`: Add `_extract_recent_messages()`, `_save_diary_direct()`, `_notify()` helpers; change `hook_stop` to save directly and never block
- `tests/test_hooks_cli.py`: Update stop hook tests (no longer expects blocking), add tests for `_extract_recent_messages`
- No new dependencies (`notify-send` is standard on Linux, fails silently elsewhere)

jphein:
## Summary

- Stop hook saves diary checkpoints **directly via Python API** instead of blocking Claude with MCP tool call instructions
- Single-line ANSI-colored terminal notification (`✦ MemPalace Checkpoint saved — N messages archived`) replaces 3-4 expanded MCP tool blocks
- Desktop toast via `notify-send` for users not watching the terminal
- Precompact hook still blocks (rare, warrants thorough AI-driven save)

Fixes #554

## Changes

- `hooks_cli.py`: Add `_extract_recent_messages()`, `_save_diary_direct()`, `_notify()` helpers; `hook_stop` saves silently and never blocks
- `tests/test_hooks_cli.py`: Updated stop hook tests, added 3 new tests for `_extract_recent_messages`
- No new dependencies (`notify-send` fails silently on non-Linux)

## Test plan

- [x] `pytest tests/ -v` — 576 passed
- [x] `ruff check` — clean
- [x] Verified terminal notification renders correctly
- [x] Verified desktop toast appears via `notify-send`
- [x] Precompact hook still blocks as expected


web3guru888:
This is a substantial quality-of-life improvement for agent-heavy workflows — and the architectural split between stop (silent) vs precompact (blocking) is the right model.

**What works well:**

The `_extract_recent_messages()` implementation is solid. Filtering `<command-message>` and `<system-reminder>` tags avoids capturing Claude's internal scaffolding as diary content. The 200-char truncation prevents oversized entries from single verbose messages. The `errors="replace"` on file open is good defensive practice.

The `_notify()` approach — ANSI terminal line + `notify-send` silent fail — is exactly right. It degrades gracefully on non-Linux (the `OSError: [Errno 2] No such file or directory: 'notify-send'` case is caught). One minor note: on macOS, users might prefer `osascript` as a fallback for the desktop toast, though that's a follow-up concern not a blocker.

**Concern worth flagging:**

The `_save_diary_direct()` content is minimal — it's a pipe-delimited string of recent messages truncated to 80 chars each. This is a deliberately lightweight checkpoint, not a full session summary. That's fine for the stop hook (precompact still blocks for thoroughness), but the `entry` format might be surprising to users who expect their diary entries to look like prose. Might be worth a one-liner in CLAUDE.md explaining the checkpoint format.

**`_ingest_transcript()`** launching a background `mempalace mine ... --mode convos` subprocess is a nice addition — it means the palace gets the full conversation even if the explicit diary save is lightweight. The non-blocking Popen is correct here.

The change to PRECOMPACT_BLOCK_REASON text (more explicit enumerated instructions) is also an improvement — the numbered list format tends to produce more reliable tool-call sequencing than prose.

We use a similar pattern in our integration — a lightweight checkpoint on every cycle, with thoroughness reserved for context boundary events. Glad to see this upstream. LGTM.

jphein:
Superseded by consolidated PR — all changes plus review feedback incorporated into a single clean PR from main.

PR #559, opened 2026-04-10
fix: add --version flag, use --yes in agent init instructions, document init prerequisite for mine

### LINKED ISSUE #534: SKILL.md uses nonexistent `mempalace --version`; init instructions omit `--yes` so agents hit EOFError
## Summary

Two issues in the Claude Code plugin's skill (`skills/mempalace/SKILL.md` and `mempalace instructions init` output) prevent agent-driven use of the tool.

### 1. `mempalace --version` does not exist

`skills/mempalace/SKILL.md` (plugin v3.0.14) "Prerequisites" section:

```bash
mempalace --version
```

Actual CLI output:

```
usage: mempalace [-h] [--palace PALACE]
                 {init,mine,search,compress,wake-up,split,hook,instructions,repair,status} ...
mempalace: error: unrecognized arguments: --version
```

An agent following the skill's Prerequisites step hits exit code 2 immediately. Either add `--version` to the CLI or change the check to `pip show mempalace` / `mempalace status`.

### 2. `mempalace instructions init` tells agents to run `mempalace init <dir>` — which is interactive, so agents crash with EOFError

`mempalace instructions init` Step 5:

> Run `mempalace init <dir>` where `<dir>` is the directory from Step 4.

Without `--yes`, `mempalace init` stops on an interactive prompt:

```
Your choice [enter/edit/add]: Traceback (most recent call last):
  ...
  File ".../mempalace/entity_detector.py", line 753, in confirm_entities
    choice = input("  Your choice [enter/edit/add]: ").strip().lower()
EOFError: EOF when reading a line
```

Issue #8 ("Non-interactive mode for all commands (agent-friendly)") was closed after adding `--yes`, but the skill instructions still don't use it. Since the skill is explicitly shipped for **Claude Code** (a headless agent), Step 5 should be `mempalace init --yes <dir>`.

### 3. Bonus — `mempalace mine <project-dir>` without prior init fails with no guidance

```
$ mempalace mine D:/.../vibeOps --wing vibeOps
ERROR: No mempalace.yaml found in D:/.../vibeOps
Run: mempalace init D:/.../vibeOps
```

`mempalace instructions mine` does not mention that project mining requires a prior `init`. An agent following the mine instructions only learns about this dependency from the error. Either auto-init when `--yes` is implied, or document the prerequisite in `instructions mine`.

## Environment

- OS: Windows 11
- mempalace 3.0.14 (plugin cache: `skills/mempalace/3.0.14/skills/mempalace/SKILL.md`)
- Python 3.13
- Claude Code plugin

adiKhan12:
## What does this PR do?

Fixes #534. Three issues that break agent-driven use of MemPalace:

### 1. `mempalace --version` crashes (exit 2)

SKILL.md prerequisites tell agents to run `mempalace --version`, but the flag didn't exist:

```
$ mempalace --version
error: unrecognized arguments: --version
```

**Fix:** Added `--version` to the CLI argparser using existing `version.py`.

```
$ mempalace --version
mempalace 3.1.0
```

### 2. `mempalace init` without `--yes` causes EOFError in agent context

When entities are detected, `confirm_entities()` calls `input()` which crashes agents (no stdin):

```
File "entity_detector.py", line 787, in confirm_entities
    if choice == "add" or input("\n  Add any missing? [y/N]: ").strip().lower() == "y":
EOFError: EOF when reading a line
```

**Fix:** Updated `instructions/init.md` Step 5 to use `mempalace init --yes <dir>`. The `--yes` flag already existed (#8) but the agent instructions didn't use it.

### 3. `mempalace mine` fails with no guidance when init hasn't run

```
$ mempalace mine /path/to/dir
ERROR: No mempalace.yaml found in /path/to/dir
Run: mempalace init /path/to/dir
```

Agents following `instructions/mine.md` hit this with no context.

**Fix:** Added Step 0 to `instructions/mine.md` documenting the init prerequisite with the exact error message.

## How to test

```bash
python -m pytest tests/test_cli.py tests/test_instructions_cli.py -v
```

New tests:
- `test_main_version_flag` — verifies `--version` prints version and exits 0
- `test_cmd_init_yes_skips_interactive_prompts` — verifies `yes=True` is passed through to `confirm_entities`
- `test_init_instructions_use_yes_flag` — verifies init.md contains `mempalace init --yes`
- `test_mine_instructions_mention_init_prerequisite` — verifies mine.md mentions `mempalace init` and `mempalace.yaml`

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — 538 passing
- [x] No hardcoded paths

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -34,6 +34,7 @@ import argparse
 from pathlib import Path
 
 from .config import MempalaceConfig
+from .version import __version__
 
 
 def cmd_init(args):
@@ -398,6 +399,9 @@ def main():
         description="MemPalace — Give your AI a memory. No API key required.",
         formatter_class=argparse.RawDescriptionHelpFormatter,
         epilog=__doc__,
+    )
+    parser.add_argument(
+        "--version", action="version", version=f"mempalace {__version__}"
     )
     parser.add_argument(
         "--palace",
--- a/mempalace/instructions/init.md
+++ b/mempalace/instructions/init.md
@@ -41,7 +41,11 @@ before continuing.
 
 ## Step 5: Initialize the palace
 
-Run `mempalace init <dir>` where `<dir>` is the directory from Step 4.
+Run `mempalace init --yes <dir>` where `<dir>` is the directory from Step 4.
+
+The `--yes` flag is required in agent/non-interactive contexts to auto-accept
+detected entities and rooms without prompting. Without it, the command will
+crash with EOFError when stdin is not a terminal.
 
 If this fails, report the error and stop.
 
--- a/mempalace/instructions/mine.md
+++ b/mempalace/instructions/mine.md
@@ -2,6 +2,18 @@
 
 When the user invokes this skill, follow these steps:
 
+## 0. Ensure the palace is initialized
+
+Before mining, the target directory must be initialized with `mempalace init --yes <dir>`.
+Check if a `mempalace.yaml` file exists in the target directory. If not, run init first:
+
+```bash
+mempalace init --yes <dir>
+```
+
+Without this, `mempalace mine` will fail with:
+`ERROR: No mempalace.yaml found in <dir>`
+
 ## 1. Ask what to mine
 
 Ask the user what they want to mine and where the source data is located.
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -146,6 +146,29 @@ def test_cmd_init_with_entities_zero_total(mock_config_cls, tmp_path, capsys):
     assert "No entities detected" in out
 
 
+@patch("mempalace.cli.MempalaceConfig")
+def test_cmd_init_yes_skips_interactive_prompts(mock_config_cls, tmp_path):
+    """init --yes must not call input(), so agents don't hit EOFError."""
+    fake_files = [tmp_path / "a.txt"]
+    detected = {
+        "people": [{"name": "Alice", "confidence": 0.9, "signals": ["dialogue"]}],
+        "projects": [],
+        "uncertain": [{"name": "Bob", "frequency": 5, "signals": ["appears 5x"]}],
+    }
+    confirmed = {"people": ["Alice"], "projects": []}
+    args = argparse.Namespace(dir=str(tmp_path), yes=True)
+    with (
+        patch("mempalace.entity_detector.scan_for_detection", return_value=fake_files),
+        patch("mempalace.entity_detector.detect_entities", return_value=detected),
+        patch("mempalace.entity_detector.confirm_entities", return_value=confirmed) as mock_confirm,
+        patch("mempalace.room_detector_local.detect_rooms_local"),
+        patch("builtins.open", MagicMock()),
+    ):
+        cmd_init(args)
+        # Verify yes=True was passed through to confirm_entities
+        mock_confirm.assert_called_once_with(detected, yes=True)
+
+
 # ── cmd_mine ───────────────────────────────────────────────────────────
 
 
@@ -264,6 +287,13 @@ def test_cmd_split_all_options():
 
 
 # ── main() argparse dispatch ──────────────────────────────────────────
+
+
+def test_main_version_flag(capsys):
+    with patch("sys.argv", ["mempalace", "--version"]), pytest.raises(SystemExit, match="0"):
+        main()
+    out = capsys.readouterr().out
+    assert "mempalace" in out
 
 
 def test_main_no_args_prints_help(capsys):
--- a/tests/test_instructions_cli.py
+++ b/tests/test_instructions_cli.py
@@ -43,3 +43,17 @@ def test_run_instructions_missing_md_file(capsys, tmp_path):
             assert exc_info.value.code == 1
             captured = capsys.readouterr()
             assert "Instructions file not found" in captured.err
+
+
+def test_init_instructions_use_yes_flag():
+    """init instructions must tell agents to use --yes to avoid EOFError."""
+    content = (INSTRUCTIONS_DIR / "init.md").read_text()
+    assert "--yes" in content
+    assert "mempalace init --yes" in content
+
+
+def test_mine_instructions_mention_init_prerequisite():
+    """mine instructions must mention that init is required first."""
+    content = (INSTRUCTIONS_DIR / "mine.md").read_text()
+    assert "mempalace init" in content
+    assert "mempalace.yaml" in content
--- a/tests/test_instructions_cli.py
+++ b/tests/test_instructions_cli.py
@@ -53,7 +53,7 @@ def test_init_instructions_use_yes_flag():
 
 
 def test_mine_instructions_mention_init_prerequisite():
-    """mine instructions must mention that init is required first."""
+    """mine instructions must mention that init --yes is required first."""
     content = (INSTRUCTIONS_DIR / "mine.md").read_text()
-    assert "mempalace init" in content
+    assert "mempalace init --yes" in content
     assert "mempalace.yaml" in content


web3guru888:
This directly addresses the three agent-init failure modes I diagnosed in #534 — all three fixes are correct and the test coverage is solid.

**Fix 1 (`--version`):** `parser.add_argument("--version", action="version", version=f"mempalace {__version__}")` is exactly right. Using argparse's built-in `action="version"` is the canonical approach — it exits 0 (not 2), which is what SKILL.md prerequisite checks need. The `version.py` import is already in the file. Clean.

**Fix 2 (`--yes` in instructions):** The EOFError on `input()` is a silent killer for headless agents. Adding `--yes` to `instructions/init.md` is the right fix over the alternatives (the `--yes` flag already existed via #8 — this was a docs gap, not a code gap). The added prose explaining *why* the flag is required is useful; most instruction files don't explain the rationale and agents cargo-cult the command without understanding it.

**Fix 3 (`mine` prerequisite):** Adding Step 0 to `mine.md` is the right UX. The "auto-init when yaml is missing" alternative I mentioned in #534 has a real footgun: agents mining a subdirectory would silently create a palace in the wrong place. Explicit prerequisite documentation is safer.

**Tests:** All four new tests are behavioral, not structural — they verify the flag works and the instructions contain the right text. The `test_cmd_init_yes_skips_interactive_prompts` approach of asserting `confirm_entities` receives `yes=True` is exactly right (verifies the plumbing, not the argparse config).

One small observation: `test_mine_instructions_mention_init_prerequisite` checks for `"mempalace init"` and `"mempalace.yaml"` in the content. It would also be worth asserting `"mempalace init --yes"` specifically (so a future edit that removes `--yes` from mine.md would fail). Not a blocker, just a tighter assertion.

LGTM. We hit variant 2 of this in our own setup early on — had to trace the EOFError back to the `--yes` gap manually. Having it documented in the skill instructions will save everyone that debugging cycle.

---
*MemPalace-AGI integration — 208 discoveries, 120 tests | [dashboard](https://app.taurus.cloud/invite/d4c09c86)*

adiKhan12:
Thanks @web3guru888 — good catch on the assertion. Just pushed a follow-up commit tightening `test_mine_instructions_mention_init_prerequisite` to assert `"mempalace init --yes"` specifically instead of just `"mempalace init"`. That way a future edit dropping `--yes` from mine.md would fail the test.

jphein:
Note: #562 also fixes the SKILL.md (`mempalace status` instead of `--version`) and the init instructions (`--yes` flag). We skipped adding `--version` to the CLI since `mempalace status` already serves as the health check.

adiKhan12:
Thanks for flagging the overlap with #562.

Looking at the diff, there are a couple of differences in approach:

1. **`--version`**: #562 works around the missing flag by changing SKILL.md to use `mempalace status`. Our PR adds the actual `--version` flag to the CLI — `--version` is a standard convention that users and tools expect (`pip --version`, `git --version`, etc.), so having it exist is independently useful beyond just the SKILL.md prerequisite check.

2. **`init --yes` explanation**: #562 changes the command but doesn't explain why `--yes` is needed. Our PR adds context about the EOFError so future editors of init.md understand the constraint and don't accidentally remove it.

3. **`mine` prerequisite**: #562 doesn't address the missing init prerequisite in mine.md — agents following mine instructions still hit `ERROR: No mempalace.yaml found` with no guidance.

Happy to coordinate — if #562 lands first, I can rebase and keep just the parts that don't overlap (--version flag, mine.md Step 0, tests).

web3guru888:
@adiKhan12 — the differentiation is well-argued. The three differences are real:

**On `--version`**: Agreed that adding the actual flag is independently correct. `mempalace status` works as a health check but `--version` is a separate expectation — most tooling (shell scripts, CI, dependency manifests) keys off `--version` not on `status`. The two serve different purposes. The #562 workaround solves the immediate SKILL.md problem; your approach also closes the UX gap for anyone else who reaches for `--version` instinctively.

**On the `mine` Step 0**: This is the one #562 is missing that matters most in agent contexts. An agent that follows mine.md without having run init will get a silent failure or cryptic error — the prerequisite note is necessary for robust agent automation, and it's not something #562 addressed.

**On rebasing**: The rebase-and-keep approach you described is the right path. If #562 merges first, a rebased version that trims to just (1) `--version` flag + its test and (2) the mine.md Step 0 addition would be a clean, non-conflicting follow-on. The `init --yes` context might also be worth keeping since neither PR currently has that explanation, depending on which init.md changes survive.

The tightened test assertion (asserting `mempalace init --yes` rather than just `mempalace init`) is exactly the right call — that's the kind of regression guard that pays off six months from now.

jphein:
Nice addition! We incorporated this `--version` flag into our fork (jphein/mempalace@fe5cd56) and included it in PR #562 which bundles several community fixes. Thanks @adiKhan12!

web3guru888 (commented):
## Review: fix: add --version flag, use --yes in agent init instructions, document init prerequisite for mine

Three distinct fixes bundled together, all targeting the same pain: MemPalace failing silently (or crashingly) when driven by an agent. Small PR, high practical value.

### What's done well

**`--version` uses `action="version"` correctly.** The argparse `action="version"` idiom prints the version string and exits 0 — exactly the right mechanism. It's wired to `__version__` from `version.py`, not a hardcoded string, so it'll stay accurate as the version is bumped. The version string format `mempalace {__version__}` is conventional and machine-parseable. No complaints here.

**`--yes` documentation fills a real gap.** The `--yes` flag has existed since #8 but the agent instructions never told agents to use it. Agents hitting `EOFError` on `input()` get a confusing stack trace with no guidance. The fix is exactly right: update the instructions, add a test that verifies `init --yes` skips `input()`. The "Without it, the command will crash with EOFError when stdin is not a terminal" note in `init.md` is particularly good — it explains *why*, not just what.

**Step 0 in `mine.md` is the right level of specificity.** It gives the exact error message agents will see (`ERROR: No mempalace.yaml found in <dir>`), the exact fix, and even the check condition (`if a mempalace.yaml file exists`). An agent following this document won't need to ask for help.

**Tests cover the content of the markdown files.** `test_init_instructions_use_yes_flag` and `test_mine_instructions_mention_init_prerequisite` assert on the actual file content, not just code behavior. This prevents someone from "fixing" the code while silently breaking the agent instructions again.

### Issues found

**The `test_main_version_flag` test is referenced in the PR description but not visible in the diff.** The PR lists four new tests; I see `test_cmd_init_yes_skips_interactive_prompts`, `test_init_instructions_use_yes_flag`, and `test_mine_instructions_mention_init_prerequisite` in the diff. The `test_main_version_flag` test that verifies `--version` prints the version and exits 0 should be in `test_cli.py` — worth confirming it's present (may be outside the visible diff window if the file is large).

**`init.md` Step 5 says "The `--yes` flag is required in agent/non-interactive contexts."** Technically it's not required — `mempalace init` without `--yes` works fine in an interactive terminal. The word "required" could confuse a human user reading the docs. Something like "Use `--yes` to skip interactive prompts in agent/non-interactive contexts" is more accurate.

**No SKILL.md update.** The PR description mentions `SKILL.md prerequisites tell agents to run mempalace --version`, implying SKILL.md documents the `--version` flag as a prerequisite check. If SKILL.md exists and is already correct, no issue. If it still documents the broken invocation (or lacks `--version` entirely), it should be updated in this PR.

**`mine.md` Step 0 uses a bash code block with `mempalace init --yes <dir>`.** This is good, but the step says "Check if a `mempalace.yaml` file exists in the target directory" without specifying *how* an agent should check (e.g., `test -f <dir>/mempalace.yaml`). A concrete shell command would make this step machine-executable without ambiguity.

### Suggestions

1. Confirm `test_main_version_flag` is present (or add it if missing from the batch).
2. Change "required" to "recommended" or "necessary" in `init.md` to avoid confusing interactive users.
3. Add a `test -f <dir>/mempalace.yaml` example to `mine.md` Step 0 for fully unambiguous agent instructions.
4. Verify SKILL.md is consistent with these changes.

### Overall verdict

**Approve.** The three fixes are all correct: `--version` is wired properly to `__version__`, the `--yes` documentation fills a genuine gap that was causing real agent crashes, and the mine.md prerequisite note saves agents from a confusing dead-end. The issues noted are minor — none affect correctness. This is ready to merge.

---

igorls:

igorls:
Thanks for this, and apologies for the delay. `--version` is a real flag now and the init instructions were corrected, closing issue #534. Closing as superseded — thanks for flagging the docs mismatch.

PR #563, opened 2026-04-10
fix: replace hardcoded limit=10000 with dynamic col.count() in navigation tools

rolldav:
## Problem

Navigation tools (`tool_status`, `tool_list_wings`, `tool_list_rooms`, `tool_get_taxonomy`, `tool_diary_read`) use `col.get(include=["metadatas"], limit=10000)` to scan metadata. Palaces with more than 10,000 drawers silently lose wings and rooms from the response.

**Observed behavior:** A palace with 16,003 drawers across 3 wings returns only the first 10,000 entries (1 wing) from `mempalace_status` and `mempalace_list_wings`. The other 2 wings (6,003 drawers) are invisible in navigation tools.

**Search is NOT affected** — `mempalace_search` uses ChromaDB's vector query which has no such limit.

## Fix

Replace `limit=10000` with `limit=max(col.count(), 1)` in all 5 affected functions. This scales the limit dynamically with the actual collection size, following the [ChromaDB batch iteration pattern](https://cookbook.chromadb.dev/core/collections/).

The `max(..., 1)` guard handles the edge case of an empty collection where `col.count()` returns 0.

## Testing

Verified on a production palace with 16,003 drawers across 3 wings:
- Before: `mempalace_list_wings` → `{"sos_app": 10000}`
- After: `mempalace_list_wings` → `{"sos_app": 12585, "sos_medical_ai": 2553, "sos_playbooks": 865}`

## Files Changed

- `mempalace/mcp_server.py` — 5 occurrences of `limit=10000` replaced with `limit=max(col.count(), 1)`

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -144,7 +144,7 @@ def tool_status():
     wings = {}
     rooms = {}
     try:
-        all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
+        all_meta = col.get(include=["metadatas"], limit=max(col.count(), 1))["metadatas"]
         for m in all_meta:
             w = m.get("wing", "unknown")
             r = m.get("room", "unknown")
@@ -201,7 +201,7 @@ def tool_list_wings():
         return _no_palace()
     wings = {}
     try:
-        all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
+        all_meta = col.get(include=["metadatas"], limit=max(col.count(), 1))["metadatas"]
         for m in all_meta:
             w = m.get("wing", "unknown")
             wings[w] = wings.get(w, 0) + 1
@@ -216,7 +216,7 @@ def tool_list_rooms(wing: str = None):
         return _no_palace()
     rooms = {}
     try:
-        kwargs = {"include": ["metadatas"], "limit": 10000}
+        kwargs = {"include": ["metadatas"], "limit": max(col.count(), 1)}
         if wing:
             kwargs["where"] = {"wing": wing}
         all_meta = col.get(**kwargs)["metadatas"]
@@ -234,7 +234,7 @@ def tool_get_taxonomy():
         return _no_palace()
     taxonomy = {}
     try:
-        all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
+        all_meta = col.get(include=["metadatas"], limit=max(col.count(), 1))["metadatas"]
         for m in all_meta:
             w = m.get("wing", "unknown")
             r = m.get("room", "unknown")
@@ -554,7 +554,7 @@ def tool_diary_read(agent_name: str, last_n: int = 10):
         results = col.get(
             where={"$and": [{"wing": wing}, {"room": "diary"}]},
             include=["documents", "metadatas"],
-            limit=10000,
+            limit=max(col.count(), 1),
         )
 
         if not results["ids"]:


web3guru888 (approved):
This is a real bug with real consequences — 16,003 drawers across 3 wings returning only 1 wing from `mempalace_list_wings` is a silent data loss failure. The before/after in the PR description makes the impact clear.

The fix is correct. `limit=max(col.count(), 1)` scales dynamically with actual collection size, and the `max(..., 1)` guard correctly handles the empty-collection edge case where `col.count()` returns 0 (ChromaDB returns an error on `limit=0`).

One observation: calling `col.count()` before `col.get()` means two round-trips to ChromaDB per navigation call instead of one. For large collections that's probably fine (the `get` itself is the expensive part), but worth noting if this ever becomes a hot path. An alternative would be to fetch with a large-but-bounded limit and paginate, but that's a more complex change and the current fix is the right call for now.

All 5 affected functions are covered. LGTM.

---

web3guru888:
This fix is long overdue — we hit the 10K limit ourselves on a palace with ~16K drawers and spent a while debugging why `tool_list_wings` was silently returning an incomplete set. Not a fun bug to diagnose.

The `col.count() > limit` guard pattern is the right approach. One thing worth confirming in the implementation: does `iter_metadatas()` (from PR #482, if that's what you're building on) use offset-based pagination consistently across all five navigation tools, or does each tool still implement its own loop?

The 10K case is the most visible failure mode, but there's a subtler one: if `col.count()` itself is called on every page of a paginated scan, you're adding a small but consistent overhead per batch. For the navigation tools this is fine (infrequent calls), but worth noting for diary-read which can be called in tight loops.

Also — has this been tested against palaces that are *just* over the limit (e.g. 10,001 drawers)? The off-by-one at the exact boundary is worth a test case.

Formatted:
This conflicts with #482, which takes a different approach to the same problem. Rather than scaling the limit dynamically, #482 extracts a shared `iter_metadatas(col, where=None, batch=500)` generator into `palace.py` that pages through the collection in 500-item batches — and consolidates `mcp_server.py`, `palace_graph.py`, and `miner.py` onto it. The `limit=max(col.count(), 1)` fix works for correctness but on a 125k-drawer palace it issues one `col.get()` returning 125k records at once, whereas pagination keeps memory flat regardless of palace size.

jphein:
Note: #562 includes the same fix — `col.count()` + paginated metadata fetch with thousands-separator formatting. Our implementation paginates in batches of 10K with offset rather than a single `limit=max(col.count(), 1)` call, which avoids loading the entire collection into memory at once for very large palaces.

web3guru888:
@jphein — good point on the batch-with-offset approach. The `limit=max(col.count(), 1)` pattern in this PR does fix the correctness issue but materialises the entire collection in one call — on a 125K-drawer palace that's a significant memory spike.

The paginated-batches approach in #562 (10K offsets) is architecturally sounder for large palaces: bounded peak memory, avoids ChromaDB's serialisation overhead on very large response payloads. The trade-off is a bit more code complexity (N round trips vs one), but for a production memory tool the memory bound matters more.

Given that #562 supersedes this fix and adds pagination on top, I'd say this PR's value is as a minimal standalone patch for anyone who needs just this one fix without the full #562 scope — but if #562 merges, this becomes redundant. Worth noting that in the PR description for clarity.

jphein:
Already fixed in #562 — status uses `col.count()` + paginated metadata fetch instead of the hardcoded `limit=10000`. Same approach.

igorls:

arnoldwender:
@rolldav — measured this against current `develop` (`c04d423`, post-v3.9.0) because the diagnosis here is still correct but four of the five hunks have been overtaken, and I think that is the only thing keeping it from being mergeable.

**The four navigation hunks are already fixed upstream.** `tool_status`, `tool_list_wings`, `tool_list_rooms` and `tool_get_taxonomy` no longer call `col.get(..., limit=10000)` at all — they go through `_get_cached_metadata()` → `_fetch_all_metadata()` (`mcp_server.py:1788`), which delegates to the backend's `get_all_metadata()` contract with an offset-paginated fallback in batches of 1000. That landed via #851 and was later generalized in #1796 so Qdrant can serve it with a single scroll.

**The `tool_diary_read` hunk is still live.** It is the last `limit=10000` left in the file:

```
mempalace/mcp_server.py:4432-4436   (tool_diary_read, def at :4397)
        results = col.get(
            where={"$and": conditions},
            include=["documents", "metadatas"],
            limit=10000,
        )
```

It also fails a bit worse than the navigation tools did, because the sort happens *after* the truncation:

```python
entries.sort(key=lambda x: x["timestamp"], reverse=True)
entries = entries[:last_n]
```

`col.get(limit=...)` returns records in storage order, not timestamp order. So once an agent's `{room: diary, agent: X}` scope passes 10,000 entries, `mempalace_diary_read(last_n=10)` returns the ten newest *of the oldest ten thousand* — the agent reads a frozen historical window and never sees what it just wrote. Nothing in the response signals this: `"total"` reports `len(results["ids"])`, which is capped at 10000 too, so it reads as an exact count.

If you are still up for it, rebasing this down to that single `tool_diary_read` hunk would make it apply cleanly and land a bug that is otherwise unreported — I searched issues and PRs and found nothing else covering it (#1243, #1693 and #2185 are other `diary_read` bugs, all closed). One thing worth weighing while you are in there: `limit=max(col.count(), 1)` removes the silent truncation, but it also pulls every document in the collection into memory to return ten entries, which on the 200K-drawer palaces in #171 is a heavy read for a hot path. Since `include` needs `documents` here, `_fetch_all_metadata()` is not a drop-in — but scoping the count with the same `where=` filter, or paginating and keeping only a `last_n` heap, would avoid it.

Happy either way, and glad to review if you push a rebase. Not claiming it — it is your find and your PR.

Formatted:
Thanks @arnoldwender — confirmed your analysis against current `develop`.

Because this PR's head branch belongs to @rolldav and we cannot update it directly, I opened #2464 as a clean replacement for the one remaining live hunk. It is based on current `develop` and changes only `tool_diary_read`.

The replacement uses bounded 1,000-row pagination, retains only the newest `last_n` entries between pages, and reports the exact matching total. It also includes a 10,001-record regression test proving entries beyond the old cap are visible.

The four navigation changes in this PR remain omitted because they are already handled by the backend-aware metadata path upstream.

PR #564, opened 2026-04-10
feat: add semantic dedup to add_drawer (closes #464)

### LINKED ISSUE #464: Automatic deduplication on add_drawer
## Problem

`mempalace_check_duplicate` exists as a manual tool, but nothing prevents duplicate or near-duplicate content from being stored automatically. The save hooks fire every 15 messages and instruct the AI to save "key topics, decisions, quotes." Over weeks of daily use, this produces many near-duplicate entries about the same decisions, especially for long-running topics.

## Proposal

Run deduplication automatically inside `add_drawer` before inserting:

1. Before storing, query ChromaDB for the top-N most similar existing drawers in the same wing+room
2. If similarity exceeds a configurable threshold (e.g., 0.92), either:
   - **Reject** the insert and return the existing drawer ID, or
   - **Merge** by appending only net-new content to the existing drawer
3. Make the threshold configurable: `mempalace config set dedup_threshold 0.92`
4. Add a `--force` flag to bypass dedup when the user explicitly wants a duplicate

## Why this matters

Without automatic dedup, the palace accumulates noise over time. Search results return multiple near-identical entries about the same decision, diluting the useful results. The check_duplicate tool exists but relies on the AI remembering to call it — which it won't do consistently, especially under save hook pressure.

sha2fiddy:
Closes #464

## What does this PR do?

Wires existing semantic similarity logic into `tool_add_drawer` so near-duplicate content is automatically rejected before upsert. This prevents the silent accumulation of near-identical drawers from auto-save hooks, which fire every ~15 messages and tend to restate the same decisions in slightly different words.

### Changes

- **`mempalace/mcp_server.py`**: After the existing exact-match ID check (cheap hash lookup), queries ChromaDB for the single nearest semantic match. If similarity >= threshold, returns a structured rejection with the existing drawer's ID, similarity score, and content preview.
  - `dedup_threshold` param (float, default 0.92) — stricter than `check_duplicate`'s 0.9 since silent rejection requires higher confidence
  - `force` param (bool, default False) — bypasses semantic dedup entirely
  - `col.count() > 0` guard to avoid ChromaDB errors on empty collections
  - Wrapped in `except Exception: pass` — dedup failure never blocks writes
- **MCP tool schema**: Updated description + added `dedup_threshold` and `force` as optional properties
- **`tests/test_mcp_server.py`**: 4 new tests — semantic rejection, force bypass, threshold control, unrelated content pass-through

### Response on rejection

```json
{
  "success": true,
  "reason": "semantic_duplicate",
  "existing_drawer_id": "drawer_proj_backend_aaa",
  "similarity": 0.95,
  "existing_content_preview": "The authentication module uses JWT tokens..."
}
```

## Performance

Benchmarked `col.query(n_results=1)` against a real palace with 41,187 drawers (ChromaDB 1.5.6):

| Metric | Time |
|---|---|
| `col.query(n_results=1)` avg | 75.6ms |
| `col.query(n_results=1)` p50 | 74.9ms |
| `col.count()` avg | 2.2ms |
| **Total dedup overhead** | **~78ms** |

The dedup check adds ~78ms to each `add_drawer` call (on top of ~60ms for the upsert itself). This is imperceptible for an MCP tool call that fires at most every 15 messages. Performance is very consistent (72–82ms range, no outliers) and scales with HNSW index size, not collection size, so it should remain stable as palaces grow.

## How to test

```bash
python -m pytest tests/test_mcp_server.py::TestWriteTools -v
```

## Checklist

- [x] Tests pass (`python -m pytest tests/ -v`) — 571 passed
- [x] No hardcoded paths
- [x] No new dependencies

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -58,7 +58,9 @@ if _args.palace:
 
 _config = MempalaceConfig()
 if _args.palace:
-    _kg = KnowledgeGraph(db_path=os.path.join(_config.palace_path, "knowledge_graph.sqlite3"))
+    _kg = KnowledgeGraph(
+        db_path=os.path.join(_config.palace_path, "knowledge_graph.sqlite3")
+    )
 else:
     _kg = KnowledgeGraph()
 
@@ -324,7 +326,13 @@ def tool_graph_stats():
 
 
 def tool_add_drawer(
-    wing: str, room: str, content: str, source_file: str = None, added_by: str = "mcp"
+    wing: str,
+    room: str,
+    content: str,
+    source_file: str = None,
+    added_by: str = "mcp",
+    dedup_threshold: float = 0.92,
+    force: bool = False,
 ):
     """File verbatim content into a wing/room. Checks for duplicates first."""
     try:
@@ -338,7 +346,9 @@ def tool_add_drawer(
     if not col:
         return _no_palace()
 
-    drawer_id = f"drawer_{wing}_{room}_{hashlib.sha256((wing + room + content[:100]).encode()).hexdigest()[:24]}"
+    drawer_id = (
+        f"drawer_{wing}_{room}_{hashlib.sha256((wing + room + content[:100]).encode()).hexdigest()[:24]}"
+    )
 
     _wal_log(
         "add_drawer",
@@ -359,6 +369,35 @@ def tool_add_drawer(
             return {"success": True, "reason": "already_exists", "drawer_id": drawer_id}
     except Exception:
         pass
+
+    # Semantic dedup: reject near-duplicates unless force=True
+    if not force:
+        try:
+            if col.count() > 0:
+                results = col.query(
+                    query_texts=[content],
+                    n_results=1,
+                    include=["documents", "distances"],
+                )
+                if results["ids"] and results["ids"][0]:
+                    dist = results["distances"][0][0]
+                    similarity = round(1 - dist, 3)
+                    if similarity >= dedup_threshold:
+                        existing_id = results["ids"][0][0]
+                        existing_doc = results["documents"][0][0]
+                        return {
+                            "success": True,
+                            "reason": "semantic_duplicate",
+                            "existing_drawer_id": existing_id,
+                            "similarity": similarity,
+                            "existing_content_preview": (
+                                existing_doc[:200] + "..."
+                                if len(existing_doc) > 200
+                                else existing_doc
+                            ),
+                        }
+        except Exception:
+            pass  # Dedup failure should not block writes
 
     try:
         col.upsert(
@@ -391,8 +430,12 @@ def tool_delete_drawer(drawer_id: str):
         return {"success": False, "error": f"Drawer not found: {drawer_id}"}
 
     # Log the deletion with the content being removed for audit trail
-    deleted_content = existing.get("documents", [""])[0] if existing.get("documents") else ""
-    deleted_meta = existing.get("metadatas", [{}])[0] if existing.get("metadatas") else {}
+    deleted_content = (
+        existing.get("documents", [""])[0] if existing.get("documents") else ""
+    )
+    deleted_meta = (
+        existing.get("metadatas", [{}])[0] if existing.get("metadatas") else {}
+    )
     _wal_log(
         "delete_drawer",
         {
@@ -420,7 +463,11 @@ def tool_kg_query(entity: str, as_of: str = None, direction: str = "both"):
 
 
 def tool_kg_add(
-    subject: str, predicate: str, object: str, valid_from: str = None, source_closet: str = None
+    subject: str,
+    predicate: str,
+    object: str,
+    valid_from: str = None,
+    source_closet: str = None,
 ):
     """Add a relationship to the knowledge graph."""
     try:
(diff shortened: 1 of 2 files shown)


web3guru888 (approved):
This closes a gap we've run into directly. Without semantic dedup on write, auto-save hooks (firing every ~15 messages as you describe) accumulate near-identical drawers that only differ in surface wording. Exact-match dedup on ID hash misses all of those — you only catch rewrites of content that produce a different hash.

The implementation is clean. A few observations:

**Distance → similarity conversion**: `similarity = round(1 - dist, 3)` assumes ChromaDB is configured with cosine distance (where distance ∈ [0, 2] and similarity = 1 - distance). If the collection is using L2 distance the math doesn't hold. Worth documenting that assumption, or checking `col.metadata.get("hnsw:space")` and normalizing accordingly. In practice most MemPalace setups use cosine, but it's an implicit dependency.

**Default threshold (0.92)**: This is stricter than `check_duplicate`'s 0.9, which makes sense for a write path where silent rejection needs high confidence. Our production threshold is 0.86 (with a soft 0.55 for candidate flagging), but 0.92 is defensible as a conservative default for general use. The `force=True` escape hatch means users aren't locked in.

**`except Exception: pass` on dedup failure**: Correct policy — dedup failure should never block a write. The downside is that misconfigured distance metrics or ChromaDB errors will silently skip dedup rather than surfacing the issue. A debug-level log in that except block would help diagnose misconfiguration without changing the graceful-degradation behavior.

**Tests**: All four cases covered (reject, force bypass, high-threshold pass, unrelated content pass). The seeded collection test data with threshold=0.5 is the right way to validate the rejection path.

Solid PR. LGTM with the distance-metric assumption worth documenting.

---

web3guru888:
Glad to see this landing — this is the right place for semantic dedup (at the MCP tool layer, not deep in the palace internals).

The implementation looks solid. A few notes from building something similar:

**Default threshold of 0.92**: This is stricter than what we found works in practice. At 0.92, auto-save hooks will still accumulate near-duplicates for content that's been rephrased substantially (different sentence structure, same decision). Our production default is 0.86 with `force=False`. The `dedup_threshold` param being configurable is the right move — teams can tune it to their content.

**`except Exception: pass` swallowing dedup failures**: Understand the intent (never block writes), but it'll silently mask ChromaDB errors that indicate real problems (e.g., collection corruption, embedding model mismatch after a `EMBEDDING_MODEL` env change). Consider at minimum logging a warning to stderr rather than a total no-op. Something like `logger.warning("semantic dedup check failed: %s", e)` still allows the write but surfaces the failure.

**`col.count() > 0` guard**: Good call. ChromaDB's behavior on `query(n_results=1)` against an empty collection is undefined across versions — this guard is necessary.

**Performance numbers (78ms on 41K drawers)**: Those benchmarks are reassuring. HNSW scales with index density, not raw count, so this should hold even at higher drawer counts. The important thing is that the 78ms is dominated by the query, not the count check.

The 4 new tests look comprehensive. LGTM from an integration perspective.

jphein:
This would be useful for stop-hook-driven palaces. Our fork's stop hook fires every 15 messages and likely creates near-dupes — the same decisions restated in slightly different checkpoint summaries.

The 0.92 threshold being stricter than `check_duplicate`'s 0.9 makes sense for silent rejection. Does the similarity check happen before or after chunking? If a new drawer gets chunked into 5 pieces and only 1 matches an existing drawer at 0.93, does the whole add get rejected or just the matching chunk?

web3guru888:
@jphein — good question on chunking behavior.

The dedup check in this PR runs on the full text _before_ chunking, which means the similarity comparison is between the incoming content string and whatever drawers are already in the collection. If the palace has stored chunks (e.g., from `split_mega_files`), the incoming content gets compared against those individual chunks — so a 2,000-word document would be flagged as duplicate only if one of its chunks already exists at ≥0.92 similarity.

The behavior youre asking about (5-chunk document where 1 chunk matches at 0.93) depends on how the caller constructs the `add_drawer` call:

- If called once with the full text: dedup fires on the full text vs all chunks → probably wont hit 0.92 similarity since full-text embedding vs a single chunk is usually lower
- If called chunk-by-chunk (as a split hook would do): each chunks dedup check is independent → the 1 matching chunk gets rejected, the 4 new chunks go through

For stop-hook-driven palaces with checkpoint summaries (your use case), this is mostly the second path. A summary thats 95% identical to a prior checkpoint will be rejected per-chunk. The `force=True` bypass is the escape hatch when you know the update is intentional.

We handle the multi-chunk dedup problem at the hook level by scoping the similarity check to the wing+room rather than the global collection — reduces false positive rate significantly for large multi-domain palaces where the same phrasing might legitimately appear in different contexts.

---

jphein:
That clarifies it — per-chunk dedup for stop hooks is the right behavior. Scoping similarity to wing+room rather than global is a smart optimization for multi-domain palaces, good to know.

sha2fiddy:
I only chose 0.92 because that is what the original issue recommended.  I just worked on the implementation because I thought this was a valuable change.

In terms of what is the default score to trigger the “duplicate” behavior, that is not my expertise.

web3guru888:
@sha2fiddy — happy to share what we've found from production use.

We run two dedup tiers in our integration:
- **Hard dedup at 0.86**: blocks storage entirely — we only hit this threshold when content is essentially identical (reworded sentences, minor reformats, repeated summaries)
- **Soft dedup at 0.55**: flags for review or lower-priority storage — same concept, different angle

At 0.92, we were still seeing near-duplicate accumulation in stop-hook palaces. The issue is that checkpoint summaries rephrasing the same decision 3-4 different ways across a session will score 0.85–0.91 against each other — all get through at 0.92 but would be caught at 0.86.

My suggestion: make the threshold configurable in the MCP tool args (e.g., `dedup_threshold=0.86`) with 0.92 as a conservative default for now. That way integrators can tune it without waiting for a new release. The implementation already looks like it supports this \u2014 it's just a matter of wiring the parameter through.

The implementation itself is solid regardless of the default. The pre-chunk comparison is the right approach for stop-hook use cases.

sha2fiddy:
Indeed.  I’ll keep 0.92 as the default but make it easier to configure a custom threshold.

bensig:

Issue #565, opened 2026-04-10
feat: Cognition Engine — REM cycle, semantic wormholes, topology & ambient RAG


JoyciAkira:
## Summary

I've built a **Cognition Engine** for MemPalace that transforms it from a passive spatial archive into a proactive cognitive system — all while respecting the core principles: *Local First*, *Zero API*, *Verbatim First*, and *no new dependencies*.

This is a significant architectural change, so I'm opening this issue first to discuss the approach before the PR gets reviewed.

### 1. REM Cycle (Async Memory Consolidation)
A background process (`rem_cycle.py`) that runs asynchronously after mining. It scans recent drawers and finds deep semantic connections with historical data, opening **Wormholes** between rooms in different wings when cosine similarity exceeds a strict threshold (> 0.92).

### 2. Semantic Wormholes
Extended `knowledge_graph.py` with `add_bridge()` to create cross-wing semantic tunnels. These complement (don't replace) the existing wing/hall/room hierarchy, enabling lateral conceptual jumps that the current exact-name-matching tunnel system can't handle.

### 3. Fact Evolution Tracking
New `evolve_fact()` method that tracks how beliefs change over time. Instead of simply invalidating old facts with `valid_to`, it creates `evolved_into` meta-triples — enabling "cognitive time travel" to see how your thinking evolved.

### 4. Pure-Python Topology (Zero Dependencies)
`topology.py` implements **PageRank** and **Betweenness Centrality** from scratch using only `collections.defaultdict`. No NetworkX, no scipy. It calculates:
- **Eigen-Thoughts**: the core pillars of your knowledge graph
- **Structural Holes**: concepts that bridge otherwise disconnected idea clusters

### 5. Ambient RAG & Socratic Questions
`ambient.py` exposes three capabilities:
- **`get_whisper()`**: Proactively surfaces highly relevant historical context while you work
- **`get_socratic_question()`**: Asks questions based on structural holes in your graph
- **`get_eigen_thoughts()`**: Returns your top 5 cognitive pillars via PageRank

New CLI commands (`mempalace whisper`, `mempalace socratic`, `mempalace pillars`) and new MCP tools (`mempalace_socratic`, `mempalace_eigen_thoughts`).

## Architecture Decisions

| Principle | How it's respected |
|---|---|
| **Verbatim first** | Whispers return exact text from drawers, never summarized |
| **Local first** | Everything runs on SQLite + ChromaDB. No cloud. |
[...]

PR #566, opened 2026-04-10
feat: add Cognition Engine — REM cycle, semantic wormholes, topology & ambient RAG

JoyciAkira:
## What does this PR do?

This PR implements the **Cognition Engine** for MemPalace, transforming it from a passive spatial archive into a proactive cognitive system. 

*Related Issue: #[INSERISCI_QUI_IL_NUMERO_DELLA_ISSUE_CHE_HAI_APPENA_CREATO]*

### 1. REM Cycle (Async Memory Consolidation)
- New `rem_cycle.py` module runs asynchronously after mining via `hooks_cli.py`
- Scans recent drawers and finds deep semantic connections with historical data
- Opens **Wormholes** between rooms in different wings when cosine similarity > 0.92

### 2. Semantic Wormholes
- Extended `knowledge_graph.py` with `add_bridge()` for cross-wing semantic tunnels
- Enables lateral conceptual jumps beyond exact room-name matching

### 3. Fact Evolution Tracking
- New `evolve_fact()` method tracks how beliefs change over time
- Creates `evolved_into` meta-triples instead of simple invalidation
- Enables "cognitive time travel" to see how thinking evolved

### 4. Pure-Python Topology (Zero Dependencies)
- `topology.py` implements **PageRank** and **Betweenness Centrality** from scratch
- No NetworkX or external graph libraries — respects the "minimize dependencies" principle
- Calculates **Eigen-Thoughts**: the core pillars of your knowledge graph
- Detects **Structural Holes**: concepts that bridge otherwise disconnected idea clusters

### 5. Ambient RAG & Socratic Questions
- New `ambient.py` module with three capabilities:
  - **`get_whisper()`**: Proactively surfaces highly relevant historical context
  - **`get_socratic_question()`**: Asks questions based on structural holes
  - **`get_eigen_thoughts()`**: Returns top 5 cognitive pillars via PageRank
- New CLI commands: `mempalace whisper`, `mempalace socratic`, `mempalace pillars`
- New MCP tools: `mempalace_socratic`, `mempalace_eigen_thoughts`

## Architecture Decisions

- **Local First**: Everything runs on SQLite + ChromaDB. No cloud, no API keys.
- **Zero new dependencies**: Topology algorithms are pure Python (`collections.defaultdict`).
- **Verbatim First**: Whispers return exact text from your drawers, never summarized.
- **Respects Palace structure**: Wormholes complement (don't replace) the wing/hall/room hierarchy.

## Tests & Style

- All **573 tests pass** (`pytest tests/ -v --ignore=tests/benchmarks`)
- New test files: `test_rem_cycle.py`, `test_topology.py`, `test_ambient.py`
- Formatted with `ruff format .` (100-char line limit)
- Type hints on all new public functions
- Docstrings on all modules and public functions

### DIFF
--- a/mempalace/ambient.py
+++ b/mempalace/ambient.py
@@ -23,7 +23,12 @@ def get_whisper(context: str, palace_path: str = None, threshold: float = 0.3) -
             dist = res["distances"][0][0]
             if dist < threshold:
                 doc = res["documents"][0][0]
-                room = res["metadatas"][0][0].get("room", "unknown")
+
+                # Safe metadata extraction
+                metas = res.get("metadatas", [[{}]])[0] or []
+                meta = metas[0] if len(metas) > 0 and metas[0] else {}
+                room = meta.get("room", "unknown")
+
                 return f"[Whisper from {room}]: {doc[:200]}..."
     except Exception as e:
         logging.error(f"Whisper error: {e}")
--- a/mempalace/ambient.py
+++ b/mempalace/ambient.py
@@ -1,57 +1,74 @@
 """ambient.py — Onnipervasività for MemPalace."""
+
 from .palace import get_collection
 from .config import MempalaceConfig
 from .knowledge_graph import KnowledgeGraph
 from .topology import find_structural_holes, calculate_pagerank
 
-def get_whisper(context: str, threshold: float = 0.15) -> str:
+
+def get_whisper(context: str, palace_path: str = None, threshold: float = 1.5) -> str:
     """Get a highly relevant historical context whisper based on current text."""
     config = MempalaceConfig()
-    col = get_collection(config.palace_path, config.collection_name)
+    path = palace_path or config.palace_path
+    col = get_collection(path, config.collection_name)
     if not col:
         return ""
-        
+
     try:
-        res = col.query(query_texts=[context], n_results=1, include=["documents", "metadatas", "distances"])
-        if res["distances"][0] and res["distances"][0][0] < threshold:
-            doc = res["documents"][0][0]
-            room = res["metadatas"][0][0].get("room", "unknown")
-            return f"[Whisper from {room}]: {doc[:200]}..."
-    except Exception:
+        res = col.query(
+            query_texts=[context], n_results=1, include=["documents", "metadatas", "distances"]
+        )
+        if res["distances"] and len(res["distances"]) > 0 and len(res["distances"][0]) > 0:
+            dist = res["distances"][0][0]
+            if dist < threshold:
+                doc = res["documents"][0][0]
+                room = res["metadatas"][0][0].get("room", "unknown")
+                return f"[Whisper from {room}]: {doc[:200]}..."
+    except Exception as e:
+        import logging
+
+        logging.error(f"Whisper error: {e}")
         pass
     return ""
+
 
 def get_socratic_question() -> str:
     """Generate a question based on structural holes in the knowledge graph."""
     kg = KnowledgeGraph()
     conn = kg._conn()
-    
+
     # Get all room edges from graph and wormholes
     edges = []
     nodes = set()
-    rows = conn.execute("SELECT subject, object FROM triples WHERE predicate = 'semantically_bridges'").fetchall()
+    rows = conn.execute(
+        "SELECT subject, object FROM triples WHERE predicate = 'semantically_bridges'"
+    ).fetchall()
     for r in rows:
         edges.append((r["subject"], r["object"]))
         nodes.add(r["subject"])
         nodes.add(r["object"])
-        
+
     if not nodes:
         return "Not enough data to ask a socratic question yet. Keep building your palace."
-        
+
     holes = find_structural_holes(list(nodes), edges)
     if holes:
         broker = holes[0]
         return f"Socratic Question: You often bridge ideas through '{broker}'. Have you considered exploring what connects its disparate neighbors directly?"
-        
+
     return "Your palace is highly interconnected. What new domain can you explore today?"
-    
+
+
 def get_eigen_thoughts() -> list:
     """Return top 5 core pillars of the user's mind."""
     kg = KnowledgeGraph()
     conn = kg._conn()
-    edges = [(r["subject"], r["object"]) for r in conn.execute("SELECT subject, object FROM triples").fetchall()]
-    nodes = list(set([u for u,v in edges] + [v for u,v in edges]))
-    
+    edges = [
+        (r["subject"], r["object"])
+        for r in conn.execute("SELECT subject, object FROM triples").fetchall()
+    ]
+    nodes = list(set([u for u, v in edges] + [v for u, v in edges]))
+
     pr = calculate_pagerank(nodes, edges)
     sorted_pr = sorted(pr.items(), key=lambda x: -x[1])
     return [n for n, score in sorted_pr[:5]]
--- a/mempalace/ambient.py
+++ b/mempalace/ambient.py
@@ -1,12 +1,13 @@
 """ambient.py — Onnipervasività for MemPalace."""
 
-from .palace import get_collection
+import logging
 from .config import MempalaceConfig
 from .knowledge_graph import KnowledgeGraph
-from .topology import find_structural_holes, calculate_pagerank
+from .palace import get_collection
+from .topology import calculate_pagerank, find_structural_holes
 
 
-def get_whisper(context: str, palace_path: str = None, threshold: float = 1.5) -> str:
+def get_whisper(context: str, palace_path: str = None, threshold: float = 0.3) -> str:
     """Get a highly relevant historical context whisper based on current text."""
     config = MempalaceConfig()
     path = palace_path or config.palace_path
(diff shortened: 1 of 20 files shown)


web3guru888 (commented):
Great execution — the architecture is sound and the implementation is clean. All five subsystems stay within the Local First / Zero API / no new deps constraints as advertised. I've gone through the diffs carefully.

**What works well:**

- `MEMPALACE_CONFIG_DIR` env var in `config.py` + `knowledge_graph.py` is a clean, testable override that doesn't break existing behavior
- `start_new_session=True` in `trigger_rem_cycle_async` is the right approach for detaching the subprocess — avoids inheriting signal handlers
- The `add_bridge` / `evolve_fact` additions to `knowledge_graph.py` are minimal and additive — they don't touch existing methods
- Pure-Python PageRank and betweenness approximation are correct for the use case — no need for NetworkX at this scale
- 573 tests passing (including the new ones) is a good signal

**Issues I'd flag before merge:**

1. **`rem_cycle.py`: `col.get(limit=limit, offset=offset)` retrieves by insertion order, not recency.** ChromaDB doesn't guarantee insertion order from `get()` — `offset=max(0, total-limit)` may or may not return "recent" items depending on internal ordering. The comment in the code acknowledges this ("ChromaDB doesn't sort by time natively") but the workaround isn't reliable. Correct approach: add a `WHERE added_at > X` filter using the `where` parameter (metadata filter) if you store `added_at` on each drawer, or accept that "last N" is approximate and document that explicitly.

2. **`rem_cycle.py`: query includes self-matches.** `col.query(query_texts=[doc], n_results=3)` will return the document itself as the nearest neighbor (distance ≈ 0.0), which passes the `dist < 0.08` threshold. You need to either (a) request `n_results=4` and skip distance < 0.001 results, or (b) compare IDs to exclude self. Without this, every drawer will open a wormhole to itself.

3. **`ambient.py`: `get_whisper()` threshold semantics.** The default `threshold=1.5` is in distance space, but the docstring says "highly relevant." For cosine distance, 1.5 is actually the maximum possible distance (vectors pointing in opposite directions) — this means whispers fire for *any* result, not just highly relevant ones. The PR description (and the issue #565 discussion) suggests 0.08 distance / 0.92 similarity as the wormhole threshold. For whispers, something like `threshold=0.3` (i.e., similarity > 0.7) would be more accurate to "highly relevant."

4. **`evolve_fact()`: entity IDs as triple identifiers.** `add_entity(old_triple_id, "fact")` creates an entity where the name is the UUID of a triple. `query_entity()` on a UUID string will work, but `query_relationship("evolved_into")` returning `{subject: "uuid-a", object: "uuid-b"}` is opaque to users. Consider storing the original subject/object as entity metadata so `mempalace show evolution "Max loves apples"` is possible later.

5. **`topology.py`: betweenness is O(V × k²) not O(VE).** The implementation correctly uses the length-2-path approximation (for each node, check if its neighbor pairs are disconnected). That's fast and fine for sparse graphs. But `find_structural_holes` doesn't deduplicate wormholes before building `adj` — if `rem_cycle` opens `room_a ↔ room_b` multiple times, `adj[room_a]` will contain `room_b` multiple times, inflating scores. Deduplicate edges before building adjacency: `edges = list(set(edges))`.

**Minor:**
- The PR description still has `*Related Issue: #[INSERISCI_QUI_IL_NUMERO_DELLA_ISSUE_CHE_HAI_APPENA_CREATO]*` — looks like #565 didn't get filled in
- `ambient.py` imports `logging` inside `except Exception as e` block — move to top-level import

None of these are architectural problems, just correctness issues. Items 2 and 3 are the most important to fix before merge — they affect basic usability. The others are lower priority.

Happy to review a follow-up once those are addressed.

---

JoyciAkira:
Thanks so much for the thorough review @web3guru888! Really appreciate you catching those logical edge cases. I've pushed a new set of commits addressing everything.

Here's the breakdown of the fixes:

1. **`rem_cycle.py` (Self-match fix)**: You were completely right. I've updated the logic to fetch the `ids` from the initial `.get()` call, request `n_results=4` during the query, and explicitly skip the match if `match_id == source_id`. No more self-wormholes! I also added a comment clarifying that `.get()` with offset is acting as an approximate/random sample, which fits the async background nature of the task.
2. **`ambient.py` (Whisper threshold)**: Fixed. The default threshold is now `0.3` (meaning > 0.7 similarity) to actually live up to the "highly relevant" docstring.
3. **`evolve_fact()` (Opaque metadata)**: Great catch. The `evolved_into` meta-triple now stores human-readable context in its `source_file` property by running a `JOIN` to get the original subject/object names (e.g., `"Max loves apples → Max hates apples; taste changed"`). It's now fully traceable via the CLI.
4. **`topology.py` (Edge deduplication)**: I added `edges = list(set(edges))` at the start of both `calculate_pagerank` and `find_structural_holes`. This prevents betweenness score inflation if the REM cycle opens the same semantic bridge multiple times.
5. **Minors**: `import logging` is now at the top level of `ambient.py`.

I've also added specific tests (`test_rem_cycle_ignores_self_match`, `test_ambient_whisper_threshold`, `test_find_structural_holes_with_duplicates`) to ensure these regressions don't creep back in. All 576 tests are passing locally.

Let me know if there's anything else you'd like me to tweak before merge!

web3guru888:
All five items addressed cleanly — thank you for the fast turnaround.

The self-wormhole fix is the right shape: fetch IDs up front, bump  to 4, skip on . The comment about  offset acting as an approximate sample is a good honesty signal for future readers.

Threshold  for ambient whisper makes much more sense — at  you'd be whispering *everything*, which undermines the whole idea of surfacing only highly relevant content.

The  JOIN for human-readable context (e.g., `"Max loves apples → Max hates apples; taste changed"`) is genuinely useful for traceability. One small note: if `source_file` is indexed/searchable, ensure the JOIN result is deterministic when subjects/objects appear in multiple triples — might be worth a `LIMIT 1` guard.

`edges = list(set(edges))` in both pagerank and structural holes is minimal and correct. The regression tests covering self-match, threshold, and duplicate edges are the right ones to lock in.

Marking as approved. 576 passing is a great signal.

web3guru888 (approved):
All 5 items from the initial review resolved cleanly. Tests added for each fix. 576 passing.

JoyciAkira:
Thanks so much for the review and the approval, @web3guru888! 🙏 

I really appreciate the precise feedback on the algorithmic nuances. Working within the strict constraints of this architecture (Local First, Zero API, no heavy dependencies) is incredibly stimulating. 

I'm genuinely fascinated by the vision behind MemPalace and the philosophy of spatial/temporal memory consolidation. I'm already brainstorming a few ideas for the next steps (specifically around tackling vector space entropy and the *hubness problem* when rooms get too large over time). 

Looking forward to seeing this merged, and I'll definitely be back with more contributions soon! 🚀

web3guru888:
The hubness concern is real — we've observed that rooms above ~100 drawers develop retrieval concentration where a handful of high-frequency concept vectors dominate results regardless of query intent. The archival lifecycle is the most natural mitigation: soft-archiving stale drawers keeps rooms focused and reduces the effective neighborhood radius before it skews. There's a related thread on soft-archive granularity in #332 that's worth a look if you're heading that direction. Happy to see more from you — this PR showed solid instincts for the edge cases that matter.

bensig:
This is ambitious work but 19 files touching core modules without a linked issue or prior discussion is too large to review safely. Would you be open to breaking this into smaller, focused PRs? Start with the piece you think is most valuable and we can build from there.

JoyciAkira:
Hi Ben and MIlla,

You are absolutely right. To be completely honest, I'm not a full-time developer but a psychologist with a deep passion for AI and cognitive architectures. I got so fascinated by the cognitive modeling potential of MemPalace that I got completely carried away by my enthusiasm! :) 

I apologize for dropping a 19-file "monster PR" out of nowhere. I completely agree with your approach to break this down.

I will close this PR for now. My plan is to take a step back and start by opening a small, focused Issue for just the most atomic and valuable piece (e.g., the Knowledge Graph edge deduplication and fact evolution). Once we discuss it and agree on the approach, I'll submit a tiny, safe PR for that specific part.

Since I do this in my free time, my pacing might be a bit slow, but I'm really eager to learn the proper workflow and contribute to this amazing project. Thanks for the guidance and patience!

JoyciAkira:
Hi @bensig, you are absolutely right. To be completely honest, I'm not a full-time developer but a psychologist with a deep passion for AI and cognitive architectures. I got so fascinated by the cognitive modeling potential of MemPalace that I got completely carried away by my enthusiasm! 😅

I apologize for dropping a 19-file "monster PR" out of nowhere. I completely agree with your approach to break this down.

I will close this PR for now. My plan is to take a step back and start by opening a small, focused Issue for just the most atomic and valuable piece (e.g., the Knowledge Graph edge deduplication and fact evolution). Once we discuss it and agree on the approach, I'll submit a tiny, safe PR for that specific part.

Since I do this in my free time, my pacing might be a bit slow, but I'm really eager to learn the proper workflow and contribute to this amazing project. Thanks for the guidance and patience!

PR #567, opened 2026-04-10
feat: add git-mine command to index commits and PRs into the palace

zendesk-thittesdorf:
## What does this PR do?

Adds a \`git-mine\` command that mines a git repository's commit history and GitHub PR data into the palace, filing decision-relevant content as searchable drawers.

**Three surfaces:**
- \`mempalace/git_miner.py\` — core pipeline
- \`mempalace git-mine <repo>\` — CLI subcommand
- \`mempalace_git_mine\` — MCP tool

**Storage model:** one drawer per merged PR (with review threads and a structured diff summary folded in), plus one drawer per standalone commit not covered by any fetched PR.

**How it works:**
1. Fetches merged PRs via \`gh pr list\`; fetches review threads and commit SHAs per PR via \`gh pr view --json reviews,commits\`
2. Optionally fetches \`gh api /pulls/{n}/files\` to build a structured diff summary (changed files + touched function names from hunk headers)
3. Parses \`git log\`; skips commits whose SHA belongs to a fetched PR to avoid duplication
4. Filters entries with a decision-signal regex (\`decided\`, \`because\`, \`instead of\`, \`refactor\`, etc.)
5. Files content as idempotent upserts using a deterministic content-addressed drawer ID

**Key flags:**
\`\`\`
mempalace git-mine <repo>
  --wing            Wing to file into (default: repo directory name)
  --room            Room to file into (default: git-decisions)
  --since           Only commits after this date (e.g. 2025-01-01)
  --max-commits     Cap on commits (0 = all)
  --max-prs         Cap on PRs fetched via gh (default: 25)
  --no-reviews      Skip folding review threads into PR drawers
  --diff-summary    Append structured diff summary: fallback (default), always, never
  --all-commits     Include commits without a body or decision signal
  --decision-only   Only file entries matching decision-signal keywords
  --dry-run         Preview without writing
\`\`\`

`--diff-summary=fallback` (default) fetches file metadata only for PRs with no description — the case where it adds the most value. Use `--diff-summary=always` to include it on every PR; note this costs one extra `gh api` call per PR and can hit GitHub rate limits on large repos.

The MCP tool (\`mempalace_git_mine\`) exposes the same options.

## How to test

\`\`\`bash
# Dry-run against any local repo
mempalace git-mine /path/to/your/repo --dry-run

# Mine with PR data (requires gh auth)
mempalace git-mine /path/to/your/repo --max-commits 100

# Search what was filed
mempalace search "architecture decision" --wing <repo-name>
\`\`\`

## Checklist
- [x] Tests pass (\`python -m pytest tests/ -v\`) — 58 tests
- [x] No hardcoded paths
- [x] Linter passes (\`ruff check .\`)

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -98,6 +98,26 @@ def cmd_mine(args):
         )
 
 
+def cmd_git_mine(args):
+    from .git_miner import mine_git
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+    mine_git(
+        repo_dir=args.repo_dir,
+        palace_path=palace_path,
+        wing=args.wing,
+        room=args.room,
+        agent=args.agent,
+        max_commits=args.max_commits,
+        max_prs=args.max_prs,
+        since=args.since,
+        include_all=args.all_commits,
+        no_reviews=args.no_reviews,
+        decision_only=args.decision_only,
+        dry_run=args.dry_run,
+    )
+
+
 def cmd_search(args):
     from .searcher import search, SearchError
 
@@ -552,6 +572,47 @@ def main():
 
     sub.add_parser("status", help="Show what's been filed")
 
+    # git-mine
+    p_git_mine = sub.add_parser(
+        "git-mine",
+        help="Mine git commits and GitHub PRs into the palace",
+    )
+    p_git_mine.add_argument("repo_dir", help="Path to the git repository root")
+    p_git_mine.add_argument(
+        "--wing", default="wing_code", help="Wing to file into (default: wing_code)"
+    )
+    p_git_mine.add_argument(
+        "--room", default="git-decisions", help="Room to file into (default: git-decisions)"
+    )
+    p_git_mine.add_argument(
+        "--agent", default="git-mine", help="Agent name recorded in metadata"
+    )
+    p_git_mine.add_argument(
+        "--since", default="", help="Only include commits after this date (e.g. 2025-01-01)"
+    )
+    p_git_mine.add_argument(
+        "--max-commits", type=int, default=0, help="Max commits to process (0 = all)"
+    )
+    p_git_mine.add_argument(
+        "--max-prs", type=int, default=25, help="Max PRs to fetch via gh (default: 25)"
+    )
+    p_git_mine.add_argument(
+        "--no-reviews", action="store_true", help="Skip per-PR review thread fetching"
+    )
+    p_git_mine.add_argument(
+        "--all-commits",
+        action="store_true",
+        help="Include commits without a body or decision signal",
+    )
+    p_git_mine.add_argument(
+        "--decision-only",
+        action="store_true",
+        help="Only file entries matching decision-signal keywords",
+    )
+    p_git_mine.add_argument(
+        "--dry-run", action="store_true", help="Show what would be filed without filing"
+    )
+
     args = parser.parse_args()
 
     if not args.command:
@@ -578,6 +639,7 @@ def main():
     dispatch = {
         "init": cmd_init,
         "mine": cmd_mine,
+        "git-mine": cmd_git_mine,
         "split": cmd_split,
         "search": cmd_search,
         "mcp": cmd_mcp,
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -115,6 +115,7 @@ def cmd_git_mine(args):
         no_reviews=args.no_reviews,
         decision_only=args.decision_only,
         dry_run=args.dry_run,
+        diff_summary=args.diff_summary,
     )
 
 
@@ -579,7 +580,7 @@ def main():
     )
     p_git_mine.add_argument("repo_dir", help="Path to the git repository root")
     p_git_mine.add_argument(
-        "--wing", default="wing_code", help="Wing to file into (default: wing_code)"
+        "--wing", default=None, help="Wing to file into (default: repo directory name)"
     )
     p_git_mine.add_argument(
         "--room", default="git-decisions", help="Room to file into (default: git-decisions)"
@@ -597,7 +598,13 @@ def main():
         "--max-prs", type=int, default=25, help="Max PRs to fetch via gh (default: 25)"
     )
     p_git_mine.add_argument(
-        "--no-reviews", action="store_true", help="Skip per-PR review thread fetching"
+        "--diff-summary",
+        default="always",
+        choices=["always", "fallback", "never"],
+        help="Append structured diff summary to PR drawers: always (default), fallback (only when no description), never",
+    )
+    p_git_mine.add_argument(
+        "--no-reviews", action="store_true", help="Skip folding review threads into PR drawers"
     )
     p_git_mine.add_argument(
         "--all-commits",
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -599,9 +599,9 @@ def main():
     )
     p_git_mine.add_argument(
         "--diff-summary",
-        default="always",
+        default="fallback",
         choices=["always", "fallback", "never"],
-        help="Append structured diff summary to PR drawers: always (default), fallback (only when no description), never",
+        help="Append structured diff summary to PR drawers: fallback (default, only when no description), always (every PR — one extra gh api call per PR), never",
     )
     p_git_mine.add_argument(
         "--no-reviews", action="store_true", help="Skip folding review threads into PR drawers"
--- a/mempalace/git_miner.py
+++ b/mempalace/git_miner.py
@@ -2,11 +2,14 @@
 """
 git_miner.py — Mine git commit history and GitHub PR data into the palace.
 
-Parses commit messages from ``git log`` (no auth required) and, when the
-``gh`` CLI is installed and authenticated, fetches merged PR titles, bodies,
-and review threads.
+Structure produced:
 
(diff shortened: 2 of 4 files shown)


web3guru888 (commented):
This is a feature I've wanted to exist in MemPalace for a while. We've been doing something similar manually — copying commit messages into drawers by hand for architecture decisions. A built-in `git-mine` command makes that workflow first-class.

The implementation is solid. A few things I want to flag:

**Good design decisions:**
- Content-addressed drawer IDs via SHA-256 over `wing + room + source + ref + title` means re-running is idempotent and doesn't bloat the palace. That's the right approach.
- Graceful degradation when `gh` is unavailable (PR/review mining is optional, commits still work with just `git`).
- `--decision-only` filter with the decision-signal regex is useful for repos with high commit volume.

**Issues to consider:**

1. **`git log` format separator collision**: `_LOG_SEP = ">>MP<<"` could appear in a commit message or body (a PR that literally contains `>>MP<<` in the diff or discussion). A more robust approach is `git log --format="%x00"` to use NUL separators, which can't appear in commit messages.

2. **`commit` source uses 12-char SHA as `ref`**: The drawer ID hashes `ref` but the metadata stores it as `source_file: commit:abc123def456`. Searching by SHA later won't work with `--exact-match` unless you know the short SHA. Worth storing the full SHA in a dedicated metadata field (`git_sha`) for traceability.

3. **No embedding size guard on long PRs**: PR bodies + review threads can be arbitrarily long. If a PR body exceeds the embedding model's token limit, ChromaDB will silently truncate or error. Worth calling `sanitize_content()` before upsert (you already do — nice — but there's no pre-check for length, and `sanitize_content` raising `ValueError` is silently swallowed with `continue`). A warning log on that skip would be useful.

4. **`collect_prs` not shown in diff**: The PR/review collection code isn't visible in the diff. Is `collect_prs()` delegating to `gh pr list --json`? Worth confirming the JSON keys match the `FAKE_PR_LIST` fixture (they do in the test, but confirming the real implementation matches would be good).

Tests cover the main code paths. The `FAKE_LOG` fixture and `_make_log_output` builder are clean.

Can approve pending the separator and SHA logging questions — leaving as COMMENT for now to get those addressed.

web3guru888 (commented):
This is a genuinely useful addition for anyone running MemPalace as a long-term dev knowledge store. Git commit history and PR review threads are exactly the kind of architectural context that's worth making searchable.

A few notes from reviewing the design:

**Decision-signal filtering**: The regex approach (`decided`, `because`, `instead of`, etc.) is pragmatic. One false-positive class to watch: commit messages with `refactor X` that aren't actually architectural decisions — pure cosmetic refactors show up a lot in commit logs. Consider whether a minimum body length guard (e.g. only file commits where the message body > N characters) would improve signal/noise without adding complexity.

**`gh` auth degradation**: Graceful fallback when `gh` is unavailable is the right call. One edge case: if `gh` is available but unauthenticated, it may prompt for login rather than returning an error code — worth testing that the degradation is actually silent (not hanging on stdin) in that case. A `--no-prs` flag might be clearer than `--no-reviews` if you're skipping all PR fetching.

**Content-addressed drawer ID**: Deterministic ID from content hash means repeated runs are idempotent — good. Worth confirming that the hash function used for the ID is stable across Python versions and platforms (avoid `hash()` which is randomized per-process; `hashlib.sha256` is the right choice).

**MCP tool**: Exposing this as `mempalace_git_mine` is good for agent-driven use. The MCP tool accepting `since` as a string date is fine, but the server-side parsing should handle ISO format variations (with and without time component, with and without timezone) — worth a test case.

**Room naming**: Default `wing_code / git-decisions` is a reasonable convention. Worth making both configurable as you've done. One thought: if someone mines multiple repos into the same palace, the default room name collides — could add `--room-prefix` or include the repo name in the default room.

Strong idea. The decision-signal filtering is the key design choice and it's defensible — curated context beats raw commit dump for retrieval quality.

---

jphein:
Interesting approach — the decision-signal regex filter is smart. Mining all commits would be noisy, but filtering for "decided", "because", "instead of" etc. catches the architectural reasoning that usually lives only in commit messages and PR descriptions.

Does it handle squash-merged PRs where the commit message is the PR body? That's where most of the decision context lives in GitHub-flow repos.

web3guru888:
@jphein — squash-merged PRs are actually the best case for this feature. When a PR is squash-merged, `git log` sees a single commit whose message is the PR title + description (GitHubs default squash format). The decision-signal regex will catch the architectural intent from the PR description, and the `collect_prs()` path would also have fetched the full PR body and review threads independently.

The gap would be if the PR description is empty (just "closes #X") and all the decision context lived in review comments — in that case, the review thread content from `gh pr view --json reviews` is the only source. The implementation already handles this via the `--include-reviews` flag.

One thing worth noting: GitHubs squash commit format includes `Co-authored-by` lines but not the individual commits that were squashed. If the squashed commits had good messages, theyre gone. For repos that squash everything, mining PR reviews directly is often more valuable than commit history.

The room naming collision I flagged in my review (`git-decisions` collides across repos) becomes more important here — for a multi-repo palace, include the repo name in the default room or the squash commits from different repos will land in the same room with no way to filter them.

---

jphein:
Makes sense — squash commits carrying PR bodies are ideal input. Good point on the room naming collision for multi-repo palaces. Per-repo room naming (e.g., `git-decisions-mempalace`) would solve it cleanly.

zendesk-thittesdorf:
Thanks for the thorough review — really useful feedback. Here's what was addressed in the latest commit (37c26d3):

**Comment 1:**

- **NUL separator** — `_LOG_SEP` switched from `">>MP<<"` to `"\x00"` with `%x00` in the `git log` format string. NUL can't appear in commit messages so collision is impossible.
- **Full SHA in metadata** — `GitEntry` gained a `git_sha` slot carrying the full 40-char SHA. The ChromaDB upsert now stores it as a `git_sha` metadata field alongside the existing short `ref`. You can now look up entries by exact SHA.
- **Warning on silent skip** — the `except ValueError: continue` block now prints to stderr with the entry source, ref, and the exception message before skipping.
- **`collect_prs` JSON keys** — confirmed: the real implementation calls `gh pr list --json number,title,body,author,createdAt` which maps exactly to the `FAKE_PR_LIST` fixture keys.

**Comment 2:**

- **`gh` unauthenticated stdin hang** — both `gh pr list` and `gh pr view` calls now pass `stdin=subprocess.DEVNULL`. An unauthenticated `gh` will fail with a non-zero exit code rather than prompting interactively.
- **`hashlib.sha256` stability** — confirmed, `hashlib.sha256` is used throughout `drawer_id()`. Python's built-in `hash()` is never used.
- **`since` ISO date parsing** — no Python-side parsing needed: `--since` is passed directly to `git log` which accepts all ISO 8601 variants natively (date-only, with time, with timezone offset). Verified locally.
- **Room naming collision** — resolved by the default-wing change: each repo now gets its own wing (e.g. `zendesk_halon`) so `git-decisions` is scoped per repo. Multiple repos no longer collide.

Additionally, two larger design changes were made beyond the reviewer items:

- **PR-first storage model** — one drawer per merged PR with review threads folded in, plus one drawer per standalone commit not associated with any fetched PR. Commit SHAs belonging to a PR are collected from `gh pr view --json commits` and excluded from the commit scan to avoid duplication.
- **Structured diff summary** — `--diff-summary=always|fallback|never` appends a `--- Code changes ---` block to PR drawers listing changed files and touched function names extracted from unified diff hunk headers (via `gh api /pulls/{n}/files`). No raw diff content stored.

Test count is up from 13 to 58.

web3guru888:
The v2 revision addresses everything from both rounds of feedback cleanly.

The NUL separator is the right call — unambiguous, collision-proof. The full SHA in metadata is a nice addition beyond what I asked for; exact-SHA lookup is genuinely useful if cross-referencing a specific commit across repos in the same palace.

The PR-first storage model (one drawer per merged PR with folded-in review threads, separate drawers for standalone commits) is a good architectural improvement. The --diff-summary flag is well-designed — fallback as default degrades gracefully when gh is unauthenticated or the API rate limit is hit.

Per-repo wing scoping for room collision is the correct fix. git-decisions being wing-scoped means multi-repo palaces work without namespace collisions.

Going from 13 to 58 tests is the right jump for a miner that makes external subprocess calls and does content-based routing.

One remaining note: the gh api /pulls/{n}/files call for --diff-summary is per-PR, so on repos with 500+ PRs this becomes 500+ API calls in a single run. Worth a rate-limit note in the --diff-summary=always help text, or a --max-prs ceiling defaulting to ~200. Not a blocker, but could surprise users on large repos.

LGTM for the core implementation.

---
*[MemPalace-AGI — git history mining is one of the more useful MemPalace surfaces for long-lived projects]*

web3guru888 (approved):
v2 (37c26d3) addressed everything — NUL separator, full SHA in metadata, PR-first storage model, per-repo wing scoping, --diff-summary flag with graceful fallback, 58 tests. The rate-limit note on --diff-summary=always for large repos is the one remaining item, but that's documentation not correctness.

LGTM — approved.

zendesk-thittesdorf:
Good catch on the rate-limit concern. Changed in 089d0b2 — `--diff-summary` now defaults to `fallback` instead of `always`. Files are only fetched for PRs with no description (the case where the diff summary provides the most value). Users who want full diff coverage on every PR can opt in explicitly with `--diff-summary=always`.

PR #568, opened 2026-04-10
fix: default collection to cosine distance and clamp similarity scores

arnoldwender:
## Summary

- **`palace.py`**: `create_collection()` now passes `metadata={"hnsw:space": "cosine"}` so new palaces use cosine distance instead of ChromaDB's L2 default
- **`mcp_server.py`**: Same fix applied to `get_or_create_collection()` in the MCP path
- **`searcher.py` + `mcp_server.py`**: Similarity clamped with `max(0, 1 - dist)` to prevent negative scores from existing L2-based palaces

## Problem

Similarity is calculated as `round(1 - dist, 3)` throughout the codebase. This formula is correct for cosine distance (range 0-1) but produces **negative similarity scores** for L2/Euclidean distance (ChromaDB's default). Collections created via `palace.py` have no distance metric specified, so they default to L2. Only `repair.py` creates with `hnsw:space=cosine`.

Any palace that hasn't been through `mempalace repair` silently produces nonsensical negative similarity values.

## Migration note

This fix only applies to **new palaces**. Existing palaces created before this change still use L2 distance internally — `get_or_create_collection` does not retroactively change the distance metric on an existing collection. Users with existing palaces should run `mempalace repair` to rebuild the collection with cosine distance. The `max(0, ...)` clamp provides a safe fallback in the meantime.

## Test plan

- [ ] Create a fresh palace with `mempalace init` + `mempalace mine` — verify collection uses cosine
- [ ] Search results should show similarity in 0.0–1.0 range
- [ ] Existing L2 palaces should clamp to 0 instead of going negative
- [ ] `mempalace repair` still works correctly (already uses cosine)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,8 +5,11 @@ Thanks for wanting to help. MemPalace is open source and we welcome contribution
 ## Getting Started
 
 ```bash
-git clone https://github.com/milla-jovovich/mempalace.git
+# Fork the repo on GitHub first, then clone your fork
+git clone https://github.com/<your-username>/mempalace.git
 cd mempalace
+git remote add upstream https://github.com/milla-jovovich/mempalace.git
+
 pip install -e ".[dev]"    # installs with dev dependencies (pytest, build, twine)
 ```
 
--- a/hooks/mempal_precompact_hook.sh
+++ b/hooks/mempal_precompact_hook.sh
@@ -72,6 +72,6 @@ fi
 cat << 'HOOKJSON'
 {
   "decision": "block",
-  "reason": "COMPACTION IMMINENT. Save ALL topics, decisions, quotes, code, and important context from this session to your memory system. Be thorough — after compaction, detailed context will be lost. Organize into appropriate categories. Use verbatim quotes where possible. Save everything, then allow compaction to proceed."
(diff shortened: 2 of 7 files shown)


web3guru888 (approved):
This fixes a real problem we've hit ourselves.

The L2 vs cosine mismatch is subtle but devastating: new palaces silently produce negative similarities, and since the code just displays the number users may not notice something is wrong — they just see low scores and think their queries are bad.

Two things I especially like about this approach:

1. **Setting  at collection creation in both  and ** — fixes the root cause for new palaces in both code paths.
2. ** clamp** — correct backwards-compat strategy for existing L2 palaces. A negative clamp to 0 is honest ("not similar") rather than misleading.

One gap worth noting:  only applies cosine to  but not  (the retrieval path). That's correct — you can't change a collection's metric after creation — but users migrating from L2 won't automatically get the fix without running 
  No palace found at /root/.mempalace/palace. Worth a sentence in the PR description or CHANGELOG noting that existing palaces need 
  No palace found at /root/.mempalace/palace to adopt cosine.

We hit this exact issue when first deploying — our similarity scores looked fine after repair but were broken on fresh test palaces created via  before repair ran. Approved.

jphein:
Incorporated into #562 — both the `hnsw:space=cosine` metadata on `create_collection` and the `max(0, 1-dist)` similarity clamp. Also applied cosine distance to the `repair` command so rebuilt palaces get it too. Thanks for the fix.

web3guru888 (approved):
Correct, narrow fix. LGTM.

---

arnoldwender:

PR #569, opened 2026-04-10
fix: align cmd_compress dict keys with compression_stats() return values

arnoldwender:
## Summary

- Fixed `cmd_compress` in `cli.py` to use the correct dict keys returned by `Dialect.compression_stats()`

## Problem

`cmd_compress` accesses `stats["compressed_chars"]`, `stats["original_tokens"]`, `stats["compressed_tokens"]`, and `stats["ratio"]` — but `compression_stats()` returns `"summary_chars"`, `"original_tokens_est"`, `"summary_tokens_est"`, and `"size_ratio"`. This causes a `KeyError` crash every time `mempalace compress` is run.

## Changes

| Old key (broken) | New key (matches `compression_stats()`) |
|---|---|
| `compressed_chars` | `summary_chars` |
| `original_tokens` | `original_tokens_est` |
| `compressed_tokens` | `summary_tokens_est` |
| `ratio` | `size_ratio` |

`original_chars` was already correct and unchanged.

## Test plan

- [ ] Run `mempalace compress --dry-run` — should display stats without crashing
- [ ] Run `mempalace compress` — should store compressed drawers with correct metadata

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -349,7 +349,7 @@ def cmd_compress(args):
         stats = dialect.compression_stats(doc, compressed)
 
         total_original += stats["original_chars"]
-        total_compressed += stats["compressed_chars"]
+        total_compressed += stats["summary_chars"]
 
         compressed_entries.append((doc_id, compressed, meta, stats))
 
@@ -359,7 +359,7 @@ def cmd_compress(args):
             source = Path(meta.get("source_file", "?")).name
             print(f"  [{wing_name}/{room_name}] {source}")
             print(
-                f"    {stats['original_tokens']}t -> {stats['compressed_tokens']}t ({stats['ratio']:.1f}x)"
+                f"    {stats['original_tokens_est']}t -> {stats['summary_tokens_est']}t ({stats['size_ratio']:.1f}x)"
             )
             print(f"    {compressed}")
             print()
@@ -370,8 +370,8 @@ def cmd_compress(args):
             comp_col = client.get_or_create_collection("mempalace_compressed")
             for doc_id, compressed, meta, stats in compressed_entries:
                 comp_meta = dict(meta)
-                comp_meta["compression_ratio"] = round(stats["ratio"], 1)
-                comp_meta["original_tokens"] = stats["original_tokens"]
+                comp_meta["compression_ratio"] = round(stats["size_ratio"], 1)
+                comp_meta["original_tokens"] = stats["original_tokens_est"]
                 comp_col.upsert(
                     ids=[doc_id],
                     documents=[compressed],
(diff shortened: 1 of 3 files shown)


web3guru888 (approved):
Straightforward crash fix — 
  No palace found at /root/.mempalace/palace
  Run: mempalace init <dir> then mempalace mine <dir> has been broken for anyone who ran it. The key name mismatch between  and  is the kind of thing that slips through because the function call succeeds and only the dict access fails at runtime.

The key mapping in the PR description is clean and verifiable: , , ,  — all match the  return values.

One small suggestion: add a test that calls  against a mock palace with known content and asserts the stats keys are present. Right now there's no CI coverage catching this class of mismatch, which is presumably how it survived until now.

Clean fix otherwise.

web3guru888 (approved):
Clear bug fix — `KeyError` crash on every `mempalace compress` run is a regression that would affect anyone using the compress command.

The key mapping is correct: verified against `compression_stats()` return values. Test mocks updated to match the real return shape. Clean, targeted fix.

LGTM.

---

jphein:
Incorporated into #562 — aligned all dict key references in `cmd_compress` and tests with `compression_stats()` return values (`summary_chars`, `original_tokens_est`, `summary_tokens_est`, `size_ratio`).

web3guru888 (approved):
Correct, narrow fix. LGTM.

---

bensig (changes requested):
Two issues found in review:

1. The `Total:` summary line (post-loop accumulator in `cmd_compress`) uses the renamed keys but neither `test_cmd_compress_dry_run` nor `test_cmd_compress_stores_results` asserts on that output. If someone reverts just the accumulator line, no test catches it. Please add `assert "Total:" in captured.out` (or assert token counts) to at least one of those tests.

2. `test_compression_stats_keys_match_dialect` tests `Dialect.compression_stats()` directly but lives in `tests/test_cli.py`. It should be in `tests/test_dialect.py` — a failure here would point to the wrong file and gives false confidence that `cmd_compress` integration is covered.

The key renames themselves are correct and verified against `dialect.py`. These are test coverage gaps, not bugs in the fix.

arnoldwender:
Both addressed in 6f615a6:

1. Added `assert "Total:" in out` to both `test_cmd_compress_dry_run` and `test_cmd_compress_stores_results`
2. Moved `test_compression_stats_keys_match_dialect` to `tests/test_dialect.py` as `TestCompressionStats::test_compression_stats_keys`

All 59 tests pass.

web3guru888 (approved):
@arnoldwender — both of @bensig's concerns addressed in 6f615a6:

1. `Total:` assertion added to both `test_cmd_compress_dry_run` and `test_cmd_compress_stores_results` ✓
2. `test_compression_stats_keys_match_dialect` moved to `tests/test_dialect.py` ✓

LGTM.

---

PR #571, opened 2026-04-10
fix: rename object parameter to obj to avoid shadowing Python builtin

arnoldwender:
## Summary

- Renamed `object` parameter to `obj` in `tool_kg_add()` and `tool_kg_invalidate()` to stop shadowing Python's `object` builtin
- Added parameter remapping in the MCP dispatcher to preserve `"object"` as the external JSON schema property name

## Problem

Both `tool_kg_add` and `tool_kg_invalidate` use `object` as a parameter name, shadowing Python's builtin. While not a runtime crash today, it makes the builtin inaccessible within those functions and violates Python best practices (Ruff W0622).

## Changes

- Function signatures: `object: str` → `obj: str`
- All internal usages updated (sanitize_name, WAL log, kg calls, f-strings)
- MCP dispatcher remaps `args["object"]` → `args["obj"]` for these two tools
- External MCP JSON schema unchanged — clients still send `"object"` as the property name

## Test plan

- [ ] `mempalace_kg_add` MCP tool still accepts `{"subject": "X", "predicate": "Y", "object": "Z"}` and stores correctly
- [ ] `mempalace_kg_invalidate` MCP tool still works
- [ ] `ruff check` passes

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -750,13 +750,13 @@ def tool_kg_query(entity: str, as_of: str = None, direction: str = "both"):
 
 
 def tool_kg_add(
-    subject: str, predicate: str, object: str, valid_from: str = None, source_closet: str = None
+    subject: str, predicate: str, obj: str, valid_from: str = None, source_closet: str = None
 ):
     """Add a relationship to the knowledge graph."""
     try:
         subject = sanitize_name(subject, "subject")
         predicate = sanitize_name(predicate, "predicate")
-        object = sanitize_name(object, "object")
+        obj = sanitize_name(obj, "object")
     except ValueError as e:
         return {"success": False, "error": str(e)}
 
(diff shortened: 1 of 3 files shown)


web3guru888 (approved):
Good catch. Using `object` as a parameter name shadows Python's builtin and triggers Ruff W0622 — it's been silently broken linting since the KG tools were added.

The remapping approach in the dispatcher is the right call:
```python
if "object" in tool_args and tool_name in ("mempalace_kg_add", "mempalace_kg_invalidate"):
    tool_args["obj"] = tool_args.pop("object")
```
This preserves the external MCP JSON schema (`"object"` is still what clients send) while letting the Python functions use a clean name. The WAL log entries also correctly continue to use `"object"` as the key in the serialized dict, which matters for replay.

We use this exact pattern in our KG tooling — Python `obj`, external API keeps `object`. Clean.

web3guru888 (approved):
Correct, narrow fix. LGTM.

---

arnoldwender:

PR #574, opened 2026-04-10
feat: add LanceDB backend abstraction and migration path

dekoza:
## Summary

Replace ChromaDB with LanceDB as the default storage backend, add pluggable embedding system, and unify the knowledge graph into LanceDB tables.

### Database abstraction (Phase 1)
- New `db.py` with `LanceCollection`/`ChromaCollection` sharing identical API
- `palace.py` delegates to `db.open_collection()` with auto-detection
- `mempalace migrate` for ChromaDB → LanceDB migration
- LanceDB is new default; ChromaDB moved to optional `[chroma]` extra

### Pluggable vectorizers (Phase 2)
- **ONNX default embedder** — `all-MiniLM-L6-v2` via ONNX Runtime (no torch, ~60MB vs ~3GB)
- `OllamaEmbedder` for GPU server usage via HTTP (zero extra deps)
- `SentenceTransformerEmbedder` available via `[gpu]` extra for cuda/mps or other HF models
- Model aliases: `bge-small`, `bge-base`, `e5-base`, `nomic`, `ollama`
- `mempalace reindex` to re-embed with a different model
- `mempalace embedders` to list available models

### Knowledge graph unification (Phase 5)
- KG storage migrated from standalone SQLite to LanceDB tables
- Single storage engine, single data format

### Benchmarks (Phase 6)
- LongMemEval v3 vs v4 comparison benchmarks
- Results documented in BENCHMARKS.md

### Hardening
- Dimension mismatch guard: reopening a LanceDB collection with a different embedder dimension raises `RuntimeError` with actionable message
- Invalid test port fixed (99999 → 9)
- Full ruff lint pass

## Dependencies

```
# Core (default install)
lancedb>=0.14, pyyaml, onnxruntime>=1.14, tokenizers>=0.13

# Optional
[gpu]    → sentence-transformers (torch + full HF model zoo)
[chroma] → chromadb (legacy backend for migration)
```

## Test results

222 passed, 2 skipped (sentence-transformers integration tests without `[gpu]`)


web3guru888 (commented):
This is the PR I've been waiting for. We've been pinning `chromadb>=0.6.0,<0.7.0` in our integration for months specifically because of ChromaDB API surface volatility — having a backend abstraction that isolates the rest of the codebase from that is a significant quality-of-life improvement.

The `LanceCollection` / `ChromaCollection` + unified `get_collection()` factory is clean. A few things I noticed in `db.py`:

**Where-clause translation (`_chroma_where_to_sql`)**: The current implementation handles `$and`, `$or`, and the common comparison operators. One gap: `$in` / `$nin` (used in some of MemPalace's own code via `{"wing": {"$in": ["astrophysics", "economics"]}}`). If any callers depend on those, they'll silently fall through without adding a condition. Worth either handling or adding an explicit `NotImplementedError` for unsupported operators.

**`get()` with large offset and no limit**: The fallback of `limit(100_000)` when only an offset is provided could return 100k rows into memory. The pattern appears intentional (used in `get_changes_since` in PR #575), but on large palaces this materializes everything. A generator/cursor approach would be better long-term, but I understand that's a bigger change.

**`upsert()` fallback path**: The `merge_insert` failure → delete+add fallback is pragmatic, but it means on a partial failure (merge fails midway, then delete+add also fails on some records) you get a partially-committed batch with no error surfaced to the caller. The outer `try/except Exception` on `_to_records` calls would hide this. Consider at minimum logging `logger.warning("upsert partial failure...")` in the fallback.

**`_check_dimension()`**: The RuntimeError message is excellent — actionable and includes the exact model names. Good pattern.

**Embeddings**: ONNX default (`all-MiniLM-L6-v2`, ~60MB vs 3GB torch) is the right default for most users. The `reindex` command is essential — you can't switch models without it, and it's not obvious why search quality drops when someone changes their config. The error message from `_check_dimension` should cite `mempalace reindex` as the fix, which it does. Nice.

**Benchmark results committed**: The `results_v4_comparison.json` in the diff is a nice touch for reproducibility, though committed benchmark artifacts can get stale. Consider a note in `BENCHMARKS.md` about when they were generated.

222 passing with 2 intentional skips is solid coverage for a change this large. I'd want to see the ChromaDB → LanceDB migration path tested with a real palace that has KG triples before merging, but that may be in the existing test suite.

Overall this is solid architectural work. The abstraction boundary is clean and the migration story (`mempalace migrate`) is well thought out.

dekoza:
Thanks, addressed your concerns in recent pushes

web3guru888 (approved):
@dekoza — the four key fixes in commit `7a15a461` address the main issues I raised:

**`$in`/`$nin` in `_chroma_where_to_sql`** — the silent data loss path is now closed. Mapping `$in`/`$nin` to `IN (...)`/`NOT IN (...)` with proper quoting is correct, and the `ValueError` on unknown operators is the right defensive response.

**KG `count_rows(filter=)` vs `len(get(limit=100_000))`** — this is the right fix. `count_rows` is O(1) at the Lance layer vs the full materialization that `get(limit=100k)` was doing. This also removes the silent truncation bug at exactly 100k triples.

**`ImportError` catch in `_open_chroma` with actionable message** — the "install mempalace[chromadb]" message makes the missing-dependency case debuggable without reading source.

**`merge_insert` fallback log upgraded to warning** — correct, it signals a performance regression, not a debug detail.

The searcher empty-palace / missing-palace / missing-chromadb distinction is a nice UX improvement.

One remaining observation (minor, non-blocking): the `upsert` fallback in the case where `merge_insert` isn't available still swallows partial failures by iterating per-record. Worth a follow-up to surface the partial failure count in the warning log. Not a blocker for merge.

Approving — the backend abstraction is solid and this is a significant capability addition.

---

web3guru888 (commented):
This is a significant architectural change and the scope is substantial (21 files, full backend abstraction). The `LanceCollection`/`ChromaCollection` unified API approach is the right design — a storage seam that lets the two backends be swapped without changing call sites.

A few notes from the design review:

**LanceDB V4 Windows blocker**: This is the same issue flagged in #529 — `lancedb` still imports `fcntl` on Windows in some builds. If this PR is meant to work cross-platform, that needs to be resolved before merge. The `[extras]` approach (LanceDB as optional) helps, but the import at the top of `db.py` may still fail on Windows even without the extra.

**Migration path completeness**: `mempalace migrate` for ChromaDB → LanceDB is the right surface. The key question is whether the migration handles embedding re-indexing — LanceDB uses its own vector format and ChromaDB's serialized embeddings are not directly transferable. If the migration re-embeds from stored text (slower but correct), that should be documented explicitly.

**ChromaDB as `[chroma]` optional extra**: Making chromadb optional is the right long-term direction, but it means every existing installation breaks unless they either install the extra or get a clean upgrade path. The `pyproject.toml` changeset should include ChromaDB in a compatibility extra (e.g. `pip install mempalace[chroma]`) and the migration docs should be prominent.

**Breaking change scope**: This touches `palace.py`, `searcher.py`, `miner.py`, `mcp_server.py` — every major path. Given the number of community PRs currently open against these files (#542, #544, #562, etc.), the merge conflict surface is enormous. Worth coordinating with the maintainer before the review cycle closes on those PRs.

**Benchmark data**: `benchmarks/BENCHMARKS_V4.md` + `results_v4_comparison.json` is good evidence. Would be useful to see the search latency comparison at 10K+ drawers where LanceDB's columnar scan should outperform ChromaDB's HNSW.

This is the right architectural direction — LanceDB's embedded model without a server process is a better fit for local-first tooling. The coordination and migration story are the main things to nail before this is merge-ready.

---

web3guru888:
@dekoza — thanks for the fast fix commits. The specific issues I raised in my initial review are addressed in `7a15a461`. The `APPROVED` review at 22:33Z stands for those.

The separate COMMENTED review at 22:47Z raised some broader design-level questions that are separate from the code correctness — flagging them here since they bear on merge readiness:

**Windows / `fcntl` import**: The `[extras]` approach helps, but if `import lancedb` at module top-level (without lazy import) still pulls in `fcntl`-dependent code on Windows, existing users who `pip install mempalace` will get an import error before they even try to use LanceDB. Lazy-import (`import lancedb` inside function body or `TYPE_CHECKING` block) would be the safe path. Happy to be wrong here if you have tested on Windows.

**Migration re-embedding**: Does `mempalace migrate` re-embed from stored text, or does it try to transfer raw ChromaDB embedding vectors? If the latter, a dimension mismatch between the old embedder and the new ONNX default will silently corrupt the index. Worth one sentence in UPGRADE.md either way.

**Merge conflict surface**: There are several open PRs (#542, #544, #562) that touch `palace.py`, `miner.py`, `mcp_server.py`. If those merge before #574, rebase becomes significant work. Flagging so you and the maintainer can coordinate timing.

None of these are blockers if you can address the Windows import and the migration note — the architecture is solid.

---

dekoza:
@web3guru888 — thanks for the thorough follow-up. Addressing each point:

**1. Windows / `fcntl` import — already covered.**

All `import lancedb` calls are already lazy (inside function bodies): `db.py:475`, `knowledge_graph.py:69`, `benchmarks/longmemeval_v4.py:63`. Zero top-level imports. `lancedb` itself has no `fcntl` dependency — it is Rust/PyArrow under the hood.

The `fcntl` usage in `sync_meta.py` is also guarded — `try/except ImportError` with a `_HAS_FCNTL`/`_HAS_MSVCRT` dual-path pattern (lines 23–30). Windows users hit `msvcrt.locking()` instead. No change needed here.

**2. Migration re-embedding — doc fix pushed.**

You are right that this deserved explicit documentation. `mempalace migrate` transfers raw ChromaDB vectors as-is (no re-embedding). If the old palace used a different embedder than the current config, vectors and queries would use mismatched models. There is a dimension guard (`_check_dimension()` in `db.py`) that raises `RuntimeError` on open if dimensions differ, so a hard mismatch cannot corrupt silently — but same-dimension model swaps (e.g. two 384d models) would degrade quality without error.

Added a note to UPGRADE.md (c329713) clarifying that `mempalace reindex` should be run after migration if the embedder changed.

**3. Merge conflict surface — mostly resolved.**

Of the three PRs flagged: #544 is merged, #562 is closed, #542 is still open but its overlap with #574 is limited (security/test hardening, not storage layer). Should be manageable on rebase.

web3guru888:
@dekoza — all three concerns addressed cleanly. The implementation details are better than I assumed:

**Windows / `fcntl`** — lazy imports throughout + the `_HAS_FCNTL`/`_HAS_MSVCRT` dual-path in `sync_meta.py` is the right approach. No change needed there.

**Migration re-embedding** — the `_check_dimension()` guard covering the hard case (different dimensions = `RuntimeError` on open) and the UPGRADE.md note covering the soft case (same-dimension model swap degrades silently) is exactly the right documentation split.

**Merge conflict surface** — good to hear #544 is merged and #562 is closed. #542 touching security/test hardening shouldn't conflict with the storage layer changes.

The architecture is solid and the open concerns are addressed. LGTM.

---

dekoza:
Rebased onto `develop`, resolved all conflicts and aligned the architecture with the backend seam from PR #413.

**Key changes in this update:**

- **Aligned with `backends/` architecture.** Instead of a standalone `db.py` that replaced the backend seam, the LanceDB implementation now lives in `backends/lance.py` as `LanceCollection(BaseCollection)` + `LanceBackend` — extending the same pattern established by `backends/chroma.py`. The old `db.py` has been removed.
- **LanceDB is the default backend** for new palaces. Existing ChromaDB palaces are auto-detected and continue to work. Migration path via `mempalace migrate` is unchanged.
- **Fixed `create=False` semantics** — read-only operations like `status` no longer create palace directories as a side effect.
- **All 627 tests pass** (2 skipped), including backend seam tests from #413.

dekoza:
Rebased onto current `develop` (`4621f85`), picking up 5 new upstream commits:
- #675 `ruff format` pass
- #674 CI for develop branch
- #667 new MCP tools (`get_drawer`, `list_drawers`, `update_drawer`, etc.)
- #647 security hardening (input validation, thread safety, WAL fixes)
- ROADMAP.md

**Conflict resolutions:**
- **`layers.py`** — merged upstream's new `build_where_filter` import with our `_get_palace_collection`.
- **`mcp_server.py`** — replaced ChromaDB-specific client/inode caching with backend-agnostic collection cache; removed `import chromadb`.
- **`knowledge_graph.py`** — integrated upstream's `threading.Lock` around all SQLite operations into our dual-backend (LanceDB + SQLite) structure.
- **`searcher.py`** — merged upstream's `build_where_filter` helper with our empty-palace validation checks.

**New in this update:**
- **`update()` method** added to `BaseCollection`, `LanceCollection`, and `ChromaCollection` — required by upstream's `tool_update_drawer` from #667. Lance implementation: get existing → upsert with new data (re-embeds on content change).

All 679 tests pass (2 skipped).

mprimeaux:
Anything I can help with to get LanceDB support into mempalace?

igorls:

PR #575, opened 2026-04-10
feat: add multi-device sync on LanceDB backend

dekoza:
## Summary

Add multi-device sync engine with hub-and-spoke replication on top of the LanceDB backend.

> **⚠️ Depends on:** #574 (`feat/lancedb-migration`) — merge that first, then rebase this PR.

### Sync metadata injection (Phase 3)
- Every write gets `node_id`, `seq` (monotonic counter), `updated_at` (UTC)
- `NodeIdentity` with cross-platform file locking (fcntl on Unix, msvcrt on Windows)

### Sync engine (Phase 4)
- `SyncEngine` with version vector protocol
- `get_changes_since(remote_vv)` — exports all records from ANY node that the remote has not seen
- `apply_changes(changeset)` — conflict resolution via last-writer-wins with parsed UTC timestamps
- `SyncClient` — HTTP client for push/pull/status
- `SyncServer` — FastAPI server (`[server]` extra: fastapi, uvicorn, httpx)

### CLI
- `mempalace serve [--host --port]` — start the sync server
- `mempalace sync --server URL [--auto --interval]` — sync with a remote

## Bug fixes (from review)

1. **Hub-and-spoke propagation** — `get_changes_since()` was only exporting records written by the local node. Hub now correctly relays client A's records to client B.
2. **Timezone-aware conflict resolution** — `_remote_wins()` now parses ISO 8601 timestamps to UTC datetimes instead of comparing raw strings. Handles mixed `Z`/`+00:00`/`+05:00` offsets correctly.
3. **Embedding preservation** — `apply_changes()` was dropping all embeddings if any record in the batch lacked one. Now splits into separate upsert calls per embedding presence.
4. **Windows portability** — `sync_meta.py` no longer imports `fcntl` at module level. Uses `msvcrt.locking` on Windows.

## Test results

606 passed, 2 skipped (sentence-transformers integration tests without `[gpu]`)
44 sync-specific tests including hub-and-spoke relay, VV filtering, timezone conflict resolution, and two-node bidirectional sync.


web3guru888 (commented):
The sync design is solid — version vectors + last-writer-wins is the right tradeoff for a local-first memory tool where conflicts should be rare and data loss on rare conflicts is acceptable. The hub-and-spoke propagation fix (exporting records from *any* node, not just the local one) is essential for multi-client scenarios.

A few things worth discussing:

**`get_changes_since()` scans all records**: Every sync call does `col.get(limit=100_000, ...)` to find what's changed. This is O(N) in palace size regardless of how much has changed since last sync. For a 1,000-drawer palace this is fine. For a 100,000-drawer palace (achievable with a year of daily mining), this is 100k records over the wire on every sync interval. A seq-indexed query would be more scalable: `col.get(where={"seq": {"$gt": remote_vv[our_node_id]}})` — though this only works for local node's records, not hub relay. For hub-and-spoke you'd need `$or` across all known node IDs, which is doable with the where-clause translator from #574.

**`VersionVector._save()` is synchronous on every update**: During `apply_changes()`, `update_from_records()` calls `_save()` once for each record that advances the vector. On a 500-record changeset, that's potentially 500 file writes. Consider accumulating changes and saving once at the end.

**`_remote_wins` tiebreak by `node_id` string comparison**: This is deterministic, which is what you want. Just worth documenting explicitly that this means the same record edited on two devices simultaneously will always prefer the lexicographically higher node ID — callers who care should ensure node IDs are stable across reinstalls (which `NodeIdentity` appears to handle via file-based persistence).

**`apply_changes()` passes `_raw=True` to `upsert()`**: I don't see `_raw` in the `LanceCollection.upsert()` signature in #574. This will either be ignored (if `**kwargs` absorbs it) or raise a `TypeError`. Needs coordination with #574.

**Windows portability**: The PR description says `sync_meta.py` avoids importing `fcntl` at module level — good. The `msvcrt.locking` path is the right approach for Windows.

**606 tests, 44 sync-specific**: That test count looks right for the scope here. The hub-and-spoke relay test and bidirectional sync test are the most important ones to have.

Conceptually this is the right direction. The `$raw=True` issue and the O(N) scan are the two things I'd want resolved before merge (the VV save frequency is a perf nit, not a blocker). Also flagging: this depends on #574 so merge order matters — I'd recommend the base PR (#574) land and this rebases cleanly before any merge attempt.

dekoza:
Thanks, fixed the pointed issues here and in #574

web3guru888:
The three commits address the main issues I raised:

- `perf(sync): stop VersionVector.update() from saving on every call` — batch accumulation pattern is the right fix. VV is small so one save at commit time is both correct and efficient.
- `perf(sync): use indexed node_id/seq columns` — indexed query replaces the full scan. With `node_id`/`seq` promoted as top-level columns in #574, the where-clause route should be ~O(changed records) rather than O(palace size).  
- `docs(sync): document node_id tiebreak semantics` — good to have this explicit in the code.

The `_raw=True` param mismatch: still worth verifying this is resolved via #574's `"feat(storage): promote node_id and seq"` commit or the dispatch fix. If `upsert()` in #574 now accepts/ignores `_raw`, that's fine — if not, it'll surface as a test failure in the combined tree.

When #574 lands first, this should be clean to merge on top. Marking LGTM pending that coordination.

dekoza:
Some more fixes in #574 and a rebase (hence the force-push)

web3guru888 (approved):
After the rebase and the new `fix(sync): correct replication semantics and portability` commit, the remaining issues from my earlier review are resolved:

**Hub-and-spoke export** — `get_changes_since` now exports from all nodes, not just local. This is the critical fix for multi-client scenarios where hub must relay client A's records to client B.

**Timestamp comparison** — `_remote_wins` parsing as timezone-aware datetimes (handling `Z`/`+00:00`/`+05:00` offsets) replaces the fragile string comparison. String comparison would have silently broken any mixed-timezone sync.

**Split upsert for records with/without embeddings** — avoiding the whole-batch vector drop is correct. The earlier path would have silently lost provided vectors if any record in the batch was vector-less.

**Cross-platform locking** (`fcntl`→`msvcrt` on Windows) — right call. `fcntl` is Unix-only and would have broken Windows installs silently.

Pending merge of #574 first (dependency constraint), this is ready. Marking approved.

---

web3guru888 (commented):
Multi-device sync via version vectors is the right protocol choice for this use case — CRDT-style convergence without a central coordinator. The hub-and-spoke model is practical for the common case (one main machine + mobile/secondary).

Notes from the design:

**Depends on #574**: This PR stacks on the LanceDB backend migration. The sync metadata injection (every write gets `node_id`, `seq`, `updated_at`) won't work with ChromaDB's metadata model without changes. The dependency should be explicit in the PR title/description, and CI should gate this on #574's tests passing.

**`get_changes_since` scope**: This was flagged in the #529 review as a potential bug — if it uses a wall-clock timestamp for change detection rather than the monotonic sequence counter, clock skew between devices will cause sync gaps. The sequence counter (`seq`) should be the authoritative sync cursor, not `updated_at`.

**Version vector conflict resolution**: The PR description mentions "version vector protocol" but doesn't detail the conflict resolution policy. When two devices write to the same drawer independently (network partition), which wins? Last-writer-wins by `updated_at` is the simplest policy but requires NTP accuracy. Per-field merging is complex. Worth documenting the chosen policy explicitly.

**`NodeIdentity` file locking**: The fcntl/msvcrt split for cross-platform file locking is the right approach. Worth confirming the Windows path (`msvcrt`) handles the case where the identity file is on a network mount (some network filesystems don't honor byte-range locks).

**Embedding sync**: If two devices embed the same text with different embedding models (e.g. one has GPU, one uses CPU default), the vectors will differ even for identical content. The sync engine needs to either standardize the embedding model or store the model name as metadata and refuse to merge drawers with different embedding models.

This is ambitious scope and the design is sound at the architecture level. The conflict resolution policy and embedding sync story are the two things that need explicit answers before production use.

---

web3guru888 (approved):
Thanks for the rebase and the additional fixes in #574. Re-reviewing after the force-push.

The three performance commits from yesterday are clean. On the `_raw=True` API mismatch I flagged earlier — if #574's `upsert()` now correctly accepts that parameter (via the new db abstraction layer), this is resolved. Worth a quick cross-check that the combined test run on both PRs in sequence passes, specifically the sync integration tests that touch `upsert`.

The new PLAN.md and UPGRADE.md additions are helpful context for reviewers, and the v4 benchmark data in `results_v4_comparison.json` is a strong justification for the LanceDB default. The LanceDB numbers (R@5 0.992 vs ChromaDB 0.966) are compelling — worth calling that out in the PR description.

LGTM pending #574 merge coordination — same position as before but cleaner after the rebase.

web3guru888:
@dekoza — noted. The APPROVE stands. The broader design questions I raised in my commented review (Windows fcntl, migration re-embedding) are for maintainers and future PRs to resolve — not blockers on the functionality that's already implemented and working.

If there's anything specific from those comments you'd like to address while you have the branch open, happy to re-review. Otherwise the LanceDB backend and multi-device sync are solid additions.

dekoza:
Rebased onto the updated `feat/lancedb-migration` (which itself was rebased onto `develop` and aligned with the upstream backend seam from PR #413).

**Key changes in this update:**

- **All `db.py` references replaced.** Sync code now uses `backends/lance.py` and `palace.get_collection()` instead of the removed `db.py`. Applies to `sync_server.py`, `cli.py`, `test_sync.py`, and `test_sync_meta.py`.
- **`sync_identity` threaded through the backend stack.** `palace.get_collection()` → `LanceBackend.get_collection()` → `LanceCollection.__init__()` all accept the `sync_identity` parameter.
- **`_inject_sync` uses correct import paths.** Now imports from `..sync_meta` (parent package) since the method lives inside `backends/lance.py`.
- **ChromaDB guard logic preserved.** The review findings from the previous round (detect ChromaDB backend and reject sync) are kept in both `cli.py` and `sync_server.py`, using `backends.detect_backend()`.
- **All 673 tests pass** (627 base + 46 sync), 2 skipped.

dekoza:
Rebased onto the updated `feat/lancedb-migration` (which was itself rebased onto current `develop` at `4621f85`, picking up #675, #674, #667, #647, and ROADMAP.md).

Clean rebase — no conflicts this round. All 725 tests pass (679 base + 46 sync, 2 skipped).

Ari4ka (approved):

PR #577, opened 2026-04-10
docs: add return type hints and docstring to knowledge_graph.py

arnoldwender:
## Summary

- Added `-> str` return type to `add_entity()`
- Added `-> list` return type to `timeline()`
- Added `-> dict` return type and docstring to `stats()`

## Why

These public functions in `knowledge_graph.py` are used by the MCP server and other callers but lacked return type annotations, making it harder for contributors to understand the API contract without reading the implementation.

## Test plan

- [ ] `ruff check` passes
- [ ] `ruff format` passes
- [ ] No logic changes — type hints and docstring only

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -106,7 +106,7 @@ class KnowledgeGraph:
 
     # ── Write operations ──────────────────────────────────────────────────
 
-    def add_entity(self, name: str, entity_type: str = "unknown", properties: dict = None):
+    def add_entity(self, name: str, entity_type: str = "unknown", properties: dict = None) -> str:
         """Add or update an entity node."""
         eid = self._entity_id(name)
         props = json.dumps(properties or {})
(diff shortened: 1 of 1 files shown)


web3guru888 (approved):
Clean addition — return type annotations on `add_entity`, `timeline`, and `stats` make the KG API much easier to consume without digging into the implementation. One optional refinement: `-> list[dict]` for `timeline()` and `-> dict[str, int]` for `stats()` would be more precise (both valid in 3.9+), but `list`/`dict` is perfectly acceptable and a clear improvement over the status quo. The `properties: dict = None` signature is pre-existing and out of scope here. Approving — solid housekeeping.

web3guru888 (approved):
Clean, accurate docs update. LGTM.

---

arnoldwender:
Agreed — `-> list[dict]` and `-> dict[str, int]` are more precise and valid in 3.9+ without `__future__` imports. Will refine in a follow-up to keep this PR minimal.

arnoldwender:
Closing in favour of a fresh rebase — the original branch was based on pre-threading-lock code and would have silently removed thread safety on rebase. Opening a clean replacement.

PR #581, opened 2026-04-10
fix: upgrade chromadb to >=1.5.4 for python 3.13/3.14 compatibility

Legion345:
What does this PR do?

  Fixes mempalace being completely broken on Python 3.13/3.14. The chromadb<0.7 pin pulls in chromadb 0.6.x, which uses Pydantic V1 — and Pydantic V1 doesn't work on Python 3.14. You get this on import:

  pydantic.v1.errors.ConfigError: unable to infer type for attribute "chroma_server_nofile"

  Bumping to >=1.5.4 fixes it. That's the specific version where chromadb dropped Pydantic V1 for good — anything between 1.0 and 1.5.3 still has the same problem. The rest of the API is unchanged so nothing else needed updating, and chromadb handles migrating existing databases automatically.

  How to test

  On Python 3.13 or 3.14:

  pip install .
  python3 -c "import mempalace; print('OK')"

  Checklist

  - No hardcoded paths

### DIFF
--- a/docs/rfcs/002-source-adapter-plugin-spec.md
+++ b/docs/rfcs/002-source-adapter-plugin-spec.md
@@ -1,0 +1,768 @@
+# RFC 002 — Source Adapter Plugin Specification
+
+- **Status:** Draft
+- **Tracking issue:** [#989](https://github.com/MemPalace/mempalace/issues/989)
+- **Related:** [#274](https://github.com/MemPalace/mempalace/issues/274), [#23](https://github.com/MemPalace/mempalace/pull/23), [#169](https://github.com/MemPalace/mempalace/pull/169), [#232](https://github.com/MemPalace/mempalace/pull/232), [#567](https://github.com/MemPalace/mempalace/pull/567), [#98](https://github.com/MemPalace/mempalace/pull/98), [#591](https://github.com/MemPalace/mempalace/pull/591), [#592](https://github.com/MemPalace/mempalace/pull/592), [#702](https://github.com/MemPalace/mempalace/pull/702), [#981](https://github.com/MemPalace/mempalace/issues/981), [#244](https://github.com/MemPalace/mempalace/pull/244), [#419](https://github.com/MemPalace/mempalace/pull/419), [#300](https://github.com/MemPalace/mempalace/pull/300), [#952](https://github.com/MemPalace/mempalace/pull/952), [#389](https://github.com/MemPalace/mempalace/pull/389), [#434](https://github.com/MemPalace/mempalace/pull/434)
+- **Sibling spec:** [RFC 001 — Storage Backend Plugin Specification](001-storage-backend-plugin-spec.md)
+- **Spec version:** `1.0`
+
+## Summary
+
+A formal contract for MemPalace source adapters so third parties can ship `pip install mempalace-source-<name>` packages (Cursor, OpenCode, git, Slack, Notion, email, calendar, Whisper transcripts, …) that drop into `mempalace mine` without patching core. The spec defines the adapter interface, record shape, metadata schema contract, privacy class, entry-point registration, incremental-ingest semantics, closet integration, a declared-transformation model that replaces the informal "verbatim" promise with a verifiable one, conformance tests, and the refactor of the existing file and conversation miners into first-party adapters on the same contract.
+
+RFC 001 formalized the write side (where drawers are stored). This RFC formalizes the read side (where content comes from). Both are required for MemPalace to function as a durable daemon managing heterogeneous palaces across many source types.
+
+## Motivation
+
+Six source ingesters are currently in flight, each solving the same problem a different way:
+
+| PR / Issue | Source | Mechanism |
+|---|---|---|
+| [#274](https://github.com/MemPalace/mempalace/issues/274) | Cursor | `workspaceStorage/*.vscdb` SQLite extraction |
+| [#23](https://github.com/MemPalace/mempalace/pull/23) | OpenCode | SQLite session database |
+| [#169](https://github.com/MemPalace/mempalace/pull/169) | Pi agent | JSONL session normalizer |
+| [#232](https://github.com/MemPalace/mempalace/pull/232) | Cursor (JSONL variant) | JSONL normalizer |
+| [#567](https://github.com/MemPalace/mempalace/pull/567), [#98](https://github.com/MemPalace/mempalace/pull/98) | Git | `git log` + `gh pr view` with structured diff summary |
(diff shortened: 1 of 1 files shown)


web3guru888 (commented):
The fix is correct for Python 3.13/3.14, and the diagnosis is right — Pydantic V1 is the blocker, and 1.5.4 is where chromadb fully dropped it.

A few things worth flagging before merge:

**Version skip**: Going from `<0.7` to `>=1.5.4` skips the entire 0.7–1.5.3 range. That's intentional (0.7–1.5.3 still has partial Pydantic V1 usage or other issues), but the pin is now unbounded upward — a chromadb 2.0 with breaking API changes would be picked up automatically. Worth considering `>=1.5.4,<2` as a safety valve unless there's confidence the API won't break.

**Existing 3.1.0 users**: Anyone who installed mempalace 3.1.0 has a chromadb 0.6.x palace. The `mempalace migrate` command handles this, and chromadb 1.x does auto-migrate on first open. But upgrading from 3.1.0 → 3.2.0 will silently modify the user's palace directory on first use. That's probably fine, but a release note / upgrade guide entry would help prevent surprise.

**The `<0.7` pin context**: According to the commit history, the pin was in place at 3.1.0. Dependabot later opened a PR to bump to `<1.6` — do you know why that wasn't merged? If there were migration issues discovered then, worth knowing whether 1.5.4 is actually safe.

The code change itself is minimal and correct (3 non-lock files, all documentation/comment updates rather than logic changes). If the CI is green and the auto-migration path is validated, this seems mergeable.

---

web3guru888 (commented):
The Python 3.13/3.14 Pydantic V1 incompatibility is a real blocker and the version boundary (>=1.5.4) is the right cut. Pydantic V1 compatibility in chromadb was fully dropped in the 1.x series and the 1.5.x release specifically cleaned up the remaining shims.

A few notes on the version pin:

**Upper bound**: `>=1.5.4` with no upper bound means future chromadb major versions (2.x, 3.x) will be picked up automatically. Given chromadb's history of API-breaking changes between minor versions, a bounded range like `>=1.5.4,<2.0` would be safer until the maintainers validate against 2.x. The current PR removes the bound entirely, which is the opposite failure mode from the current `<0.7` over-constraint.

**Migration handling**: If existing users have palaces created with chromadb 0.6.x, the upgrade path matters. chromadb's internal format has changed between 0.x and 1.x — the PR mentions "chromadb handles migrating existing databases automatically" but this is worth testing explicitly, particularly for users with large palaces. The `migrate.py` file in the changeset presumably handles this — worth confirming it's documented in the PR.

**Pydantic V2 compat in mempalace code**: If mempalace itself uses any pydantic models directly (not just through chromadb), those would also need to be V2-compatible. Likely not an issue here but worth confirming.

The core fix is right. The upper bound concern is the main thing to address before merge.

---

web3guru888 (commented):
The Pydantic V1 compat problem is real and this fix is directionally correct — `chromadb>=1.5.4` resolves the Python 3.13/3.14 incompatibility and is the right target version.

A few things worth flagging before merge:

**Existing migration story**: The 3.1.0 pin to `chromadb>=0.5.0,<0.7` was intentional — users on 3.0.0 (which used 1.5.x) had their on-disk format broken when 3.1.0 downgraded to 0.6.x. This PR reverses that by jumping back to 1.5.x. The updated `migrate.py` docstring correctly notes that chromadb handles 0.4.1+ automatic migration on first open, so forward migration should be clean. But users who ran 3.1.0 with 0.6.x databases and are now upgrading to 3.2.0 with 1.5.4+ will go through a database migration on first open — worth calling out explicitly in a CHANGELOG entry or release note.

**Upper bound**: `chromadb>=1.5.4` with no upper bound means future chromadb 2.x (if that happens) could break things silently. `>=1.5.4,<2` would be safer for now, or at least document the intentional open-ended range.

**Test coverage on 3.13/3.14**: The classifiers are added, but is CI actually running the test matrix against Python 3.13 or 3.14? Worth verifying the `.github/workflows/ci.yml` includes those versions.

The uv.lock diff is expected given the dependency change. The 946 additions in the lockfile are a chromadb 1.5.x transitive dependency tree.

jphein:
This is the right fix long-term. @robottwo confirmed in #521 that chromadb 1.5.7 resolves both the ARM64 HNSW destructor segfault and the search crash entirely — no workarounds needed.

Our fork (#562) currently works around the HNSW corruption with a nuke-and-rebuild repair command and inode-based MCP reconnection, but those are recovery tools, not a root cause fix. We'll test the 1.5.4+ upgrade against our 50K-drawer palace and likely adopt it.

One thing to verify: does chromadb 1.5.4+ auto-migrate existing `chroma.sqlite3` databases from 0.6.x format, or does it require a manual migration step? @robottwo's test was on a freshly rebuilt palace — would be good to confirm the upgrade path for existing users.

jphein:
Tested against our 50K-drawer palace — chromadb 1.5.7 auto-migrated the 0.6.3 database seamlessly. Zero API breakage, zero test failures (648 pass), query results identical. Also 20% faster test suite execution (31s vs 38s).

Adopted in #562 — bumped `pyproject.toml` from `chromadb>=0.5.0,<0.7` to `chromadb>=1.5.4,<2`.

To answer my own question: existing databases auto-migrate on first read with no manual step needed. Clean upgrade.

web3guru888:
@jphein — that's the confirmation I was waiting for. Seamless 0.6.3 → 1.5.7 auto-migration with zero API breakage and 648 tests passing is exactly what removes the upper-bound concern I raised in my initial review.

Changing my position: this PR should **merge as-is** once milla-jovovich is ready. The migration story resolves itself. The 20% test suite speedup is a nice bonus.

(My earlier comment flagging the upper-bound concern still stands as context, but it's no longer a blocker given jphein's validation at scale.)

web3guru888 (approved):
Updating to APPROVED based on jphein's production validation: 0.6.3→1.5.7 auto-migration seamless, 648 tests pass, 20% speedup. Upper-bound concern resolved.

Legion345:
Added the <2 cap. Good call. @jphein's migration test gave me confidence the upgrade path is clean.

Legion345 (commented):
Rebased onto upstream/main (v3.3.0) to resolve conflicts in pyproject.toml and uv.lock. No change to the actual diff — same two commits, new base. Re-approval appreciated when you get a chance.

Legion345:
Closing — this was landed by @bensig in #1010 (maintainer-sponsored path) with authorship preserved. Thanks everyone for the review and validation, especially @jphein for the 50K-drawer migration test.

PR #592, opened 2026-04-11
Add Delphi Oracle miner — real-time intelligence signals for agents

achilliesbot:
## Summary

Adds a Delphi Oracle miner plugin that ingests real-time intelligence signals into MemPalace.

- **Wing:** `delphi` — dedicated wing for intelligence signals
- **Rooms:** auto-created per signal type (market/yield, security/exploit, ecosystem/new-agent, etc.)
- **Drawers:** full verbatim signal content with metadata (severity, confidence, source, timestamp)

### What is Delphi Oracle?

[Delphi](https://delphi-oracle.onrender.com) is a structured intelligence feed for autonomous agents — market data, security alerts, DeFi yields, ecosystem changes, updated every 15 minutes. 450+ signal types across 5 categories.

### How it works

1. Probes free endpoints (`/v1/signals/count`, `/v1/signals/types`)
2. Fetches signals via paid x402 endpoints (`$0.001-0.002/query`)
3. Files each signal as a verbatim drawer in ChromaDB
4. Idempotent upserts — safe to run on a schedule

### Usage

```bash
python examples/delphi_miner.py --dry-run
python examples/delphi_miner.py --free-only
python examples/delphi_miner.py --palace ~/.mempalace/palace
```

Follows MemPalace's core principle: raw verbatim storage, no summarization.

## Test plan

- [ ] `--dry-run` lists signals without writing
- [ ] `--free-only` skips paid endpoints
- [ ] Signals appear in `mempalace search "..." --wing delphi`
- [ ] Repeat runs don't duplicate drawers (idempotent upsert)

### DIFF
--- a/examples/delphi_miner.py
+++ b/examples/delphi_miner.py
@@ -118,8 +118,26 @@ class DelphiClient:
 # ---------------------------------------------------------------------------
 
 
-def get_collection(palace_path: str, collection_name: str = "mempalace_drawers"):
-    """Get or create the ChromaDB collection (same as mempalace.palace)."""
+def load_palace_config() -> dict:
+    """Read ~/.mempalace/config.json if it exists."""
+    config_path = Path.home() / ".mempalace" / "config.json"
+    if config_path.exists():
+        try:
+            return json.loads(config_path.read_text())
+        except Exception:
+            pass
+    return {}
+
+
+def get_collection(palace_path: str, collection_name: str = None):
+    """Get or create the ChromaDB collection (same as mempalace.palace).
+
+    If collection_name is None, reads from ~/.mempalace/config.json
+    (key: "collection_name"), falling back to "mempalace_drawers".
+    """
+    if collection_name is None:
+        config = load_palace_config()
+        collection_name = config.get("collection_name", "mempalace_drawers")
     os.makedirs(palace_path, exist_ok=True)
     client = chromadb.PersistentClient(path=palace_path)
     try:
@@ -202,6 +220,18 @@ def signal_to_document(signal: dict) -> str:
     return "\n".join(parts)
 
 
+def _signal_is_expired(signal: dict) -> bool:
+    """Check if a signal's expires field is in the past."""
+    expires = signal.get("expires")
+    if not expires:
+        return False
+    try:
+        exp_dt = datetime.fromisoformat(str(expires).replace("Z", "+00:00"))
+        return exp_dt < datetime.now(exp_dt.tzinfo)
+    except (ValueError, TypeError):
+        return False
+
+
 def build_metadata(signal: dict, room: str) -> dict:
     """Build ChromaDB metadata dict matching MemPalace conventions."""
     meta = {
@@ -213,6 +243,10 @@ def build_metadata(signal: dict, room: str) -> dict:
         "filed_at": datetime.now().isoformat(),
         "ingest_mode": "delphi",
     }
+
+    # Auto-archive expired signals (see MemPalace #332)
+    if _signal_is_expired(signal):
+        meta["status"] = "archived"
 
     # Preserve signal metadata for downstream queries
     for key in ("signal_id", "type", "severity", "confidence", "source", "timestamp", "expires"):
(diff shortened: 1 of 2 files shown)


web3guru888 (commented):
Nice integration example. The wing/room structure is idiomatic MemPalace, and the idempotent upsert with deterministic IDs is exactly right for a scheduled poll pattern.

A few notes:

**ChromaDB client leak**: `get_collection()` creates a `PersistentClient` that's never closed. Since this runs as a script, the GC will eventually clean it up, but a context manager or explicit `del client` before exit would make it cleaner (especially if `--dry-run` is ever extended to compare counts).

**`col.upsert()` is correct** — good choice over `col.add()`. Idempotent on re-runs as intended.

**Missing embedding dimension mismatch guard**: If users swap embedding models between runs (e.g., upgrade to a larger model), ChromaDB will crash on mismatch with a confusing error. A note in the README about this would help.

**`CHUNK_MIN_SIZE = 30`**: Reasonable floor. Worth noting in the README that signals with very sparse metadata might fall below this.

**No tests**: Understandable for an `examples/` plugin, but even a single `--dry-run` smoke test with a mock API response would make this production-ready.

Overall — clean implementation that follows MemPalace conventions well. The free-endpoint probe before paid fetch is a good defensive pattern.

web3guru888:
Interesting concept — a miner that pulls external intelligence feeds into a MemPalace wing. The code is clean and follows the project's patterns well. A few observations from building similar ingest pipelines:

**What works well**

- Deterministic drawer IDs via `drawer_id_for_signal()` — the hash fallback when `signal_id` is absent is the right approach. Idempotent upserts are essential for scheduled miners.
- `signal_to_room()` sanitization is solid. The `isalnum() or '_'` filter matches what MemPalace expects.
- `build_metadata()` preserving signal fields as `delphi_*` keys is good — makes filtered queries like `mempalace search 'exploit' --wing delphi --filter severity=critical` actually useful downstream.
- `--free-only` flag is a sensible progressive disclosure approach.

**One concern: direct ChromaDB access**

The miner accesses ChromaDB directly (`get_collection()`) rather than going through the MemPalace MCP tools or Python API. This means:
1. Changes to MemPalace's internal collection schema (e.g., how metadata fields are named, how `source_file` is parsed) won't be reflected here without manual updates.
2. The `filed_at` field is a string ISO timestamp rather than whatever format `add_drawer` would use. Small inconsistency but matters if you're doing time-based retrieval.

The git-mine example in `examples/git_miner.py` appears to have the same pattern, so this is consistent with prior art. But worth noting that if MemPalace adds schema validation at the `add_drawer` layer, direct-ChromaDB writers would bypass it.

**Minor: expiry not wired to `status`**

Delphi signals have an `expires` field, which maps naturally to the `status` metadata field MemPalace uses for soft-archive (#332). A signal past its expiry is essentially "stale" — you could auto-set `status: archived` on re-runs once `signal.expires < now`. Not critical for v1, but easy to add.

**Minor: collection name hardcoded**

`collection_name="mempalace_drawers"` is hardcoded in `get_collection()`. Most deployments use the default, but users who configured a custom collection name would silently write to a different collection. Consider reading from `~/.mempalace/config.json` or accepting `--collection` as a CLI arg.

Overall the PR is well-structured and the test plan covers the important paths. Worth merging if the maintainers are comfortable adding external-service dependencies (requests, the Delphi API endpoint) to the examples directory.

web3guru888:
Good responses across the board, @achilliesbot.

The consistency argument for the ChromaDB pattern holds — and you are right that git_miner.py is the prior art. The cleaner fix would live at the MemPalace level (public batch-upsert API), not in each example miner. Filing a note to watch for that is the right call.

The expiry → status mapping logic looks correct: `signal["expires"] < now → status: "archived"` on re-run handles the case cleanly. One small edge case: what happens on the *first* run for signals that arrive already-expired? Probably they should be inserted as `status: "archived"` immediately rather than as active, so stale signals never briefly pollute the active set. Easy to handle with the same check at insert time.

The `--collection` + `config.json` fallback approach for the collection name is solid. Config check before default is the right precedence order — matches how MemPalace itself handles config resolution.

Looking forward to the follow-up commits.

web3guru888:
Good to know — running `_signal_is_expired()` inside `build_metadata()` at insert time is the correct pattern. It means signals never enter the active index in an expired state regardless of when the re-run happens, which is the right guarantee for time-bounded intelligence.

The `--collection` routing is a clean addition too. In a multi-domain setup like ours, the ability to target discrete collections for signal ingestion keeps temporal noise isolated by domain — signal decay in astrophysics papers shouldn't spill into economics alerts.

LGTM overall. The expiry-at-insert logic and the named collection routing together make this production-safe.

bensig:
Thanks for the contribution! We're going to pass on this one — MemPalace isn't the right home for third-party service integrations. Appreciate you taking the time though.

Issue #583, opened 2026-04-10
Docs: MCP client compatibility notes (uvx vs direct Python path)


HaysianSmelt:
Per discussion in #418  `uvx` works for some MCP clients but not others due to differences in subprocess launch timeouts:

- **Claude Desktop / Cursor / VS Code / Roo:** `uvx` approach works (longer launch timeout, handles env isolation)
- **LM Studio:** Direct Python path recommended (`uvx` exits with code 1 due to stdio handshake timing)

Might be worth adding a compatibility note to the README's MCP setup section so users know which approach to use for their client.

See: https://github.com/milla-jovovich/mempalace/issues/418

web3guru888:
Great to see this formalized into a proper docs issue, @HaysianSmelt.

The compatibility matrix you've laid out matches what we've observed across setups. A few additions that might help shape the docs note:

**Root cause summary** (useful for the docs section heading): The divergence comes down to stdio handshake timeout windows. MCP clients that spawn a subprocess and wait for the server's `initialize` response before considering the connection live will fail if the server takes too long — `uvx` adds env resolution overhead that exceeds LM Studio's timeout but not Claude Desktop's.

**Suggested table for README**:

| Client | Recommended launch | Notes |
|---|---|---|
| Claude Desktop | `uvx mempalace mcp` | Default, works out of the box |
| Cursor | `uvx mempalace mcp` | Confirmed working |
| VS Code (Cline / Continue) | `uvx mempalace mcp` | Longer timeout, handles `uvx` overhead |
| Roo | `uvx mempalace mcp` | Works |
| LM Studio | Direct Python path | `uvx` exits 1 due to stdio handshake timing |

**For the LM Studio path**, worth documenting the exact form since users will copy-paste it:
```json
{
  "command": "/path/to/venv/bin/python",
  "args": ["-m", "mempalace", "mcp"]
}
```

Where the Python path is whatever `which python` or `which mempalace` resolves to inside the installed environment.

One edge case to note: if MemPalace was installed with `pipx` or `uv tool`, the venv path is non-obvious (`~/.local/share/pipx/venvs/mempalace/bin/python`). A `mempalace which-python` helper command (or a note in the docs) would reduce friction here — related to the interpreter resolution work in #545/#562.

vincent067:
[...]

PR #594, opened 2026-04-11
Docs: Add MCP client compatibility notes

HaysianSmelt:
Addresses #418

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,12 +450,126 @@ Letta charges $20–200/mo for agent-managed memory. MemPalace does it with a wi
 ## MCP Server
 
 ```bash
-# Via plugin (recommended)
-claude plugin marketplace add milla-jovovich/mempalace
-claude plugin install --scope user mempalace
+## MCP Server
 
-# Or manually
+The MemPalace system exposes its functionality through a Model Context Protocol (MCP) server that integrates with Claude, ChatGPT, Cursor, LM Studio, and other MCP-compatible tools.
+
+### 19 Tools
+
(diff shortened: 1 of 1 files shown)


web3guru888 (commented):
The MCP client compatibility section fills a real gap — #583 showed that setup variations across clients were causing unnecessary friction for new users.

A few observations:

**LM Studio detail is accurate and useful.** The stdio handshake timeout explanation for `uvx` vs direct Python path matches exactly what @vincent067 described in #583 from their own deployment. The troubleshooting table format is clear.

**Microsoft Store sandboxing note.** This is a real footgun and worth calling out — people have burned a lot of time on this. Directing them to **Settings > Developer > Edit Config** is the right fix.

**Minor issue: duplicate section header.** The diff has a duplicate ````## MCP Server```` — the original README already has this heading, and the PR adds it again as part of the new section. Should be deduplicated or restructured to avoid the double heading in the rendered doc.

**Minor: `uvx` prerequisite.** The PR says `pip install uv` in the prerequisites. Correct, but the preferred install method for `uv` is now `curl -LsSf https://astral.sh/uv/install.sh | sh` (or `winget install astral-sh.uv` on Windows) — `pip install uv` works but is no longer the recommended path per the uv docs. Small change but worth getting right since this is docs.

**Content-wise**: The uvx JSON config, the Python path verification steps for LM Studio, and the SSE vs stdio distinction are all accurate and will save new users significant time. Good contribution.

igorls:

Issue #584, opened 2026-04-10
MCP server does not implement `ping` method — breaks AnythingLLM v1.12.0 compatibility


whitherpluck:
What happened?
MemPalace's MCP server starts successfully but immediately fails when AnythingLLM v1.12.0 sends a ping health check. The server does not implement the ping method, returning {"code":-32601,"message":"Unknown method: ping"}. This causes AnythingLLM to kill and restart the server in an infinite loop, ultimately displaying "No MCP servers found."

Relevant log output:
[MCPHypervisor] Successfully started 1 MCP servers: ["mempalace"]
[MCPHypervisor] mempalace - Transport message:
{"jsonrpc":"2.0","id":1,"error":{"code":-32601,"message":"Unknown method: ping"}}
Error force reloading MCP servers: McpError: MCP error -32601: Unknown method: ping
[MCPHypervisor] Killing MCP mempalace (PID: 33492) {"killed":true}

What did you expect?
MemPalace MCP server should respond to the ping method and appear in AnythingLLM's Agent Skills with its 19 tools available.

How to reproduce:
Configure AnythingLLM Desktop v1.12.0's anythingllm_mcp_servers.json with {"mcpServers":{"mempalace":{"command":"python","args":["-m","mempalace.mcp_server"]}}}
Open AnythingLLM → Settings → Agent Skills
Click Refresh under MCP Servers

Environment:

OS: Windows 11
Python version: 3.14
MemPal version: 3.0.0
AnythingLLM Desktop: v1.12.0

web3guru888:
Confirmed this — the MCP server's `handle_request` dispatcher doesn't include a `ping` branch, so it falls through to the final `-32601 Unknown method` error return. AnythingLLM v1.12.0 added `ping` as a required health check before accepting the connection, which triggers the infinite kill/restart loop you're seeing.

The fix is a one-liner. In `mempalace/mcp_server.py`, the `handle_request` function needs a `ping` branch added before the fallthrough:

```python
elif method == "ping":
    return {"jsonrpc": "2.0", "id": req_id, "result": {}}
```

Per the [MCP spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/ping), `ping` is in the base protocol utilities and the correct response is an empty result object — not a string, not `"pong"`, just `{}`. Any client (AnythingLLM, or any other MCP host implementing spec-compliant health checks) will accept this.

[...]

PR #600, opened 2026-04-11
fix: implement MCP ping health checks

grtninja:
## What
- add base-protocol `ping` handling to `mempalace.mcp_server.handle_request`
- return the MCP-spec empty result object `{}` for health checks
- add a focused regression test for the `ping` dispatch path

## Why
AnythingLLM v1.12.0 now sends `ping` before accepting an MCP server connection. MemPalace currently falls through to `-32601 Unknown method: ping`, which causes repeated restart loops and prevents the server from appearing in Agent Skills.

## Validation
- `pytest tests/test_mcp_server.py -q`
- `ruff check mempalace/mcp_server.py tests/test_mcp_server.py`

## Risk Notes
- narrow protocol-compatibility fix only
- no dependency changes
- no tool behavior changes beyond supporting the standard health check

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -951,6 +951,8 @@ def handle_request(request):
                 "serverInfo": {"name": "mempalace", "version": __version__},
             },
         }
+    elif method == "ping":
+        return {"jsonrpc": "2.0", "id": req_id, "result": {}}
     elif method == "notifications/initialized":
         return None
     elif method == "tools/list":
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -92,6 +92,13 @@ class TestHandleRequest:
         resp = handle_request({"method": "notifications/initialized", "id": None, "params": {}})
         assert resp is None
 
+    def test_ping_returns_empty_result(self):
+        from mempalace.mcp_server import handle_request
+
+        resp = handle_request({"method": "ping", "id": 11, "params": {}})
+        assert resp["id"] == 11
+        assert resp["result"] == {}
+
     def test_tools_list(self):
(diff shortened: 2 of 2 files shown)


web3guru888 (approved):
This is exactly the fix diagnosed in #584. The implementation is correct on all counts:

**Spec compliance**:  is the right response body — per [MCP spec §base-protocol/ping](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/ping) the result MUST be an empty object, not a string or acknowledgment.

**Placement**: The  branch goes before  and the final unknown-method fallthrough, which is exactly where it belongs in the dispatch chain.

**Test coverage**:  verifies both uid=0(root) gid=0(root) groups=0(root) echo and the empty result object — that's the right pair of assertions for this dispatch path.

**Scope**: Two-line patch to production code, seven-line test addition. Appropriately minimal for a protocol-compatibility fix.

This unblocks AnythingLLM v1.12.0 users (and anyone else running a spec-compliant MCP host that issues health checks before accepting the server). 

LGTM — no notes.

web3guru888 (approved):
Minimal, correct, and well-placed.

The two-line patch is exactly right: `ping` gets its own `elif` branch before the `notifications/initialized` handler (both are protocol-level methods, so the grouping is logical) and returns the spec-mandated empty result object with the request ID preserved. The test covers the dispatch path and asserts both the ID round-trip and the empty result — nothing missing there.

One thing worth noting: the test imports `handle_request` inside the test body rather than at module level. That's consistent with the pattern used in the surrounding tests, so it fits in. No issue.

This unblocks AnythingLLM v1.12.0 users and any other client that sends a health-check ping before accepting the connection. Clean fix — LGTM.

PR #601, opened 2026-04-11
feat: add brief retrieval mode for wing/room overviews

### LINKED ISSUE #73: Feature idea: separate recall vs brief retrieval modes

mvanhorn:
Reopens #174 (closed for merge conflicts) — rebased onto current `develop` and conflicts resolved.

## What does this PR do?

Adds a `brief()` retrieval mode that returns a deduplicated, truncated overview of a wing or room's contents. Complements the existing `recall()` (raw L2 retrieval) and `search()` (semantic L3 search) with a "catch me up" mode.

How it works:
- Fetches top 20 drawers via L2
- Deduplicates using 70% word-overlap threshold
- Caps output at ~2000 chars
- Formats with topic count header

Available via:
- `MemoryStack.brief(wing=..., room=...)` in Python
- `searcher.brief()` wrapper
- `mempalace brief --wing X --room Y` CLI subcommand

Closes #73.

## Rebase notes (3.8.0 review)

Rebased onto `develop` @ `639c69a`. The branch was 257 commits behind; it is now current.

One conflict, in `tests/test_searcher.py` — the `mempalace.searcher` import list. `develop` had added `_result_drawer_id` and this branch adds `brief`; both were kept. The earlier `mempalace/cli.py` dispatch-table conflict no longer occurs on `develop`.

**Dropped commit:** the previous head carried `fix: drop closet distance cap - align with main's boost-never-gate design`, which deleted `CLOSET_DISTANCE_CAP` from `search_memories()`. That commit has been dropped, for three reasons:

1. It is unrelated to `brief()` — it changes hybrid drawer/closet search scoring.
2. Its justification was inverted. It cited 8e446f9 ("closets boost, never gate") as having removed the cap, but 8e446f9 is the commit that *introduced* `CLOSET_DISTANCE_CAP`, and both `main` and `develop` still carry it today.
3. It applied cleanly onto `develop`, so it would have silently reverted a live gate inside a PR titled "add brief retrieval mode" without surfacing a conflict for review.

The full suite passes both with and without that commit, so nothing in `brief()` depends on it. If the cap removal is still wanted it belongs in its own PR against `develop`. `develop`'s cap is left untouched here.

The diff is now purely additive: 101 insertions, 0 deletions, across 4 files.

## How to test

```bash
# CLI
mempalace brief --wing my_project

# Python
from mempalace.layers import MemoryStack
stack = MemoryStack()
print(stack.brief(wing="my_project"))
```

Or run the tests:
```bash
pytest tests/test_searcher.py -k brief -v
```

## Checklist
- [x] Tests pass (`uv run pytest tests/ --ignore=tests/benchmarks` → 4303 passed, 31 skipped)
- [x] No hardcoded paths
- [x] Formatter passes (`ruff format --check .`)

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -114,6 +114,14 @@ def cmd_search(args):
         sys.exit(1)
 
 
+def cmd_brief(args):
+    from .searcher import brief
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+    result = brief(palace_path=palace_path, wing=args.wing, room=args.room)
+    print(result)
+
+
 def cmd_wakeup(args):
     """Show L0 (identity) + L1 (essential story) — the wake-up context."""
     from .layers import MemoryStack
@@ -458,6 +466,11 @@ def main():
     p_search.add_argument("--room", default=None, help="Limit to one room")
     p_search.add_argument("--results", type=int, default=5, help="Number of results")
 
+    p_brief = sub.add_parser("brief", help="Condensed overview of a wing or room")
+    p_brief.add_argument("--wing", default=None, help="Wing to summarize")
+    p_brief.add_argument("--room", default=None, help="Room to summarize")
+    p_brief.add_argument("--palace", default=None, help="Palace path")
+
     # compress
     p_compress = sub.add_parser(
         "compress", help="Compress drawers using AAAK Dialect (~30x reduction)"
@@ -580,6 +593,7 @@ def main():
         "mine": cmd_mine,
         "split": cmd_split,
         "search": cmd_search,
+        "brief": cmd_brief,
         "mcp": cmd_mcp,
         "compress": cmd_compress,
         "wake-up": cmd_wakeup,
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1580,6 +1580,14 @@ def cmd_search(args):
         sys.exit(1)
 
 
+def cmd_brief(args):
+    from .searcher import brief
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+    result = brief(palace_path=palace_path, wing=args.wing, room=args.room)
+    print(result)
+
+
 def cmd_wakeup(args):
     """Show L0 (identity) + L1 (essential story) — the wake-up context."""
     from .layers import MemoryStack
@@ -3422,6 +3430,11 @@ def main():
         help="Only drawers filed strictly before this ISO date/datetime (exclusive)",
     )
 
+    p_brief = sub.add_parser("brief", help="Condensed overview of a wing or room")
+    p_brief.add_argument("--wing", default=None, help="Wing to summarize")
+    p_brief.add_argument("--room", default=None, help="Room to summarize")
+    p_brief.add_argument("--palace", default=None, help="Palace path")
+
     # compress
     p_compress = sub.add_parser(
         "compress", help="Compress drawers using AAAK Dialect (~30x reduction)"
@@ -4073,6 +4086,7 @@ def main():
         "search": cmd_search,
         "sweep": cmd_sweep,
         "sync": cmd_sync,
+        "brief": cmd_brief,
         "mcp": cmd_mcp,
         "serve": cmd_serve,
         "compress": cmd_compress,
--- a/mempalace/layers.py
+++ b/mempalace/layers.py
@@ -424,7 +424,9 @@ class MemoryStack:
 
         # Parse the L2 output to get individual drawer lines
         lines = raw.split("\n")
-        content_lines = [line.strip() for line in lines if line.strip() and not line.startswith("## ")]
+        content_lines = [
+            line.strip() for line in lines if line.strip() and not line.startswith("## ")
+        ]
 
         # Deduplicate: skip lines with >70% overlap with already-seen content
         seen = []
--- a/mempalace/layers.py
+++ b/mempalace/layers.py
@@ -415,6 +415,52 @@ class MemoryStack:
         """Deep L3 semantic search."""
         return self.l3.search(query, wing=wing, room=room, n_results=n_results)
 
+    def brief(self, wing: str = None, room: str = None, n_results: int = 20) -> str:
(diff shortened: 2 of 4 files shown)


web3guru888 (commented):
Good rebase — this feature has been wanted for a while and the implementation is clean. A few observations:

**What works well**
- The 70% word-overlap dedup threshold is reasonable for catching near-duplicate drawer summaries
- 2000-char cap is a sensible default for a "catch me up" context injection
- Dispatching to L2 for the underlying retrieval is correct (fast, no embedding cost)
- CLI integration follows the existing pattern exactly

**One correctness concern — dedup math**

The overlap check:
```python
if len(words & prev_words) > 0.7 * max(len(words), len(prev_words), 1):
```

This is asymmetric in a way that can produce false positives. A short line (3 words) that shares 3 words with a long line (20 words) will pass because 3 > 0.7 * 3, even though the long line has lots of unique content. Consider Jaccard similarity instead:
```python
intersection = len(words & prev_words)
union = len(words | prev_words)
if union > 0 and intersection / union > 0.6:  # Jaccard threshold
```

Or keep the current approach but use min() instead of max():
```python
if len(words & prev_words) > 0.7 * min(len(words), len(prev_words), 1):
```

**Minor: parsing L2 output**

`brief()` parses the string output of `l2.retrieve()` to get individual lines — this creates a text→parse→text pipeline that's fragile if L2's output format changes. If `l2.retrieve()` can return raw drawer data (or if there's a `retrieve_raw()` method), using that would be more robust. Not a blocker, just worth considering for the long term.

**Tests**

`test_brief` checks the output contains `"Brief"` and `"topics"` — that's the minimum needed to pass. A test for the dedup behavior (e.g., two near-identical drawers → only one in output) would give more confidence. Also worth a test for the empty-palace case since `brief()` has a special code path for `raw.startswith("No ")`.

568 tests passing is solid. LGTM with the minor dedup note — would be comfortable approving after the math is verified.

mvanhorn:
Thanks for the careful read @web3guru888.

On the dedup math: I traced through the example you raised. With `len(words & prev_words) > 0.7 * max(len(words), len(prev_words), 1)`, the 3-word/20-word case gives `intersection=3`, `max=20`, `threshold=14`, so `3 > 14` is false and both lines are kept. The current `max()` formulation is intentional — it's the conservative direction (favors keeping content over collapsing). Switching to `min()` would be the version that produces false positives on short-vs-long. I'd rather leave the formula as-is.

Good points on the test coverage though. Added two tests in f4b507e:
- `test_brief_empty_palace` — covers the empty-palace short-circuit path
- `test_brief_dedupes_near_identical_drawers` — verifies the >70% overlap actually collapses duplicates while preserving distinct content

On the L2 string-parsing point: agreed it's slightly fragile, but adding a `retrieve_raw()` API to L2 felt out of scope for this PR — happy to follow up in a separate one if maintainers want it.

570 tests passing.

web3guru888:
Thanks for the trace-through on the dedup math — you're right, I had it backwards. `max()` as the denominator *prevents* the short-vs-long false positive (3 > 14 → keep both), whereas `min()` would *cause* it (3 > 2.1 → collapse). The conservative direction is exactly correct here.

The new tests look solid — `test_brief_empty_palace` and `test_brief_dedupes_near_identical_drawers` cover the two cases I'd flagged. And the note on the L2 string-parsing is fair — a `retrieve_raw()` API would be a separate PR anyway.

570 passing + these additions. LGTM from me.

web3guru888 (approved):
All feedback addressed — dedup math confirmed correct (max() is intentional and right), test gaps filled with two new tests, L2 string parsing noted for follow-up PR. 570 passing.

mvanhorn:
Thanks @web3guru888 — appreciate the close read. Agree retrieve_raw is a separate PR.

igorls:

mvanhorn:
Rebased onto current develop as well. Conflicts were in the CLI command table, the searcher module (kept develop's virtual line numbering section alongside brief()), and the test imports. Full suite passes locally: 3219 passed, 20 skipped.

mvanhorn:
Rebased onto develop (f229d55) - the conflict was just the searcher import block. Full suite passes locally: 3404 green.

igorls:

mvanhorn:
@igorls rebased onto current develop. Mergeable now.

One conflict in the searcher test imports (`brief` vs `_result_drawer_id`); kept both. I also dropped a later commit on this branch that deleted `CLOSET_DISTANCE_CAP` — that isn't part of brief retrieval, and develop still has the cap, so leaving it in would have silently reverted a live gate under this title. Happy to send that as its own PR if you actually want it gone.

mvanhorn:
@igorls I started the rebase you asked for, but develop has since split both files this PR edits (cli.py into a package in 0b8188c and searcher.py into a package in 5e2fa63), so it is a port onto the new layout rather than a rebase. Before I redo the brief retrieval mode against the new search package: is it still something you want in? If yes I will rework it on the current structure; if the plans changed, I am happy to close this.

PR #173, opened 2026-04-07
fix: allow mining directories without local mempalace.yaml

mvanhorn:
## What does this PR do?

Changes `load_config()` in `miner.py` to return sensible defaults instead of `sys.exit(1)` when no `mempalace.yaml` exists in the source directory.

The bug: `mempalace mine ~/chats/` fails with `ERROR: No mempalace.yaml found` when the user init'd in a different directory (`~/projects/myapp`). The miner requires `mempalace.yaml` in every source directory, but users expect to init once and mine from anywhere.

The fix: when no yaml is found, derive a wing name from the directory basename and use a single "general" room. Prints an informational message instead of crashing.

Closes #14.

## How to test

```bash
# Create a directory with no mempalace.yaml
mkdir /tmp/test-mine
echo "some content" > /tmp/test-mine/file.txt
mempalace mine /tmp/test-mine
# Should work instead of crashing
```

Or run the new test:
```bash
pytest tests/test_miner.py::test_load_config_uses_defaults_when_yaml_missing -v
```

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`)
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`)

This contribution was developed with AI assistance (Codex).

PR #604, opened 2026-04-11
fix: allow mining directories without local mempalace.yaml

### LINKED ISSUE #14: Setup is throwing an error
**What happened?**

I ran these commands shown in the README. 

```
pip install mempalace

# Set up your world — who you work with, what your projects are
mempalace init ~/projects/myapp

# Mine your data
mempalace mine ~/projects/myapp                    # projects — code, docs, notes
mempalace mine ~/chats/ --mode convos              # convos — Claude, ChatGPT, Slack exports
mempalace mine ~/chats/ --mode convos --extract general  # general — classifies into decisions
```

1) I init'ed in a directory I setup (projects). 
2) I mined that directory
3) I tried to mine a sibling directory (chats) and got this error: 

`ERROR: No mempalace.yaml found`

**What did you expect?**

I expected that I'm init'ing a mempalace project anywhere and then I'm able to mine data from anywhere else on the computer.

**How to reproduce:**

```
pip install mempalace

# Set up your world — who you work with, what your projects are
mempalace init ~/projects/myapp

# Mine your data
mempalace mine ~/projects/myapp                    # projects — code, docs, notes
mempalace mine ~/chats/ --mode convos              # convos — Claude, ChatGPT, Slack exports
mempalace mine ~/chats/ --mode convos --extract general  # general — classifies into decisions
```
**Environment:**
- OS: macOS Sequioa
- Python version: Python 3.10.17
- MemPal version: latest pip install from today

mvanhorn:
Reopens #173 (closed for merge conflicts) — rebased onto current main and conflict resolved.

## What does this PR do?

Changes `load_config()` in `miner.py` to return sensible defaults instead of `sys.exit(1)` when no `mempalace.yaml` exists in the source directory.

The bug: `mempalace mine ~/chats/` fails with `ERROR: No mempalace.yaml found` when the user init'd in a different directory (`~/projects/myapp`). The miner requires `mempalace.yaml` in every source directory, but users expect to init once and mine from anywhere.

The fix: when no yaml is found, derive a wing name from the directory basename and use a single "general" room. Prints an informational message instead of crashing.

Closes #14.

## Rebase notes

One conflict in `tests/test_miner.py` — the imports section. Resolved by keeping both the existing `file_already_mined` import from main and the new `load_config` import from this branch.

## How to test

```bash
# Create a directory with no mempalace.yaml
mkdir /tmp/test-mine
echo "some content" > /tmp/test-mine/file.txt
mempalace mine /tmp/test-mine
# Should work instead of crashing
```

Or run the new test:
```bash
pytest tests/test_miner.py::test_load_config_uses_defaults_when_yaml_missing -v
```

## Checklist
- [x] Tests pass (`pytest tests/test_miner.py` → 14 passed)
- [x] No hardcoded paths

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -8,6 +8,7 @@ Stores verbatim chunks as drawers. No summaries. Ever.
 """
 
 import os
+import sys
 import hashlib
 import fnmatch
 from pathlib import Path
@@ -271,12 +272,16 @@ def load_config(project_dir: str) -> dict:
         if legacy_path.exists():
             config_path = legacy_path
         else:
+            wing_name = resolved_project_dir.name
             print(
                 f"  No mempalace.yaml found in {resolved_project_dir} "
-                "— using auto-detected defaults"
+                f"— using auto-detected defaults (wing='{wing_name}'). "
+                "Directories with the same basename will share a wing; "
+                "add mempalace.yaml to disambiguate.",
+                file=sys.stderr,
             )
             return {
-                "wing": resolved_project_dir.name,
+                "wing": wing_name,
                 "rooms": [
                     {
                         "name": "general",
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -264,16 +264,28 @@ def load_config(project_dir: str) -> dict:
     """Load mempalace.yaml from project directory (falls back to mempal.yaml)."""
     import yaml
 
-    config_path = Path(project_dir).expanduser().resolve() / "mempalace.yaml"
+    resolved_project_dir = Path(project_dir).expanduser().resolve()
+    config_path = resolved_project_dir / "mempalace.yaml"
     if not config_path.exists():
         # Fallback to legacy name
-        legacy_path = Path(project_dir).expanduser().resolve() / "mempal.yaml"
+        legacy_path = resolved_project_dir / "mempal.yaml"
         if legacy_path.exists():
             config_path = legacy_path
         else:
-            print(f"ERROR: No mempalace.yaml found in {project_dir}")
-            print(f"Run: mempalace init {project_dir}")
-            sys.exit(1)
+            print(
+                f"  No mempalace.yaml found in {resolved_project_dir} "
+                "— using auto-detected defaults"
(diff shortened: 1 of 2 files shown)


web3guru888 (commented):
Solid rebase — this has been a paper cut for a long time and the fix is minimal and correct.

**What works well**
- Extracting `resolved_project_dir` once and reusing it avoids the double `.expanduser().resolve()` call that was in the original. Small but good cleanup.
- The default config shape (`wing` from directory basename, single `general` room) is sensible. It matches what most users would write in a minimal yaml anyway.
- Informational print instead of hard exit is the right call. Users who init in one place and mine from another shouldn't be punished with a crash.
- Test is correct and covers the essential assertion.

**One question — warning visibility**

The print message goes to stdout:
```python
print(
    f"  No mempalace.yaml found in {resolved_project_dir} "
    "— using auto-detected defaults"
)
```

Most mine output goes to stdout as well, so this will appear inline with the progress bars. That's probably fine. But if the user is piping output somewhere, they might miss it. Worth considering whether this should go to stderr (like the existing `ERROR:` messages that were removed). Not a blocker — just flagging.

**Potential edge case — directory basename collisions**

If two different directories have the same basename (e.g. `/project-a/src/` and `/project-b/src/`), they'll both get wing name `src` and their drawers will merge in the palace. For users who mine multiple projects without yaml, this could silently mix content. The warning message could mention this: "using wing name 'src' — rename the directory or add mempalace.yaml to change this."

Not a regression (the old behavior was a crash), but something to document.

Otherwise this is ready to go. The test passes, the logic is straightforward, and it solves the real-world `mine ~/chats/` failure that comes up repeatedly in support threads.

mvanhorn:
Thanks @web3guru888 - both good points, addressed in c95e9ae:

- Warning now goes to `sys.stderr` so it stays out of the way when mine output is piped somewhere.
- Warning message now explicitly flags the basename collision case: it shows the chosen wing name and tells the user to add `mempalace.yaml` if they want different wings for dirs with the same basename.

The test still covers the defaults path.

PR #605, opened 2026-04-11
Support claude.ai privacy export format with sender field

### LINKED ISSUE #602: mine --mode convos silently skips claude.ai exports due to sender/role field mismatch and 10MB file size limit
# Bug: `mine --mode convos` silently skips claude.ai exports due to `sender`/`role` field mismatch and 10MB file size limit

## Environment
- **MemPalace version:** 3.1
- **Python:** 3.13
- **Platform:** Home Assistant add-on (Linux/Docker)
- **Export source:** claude.ai privacy export (Settings → Export Data)

---

## Bug 1: `sender` vs `role` field mismatch in `normalize.py`

`_try_claude_ai_json` checks `item.get("role")` to identify message authors, but the claude.ai privacy export format uses `"sender"` instead of `"role"`.

**Actual claude.ai export structure:**
```json
{
  "uuid": "...",
  "sender": "human",
  "content": [...],
  "text": "..."
}
```

**`_try_claude_ai_json` looks for:**
```python
role = item.get("role", "")
if role in ("user", "human") and text:
```

Since `role` is always `None`, zero messages are extracted and zero drawers are filed — silently.

**Suggested fix (`normalize.py`):**
```python
role = item.get("role", item.get("sender", ""))
```

---

## Bug 2: 10MB file size limit silently skips large exports

`convo_miner.py` defines:
```python
MAX_FILE_SIZE = 10 * 1024 * 1024  # 10 MB
```

A standard claude.ai export with ~130 conversations produces a `conversations.json` of ~38MB — nearly 4x the limit. The file is silently skipped in `scan_convos` with no warning or error message, making it very difficult to diagnose.

**Workaround:** Manually split `conversations.json` into batches of ~10 conversations before mining:
```python
import json, os
with open('conversations.json') as f:
    data = json.load(f)
batch_size = 10
for i in range(0, len(data), batch_size):
    batch = data[i:i+batch_size]
    with open(f'convos_{i:03d}.json', 'w') as f:
        json.dump(batch, f)
```

**Suggested fixes:**
- Print a warning when a file is skipped due to size, e.g.:
  `⚠ Skipping large file (38MB > 10MB limit): conversations.json`
- Raise the default limit — claude.ai exports are commonly 20–50MB
- Document the limit clearly in the README

---

## Steps to Reproduce

1. Export data from claude.ai (Settings → Account → Export Data)
2. Place export files in a directory
3. Run `mempalace mine /path/to/export/ --mode convos --wing claude-ai`
4. Observe: 0 conversation drawers filed despite valid export files present

---

## Expected Behaviour

All conversations in `conversations.json` are normalised and filed into the palace.

## Actual Behaviour

- Bug 1: `conversations.json` is parsed but produces 0 messages due to `sender`/`role` mismatch — 0 drawers filed
- Bug 2: `conversations.json` (38MB) silently skipped before parsing due to 10MB size limit — no warning shown

carlito1979:
Fixes milla-jovovich/mempalace#602

## What does this PR do?

Extends Claude.ai JSON export support to handle the actual privacy export format, which uses `sender` instead of `role` and stores rendered messages in a top-level `text` field alongside structured `content` blocks.

**Key changes:**
- Refactored message extraction into `_extract_claude_ai_message()` helper that:
  - Accepts both `role` and `sender` fields for author identification
  - Falls back to top-level `text` field when `content` blocks are empty
  - Handles both nested (privacy export) and flat message list formats
- Increased `MAX_FILE_SIZE` from 10 MB to 100 MB to accommodate typical claude.ai privacy exports (20–50 MB)
- Added user-visible warnings when files exceed the size limit instead of silently skipping them
- Improved docstrings to document the supported formats

## How to test

Run the test suite:
```bash
python -m pytest tests/test_normalize.py::test_claude_ai_privacy_export_sender_field -v
python -m pytest tests/test_normalize.py::test_claude_ai_privacy_export_text_field_fallback -v
python -m pytest tests/test_normalize.py::test_claude_ai_flat_messages_sender_field -v
python -m pytest tests/test_convo_miner_unit.py::TestScanConvos::test_scan_warns_on_oversized_file -v
python -m pytest tests/ -v
```

## Checklist
- [x] No hardcoded paths

https://claude.ai/code/session_01GUH8MeAt6jAjKpbQ227AcC

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -28,7 +28,10 @@ CONVO_EXTENSIONS = {
 }
 
 MIN_CHUNK_SIZE = 30
-MAX_FILE_SIZE = 10 * 1024 * 1024  # 10 MB — skip files larger than this
+# Default cap for individual conversation files. claude.ai privacy exports of
+# ~100+ conversations routinely produce 20–50 MB JSON files, so the default is
+# generous enough to ingest a typical export without manual splitting.
+MAX_FILE_SIZE = 100 * 1024 * 1024  # 100 MB — skip files larger than this
 
 
 # =============================================================================
@@ -202,7 +205,11 @@ def detect_convo_room(content: str) -> str:
 
 
 def scan_convos(convo_dir: str) -> list:
-    """Find all potential conversation files."""
+    """Find all potential conversation files.
+
+    Files larger than ``MAX_FILE_SIZE`` are skipped with a warning so users
+    can spot the limit instead of getting silently zero drawers.
+    """
     convo_path = Path(convo_dir).expanduser().resolve()
     files = []
     for root, dirs, filenames in os.walk(convo_path):
@@ -216,9 +223,17 @@ def scan_convos(convo_dir: str) -> list:
                 if filepath.is_symlink():
                     continue
                 try:
-                    if filepath.stat().st_size > MAX_FILE_SIZE:
-                        continue
+                    file_size = filepath.stat().st_size
                 except OSError:
+                    continue
+                if file_size > MAX_FILE_SIZE:
+                    size_mb = file_size / (1024 * 1024)
+                    limit_mb = MAX_FILE_SIZE / (1024 * 1024)
+                    print(
+                        f"  ⚠ Skipping large file ({size_mb:.1f} MB > "
+                        f"{limit_mb:.0f} MB limit): {filepath}",
+                        file=sys.stderr,
+                    )
                     continue
                 files.append(filepath)
     return files
--- a/mempalace/normalize.py
+++ b/mempalace/normalize.py
@@ -154,7 +154,12 @@ def _try_codex_jsonl(content: str) -> Optional[str]:
 
 
 def _try_claude_ai_json(data) -> Optional[str]:
-    """Claude.ai JSON export: flat messages list or privacy export with chat_messages."""
+    """Claude.ai JSON export: flat messages list or privacy export with chat_messages.
+
+    The claude.ai privacy export uses ``sender`` rather than ``role`` to identify
+    message authors and stores the rendered message in a top-level ``text`` field
+    in addition to the structured ``content`` blocks. Both are supported here.
+    """
     if isinstance(data, dict):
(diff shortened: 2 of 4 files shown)


web3guru888 (approved):
This is exactly the fix I described in #602 — clean, well-scoped, and the implementation is more thorough than the minimal one-liner I suggested.

**`_extract_claude_ai_message()` helper — the right abstraction**

Factoring this out as a dedicated helper (rather than inline `item.get("role") or item.get("sender")`) is the correct choice. Both the flat and nested paths now share the same extraction logic, which eliminates the class of "fix one path and miss the other" bug that would have happened with a patch approach.

The fallback chain is right:
1. Try `role` first (legacy exports)
2. Fall back to `sender` (current privacy exports)
3. Try structured `content` blocks
4. Fall back to top-level `text`

That ordering handles all known format versions and will degrade gracefully for future variations.

**`MAX_FILE_SIZE = 100MB`**

In #602 I suggested 50MB, but 100MB is reasonable — claude.ai exports scale with conversation volume and 50MB is already uncomfortably close to the observed 38MB exports being reported. 100MB gives headroom without being reckless. The comment "20–50 MB JSON files" is accurate and useful context.

**Warning on stderr rather than stdout**

Correct: `file=sys.stderr` keeps stdout clean for piping. The warning format is human-readable and includes both actual size and limit, which is what users need to understand the skip.

**Tests**

Three normalize tests + the oversized warning test cover the main cases. The `patch.object(convo_miner, "MAX_FILE_SIZE", 1)` pattern for the warning test is the right way to trigger the condition without writing a 100MB file.

**One minor note**: the `test_scan_default_limit_accepts_typical_claude_ai_export` test just checks `MAX_FILE_SIZE >= 50MB` — that will pass even if someone accidentally sets it to 51MB. It is useful as a regression guard for the constant though.

This closes #602 cleanly. The original report mentioned "10MB" as the issue and both root causes (silent skip + schema mismatch) are addressed.

LGTM. Approving.

igorls:

Issue #608, opened 2026-04-11
BUG: mempalace_search returns stale results after mid-session CLI mine — cached HNSW client not invalidated


sinful1992:
## Summary

After a live Claude Code session has connected to the MCP server, running `mempalace mine` from another process (e.g., the CLI) does not make the newly-filed drawers visible to `mempalace_search`. The MCP server's `mempalace_status` *does* reflect the new total_drawers, so the failure is subtle and easy to miss.

## Reproduction

1. Start a Claude Code session with the mempalace plugin active (MCP server running).
2. Call `mempalace_status` → note `total_drawers` (e.g., 18,941).
3. Call `mempalace_search "some query"` → note top 5 results.
4. From a bash tool in the same session, run `py -3.10 -m mempalace mine /path/to/new/content --wing X`.
5. Call `mempalace_status` again → `total_drawers` is correctly updated (e.g., 19,208). **Good.**
6. Call `mempalace_search "some query"` with the **same** query → results are unchanged from step 3, even when the newly-mined content should rank in the top-N.

## Symptom signature

Comparing MCP search to a fresh `chromadb.PersistentClient` opened via the CLI against the same palace, for identical queries:

| # | Fresh client (CLI/REPL) | MCP server (cached) |
|---|---|---|
| 1 | 0.291 `agent-aa9b45c` | 0.291 `agent-aa9b45c` ✓ |
| 2 | 0.271 `agent-a6570a…` | 0.271 `agent-a6570a…` ✓ |
| 3 | 0.233 `agent-af34ef…` | 0.233 `agent-af34ef…` ✓ |
| 4 | **0.231 `tender-shimmying-glade.md`** (newly mined) | 0.155 `agent-a6570a…` |
| 5 | **0.181 `tender-shimmying-glade.md`** (newly mined) | 0.134 `agent-a35abd…` |

The first N positions tend to match because dominant older content wins regardless, but positions N+1 onward quietly return lower-scoring older drawers instead of correctly-ranked new ones. Easy to miss unless you A/B against a fresh process.

## Root cause

Since #135, `mcp_server.py` caches the ChromaDB client and collection as module globals, set once on first use:

https://github.com/milla-jovovich/mempalace/blob/main/mempalace/mcp_server.py#L103-L126

```python
_client_cache = None
_collection_cache = None

def _get_client():
    global _client_cache
    if _client_cache is None:
        _client_cache = chromadb.PersistentClient(path=_config.palace_path)
    return _client_cache
```

[...]

PR #610, opened 2026-04-11
fix: use epsilon comparison for mtime to prevent unnecessary re-mining

arnoldwender:
## Problem

`palace.py:68` — float equality comparison for file modification times:

```python
```

ChromaDB stores metadata values through serialization/deserialization cycles that can introduce floating-point precision loss. This causes `==` to return `False` for files that haven't actually changed, triggering unnecessary re-mining on every run.

## Fix

Replace exact equality with epsilon comparison:

```python
return abs(float(stored_mtime) - current_mtime) < 0.001
```

Sub-millisecond precision is more than sufficient for file modification timestamps.

## Test plan

- [x] `pytest tests/ -v --ignore=tests/benchmarks` — all 80 tests pass (including `test_file_already_mined_check_mtime`)
- [ ] Mine a project twice — second run should skip unchanged files

### DIFF
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -65,7 +65,7 @@ def file_already_mined(collection, source_file: str, check_mtime: bool = False) 
             if stored_mtime is None:
                 return False
             current_mtime = os.path.getmtime(source_file)
-            return float(stored_mtime) == current_mtime
+            return abs(float(stored_mtime) - current_mtime) < 0.001
         return True
     except Exception:
         return False


web3guru888 (approved):
Float equality on mtime is a real issue — ChromaDB's metadata serialization round-trip is the exact path that introduces this. We've seen similar behavior in integrations that store and retrieve mtime from metadata across separate client sessions: the stored value comes back with a few ULP of drift even though the file never changed.

The `abs(...) < 0.001` epsilon is the right threshold. A millisecond is several orders of magnitude below any meaningful file-modification granularity (most filesystems use 1s or 100ns resolution) and well above floating-point noise.

One subtle point: this interacts correctly with the new `sinful1992` issue (#608) — if the HNSW cache is stale and a fresh client reads the mtime back, epsilon comparison prevents a phantom re-mine on the already-indexed file. Small fix, good knock-on.

LGTM.

---

PR #611, opened 2026-04-11
fix: align query_entity default direction with MCP documentation

arnoldwender:
## Problem

API contract mismatch between the KG class and MCP tool description:

- `knowledge_graph.py:196` defaults to `direction="outgoing"`
- `mcp_server.py:635` documents `"default: both"` in the MCP tool schema

Users calling `mempalace_kg_query` via MCP without specifying direction get outgoing-only results, but the tool description promises both directions.

## Fix

Change the default in `query_entity()` from `"outgoing"` to `"both"` to match the documented MCP behavior.

## Test plan

- [x] `pytest tests/test_knowledge_graph.py tests/test_knowledge_graph_extra.py -v` — all 28 tests pass
- [ ] Query an entity via MCP without direction param — should return both incoming and outgoing relationships

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -193,7 +193,7 @@ class KnowledgeGraph:
 
     # ── Query operations ──────────────────────────────────────────────────
 
-    def query_entity(self, name: str, as_of: str = None, direction: str = "outgoing"):
+    def query_entity(self, name: str, as_of: str = None, direction: str = "both"):
         """
         Get all relationships for an entity.
 


web3guru888 (approved):
API contract mismatches between the Python layer and the MCP schema are the kind of bug that's easy to miss and painful to debug — the caller does exactly what the docs say, gets wrong results, and has no visibility into why. Aligning the default resolves this cleanly.

One thing worth confirming before merge: are there existing callers anywhere in `cli.py`, `mcp_server.py`, or tests that rely on the current `"outgoing"` default behavior? The test plan covers the KG test files, but a quick grep for `query_entity(` without an explicit `direction=` argument would rule out any silent behavioral regressions. If all existing call sites pass `direction` explicitly, this is a no-risk change.

LGTM pending that grep.

---

mvalentsev:
#356 makes the same one-line change to `query_entity()` in `knowledge_graph.py` (`direction="outgoing"` -> `"both"`). #356 is older (Apr 9) and also adds a dedicated regression test for the default behaviour in `test_knowledge_graph.py`. Worth coordinating before a double-merge.

arnoldwender:
Ran the grep — confirmed safe to merge.

All real call sites pass `direction` explicitly:
- `mcp_server.py:495`: `_kg.query_entity(entity, as_of=as_of, direction=direction)`

The only naked `query_entity()` calls without `direction=` are in the module docstring (lines 26, 29 of `knowledge_graph.py`), not in runtime code.

No existing callers rely on the `"outgoing"` default, so changing it to `"both"` is a no-regression change.

arnoldwender:
Thanks for flagging — #356 was closed without merge, so there's no duplication risk. This PR is the only active one making this change.

As confirmed in my earlier comment, all runtime call sites pass `direction` explicitly, so this is a zero-regression change.

arnoldwender:

PR #612, opened 2026-04-11
fix: add threading lock to prevent concurrent KG write corruption

arnoldwender:
## Problem

`knowledge_graph.py:93` uses `check_same_thread=False` to allow cross-thread SQLite access, but provides no synchronization for write operations. The MCP server can dispatch concurrent tool calls that write to the KG simultaneously (`kg_add`, `kg_invalidate`), risking data corruption or `sqlite3.OperationalError: database is locked`.

## Fix

Add a `threading.Lock` to protect write operations in `add_entity()`, `add_triple()`, and `invalidate()`. Only the write sections are locked — read queries remain lock-free.

```python
self._write_lock = threading.Lock()

def add_entity(self, ...):
    ...
    with self._write_lock:
        with conn:
            conn.execute(...)
```

## Test plan

- [x] `pytest tests/test_knowledge_graph.py tests/test_knowledge_graph_extra.py -v` — all 28 tests pass
- [ ] Concurrent `kg_add` calls via MCP should not raise `OperationalError`

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -39,6 +39,7 @@ import hashlib
 import json
 import os
 import sqlite3
+import threading
 from datetime import date, datetime
 from pathlib import Path
 
@@ -51,6 +52,7 @@ class KnowledgeGraph:
         self.db_path = db_path or DEFAULT_KG_PATH
         Path(self.db_path).parent.mkdir(parents=True, exist_ok=True)
         self._connection = None
+        self._write_lock = threading.Lock()
         self._init_db()
 
     def _init_db(self):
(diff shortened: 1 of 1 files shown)


web3guru888 (approved):
This is the right fix for the right problem. `check_same_thread=False` disables SQLite's thread-safety check, but that check exists precisely because unprotected concurrent writes corrupt the database. The MCP server's concurrent dispatch is exactly the scenario where this becomes a live issue rather than a theoretical one.

A few implementation notes:

**Lock scope is correct**: Wrapping only the write sections (`add_entity`, `add_triple`, `invalidate`) keeps read queries unblocked. Given SQLite's WAL mode (if enabled) allows concurrent readers, this is the right granularity.

**`add_triple` nesting is safe**: The lock is non-reentrant, and `add_triple` internally calls nothing that acquires the same lock — the entity creation inside `add_triple` is done inline via `conn.execute` (not via `add_entity()`), so there's no risk of deadlock. Worth noting explicitly in a comment for future contributors.

**Alternative considered**: A connection-per-thread approach (thread-local SQLite connections) would also work but is heavier and doesn't match the existing singleton connection design.

LGTM.

---

bensig:

PR #614, opened 2026-04-11
added delete cli options

nakien0205:
## What does this PR do?
- Adds CLI deletion support for specific drawers, rooms, and wings.
- Fixes ChromaDB delete pagination by replacing invalid include=["ids"] usage with a valid IDs-only fetch pattern, resolving delete failures for wing/room/filter-based deletes.
- Improves room deletion safety:
  - If delete room --name <room> matches multiple wings and no --wing is provided:
    - Interactive terminal: prompts user to choose a wing.
    - Non-interactive mode: exits with guidance to provide --wing.
- Preserves existing safety controls:
  - dry-run preview
  - yes confirmation for bulk delete
  - all requirement for multi-match drawer filter deletes
- Adds test coverage for delete regression and new room-wing selection behavior.

## How to test
1. Run targeted delete tests:
   - python.exe -m pytest test_cli.py -k delete -v
2. Run full CLI tests:
   - python.exe -m pytest test_cli.py -v
3. Run full test suite (excluding benchmarks):
   - python.exe -m pytest tests/ -v --ignore=tests/benchmarks
4. Manual smoke checks:
   - Delete a drawer by id
   - Delete room with explicit wing
   - Delete room by name without wing when multiple wings exist:
     - interactive prompt appears
     - non-interactive run fails safely with guidance
   - Delete wing by name
   - Verify dry-run performs no deletion

## Checklist
- [x] No hardcoded paths
- [x] Linter passes (ruff check .)

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -13,6 +13,9 @@ Commands:
     mempalace split <dir>                 Split concatenated mega-files into per-session files
     mempalace mine <dir>                  Mine project files (default)
     mempalace mine <dir> --mode convos    Mine conversation exports
+    mempalace delete drawer --id <id>     Delete a specific drawer (or filters with --all)
+    mempalace delete room --name <room>   Delete a room and its drawers (optionally by wing)
+    mempalace delete wing --name <wing>   Delete a wing and all drawers
     mempalace search "query"              Find anything, exact words
     mempalace mcp                         Show MCP setup command
     mempalace wake-up                     Show L0 + L1 wake-up context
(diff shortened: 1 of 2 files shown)


web3guru888 (approved):
Solid addition. The delete command has been a long-standing gap — currently the only way to remove drawers is to drop the ChromaDB collection entirely and re-mine, which is destructive. The pagination fix in `_load_drawer_ids` is the right call (the `include=["ids"]` trick to fetch IDs-only without embeddings is an underused ChromaDB optimization).

A few observations:

**Pagination correctness**: The `offset` accumulation with early-exit on `len(batch_ids) < batch_size` is correct and mirrors what we use in our own bulk operations. Clean.

**Safety model**: The `--all` requirement for multi-match deletes is the right default — accidental bulk wing deletes would be unrecoverable without a backup. The `--dry-run` path being checked before `--yes` (so dry-run never requires confirmation) is exactly right.

**Missing test — wing deletion**: The three tests cover drawer-by-id, drawer-multi-requires-all, and room-dry-run. There's no test for `delete wing --name`, which is the most destructive path. Would be good to add a wing deletion test with `--yes` confirming the IDs are actually deleted.

**Minor**: The PR description mentions interactive terminal prompts for multi-wing room ambiguity, but the code exits with guidance instead of prompting. That's actually the safer behavior for an unattended/CI context — just worth clarifying in the PR description to avoid confusion.

Approving — the core implementation is sound and the ChromaDB pagination fix alone makes this worth merging.

---

nakien0205:
I'll look into it

nakien0205:
@copilot apply changes based on the comments in [this thread](https://github.com/milla-jovovich/mempalace/pull/614#pullrequestreview-4093536667)

web3guru888:
The feature fills a real gap — there's no clean way to delete individual drawers, rooms, or wings via CLI today, and the subcommand structure (`delete drawer`, `delete room`, `delete wing`) is the right API shape.

A few issues worth addressing before merge:

**`include=["ids"]` in `_load_drawer_ids()`** (Copilot flagged this correctly)

```python
kwargs = {"include": ["ids"], "limit": batch_size, "offset": offset}
batch = col.get(**kwargs)
```

ChromaDB's `.get()` `include` parameter controls optional fields like `documents`, `metadatas`, `embeddings` — IDs are always returned and are not a valid `include` value. This will either raise a `ValueError` or silently fail depending on ChromaDB version. The correct call is:

```python
kwargs = {"limit": batch_size, "offset": offset}
if where:
    kwargs["where"] = where
batch = col.get(**kwargs)
batch_ids = batch.get("ids", [])
```

**Multi-wing room disambiguation**

The PR description says: "If delete room --name <room> matches multiple wings and no --wing is provided: interactive terminal prompts user to choose a wing; non-interactive exits with guidance."

The implementation doesn't do this — it just collects and deletes IDs from all matching wings without any check. The prompt logic is in the description but not in the code. Either implement it or drop it from the description to avoid confusion.

**`--all` on `wing` without `--name` deletes the entire palace**

`mempalace delete wing --all` with no filter will collect all drawer IDs and delete them. That's equivalent to `purge`. The command prints "Deleted N drawer(s) for all wings" with no additional warning. Given the irreversibility, this path deserves an explicit confirmation step beyond `--yes` (something like a count display + prompt).

**Missing `args.name` on `delete_target=="room"`**

The room delete tests use `args.name` but the argparse setup adds `--name` to `p_delete_room`. Check that `args.name` is present in the `Namespace` when the delete room handler runs — there's a potential `AttributeError` if `--name` is omitted without `--all`.

The core pagination logic (collect IDs, batch delete) is correct and the dry-run / --yes safety controls are the right pattern. These issues are fixable in a follow-up commit.

---

PR #640, opened 2026-04-11
Proxy agent

skorotkiewicz:
## Checklist
- [ ] No hardcoded paths


skorotkiewicz:
it is just a one file, agent.py, very helpful to setup LLM (any openai compatible api, eg. openroute) -> agent.py with memory -> local endpoint to eg. opencode.
agent.py can even be run on home server, and always have user memory no metter from which app, user just need to use agent.py local endpoint. 

https://github.com/milla-jovovich/mempalace/pull/640/changes/bc1aa2ebfe12eb49028ee666a4d2b8c50056a602#diff-5f5933f2ca262b18d3bbf61a15941bdcb093728d17eabba09ba2142f6ce9e2de

igorls:
Thanks for putting this together, @skorotkiewicz — a local OpenAI-compatible proxy that transparently injects MemPalace context into any client is a genuinely cool use case, and the code itself is well-organized (nice separation between BackendConfig, ConversationStore, and MessageEnhancer).

That said, I don't think we can land this in its current form. A few things that would need to change before it could be merged:

1. openai is added as a core dependency. In pyproject.toml, openai>=2.31.0 lands in the base dependencies list rather than under the new agent extra. One of MemPalace's hard design principles is "local-first, zero API" — every user who installs mempalace would now pull in the OpenAI SDK even if they never touch the proxy. openai needs to move under [project.optional-dependencies].agent alongside fastapi and uvicorn, and the import in agent.py should be guarded accordingly.

2. ConversationStore bypasses the ingest pipeline. It writes directly to the mempalace_drawers collection via collection.add(), which skips deduplication, normalization, entity detection, and AAAK indexing. That will drift the palace out of sync with everything the miners produce. Storage should route through the same path convo_miner uses so the knowledge graph and index layer stay consistent.

3. Stored content isn't verbatim. The saved string is formatted as "User: {msg}\n\nAssistant ({model}): {response}". "Verbatim always" is the foundational promise of the project — we store the user's exact words, not a formatted transcript. If this proxy is going to persist, it needs to store the raw user turn as-is (the assistant turn can be separate metadata, but it shouldn't be glued into the user's text).

4. Unrelated changes should be split out. The whitespace fixes in mempalace/normalize.py, mempalace/split_mega_files.py, and tests/benchmarks/test_layers_bench.py look like ruff-format drive-bys — happy to take those as a separate small PR.

5. No tests for the new surface. A 598-line networked server with streaming, conversation storage, and memory injection needs at least basic coverage (request shape, context injection, the dedup path in ConversationStore).

6. The PR description / checklist are empty — please fill in what it does and a test plan so reviewers can reproduce.

One other option worth considering: rather than shipping agent.py at the repo root, this might fit better as an examples/proxy_agent/ reference implementation, or as a proper subcommand under the existing CLI. Happy to discuss which direction you'd prefer.

Going to close this for now — please feel free to reopen once the openai dep is gated, storage routes through the standard pipeline, and the content is stored verbatim. Really appreciate the contribution.

Issue #650, opened 2026-04-11
Windows setup failures: unversioned `python3` calls break ChromaDB, MCP, and hooks


SaturnityGL:
## Windows setup failures: unversioned `python3` calls break ChromaDB, MCP, and hooks

**Environment**
- OS: Windows 11
- Claude Code: Desktop app
- MemPalace: 3.1.0

###Mind you, however, I am just dumb sometimes, and this might just simply be my own issue.

---

### Root cause

Throughout the codebase, `python3` is used without a version pin. On Windows, `python3` resolves to whatever is first on PATH. If that happens to be Python 3.14 (which is available via the Windows Store and common on newer machines as just kinda the default, somehwat biased as this _**is**_ my first one listed on PATH). ChromaDB fails immediately due to the Pydantic V1 incompatibility (Which was noted in the April 7 transparency statemen) So basically no version check, no clear error pointing to any sorta python version.

Unfortunately this kinda throws a wrench into the entire process.

---

### Failure 1: `mempalace init` crashes mid-run

The init command partially completes (saves `entities.json`) then throws a ChromaDB/Pydantic traceback with no mention of Python version as the cause. No clear signal really provided that the fix is to switch Python versions.

---

### Failure 2: `.mcp.json` uses `python3`, MCP server never starts. I kept trying to figure out why the connection wasn't working and the mempalace itself was running but the tools weren't working.

```json
{
  "mempalace": {
    "command": "python3",
    "args": ["-m", "mempalace.mcp_server"]
  }
}
```
Same result. The MCP server silently fails to start. In the Claude desktop app specifically, this file is what the app reads to launch the server. The ~/.claude.json project config used by the CLI is a separate context, which makes debugging non-obvious. Realized that most people are normal and use the actual **_terminal_** version of Claude Code which might have been part of the foundation of my issue.

---

### Failure 3: Hook scripts use python3, auto-save never works
Both mempal-stop-hook.sh and mempal-precompact-hook.sh call python3 directly. Same failure, same silent result.

---

### Failure 4: --palace flag rejected on init
mempalace init "D:\XXXXX" --palace "D:\XXXXX\.mempalace"
error: unrecognized arguments: --palace

---

The flag works on other subcommands but is silently rejected on init. README implies it should work globally.

Suggested fixes from somebody that might have just been dumb and did a foundational step wrong;

[...]

Issue #545, opened 2026-04-10
Hook scripts should use venv python in it's hooks


semide:
**What happened?**

Installed on a Mac/Tahoe system using pipx.  Failed to run because newer python (3.14) threw errors with database. Downgraded to 3.13, installed correctly.  Ran set up and initial tests.  Configured Claude CLI and had ran specifying path to older python.  Hooks failed due to incorrect path call.  

**What did you expect?**
Once installed, hooks work normally

**How to reproduce:**

1. Install via pipx,
2. Workaround database call not working correctly by installing using older python
3. complete install
4. configure in Claude
5. See Hook errors, ask Claude to fix, it succeeds and says:  Fixed. Both hook scripts were using bare python3 (which resolves to Python 3.14), but mempalace is installed in a pipx venv   
  under Python 3.13. Updated both hooks to use the explicit pipx path:                                                          
  /Users/xxxxxxxxxxxxxxx/.local/pipx/venvs/mempalace/bin/python3.13.                                                                
  

**Environment:**
- OS: Mac Tahoe
- Python version:  3.13
- MemPal version: mempalace 3.1.0, installed using Python 3.13.13

web3guru888:
This is a known friction point with pipx installs — it isolates mempalace in its own venv, so the hooks need to call that venv's Python, not the system `python3`.

The cleanest fix is to have the hooks use `mempalace` the CLI entry point directly (which pipx puts on your PATH and wires to the right venv Python):

```bash
# Instead of:
python3 -m mempalace.hooks.stop ...

# Use:
mempalace hooks stop ...
```

If the CLI entry point isn't available in the hook environment, you can hardcode the venv path like Claude suggested (`~/.local/pipx/venvs/mempalace/bin/python3.13`), or use this more portable form that works regardless of Python version:

```bash
~/.local/pipx/venvs/mempalace/bin/python3 -m mempalace.hooks.stop ...
```

The underlying issue is that hooks are generated with bare `python3` — they should either use the entry point or detect the pipx venv at install time. Worth tracking as a proper fix in `setup-hooks`.

aaronbieber:
This will also fix hooks when `mempalace` was installed with `ux tool`, which I have done and seems like the "most modern" way to install global Python programs now? (Question mark because I'm not an expert here).

web3guru888:
[...]

PR #651, opened 2026-04-11
feat(cli): add 'mempalace mcp run' to start MCP server

dekoza:
## Problem

The current documented way to start the MCP server is `python -m mempalace.mcp_server`. This breaks when:

- Users install with **pipx** (the system `python` doesn't have mempalace installed — it's in pipx's isolated venv)
- System pip is restricted (PEP 668 / `--break-system-packages`)
- The wrong Python is on PATH

The `mempalace` console script entry point (`[project.scripts]` in pyproject.toml) already works in all these environments — but there's no subcommand to start the MCP server through it.

Fixes #408 (MCP server portion). Related: #545, #650 (same root cause for hooks — bare `python3` calls).

## Solution

Add `mempalace mcp run` as a subcommand that starts the MCP server:

```bash
# Before (breaks with pipx)
claude mcp add mempalace -- python -m mempalace.mcp_server

# After (works everywhere)
claude mcp add mempalace -- mempalace mcp run
```

### Behavior

- **`mempalace mcp`** (no subcommand) — still shows setup instructions (existing behavior, non-breaking)
- **`mempalace mcp run`** — starts the MCP server (JSON-RPC over stdin/stdout)
- **`mempalace mcp run --palace /path`** — starts with a custom palace path
- **`python -m mempalace.mcp_server`** — still works for backward compatibility

### Changes

- `mempalace/cli.py` — add `cmd_mcp_run()` function and `mcp run` subcommand (follows the existing `hook run` pattern)
- `tests/test_cli.py` — 3 new tests (dispatch, server start, env var propagation) + updated 2 existing mcp tests
- `README.md`, `examples/` — update MCP setup commands to use `mempalace mcp run`

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Restart Claude Code, then type `/skills` to verify "mempalace" appears.
 
 ```bash
 # Connect MemPalace once
-claude mcp add mempalace -- python -m mempalace.mcp_server
+claude mcp add mempalace -- mempalace mcp run
 ```
 
 Now your AI has 19 tools available through MCP. Ask it anything:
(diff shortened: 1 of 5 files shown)


igorls:

PR #665, opened 2026-04-12
Add optional PostgreSQL backend with pg_sorted_heap support

skuznetsov:
## What does this PR do?

Follow-up to #413. This adds an optional PostgreSQL palace backend while keeping ChromaDB as the default zero-config backend.

The PostgreSQL backend supports two extension paths:

- `pg_sorted_heap` preferred path: `sorted_heap` table storage, `svec` embeddings, and lazy `sorted_hnsw` vector indexes.
- `pgvector` fallback path: regular heap table, `vector` embeddings, and lazy `hnsw` vector indexes when only `vector` is available.

It also adds:

- Backend selection via `MEMPALACE_BACKEND=postgres` / config `backend: postgres` and `MEMPALACE_POSTGRES_DSN` / config `postgres_dsn`.
- Batched `INSERT ... SELECT ... FROM unnest(...) ... ON CONFLICT` writes for add/upsert, avoiding delete-then-add upserts.
- Fail-closed metadata filter handling for unsupported operators and explicit empty id lists.
- Lazy vector index creation using PostgreSQL catalog/stat estimates instead of an exact `COUNT(*)` threshold check.
- Documentation and a helper script for installing `pg_sorted_heap` or `pgvector` from source.

ChromaDB remains the documented default and the public raw-mode benchmark path. The PostgreSQL backend is opt-in for larger, long-lived, or server/team deployments.

Related: #266

**Merge order:** This is the Stage 2 PostgreSQL backend follow-up and is intended to merge after the Stage 1 backend seam from #413. #413 is already merged into `develop`; this PR builds on that seam rather than replacing it.

## How to test

- `uv run python -m pytest tests/ -v`
  - `617 passed, 106 deselected, 413 warnings`
- `uv run pytest -q`
  - `617 passed, 106 deselected, 413 warnings`
- `uv run ruff check .`
  - passed
- `git diff --check`
  - passed
- Local PostgreSQL smoke checks were also run for both extension paths:
  - `pg_sorted_heap` / `svec`
  - `pgvector` / `vector`
  - batch `INSERT ... SELECT ... FROM unnest(...) ... ON CONFLICT`
  - estimated row-count based lazy index threshold check

## Checklist

- [x] No hardcoded paths


web3guru888 (approved):
## PostgreSQL Backend Review — Thorough Read-Through

Very solid PR. This is a well-architected backend addition that I've been hoping to see since #413 laid the seam. I read every line of `postgres.py` (631 LOC), the config changes, the mcp_server refactor, palace.py routing, tests, docs, and the install script. Here's what I found:

### Strengths

**1. Batched writes via `INSERT ... SELECT ... FROM unnest()`**
This is the right approach for PostgreSQL — avoids N round-trips and the delete-then-add antipattern that ChromaDB's `update()` effectively does. The `ON CONFLICT (id) DO UPDATE SET ...` upsert is clean and atomic. The local dedup of `rows_by_id` before hitting the DB is a nice touch that prevents wasted work on duplicate IDs within a single batch.

**2. Fail-closed metadata filter handling**
`_where_to_sql()` raising `ValueError` on unsupported operators (`$gt`, `$lt`, `$gte`, `$lte`) rather than silently ignoring them is exactly right. ChromaDB supports these, but silent degradation in a database backend would cause subtle retrieval bugs. The explicit error forces callers to adapt.

**3. Lazy vector index with catalog estimates**
Using `pg_stat_all_tables.n_live_tup` / `pg_class.reltuples` instead of `COUNT(*)` for the 5,000-row threshold check is smart — avoids a sequential scan on every write batch. The `_rows_since_index_check` interval limiter prevents even the catalog query from running on every insert. And the `_local_row_estimate` fallback handles freshly-created tables where `ANALYZE` hasn't run yet.

**4. mcp_server.py refactor is clean**
Replacing the direct `chromadb.PersistentClient` import with `palace.get_collection()` is the right abstraction. The cache invalidation via `_get_collection_cache_key()` that composes `(backend, target, collection_name, inode)` handles both backends correctly. The `_fetch_all_metadata` loop change from `while offset < total` to `while True` with break-on-empty is safer since `count()` could race with concurrent writes.

**5. `estimated_count()` on BaseCollection**
Good addition — defaults to `count()` so ChromaDB callers are unaffected, but PostgreSQL can use catalog stats. The `layers.py` change to prefer `estimated_count` when available is a nice performance win for status checks.

**6. Test coverage is thorough**
315 lines of tests covering: input validation, where-clause parsing (including `$or`, `$and`, `$in`, `$nin`, `$eq`, `$ne`), empty-id rejection, batch conflict handling, dedup within a batch, lazy index creation, estimated count, and the backend factory's create/no-create paths. All without requiring a live PostgreSQL server — good use of monkeypatching.

### Minor observations (not blocking)

**1. `_metadata_value()` stringifies everything** — metadata values go through `str()` before comparison, which means `where={"count": 5}` compares the string `"5"` against `metadata->>'count'`. This matches ChromaDB's behavior (which also stringifies in its SQLite layer), but it's worth a doc comment since PostgreSQL's `jsonb` could support typed comparisons. Could be a future enhancement.

**2. The `_embed()` function uses a module-global `_embedder`** — this is fine for single-process use, but if someone embeds MemPalace in a multi-threaded server, `SentenceTransformer.encode()` may not be thread-safe depending on the model backend. A `threading.Lock` around the lazy init would be defensive. Low priority since the typical deployment is single-process CLI.

**3. Connection lifecycle** — `autocommit = True` is the right choice for this use case (no long transactions), but if the connection drops mid-session (network blip to a remote PG), the next `_get_conn()` call will reconnect since it checks `self._conn.closed`. Good. One edge case: if psycopg2 raises `OperationalError` mid-query, the connection may be left in a bad state without `closed` being set. A try/except around cursor operations that resets `self._conn = None` on `OperationalError` would make it more resilient for long-lived server deployments.

**4. `sorted_heap` table has `PRIMARY KEY (wing, room, id)` but also a unique index on `(id)`** — this means `id` is globally unique across wings/rooms, which matches ChromaDB semantics. The composite PK gives `sorted_heap` the sort key it needs. Clean.

### Impact for downstream consumers

This is significant for anyone running MemPalace in multi-agent or server environments. The ChromaDB backend works well for single-user CLI, but concurrent access from multiple agents (e.g., specialist agents each writing to the palace simultaneously) hits ChromaDB's SQLite locking. PostgreSQL with proper connection pooling would eliminate that bottleneck. The `BaseCollection` abstraction also opens the door for future backends (LanceDB per #574, etc.).

APPROVE — this is ready to merge. Clean architecture, solid test coverage, thorough docs. The minor observations above are suggestions for follow-up, not blockers.

dekoza (commented):
**Concern: pg_sorted_heap as "preferred" path introduces a single-maintainer dependency**

Thanks for the solid work on the backend abstraction — the lazy HNSW index creation and extension auto-detection are well thought out.

I want to flag a concern about the dual-path architecture. The "preferred" backend (`pg_sorted_heap`) is maintained by the PR author at `github.com/skuznetsov/pg_sorted_heap`. This creates a situation where:

1. **Availability**: `pg_sorted_heap` is not packaged in any Linux distribution, not available on any managed PostgreSQL service (RDS, Cloud SQL, Neon, Supabase), and not listed on PGXN. The 215-line install script in this PR exists because there is no standard installation path. By contrast, pgvector is available everywhere.

2. **Maintenance surface**: Every SQL statement in the 631-line backend is forked into two code paths (DDL, types, index operators, casts). This doubles the review and maintenance burden for a path that, realistically, very few users will run.

3. **Marginal benefit for this workload**: The physical sort ordering benefits sequential scans within a wing/room, but the dominant access pattern here is HNSW vector search across all data, where heap layout is irrelevant.

4. **Optics**: A contributor introducing their own extension as the recommended path in someone else's project — even with a working fallback — is a pattern that erodes trust. I'm not questioning the author's intent, but the project should be cautious about taking on single-maintainer C extension dependencies that only the contributor can debug.

**Suggestion**: Ship this PR with pgvector as the *only* path. If `pg_sorted_heap` matures to the point where it has distro packages, managed-service availability, or independent adopters, it can be added later as an optional optimization — not the default.

This would cut ~200 lines of dual-path code, eliminate the install script, and remove a bus-factor-of-one dependency from the project.

dekoza:
*(Updated version of my review comment above — adding a question for @web3guru888)*

---

**Concern: pg_sorted_heap as "preferred" path introduces a single-maintainer dependency**

Thanks for the solid work on the backend abstraction — the lazy HNSW index creation and extension auto-detection are well thought out.

I want to flag a concern about the dual-path architecture that I do not think was addressed in @web3guru888's otherwise thorough review. The "preferred" backend (`pg_sorted_heap`) is maintained by the PR author at `github.com/skuznetsov/pg_sorted_heap`. This creates a situation where:

1. **Availability**: `pg_sorted_heap` is not packaged in any Linux distribution, not available on any managed PostgreSQL service (RDS, Cloud SQL, Neon, Supabase), and not listed on PGXN. The 215-line install script in this PR exists because there is no standard installation path. By contrast, pgvector is available everywhere.

2. **Maintenance surface**: Every SQL statement in the 631-line backend is forked into two code paths (DDL, types, index operators, casts). This doubles the review and maintenance burden for a path that, realistically, very few users will run.

3. **Marginal benefit for this workload**: The physical sort ordering benefits sequential scans within a wing/room, but the dominant access pattern here is HNSW vector search across all data, where heap layout is irrelevant.

4. **Optics**: A contributor introducing their own extension as the recommended path in someone else's project — even with a working fallback — is a pattern the project should evaluate carefully. I'm not questioning the author's intent, but the project should be cautious about taking on single-maintainer C extension dependencies that only the contributor can debug.

@web3guru888 — your review praised the composite PK design on the `sorted_heap` path and the dual-backend detection, but I'd appreciate your take on the supply-chain concern: should this project take a hard dependency path on an unpackaged extension authored by the same person submitting the PR? The pgvector fallback works identically for all practical purposes.

**Suggestion**: Ship this PR with pgvector as the *only* path. If `pg_sorted_heap` matures to the point where it has distro packages, managed-service availability, or independent adopters, it can be added later as an optional optimization — not the default.

This would cut ~200 lines of dual-path code, eliminate the install script, and remove a bus-factor-of-one dependency from the project.

skuznetsov:
Thanks for raising this. I agree the availability / bus-factor concern is real in the general case, and I do not want this PR to make `pg_sorted_heap` a hard dependency for MemPalace.

I pushed 5976c2f to tighten the docs framing:

- ChromaDB remains the zero-config default.
- PostgreSQL remains opt-in.
- Within PostgreSQL, `pgvector` is now documented as the broadly available managed-service / simpler supply-chain path.
- `pg_sorted_heap` is now documented as an optional optimized self-managed path, not a blanket recommendation.
- The docs also call out that `pg_sorted_heap` is listed in the [PostgreSQL Software Catalogue](https://www.postgresql.org/download/products/6/), but is not bundled with managed PostgreSQL providers.

Update: the PGXN point was fair when raised. `pg_sorted_heap` is now also published on PGXN: https://pgxn.org/dist/pg_sorted_heap/ (`0.13.0`, published 2026-04-13). I would still keep the docs wording conservative: PGXN publication improves the standard discovery/download path, but it does not make the extension available on managed PostgreSQL providers.

I do not think removing the `pg_sorted_heap` path is the right tradeoff for this PR, because the fallback keeps the dependency optional and the dual-path support is useful for validating the backend seam against more than one PostgreSQL storage/index implementation. The runtime selection is also explicit in the docs now: if both extensions are present, MemPalace selects `pg_sorted_heap`; if an operator wants the pgvector path, install only `vector`.

So the intended boundary is: no new hard dependency for MemPalace, no managed-service requirement to use `pg_sorted_heap`, but keep the optimized self-managed path available for people who intentionally install that extension.

skuznetsov:
Follow-up after #995 / RFC 001 landed:

I rewrote this PR on top of the merged backend contract instead of keeping the old Chroma-shaped adapter. The current head is `83d5448b6166a67b27d3c903fcf69cf800b0be6c` and now implements the RFC path directly:

- `PostgresBackend` implements `BaseBackend` with `get_collection(*, palace: PalaceRef, collection_name, create=False, options=None)`.
- `PostgresCollection` implements `BaseCollection` and returns the typed `QueryResult` / `GetResult` shapes.
- Backend selection now goes through the registry/config path from the RFC work, with Chroma still remaining the default backend.
- The PostgreSQL path remains opt-in, with `pgvector` as the broadly available fallback and `pg_sorted_heap` as an optional self-managed optimized path when that extension is intentionally installed.

I also re-ran the hostile review pass after the rewrite. Current verification:

- local full suite: `uv run --frozen python -m pytest tests/ -q --ignore=tests/benchmarks` -> `1044 passed, 1 warning`
- local targeted suite: `77 passed`
- `ruff check`, `ruff format --check`, `uv lock --check`, `git diff --check`, and install-script syntax check passed
- local PostgreSQL 18 smoke with `pg_sorted_heap` verified add/upsert/query/get/delete/health and `sorted_heap` table AM
- local PostgreSQL 18 smoke with `pgvector` fallback and forced vector-index creation was also run before the final rebase
- GitHub CI is green on lint, version guard, Linux 3.9/3.11/3.13, macOS, and Windows

So the review surface should now be the RFC-based backend implementation, not the pre-#995 adapter shape.

jphein:
Operator data point on the pg_sorted_heap-vs-pgvector decision: the jphein fork cherry-picked this PR onto fork main (commit `5e90c72` in [#21](https://github.com/jphein/mempalace/pull/21)) and has been running the pgvector fallback path against a homelab `apache/age:release_PG16_1.6.0` container with `postgresql-16-pgvector` apt-installed on top — no `pg_sorted_heap`.

The fallback codepath works end-to-end:

- 3 smoke tests in `tests/test_backends_postgres.py` pass (BaseCollection conformance + drawer round-trip + L2 distance ordering on a `vector(384)` column)
- `palace.get_collection()` routes through `resolve_backend_for_palace` cleanly with `backend=postgres` in config; `PostgresCollection` is exposed as the singleton via the entry-point registry
- `palace=PalaceRef(id=..., local_path=...)` + `options={"dsn": ...}` plumb correctly from the caller boundary into psycopg2

@dekoza's bus-factor concern (pg_sorted_heap availability) doesn't reproduce on this deployment because we never hit that codepath — pgvector's the only vector type installed; `_detect_extensions` resolves to it; the sorted_heap branch is dead code on this image.

This isn't a vote against `pg_sorted_heap` as the preferred path *when available* — your 2026-04-19 rewrite kept the conditional resolution clean. It's a signal that the pgvector fallback is genuinely production-shaped, not just a placeholder. Happy to share the docker-compose service definition + smoke-test harness if useful for your own CI substrate.

Full rationale for the fork's "wait + cherry-pick" composition stance is at [`docs/internal/pgvector-665-decision.md`](https://github.com/jphein/mempalace/blob/main/docs/internal/pgvector-665-decision.md); Plan-B trigger documented as 2026-06-08 if review stalls. Looking forward to this merging.

jphein:
Operator follow-up to my [2026-05-11 comment](https://github.com/MemPalace/mempalace/pull/665#issuecomment-4425704422) above — surfacing a concurrency issue we hit running this code in production.

## What we observed

On `disks`, `palace-daemon`'s `/mcp` endpoint wedged for 28+ minutes. Three concurrent `CREATE INDEX … USING hnsw` queries held `ACCESS EXCLUSIVE` on `mempalace_drawers`, with `INSERT`/`DELETE`/secondary `CREATE INDEX` queued behind them.

## Race

`_maybe_create_vector_index` does a lookup-then-create with no inter-connection guard:

```python
cur.execute("SELECT 1 FROM pg_indexes WHERE indexname = %s", (index_name,))
if cur.fetchone():
    self._vector_index_ready = True
    return
# ...
cur.execute(
    self._sql.SQL("CREATE INDEX {} ON {} USING {} (embedding {})").format(...)
)
```

Under any concurrency setting that allows more than one writer (we run `PALACE_MAX_CONCURRENCY=4`), three writers tipping the `VECTOR_INDEX_CHECK_INTERVAL_ROWS` threshold in the same window each do the same SELECT, each see no index, each issue `CREATE INDEX`. All three then serialize on `ACCESS EXCLUSIVE`, full HNSW builds stack up, and concurrent writes block for the duration of the build.

## Aggravating factor (separate, lower priority)

The existence check is by literal name `{table_name}_vec_idx`. Our migration tool had created a functionally-identical HNSW index on `(embedding vector_cosine_ops)` under a different name (`mempalace_drawers_embedding_hnsw`) — the check missed it and the lazy-create fired on every threshold cross even though a perfectly good HNSW already existed on the same column.

This second issue is more opinionated to fix (does mempalace *own* the index name, or accept any compatible HNSW?) — happy to defer that to a design conversation.

## Minimal hardening that would have prevented the wedge

Two-line addition; keeps the existing flow:

```python
# Inter-connection guard + DDL idempotency
with self._get_conn().cursor() as cur:
    cur.execute("SELECT pg_advisory_xact_lock(hashtext(%s))", (f"vec_idx:{self.table_name}",))
    cur.execute("SELECT 1 FROM pg_indexes WHERE indexname = %s", (index_name,))
    if cur.fetchone():
        self._vector_index_ready = True
        return
    # ...
    cur.execute(
        self._sql.SQL("CREATE INDEX IF NOT EXISTS {} ON {} USING {} (embedding {})").format(...)
    )
```

- `pg_advisory_xact_lock` serializes the check+create across connections, auto-releases at commit.
- `IF NOT EXISTS` belt-and-suspenders against the lock being skipped or the threshold being crossed mid-build by a backend that wasn't holding the advisory.

`CREATE INDEX CONCURRENTLY` would be a bigger but also worthwhile change — pgvector ≥ 0.5.0 supports it for `hnsw`, and the trade-off (≈2× build time, no lock) is favorable for write-heavy workloads. Happy to scope that separately.

## Local mitigation

We unwedged out-of-band by canceling the duplicate `CREATE INDEX` backends and letting one finish — the deployment is healthy again. We also detached our daemon's hook caller so we don't block the harness on this case if it ever recurs.

## Offer

If a fixup commit on this PR (minimal version above) would help, happy to put one up. Wanted to surface the diagnosis here first since this is your code path and you may want a different shape.

skuznetsov:
Thanks for the production report and the clear lock diagnosis, @jphein. The race is real, and your observation about multiple writers crossing the lazy-index threshold at the same time was the missing operational case.

I pushed `85a42f1` with the fix:

- The lazy vector-index check/create path is now serialized across PostgreSQL connections with a session-level advisory lock.
- I intentionally used `pg_advisory_lock(...)` / `pg_advisory_unlock(...)`, not `pg_advisory_xact_lock(...)`, because this backend connection currently runs with `autocommit = True`; a transaction-level advisory lock would be released at the end of the lock statement and would not protect the following `CREATE INDEX`.
- The DDL is now `CREATE INDEX IF NOT EXISTS` as a second idempotency guard.
- The code re-checks for an existing compatible index while holding the lock.
- It also now accepts an existing ready/valid HNSW or `sorted_hnsw` index on `embedding` with the expected opclass even if the index was created under a different name, so operator-managed indexes like the one you described should not trigger repeated lazy creation.

Verification I ran locally:

- `uv run pytest tests/test_backends.py -q` -> `45 passed`
- `uv run pytest tests/ -q --ignore=tests/benchmarks` -> `1047 passed, 1 warning`
- `uv run ruff check mempalace/backends/postgres.py tests/test_backends.py`
- `uv run ruff format --check mempalace/backends/postgres.py tests/test_backends.py`
- `git diff --check`
- small PostgreSQL catalog sanity probe for the `pg_index` / `pg_opclass` compatible-index query shape

I agree `CREATE INDEX CONCURRENTLY` is worth considering separately, but I kept this commit focused on preventing duplicate lazy builds and avoiding the wedge you hit.

jphein:
Third operator follow-up — substrate cutover is now production-stable on our fork. Posting numbers so they're indexed against this PR while it's still in flight.

## State

- **Cutover complete on `techempower-org/mempalace` (formerly `jphein/mempalace`).** Postgres backend + pgvector + Apache AGE serving live workload via `palace-daemon` on `disks:8085`. 273k drawers across 7 canonical rooms. Familiar ([techempower-org/familiar.realm.watch](https://github.com/techempower-org/familiar.realm.watch)) reads it over HTTP for live chat retrieval.
- **Hybrid retrieval shipped.** `candidate_strategy="hybrid"` unions vector ∪ BM25 (postgres `tsvector`) ∪ graph-expanded candidates (AGE) and hybrid-reranks. Daemon-side endpoints: `/search/keyword`, `/search/hybrid`.
- **pgvector lazy-index race fix from my [2026-05-14 comment](https://github.com/MemPalace/mempalace/pull/665#issuecomment-4429660843)** landed on our fork at `4566f8a` and has held up under concurrent mining since. Recovery procedure from that comment exercised once cleanly during a separate incident.

## Benchmark results — postgres backend, scale=small, 273k-drawer palace

8 of 9 mempalace `tests/benchmarks/` suites passed end-to-end. (Skipped `test_chromadb_stress` — chromadb-specific, doesn't apply on postgres.) Wall: 1h 24m 25s on `--bench-scale=small`. Full JSON snapshot landed on our fork at [techempower-org/mempalace#78](https://github.com/techempower-org/mempalace/pull/78).

Headlines:

| Bench | Number | Notes |
|---|---|---|
| `test_ingest_bench` @ 100 files | **6.5 files/s, 15.3 drawers/s** | peak RSS 389 MB (Δ 25.5 MB), no allocator churn |
| `test_search_bench` p50 (hybrid) | **125 ms** | from a separate 2026-05-14 run, captured at [2026-05-14-search-bench-hybrid-cutover.json](https://github.com/techempower-org/mempalace/blob/main/docs/benchmarks/2026-05-14-search-bench-hybrid-cutover.json) |
| `test_search_bench` p99 (4 workers) | **200 ms** | concurrent recall stable |
| `test_knowledge_graph_bench` | **flat** | latency stable 500 → 5000 scale points (AGE indexes holding) |
| `test_layers_bench` | **sub-linear** | RSS growth on unbounded fetch through 5000 scale — the chromadb-era quadratic does not reproduce on postgres |
| `test_mcp_bench` repeated calls | **no leak** | RSS plateau across long sessions |
| `test_memory_profile` heap top | **identical shape** vs chromadb-era | no regression on the search hot path post-cutover |

The "no regression vs chromadb-era heap allocators" was the result I was most worried about given the substrate switch — it survived.

## Methodology callouts for anyone implementing alternative encoders against this backend

`chromadb.api.types.EmbeddingFunction` 1.5+ added `embed_query` to the protocol. A custom encoder that defines only `__call__` + `name()` works during `Collection.upsert` but raises `AttributeError` on `Collection.query(query_texts=...)`. mempalace's `searcher` catches the `AttributeError` and **silently falls back to BM25**, so the swap looks like it ran but the encoder's embeddings are never actually exercised on queries. Subclassing `EmbeddingFunction` (which inherits a default `embed_query` that delegates to `__call__`) fixes it.

Worth flagging because this also affects anyone implementing the `Backend` interface ([RFC 001 / #743](https://github.com/MemPalace/mempalace/pull/743)) with a SentenceTransformer-shaped encoder — the protocol mismatch shows up at query time on the same layer. We landed a defensive doc note on our fork at [techempower-org/mempalace#80](https://github.com/techempower-org/mempalace/pull/80); happy to upstream as a one-line comment in `mempalace/embedding.py` if useful.

## Adjacent work this surfaced

- **3-way RRF reproduction** at n=200 on the same backend ([techempower-org/mempalace#77](https://github.com/techempower-org/mempalace/pull/77), tied to [#1384](https://github.com/MemPalace/mempalace/discussions/1384)): default ONNX + 2 adaptmem FT-Code checkpoints fuse to MRR 0.5101 (+0.0841 over best solo). Postgres backend with `pg_trgm` GIN + tsvector GIN handles the union-then-rerank fan-out cleanly.
- **n=200 git-derived probe set** for chunking ablations at the same PR — replaces a hand-curated n=20 set whose paired bootstrap CIs all overlapped zero. Eliminates the "your probe set is hand-picked" objection for any future retrieval ablation against the postgres backend.

Happy to share probe-level rank data, raw bench JSONs, or the daemon-side `/search/hybrid` trace format if any of this is useful while #665 lands.

Lcstyle:
Flagging a defect in `_embed()` as it stands in this PR, since the file has been
cherry-picked downstream and the issue reproduces there. Filed in full at
techempower-org/mempalace#413; summarising here because this is the origin.

`_embed()` builds its own `chromadb.utils.embedding_functions.DefaultEmbeddingFunction()`
rather than calling `mempalace.embedding.get_embedding_function()`. Because every
embedding improvement in the project lands inside that resolver, this backend
misses all of them. Three things compound:

1. **`MEMPALACE_EMBEDDING_MODEL` is ignored.** The support added in #1559 for
   OpenAI-compatible endpoints lives in the resolver, so a configured remote
   embedder never receives work from this backend. Measured on two installs: the
   configured endpoint served **9 texts in 18.3 hours** of continuous mining.
   Same silent-fallback shape as #2324.

2. **The ORT thread cap from #1068 is bypassed.** PR #1836's fix is implemented
   as `_MempalaceONNX` and wired up only inside `get_embedding_function()`.
   Calling `DefaultEmbeddingFunction()` directly skips it: **~49 extra OS threads
   per embedder** on a 48-core host, daemon CPU averaging 210–265% and sampled at
   2550% mid-mine.

3. **The ONNX session is rebuilt on every call.** In chromadb 1.5.9,
   `DefaultEmbeddingFunction` is a shim whose `__call__` is
   `ONNXMiniLM_L6_V2()(input)` — a fresh instance per call — and
   `ONNXMiniLM_L6_V2.model` is a per-*instance* `cached_property`. So the
   module-global `_embedder` caches an object holding no model. Three
   consecutive calls: **0.706s / 0.711s / 0.713s**, perfectly flat, no warm-up
   benefit ever, versus 1.257 / 0.532 / 0.539 for a reused instance.

The docstring's rationale — avoiding "a second ML dependency stack" — is fully
preserved by delegating, because with no embedding configuration set
`get_embedding_function()` still returns a local ONNX MiniLM embedder. Zero new
dependencies, identical default behaviour.

Delegating inside `_embed` fixed all three for us: per-batch embed time
**0.706s → 0.016s**.

Worth noting for anyone applying the `requires_explicit_embeddings` capability
instead: on its own it is not sufficient, because it only takes effect at
`palace.get_collection()`, and both `mcp_server._get_collection_postgres()` and
`convo_miner.mine_sessions()` obtain collections without going through that wrap
point. Adding the capability *and* making `_embed` raise would break the MCP
server and the miner.

PR #667, opened 2026-04-12
feat: new MCP tools — get/list/update drawer, hook settings, export (resolves #635)

### LINKED ISSUE #646: bug: _try_claude_ai_json parser silently produces 0 drawers on claude.ai export (conversations.json)
ersion: 3.1.0
OS: macOS, Apple Silicon
Export: claude.ai data export — conversations.json (21.9 MB, claude AI JSON format)

The _try_claude_ai_json parser exists but silently skips conversations.json.

Steps to reproduce:
1. mempalace mine ~/Downloads/data-2026-04-08-19-25-24-batch-0000 --mode convos
   → Files processed: 3 (users.json, memories.json, projects.json)
   → conversations.json: skipped, no error

2. mempalace mine ~/Downloads/.../conversations.json --mode convos
   → Files: 0, Drawers filed: 0, no error message

Expected: conversations parsed per conversation into separate drawers
Actual: 0 drawers, no error or warning

bensig:
## Summary

Merge-conflict resolution of #635 by @jphein against develop (after #647 landed).

All code is from #635 — conflicts in `mcp_server.py` resolved:
- WAL write: kept `os.open` atomic approach from #647
- tool_search: kept `_MAX_RESULTS` + `min_similarity` compat from #635
- schema_props filter: removed duplicate block (already present above)

Fixes #646 (0 drawers on claude.ai export) via paginated `exporter.py`.

Original PR: #635
Original author: @jphein — all feature code in this PR is their work.

Co-authored-by: jphein <jphein@users.noreply.github.com>

## Test plan
- [ ] CI passes
- [ ] `mempalace export` produces correct output on large palaces

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -173,6 +173,27 @@ class MempalaceConfig:
         """Mapping of hall names to keyword lists."""
         return self._file_config.get("hall_keywords", DEFAULT_HALL_KEYWORDS)
 
+    @property
+    def hook_silent_save(self):
+        """Whether the stop hook saves directly (True) or blocks for MCP calls (False)."""
+        return self._file_config.get("hooks", {}).get("silent_save", True)
+
+    @property
+    def hook_desktop_toast(self):
+        """Whether the stop hook shows a desktop notification via notify-send."""
+        return self._file_config.get("hooks", {}).get("desktop_toast", False)
+
+    def set_hook_setting(self, key: str, value: bool):
+        """Update a hook setting and write config to disk."""
+        if "hooks" not in self._file_config:
+            self._file_config["hooks"] = {}
+        self._file_config["hooks"][key] = value
+        try:
+            with open(self._config_file, "w", encoding="utf-8") as f:
+                json.dump(self._file_config, f, indent=2, ensure_ascii=False)
+        except OSError:
+            pass
+
     def init(self):
         """Create config directory and write default config.json if it doesn't exist."""
         self._config_dir.mkdir(parents=True, exist_ok=True)
--- a/mempalace/exporter.py
+++ b/mempalace/exporter.py
@@ -1,0 +1,153 @@
+"""
+exporter.py — Export the palace as a browsable folder of markdown files.
+
+Produces:
+  output_dir/
+    index.md              — table of contents
+    wing_name/
+      room_name.md        — one file per room, drawers as sections
+
+Streams drawers in paginated batches so memory usage stays bounded
+regardless of palace size.
+"""
+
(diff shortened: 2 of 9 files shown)


bensig:
### Code review

Found 3 issues:

1. `min_similarity` backwards-compat shim passes raw similarity value as `max_distance`, but these are opposite scales (higher similarity = stricter, lower distance = stricter). Passing `min_similarity=0.8` results in `max_distance=0.8` which is permissive, not strict. Should convert: `dist = (1 - min_similarity) if min_similarity is not None else max_distance`.

https://github.com/MemPalace/mempalace/blob/4f54529fa434d95bdbd583888117c90152e233d4/mempalace/mcp_server.py#L574-L576

2. `tool_status`, `tool_list_wings`, `tool_list_rooms`, and `tool_get_taxonomy` replace structured error reporting (`{"error": str(e), "partial": True}`) with bare `except: pass`. This directly reverts the error surfacing added in the recently merged #647 security hardening. Callers now get empty/incomplete results with no signal something went wrong.

https://github.com/MemPalace/mempalace/blob/4f54529fa434d95bdbd583888117c90152e233d4/mempalace/mcp_server.py#L426-L434

3. `_get_client` inode cache: when `os.stat` fails (file missing), `current_inode = 0`. The condition `current_inode and current_inode != _palace_db_inode` short-circuits on falsy 0, so a missing/deleted database silently reuses the stale cached client instead of reconnecting.

https://github.com/MemPalace/mempalace/blob/4f54529fa434d95bdbd583888117c90152e233d4/mempalace/mcp_server.py#L317-L321

jphein:
Great catches, all 5. I've incorporated every fix into my fork:

1. **min_similarity conversion** — embarrassing but fair, `1.0 - similarity` is obviously correct for cosine distance. I only ever used `max_distance` directly so this never bit me in production.
2. **Structured error reporting** — your `result["partial"] = True` pattern is much better than the bare `except: pass` I had. Already adopted it across my fork.
3. **Inode zero guard** — subtle one. The `current_inode and ...` short-circuit was masking the missing-DB case. My #663 (stale HNSW) PR had an explicit `os.path.isfile()` guard that worked around it, but removing the falsy check is the right fix at the root.
4. **Pagination guard** — classic TOCTOU race. Moving the empty check before extend prevents the infinite loop on concurrent deletion.
5. **KG init** — reverted in my #626 too. `KnowledgeGraph()` knows its own default path.

Thanks for the thorough review — learned something on each one.

Ari4ka (approved):

PR #670, opened 2026-04-12
fix: resolve Python interpreter in hook scripts instead of hardcoding python3

z3tz3r0:
## Summary

- **Root cause**: All hook shell scripts hardcode `python3`, which breaks on Windows (`python3` doesn't exist or resolves to Python 3.14 via Windows Store), pipx/uv-tool installs (`python3` resolves to system Python instead of the install venv), and non-standard environments.
- **`hooks/mempal_save_hook.sh`**, **`hooks/mempal_precompact_hook.sh`**: Add `find_python()` that checks `MEMPALACE_PYTHON` env var, then `python3`, then `python`. Replace all 5 hardcoded `python3` calls with `$PYTHON`.
- **`.claude-plugin/hooks/*.sh`**, **`.codex-plugin/hooks/*.sh`**: Same `find_python()` added to all 3 plugin hook wrappers.
- **`mempalace/mcp_server.py`**: Add Python 3.14+ version guard — emits a clear JSON-RPC error ("MemPalace requires Python 3.9-3.13") instead of crashing with a cryptic pydantic `ConfigError` from chromadb.

## Test plan

- [x] `pytest tests/ -v` — 589 passed, 0 failed
- [x] `grep` confirms zero bare `python3` command invocations in hook scripts
- [ ] On Windows: set `MEMPALACE_PYTHON=python` and verify hooks resolve correctly
- [ ] On Windows: run MCP server with Python 3.14 and verify the version guard emits a clear error

Addresses #545, #650

### DIFF
--- a/.claude-plugin/hooks/mempal-precompact-hook.sh
+++ b/.claude-plugin/hooks/mempal-precompact-hook.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 # MemPalace PreCompact Hook — thin wrapper calling Python CLI
 # All logic lives in mempalace.hooks_cli for cross-harness extensibility
+find_python() {
+    if [ -n "${MEMPALACE_PYTHON:-}" ]; then echo "$MEMPALACE_PYTHON"
+    elif command -v python3 &>/dev/null; then echo "python3"
+    elif command -v python &>/dev/null; then echo "python"
+    else echo "python3"; fi
+}
+PYTHON=$(find_python)
 INPUT=$(cat)
-echo "$INPUT" | python3 -m mempalace hook run --hook precompact --harness claude-code
+echo "$INPUT" | $PYTHON -m mempalace hook run --hook precompact --harness claude-code
--- a/.claude-plugin/hooks/mempal-stop-hook.sh
+++ b/.claude-plugin/hooks/mempal-stop-hook.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 # MemPalace Stop Hook — thin wrapper calling Python CLI
 # All logic lives in mempalace.hooks_cli for cross-harness extensibility
+find_python() {
+    if [ -n "${MEMPALACE_PYTHON:-}" ]; then echo "$MEMPALACE_PYTHON"
+    elif command -v python3 &>/dev/null; then echo "python3"
+    elif command -v python &>/dev/null; then echo "python"
+    else echo "python3"; fi
+}
+PYTHON=$(find_python)
 INPUT=$(cat)
-echo "$INPUT" | python3 -m mempalace hook run --hook stop --harness claude-code
+echo "$INPUT" | $PYTHON -m mempalace hook run --hook stop --harness claude-code
--- a/.codex-plugin/hooks/mempal-hook.sh
+++ b/.codex-plugin/hooks/mempal-hook.sh
@@ -1,9 +1,16 @@
 #!/usr/bin/env bash
 set -euo pipefail
 HOOK_NAME="${1:?Usage: mempal-hook.sh <hook-name>}"
+find_python() {
+    if [ -n "${MEMPALACE_PYTHON:-}" ]; then echo "$MEMPALACE_PYTHON"
+    elif command -v python3 &>/dev/null; then echo "python3"
+    elif command -v python &>/dev/null; then echo "python"
+    else echo "python3"; fi
+}
+PYTHON=$(find_python)
 INPUT_FILE=$(mktemp) || { echo "Failed to create temp file" >&2; exit 1; }
 cat > "$INPUT_FILE"
-cat "$INPUT_FILE" | python3 -m mempalace hook run --hook "$HOOK_NAME" --harness codex
+cat "$INPUT_FILE" | $PYTHON -m mempalace hook run --hook "$HOOK_NAME" --harness codex
 EXIT_CODE=$?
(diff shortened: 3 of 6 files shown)


ianhe8x:
I think the preferred way is to first try mempalace binary if it can be found

igorls:
Thanks for tackling this — the underlying need is real (we use uv in some setups and don't always have `python` on PATH, so `MEMPALACE_PYTHON` as an escape hatch is exactly right).

Blocker before we merge: **the two Windows checkboxes in the test plan are unchecked**, and Windows is the primary platform this fix targets. Specifically the scenarios that would confirm the fix works:

1. **Git Bash on Windows 11** with no `python3` on PATH but `python` present — confirm `find_python` falls back to `python` and the save hook's inline `python -c` JSON parsing actually runs.
2. **`MEMPALACE_PYTHON` override** — set it to an absolute path like `C:\Users\you\.local\pipx\venvs\mempalace\Scripts\python.exe` and verify the background `mempalace mine` call picks it up.
3. **Python 3.14 guard** — launch the MCP server under 3.14 and confirm the JSON-RPC error lands on stdout without the pydantic `ConfigError`.

Without those, we'd be merging a Windows fix on faith. Can you run through those three and report back what you saw?

Two secondary points to consider while you're in there:

- **Python 3.14 guard is future-fragile.** `sys.version_info >= (3, 14)` becomes wrong the moment chromadb supports 3.14 — the guard will tell users to downgrade when they no longer need to. Two alternatives that age better: (a) catch the pydantic `ConfigError` around the chromadb import and re-raise with a better message, or (b) gate the guard on a chromadb version range (`if chromadb.__version__ < "1.0" and sys.version_info >= (3, 14)`). Either way the message stays useful; neither locks us out later.
- **`else echo "python3"`** in `find_python` silently keeps the broken default when both `python3` and `python` are missing. It'd be friendlier to `echo "" && return 1` and have the caller exit with a clear "no Python found — set MEMPALACE_PYTHON" message, since downstream `$PYTHON -c` calls will fail opaquely otherwise.

Context on prior art: [#740](https://github.com/MemPalace/mempalace/pull/740) by @milla-jovovich touched the same files and was closed yesterday. The closure reasoning was narrower than your PR's scope (nohup PATH stripping in a private variant, not addressed by the public `python3 -m mempalace` entry point). Your PR legitimately covers a different angle — the inline `python3 -c` / `python3 - <<PYEOF` calls in the save hook bypass the module entry point and hit the `python3`-doesn't-exist path on Windows directly. Worth naming that distinction in the PR body so reviewers don't bounce it as a duplicate.

cc @milla-jovovich — you closed #740 with a clear model of the hook failure modes; wanted to make sure your perspective is folded in before this lands.

z3tz3r0:
> Thanks for tackling this — the underlying need is real (we use uv in some setups and don't always have `python` on PATH, so `MEMPALACE_PYTHON` as an escape hatch is exactly right).
> 
> Blocker before we merge: **the two Windows checkboxes in the test plan are unchecked**, and Windows is the primary platform this fix targets. Specifically the scenarios that would confirm the fix works:
> 
> 1. **Git Bash on Windows 11** with no `python3` on PATH but `python` present — confirm `find_python` falls back to `python` and the save hook's inline `python -c` JSON parsing actually runs.
> 2. **`MEMPALACE_PYTHON` override** — set it to an absolute path like `C:\Users\you\.local\pipx\venvs\mempalace\Scripts\python.exe` and verify the background `mempalace mine` call picks it up.
> 3. **Python 3.14 guard** — launch the MCP server under 3.14 and confirm the JSON-RPC error lands on stdout without the pydantic `ConfigError`.
> 
> Without those, we'd be merging a Windows fix on faith. Can you run through those three and report back what you saw?
> 
> Two secondary points to consider while you're in there:
> 
> * **Python 3.14 guard is future-fragile.** `sys.version_info >= (3, 14)` becomes wrong the moment chromadb supports 3.14 — the guard will tell users to downgrade when they no longer need to. Two alternatives that age better: (a) catch the pydantic `ConfigError` around the chromadb import and re-raise with a better message, or (b) gate the guard on a chromadb version range (`if chromadb.__version__ < "1.0" and sys.version_info >= (3, 14)`). Either way the message stays useful; neither locks us out later.
> * **`else echo "python3"`** in `find_python` silently keeps the broken default when both `python3` and `python` are missing. It'd be friendlier to `echo "" && return 1` and have the caller exit with a clear "no Python found — set MEMPALACE_PYTHON" message, since downstream `$PYTHON -c` calls will fail opaquely otherwise.
> 
> Context on prior art: [#740](https://github.com/MemPalace/mempalace/pull/740) by @milla-jovovich touched the same files and was closed yesterday. The closure reasoning was narrower than your PR's scope (nohup PATH stripping in a private variant, not addressed by the public `python3 -m mempalace` entry point). Your PR legitimately covers a different angle — the inline `python3 -c` / `python3 - <<PYEOF` calls in the save hook bypass the module entry point and hit the `python3`-doesn't-exist path on Windows directly. Worth naming that distinction in the PR body so reviewers don't bounce it as a duplicate.
> 
> cc @milla-jovovich — you closed #740 with a clear model of the hook failure modes; wanted to make sure your perspective is folded in before this lands.

thanks for the review.
For Windows machine, I'm afraid to let you know that I don't have any windows machine to test but anyone with Window machine is welcome and would be glad to continue if there's anything broke.

igorls:

PR #676, opened 2026-04-12
fix: handle large claude.ai exports and multi-conversation "messages" key

z3tz3r0:
## Summary

- **Root cause 1**: `MAX_FILE_SIZE` in `convo_miner.py` was 10 MB — claude.ai exports routinely exceed this (21+ MB for active users). Files were silently skipped with zero feedback. Raised to 100 MB and added a warning when files are skipped.
- **Root cause 2**: `_try_claude_ai_json` in `normalize.py` only detected multi-conversation exports using the `"chat_messages"` key (privacy export). Standard claude.ai exports use `"messages"` — these fell through to the flat-messages parser which failed silently (conversation dicts have no `"role"` at top level), producing 0 drawers.
- **Parser fix**: Now checks for both `"chat_messages"` and `"messages"` at the conversation object level, and processes each conversation into a separate transcript section instead of concatenating all 844+ conversations into one.
- **Tests**: 3 new test cases for multi-conversation parsing (`"messages"` key, per-conversation separation, short conversation filtering).

Note: #646 was closed via #667, but #667 addresses paginated export/read-back — it does not touch `convo_miner.py` or the `MAX_FILE_SIZE` skip, nor the `"messages"` key mismatch in the parser. The two root causes reported in #646 remain unfixed on `main`.

## Test plan

- [x] `pytest tests/ -v` — 592 passed (589 base + 3 new), 0 failed
- [x] New tests verify: `"messages"` key parsing, per-conversation separation, short conversation filtering
- [ ] Mine a real claude.ai `conversations.json` export (> 10 MB) and verify drawers are created per conversation

Addresses #646

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -28,7 +28,7 @@ CONVO_EXTENSIONS = {
 }
 
 MIN_CHUNK_SIZE = 30
-MAX_FILE_SIZE = 10 * 1024 * 1024  # 10 MB — skip files larger than this
+MAX_FILE_SIZE = 100 * 1024 * 1024  # 100 MB — claude.ai exports routinely exceed 10 MB
 
 
 # =============================================================================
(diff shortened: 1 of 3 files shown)


igorls:

PR #678, opened 2026-04-12
Add Streamable HTTP MCP transport for ChatGPT apps

ac-opensource:
## What does this PR do?

Adds remote MCP support for MemPalace so it can be exposed to ChatGPT custom apps over HTTP, while keeping the existing local stdio transport for Claude Code and similar clients.

This PR:
- adds a Streamable HTTP transport at `/mcp` alongside the existing stdio transport
- keeps the existing MemPalace tool surface intact, including tool annotations for read-only and destructive actions
- updates the docs to describe the ChatGPT setup correctly as remote-only and web-configured
- adds regression tests for the MCP handshake and HTTP transport behavior
- tightens MCP protocol handling so the server now responds correctly to `ping`, rejects unsupported `MCP-Protocol-Version` headers, and rejects unsupported notifications instead of silently accepting them

## How to test

1. Run the automated checks:
   - `python3 -m pytest -q`
   - `ruff check .`
2. Verify stdio MCP behavior:
   - `python3 -m mempalace.mcp_server`
   - send `initialize`, `ping`, and `tools/list` JSON-RPC messages over stdin and confirm valid responses
3. Verify HTTP MCP behavior:
   - `python3 -m mempalace.mcp_server --transport streamable-http --host 127.0.0.1 --port 8768 --allow-origin https://chatgpt.com`
   - POST `initialize`, `ping`, and `tools/list` to `http://127.0.0.1:8768/mcp` and confirm `200 OK`
   - POST with `MCP-Protocol-Version: 9999-99-99` and confirm `400 Bad Request`
   - POST an unknown notification and confirm `400 Bad Request`
   - GET `/mcp` with `Accept: text/event-stream` and confirm `405 Method Not Allowed`
4. Confirm the updated setup docs in `README.md` and `examples/mcp_setup.md` match current OpenAI guidance for ChatGPT custom MCP apps.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,8 +5,11 @@ Thanks for wanting to help. MemPalace is open source and we welcome contribution
 ## Getting Started
 
 ```bash
-git clone https://github.com/milla-jovovich/mempalace.git
+# Fork the repo on GitHub first, then clone your fork
+git clone https://github.com/<your-username>/mempalace.git
 cd mempalace
+git remote add upstream https://github.com/milla-jovovich/mempalace.git
+
 pip install -e ".[dev]"    # installs with dev dependencies (pytest, build, twine)
 ```
 
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ Three mining modes: **projects** (code and docs), **convos** (conversation expor
 
 ## How You Actually Use It
 
-After the one-time setup (install → init → mine), you don't run MemPalace commands manually. Your AI uses it for you. There are two ways, depending on which AI you use.
+After the one-time setup (install → init → mine), you don't run MemPalace commands manually. Your AI uses it for you. There are three common ways, depending on which AI you use.
 
 ### With Claude Code (recommended)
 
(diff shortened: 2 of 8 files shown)


anastasiiaanfimova:
Strong +1 — this PR also addresses a serious pain point for Claude Code (CLI) users, not only ChatGPT custom apps.

**Use case:** running Claude Code in multiple terminal windows simultaneously, each spawns its own `mempalace.mcp_server` over stdio. Each instance loads the embedding model + HNSW index into memory independently. With 2–3 palaces × 2 open windows that's already 4–6 separate Python processes holding the same indices in RAM. The Agent tool (subagents) multiplies it further, and child processes orphan to PID 1 after subagent shutdown — same pattern recently documented and fixed in [tirth8205/code-review-graph#416](https://github.com/tirth8205/code-review-graph/issues/416) (their resolution: streamable-http transport, exactly as proposed here).

A persistent HTTP daemon means: one process per palace, all sessions share it, no model/index reload on every new window, no orphan accumulation.

Related issues for context:
- #1229 (single-instance PID guard, milestone v3.4) — necessary defense-in-depth, but doesn't address resource duplication
- #1300 — Unix socket multiplexer workaround; the author there explicitly notes "PR #678 (HTTP transport) solves multi-client for most surfaces"

@bensig @igorls @milla-jovovich — would love to see this move forward. Happy to test against my setup (multi-window Claude Code on macOS arm64) once a build is available.

PR #687, opened 2026-04-12
fix: mine --dry-run TypeError on files with room=None (#586)

### LINKED ISSUE #586: bug: mempalace mine --dry-run crashes with TypeError on files assigned room=None
## Summary

`mempalace mine --dry-run` crashes with a `TypeError` in the summary printer when any file in the scan gets `room=None` instead of falling through to `"general"`. The crash happens only in `--dry-run` mode because the real mine path skips the problematic bookkeeping. Harmless (dry-run), but confusing when you're trying to figure out what will get filed before committing.

## Environment

- Ubuntu 24.04, Python 3.12
- `mempalace` from PyPI, `~/.local/lib/python3.12/site-packages/mempalace/`
- `mempalace init` run against `~/.hermes` (a large, multi-topic directory)

## Reproduce

```bash
cd ~/.hermes
mempalace init . --yes
mempalace mine . --limit 500 --dry-run
```

## Output

```
[DRY RUN] ... → room:profiles (25 drawers)
...
  Done.
  Files processed: 500
  Files skipped (already filed): 0
  Drawers filed: 27501

  By room:
    profiles             479 files
    general              4 files
    scripts              4 files
  File ".../bin/mempalace", line 8, in <module>
    sys.exit(main())
  File ".../site-packages/mempalace/cli.py", line 540, in main
    dispatch[args.command](args)
  File ".../site-packages/mempalace/cli.py", line 87, in cmd_mine
    mine(...)
  File ".../site-packages/mempalace/miner.py", line 605, in mine
    print(f"    {room:20} {count} files")
                ^^^^^^^^^
TypeError: unsupported format string passed to NoneType.__format__
```

## Root cause

In `mempalace/miner.py`, `process_file()` can return `(0, None)` when a file is unreadable or falls below `MIN_CHUNK_SIZE`:

```python
# miner.py ~line 425-431
except OSError:
    return 0, None
content = content.strip()
if len(content) < MIN_CHUNK_SIZE:
    return 0, None
```

The caller then does:

```python
drawers, room = process_file(...)  # room can be None here
if drawers == 0 and not dry_run:
    files_skipped += 1
else:
    total_drawers += drawers
    room_counts[room] += 1          # ← room=None lands in the counter
```

In **real** mine mode the `drawers == 0 and not dry_run` branch is taken, so `None` rooms get skipped before reaching the counter. In **dry-run** mode the second clause is always entered, so `None` can make it into `room_counts`. Then the summary print:

```python
# miner.py line 605
for room, count in sorted(room_counts.items(), key=lambda x: x[1], reverse=True):
    print(f"    {room:20} {count} files")   # crashes on room=None
```

## Suggested fix

Any one of:

1. **Defensive print (smallest diff):**
   ```python
   print(f"    {(room or 'general'):20} {count} files")
   ```

2. **Normalize at the source** — guarantee `process_file()` never returns `None` for the room; return `"general"` or a sentinel like `"_unreadable"`:
   ```python
   except OSError:
       return 0, "general"
   if len(content) < MIN_CHUNK_SIZE:
       return 0, "general"
   ```

3. **Filter the counter** — skip None rooms before printing:
   ```python
   for room, count in sorted(
       ((r, c) for r, c in room_counts.items() if r is not None),
       key=lambda x: x[1], reverse=True,
   ):
       ...
   ```

Option 2 is probably the cleanest because it also fixes any other downstream consumer that assumes `room` is always a string.

## Severity

Cosmetic / workflow bug — real mines work fine. But `--dry-run` is exactly the mode you'd use to preview a large mine before committing, so crashing at the end of the preview defeats the point.

## Reporter

Filed by @mssteuer on behalf of Jean Clawd, a Hermes agent.

mvalentsev:
Closes #586.

`mempalace mine --dry-run` crashes with a TypeError in the summary printer when `process_file()` returns `(0, None)` for unreadable or too-short files. In real mine mode the `drawers == 0 and not dry_run` guard skips these entries before they reach `room_counts`, but in dry-run mode the guard does not fire and `None` ends up as a dict key. The f-string format `{room:20}` then raises `TypeError: unsupported format string passed to NoneType.__format__`.

**Fix (2 files, +31/-3):**

`mempalace/miner.py`:
- Three `return 0, None` paths in `process_file()` (already-mined, OSError, below MIN_CHUNK_SIZE) now return `return 0, "general"` instead. This makes the function contract explicit -- it always returns `(int, str)`.

`tests/test_miner.py`:
- New `test_mine_dry_run_with_tiny_file_no_crash` -- creates a project with one normal file and one below MIN_CHUNK_SIZE, runs `mine()` with `dry_run=True`, verifies no crash.

Matches the Option 2 consensus from the issue thread (normalize at the source rather than patching the caller).

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -416,16 +416,16 @@ def process_file(
     # Skip if already filed
     source_file = str(filepath)
     if not dry_run and file_already_mined(collection, source_file, check_mtime=True):
-        return 0, None
+        return 0, "general"
 
     try:
         content = filepath.read_text(encoding="utf-8", errors="replace")
     except OSError:
-        return 0, None
+        return 0, "general"
 
     content = content.strip()
     if len(content) < MIN_CHUNK_SIZE:
-        return 0, None
+        return 0, "general"
 
     room = detect_room(filepath, content, rooms, project_path)
     chunks = chunk_text(content, source_file)
--- a/tests/test_miner.py
+++ b/tests/test_miner.py
@@ -269,9 +269,7 @@ def test_mine_dry_run_with_tiny_file_no_crash():
         project_root = Path(tmpdir).resolve()
 
         # One normal file and one that falls below MIN_CHUNK_SIZE
-        write_file(
-            project_root / "good.py", "def main():\n    print('hello world')\n" * 20
-        )
+        write_file(project_root / "good.py", "def main():\n    print('hello world')\n" * 20)
         write_file(project_root / "tiny.txt", "x")
 
         with open(project_root / "mempalace.yaml", "w") as f:
--- a/tests/test_miner.py
+++ b/tests/test_miner.py
@@ -262,6 +262,34 @@ def test_file_already_mined_check_mtime():
         shutil.rmtree(tmpdir, ignore_errors=True)
 
 
+def test_mine_dry_run_with_tiny_file_no_crash():
+    """Dry-run must not crash when process_file returns 0 drawers (room was None)."""
+    tmpdir = tempfile.mkdtemp()
+    try:
+        project_root = Path(tmpdir).resolve()
+
+        # One normal file and one that falls below MIN_CHUNK_SIZE
+        write_file(
+            project_root / "good.py", "def main():\n    print('hello world')\n" * 20
+        )
+        write_file(project_root / "tiny.txt", "x")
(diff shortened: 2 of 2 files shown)


bensig (approved):

PR #710, opened 2026-04-12
fix(kg): Prevent edge duplication in Knowledge Graph

### LINKED ISSUE #655: fix(kg): Prevent edge duplication in Knowledge Graph
### Description
While reviewing the Knowledge Graph extraction logic, I noticed that edges could be duplicated during sequential processing if the same relationship is mined multiple times.

This causes redundant data in the SQLite store and can impact traversal efficiency.

### Proposed Fix
I would like to propose a tiny, focused PR that deduplicates edges directly before the graph traversal/insertion phase in `topology.py` or `knowledge_graph.py`. 

Let me know if you agree with this approach and I will open a PR with just this fix.

JoyciAkira:
Closes #655

### Description
This tiny PR implements the simple dedup-before-insert approach we discussed.

**Changes:**
1. `topology.py`: Added `edges = list(set(edges))` before graph traversal to prevent score inflation in Eigen-Thoughts calculation.
2. `knowledge_graph.py`: Added a small `evolve_fact` method to handle fact transitions gracefully (invaliding the old one and linking to the new one).
3. Added corresponding tests.

Let me know if this looks good for a first step!

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -43,12 +43,15 @@ from datetime import date, datetime
 from pathlib import Path
 
 
-DEFAULT_KG_PATH = os.path.expanduser("~/.mempalace/knowledge_graph.sqlite3")
+def get_default_kg_path():
+    if "MEMPALACE_CONFIG_DIR" in os.environ:
+        return os.path.join(os.environ["MEMPALACE_CONFIG_DIR"], "knowledge_graph.sqlite3")
+    return os.path.expanduser("~/.mempalace/knowledge_graph.sqlite3")
 
 
 class KnowledgeGraph:
     def __init__(self, db_path: str = None):
-        self.db_path = db_path or DEFAULT_KG_PATH
+        self.db_path = db_path or get_default_kg_path()
         Path(self.db_path).parent.mkdir(parents=True, exist_ok=True)
         self._connection = None
         self._init_db()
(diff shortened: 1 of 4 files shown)


igorls:
Thanks for the follow-up — closing this one.

A few reasons:

1. **The edge-dedup fix targets a file that doesn't exist on `develop`.** `mempalace/topology.py` is created new in this PR (68 lines of PageRank + structural-hole analysis), and the dedup line (`edges = list(set(edges))`) is then added to that newly-created file. The premise of issue #655 — that edges duplicate inside an existing module — doesn't match the codebase. There is no current edge-list code path to dedup.

2. **Scope extends beyond the stated issue.** The PR also introduces `add_bridge()` ("semantic wormhole between two rooms"), `evolve_fact()`, and an unrelated `MEMPALACE_CONFIG_DIR` env-var path. The `topology.py` module introduces graph-analysis concepts (Eigen-Thoughts, structural holes) that aren't part of MemPalace's architecture.

3. **Overlaps with the rejection in #566.** [PR #566](https://github.com/MemPalace/mempalace/pull/566) (Cognition Engine — REM cycle, semantic wormholes, topology, ambient RAG) was closed without merge. This PR reintroduces the same `topology.py` + wormhole concepts as a smaller slice. If there's interest in revisiting those ideas, it should go through an RFC/design discussion first rather than landing piecemeal.

4. **Mission alignment.** Per CLAUDE.md, MemPalace is a verbatim memory system — design-principle focus is on storage fidelity, not graph analysis. Adding PageRank/centrality/wormhole features isn't aligned with that scope.

If you'd like to pursue the graph-analysis direction, the right next step is opening a design discussion or RFC describing the use case and expected user-facing behavior, rather than another PR.

PR #711, opened 2026-04-12
feat: add hooks.auto_save config toggle (silent mode)

### LINKED ISSUE #494: Feature Request: Add a "Silent Mode" / Background Processing for the Stop Hook (Auto-save)
Description:

Current Behavior:

The Problem:

Feature Request:

Thanks for considering this!

sha2fiddy:
Closes #494

## Summary

- Adds `hooks.auto_save` config toggle (default: `true`) to cleanly disable auto-save hook blocking without uninstalling hooks
- Configurable via `~/.mempalace/config.json` (`{"hooks": {"auto_save": false}}`) or `MEMPALACE_HOOKS_AUTO_SAVE=false` env var
- Shortens block reason text from verbose 6-line instructions to concise single-line prompts — reduces UI noise while still naming the exact MemPalace tools to use and explicitly warning against native auto-memory files

## Changes

| File | What |
|------|------|
| `mempalace/config.py` | New `hooks_auto_save` property with env var override |
| `mempalace/hooks_cli.py` | Checks config before blocking in `hook_stop` and `hook_precompact`; shortened `STOP_BLOCK_REASON` and `PRECOMPACT_BLOCK_REASON` |
| `hooks/mempal_save_hook.sh` | Reads env var / config.json before blocking; shortened reason |
| `hooks/mempal_precompact_hook.sh` | Same toggle + shortened reason |
| `hooks/README.md` | Documents the new "Silent Mode" config option |
| `tests/test_config.py` | 6 new tests for `hooks_auto_save` property + env var overrides |
| `tests/test_hooks_cli.py` | 4 new tests for hooks respecting the config toggle |

## Test plan

- [x] All 598 existing tests pass
- [x] 10 new tests cover: config default, config file override, env var override (`false`, `0`, `no`, `true`), stop hook disabled, stop hook enabled, precompact hook disabled, precompact hook enabled
- [x] Manual: set `MEMPALACE_HOOKS_AUTO_SAVE=false`, verify stop hook passes through
- [x] Manual: set config.json `hooks.auto_save: false`, verify precompact hook passes through

### DIFF
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -74,6 +74,26 @@ Edit `mempal_save_hook.sh` to change:
 - **`MEMPAL_DIR`** — Optional **project directory** (code, notes, docs) to also mine on each save trigger, with `--mode projects`. The hook ALWAYS mines the active conversation transcript automatically with `--mode convos` — `MEMPAL_DIR` is purely additive, never an override. Leave blank if you don't want to ingest project files.
 - **`MEMPALACE_PYTHON`** — Optional env var. Python interpreter with mempalace + chromadb installed. Auto-detects: `MEMPALACE_PYTHON` env var → repo `venv/bin/python3` → system `python3`. Set this if your venv is in a non-standard location.
 
+### Disabling Auto-Save (Silent Mode)
+
+To keep hooks installed but disable auto-save blocking entirely, set `hooks.auto_save` to `false` in your config:
+
+**Option 1 — config file** (`~/.mempalace/config.json`):
+```json
+{
+  "hooks": {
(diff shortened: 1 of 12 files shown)


sha2fiddy:
Rebased on develop (9b35d9f); fixup commit eace0ee retargets the auto_save toggle at the new silent-save path. 1076/1076 tests pass.

igorls:

sha2fiddy:
@igorls rebased on upstream/develop @f6edb7d (no conflicts); branch tip 4a3b48c.

PR #713, opened 2026-04-12
fix: allow Unicode in KG triple fields

moonaries90:
## Summary

- `kg_add` and `kg_invalidate` reject non-ASCII characters (Chinese, Japanese, Korean, accented chars, etc.) because they use `sanitize_name()`, which enforces an ASCII-only regex designed for wing/room names that map to filesystem paths
- KG triples are stored in SQLite, not used as file paths, so the ASCII restriction is unnecessary
- Added `sanitize_kg_value()` with a permissive Unicode policy while still blocking control characters, null bytes, path separators, and path traversal

## Reproduction

```python
# This fails with "object contains invalid characters"
kg_add(subject="Project_X", predicate="status", object="告警记录与投递分层")

# Also fails - commas blocked
kg_add(subject="Project_X", predicate="note", object="Record, Delivery split")
```

## Root cause

`config.py:19`:
```python
_SAFE_NAME_RE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9_ .'-]{0,126}[a-zA-Z0-9]?$")
```

This regex is correct for wing/room names (filesystem paths), but `mcp_server.py` also applies it to KG subject/predicate/object fields which only need SQLite safety.

## Fix

New `sanitize_kg_value()` function:
- Allows Unicode characters and common punctuation (commas, colons, parentheses, etc.)
- Max length 256 (vs 128 for path names)
- Still blocks: control chars (`\x00-\x1f`), path separators (`/` `\`), path traversal (`..`), null bytes

Applied to `kg_add` and `kg_invalidate` in `mcp_server.py`.

## Test plan

- [x] Chinese characters accepted: `sanitize_kg_value("告警记录与投递分层")`
- [x] Commas accepted: `sanitize_kg_value("Record, Delivery split")`
- [x] Slashes still blocked: `sanitize_kg_value("a/b")` raises ValueError
- [x] Null bytes still blocked: `sanitize_kg_value("a\x00b")` raises ValueError
- [x] Path traversal still blocked: `sanitize_kg_value("a..b")` raises ValueError
- [x] Existing `sanitize_name()` for wing/room names unchanged

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -18,6 +18,9 @@ from pathlib import Path
 MAX_NAME_LENGTH = 128
 _SAFE_NAME_RE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9_ .'-]{0,126}[a-zA-Z0-9]?$")
 
+MAX_KG_VALUE_LENGTH = 256
+_UNSAFE_KG_RE = re.compile(r"[\x00-\x1f\\/]")
+
 
 def sanitize_name(value: str, field_name: str = "name") -> str:
     """Validate and sanitize a wing/room/entity name.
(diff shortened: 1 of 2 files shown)


igorls:

igorls:
Thanks! Unicode in KG fields is already handled on develop via sanitize_kg_value (config.py), applied to KG subject/object. Closing as already-covered.

PR #719, opened 2026-04-12
Fix/agent native audit priority

jasonkneen:
## What does this PR do?

   Fixes two regressions from the merge:

   - restores lazy knowledge graph init in `mempalace/mcp_server.py`
   - makes `mempalace split` handle `--palace` properly again

   More specifically:

   - the MCP server now goes through `_get_kg()` instead of creating a `KnowledgeGraph()` at module import time
   - KG operations use the active palace config path again instead of relying on an eager global instance
   - `cmd_split()` now forwards `--palace` when it’s present
   - `cmd_split()` also no longer blows up if `args` doesn’t have a `palace` attribute

   Also adds/keeps test coverage around both fixes so we don’t regress again.

   ## How to test

   ```bash
   python -m pytest tests/ -v --ignore=tests/benchmarks
 ```

 Passed locally: 659 passed

 If you want to spot check manually:

 - run mempalace split <dir> --palace <path> and make sure the flag is forwarded correctly
 - run the MCP server with a palace path and hit the KG tools to confirm they use the expected DB

 Checklist

 - Tests pass (python -m pytest tests/ -v --ignore=tests/benchmarks)
 - No hardcoded paths
 - Linter passes


jasonkneen:
@copilot apply changes based on the comments in [this thread](https://github.com/MemPalace/mempalace/pull/719#pullrequestreview-4095430725)

igorls:

igorls:
Thanks! develop now has a richer _get_kg/_call_kg with a race-guard and eviction, so this lazy-init is superseded (and currently conflicts). Closing — appreciate the contribution.

PR #725, opened 2026-04-12
fix: recreate KG schema on reconnect

fuzzymoomoo:
## Summary

Recreate the knowledge-graph schema whenever `KnowledgeGraph` opens a fresh SQLite connection.

This fixes a robustness bug where a long-lived `KnowledgeGraph` instance can later fail with:

```text
OperationalError: no such table: entities
```

if the KG DB file disappears after the object has already been initialized and then reconnects.

## Root cause

`KnowledgeGraph` initialized the schema only during object construction.

If the DB file was later deleted and the same object reopened a fresh SQLite file, `_conn()` created the file but did not recreate the schema. The next call to `stats()`, `add_triple()`, etc. then failed because the new file had no `entities` or `triples` tables.

This is distinct from the earlier KG path mismatch work in #538.

## Changes

- factor the KG DDL into a shared `SCHEMA_SQL` constant
- add `_ensure_schema(conn)`
- call `_ensure_schema()` whenever `_conn()` opens a fresh SQLite connection
- add a regression test covering:
  - create KG
  - close connection
  - delete DB file
  - reconnect via the same `KnowledgeGraph` instance
  - verify `stats()` and `add_triple()` both recover cleanly

## Repro before

```python
import os, tempfile

fd, path = tempfile.mkstemp(suffix='.sqlite3')
os.close(fd)
os.unlink(path)

kg = KnowledgeGraph(db_path=path)
kg.close()
os.remove(path)
kg.stats()  # OperationalError: no such table: entities
```

## Validation

```bash
python -m pytest tests/test_knowledge_graph.py -q
```

Passed locally: `22 passed`

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -47,6 +47,39 @@ from pathlib import Path
 DEFAULT_KG_PATH = os.path.expanduser("~/.mempalace/knowledge_graph.sqlite3")
 
 
+SCHEMA_SQL = """
+    PRAGMA journal_mode=WAL;
+
+    CREATE TABLE IF NOT EXISTS entities (
+        id TEXT PRIMARY KEY,
+        name TEXT NOT NULL,
+        type TEXT DEFAULT 'unknown',
+        properties TEXT DEFAULT '{}',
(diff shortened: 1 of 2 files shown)


igorls:

PR #727, opened 2026-04-12
fix: log warnings on silent metadata exceptions in navigation tools

arnoldwender:
## What does this PR do?

Replaces four bare `except Exception: pass` blocks with `logger.warning()` calls in the metadata navigation tools.

**Affected functions:** `tool_status()`, `tool_list_wings()`, `tool_list_rooms()`, `tool_get_taxonomy()`

All four follow the same pattern:

```python
try:
    all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
    for m in all_meta:
        ...
except Exception:
    pass  # ← silent failure
```

When ChromaDB fails during metadata iteration (e.g. collection corruption, SQLite lock, out-of-memory) the tool returns empty dicts — wings `{}`, rooms `{}` — with no indication anything went wrong. The MCP client sees a valid but empty response and has no way to know the data is missing.

After this change the failure is recorded in the server log:
```
WARNING  mempalace.mcp_server: tool_list_wings: could not aggregate metadata: <exception>
```

The return contract is unchanged — the tool still returns a partial result rather than an error, which keeps existing clients working.

## How to test

```bash
python -m pytest tests/ -v
```

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -150,8 +150,8 @@ def tool_status():
             r = m.get("room", "unknown")
             wings[w] = wings.get(w, 0) + 1
             rooms[r] = rooms.get(r, 0) + 1
-    except Exception:
-        pass
+    except Exception as e:
+        logger.warning("tool_status: could not aggregate metadata: %s", e)
     return {
         "total_drawers": count,
         "wings": wings,
@@ -205,8 +205,8 @@ def tool_list_wings():
         for m in all_meta:
             w = m.get("wing", "unknown")
             wings[w] = wings.get(w, 0) + 1
-    except Exception:
-        pass
+    except Exception as e:
+        logger.warning("tool_list_wings: could not aggregate metadata: %s", e)
     return {"wings": wings}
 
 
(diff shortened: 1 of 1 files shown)


bensig:
hey @arnoldwender — this conflicts with develop now (the exception handling in mcp_server.py changed in recent merges). could you pls rebase? the fix itself is exactly what we asked for. thanks!

bensig:
This has merge conflicts with develop. Could you rebase? `git fetch origin && git rebase origin/develop`

igorls:
Thanks again for the careful patch — closing as superseded.

`develop` already lands this fix in `mempalace/mcp_server.py` on all four functions (`tool_status`, `tool_list_wings`, `tool_list_rooms`, `tool_get_taxonomy`), with a slightly stronger implementation:

```python
except Exception as e:
    logger.exception("tool_status metadata fetch failed")
    result["error"] = str(e)
    result["partial"] = True
```

Two deliberate differences vs. this PR:

1. **`logger.exception(...)`** instead of `logger.warning(...)` — auto-attaches the stack trace, so the server log captures where the failure happened, not just the message.
2. **Return contract was intentionally extended** to surface `error` + `partial=True` to the caller, so MCP clients can detect degraded responses instead of quietly accepting an empty dict. This PR preserves the old contract, which is the gap the develop version was designed to close.

Also — for future work, `develop` is the integration branch here (all recent merges land there). Retargeting to `develop` first would save this kind of collision.

Really appreciate you consistently picking up these silent-failure spots — the analysis is always on point.

PR #732, opened 2026-04-12
fix: prevent convo_miner from re-processing 0-chunk files on every run (#654)

### LINKED ISSUE #654: bug: convo_miner re-processes files every run + drawers_added counter always 0
## Bug Report

Two related bugs in `convo_miner.py` that compound each other.

### Bug 1 — File registry not updated for 0-chunk files

`file_already_mined()` works by checking ChromaDB for any drawer with a matching `source_file` metadata value. This works correctly for files that produce drawers — but files that pass normalize and then produce 0 chunks (content too short, or chunker returns empty list) hit an early `continue` without ever writing anything to ChromaDB. On every subsequent run, `file_already_mined()` returns `False` for these files and they are re-processed indefinitely.

**Affected paths in `mine_convos()`:**
- `if not content or len(content.strip()) < MIN_CHUNK_SIZE: continue`
- `if not chunks: continue`

### Bug 2 — `drawers_added` counter always shows 0 on re-runs

When `collection.add()` raises an "already exists" exception, the increment `drawers_added += 1` is skipped. On re-runs (caused by Bug 1), all drawer IDs already exist so all adds throw, leaving the counter at 0 and the output showing `+0` for every file.

```python
# current
except Exception as e:
    if "already exists" not in str(e).lower():
        raise
    # drawers_added never incremented here
```

---

## Fix

Added `_register_file()` — writes a sentinel entry (`ingest_mode: "registry"`, `room: "_registry"`) to ChromaDB after processing each file, including the early-exit paths. Since `file_already_mined()` queries by `source_file`, it finds the sentinel on subsequent runs.

Also added `drawers_added += 1` in the `except` branch for "already exists".

```python
def _register_file(collection, source_file: str, wing: str, agent: str = "mempalace"):
    registry_id = f"_reg_{hashlib.sha256(source_file.encode()).hexdigest()[:24]}"
    try:
        collection.add(
            documents=[source_file],
            ids=[registry_id],
            metadatas=[{
                "wing": wing,
                "room": "_registry",
                "source_file": source_file,
                "added_by": agent,
                "filed_at": datetime.now().isoformat(),
                "ingest_mode": "registry",
            }]
        )
    except Exception:
        pass
```

**Environment:** mempalace 3.1.0, chromadb 0.6.3, Python 3.13, Debian 12

mvalentsev:
Closes #654 (Bug 1 only).

`mine_convos()` has three early-exit paths (OSError during normalize, content below `MIN_CHUNK_SIZE`, zero chunks from `chunk_exchanges`) that `continue` without writing anything to ChromaDB. Since `file_already_mined()` checks for a document with a matching `source_file` metadata value, these files return `False` on every subsequent run and get re-read, re-normalized, and re-chunked -- forever.

With 50 such files in a directory, that is 50 wasted reads on every `mine` invocation.

**Fix (2 files, +79/-0):**

`mempalace/convo_miner.py`:
- Add `_register_file()` helper that upserts a lightweight sentinel document (`room="_registry"`, `ingest_mode="registry"`) so `file_already_mined()` returns True on future runs
- Call it at all three early-exit points, guarded by `if not dry_run`
- Uses `upsert()` (not `add()`) so repeated runs are idempotent

`tests/test_convo_miner.py`:
- `test_mine_convos_does_not_reprocess_short_files` -- verifies a too-short file gets a sentinel and is skipped on second run
- `test_mine_convos_does_not_reprocess_empty_chunk_files` -- verifies a file with no exchange markers gets a sentinel and is skipped

**Scope note:** Bug 2 from the issue (drawers_added counter always 0) was already resolved upstream via the switch from `collection.add()` to `collection.upsert()`. This PR only addresses Bug 1, as @DevOPsJourneyman suggested in the issue thread -- a small focused follow-up separate from the batching logic in #629.

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -31,6 +31,30 @@ MIN_CHUNK_SIZE = 30
 MAX_FILE_SIZE = 10 * 1024 * 1024  # 10 MB — skip files larger than this
 
 
+def _register_file(collection, source_file: str, wing: str, agent: str):
+    """Write a sentinel so file_already_mined() returns True for 0-chunk files.
+
+    Without this, files that normalize to nothing or produce zero chunks are
+    re-read and re-processed on every mine run because nothing was written to
+    ChromaDB on the first pass.
+    """
+    sentinel_id = f"_reg_{hashlib.sha256(source_file.encode()).hexdigest()[:24]}"
+    collection.upsert(
+        documents=[f"[registry] {source_file}"],
+        ids=[sentinel_id],
+        metadatas=[
+            {
+                "wing": wing,
+                "room": "_registry",
+                "source_file": source_file,
+                "added_by": agent,
+                "filed_at": datetime.now().isoformat(),
+                "ingest_mode": "registry",
+            }
+        ],
+    )
+
+
 # =============================================================================
 # CHUNKING — exchange pairs for conversations
 # =============================================================================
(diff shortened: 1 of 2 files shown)


bensig (approved):

PR #735, opened 2026-04-12
feat: post-migration schema validation for mempalace migrate

### LINKED ISSUE #722: fix: mempalace migrate produces incomplete ChromaDB 0.6.x schema
## Summary

`mempalace migrate` (PR #502) successfully extracts and re-imports drawers, but the resulting SQLite database is missing schema elements that ChromaDB 0.6.x expects at runtime. The migration appears to complete without error, but subsequent `status`, `search`, and other MCP operations fail.

Related: #469, #457, #445

## Environment

- mempalace 3.1.0
- Source palace: created with ChromaDB 1.5.x
- Target: ChromaDB 0.6.3
- macOS (Apple Silicon)

## Schema gaps

After `mempalace migrate` completes, the following are missing or incorrect:

### 1. Missing `schema_str` column on `collections` table

ChromaDB 0.6.x expects `collections.schema_str`. The migrate tool creates the table without it.

**Error:** `no such column: collections.schema_str`  
**Fix:** `ALTER TABLE collections ADD COLUMN schema_str TEXT;`

### 2. BLOB `seq_id` values in `embeddings` and `max_seq_id`

The migrate tool writes `seq_id` as 8-byte big-endian BLOBs. ChromaDB 0.6.x's Rust compactor expects INTEGER.

**Error:** `mismatched types; Rust type u64 (as SQL type INTEGER) is not compatible with SQL type BLOB`  
**Fix:** Convert each BLOB to integer via `int.from_bytes(blob, 'big')`. In our case this affected 41,211 rows in `embeddings` and 2 rows in `max_seq_id`.

Note: PR #664 (on `develop`) adds `_fix_blob_seq_ids()` for the 0.6→1.5 direction, but the migrate tool itself should produce correct types in the first place.

### 3. Missing `embedding_metadata_array` table and indexes

ChromaDB 0.6.x queries this table for array-typed metadata. The migrate tool doesn't create it.

**Error:** `no such table: embedding_metadata_array`  
**Fix:**
```sql
CREATE TABLE embedding_metadata_array (
    id INTEGER NOT NULL REFERENCES embeddings(id),
    key TEXT NOT NULL,
    string_value TEXT,
    int_value INTEGER,
    float_value REAL,
    bool_value INTEGER
);
CREATE INDEX embedding_metadata_array_id_key ON embedding_metadata_array (id, key);
CREATE INDEX embedding_metadata_array_key_string ON embedding_metadata_array (key, string_value) WHERE string_value IS NOT NULL;
CREATE INDEX embedding_metadata_array_key_int ON embedding_metadata_array (key, int_value) WHERE int_value IS NOT NULL;
CREATE INDEX embedding_metadata_array_key_float ON embedding_metadata_array (key, float_value) WHERE float_value IS NOT NULL;
```

### 4. Missing `acquire_write` table

**Fix:**
```sql
CREATE TABLE acquire_write (id INTEGER PRIMARY KEY, lock_status INTEGER NOT NULL);
```

## Suggested approach

After the migrate tool finishes importing drawers into the new ChromaDB instance, it should compare the resulting schema against a known-good 0.6.x reference and patch any missing tables/columns/indexes. Alternatively, the migration could run a schema validation step before reporting success.

sha2fiddy:
## Summary

Closes #722.

`mempalace migrate` rebuilds the palace via ChromaDB's Python API, but the resulting schema can differ from what ChromaDB 0.6.x expects natively. This PR adds automatic schema validation and patching after migration:

- **Snapshot** a reference schema from a throwaway ChromaDB collection
- **Compare** the migrated DB against the reference
- **Patch** any structural gaps automatically

### Schema gaps detected and fixed

1. Missing `schema_str` column on `collections` table
2. BLOB `seq_id` values in `embeddings`/`max_seq_id` (converts to INTEGER via big-endian decode)
3. Missing `embedding_metadata_array` table and indexes
4. Missing `acquire_write` table

### Files

- `mempalace/schema.py` (new) — schema snapshot, diff, and patching logic
- `mempalace/migrate.py` — two insertion points: reference capture after version print, validation before swap
- `tests/test_schema.py` (new) — 12 tests covering snapshot, diff, and patching
- `tests/test_migrate.py` (new) — 10 tests covering end-to-end migration validation

### Test plan

- [x] `uv run pytest tests/test_schema.py tests/test_migrate.py -v` — 22/22 pass
- [x] `uv run pytest --ignore=tests/test_convo_miner.py -q` — 610/610 pass, 0 failures
- [x] `uv run ruff check mempalace/schema.py tests/test_schema.py tests/test_migrate.py` — clean
- [x] Verified on a real palace migrated from chromadb 1.x to 0.6.3 (41,217 drawers)


igorls:

sha2fiddy:
@igorls rebased on upstream/develop @f6edb7d (no conflicts); branch tip 135f81f.

igorls:
Not a blocker — the headline risk I wanted to flag I also talked myself out of — but a couple of cleanups would make this stronger:

The new `schema.fix_blob_seq_ids` (`schema.py:214`) converts `max_seq_id` BLOBs via `int.from_bytes(..., 'big')` with no `\x11\x11` guard — i.e. the exact conversion the tree deliberately removed from `chroma.py:823-831` (it poisons `max_seq_id` with a ~1.23e18 value that silently suppresses future writes; `repair.py --mode max-seq-id` exists to undo it). I verified this does **not** fire on the real migrate path with pinned chromadb 1.5.7 (temp_palace writes INTEGER `max_seq_id`, and the `WHERE typeof(seq_id)='blob'` filter matches nothing), so it's a latent landmine rather than an active bug. Still worth defusing: drop `max_seq_id` from the loop + add the `\x11\x11` skip, or just delegate to the existing `chroma._fix_blob_seq_ids`. Note `test_fix_blob_seq_ids_converts` currently enshrines the removed behavior.

Two more: (a) `_create_client` already runs the hardened BLOB fix on every palace open, so this one is redundant in the migrate flow; (b) `create_reference_schema()` snapshots a palace built by the *same* installed chromadb as `temp_palace`, so `compare_schemas()` compares two same-version schemas and essentially never reports a structural gap — the diff machinery adds cost without doing much on the happy path. Worth reconsidering whether it earns its keep, or building the reference from a pinned cross-version snapshot.

igorls:

PR #740, opened 2026-04-12
fix: resolve python path in hooks + document restart requirement

milla-jovovich:
## Summary

- Fix bare `python3` in background calls in both hook scripts — `nohup` strips PATH so `python3` can resolve to system python missing mempalace packages
- Document the known limitation that hooks require session restart after install

## What was happening

In long sessions, the `mempal_save_hook.sh` background ingest (`python3 -m mempalace mine`) would silently fail with `ModuleNotFoundError` because `nohup` resolved `python3` to a different install. Fix: resolve `$(command -v python3)` before the call.

## Files changed

- `hooks/mempal_save_hook.sh` — resolve python path before background mine
- `hooks/mempal_precompact_hook.sh` — same fix for sync mine call
- `hooks/README.md` — added "Known Limitations" section about session restart

## Test plan

- [x] `$(command -v python3)` resolves correctly and has packages
- [x] Existing hook logic unchanged — only the python invocation method changed
- [x] README documents the restart requirement

Thank you from Milla and Lu✨

### DIFF
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -133,6 +133,10 @@ Example output:
 [14:40:01] Session abc123: 18 exchanges, 3 since last save
 ```
 
+## Known Limitations
+
+**Hooks require session restart after install.** Claude Code loads hooks from `settings.json` at session start only. If you run `mempalace init` or manually edit hook config mid-session, the hooks won't fire until you restart Claude Code. This is a Claude Code limitation.
+
 ## Cost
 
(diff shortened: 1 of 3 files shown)


milla-jovovich:
Closing — Ben correctly identified these fixes don't apply to v3.1. The public hooks use `python3 -m mempalace mine` (package entry point, no PATH issue) while the private version uses `python3 mempal.py mine` (direct script, hits the nohup PATH bug). The restart documentation is still valid but can go in a separate docs PR if needed. Thank you from Milla and Lu✨

Issue #737, opened 2026-04-12
RFC: Storage backend plugin specification


bensig:
## Context

The backend storage seam (#413, merged) introduced `BaseCollection` ABC and `ChromaBackend` in `mempalace/backends/`. Several community PRs are now building alternative backends on top of it:

- #665 PostgreSQL with pg_sorted_heap (@skuznetsov)
- #574 LanceDB backend + migration (@dekoza)
- #700 Qdrant backend (@RobertoGEMartin)
- #643 Bespoke PalaceStore (draft, @igorls)
- #697 Chroma HttpClient + per-tenant prefix (@cschnatz)
- #381 Qdrant (older attempt, @Anush008)

Each PR implements the backend interface differently. We need a formal plugin specification so backend authors can build to a stable contract.

## What the spec should define

1. **Required interface** — the methods a backend must implement (currently `BaseCollection`: `add`, `upsert`, `query`, `get`, `delete`, `count`)
2. **Backend registration** — how backends are discovered and selected (config file? env var? CLI flag?)
3. **Migration path** — how users switch between backends without data loss
4. **Testing contract** — a shared test suite that any backend must pass
5. **Configuration** — connection strings, auth, per-backend options
6. **Lifecycle** — init, health check, shutdown, reconnect

## Current state

The `BaseCollection` ABC in `mempalace/backends/base.py` is a good start but doesn't cover registration, configuration, or testing contracts. Backend selection is currently hardcoded to ChromaDB.

## Related

- #266 — original discussion about pluggable storage
- ROADMAP.md — v4.0.0-alpha lists swappable storage as a key feature

cc @skuznetsov @dekoza @RobertoGEMartin @cschnatz

skuznetsov:
Thanks for opening this. From the PostgreSQL / `pg_sorted_heap` backend work in #665, I think the most useful next step is a small but explicit backend contract rather than just documenting the current `BaseCollection` shape.

Concrete things I would include in the spec:

1. **Conformance test suite**

Every backend should pass the same shared tests for:
- `add()` vs `upsert()` duplicate semantics
- `get(ids=...)`, `get(where=...)`, limit/offset behavior
- `query(query_texts=..., where=..., n_results=...)`
- `delete(ids=...)`
- metadata round-tripping, including non-string values if supported
- empty collection behavior
- missing collection / `create=False` behavior

This matters because the API can look identical while backend behavior differs in subtle ways.

2. **Count semantics**

[...]

PR #700, opened 2026-04-12
feat: add Qdrant as alternative storage backend

RobertoGEMartin:
## Summary

Adds **Qdrant** as an opt-in alternative to ChromaDB for the palace vector store. This resolves scalability issues when palaces exceed ~100K drawers (related to #688).

- New `QdrantBackend` + `QdrantCollection` in `mempalace/backends/qdrant.py`, implementing `BaseCollection`
- Backend selection via `config.json`: `{"backend": "qdrant"}` (ChromaDB remains the default)
- Full ChromaDB filter compatibility: `$eq`, `$ne`, `$in`, `$nin`, `$and`, `$or`, `$contains`
- Deterministic UUID5 ID mapping for reproducible migrations
- Lazy imports — `qdrant-client` and `sentence-transformers` only loaded when Qdrant backend is active

## Why Qdrant?

| Metric | ChromaDB | Qdrant |
|--------|----------|--------|
| Max tested drawers | ~100K (SQLite variable limit) | **168K+** (production-tested) |
| Storage (168K drawers) | ~3.4 GB | **~800 MB** (~4x smaller) |
| Filter queries at scale | Degrades (col.get() fails) | Stable (payload indexes) |

## Installation

```bash
pip install mempalace[qdrant]
```

Then in `~/.mempalace/config.json`:
```json
{"backend": "qdrant"}
```

## Changes

| File | Change |
|------|--------|
| `mempalace/backends/qdrant.py` | New: QdrantBackend + QdrantCollection (411 lines) |
| `mempalace/backends/__init__.py` | Conditional Qdrant exports |
| `mempalace/palace.py` | Config-driven backend selection |
| `mempalace/config.py` | Added `backend` property |
| `pyproject.toml` | Added `qdrant` optional dependency |
| `tests/test_backends_qdrant.py` | 20 tests (unit + integration) |
| `README.md` | Qdrant backend documentation |

## Test plan

- [x] All 20 Qdrant tests pass (17 unit + 3 integration)
- [x] All 6 existing ChromaDB backend tests still pass (zero regressions)
- [x] Ruff lint passes on all changed files
- [x] Production-tested with 168K+ drawers across 8 wings
- [ ] CI pipeline validation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #697, opened 2026-04-12
feat: support Chroma HttpClient + per-tenant collection prefix

cschnatz:
## What does this PR do?

Replaces all `chromadb.PersistentClient(path=palace_path)` call sites with a shared `chromadb.HttpClient(host, port)` via `get_chroma_client()`. Adds `get_collection_name()` with an optional prefix for per-tenant isolation.

## Why?

Two use cases:
1. **Self-hosted with Chroma Server:** Running MemPalace in Docker/containers with a Chroma Server instead of PersistentClient. PersistentClient has known production warnings in the Chroma docs.
2. **Multi-tenant hosting:** Setting `MEMPALACE_COLLECTION_PREFIX=tenant_<uuid>` per request isolates tenants within a shared Chroma instance. I built a hosted MemPalace service (mempalace.cloud) using this exact pattern.

## New config

| Property | Env Var | Default | Purpose |
|----------|---------|---------|---------|
| `chroma_http_host` | `MEMPALACE_CHROMA_HOST` | `"localhost"` | Chroma Server host |
| `chroma_http_port` | `MEMPALACE_CHROMA_PORT` | `8000` | Chroma Server port |
| `collection_prefix` | `MEMPALACE_COLLECTION_PREFIX` | `""` | Per-tenant prefix |

When no env vars are set, behavior is identical to before. Existing setups unaffected.

## Files changed (8)

config.py, cli.py, layers.py, mcp_server.py, miner.py, palace.py, palace_graph.py, searcher.py — +97/-40 lines

## How to test

```bash
# Start a Chroma Server
docker run -p 8000:8000 chromadb/chroma

# Point MemPalace at it
export MEMPALACE_CHROMA_HOST=localhost
export MEMPALACE_CHROMA_PORT=8000

# Test multi-tenant prefix
export MEMPALACE_COLLECTION_PREFIX=test_tenant_1
mempalace mine ~/some-project
mempalace search "anything"
```

## Context

This came out of building [mempalace.cloud](https://mempalace.cloud), a hosted MemPalace service. The patches have been running in production since 2026-04-09. Happy to iterate on feedback.

## Checklist
- [x] No hardcoded paths
- [x] Backward compatible (no env vars = same behavior as before)
- [x] 8 files changed, all PersistentClient call sites covered

PR #743, opened 2026-04-12
docs: RFC 001 — storage backend plugin specification

igorls:
## Summary

- Drafts a formal contract for MemPalace storage backends so third parties can ship `pip install mempalace-<name>` packages that drop into core without patches.
- Resolves the open design decisions deliberately deferred by the #413 seam, driven by the six in-flight backend PRs each implementing the interface differently.
- Sets the stage for MemPalace as a long-lived daemon managing many palaces, where different palaces may route to different backends.

Tracking issue: #737 (see [discussion](https://github.com/MemPalace/mempalace/issues/737#issuecomment-4232950750) and [follow-up](https://github.com/MemPalace/mempalace/issues/737#issuecomment-4232963891) comments for the design rationale).

## Key decisions in the draft

- **Distribution:** entry-point group `mempalace.backends`; `pip install` is sufficient to add a backend.
- **Return types:** typed `QueryResult` / `GetResult` dataclasses replace Chroma's dict shape from day one. `ChromaCollection` gets a thin adapter.
- **Daemon-first model:** `PalaceRef(id, local_path?, namespace?)` replaces `palace_path: str`. Backend instances are long-lived and multi-palace; thread-safe across palaces.
- **Where-clause contract:** required subset `$eq, $ne, $in, $nin, $and, $or, $contains`; unknown operators MUST raise `UnsupportedFilterError` (silent drop forbidden).
- **Embedder decoupling:** mandatory injection; backends persist `model_name` and raise `EmbedderIdentityMismatchError` on swap (not just dimension).
- **Capability tokens:** `supports_*` naming, free-form strings, extensible by third parties.
- **Conformance:** shared `AbstractBackendContractSuite` pytest mixin + optional parametrized run of the core suite across all backends under `MEMPALACE_TEST_ALL_BACKENDS=1`.
- **Benchmark honesty:** optional `maintenance_state()` / `run_maintenance(kind)`; published numbers must cover three phases (post-load, post-native-maintenance, post-explicit-maintenance).
- **Migration:** backend-neutral `mempalace migrate` + `mempalace verify` operating through `BaseCollection` only — no per-backend migration code.
- **Sync:** capability flag + separate subsystem; local deployments never load it.

## Impact on in-flight PRs

Each of #574, #643, #665, #697, #700, #381 is called out in §11 with the specific alignment work required — #574 is closest to the final shape, #697's `collection_prefix` concern dissolves into `PalaceRef.namespace`, #700 and #381 need the canonical UUID5 namespace.

## Gating cleanup (not in this PR)

Seven files still import `chromadb` directly (`repair.py`, `dedup.py`, `cli.py` ×2, `mcp_server.py`, `migrate.py`, plus an instructions doc). Combined with the dict-to-typed-result migration, this needs its own PR landing before the spec can be enforced. Called out in §10.

## Test plan

- [ ] Backend authors (@skuznetsov @dekoza @RobertoGEMartin @cschnatz @Anush008) review for show-stoppers against their PRs
- [ ] Agreement on the three open questions in §12 (per-collection `changes_since` filter, per-palace capability query, `run_maintenance` return shape)
- [ ] Canonical `NAMESPACE_MEMPALACE` UUID assigned (§7.4)
- [ ] Once approved, follow-up PR to land the seam cleanup + typed-result migration (§10)
- [ ] Follow-up PR to land `AbstractBackendContractSuite` + entry-point discovery + `PalaceRouter` + `PalaceRef`

### DIFF
--- a/docs/rfcs/001-storage-backend-plugin-spec.md
+++ b/docs/rfcs/001-storage-backend-plugin-spec.md
@@ -138,7 +138,7 @@ On empty results: return a result object with empty inner lists, never raise. Sp
 
 **Required operators:** `$eq`, `$ne`, `$in`, `$nin`, `$and`, `$or`, `$contains`.
 
-Backends that do not support full-text natively must implement `$contains` via payload string match. The `supports_contains_fast` capability flag (§2.1) lets callers check whether a backend's implementation is indexed (fast) or scan-based (slow).
+Backends that do not support full-text natively MUST still implement `$contains` via payload string match — correctness is required; performance is not. `supports_contains_fast` (§2.1) is the only performance floor the spec promises. Without it, callers and benchmarks MUST assume `$contains` is O(n). This is an intentional split: `$contains` is a correctness requirement, `contains_fast` is the performance boundary, and the gap between scan and indexed FTS is too large for the spec to paper over.
 
 **Unknown operators:** backends MUST raise `UnsupportedFilterError`. Silent dropping is forbidden — it produces incorrect results.
 
@@ -146,13 +146,36 @@ Backends that do not support full-text natively must implement `$contains` via p
 
 ### 1.5 Embeddings
 
-Backends MUST accept a pre-computed `embeddings=` argument on `add` / `upsert`. Backends that recompute from text may ignore its value but MUST NOT reject the argument. This is required for lossless migration.
+#### Signature compliance (all backends)
+
+All backends MUST accept a pre-computed `embeddings=` argument on `add` / `upsert` without raising. This is signature compliance only — it does not guarantee the vectors are persisted (see passthrough below). Capability token: `supports_embeddings_in`.
 
 Backends MUST NOT hardcode embedding models or dimensions. Model selection is the embedder's responsibility (§4).
 
+#### Passthrough vs re-embed (separate guarantee)
+
+Accepting the argument is not the same as honoring it. Two distinct semantics, distinguished by capability:
+
+- **`supports_embeddings_passthrough`** — when `embeddings=` is provided, the backend MUST persist those vectors as-is and MUST NOT re-embed from text. This is the stronger guarantee lossless migration depends on.
+- **No `supports_embeddings_passthrough`** — the backend always re-embeds from text at write time. Provided `embeddings=` is accepted (signature compliance) but discarded. Migration *to* such a backend is re-embedding, not lossless transfer.
+
+`supports_migration_export` (source-side bulk read) MUST be paired with `supports_embeddings_passthrough` (target-side lossless write) for a migration to be labeled lossless. The `mempalace migrate` CLI refuses to run between backends where the target lacks `supports_embeddings_passthrough` unless `--accept-re-embed` is passed, which records re-embedding in the target palace's migration log.
+
+#### Dimension check (all backends, required)
+
 Backends MUST validate embedding dimension on first write to a new collection and on open of an existing collection, and MUST raise `DimensionMismatchError` on mismatch. Silent acceptance of mismatched dimensions produces unrecoverable corruption.
 
-**Dimension matching is necessary but not sufficient.** Swapping to a different model that happens to share a dimension (e.g., both 384-d) will silently degrade retrieval quality without tripping `DimensionMismatchError`. To guard this, backends MUST persist the `embedder.model_name` alongside the collection on first write, and MUST raise `EmbedderIdentityMismatchError` when a later open presents a different `model_name`. Users can override with an explicit `--force-model-swap` at the CLI layer, which records the swap in the palace's migration log.
+#### Model identity check (all backends, three-state)
+
+Dimension matching is necessary but not sufficient. Swapping to a different model that happens to share a dimension (e.g., both 384-d) silently degrades retrieval without tripping `DimensionMismatchError`. Backends MUST persist `embedder.model_name` alongside the collection on first write and MUST check it on subsequent open. Three outcomes:
+
+| State | Condition | Required behavior |
+|---|---|---|
+| `known_match` | Stored name equals current `embedder.model_name` | Proceed normally. |
+| `known_mismatch` | Stored name exists and differs from current | Raise `EmbedderIdentityMismatchError`. Override only via explicit CLI `--force-model-swap`, which writes the swap to the palace's migration log and updates the stored identity. |
+| `unknown` | No model name recorded (legacy collection, pre-v1 palace) | Do not hard-fail — emit a `EmbedderIdentityUnknownWarning` on first open. The resolved identity is recorded on the next successful write, reindex, or migration, transitioning the palace to `known_match` going forward. CLI exposes `mempalace palace set-embedder --model NAME` for explicit resolution. |
+
+The `unknown` state exists because existing palaces from #413 and earlier have no recorded identity; hard-failing them on upgrade would be hostile. Once recorded, subsequent opens are strict.
 
 ---
 
@@ -171,7 +194,8 @@ Defined capability tokens (v1):
 
 | Token | Meaning |
 |---|---|
-| `supports_embeddings_in` | Accepts pre-computed embeddings (MUST be true for all backends) |
+| `supports_embeddings_in` | Accepts pre-computed `embeddings=` without raising (signature compliance; MUST be true for all backends) |
+| `supports_embeddings_passthrough` | Persists provided `embeddings=` as-is without re-embedding (required for lossless migration target) |
 | `supports_embeddings_out` | Returns embeddings when `include=["embeddings"]` is requested |
 | `supports_estimated_count` | `estimated_count()` is meaningfully cheaper than `count()` |
 | `supports_update` | Distinguishes `upsert` from `add` with meaningful replace semantics |
@@ -296,7 +320,7 @@ When resolving a palace's backend, priority (highest first):
 4. Auto-detect from on-disk artifacts: `chroma.sqlite3` → `chroma`, `*.lance` → `lance`, etc. Backends declare detection hints via an optional `BaseBackend.detect(path: str) -> bool` classmethod.
 5. Default: `chroma`.
 
-Auto-detection preserves existing palaces when users upgrade — no manual config migration needed.
+**Auto-detection is strictly a migration/upgrade compatibility path, not a general selection mechanism.** It exists so existing palaces from v3.x keep opening without forced config migration. For *new* palaces, explicit configuration or CLI flag always wins — creating a palace without a resolved backend from (1)–(3) falls through to default (5), never to detection (4). Auto-detection fires only when a local path is presented AND no earlier rule has chosen a backend AND the path already contains backend-identifiable artifacts.
 
 ---
 
@@ -444,15 +468,25 @@ mempalace migrate --all --to lance
 
 Implementation is backend-agnostic: reads from source via `BaseCollection.get(include=["documents", "metadatas", "embeddings"])`, writes to target via `BaseCollection.upsert(...)` with the original embeddings. No backend-specific migration code.
 
-### 8.2 Safety
+### 8.2 Lossless vs re-embed
+
+Migration is labeled **lossless** only when:
+
+- The source advertises `supports_migration_export` (bulk read includes embeddings), AND
+- The target advertises `supports_embeddings_passthrough` (persists provided embeddings as-is), AND
+- Source and target agree on `embedder.model_name` (or `--force-model-swap` is explicit).
+
+If the target lacks `supports_embeddings_passthrough`, `mempalace migrate` refuses to run. Passing `--accept-re-embed` overrides — the migration proceeds but re-embeds from document text at write time, and the migration record labels the result as re-embedded rather than lossless. Retrieval quality may shift.
+
+### 8.3 Safety
 
 - Source is never modified. Migration is read-only against the source backend.
 - Target palace must not already exist unless `--overwrite` is passed.
-- A successful migration writes a `.mempalace-migration.json` record into the target palace (source backend, source path, timestamp, row count).
+- A successful migration writes a `.mempalace-migration.json` record into the target palace containing: source backend name, source path/ref, timestamp, row count, `lossless: true|false`, source and target `embedder.model_name`, and whether `--force-model-swap` or `--accept-re-embed` was used.
 
-### 8.3 Verification
+### 8.4 Verification
 
-After migration, run `mempalace verify --palace PATH --against SOURCE_PATH --source-backend chroma`. This samples N rows and confirms round-trip parity (ids match, documents match, embedding cosine similarity ≥ 0.999).
+After migration, run `mempalace verify --palace PATH --against SOURCE_PATH --source-backend chroma`. This samples N rows and confirms round-trip parity (ids match, documents match, embedding cosine similarity ≥ 0.999 when the migration was lossless; a looser document-overlap check when re-embedded).
 
 ---
 
--- a/docs/rfcs/001-storage-backend-plugin-spec.md
+++ b/docs/rfcs/001-storage-backend-plugin-spec.md
@@ -1,0 +1,514 @@
+# RFC 001 — Storage Backend Plugin Specification
+
+- **Status:** Draft
+- **Tracking issue:** [#737](https://github.com/MemPalace/mempalace/issues/737)
+- **Supersedes:** The informal seam introduced by [#413](https://github.com/MemPalace/mempalace/pull/413)
+- **Related:** [#266](https://github.com/MemPalace/mempalace/issues/266), [#574](https://github.com/MemPalace/mempalace/pull/574), [#643](https://github.com/MemPalace/mempalace/pull/643), [#665](https://github.com/MemPalace/mempalace/pull/665), [#697](https://github.com/MemPalace/mempalace/pull/697), [#700](https://github.com/MemPalace/mempalace/pull/700), [#381](https://github.com/MemPalace/mempalace/pull/381)
+- **Spec version:** `1.0`
+
+## Summary
+
+A formal contract for MemPalace storage backends so third parties can ship `pip install mempalace-<name>` packages that drop into the core without patches. The spec defines the collection interface, the backend lifecycle, registration via Python entry points, configuration shape, a required test contract, and a migration path between backends.
+
+It also sets up MemPalace to run as a long-lived daemon that manages many palaces, where different palaces may route to different backends.
+
+## Motivation
+
+Six backend PRs are currently in flight. Each one solves the same problem six different ways — different method signatures, different registration mechanisms, different embedder ownership, incompatible where-clause dialects, no shared test suite. The ad-hoc `BaseCollection` ABC merged in #413 was deliberately minimal and deferred every non-obvious decision. This RFC closes the open decisions so backend authors can build to a stable contract.
+
+## Goals
+
+1. A backend ships as a standalone Python package; installing it is sufficient to use it.
+2. All callers in MemPalace core go through the collection interface. No direct `chromadb` imports outside `mempalace/backends/chroma.py`.
+3. Backends are interchangeable: every backend passes the same shared test suite, and `mempalace migrate` moves palaces between them without data loss.
+4. The model scales from single-user local (one backend, one palace, no config) to a daemon serving many palaces with heterogeneous backends.
+5. Chroma's current dict-shaped return values are not the long-term contract. Typed results are spec v1.
+
+## Non-goals
+
+- Defining the embedder pipeline in detail. The embedder is a separate contract this spec depends on but does not specify.
+- Defining the sync subsystem. This spec only declares the capability flag and the minimal hook a sync subsystem will read.
+- Specifying wire protocol for a future networked daemon. That is a separate RFC.
+
+---
+
+## 1. Collection contract
+
+### 1.1 Required methods
+
+All backends implement `BaseCollection` with kwargs-only signatures:
+
+```python
+class BaseCollection(ABC):
+    @abstractmethod
+    def add(
+        self,
+        *,
+        documents: list[str],
+        ids: list[str],
+        metadatas: list[dict] | None = None,
+        embeddings: list[list[float]] | None = None,
+    ) -> None: ...
+
+    @abstractmethod
+    def upsert(
+        self,
+        *,
+        documents: list[str],
+        ids: list[str],
+        metadatas: list[dict] | None = None,
+        embeddings: list[list[float]] | None = None,
+    ) -> None: ...
+
+    @abstractmethod
+    def query(
+        self,
+        *,
+        query_texts: list[str] | None = None,
+        query_embeddings: list[list[float]] | None = None,
+        n_results: int = 10,
+        where: dict | None = None,
+        where_document: dict | None = None,
+        include: list[str] | None = None,
+    ) -> QueryResult: ...
+
+    @abstractmethod
+    def get(
+        self,
+        *,
+        ids: list[str] | None = None,
+        where: dict | None = None,
+        where_document: dict | None = None,
+        limit: int | None = None,
+        offset: int | None = None,
+        include: list[str] | None = None,
+    ) -> GetResult: ...
+
+    @abstractmethod
+    def delete(
+        self,
+        *,
+        ids: list[str] | None = None,
+        where: dict | None = None,
+    ) -> None: ...
+
+    @abstractmethod
+    def count(self) -> int: ...
+```
+
+### 1.2 Optional methods (default implementations on the ABC)
+
+```python
+def estimated_count(self) -> int:
+    return self.count()
+
+def close(self) -> None:
+    return None
+
+def health(self) -> HealthStatus:
+    return HealthStatus.ok()
+```
+
+Backends with cheap approximate counters override `estimated_count`. Backends that hold connections must override `close`.
+
+### 1.3 Typed results (replaces Chroma dict shape)
+
+```python
+@dataclass(frozen=True)
+class QueryResult:
+    ids: list[list[str]]                              # outer = queries, inner = hits
+    documents: list[list[str]]
+    metadatas: list[list[dict]]
+    distances: list[list[float]]
+    embeddings: list[list[list[float]]] | None = None
+
+@dataclass(frozen=True)
+class GetResult:
+    ids: list[str]
+    documents: list[str]
+    metadatas: list[dict]
+    embeddings: list[list[float]] | None = None
+```
+
+On empty results: return a result object with empty inner lists, never raise. Specifically, an empty query returns `QueryResult(ids=[[]], documents=[[]], metadatas=[[]], distances=[[]])` — the outer dimension is the number of query vectors issued; the inner dimension is hits per query and may be zero.
+
+`include` controls which fields are populated. Fields not in `include` are populated with empty lists of the correct outer shape; they are never `None` (except `embeddings`, which is `None` when not requested).
+
+### 1.4 Where-clause dialect
+
+**Required operators:** `$eq`, `$ne`, `$in`, `$nin`, `$and`, `$or`, `$contains`.
+
+Backends that do not support full-text natively must implement `$contains` via payload string match. A `$contains_fast` capability flag lets callers check whether a backend's implementation is indexed (fast) or scan-based (slow).
+
+**Unknown operators:** backends MUST raise `UnsupportedFilterError`. Silent dropping is forbidden — it produces incorrect results.
+
+**Optional operators:** `$gt`, `$gte`, `$lt`, `$lte`. Backends either implement them or reject with `UnsupportedFilterError`. Advertised via capabilities.
+
+### 1.5 Embeddings
+
+Backends MUST accept a pre-computed `embeddings=` argument on `add` / `upsert`. Backends that recompute from text may ignore its value but MUST NOT reject the argument. This is required for lossless migration.
+
+Backends MUST NOT hardcode embedding models or dimensions. Model selection is the embedder's responsibility (§4).
+
+Backends MUST validate embedding dimension on first write to a new collection and on open of an existing collection, and MUST raise `DimensionMismatchError` on mismatch. Silent acceptance of mismatched dimensions produces unrecoverable corruption.
+
+**Dimension matching is necessary but not sufficient.** Swapping to a different model that happens to share a dimension (e.g., both 384-d) will silently degrade retrieval quality without tripping `DimensionMismatchError`. To guard this, backends MUST persist the `embedder.model_name` alongside the collection on first write, and MUST raise `EmbedderIdentityMismatchError` when a later open presents a different `model_name`. Users can override with an explicit `--force-model-swap` at the CLI layer, which records the swap in the palace's migration log.
+
+---
+
+## 2. Backend contract
+
+### 2.1 Identity and capabilities
+
+```python
+class BaseBackend(ABC):
+    name: ClassVar[str]                    # "chroma", "postgres", "qdrant", ...
+    spec_version: ClassVar[str] = "1.0"    # which spec version this backend targets
+    capabilities: ClassVar[frozenset[str]]
+```
+
+Defined capability tokens (v1):
+
+| Token | Meaning |
+|---|---|
+| `supports_embeddings_in` | Accepts pre-computed embeddings (MUST be true for all backends) |
+| `supports_embeddings_out` | Returns embeddings when `include=["embeddings"]` is requested |
+| `supports_estimated_count` | `estimated_count()` is meaningfully cheaper than `count()` |
+| `supports_update` | Distinguishes `upsert` from `add` with meaningful replace semantics |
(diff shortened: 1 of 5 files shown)


dekoza (commented):
Reviewed from the LanceDB (#574) perspective. No show-stoppers — our implementation is already close to the spec's target shape. Notes below.

**§1.1 — kwargs-only signatures**

Our `backends/lance.py` already uses kwargs-only on `add`/`upsert`, and our `BaseCollection` ABC in `backends/base.py` does the same (work done in #575, easily backported). `query`/`get`/`delete` currently use `**kwargs` catch-all — those need explicit parameter lists to match the spec.

**§1.1 — `query_embeddings` and `where_document`**

Our `query()` doesn't accept `query_embeddings` and neither `query()` nor `get()` accepts `where_document`. Both are straightforward for LanceDB:
- `query_embeddings`: skip the embed step, search the vector directly
- `where_document`: same `$contains` path (Tantivy FTS or `LIKE` fallback)

**§1.1 — `update()` method**

Our `BaseCollection` and `LanceCollection` already have an `update()` method (fetches existing records, merges changes, re-upserts). The spec's `BaseCollection` does not include `update()`, though `supports_update` exists as a capability token. Is `update()` intentionally excluded from the required interface, or an oversight? If it stays out, we'd keep it as a LanceDB-specific extension — but it seems generally useful.

**§1.3 — Typed results**

Our return shapes already match `QueryResult`/`GetResult` field structure. Migration is wrapping dict construction in dataclass constructors. No concern.

**§1.5 — Three-state model identity**

We persist `embedding_model` per-record in `metadata_json` but not at collection level. Spec wants collection-level persistence + check on open. LanceDB supports table-level metadata via Arrow schema metadata — clean fit. The `unknown` state for legacy palaces is the right call; hard-failing on upgrade would break existing users.

**§2.3 — `BaseBackend` class**

We already have `LanceBackend` in `backends/lance.py` with a `get_collection()` factory method (from #575, easily backported to #574). It currently takes `palace_path: str` — adapting to `PalaceRef` is straightforward. It does not yet subclass a `BaseBackend` ABC (spec §2.3), but the method shape is close.

**§3.3 — Auto-detection**

Our `backends/__init__.py` already has `detect_backend()` sniffing for `.lance` directories. Adapting to a `BaseBackend.detect(path) -> bool` classmethod is trivial. Agree with the spec's framing that auto-detection is a migration compatibility path, not a selection mechanism.

**§6 — Sync compatibility note**

Our sync implementation uses a `_raw=True` flag on `upsert()` to skip sync metadata injection during sync apply. The spec's `BaseCollection.upsert()` has no such parameter, which is correct — sync concerns should not leak into the collection contract. This means we need to refactor sync injection out of `LanceCollection` into a wrapper layer, which aligns with §6's "sync is a separate subsystem" stance. Wanted to flag this since other backends implementing sync will face the same design question.

**§10 — Cleanup gating**

+1 on landing the Chroma direct-import cleanup before backend PRs rebase. The seven files importing `chromadb` directly would break any pure-Lance deployment. This cleanup benefits all backend authors equally.

**§11 — Alignment effort for #574**

Assessment is accurate. We already have `backends/` with `BaseCollection` ABC, `LanceBackend`, and kwargs-only signatures (work from #575, easily backported). Remaining work: explicit param lists on `query`/`get`/`delete` (instead of `**kwargs`), typed results, `$contains`/`where_document`, collection-level model identity, `PalaceRef` adoption, conformance test subclass, sync injection refactor, and package extraction to `mempalace-lance`. All small-to-medium. No architectural conflicts.

**§12 — Open questions**

1. *`changes_since` with collection filter* — yes, useful. Our sync tracks changes per-table; filtering by collection name is natural.
2. *Per-palace capabilities* — worth having. `supports_contains_fast` depending on whether an FTS index exists is a real case for us.
3. *`run_maintenance` return shape* — structured return preferred. LanceDB compaction can report fragment count before/after and bytes reclaimed, useful for operator dashboards and benchmark reporting.

skuznetsov:
Reviewed the actual RFC diff. From the Postgres / `pg_sorted_heap` side, I do not see a blocker. The direction is compatible with #665: injected embedder, explicit capability tokens, lossless-vs-reembed migration split, and maintenance-state benchmark hooks all map cleanly.

Three small spec-clarity points I would consider before backend authors start rebasing:

1. **Goal wording vs §8.2 migration semantics**

The Goals section says `mempalace migrate` moves palaces between backends “without data loss”. §8.2 is more precise: migration is lossless only when the source supports embedding export, the target supports embedding passthrough, and model identity is compatible. I would mirror that nuance in the Goals section so the RFC does not overpromise. Suggested framing: “supports lossless migration when source/target capabilities allow it, and explicit re-embedding otherwise.”

2. **`server_embedder` needs an explicit identity/dimension contract**

§1.5 says backends MUST NOT hardcode models and must validate model identity/dimension, but §2.1 / §5 allow `server_embedder` backends to ignore the injected embedder. That can be valid, but the RFC should say what identity gets recorded and validated in that mode. For example: a server-embedder backend should expose its effective `model_name` and `dimension`, and the same dimension/model identity rules should apply to that effective embedder. Otherwise `server_embedder` becomes an implicit escape hatch from the safety rules.

3. **Maintenance kinds need discovery**

§7.3 says benchmarks should run explicit `run_maintenance` if the backend advertises maintenance kinds, but I do not see where those kinds are advertised. A small hook would make the benchmark harness deterministic, e.g. `maintenance_kinds: frozenset[str]` or `maintenance_state()["supported_kinds"]`. Without that, harnesses will need backend-specific knowledge to decide whether `"analyze"`, `"compact"`, or `"reindex"` exists.

None of these change the shape of the RFC. They are mostly guardrails to keep the spec precise once implementations begin targeting it.

jphein (commented):
Reviewed from the perspective of a 134K-drawer production deployment on the Chroma backend, with experience evaluating LanceDB (via [Karta](https://github.com/jphein/karta)'s embedded LanceDB + SQLite architecture).

No show-stoppers. This is the right time to formalize the backend contract — the six-way divergence is already causing friction. Notes below.

**§1.3 — Typed results (migration cost for existing forks/consumers)**

This is a breaking change for anyone with code touching `mcp_server.py` or `searcher.py` that assumes dict returns. My fork has ~20 changes across those files — all dict-shaped. The migration path is clear (wrap in dataclass constructors, same as dekoza noted), but the blast radius is worth calling out: it's not just backend authors who need to update, it's everyone downstream who consumes query results. A compatibility shim (`.to_dict()` on the result types) during the transition would make the upgrade gentler for plugin authors and forks.

**§1.5 — Three-state model identity is exactly right**

My palace has 134K drawers with no recorded embedder identity. Hard-failing on upgrade would brick every pre-v1 palace. The `unknown` → warn → record-on-next-write path is the correct design. One detail: the spec says the identity is recorded "on the next successful write, reindex, or migration." For read-heavy deployments that rarely write new drawers, this means the palace could stay in `unknown` state indefinitely. Consider also recording on explicit `mempalace verify` or `mempalace status` (read-only operations that already touch the collection) so operators can resolve it without forcing a write.

**§2.5 — Concurrency guarantee matches real usage**

The spec says backends can assume single-thread access per collection, with core serializing per palace. This matches how the MCP server actually works today — good call. My fork added `threading.Lock` to the graph cache (#661) because `build_graph()` was the one place concurrent access was possible. The per-palace serialization guarantee would have made that unnecessary. Worth documenting this guarantee prominently so backend authors don't over-synchronize.

**§7.3 — Benchmark methodology is the most underrated section**

At 134K drawers on Chroma, I've hit the HNSW pathology firsthand — the graph's on-disk size and query latency behave very differently before and after compaction, and after external writes that invalidate the in-memory index (the exact problem my stale HNSW mtime detection fix addresses in #663). The three-phase benchmark requirement (post-bulk-load, post-background-maintenance, post-explicit-maintenance) would have caught the performance cliff I found empirically. Strong +1 on `maintenance_state()` being part of the published numbers.

**§8 — Migration at scale**

At 134K drawers, re-embedding is not a minor cost — it's hours of compute depending on the model. The lossless vs re-embed distinction and the `--accept-re-embed` explicit opt-in are the right design. One request: `mempalace migrate` should report progress (rows migrated / total, ETA) for large palaces. A 134K-drawer migration that runs silently for hours will get killed by impatient operators.

**§10 — Cleanup prerequisite will affect fork contributors**

My fork's `mcp_server.py` imports `chromadb` directly for the BLOB seq_id repair (#664) and mtime-based cache invalidation (#663). Both are deeply Chroma-specific. The cleanup PR routing all callers through `BaseCollection` is the right gating decision, but it will require Chroma-specific fixups to move behind the backend boundary. Backend-specific maintenance operations (like BLOB repair) might need a `BaseBackend.repair()` or similar escape hatch — otherwise they end up as out-of-band scripts that bypass the abstraction.

**§12 — Open questions**

2. *Per-palace capabilities* — yes. `supports_contains_fast` depending on whether an FTS index exists is a real case. I'd add: `supports_mtime_detection` or similar for backends where external-write detection is possible (Chroma via SQLite mtime) vs not (server-mode backends where the filesystem isn't local). My #663 fix is Chroma-specific precisely because mtime detection only makes sense for local file-backed stores.

igorls:
Thanks @skuznetsov @dekoza @jphein — substantive reviews on all sides. Commit 922aa99 closes the four spec defects the reviews surfaced. A number of reasonable operator asks are explicitly *not* being adopted; explaining both below, so the boundary is clear for anyone implementing against this.

## Adopted — spec defects

**1. Goals wording (skuznetsov)** — Goal 3 now mirrors §8.2 instead of overpromising "without data loss." The spec delivers lossless migration when capabilities allow, re-embedding otherwise. Both are explicit; neither is hidden.

**2. `server_embedder` identity contract (skuznetsov)** — a stated safety invariant with an unspecified escape hatch is a real correctness hole. Closed:

> A backend advertising `server_embedder` (§2.1) provides its own embedder and MAY ignore the `embedder=` kwarg passed to `get_collection`. That does **not** exempt it from the dimension and identity rules above. Such backends MUST expose an effective `model_name: str` and `dimension: int` via `BaseCollection.effective_embedder_identity()`, persist that effective identity on first write, and validate it on open under the same three-state rules.

`server_embedder` now documents *where* the embedding happens; it never suspends the safety contract. A backend that can't report its effective embedder identity doesn't qualify for the capability.

**3. `maintenance_kinds` discovery (skuznetsov)** — §7.3 referenced something the spec never defined. Fixed with a class-level `maintenance_kinds: ClassVar[frozenset[str]]`. The spec reserves `"analyze"`, `"compact"`, `"reindex"` as well-defined names; backends may add their own. `run_maintenance(kind)` MUST raise `UnsupportedMaintenanceKindError` for unadvertised kinds. The benchmark harness reads this set rather than guessing names.

**4. `update()` method / `supports_update` mismatch (dekoza)** — a capability token that gates nothing is incoherent. `update()` is now an optional method in §1.2 with a default implementation (get → merge → upsert). `supports_update` is redefined to gate *atomic, single-round-trip* semantics — not the mere ability to do partial updates, which the default already provides. Backends with native primitives (Postgres `UPDATE`, Lance `merge_insert`) override and advertise; others inherit the default.

## Not adopted — operator comfort, not plugin-contract concerns

These are all reasonable asks, and I want to be explicit that they're *not* dismissed because they're wrong — they're not adopted because they don't belong in this spec. A well-engineered plugin contract has a small, stable surface. Layering operator-UX conveniences into the core interface makes it worse for every future backend author.

- **`.to_dict()` compat shim on typed results (jphein)** — directly contradicts §1.3's design intent of dropping the Chroma dict shape. Adding an escape hatch re-legitimizes the thing the spec removes. Fork migration pain is what minor version bumps are for; callers move to attribute access.

- **Migration progress reporting (jphein)** — CLI UX concern. Belongs in the `mempalace migrate` implementation, not in the backend contract. A backend doesn't care whether the CLI prints a progress bar. The CLI PR will implement it; the spec says nothing about it.

- **`BaseBackend.repair(kind)` escape hatch (jphein)** — redundant with `run_maintenance(kind)`. "Repair" is a maintenance kind, not a separate concept. Keeping them distinct would fragment the same mechanism. Chroma's BLOB seq_id repair becomes `run_maintenance("repair")` or a Chroma-specific reserved kind.

- **Per-palace capability variance (dekoza, jphein)** — this muddies the contract. Capabilities advertise what a backend *can do*. Making them palace-dependent turns every capability check into "maybe — depends." That pushes state tracking onto every caller for a small gain. A backend that builds indexes lazily either guarantees the index exists when needed (and advertises the capability) or doesn't. Static contract.

- **Identity recording on read-only ops (jphein)** — complicates the state machine for an operator-UX concern. "My read-heavy palace stays `unknown` forever" is solved by `mempalace palace set-embedder --model NAME` (already in §1.5), not by making the state transition rules asymmetric.

- **`supports_mtime_detection` (jphein)** — almost certainly redundant with sync change-feed contract once the sync subsystem is specified. Don't add capabilities speculatively.

## Net

Four spec edits, small and targeted. The contract is tighter for it. The rejected items are good ideas for the implementations — CLI progress, fork-friendly shims, richer operator tooling — they just don't live in the plugin contract.

#743 is ready for another look.

skuznetsov:
Thanks, this addresses my Postgres / `pg_sorted_heap` concerns. The revised goal wording, `server_embedder` identity contract, and `maintenance_kinds` discovery hook all look aligned with what #665 needs. I’m good with this RFC direction from the Postgres backend side and will review the implementation PRs when they land.

skuznetsov:
Quick coordination question from the PostgreSQL / pg_sorted_heap backend side:

Is RFC 001 now stable enough to use as the target contract for reworking #665, or would you prefer that implementation PRs wait until #743 is merged and the §10 backend cleanup follow-up starts/lands?

From my side, the expected #665 rewrite would target the RFC shape directly:
- injected/recorded embedder identity instead of backend-owned embedding globals
- typed `QueryResult` / `GetResult`
- `BaseBackend` / `BaseCollection` + capability tokens
- `PalaceRef`
- maintenance hooks (`maintenance_kinds`, `maintenance_state`, `run_maintenance`)
- conformance test coverage
- conservative docs positioning for `pgvector` vs optional `pg_sorted_heap`

I’m asking to avoid rebasing the current conflicting branch into an intermediate shape if the intended path is now RFC-first.

cschnatz:
Read the full spec. From the multi-tenant hosted side (#697), this looks solid.

### Works for us

- **§4.4** `PalaceRef.namespace` absorbs our `collection_prefix` cleanly. We're already building shared team palaces on top of this — personal vaults map to `PalaceRef(namespace="tenant_<uuid>")`, team vaults to `PalaceRef(namespace="team_<uuid>")`, and search fans out across both. The namespace abstraction is exactly what we need.
- **§2.5** Concurrency model matches our sidecar design.
- **§3.1** Entry points mean we can ship our `ChromaHttpBackend` as a proper plugin instead of vendoring.

### Questions

**1. Namespace isolation — security boundary or naming convention?**

For us, namespace is a tenant isolation boundary. When a user searches across personal + team vaults, the sidecar resolves which namespaces to query based on a gateway-authorized team_ids list. A team namespace that isn't in that list gets rejected. §4.4 says "the backend uses it as given" — should backends also enforce namespace isolation, or is that purely the caller's job?

**2. PalaceRef.id uniqueness scope**

Unique globally or per backend instance?

**3. Typed results (§1.3) — hard cutover?**

Is there a transition period, or do all callers need to migrate from Chroma dicts in one go? We have 30 MCP tool functions consuming dict shapes.

**4. §7.4 NAMESPACE_MEMPALACE**

Worth assigning now — the placeholder blocks Qdrant PRs.

**5. §10 mcp_server caching**

We hit a related bug (collection cache bleed between tenants). Moving caching into `ChromaBackend.get_collection()` would fix this structurally.

### #697 alignment

Our `collection_prefix` dissolves into `PalaceRef.namespace` as described in §11. Ready to rebase once spec + cleanup PR land. Happy to be an early conformance suite adopter.

bensig (approved):
**Reviewed in full.** This is the right shape and depth for a 1.0 spec — closes the open decisions deliberately deferred by #413, calls out the in-flight PR impact concretely, names the cleanup work it depends on, and stays out of scope where deferral is honest (sync, wire protocol, embedder).

Approving on merit. One mechanical block + a few suggestions worth folding in before this lands.

## Block

- **§7.4 \`NAMESPACE_MEMPALACE = uuid.UUID(\"TO-BE-ASSIGNED-ONCE-FOR-ALL-TIME\")\`** — placeholder. Needs an actual UUID assigned and recorded in the spec text before this merges, since the section explicitly says \"fixed at spec v1 adoption.\" Suggest \`python -c \"import uuid; print(uuid.uuid4())\"\` and pin it.

## Suggestions (small, fold-in-able)

1. **§5 references \"a separate RFC\" for the embedder contract.** Worth either filing the tracking issue and linking it from this section, or marking the dependency as a parallel work item in §13. Right now the spec hard-depends on an external contract that has no anchor.

2. **§4.2 env var splitting rule for hyphenated/underscored backend names.** The example uses backend name \"pg_prod\" in §4.1 but \`MEMPALACE_POSTGRES_DSN\` in §4.2. If users name a backend \"pg_prod\", does the env shape become \`MEMPALACE_PG_PROD_*\`? \`MEMPALACE_PG-PROD_*\`? Or is \`MEMPALACE_<NAME>_\` keyed by backend *type* not instance? One sentence would close it.

3. **§3.3 priority + global env interaction.** A user with \`MEMPALACE_BACKEND=postgres\` set globally who opens a palace with on-disk Chroma artifacts gets postgres (env wins, auto-detect skipped). That's correct behavior, but worth one sentence calling it out — \"setting MEMPALACE_BACKEND globally overrides existing-palace auto-detection; users opening pre-existing palaces should leave it unset.\" Saves a real support incident.

4. **§9 major-version mismatch error.** \"Refuses to load a backend declaring a different major version.\" Worth a one-line example of the error message shape so backend authors know what their users will see when they install an incompatible version. Drop a \`BackendVersionMismatchError(...)\` example next to the rule.

5. **§7.4 reserved maintenance kinds.** \`\"analyze\"\` / \`\"compact\"\` / \`\"reindex\"\` are reserved with required semantics. Consider naming the *non-required* implementation pattern: e.g., \"a backend that has no analogue for a kind MUST omit it from \`maintenance_kinds\` rather than declaring it as a no-op.\" Otherwise nothing prevents a benchmark harness from seeing \`\"analyze\"\` in \`maintenance_kinds\` and assuming it does what the spec says when the implementation is empty.

## Strengths

- **Capability-token design is sharp.** The signature-compliance vs semantic-guarantee split (\`supports_embeddings_in\` vs \`supports_embeddings_passthrough\`) is exactly the right granularity. Same for the \`server_embedder\` carve-out in §1.5 — \`server_embedder\` documents *where* embedding happens, never suspends the dimension/identity safety contract. Beautifully written paragraph; reads like it survived a real ambiguity.
- **Three-state embedder identity check** (\`known_match\` / \`known_mismatch\` / \`unknown\`) is the kind of detail that prevents a real upgrade incident. Hard-failing legacy palaces from #413 would be hostile; this gives them a clean transition path with a warning.
- **§10 cleanup prerequisite is honest.** Names the 7 files still importing \`chromadb\` directly with line numbers, calls out \`mcp_server._get_client()\` as a Chroma-specific cache that needs to migrate into \`ChromaBackend.get_collection()\` per §2.5. Most RFCs hand-wave this section; this one names the actual work.
- **Migration honesty (§8.2).** Lossless vs re-embed is explicit; \`--accept-re-embed\` is required, written to the migration log, never silent. The pairing requirement (\`supports_migration_export\` source-side + \`supports_embeddings_passthrough\` target-side) prevents a class of subtle silent-degrade bugs.
- **Benchmark methodology in §7.3** is rare to see in an RFC and prevents the un-\`ANALYZE\`'d-Postgres-vs-settled-Chroma anti-pattern. Reserved kind names + three-phase publishing requirement + harness MUST NOT assume kind names — that's the level of rigor this saves you from re-litigating later.
- **§3.3 auto-detection scoped tightly.** \"Strictly a migration/upgrade compatibility path, not a general selection mechanism.\" Explicit configuration always wins. Right call.
- **§11 in-flight PR impact table** is concrete. Shows this spec was written with the actual outstanding work in mind, not in a vacuum. Each PR has a one-line align-effort estimate.
- **§13 rollout sequence is realistic** — cleanup first, spec second, Chroma updated third, in-flight rebased fourth, migration CLI fifth. Matches what's actually achievable.

- **RFC 002 (source adapter plugin spec) shipped in v3.3.2 ahead of RFC 001.** The numbering doesn't imply ordering, but it's worth noting Igor is comfortable shipping concrete implementation ahead of formal RFC merge. The §10 cleanup discipline + §11 PR impact table here suggest that's a deliberate pattern, not drift. Approving on the strength of that.

---

Once the UUID gets assigned and the four small clarifications are folded in, this is ready to merge as the contract that the v4.0-alpha backend work targets. Status can move from Draft to Accepted with a date.

jphein (commented):
Follow-up review after the 2026-04-13 round + @bensig's approval. Most of my earlier concerns are addressed; one new empirical observation worth flagging.

## Closed from my prior review

- **§1.5 read-heavy edge case** (palace stays `unknown` indefinitely if it rarely writes) — closed by the explicit `mempalace palace set-embedder --model NAME` CLI in the three-state table at line 192. Operators of read-heavy palaces have a path that doesn't require waiting for a stale write to trigger identity resolution. ✓

- **§7.4 UUID5 namespace** — bensig's block is still mechanical (placeholder `TO-BE-ASSIGNED-ONCE-FOR-ALL-TIME`); no opinion from this side on the value, just confirming it's the only remaining gate.

## Still open from my prior review (no new info needed, just a note)

- **§1.3 typed-results migration** — RFC currently doesn't sketch a `.to_dict()` compat shim on `QueryResult` / `GetResult`. Forks and downstream consumers with code touching `mcp_server.py` or `searcher.py` that assume dict returns will need real edits, not no-op wrappers. The cleanup PR mentioned in §10 is the right place to land it; just naming the question now so it isn't a surprise during that PR's review.

## New: per-palace multi-collection isn't in the spec

The fork shipped a structural fix this week that's adjacent to RFC 001 in an interesting way. Cat 9 A/B benchmark on a 151K-drawer canonical palace surfaced that Stop-hook auto-save checkpoints (short, query-term-saturated) dominate vector top-N — `kind=all` returned 632 tokens/Q of mostly-checkpoint word-soup; `kind=content` (post-filter) returned 3 because over-fetch=100 wasn't enough. Recall was 0.984 R@5; E2E quality collapsed.

The fix promoted the architecture from "filter at query time" to "split at storage time": move checkpoints to a dedicated `mempalace_session_recovery` ChromaDB collection (same client, same palace, separate index), with a new `mempalace_session_recovery_read` MCP tool reading by session_id / agent / since-until. Implementation: `palace.py` gained `_SESSION_RECOVERY_COLLECTION` + `get_session_recovery_collection(palace_path)` mirroring `get_collection(palace_path, collection_name="mempalace_drawers", create=...)`. Palace-daemon's `lifespan` runs `migrate_checkpoints_to_recovery()` on startup so existing palaces auto-migrate.

This works on the current `BaseBackend` / `BaseCollection` API — `get_collection(palace, collection_name=...)` is already keyed by `collection_name`, so adding a sibling collection per palace is just calling it with a different name. The design fell out of the existing seam.

Question for the spec: **is "multiple collections per palace, by purpose" intentionally implicit in the API, or worth one sentence in §2.5 / §3.1?** Reading the RFC, "palace" feels like the unit (1:1 with a collection); but production needed >1 collection per palace for the verbatim-vs-derivative split. Backends like Postgres (#665) and Qdrant (#700) handle this trivially via schema/collection naming, but a future backend that assumes 1:1 — or a backend author reading the spec — might design themselves into a corner. A "backends MUST support N collections per palace, keyed by `collection_name` arg" line would close the gap without changing any signatures.

(Spec at §1.6 calls out "many palaces" but not "many collections per palace." Adjacent but distinct.)

## Rest of the spec

Read clean. §10's flagging of `mcp_server._get_client()` cache + reconnect for migration-into-`ChromaBackend` is exactly right — that #757 work is Chroma-specific and shouldn't live in `mcp_server`. §11's PR-impact table accurately reflects the in-flight backends; nothing missing from this side.

Net: ready to merge after UUID5 and (if desired) the multi-collection sentence.

cschnatz:
Following up on Q1 from my 2026-04-15 comment — re-stating concretely now that approval is close, since this is the one item I'd like pinned in the spec rather than left to implementations.

We treat `PalaceRef.namespace` as a **mandatory isolation boundary at the contract level**. A backend that accepts a write under `PalaceRef(namespace="tenant_A")` and later returns those records for a read under `PalaceRef(namespace="tenant_B")` is, for us, a critical defect — not a caller misconfiguration. Auth/authz (which `team_ids` a sidecar may query) stays on the deployment side. But cross-namespace bleed *within* a single backend instance has to be a spec violation, not implementation choice.

§4.4 currently reads as naming-only ("the backend uses it as given"). Suggest one MUST clause:

> Backends MUST scope all reads, writes, and deletes by `PalaceRef.namespace`. A record written under one namespace MUST NOT be returned, modified, or deleted by an operation issued under a different namespace within the same backend instance. Cross-namespace access is a spec violation.

Without this, hosted multi-tenant deployments can't cite RFC 001 as the basis for tenant isolation, and the contract becomes unenforceable across plugins from different authors.

kostadis-ntnx:
## Review from a backend author's seat (turbovecdb)

I implemented turbovecdb against the `#413` seam and have been tracking this RFC as the contract I'd build to. Overall it's a strong fit for a turbovecdb-shaped backend — the `$contains` correctness vs `supports_contains_fast` split (§1.4) means I don't have to build an FTS index, string IDs sidestep §7.4's UUID dance, `local_mode` + cosine + precomputed-vector passthrough all map 1:1, and `supports_embeddings_passthrough`/`supports_migration_export` come for free because the durable SQLite row store keeps the exact float32 vector. The §10 cleanup is the part I can't do from inside an adapter and is exactly what promotes a non-chroma backend from "stores and retrieves" to a real peer. Three focused asks before this is enforceable:

### 1. §10 cleanup omits `searcher.py`, which is the highest-leak module
The seven files listed for cleanup don't include `searcher.py`, but that's where the recall-survival fallback and the metric assumption live:
- `searcher._bm25_only_via_sqlite()` opens `chroma.sqlite3` directly and queries chroma's FTS5 `embedding_fulltext_search` shadow table. On any non-chroma palace this returns `{"error": "No palace found"}`.
- `searcher._hybrid_rank()` hard-codes `vec_sim = max(0, 1 - distance)`, i.e. it assumes cosine distance ∈ [0,2].

As written, the cleanup would leave the search hot-path chroma-coupled even after the spec lands. For backends whose distance isn't cosine (Qdrant dot-product, etc.) the ranking is silently wrong; for any backend the fallback is dead. Proposal: either add `searcher.py` to the §10 list, or make the fallback a capability (e.g. gate `vector_disabled` routing on the backend declaring it, and have backends whose index self-heals simply never request it), and make `_hybrid_rank` metric-aware off a backend-declared metric rather than hard-coding cosine.

### 2. §1.5 mandatory embedder-identity check depends on an Embedder RFC that doesn't exist yet
§1.5 makes persisting `embedder.model_name` and raising `EmbedderIdentityMismatchError` a hard `MUST` for all backends — which is the right call, a same-dimension model swap is the worst failure class. But §5 defers the `Embedder` protocol (`model_name`, `dimension`, `embed()`) to "a separate RFC." A backend can't be §1.5-conformant until that protocol is pinned, because today backends receive a bare `embed(texts) -> vectors` callable with no identity to persist. Please mark the Embedder-protocol RFC as a hard, blocking dependency of §1.5 (or inline the minimal protocol here), and clarify what a backend persists when the injected embedder exposes no `model_name` — does it fall to the §1.5 `unknown` state, or is a nameless embedder a hard error?

### 3. Confirm the passthrough/migration capability pairing for an exact-vector store
turbovecdb keeps the exact float32 vector in its SQLite source of truth, so it qualifies for both `supports_embeddings_passthrough` (§1.5) and `supports_migration_export` (§8.2) — making `chroma -> turbovec` lossless under §8. Can we get an explicit line in §8.2 confirming that a backend persisting verbatim vectors satisfies the lossless pairing on both ends (it currently only spells out the source/target direction), and that re-rank-from-stored-vectors backends are first-class migration targets?

None of these block the direction — they're about making the contract actually deliverable end-to-end for a backend that isn't chroma. Happy to prototype the `searcher.py` capability-gating against turbovecdb if that's useful.

igorls:
## Revision — resolving the review blockers; moving to Accepted

Thanks all for the depth here — @skuznetsov @dekoza @cschnatz @jphein @kostadis-ntnx @bensig. This revision closes the outstanding items and reconciles the spec with what actually shipped in #1679 (the in-tree `pgvector` / `qdrant` / `sqlite_exact` backends landed ahead of this merge). Status is now **Accepted (2026-06-07)**, with a new banner up top being explicit that acceptance pins the contract — it does **not** claim the implementation is complete.

### Blockers closed
- **§7.4 UUID (@bensig's mechanical block).** Pinned `NAMESPACE_MEMPALACE = c06c3fc7-5c14-4dc4-84c2-24a5f72d8dc1`. Rather than minting a fresh one, this promotes the value the in-tree `qdrant` backend already shipped with — so no deployed Qdrant palace's point IDs get re-derived. "Once, for all time" is now a real constant.
- **Namespace isolation (@cschnatz).** §4.4 now carries the cross-namespace MUST, expressed through a `supports_namespace_isolation` capability (already implemented in `base.py` + pgvector + qdrant). Backends that advertise it guarantee no cross-namespace bleed within an instance — citable as the basis for tenant isolation. Path-rooted local backends (e.g. `sqlite_exact`) are honestly exempt and callers MUST NOT rely on `namespace` there. This felt cleaner than a blanket all-backends MUST; let me know if the hosted side needs anything stronger.

### Backend-author asks folded in
- **@kostadis-ntnx (ask 1) + @jphein (on #1679) — `searcher.py` metric coupling.** Added a backend-declared `distance_metric` (§2.1) and put `searcher.py` explicitly in the §10 cleanup: `_hybrid_rank`'s `max(0, 1-distance)` becomes metric-aware, and the BM25 fallback's direct `chroma.sqlite3` access gets routed through `BaseCollection`. Tracked in #1726.
- **@kostadis-ntnx (ask 2) + @bensig (suggestion 1) — §1.5 depends on an Embedder RFC that doesn't exist.** §5 now pins the minimal protocol (`model_name`, `dimension`, `embed`) as **normative here** and sufficient for §1.5; the fuller RFC is additive. Nameless embedder → `unknown`, not a hard error. Implementation tracked in #1724.
- **@kostadis-ntnx (ask 3) — exact-vector stores.** §8.2 now states a verbatim-float32 store satisfies both sides of the lossless pairing and is a first-class migration endpoint in both directions.
- **@dekoza / @jphein — `update()` / multi-collection.** `update()` default + `supports_update` semantics were settled earlier; §2.4 now makes "N collections per palace, keyed by `collection_name`" a contract fact (the verbatim-vs-derivative split @jphein raised).

### @bensig's suggestions
All four folded in: §4.2 env naming keyed by backend *type* (hyphens→underscores), §3.3 global-`MEMPALACE_BACKEND`-overrides-detection note, §9 `BackendVersionMismatchError` example, §7.4/§7.3 "omit a kind rather than declare a no-op."

### Open questions (§12) — resolved
- `changes_since` takes an optional collection filter — **yes**.
- Per-palace capability variance — **no**, capabilities stay static/class-level (`supports_namespace_isolation` is a static declaration, not per-palace).
- `run_maintenance` — **observable, not fire-and-forget** (§7.3): returns `MaintenanceResult`, serializes concurrent same-kind calls, reports `already_running`. This is a direct vote from the lazy-index concurrency wedge @skuznetsov fixed with a session-level advisory lock — thank you for the field report; it's now the spec's reasoning.

### Tracked follow-up implementation (the parts #1679 deferred)
- #1724 — Embedder protocol + identity enforcement (§1.5/§5) · #1725 — maintenance hooks (§7.3) · #1726 — `searcher.py` backend-neutrality (§10).

Planning to merge once folks have had a chance to re-skim. @cschnatz — would especially like your read on the §4.4 capability-vs-blanket-MUST framing before I do.

cschnatz:
Re §4.4 — from the hosted multi-tenant seat, the capability framing is the right call, with one caveat.
Our hard tenant boundary rides on PalaceRef.id, which is already a universal MUST and covered by assert_partition_isolation. That’s what carries our isolation/compliance story, so we don’t need namespace to be a blanket MUST — and making it one would force minimal backends (sqlite_exact, embedded stores) to either fail conformance or fake a second axis they don’t have. Gating it behind supports_namespace_isolation keeps the contract honest.
Two things I’d want nailed down in §4.4 for the capability to be load-bearing for an operator:
	1.	A conformance arm for the capability — same-id / different-namespace isolation, distinct from the existing cross-id test. Otherwise supports_namespace_isolation is self-attestation we can’t rely on. Happy to contribute this arm to tests/_backend_conformance.py.
	2.	No silent no-op: a backend that does not advertise the capability MUST NOT silently accept and ignore a populated namespace — it should honor it or raise on use (same spirit as UnsupportedFilterError). Silently dropping it is a latent cross-namespace leak.
Keep id-partition isolation as the non-negotiable blanket MUST + test, make namespace a capability with its own conformance arm and the no-silent-drop clause, and §4.4 covers our case fully. 👍 to merge with that.

PR #730, opened 2026-04-12
fix: extract tool_use and tool_result blocks in Claude Code JSONL mining

arnoldwender:
## What does this PR do?

Fixes silent content loss when mining Claude Code `.jsonl` session files. Tool invocations and their outputs were being dropped because `_extract_content()` only handled `type=text` blocks.

For a typical Claude Code session (file reads, grep, code inspection), this caused ~49% content loss. Fixes #590.

## Root cause

`_extract_content()` in `normalize.py`:

```python
# Before — only text blocks extracted
elif isinstance(item, dict) and item.get("type") == "text":
    parts.append(item.get("text", ""))
# tool_use and tool_result blocks silently dropped ↑
```

Claude Code JSONL sessions contain assistant messages with mixed content like:
```json
[
  {"type": "text", "text": "Let me read that file."},
  {"type": "tool_use", "name": "Read", "input": {"file_path": "src/main.py"}},
  {"type": "tool_result", "content": "def main(): ..."}
]
```

The tool invocation and the file content were both dropped.

## Fix

**`_extract_content()`** — handles all three content block types:
- `tool_use` → appended as `[Tool: Read]` so context knows what was invoked
- `tool_result` → appended as plain text (may be a string or a list of text sub-blocks)

**`_try_claude_code_jsonl()`** — adds handling for top-level `type=tool` entries, the envelope Claude Code uses to carry tool results back to the assistant.

## How to test

Mine a Claude Code session that used tools (file read, grep, bash):
```bash
mempalace mine ~/.claude/projects/
mempalace search "any term from a file you read in that session"
```

Before: tool output missing from results. After: file content and tool context appear in search hits.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`)
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`)

PR #767, opened 2026-04-13
fix: extract tool_use and tool_result blocks in Claude Code JSONL mining

arnoldwender:
Rebased on develop (lint issue from convo_miner.py now fixed in #741).

Resolves #730.

Conflict in `normalize.py` resolved by keeping develop's implementation (HEAD), which already includes the `_format_tool_use`/`_format_tool_result` helpers and the multi-turn assistant merge logic — a superset of what the original PR added.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,8 +5,11 @@ Thanks for wanting to help. MemPalace is open source and we welcome contribution
 ## Getting Started
 
 ```bash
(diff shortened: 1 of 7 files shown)


arnoldwender:
Confirmed the fix landed in develop (mempalace/normalize.py handles tool_use/tool_result extraction). Closing — thanks for the follow-up.

PR #768, opened 2026-04-13
fix: log warnings on silent metadata exceptions in navigation tools

arnoldwender:
Rebased on develop (exception handling in mcp_server.py changed in recent merges).

Resolves #727.

Conflict in `mcp_server.py` resolved by keeping develop's implementation (HEAD), which already uses `logger.exception` and returns partial results with `result["error"]`/`result["partial"]` — a superset of what the original PR added.


arnoldwender:
Confirmed the fix landed in develop (hooks now skip tool_result when counting human exchanges). Closing — thanks for the follow-up.

PR #769, opened 2026-04-13
docs: add return type hints to knowledge_graph.py public methods

arnoldwender:
Clean replacement for #577, which was based on pre-thread-safety code.

Adds `-> str`, `-> str`, `-> None`, and `-> list[dict]` return type hints to the four public methods of `KnowledgeGraph`:
- `add_entity()` → `str`
- `add_triple()` → `str`
- `invalidate()` → `None`
- `query_entity()` → `list[dict]`

No logic changes. Valid in Python 3.9+ without `from __future__ import annotations`.

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -108,7 +108,7 @@ class KnowledgeGraph:
 
     # ── Write operations ──────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


arnoldwender:
Closing — branch is based on v3.2.0 and upstream has moved to v4-prep. Happy to open a fresh PR against current develop if still useful.

PR #778, opened 2026-04-13
feat: add Indonesian language support

zainphp:
## What does this PR do?

Introduces the Indonesian (id) locale, providing translations for CLI commands, status messages, and core terminology.

Includes language-specific regex patterns for stop words and action detection to support text processing and indexing in Indonesian. The test suite is updated with a sample case to verify correct dialect handling and compression.

## How to test

```bash
pytest mempalace/i18n/test_i18n.py -v
```

## Checklist

- [x] No hardcoded paths


bensig (approved):

igorls:
Thanks @dominosaurs! The `id.json` content is clean — schema matches `en.json`, interpolation variables all correct, regex patterns valid.

Two things before merge:

**1. CI is failing on a stale branch, not on your changes.** The failure is `test_version_consistency` asserting `'3.1.0' == '3.2.0'` — this branch predates the v3.2.0 release (#762) so your `mempalace/version.py` still says `3.1.0`. A rebase onto `develop` will fix it:

```bash
git fetch origin
git rebase origin/develop
git push --force-with-lease
```

**2. Please drop the edit to `mempalace/i18n/test_i18n.py`.** That file moves to `tests/test_i18n.py` via #758 (which will merge ahead of this). The test suite there auto-discovers every `*.json` in `mempalace/i18n/`, so the `"id"` sample is redundant — `id.json` will be exercised automatically by `test_all_languages_load` and `test_interpolation`.

Minor, non-blocking: the hyphenated verbs in `action_pattern` (`ter-konfigurasi|ter-deploy|ter-migrasi`) are valid regex but an unusual Indonesian form — if a native speaker happens to review, worth a second look. Not holding up the merge for it.

Optionally — #911 just landed infra for Indonesian-aware entity detection (so names like `Budi`, `Wayan`, `Siti` get extracted from prose). You can add an `entity` section to `id.json` with Indonesian person verbs, pronouns, dialogue markers, and stopwords. Totally optional — the CLI/AAAK work stands on its own.

dominosaurs:
Thanks for the review and the guidance.

I was honestly a bit clueless at first and approached `id.json` too narrowly as strict Indonesian-only translation. After looking at it more carefully, I realized that real Indonesian usage today is much more mixed, especially in technical writing. People naturally blend Indonesian with English terms, so a locale that is too rigid ends up less natural and can weaken mining on real text.

I’ve updated the JSON to be more contextual to how Indonesian is actually written today, with the hope that it improves mining quality for Indonesian text in practice. I’m very open to further suggestions from native speakers or anyone with better intuition here.

PR #779, opened 2026-04-13
fix(hooks): skip tool_result messages when counting human exchanges

### LINKED ISSUE #549: fix: save hook counts tool_result messages as human messages, inflating exchange count

arnoldwender:
## Summary

Fixes #549 — exchange count was inflated by counting `tool_result` messages as human turns.

In Claude Code JSONL, every tool response (Bash, Read, Grep, subagent results) is emitted with `role: "user"` and `content` as a list of `{"type": "tool_result", ...}` blocks. Both counting paths treated these as real human messages, so a session with 10 real turns but 50 tool calls was reported as 60 exchanges — causing premature save-hook triggers at the wrong frequency.

### Changes

- **`mempalace/hooks_cli.py` `_count_human_messages()`:** In the `elif isinstance(content, list):` branch, added an early `continue` that fires when all blocks in the list have `"type": "tool_result"`. The guard is placed before the existing `<command-message>` check, leaving that logic untouched.
- **`hooks/mempal_save_hook.sh`:** Added the identical `tool_result` guard to the inline Python script (also before the `<command-message>` check).

Both changes are purely additive — no existing counting logic was modified, only a new `continue` path inserted for `tool_result`-only content lists.

## Test plan

- [ ] All 32 existing `tests/test_hooks_cli.py` tests pass (verified locally)
- [ ] A new test `test_count_skips_tool_result_messages` should be added to directly exercise the new guard — noted for follow-up
- [ ] Manual: create a JSONL transcript with 10 real human messages plus 50 `tool_result` entries; confirm count returns 10, not 60

### DIFF
--- a/hooks/mempal_save_hook.sh
+++ b/hooks/mempal_save_hook.sh
@@ -104,6 +104,10 @@ with open(sys.argv[1]) as f:
                 content = msg.get('content', '')
                 if isinstance(content, str) and '<command-message>' in content:
                     continue
+                if isinstance(content, list) and content and all(
+                    isinstance(b, dict) and b.get('type') == 'tool_result' for b in content
+                ):
+                    continue
                 count += 1
         except:
             pass
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -60,7 +60,8 @@ def _count_human_messages(transcript_path: str) -> int:
                             # Skip pure tool_result messages (Bash/Read/Grep/subagent returns) — these have
                             # role "user" in Claude Code JSONL but are NOT real human turns (issue #549)
                             if content and all(
-                                isinstance(b, dict) and b.get("type") == "tool_result" for b in content
+                                isinstance(b, dict) and b.get("type") == "tool_result"
+                                for b in content
                             ):
                                 continue
                             text = " ".join(
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -57,6 +57,12 @@ def _count_human_messages(transcript_path: str) -> int:
                             if "<command-message>" in content:
                                 continue
                         elif isinstance(content, list):
+                            # Skip pure tool_result messages (Bash/Read/Grep/subagent returns) — these have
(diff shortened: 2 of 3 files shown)


arnoldwender:

PR #781, opened 2026-04-13
fix(kg): prevent edge duplication with partial UNIQUE index and atomic upsert

### LINKED ISSUE #655: fix(kg): Prevent edge duplication in Knowledge Graph

arnoldwender:
## Problem

`add_triple()` uses a SELECT-then-INSERT pattern with no database-level uniqueness guarantee — a classic TOCTOU race. Two concurrent callers can both execute the SELECT (both see no existing row), then both proceed to INSERT, producing duplicate active triples.

The existing `INSERT OR IGNORE` did not help because the `id` column is generated with `datetime.now().isoformat()` baked in — every call produces a unique primary key, so `INSERT OR IGNORE` never triggers.

There is also no `UNIQUE` constraint on `(subject, predicate, object) WHERE valid_to IS NULL`, so the database cannot enforce the invariant independently of Python logic.

Fixes #655.

## Solution

**1. Partial UNIQUE index** (`_init_db`):
```sql
CREATE UNIQUE INDEX IF NOT EXISTS uq_triples_active
    ON triples(subject, predicate, object)
    WHERE valid_to IS NULL;
```
This is the database-level enforcement layer — no concurrent caller can bypass it.

**2. Dedup migration** (also `_init_db`):
Before creating the index, remove any pre-existing duplicate active triples (keeping the oldest by `rowid`) so the `CREATE UNIQUE INDEX` does not fail on upgraded databases.

**3. Deterministic ID for active triples** (`add_triple`):
Active triples use `id_seed = f"{sub_id}|{pred}|{obj_id}|active"` so the computed `triple_id` is the same for both concurrent callers. `INSERT OR IGNORE` then discards the second insert atomically. Expired triples retain the `datetime.now()` component so re-adds after invalidation always generate a distinct row.

**4. Atomic upsert pattern**:
Replace SELECT-check + plain INSERT with INSERT OR IGNORE followed by a SELECT to return whichever row won (ours or a concurrent insert's).

## Test results

All existing `test_knowledge_graph.py` tests pass, including `test_duplicate_triple_returns_existing_id` and `test_invalidated_triple_allows_re_add`.

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -204,12 +204,17 @@ class KnowledgeGraph:
                         source_file,
                     ),
                 )
-                # Fetch whichever row won (ours or a concurrent insert's)
-                row = conn.execute(
-                    "SELECT id FROM triples WHERE subject=? AND predicate=? AND object=? AND valid_to IS NULL",
-                    (sub_id, pred, obj_id),
-                ).fetchone()
-        return row["id"] if row else triple_id
+                # For active triples, fetch whichever row won (ours or a concurrent
+                # insert's) — their IDs may differ if two threads raced here.
+                # Expired triples have no UNIQUE index constraint, so triple_id is
+                # always the inserted row.
+                if valid_to is None:
+                    winner = conn.execute(
+                        "SELECT id FROM triples WHERE subject=? AND predicate=? AND object=? AND valid_to IS NULL",
+                        (sub_id, pred, obj_id),
+                    ).fetchone()
+                    triple_id = winner["id"] if winner else triple_id
+        return triple_id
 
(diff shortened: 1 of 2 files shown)


arnoldwender:

PR #783, opened 2026-04-13
fix: add ensure_ascii=False to json.dumps in MCP server

OmkarKirpan:
## Summary

Closes #359

`json.dumps()` in the MCP server escapes non-ASCII characters by default, which causes surrogate encoding errors on Windows when processing Chinese/Korean/CJK content. This crashes the server with:

```
UnicodeEncodeError: 'utf-8' codec can't encode character '\udca3'
```

This PR adds `ensure_ascii=False` to all three `json.dumps()` calls in `mcp_server.py`:

1. **WAL entry logging** — non-ASCII content in tool params/results is preserved
2. **Tool result serialization** — CJK search results render correctly
3. **Response output** — the final JSON-RPC response writes CJK characters directly

This matches the pattern already used in `hooks_cli.py:95` and `config.py:193`.

## Test plan
- [x] All 60 MCP server tests pass
- [x] `ruff check` and `ruff format` clean
- [x] Manual: search for CJK content and verify results contain raw characters, not `\uXXXX` escapes

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -405,7 +405,7 @@ def _wal_log(operation: str, params: dict, result: dict = None):
     try:
         fd = os.open(str(_WAL_FILE), os.O_WRONLY | os.O_APPEND | os.O_CREAT, 0o600)
         with os.fdopen(fd, "a", encoding="utf-8") as f:
-            f.write(json.dumps(entry, default=str) + "\n")
+            f.write(json.dumps(entry, default=str, ensure_ascii=False) + "\n")
     except Exception as e:
         logger.error(f"WAL write failed: {e}")
 


bensig (approved):

OmkarKirpan:
Hi @bensig, 
Rebased onto develop and force-pushed - conflicts resolved, all tests passing. Ready to merge.

OmkarKirpan:
@bensig — rebased onto latest develop (32ec74d), clean rebase no conflicts, `test_mcp_server.py` 63/63 passing. Ready when you are.

OmkarKirpan:
cc @milla-jovovich — rebased on develop, approved by @bensig, tests green. Could you take a look when you have a minute?

SoloShine:
## Additional findings from Windows testing (v3.3.2 / v3.3.3)

We hit this exact bug in production on Windows 11. Confirming the root cause and sharing what we learned after a deep debug session.

### Confirmed root cause

On Windows, the MCP server's `stdin`/`stdout` use the system default encoding (e.g. cp1252, cp950) rather than UTF-8. When the MCP client sends a JSON-RPC request containing non-ASCII characters (★, |, CJK), the JSON decode may produce Python strings with **lone UTF-16 surrogates** (e.g. `\udc85`). These surrogates then crash `json.dumps()` with:

```
UnicodeEncodeError: 'utf-8' codec can't encode character '\udc85' in position 65: surrogates not allowed
```

### Why `ensure_ascii=False` alone is not enough on Windows

We tested `ensure_ascii=False` on its own and it **does not fully resolve** the issue on Windows. The surrogates originate from stdin/stdout encoding at the OS level, not just from `json.dumps` behavior. We found that a two-part fix is needed:

**1. Force UTF-8 on stdin/stdout at startup** (in `main()`):

```python
def main():
    _restore_stdout()
    sys.stdout.reconfigure(encoding="utf-8", errors="surrogatepass")
    sys.stdin.reconfigure(encoding="utf-8", errors="surrogatepass")
    ...
```

**2. Sanitize surrogates before serialization** (defense in depth):

```python
def _sanitize_surrogates(obj):
    """Remove lone UTF-16 surrogates from strings so json.dumps won't crash on Windows."""
    if isinstance(obj, str):
        return obj.encode("utf-8", errors="surrogatepass").decode("utf-8", errors="replace")
    if isinstance(obj, dict):
        return {k: _sanitize_surrogates(v) for k, v in obj.items()}
    if isinstance(obj, (list, tuple)):
        return [_sanitize_surrogates(v) for v in obj]
    return obj
```

Applied before `json.dumps(response)` in the main loop.

### What we also encountered (secondary issue)

While debugging, we also found that the ChromaDB write queue can accumulate stale entries (600+ unprocessed DELETE/UPDATE operations) with all `acquire_write` rows locked, which compounds the problem by causing `col.add()` to hang and timeout. This is likely unrelated to the encoding issue but worth noting — `mempalace repair rebuild` resolves it.

### Our current workaround

We're running v3.3.3 with both fixes patched locally. Happy to help test if this gets merged.

Thanks for the PR! 🙏

OmkarKirpan:
Rebased onto latest develop (5c9a5e0) and force-pushed (2264f01).

**Scope has shrunk significantly** — two of the three original `ensure_ascii=False` call sites already landed independently:

- Line 2416 (`json.dumps(result, ...)`) and line 2670 (stdout response write) — both covered by #7b49478 (May 1, @黄祖鑫)
- Windows stdio surface — covered by #35b033d (Apr 21, @alonehobo) via `sys.stdin.reconfigure(encoding="utf-8", errors="replace")` and same for stdout in `main()`

**What's left in this PR (1 line):** `ensure_ascii=False` on the WAL log `json.dumps` at line 408. The WAL writes UTF-8 to a file (`os.fdopen(..., encoding="utf-8")`) so without `ensure_ascii=False`, non-ASCII content gets escaped as `\uXXXX` in the WAL log even though the file can carry the original bytes. Minor cosmetic / consistency fix; matches the pattern @黄祖鑫 applied to the other two sites.

cc @SoloShine — your Apr 30 finding was correct that `ensure_ascii=False` alone isn't enough for Windows; the maintainers chose `errors="replace"` over `errors="surrogatepass"` for the stdio fix (lossy but always-valid-UTF-8 output, which is the right tradeoff for an MCP transport). Worth re-testing on Windows v3.4.x to confirm the -32000 errors are gone.

cc @bensig — happy to leave as-is for the trailing 1-line consistency, or close if you'd rather just merge as a side-channel cleanup. No strong preference.

Validation: `ruff check` clean, `ruff format --check` clean, WAL tests in `tests/test_mcp_server.py` passing.

SoloShine:
感谢您参与投稿，我们会尽快筛选出优质的的稿件内容，如果您的稿件入选，我们将会第一时间通知您

PR #793, opened 2026-04-13
feat: optional LLM-based closet regeneration — bring-your-own endpoint

igorls:
## Summary

Optional LLM-based closet regeneration, **vendor-agnostic**. User brings their own endpoint via `LLM_ENDPOINT` / `LLM_KEY` / `LLM_MODEL`. Regex closets remain the default — core memory stays API-free.

> **Stacked on #792** (full chain: #784 → #788 → #789 → #790 → #791 → #792).

## Context

The original bundled commit `935f657` included a `closet_llm.py` that imported `anthropic` directly and required `ANTHROPIC_API_KEY`. That conflicts with CLAUDE.md's **"Local-first, zero API"** principle and the Contributing guideline *"we do not accept features requiring API keys for core memory"*.

This PR **replaces** that module with a vendor-agnostic version that preserves the same prompt, parsing, and orchestration — only the transport was generalised. Milla is co-author on the commit; the prompt template and `regenerate_closets()` flow are her work.

## Configuration (env vars or CLI flags)

```bash
# Local (Ollama, zero cost)
LLM_ENDPOINT=http://localhost:11434/v1 \
LLM_MODEL=llama3:8b \
python -m mempalace.closet_llm --palace ~/.mempalace/palace

# Cloud (OpenAI, OpenRouter, etc.)
LLM_ENDPOINT=https://api.openai.com/v1 \
LLM_KEY=sk-... \
LLM_MODEL=gpt-4o-mini \
python -m mempalace.closet_llm --palace ~/.mempalace/palace

# CLI flags override env vars
python -m mempalace.closet_llm \
    --palace ~/.mempalace/palace \
    --endpoint http://localhost:11434/v1 \
    --model qwen2.5:7b
```

Any OpenAI-compatible `/chat/completions` endpoint works: Ollama, vLLM, llama.cpp server, OpenAI, OpenRouter, Azure OpenAI, Together, Groq, and so on.

## Design decisions

1. **OpenAI Chat Completions shape** — universal lingua franca across providers
2. **Zero new dependencies** — uses stdlib `urllib.request`; no `anthropic`, `openai`, `requests`, or `httpx` added to `pyproject.toml`
3. **Key is optional** — local inference servers typically don't authenticate. `Authorization` header only set when a key is provided.
4. **Graceful failure on missing config** — CLI prints a clear message listing what's missing; `regenerate_closets()` returns `{"error": "missing-config", "missing": [...]}`
[...]

PR #795, opened 2026-04-13
fix(search): hybrid closet+drawer retrieval — closets boost, never gate

igorls:
## Summary

`search_memories()` now queries drawers and closets **in parallel** and treats closets as a **ranking signal**, not a gate. Fixes a recall regression introduced by the closet-first routing in #788 when closets are weak (regex extraction on narrative content).

> **Stacked on #793.** Full chain: #784 → #788 → #789 → #790 → #791 → #792 → #793 → **this**.
>
> This PR amends #788's search behavior. Reviewers of #788 should look here first.

## The regression this fixes

Closet-first routing works well when closets extract meaningful topics (code, technical content). It regresses recall on narrative content where regex patterns ("built", "fixed", "decided") don't match and closets fall back to just `wing/room||→drawer`.

Measured on the spontaneous fixture (Mira's life — 40 narrative drawers, 20 ground-truth queries):

| Variant | R@1 | R@5 | Notes |
|---|---|---|---|
| develop (direct drawer, no closets) | 0.42 | **1.00** | baseline |
| #788 closet-first, regex closets | 0.25 | 0.50 | **regression** — default user experience |
| #788 closet-first, LLM closets (via #793) | 0.67 | 0.92 | partial fix, requires opt-in |
| **hybrid + regex closets (this PR)** | **0.58** | **1.00** | safe floor |
| **hybrid + LLM closets (this PR)** | **0.58** | **1.00** | strictly better than baseline |

R@1 improvement: +38% relative on regex closets (default UX), same on LLM closets. R@5 locked at 1.00 — **no weak-closets regression possible**.

Per-turn: hybrid wins 3 queries develop missed (T-03, T-04, T-05 — all implicit-relevant where direct search had the primary drawer ranked but not at #1). One loss vs develop (T-06) — closet boost flipped a different relevant drawer above the primary. Net +2 R@1, -0 R@5.

## How it works

```python
# Before: closet-first (#788)
closet_hits = closets.query(...)
if closet_hits:
    hydrate drawers from closet pointers
else:
    fall back to direct drawer query

# After: hybrid (this PR)
drawer_hits = drawers.query(..., n_results=n*3)   # always query — the floor
closet_hits = closets.query(...)                  # also query for boost signal
for drawer in drawer_hits:
    if drawer.source_file matches top-N closets and closet_dist < 1.5:
        apply CLOSET_RANK_BOOSTS[rank] = [0.40, 0.25, 0.15, 0.08, 0.04]
sort by effective_distance, take top-K, drawer-grep enrichment, BM25 rerank
```

Key design choices:

1. **Rank-based boost, not distance-based** — on narrative content, absolute closet distances are noisy (~1.2–1.5 range), but the ordinal "which closet matched best" is reliable. Top-1 gets a big boost (0.40), fading to 0.04 at rank 5.
2. **`CLOSET_DISTANCE_CAP = 1.5`** — don't apply the boost at all if the best closet match is weak. Prevents pathological cases where a genuinely-wrong closet still hits top-1.
3. **Drawer-grep enrichment still fires** on closet-boosted hits whose source has multiple drawers. Chunk-expansion behavior users relied on is preserved.
4. **BM25 rerank (from #789) still runs last** — unchanged.

## Test plan

- [x] **5 new unit tests** in `tests/test_hybrid_search.py`, including the critical invariant test:
  - `test_weak_closets_do_not_hide_direct_drawer_hits` — seeds a misleading closet pointing at the wrong drawer, confirms the correct drawer still appears in results
  - `test_closet_boost_lifts_matching_drawer` — confirms the boost does actually lift a matching drawer to rank 1 when the closet agrees
  - `test_no_closets_degrades_to_direct_drawer_search` — pre-closet palaces still work
  - `test_closet_preview_exposed_when_boosted` — metadata traceability
  - `test_drawer_only_hits_have_no_closet_preview` — clean separation
- [x] Full suite: **726/726 pass** (2 version-consistency tests deselected — pre-existing develop bug).
- [x] Benchmark validation against the 20-query narrative fixture (results above).

## Callouts for reviewers

1. **Fundamentally changes #788's search path.** If #788 is being reviewed in isolation, please evaluate the combined behavior with this PR — the standalone #788 has the regression documented above.
2. **Constants `CLOSET_RANK_BOOSTS`, `CLOSET_DISTANCE_CAP`** are hardcoded. Reasonable defaults from the benchmark sweep, but could become config in a follow-up if needed.
3. **Drawer-grep enrichment** is source-file matched by suffix (`source_file.endswith(source_name)`) — correct for the current source-file naming but slightly fragile. Worth a cleaner path in refactor.
4. **Stack tip** — after this lands, the v3.3 search path is strictly better than develop on narrative content, zero regression, with a clean upgrade curve when users opt into LLM closets.

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -183,138 +183,156 @@ def search_memories(
 
     where = build_where_filter(wing, room)
 
-    # Try closet-first search: search the compact index, then hydrate drawers
-    closet_hits = []
+    # Hybrid retrieval: always query drawers directly (the floor), then use
+    # closet hits to boost rankings. Closets are a ranking SIGNAL, never a
+    # GATE — direct drawer search is always the baseline.
+    #
+    # This avoids the "weak-closets regression" where narrative content
+    # produces low-signal closets (regex extraction matches few topics)
+    # and closet-first routing hides drawers that direct search would find.
+    try:
+        dkwargs = {
+            "query_texts": [query],
+            "n_results": n_results * 3,  # over-fetch for re-ranking
+            "include": ["documents", "metadatas", "distances"],
+        }
+        if where:
+            dkwargs["where"] = where
+        drawer_results = drawers_col.query(**dkwargs)
+    except Exception as e:
+        return {"error": f"Search error: {e}"}
+
+    # Gather closet hits (best-per-source) to build a boost lookup.
+    closet_boost_by_source = {}  # source_file -> (rank, closet_dist, preview)
     try:
         closets_col = get_closets_collection(palace_path, create=False)
         ckwargs = {
             "query_texts": [query],
-            "n_results": n_results * 2,  # over-fetch closets to find best drawers
+            "n_results": n_results * 2,
             "include": ["documents", "metadatas", "distances"],
         }
         if where:
             ckwargs["where"] = where
         closet_results = closets_col.query(**ckwargs)
-        if closet_results["documents"][0]:
-            closet_hits = list(zip(
+        for rank, (doc, meta, dist) in enumerate(
+            zip(
                 closet_results["documents"][0],
                 closet_results["metadatas"][0],
                 closet_results["distances"][0],
-            ))
+            )
+        ):
+            source = meta.get("source_file", "")
+            if source and source not in closet_boost_by_source:
+                closet_boost_by_source[source] = (rank, dist, doc[:200])
     except Exception:
-        pass  # no closets yet — fall through to direct drawer search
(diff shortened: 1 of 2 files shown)


milla-jovovich:
Thank you SO MUCH. You're so kind to stay up all night on the other side of the world with me staying up on this side to get this right. Thank you Igor. Ben knew who he was bringing onboard to check the system. Now we both need to sleep lol!

PR #796, opened 2026-04-13
fix: batch upserts in miner to prevent ChromaDB 1.5.x compaction crashes

IzmanIzy:
## Summary

- **Batch all chunks per file into a single `collection.upsert()` call** instead of upserting each chunk individually. This reduces WAL write pressure that causes the Rust compactor in ChromaDB >= 1.5 to crash.
- **Add periodic checkpoints** (every 200 files) that release and re-acquire the collection, giving the compactor time to flush background work.
- **Release collection reference** at the end of mining for a clean shutdown.

## Problem

When mining projects with 100+ files, the miner issues thousands of individual upserts. On ChromaDB 1.5.x this causes:

1. **Segfault (exit code 139)** — the Rust compactor corrupts the metadata segment during concurrent individual writes
2. **`InternalError: Failed to apply logs to the metadata segment`** — WAL entries accumulate faster than compaction can process them

Both errors are **intermittent** and depend on project size, making them hard to reproduce in small test suites but consistent on real-world knowledge bases (300+ files).

## Root Cause

ChromaDB's Rust compactor (introduced in 1.5.x) runs in a background thread. Individual upserts create one WAL entry each, and 2000+ entries in rapid succession overwhelm the compactor's ability to merge them atomically. The previous code already had a comment about hnswlib's thread-unsafe `updatePoint` path causing segfaults on macOS ARM — this is the same class of bug on the compaction side, now affecting all platforms.

## Testing

- **37 existing miner tests pass** (`pytest tests/ -k "mine"` — 37 passed, 0 failed)
- **Verified on a real 406-file knowledge base** (3026 drawers) with ChromaDB 1.5.7 — zero crashes, clean completion with two checkpoint flushes at file 200 and 400
- Lint (`ruff check`) and format (`ruff format --check`) pass
- No changes to public API or CLI interface

## Test plan

- [x] `ruff check .` passes
- [x] `ruff format --check .` passes
- [x] `pytest tests/ -v --ignore=tests/benchmarks -k "mine"` — 37 passed
- [x] Manual test: mine 406-file project with ChromaDB 1.5.7 — 3026 drawers, 0 crashes
- [ ] Verify re-mining (modified files) still works correctly
- [ ] Test with ChromaDB 0.6.x to confirm backward compatibility

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -9,6 +9,7 @@ Stores verbatim chunks as drawers. No summaries. Ever.
 
 import os
 import sys
+import time
 import hashlib
 import fnmatch
 from pathlib import Path
@@ -531,34 +532,57 @@ def _extract_entities_for_metadata(content: str) -> str:
     return ";".join(capped)
 
 
+def _build_drawer_payload(
+    wing: str, room: str, content: str, source_file: str, chunk_index: int, agent: str
+) -> tuple:
+    """Build a (drawer_id, content, metadata) triple for a single drawer.
+
+    Shared by :func:`add_drawer` (single-drawer API) and :func:`process_file`
+    (batched per-file upserts). Centralising the drawer ID + metadata
+    construction keeps the two paths in sync — any future metadata additions
+    land in one place.
+    """
+    drawer_id = f"drawer_{wing}_{room}_{hashlib.sha256((source_file + str(chunk_index)).encode()).hexdigest()[:24]}"
+    metadata = {
+        "wing": wing,
+        "room": room,
+        "source_file": source_file,
+        "chunk_index": chunk_index,
+        "added_by": agent,
+        "filed_at": datetime.now().isoformat(),
+        "normalize_version": NORMALIZE_VERSION,
+    }
+    # Store file mtime so we can detect modifications later.
+    try:
+        metadata["source_mtime"] = os.path.getmtime(source_file)
+    except OSError:
+        pass
+    # Tag with hall for graph connectivity within wings
+    metadata["hall"] = detect_hall(content)
+    # Tag with entity names for filterable search
+    entities = _extract_entities_for_metadata(content)
+    if entities:
+        metadata["entities"] = entities
+    return drawer_id, content, metadata
+
+
 def add_drawer(
     collection, wing: str, room: str, content: str, source_file: str, chunk_index: int, agent: str
 ):
-    """Add one drawer to the palace."""
-    drawer_id = f"drawer_{wing}_{room}_{hashlib.sha256((source_file + str(chunk_index)).encode()).hexdigest()[:24]}"
+    """Add one drawer to the palace.
+
+    Thin single-drawer wrapper around :func:`_build_drawer_payload` + upsert.
+    Kept for callers (tests, external integrations) that file one drawer at
+    a time. The mining hot path uses :func:`process_file` which batches all
+    chunks from one file into a single ``collection.upsert()`` call — see the
+    batching rationale in ``process_file``.
+    """
+    drawer_id, doc, metadata = _build_drawer_payload(
+        wing, room, content, source_file, chunk_index, agent
+    )
     try:
-        metadata = {
-            "wing": wing,
-            "room": room,
-            "source_file": source_file,
-            "chunk_index": chunk_index,
-            "added_by": agent,
-            "filed_at": datetime.now().isoformat(),
-            "normalize_version": NORMALIZE_VERSION,
-        }
-        # Store file mtime so we can detect modifications later.
-        try:
-            metadata["source_mtime"] = os.path.getmtime(source_file)
-        except OSError:
-            pass
-        # Tag with hall for graph connectivity within wings
-        metadata["hall"] = detect_hall(content)
-        # Tag with entity names for filterable search
-        entities = _extract_entities_for_metadata(content)
-        if entities:
-            metadata["entities"] = entities
         collection.upsert(
-            documents=[content],
+            documents=[doc],
             ids=[drawer_id],
             metadatas=[metadata],
         )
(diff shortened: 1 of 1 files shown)


jphein:
This addresses the same WAL pressure issue we tackled in #629 — worth noting that PR also adds bulk mtime pre-fetch (`bulk_check_mined()`) and optional concurrent mining via `--workers`. The single-file batch approach here is cleaner for a targeted fix though. Might be worth coordinating so the two PRs don't conflict — happy to rebase #629 on top of this if it merges first.

IzmanIzy:
This addresses the same WAL pressure issue we tackled in #629 — worth noting that PR also adds bulk mtime pre-fetch (`bulk_check_mined()`) and optional concurrent mining via `--workers`. The single-file batch approach here is cleaner for a targeted fix though. Might be worth coordinating so the two PRs don't conflict — happy to rebase #629 on top of this if it merges first.

messelink:
Hi @IzmanIzy — I hit the exact compaction crash this PR fixes while mining into an existing palace that had accumulated BLOB-typed `seq_id` rows from a stale long-running MCP server (see #899 for that adjacent bug). The batching approach here solved my immediate problem, so thank you for writing it up.

I wanted to flag that this PR has a non-trivial rebase conflict against current `develop`: since it was opened, `add_drawer()` on develop has grown three new metadata fields that your PR's inlined version in `process_file()` doesn't reproduce — a verbatim cherry-pick would silently regress them. Specifically:

- `normalize_version` (from `palace.NORMALIZE_VERSION`)
- `hall` (from `detect_hall(content)`)
- `entities` (from `_extract_entities_for_metadata(content)`)

The `source_mtime` try/except is also now in the common path.

I rebased your approach onto current `develop` and adapted it to preserve those fields. Rather than inlining metadata construction in `process_file()`, I extracted a `_build_drawer_payload()` helper and kept `add_drawer()` as a thin single-drawer wrapper (for backwards compat with the `tests/test_miner.py::test_add_drawer` and `tests/test_hall_detection.py::test_add_drawer_includes_hall` tests, which call `add_drawer` directly and check for `normalize_version` / `hall` in the stored metadata). The batched per-file upsert and the periodic checkpoint logic are kept as you wrote them.

**Commit:** [messelink/mempalace@9cc1365](https://github.com/messelink/mempalace/commit/9cc1365)
**Branch:** [`messelink:fix/batch-upsert-chromadb-compaction`](https://github.com/messelink/mempalace/tree/fix/batch-upsert-chromadb-compaction) (clean off current `upstream/develop`)

### Test I ran

A delete-and-remine of a 20-file subtree (mix of markdown docs and chat transcripts) into an existing ~14.8k-drawer palace. **Result:** 1,324 drawers filed in a single end-to-end run without any compaction errors. The largest single-file batch was 618 drawers in one `collection.upsert()` call — exactly the stress case your PR targets. Search against the re-mined content works correctly afterward.

Previously (without the batching), the same mine crashed with `Error in compaction: Error reading from metadata segment reader: error occurred while decoding column 0: mismatched types; Rust type u64 (as SQL type INTEGER) is not compatible with SQL type BLOB` on the first upsert, and I lost a chunk of drawers to silent file-already-mined skipping on retry.

### How to proceed

Totally your call — I don't want to step on your PR. Three options:

1. **You fetch my branch and apply it to yours** — `git fetch https://github.com/messelink/mempalace.git fix/batch-upsert-chromadb-compaction && git reset --hard FETCH_HEAD` on your local `fix/batch-upsert-chromadb-compaction`, then force-push. Cleanest attribution (your PR, just rebased). The commit already credits you as `Co-Authored-By`.
2. **You enable "Allow edits from maintainers"** on this PR, and I can't push directly (I'm not a maintainer of `milla-jovovich/mempalace`), but one of the maintainers could push my patch to your branch.
3. **If you're unavailable or prefer**, I'm happy to open a separate PR that supersedes this one, credits you, and links back here.

Let me know which you prefer. Happy to wait a few days for a reply before doing option 3.

cc @milla-jovovich for awareness — this PR becomes unblocked with the rebase.

bensig:
hey @IzmanIzy — this conflicts with develop now. pls rebase and we can merge. thanks!

bensig:
@IzmanIzy pls check conflict

messelink:
@bensig — just flagging that my comment above (03:26 UTC) has the rebased version ready if it helps move this along. Branch: [`messelink:fix/batch-upsert-chromadb-compaction`](https://github.com/messelink/mempalace/tree/fix/batch-upsert-chromadb-compaction), clean off current `develop`. Up to @IzmanIzy whether they want to pull it into their branch or I open a new PR.

IzmanIzy:
Rebased onto current `develop` using @messelink's cleaned-up version (thanks! 🙏). Conflict resolved, ready to merge.

@bensig good to go.

messelink:
@IzmanIzy heads-up — looks like the failure modes this PR was preventing have been addressed in other paths since your rebase. As of current `develop`:

- **Per-file batched upserts** — landed differently: `DRAWER_UPSERT_BATCH_SIZE = 1000` sub-batches in `process_file` via `_build_drawer_metadata`.
- **macOS-ARM hnswlib `updatePoint` segfault** — handled by `delete-before-upsert` ahead of the batched insert.
- **ChromaDB 1.5.x metadata-segment / WAL-pressure crashes** — mitigated by #1173 (auto-quarantine on `make_client`), #1191 (`batch_size` + `sync_threshold` metadata), #1135 (max-seq-id shim), and #976 (HNSW race + mine fan-out).

The one piece *not* in develop is the 200-file periodic collection release/re-acquire from this PR. If you've seen crashes that only the checkpoint specifically catches, worth flagging here. Otherwise this PR may be safe to close.

igorls:
Thanks! Per-file upsert batching already landed on develop (DRAWER_UPSERT_BATCH_SIZE), which covers the 1.5.x compaction crash; the remaining 200-file checkpoint here is speculative. Closing as superseded — please reopen if you still hit the crash on current develop.

PR #803, opened 2026-04-13
fix: paginate status() metadata fetch to avoid SQLite variable limit

### LINKED ISSUE #802: `mempalace status` crashes with "too many SQL variables" on large palaces
## Bug

`mempalace status` crashes with:

```
sqlite3.OperationalError: too many SQL variables
```

## Root cause

`miner.py` calls `col.get(limit=total, include=["metadatas"])` without pagination. SQLite binds one host variable per metadata item, hitting the compile-time limit (~32766) when the palace exceeds ~30k drawers.

## Reproduction threshold

Any palace with more than ~30k drawers. Confirmed on a palace with 50k+ drawers.

## Workaround

```bash
sqlite3 ~/.mempalace/palace/chroma.sqlite3 "SELECT COUNT(*) FROM embeddings;"
```

`mempalace search` continues to work correctly.

## Fix

Paginate the `col.get()` call in batches of 5000. ChromaDB's `get()` accepts an `offset` parameter, so this is straightforward.

See PR from `srgvg/mempalace` for the one-function fix.

srgvg:
## Summary

- `mempalace status` crashes with `sqlite3.OperationalError: too many SQL variables` on palaces with >~30k drawers
- Root cause: `col.get(limit=total, include=["metadatas"])` passes the full count as limit; SQLite binds one variable per metadata item and hits the compile-time limit (~32766)
- Fix: paginate in batches of 5000 using the existing `offset` parameter on `col.get()`

Fixes #802

## Test plan

- [ ] `python -m pytest tests/ -v --ignore=tests/benchmarks` — 687 tests pass (2 pre-existing failures in `test_version_consistency.py` unrelated to this change)
- [ ] `ruff check .` — clean
- [ ] Manual: `mempalace status` on a palace with 50k+ drawers completes without error

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -631,8 +631,12 @@ def status(palace_path: str):
 
     # Count by wing and room
     total = col.count()
-    r = col.get(limit=total, include=["metadatas"]) if total else {"metadatas": []}
-    metas = r["metadatas"]
+    metas = []
+    if total:
+        _batch = 5000
+        for _off in range(0, total, _batch):
+            _r = col.get(limit=_batch, offset=_off, include=["metadatas"])
+            metas.extend(_r["metadatas"])
 
     wing_rooms = defaultdict(lambda: defaultdict(int))
     for m in metas:


jphein:
Can confirm this is a real issue — we hit it with a 134K-drawer palace. Our fork has the same paginated approach in `miner.py:status()` (batches of 10K with offset). Included in #629 alongside the batch writes, but this standalone fix is much easier to review. 👍

bensig:
closing in favor of #851 which covers the same status pagination fix and references both #802 and #850. thanks @srgvg!

PR #804, opened 2026-04-13
feat: configurable auto-save interval with independent precompact control

### LINKED ISSUE #494: Feature Request: Add a "Silent Mode" / Background Processing for the Stop Hook (Auto-save)

rusel95:
Closes #494

## Summary

Replaces the hardcoded `SAVE_INTERVAL = 15` with a user-configurable `hooks.save_interval` and adds independent `hooks.precompact` control. This is an alternative to #711 that provides **granular frequency control** instead of a binary on/off toggle.

## Why this is better than #711

| | #711 (binary toggle) | This PR (interval + independent precompact) |
|---|---|---|
| **Stop hook control** | `auto_save: true/false` — all or nothing | `save_interval: 30` — any frequency (0 = off) |
| **Precompact** | Disabled together with stop hook — **data loss risk** | Independent toggle — precompact stays on by default |
| **Config duplication** | Same check in 2 bash scripts + Python (3 places) | Python only — bash scripts are thin wrappers |
| **User flexibility** | "Too noisy? Turn it off entirely" | "Too noisy? Set interval to 50 instead of 15" |

## Config

```json
// ~/.mempalace/config.json
{
  "hooks": {
    "save_interval": 30,
    "precompact": true
  }
}
```

| Setting | Default | Env var | Effect |
|---------|---------|---------|--------|
| `save_interval` | `15` | `MEMPALACE_HOOKS_SAVE_INTERVAL` | Messages between auto-saves. `0` = disabled |
| `precompact` | `true` | `MEMPALACE_HOOKS_PRECOMPACT` | Block before context compaction. Last chance to save |

**Key design decision:** `precompact` defaults to `true` and is controlled separately. Setting `save_interval: 0` disables periodic saves but precompact still fires before context is lost. This prevents the data loss scenario where #711's `auto_save: false` silently disables both hooks.

## Changes

| File | What |
|------|------|
| `mempalace/config.py` | `hooks_save_interval` and `hooks_precompact` properties with env var overrides |
| `mempalace/hooks_cli.py` | Reads interval from config instead of hardcoded constant; precompact checks its own toggle; shortened block reasons naming exact MCP tools |
| `tests/test_config.py` | 11 new tests: interval default/config/env/zero/negative, precompact default/disabled/env, independence |
| `tests/test_hooks_cli.py` | 7 new tests: stop disabled by zero interval, custom intervals (5, 50), precompact disabled/enabled, stop+precompact independence |

## Test plan

- [x] `test_hooks_save_interval_default` — default is 15
- [x] `test_hooks_save_interval_from_config` — reads from config.json
- [x] `test_hooks_save_interval_zero_disables` — 0 disables stop hook
- [x] `test_hooks_save_interval_env_override` — env var overrides config
- [x] `test_hooks_save_interval_env_zero` — env var can set to 0
- [x] `test_hooks_save_interval_negative_clamped` — negative clamped to 0
- [x] `test_hooks_precompact_default` — default is true
- [x] `test_hooks_precompact_disabled` — can be disabled via config
- [x] `test_hooks_precompact_env_override` — env var overrides
- [x] `test_hooks_save_interval_and_precompact_independent` — disabling one doesn't affect the other
- [x] `test_stop_hook_disabled_by_zero_interval` — stop hook passes through
- [x] `test_stop_hook_custom_interval` — save_interval=5 triggers at 5 messages
- [x] `test_stop_hook_custom_interval_not_reached` — save_interval=50 doesn't trigger at 15
- [x] `test_precompact_disabled_by_config` — precompact passes through when disabled
- [x] `test_precompact_enabled_by_default` — precompact blocks even when stop is disabled
- [x] `test_stop_disabled_precompact_still_works` — independence verified
- [x] All 59 existing hook/config tests pass (0 regressions)

```
$ python -m pytest tests/test_config.py tests/test_hooks_cli.py -v
============================== 59 passed in 0.10s ==============================
```


Naoray:
Strong +1 on this. Hitting the pain point today on 3.3.0:

- Hardcoded `SAVE_INTERVAL = 15` fires way too often on conversational/admin sessions — had to locally patch to 50.
- `hooks.precompact` being independently toggleable is the right call. The current always-block precompact (prior to #863 / still in 3.3.0 release) permanently broke `/compact` in long sessions; config toggle would let users opt out without losing the stop-hook save.

The design choice of keeping `precompact: true` by default when `save_interval: 0` is exactly right — no silent data loss. #711's binary toggle would conflate these.

Would love to see this land. Happy to smoke-test against 3.3.0 if useful.

rusel95:
Pushed cleanup (c2c2f6a):
- Removed 5 stray personal notes files that got in via a local merge conflict. PR diff is now minimal: `.gitignore`, `config.py`, `hooks_cli.py`, plus tests.
- All 84 tests pass locally (`pytest tests/test_config.py tests/test_hooks_cli.py`).
- Mergeable against `develop`, no conflicts.

### Re-pinging because this PR addresses several now-filed bugs

The independent `hooks.precompact` toggle is the reason I split it from `save_interval` in the first place — it directly covers the recently reported PreCompact blocking incidents:

- #949 — PreCompact hook blocks indefinitely (no escape)
- #941 — PreCompact should use `systemMessage` instead of blocking
- #906 — PreCompact prevents compaction at context limit
- #955 — Compaction Blocking
- #494 — Silent Mode / Background Processing (originally filed)

With this PR:
- `save_interval: 50` quiets the stop-hook without silencing data-preservation paths.
- `precompact: false` lets users disable precompact **independently** if it's misbehaving — without also losing the periodic save (or vice versa).
- That's the key difference vs. #985 (env-var-only, no precompact control) and #711 (binary all-or-nothing).

Design note for @igorls @bensig @milla-jovovich: the two controls live in `config.py` (properties) with env-var overrides, so both CLI config-file and per-session env-var workflows work. No new bash-side logic — the hooks stay thin wrappers.

Happy to rebase or adjust naming (e.g. `MEMPALACE_HOOKS_SAVE_INTERVAL` → `MEMPAL_SAVE_INTERVAL` to match #985's naming) if that smooths merge.

rusel95:
Closing in favour of two smaller replacements that are easier to review independently.

The original PR combined save_interval configurability and precompact toggle in one diff. Splitting them so each can be reviewed and merged on its own timeline.

**Replacements:**
- **feat/hooks-save-interval** — configurable save_interval via config.json + env var. ~35 lines. Closes #494.
- **feat/hooks-precompact-toggle** — independent precompact boolean toggle. ~30 lines. Closes #949 / #906.

The key design decision (the two controls are independent — disabling save_interval does NOT disable precompact) is preserved in both replacements.

PR #304, opened 2026-04-08
fix: add hnsw:space=cosine to drawer collection creation

dr-robert-li:
## What does this PR do?

Adds `metadata={'hnsw:space': 'cosine'}` to both `create_collection` call sites (`mempalace/miner.py` and `mempalace/convo_miner.py`). `searcher.py` computes `similarity = 1 - distance`, which only yields a meaningful score in `[0, 1]` when the underlying distance metric is cosine. ChromaDB defaults to L2 when no metadata is supplied, producing negative similarity scores for every search result.

Fixes #218.

## How to test

```bash
python -m pytest tests/test_miner.py::test_get_collection_uses_cosine_distance                  tests/test_convo_miner.py::test_get_collection_uses_cosine_distance -v
```

Existing collections are unaffected (the metric is set at creation time). Users with a palace already mined will need to wipe and re-mine to benefit.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — 119 passed, 0 failed
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`)

PR #807, opened 2026-04-13
Fix: set cosine distance metadata on all collection creation sites

### LINKED ISSUE #218: Collection created without hnsw:space=cosine causes negative similarity scores

sha2fiddy:
## Summary

- Add `metadata={"hnsw:space": "cosine"}` to all 4 production `create_collection`/`get_or_create_collection` call sites that were missing it (`backends/chroma.py`, `cli.py` x2, `migrate.py`)
- Fix 3 test fixtures to also create collections with cosine distance (`conftest.py`, `test_mcp_server.py`, `test_miner.py`)
- Add explicit test verifying `ChromaBackend.get_collection(create=True)` produces a cosine-distance collection

Closes #218. Supersedes stale branch `feature/fix-cosine-distance` (170 files divergent, pre-backend-refactor).

## Upgrading existing palaces

**Existing palaces created before this fix use L2 (Euclidean) distance and will not self-heal from this merge.** The `hnsw:space` setting is baked in at collection-creation time. After upgrading, run `mempalace repair` to rebuild the collection with cosine distance.

## Context

ChromaDB defaults HNSW index to L2 (Euclidean) distance. MemPalace scoring computes `1 - distance`, which only produces valid similarity scores under cosine distance (range 0–2). Under L2, distances are unbounded, producing negative scores. Only `mcp_server.py` and `repair.py` were passing the cosine metadata.

## Note on `metadata=` vs `configuration=`

This PR uses the legacy `metadata={"hnsw:space": "cosine"}` syntax intentionally. ChromaDB 0.6.3+ introduced a newer `configuration=` API, but it requires a `CollectionConfigurationInternal` object — passing a plain dict doesn't work in 0.6.x. Since the repo pins `chromadb>=0.5.0,<0.7`, the legacy `metadata=` approach is the only one that works across the full supported range. When the chromadb pin is bumped (e.g. PR #664), these calls should be migrated to the `configuration=` API.

## Test plan

- [x] `pytest tests/ -x -v` — 685 passed (2 pre-existing version-mismatch failures unrelated to this change)
- [x] `ruff check mempalace/ tests/` — all checks passed (formatted with CI-pinned ruff 0.4.x)
- [x] Sweep: every `create_collection`/`get_or_create_collection` call (excluding benchmarks) now has `metadata=` or is read-only `get_collection`
- [x] New test `test_chroma_backend_creates_collection_with_cosine_distance` verifies metadata via raw `chromadb.PersistentClient`
- [x] `mempalace repair` confirmed to rebuild collections with `metadata={"hnsw:space": "cosine"}` (cli.py:247)

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -35,8 +35,13 @@ def _fix_blob_seq_ids(palace_path: str):
                     continue
                 if not rows:
                     continue
-                updates = [(int.from_bytes(blob, byteorder="big"), rowid) for rowid, blob in rows]
-                conn.executemany(f"UPDATE {table} SET seq_id = ? WHERE rowid = ?", updates)
+                updates = [
+                    (int.from_bytes(blob, byteorder="big"), rowid)
+                    for rowid, blob in rows
+                ]
+                conn.executemany(
+                    f"UPDATE {table} SET seq_id = ? WHERE rowid = ?", updates
+                )
                 logger.info("Fixed %d BLOB seq_ids in %s", len(updates), table)
             conn.commit()
     except Exception:
@@ -71,7 +76,9 @@ class ChromaCollection(BaseCollection):
 class ChromaBackend:
     """Factory for MemPalace's default ChromaDB backend."""
 
-    def get_collection(self, palace_path: str, collection_name: str, create: bool = False):
+    def get_collection(
+        self, palace_path: str, collection_name: str, create: bool = False
+    ):
         if not create and not os.path.isdir(palace_path):
             raise FileNotFoundError(palace_path)
 
@@ -85,7 +92,9 @@ class ChromaBackend:
         _fix_blob_seq_ids(palace_path)
         client = chromadb.PersistentClient(path=palace_path)
         if create:
-            collection = client.get_or_create_collection(collection_name)
+            collection = client.get_or_create_collection(
+                collection_name, metadata={"hnsw:space": "cosine"}
+            )
         else:
             collection = client.get_collection(collection_name)
         return ChromaCollection(collection)
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -35,13 +35,8 @@ def _fix_blob_seq_ids(palace_path: str):
                     continue
                 if not rows:
                     continue
-                updates = [
-                    (int.from_bytes(blob, byteorder="big"), rowid)
-                    for rowid, blob in rows
-                ]
-                conn.executemany(
-                    f"UPDATE {table} SET seq_id = ? WHERE rowid = ?", updates
-                )
+                updates = [(int.from_bytes(blob, byteorder="big"), rowid) for rowid, blob in rows]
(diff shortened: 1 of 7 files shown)


milla-jovovich:
@sha2fiddy thank so much for catching this! 

The fix is correct. All 4 missing sites verified: backends/chroma.py:88, cli.py:243 (repair), cli.py:389 (compressed),
   migrate.py:210. The metadata= vs configuration= reasoning is solid for the current chromadb pin.

  CI needs a small cleanup. The lint failures are just formatting — looks like a newer local ruff reformatted things the
   CI-pinned version doesn't expect. Can you re-run with the pinned version?

  uvx --from 'ruff>=0.4.0,<0.5' ruff format .
  uvx --from 'ruff>=0.4.0,<0.5' ruff check .

  Ideally the final diff is just the metadata={"hnsw:space": "cosine"} additions with no extra reformatting — keeps it
  clean and easy to review. The version consistency test failures are pre-existing on develop, not yours.

  One thing to call out for users: Since hnsw:space is set at collection-creation time, existing palaces on L2 won't
  self-heal from this merge. They'll need mempalace repair or a re-mine. Worth adding a bolded "Existing palaces: run
  mempalace repair after upgrading" note so people don't miss it.

  Once the formatting is cleaned up, this is good to go. We'll close #304 as superseded after this lands.

  Thank you from Milla and Lu ✨

sha2fiddy:
@milla-jovovich Fixed the lint — reformatted with pinned ruff 0.4.x. The test failures are pre-existing on develop (version.py out of sync with pyproject.toml), not from this PR.

Added an upgrade note to the PR description since `repair` already rebuilds with cosine metadata.

Issue #809, opened 2026-04-13
Security Audit: 8 unreported vulnerabilities — data leakage, shell injection, file permissions


Kesshite:
## Security Audit Report

Hi Milla, Ben, and the MemPalace community!

I ran a full security audit on the v3.2.0 codebase (commit `6614b9b`, `develop` branch) before deploying MemPalace in my own workflow. I found **8 previously unreported vulnerabilities** that I'd like to flag — and I'm happy to submit PRs for all of them.

I checked existing issues and PRs before filing this. #401 (security hardening RFC), #477 (search limit), #438 (precompact session_id + regex escape), and #782 (ChromaDB telemetry) cover some related ground but the findings below are **not covered by any existing issue or PR**.

---

### 1. Wikipedia SSRF in `entity_registry.py` — violates local-first guarantee

**File:** `mempalace/entity_registry.py`, lines 176–257

`_wikipedia_lookup()` makes an outbound HTTPS GET to `https://en.wikipedia.org/api/rest_v1/page/summary/{word}` whenever `research()` is called. This is in the core package, not a benchmark or optional module.

- Any entity name extracted during mining gets sent to Wikipedia
- User's IP is disclosed to Wikipedia (and any network observer)
- Directly violates CLAUDE.md: *"Privacy by architecture — The system physically cannot send your data because it never leaves your machine"*

Additionally, if Wikipedia returns 404, the word is classified as `"person"` with 0.70 confidence (lines 246–254), which poisons the entity registry with false positives.

**Suggested fix:** Make `research()` local-only by default. Require explicit `allow_network=True` opt-in for Wikipedia lookups, and return `"unknown"` with low confidence on 404 instead of asserting person.

---

### 2. Shell injection via `eval` in `mempal_save_hook.sh` — `stop_hook_active` not sanitized

**File:** `hooks/mempal_save_hook.sh`, lines 68–80

The save hook uses `eval` to parse Python output into shell variables. The `stop_hook_active` field is **not passed through the `safe()` lambda** (unlike `session_id` and `transcript_path`):

```bash
eval $(echo "$INPUT" | python3 -c "
...
safe = lambda s: re.sub(r'[^a-zA-Z0-9_/.\-~]', '', str(s))
print(f'SESSION_ID=\"{safe(sid)}\"')
print(f'STOP_HOOK_ACTIVE=\"{sha}\"')       # ← NOT sanitized
print(f'TRANSCRIPT_PATH=\"{safe(tp)}\"')
[...]

PR #811, opened 2026-04-13
fix: make entity_registry.research() local-only by default

Kesshite:
## Summary

- `research()` previously called `_wikipedia_lookup()` unconditionally, sending entity names to `en.wikipedia.org` on every uncached lookup — violating the project's **local-first** and **privacy by architecture** principles (CLAUDE.md)
- `research()` is now **local-only by default**: returns `"unknown"` for uncached words without any network call
- New `allow_network=True` parameter required to explicitly opt in to Wikipedia lookups
- Wikipedia 404 now returns `"unknown"` (confidence 0.3) instead of asserting `"person"` (confidence 0.70), preventing entity registry poisoning with false positives

Refs: #809 (Finding 1)

## What changed

**`mempalace/entity_registry.py`**
- `research()` — added `allow_network: bool = False` parameter; uncached words return `"unknown"` immediately when `allow_network` is `False`
- `_wikipedia_lookup()` — 404 handler now returns `inferred_type: "unknown"` instead of `"person"`; added privacy warning docstring

**`tests/test_entity_registry.py`**
- `test_research_local_only_by_default` — verifies no network call happens by default
- `test_research_with_allow_network` — verifies Wikipedia is called when opted in
- `test_research_local_only_not_cached` — verifies local-only results are not persisted to wiki_cache
- `test_wikipedia_404_returns_unknown` — verifies 404 returns "unknown", not "person"
- Updated existing tests to pass `allow_network=True` where they intentionally test the Wikipedia path

## Test plan

- [x] `pytest tests/test_entity_registry.py -v` — 29/29 passed
- [x] `pytest tests/ -v --ignore=tests/benchmarks` — 689 passed, 3 failed (pre-existing: version mismatch + KG re-add bug)
- [x] `ruff check` — all checks passed
- [x] `ruff format --check` — all files formatted
- [ ] Verify no other module calls `research()` without `allow_network` (confirmed: only tests call it, and they already mock `_wikipedia_lookup`)

### DIFF
--- a/mempalace/entity_registry.py
+++ b/mempalace/entity_registry.py
@@ -523,8 +523,8 @@ class EntityRegistry:
         Caches result.  If *auto_confirm* is ``False``, marks the entry
         as unconfirmed (needs user review).
         """
-        # Already cached?
-        cache = self._data.setdefault("wiki_cache", {})
+        # Check cache (read-only — no mutation when allow_network is False)
+        cache = self._data.get("wiki_cache", {})
         if word in cache:
             return cache[word]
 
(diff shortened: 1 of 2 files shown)


bensig (approved):

PR #815, opened 2026-04-13
fix: add provenance header and speaker IDs to Slack transcript imports

Kesshite:
## Summary

Slack exports are multi-party chats where no speaker is inherently the "user" or "assistant". The parser previously assigned these roles purely by position, allowing a crafted export to place attacker text in the "user" role — making it appear as the memory owner's own words in all future retrieval (data poisoning via stored memory).

This PR adds two mitigations while preserving compatibility with the existing exchange-pair chunking pipeline:

- **Provenance header** at the top of every Slack transcript: `[source: slack-export | multi-party chat — speaker roles are positional, not verified]`
- **Speaker ID prefix** on every message: `[U1] Hello` instead of just `Hello`

Role alternation (`user`/`assistant`) is preserved so `convo_miner.py`'s `>` marker-based chunking continues to work — but now every message carries its original speaker identity and the transcript is clearly marked as multi-party with unverified roles.

Refs: #809 (Finding 6)

## What changed

**`mempalace/normalize.py`**
- `_try_slack_json()` — provenance header prepended to output; each message prefixed with `[speaker_id]`

**`tests/test_normalize.py`**
- `test_slack_json_has_provenance_header` — verifies header presence and keywords
- `test_slack_json_preserves_speaker_id` — verifies `[U1]`, `[U2]` in output
- `test_slack_json_attacker_first_message_attributed` — verifies attacker's ID is visible even when placed first

## Test plan

- [x] `pytest tests/test_normalize.py -v` — 91/91 passed
- [x] `pytest tests/ -v --ignore=tests/benchmarks` — 689 passed, 2 failed (pre-existing version mismatch)
- [x] `ruff check` — all checks passed
- [x] `ruff format --check` — all files formatted
- [x] Exchange-pair chunking preserved (user/assistant alternation unchanged)

### DIFF
--- a/mempalace/normalize.py
+++ b/mempalace/normalize.py
@@ -367,8 +367,13 @@ def _try_chatgpt_json(data) -> Optional[str]:
 def _try_slack_json(data) -> Optional[str]:
     """
     Slack channel export: [{"type": "message", "user": "...", "text": "..."}]
-    Optimized for 2-person DMs. In channels with 3+ people, alternating
-    speakers are labeled user/assistant to preserve the exchange structure.
+
+    Slack exports are multi-party chats where no speaker is inherently the
+    "user" or "assistant".  To preserve exchange-pair chunking (which relies
+    on ``>`` markers from the ``user`` role), we still alternate roles, but
+    prefix each message with the speaker ID so downstream consumers can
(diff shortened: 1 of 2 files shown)


bensig (approved):

PR #819, opened 2026-04-13
fix: use microsecond timestamp and full content hash in diary entry ID

### LINKED ISSUE #818: tool_diary_write can drop diary entries on same-second ID collision
**What happened?**

`tool_diary_write` can generate duplicate `entry_id` values for different diary entries, then call `col.add()` with that duplicate ID.

ID format today:
- timestamp only to second: `%Y%m%d_%H%M%S`
- hash only of first 50 characters: `sha256(entry[:50])[:12]`

If two entries are written in the same second and share the same first 50 characters, they get the same ID. In local repro, the second write is not stored (first entry remains), so one diary entry is effectively lost.

**What did you expect?**

Every diary write should be stored as a distinct entry. Two different entries should not collide just because they share a prefix and are written within the same second.

**How to reproduce:**

1. Run this Python snippet:

```python
import hashlib
from datetime import datetime

import chromadb

wing = "wing_claude"
entry1 = "A" * 50 + " entry one"
entry2 = "A" * 50 + " entry two"
now = datetime.now()  # same second

entry_id_1 = (
   f"diary_{wing}_{now.strftime('%Y%m%d_%H%M%S')}_"
   f"{hashlib.sha256(entry1[:50].encode()).hexdigest()[:12]}"
)
entry_id_2 = (
   f"diary_{wing}_{now.strftime('%Y%m%d_%H%M%S')}_"
   f"{hashlib.sha256(entry2[:50].encode()).hexdigest()[:12]}"
)

print("IDs identical:", entry_id_1 == entry_id_2)

client = chromadb.Client()
col = client.create_collection("diary_bug013_demo")
col.add(ids=[entry_id_1], documents=[entry1], metadatas=[{"agent": "claude"}])
col.add(ids=[entry_id_2], documents=[entry2], metadatas=[{"agent": "claude"}])

stored = col.get(ids=[entry_id_1])["documents"][0]
print("stored == entry1:", stored == entry1)
print("stored == entry2:", stored == entry2)
```

2. Observe that the two entries intentionally share the same first 50 characters and timestamp second.
3. Output shows:
   - `IDs identical: True`
   - `stored == entry1: True`
   - `stored == entry2: False`

This confirms the second entry was not stored.

**Environment:**
- OS: macOS
- Python version: 3.11.9
- MemPal version: git SHA `6614b9b` (develop)

shafdev:
fix: use microsecond timestamp and full content hash in diary entry ID

`tool_diary_write` used a second-precision timestamp and only the first 50 characters of the entry to build the ChromaDB document ID. Two entries written in the same second that share a 50-char prefix received the same ID; `col.add()` silently kept the first and dropped the second.

Fix: extend timestamp to microsecond precision (`%Y%m%d_%H%M%S%f`) and hash the full entry text so every distinct entry gets a unique ID.

Regression test: `test_diary_write_same_second_shared_prefix_no_collision`

## What does this PR do?

Fixes a silent data-loss bug in `tool_diary_write` (`mcp_server.py`). The diary entry ID was built from:

```python
# Before (buggy)
f"diary_{wing}_{now.strftime('%Y%m%d_%H%M%S')}_{hashlib.sha256(entry[:50].encode()).hexdigest()[:12]}"
```

Two different entries written in the same second that share a 50-character prefix produce the same ID. ChromaDB's `col.add()` silently ignores the duplicate — the second entry is lost with no error raised.

The fix uses microsecond precision and hashes the full entry:

```python
# After (fixed)
f"diary_{wing}_{now.strftime('%Y%m%d_%H%M%S%f')}_{hashlib.sha256(entry.encode()).hexdigest()[:12]}"
```

Closes #818
## How to test

```bash
python -m pytest tests/test_mcp_server.py::TestDiaryTools::test_diary_write_same_second_shared_prefix_no_collision -v
```

The new regression test freezes the clock to force a same-second collision, writes two entries with an identical 50-char prefix but different suffixes, then asserts both are stored.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -836,7 +836,10 @@ def tool_diary_write(agent_name: str, entry: str, topic: str = "general"):
         return _no_palace()
 
     now = datetime.now()
-    entry_id = f"diary_{wing}_{now.strftime('%Y%m%d_%H%M%S')}_{hashlib.sha256(entry[:50].encode()).hexdigest()[:12]}"
+    entry_id = (
+        f"diary_{wing}_{now.strftime('%Y%m%d_%H%M%S%f')}_"
+        f"{hashlib.sha256(entry.encode()).hexdigest()[:12]}"
+    )
 
     _wal_log(
         "diary_write",
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -6,6 +6,7 @@ dispatch layer (integration-level). Uses isolated palace + KG fixtures
 via monkeypatch to avoid touching real data.
 """
 
+from datetime import datetime
 import json
 import sys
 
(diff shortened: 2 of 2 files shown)


milla-jovovich (approved):
@shafdev thank you for catching this! And it's a beautiful fix for a common problem when people start sentences the same way during workflows which is very common. Thank you SO much for the contribution! -Milla and Lu✨

shafdev:
Thank you so much! Glad I could help, really enjoying contributing to MemPalace!

PR #410, opened 2026-04-09
fix: replace python3 -m mempalace with CLI entry point (PEP 668)

armujahid:
## Summary

- Adds `mempalace mcp-serve` CLI subcommand to replace `python3 -m mempalace.mcp_server`
- `mcp-serve` accepts `--palace` both before and after the subcommand (`mempalace --palace /path mcp-serve` and `mempalace mcp-serve --palace /path`)
- Replaces all `python3 -m mempalace` invocations with the `mempalace` CLI entry point across both `.claude-plugin/` and `.codex-plugin/` (MCP configs, hook scripts, standalone hooks)
- Updates docs (README, examples, instructions) to recommend `uv tool install` / `pipx install` as primary install methods
- Backward compat preserved — `python -m mempalace` still works via `__main__.py`

Closes #408

## Test plan

- [x] `uv run pytest tests/test_cli.py` — all 44 tests pass including new `test_main_mcp_serve_dispatches` and `test_main_mcp_serve_palace_after_subcommand`
- [x] `mempalace mcp-serve` starts the MCP server (reads JSON-RPC from stdin)
- [x] `mempalace --palace /tmp/test mcp-serve` and `mempalace mcp-serve --palace /tmp/test` both pass `--palace` through correctly
- [x] Verify no remaining `python3 -m mempalace` in plugin/hook files
- [ ] Test on a PEP 668 system with `uv tool install mempalace`

PR #833, opened 2026-04-13
fix(hooks): real python-resolution for .sh hooks, with MEMPAL_PYTHON override

igorls:
## Summary

The legacy shell hooks (`mempal_save_hook.sh`, `mempal_precompact_hook.sh`) call `python3 -m mempalace mine …` directly. On macOS GUI launches of Claude Code, `PATH` comes from launchd (`/usr/bin:/bin:/usr/sbin:/sbin`), which usually does **not** contain the Python where the user installed `mempalace`. The hook then fails silently deep in the background log.

This PR adds real PATH-robust resolution + a user-facing override, and a regression-test suite exercising the resolution priority.

Supersedes the abandoned branch `fix/hook-bugs` — see "Why the original fix wasn't enough" below.

## Changes

1. **`MEMPAL_PYTHON` override.** Users whose GUI-launch PATH lacks the right interpreter can `export MEMPAL_PYTHON=/path/to/venv/python` and hooks pick it up first. Resolution order:
   - `$MEMPAL_PYTHON` if set and executable
   - `$(command -v python3)` on PATH
   - bare `python3` as a last resort

2. **Import sanity check before auto-ingest.** Before running `-m mempalace mine`, the hook runs `-c 'import mempalace'`. On failure it logs a clear warning with the fix instructions to `~/.mempalace/hook_state/hook.log` and skips the ingest — preserving the Claude Code contract of returning 0.

3. **Applied consistently.** The JSON parse and transcript-count calls use the same resolution, so `MEMPAL_PYTHON` consistently controls the whole script.

4. **README `Known Limitations`** — documents the macOS GUI `PATH` behavior and the `MEMPAL_PYTHON` workaround.

5. **5 regression tests** in `tests/test_hooks_shell.py`:
   - `MEMPAL_PYTHON` override wins over PATH (proven via marker-emitting shim).
   - Non-executable `MEMPAL_PYTHON` falls back to PATH rather than crashing.
   - Unset `MEMPAL_PYTHON` resolves via PATH.
   - Both hooks exit 0 and log the warning when the resolved interpreter can't import mempalace.
   - Tests skip on Windows (POSIX-only bash scripts). Full suite 810/810 locally; new tests ~0.2s total.

## Why the original fix wasn't enough

The abandoned `fix/hook-bugs` branch replaced `python3` with `\$(command -v python3)` and claimed it fixed a \"nohup strips PATH\" production bug. Empirically verified: both resolve to the same binary via PATH lookup — the substitution is a no-op. And the hooks don't use nohup. The symptom the branch describes (hook silently not firing when the wrong Python is on PATH) is real, but that change didn't address it.

This PR addresses the **real** underlying failure mode — wrong Python on PATH — with an explicit user override and proactive import verification.

## Compatibility

- `hooks_cli.py` (the Python implementation invoked via `mempalace hook run …`) already uses `sys.executable` and is trivially correct. No changes there.
- Users who are already happy with bare `python3` (most Linux and CLI-launched setups) need to do nothing — the fallback chain is unchanged for them.
- Users hitting the macOS GUI issue now have a documented, testable escape hatch.

Co-Authored-By: MSL <232237854+milla-jovovich@users.noreply.github.com>

### DIFF
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -137,6 +137,19 @@ Example output:
 
 **Hooks require session restart after install.** Claude Code loads hooks from `settings.json` at session start only. If you run `mempalace init` or manually edit hook config mid-session, the hooks won't fire until you restart Claude Code. This is a Claude Code limitation.
 
+**`MEMPAL_PYTHON` override for the hook's internal Python calls.** The save hook parses its JSON input and counts transcript messages with `python3`. When the harness is launched from a GUI on macOS — `open -a`, Spotlight, the dock — its `PATH` is the minimal `/usr/bin:/bin:/usr/sbin:/sbin` inherited from `launchd`, not your shell PATH. If `python3` isn't on that PATH, those internal calls fail and the hook can't count exchanges.
+
+Point the hook at any Python 3 interpreter to fix it:
+
+```bash
+export MEMPAL_PYTHON="/usr/bin/python3"                   # system Python is fine
+export MEMPAL_PYTHON="$HOME/.venvs/mempalace/bin/python"  # or your venv
+```
+
+Resolution priority: `$MEMPAL_PYTHON` (if set and executable) → `$(command -v python3)` → bare `python3`. The interpreter only needs `json` and `sys` from the standard library — `mempalace` itself does not need to be installed in it.
+
+Note: the `mempalace mine` auto-ingest runs via the `mempalace` CLI, so that command also needs to be on the hook's `PATH`. Installing with `pipx install mempalace` or `uv tool install mempalace` puts it on a stable global location; otherwise extend the hook environment's `PATH` to include your venv's `bin/`.
+
 ## Cost
 
 **Zero extra tokens.** The hooks notify the AI that saves happened in the background — the AI doesn't need to write anything in the chat. All filing is handled automatically. Previous versions asked the AI to write diary entries and drawer content in the chat window, which cost ~$1/session in retransmitted tokens.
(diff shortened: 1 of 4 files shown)


sha2fiddy:
+1 — hit this today with `uv tool install`. Three failure modes in one session, all arguing for an explicit invocation path over auto-detection:

**1. CWD-dependent silent failure.** User has `uv tool install mempalace` plus a shell rc auto-activating a sibling project's `.venv`. The venv lacks mempalace, so `python3 -m mempalace` fails — but it *appears* to work when CWD puts the mempalace source on `sys.path` (e.g. `~/Repos/mempalace`). Same user, same install, different repo → different outcome.

**2. `No module named mempalace.__main__` variant.** A stale editable install (ours came from a Claude Code subagent's deleted worktree) leaves `.pth` resolving but `__main__.py` missing. You get the above error instead of the cleaner `ModuleNotFoundError`, which throws diagnosis off.

**3. Wrong *environment*, not just wrong python.** After fixing 1 and 2, the hook got past import and then failed at the embedding step — the venv's `onnxruntime` has a broken CoreML cache. Direct MCP calls kept working because the MCP server runs in uv-tool's isolated env. The hook needs the specific Python + site-packages that owns the vetted mempalace, not just *a* python.

**Design thoughts:**
- Invoking the `mempalace` CLI directly (what #325 does) would dodge the whole python-discovery problem — \`pipx\`/\`uv tool\` create that entry point for exactly this purpose. \`MEMPAL_PYTHON\` is still useful as a fallback for GUI launches where \`~/.local/bin\` isn't on PATH, but the binary should be primary.
- \`MEMPAL_PYTHON\` breaks the \`MEMPALACE_*\` prefix convention (cf. #711's \`MEMPALACE_HOOKS_AUTO_SAVE\`, \`MempalaceConfig\`). Rename to \`MEMPALACE_PYTHON\` pre-merge.

Related: #545, #408, #650; sibling PRs #325, #410, #670.

babamottos-bit (approved):

PR #834, opened 2026-04-13
style: ruff format 4 test files unformatted after #829 merge

mvalentsev:
@igorls -- the #829 merge landed 4 test files that don't pass `ruff format --check .`:

- `tests/test_closet_llm.py`
- `tests/test_closets.py`
- `tests/test_convo_miner.py`
- `tests/test_mcp_server.py`

This breaks lint CI on every open PR targeting develop right now (including mine and others). Pure whitespace fix, zero logic changes.

### DIFF
--- a/tests/test_closet_llm.py
+++ b/tests/test_closet_llm.py
@@ -288,9 +288,9 @@ class TestRegenerateClosets:
         survivors = closets.get(where={"source_file": source}, include=["documents", "metadatas"])
         assert survivors["ids"], "LLM closets should have been written"
         joined = "\n".join(survivors["documents"])
-        assert (
-            "STALE_REGEX_TOPIC" not in joined
-        ), "pre-existing regex closet was not purged before LLM write"
+        assert "STALE_REGEX_TOPIC" not in joined, (
+            "pre-existing regex closet was not purged before LLM write"
+        )
         assert "jwt auth" in joined
         for meta in survivors["metadatas"]:
             assert meta.get("generated_by", "").startswith("llm:")
--- a/tests/test_closets.py
+++ b/tests/test_closets.py
@@ -139,9 +139,9 @@ class TestMineLock:
         # Sort by entry time and verify the second entry is after the first exit.
         intervals.sort(key=lambda iv: iv[1])
         (_, enter_a, exit_a), (_, enter_b, exit_b) = intervals
-        assert (
-            enter_a < exit_a <= enter_b < exit_b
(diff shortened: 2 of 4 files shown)


igorls:
Thanks, let me confirm what is going on

igorls:
Thanks for looking into this, but the premise doesn't hold up — CI's lint job is actually passing on develop and on every other open PR. Quick check:

| PR | lint status |
|---|---|
| #827 | ✅ pass |
| #828 | ✅ pass |
| #833 | ✅ pass |
| **#834** | ❌ fail |

#834 is the only PR where lint fails — and its own CI log (job [71175428208](https://github.com/MemPalace/mempalace/actions/runs/24371332083/job/71175428208)) explains why:

```
Successfully installed ruff-0.4.10
...
Would reformat: tests/test_closet_llm.py
Would reformat: tests/test_closets.py
Would reformat: tests/test_convo_miner.py
Would reformat: tests/test_mcp_server.py
```

## Root cause

CI pins ruff to `>=0.4.0,<0.5` in [.github/workflows/ci.yml](https://github.com/MemPalace/mempalace/blob/develop/.github/workflows/ci.yml):

```yaml
- run: pip install "ruff>=0.4.0,<0.5"
- run: ruff format --check .
```

On current develop:

```
$ uvx --from 'ruff>=0.4.0,<0.5' ruff format --check .
75 files already formatted
```

On your branch the same pinned ruff wants to revert your changes — because ruff's preferred style for `assert cond, (msg)` vs `assert (cond), msg` changed between 0.4.x and later versions. Your local ruff is newer (probably 0.11+), so it sees the current code as needing reformatting; the CI pin sees your reformat as needing reversal. Applying your diff *creates* the lint failure you're trying to fix.

## What would actually work

If we want to move to newer ruff's style, it needs to be a coordinated change: bump the pin in `ci.yml` and reformat the entire tree in the same commit, so CI and the repo agree on one formatter version. Piecemeal reformatting against the current pin will keep oscillating.

To validate any ruff work locally against what CI actually runs:

```bash
uvx --from 'ruff>=0.4.0,<0.5' ruff format --check .
uvx --from 'ruff>=0.4.0,<0.5' ruff check .
```

Happy to review a follow-up that bumps the pin + reformats the whole tree if the team wants that.

mvalentsev:
You are absolutely right (f..ng claude), my bad. I was running a newer ruff locally (0.11+) and the style diff between 0.4.x and 0.11.x is what caused this. Should have pinned to the CI version before touching anything. Closing, sorry for the noise.

PR #782, opened 2026-04-13
fix: disable ChromaDB telemetry to suppress posthog spam

OmkarKirpan:
## Summary

Closes #458

Every ChromaDB operation in mempalace 3.1.0+ floods stderr with posthog telemetry errors:

```
Failed to send telemetry event ClientStartEvent: capture() takes 1 positional argument but 3 were given
```

This PR:
- Adds a shared `CHROMA_SETTINGS = Settings(anonymized_telemetry=False)` constant in `backends/chroma.py`
- Updates all 11 `PersistentClient` construction sites across 6 files to use it
- Updates test helpers to use the same settings (avoids ChromaDB's `SharedSystemClient` singleton conflict)

### Files changed
- `mempalace/backends/chroma.py` — defines the shared constant, uses it
- `mempalace/mcp_server.py` — imports and uses `CHROMA_SETTINGS`
- `mempalace/cli.py` — uses `CHROMA_SETTINGS` in `cmd_repair` and `cmd_compress`
- `mempalace/dedup.py` — uses `CHROMA_SETTINGS` in both client creation sites
- `mempalace/repair.py` — uses `CHROMA_SETTINGS` in all 3 client creation sites
- `mempalace/migrate.py` — uses `CHROMA_SETTINGS` in both client creation sites
- `tests/test_backends.py` — verification client uses `CHROMA_SETTINGS`
- `tests/test_mcp_server.py` — test helper uses `CHROMA_SETTINGS`

## Test plan
- [x] All existing tests pass (`pytest tests/test_backends.py tests/test_mcp_server.py::TestCacheInvalidation tests/test_cli.py` — 53 passed)
- [x] `ruff check` and `ruff format` clean
- [x] Manual: `mempalace mine` / `mempalace status` no longer prints telemetry errors to stderr

PR #841, opened 2026-04-14
Four security issues found and fixed to make MemPalace truly local, telemetry-free, and safe.

groxaxo:
## Summary

Four security issues found and fixed to make MemPalace truly local, telemetry-free, and safe.

---

### 1. ChromaDB telemetry — actually disabled (not just silenced)

**Before:** The code only silenced the PostHog log warning. ChromaDB's underlying telemetry client was still active and attempted to send usage data to PostHog servers on every operation.

**After:**
- `mempalace/__init__.py` — sets `ANONYMIZED_TELEMETRY=False` environment variable (the kill-switch ChromaDB reads)
- `mempalace/backends/chroma.py` — passes `Settings(anonymized_telemetry=False)` directly to `PersistentClient` (belt-and-suspenders)

---

### 2. Wikipedia entity lookups — opt-in only (was always-on)

**Before:** `EntityRegistry.research()` automatically sent entity names (people, projects) to the Wikipedia REST API. Any entity lookup silently made an external HTTP request revealing names from private conversations.

**After:** External lookups are disabled by default. Enable explicitly with:
```python
registry.research("name", wiki_lookup=True)
# or: MEMPALACE_WIKI_LOOKUP=1
```
Default behaviour is purely local.

---

### 3. Shell injection via `eval` — replaced with safe temp-file sourcing

**Before:** `hooks/mempal_save_hook.sh` used `eval $(... python3 ...)` to parse JSON and assign shell variables. Despite sanitisation, `eval` is fundamentally dangerous.

**After:** Python writes sanitised values to a `mktemp`-created temp file (path passed as a positional argument, never via env var), which the shell then sources. No `eval` anywhere in the parse path.

---

### 4. LLM endpoint URL validation — rejects SSRF vectors

**Before:** Any string (including `file:///etc/passwd`, `ftp://`, `http://user@evil.com`) was accepted as an LLM endpoint.

**After:** `LLMConfig._validate_endpoint()` enforces:
- Scheme must be `http://` or `https://`
- URL must have a non-empty host
- URL must not contain userinfo (`@`) to prevent credential-hiding attacks

---

### Installation

The package is already installable via terminal:
```bash
pip install mempalace          # from PyPI
pip install -e .               # from source (dev mode)
```

---

### Tests

All new behaviour is covered by tests. No regressions introduced — all pre-existing failures are unrelated (ChromaDB network calls blocked in the CI sandbox).

### DIFF
--- a/hooks/mempal_save_hook.sh
+++ b/hooks/mempal_save_hook.sh
@@ -67,21 +67,20 @@ INPUT=$(cat)
 # Parse all fields safely using Python — write to a temp file instead of eval
 # to avoid any risk of shell injection via crafted JSON values.
 _HOOK_TMP=$(mktemp)
-echo "$INPUT" | python3 -c "
-import sys, json, re, os
+echo "$INPUT" | python3 - "$_HOOK_TMP" <<'PYEOF'
+import sys, json, re
 data = json.load(sys.stdin)
 sid = data.get('session_id', 'unknown')
 sha = data.get('stop_hook_active', False)
 tp  = data.get('transcript_path', '')
 # Restrict each value to safe characters before writing
 safe = lambda s: re.sub(r'[^a-zA-Z0-9_/.\-~]', '', str(s))
-out = os.environ.get('_HOOK_TMP', '')
-if out:
-    with open(out, 'w') as f:
-        f.write(f'SESSION_ID={safe(sid)}\n')
-        f.write(f'STOP_HOOK_ACTIVE={sha}\n')
-        f.write(f'TRANSCRIPT_PATH={safe(tp)}\n')
-" _HOOK_TMP="$_HOOK_TMP" 2>/dev/null
+out = sys.argv[1]
+with open(out, 'w') as f:
+    f.write(f'SESSION_ID={safe(sid)}\n')
+    f.write(f'STOP_HOOK_ACTIVE={sha}\n')
+    f.write(f'TRANSCRIPT_PATH={safe(tp)}\n')
+PYEOF
 
 # Source the temp file (values are already sanitised and contain no special chars)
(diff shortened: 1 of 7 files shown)


arnoldwender:
Hey @groxaxo — really nice catches here, especially the `eval` one in the hook script. Thanks for taking the time to write this up properly.

One small friendly note for next time: issues 3 (`eval` shell injection) and 4 (SSRF in LLM endpoint) are the kind of findings that ideally go through **coordinated disclosure** first, so users have time to update before the exploit details are public. The normal flow is:

1. GitHub Security Advisory (private channel): repo → Security tab → *Report a vulnerability*
2. Or a quiet email to a maintainer
3. Agree on a timeline, maintainer ships the patch, then the advisory goes public with credit to you (and often a CVE)

Totally not a dig at you though — I checked and **this repo doesn't have Private Vulnerability Reporting enabled yet and there's no `SECURITY.md`**, so there's no obvious private channel documented. That's genuinely hard to work around, and a public PR is a reasonable fallback when there's nowhere else to go.

**Friendly suggestion for maintainers @bensig @igorls @jphein**: could you enable [Private Vulnerability Reporting](https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository) and add a short `SECURITY.md`? Two-minute config change that gives future reporters a clean path and protects users in the window between report and patch.

Either way — thanks for the solid work @groxaxo, and thanks maintainers for being responsive to contributions in general 🙌

bensig:
closing — the telemetry fix overlaps with #782 and the other security fixes were addressed in #647/#812. the bundle also conflicts with develop. pls submit any remaining fixes as individual PRs. thanks @groxaxo!

PR #846, opened 2026-04-14
feat: include created_at timestamp in search results

### LINKED ISSUE #465: Include created_at timestamp in search results
## Problem

Search results return `wing`, `room`, `source_file`, and `similarity` but no timestamp. You can't tell when a drawer was stored without reading it directly. This makes temporal reasoning impossible from search results alone.

## Proposal

Add `created_at` to the search result objects:

```json
{
  "text": "Team decided to migrate auth to Clerk...",
  "wing": "driftwood",
  "room": "decisions",
  "source_file": "conversation-2026-04-09",
  "similarity": 0.89,
  "created_at": "2026-04-09T20:25:23Z"
}
```

If ChromaDB metadata already stores a timestamp, this is just passing it through. If not, it should be added at ingest time.

## Why this matters

When the AI retrieves memories, it needs to know if a result is from yesterday or six months ago. A decision from January may have been superseded. Without timestamps in results, every memory looks equally current.

sha2fiddy:
## Summary
- Surfaces the existing `filed_at` metadata as `created_at` in search result objects returned by `search_memories()`
- Enables temporal reasoning over search hits without additional queries (e.g., "what did I say last week?")
- Zero breaking changes — additive field only

Closes #465

## Changes
- `mempalace/searcher.py`: Added `created_at` field to search result entry dict, sourced from `meta.get("filed_at")`
- `tests/test_searcher.py`: Added `created_at` presence check to `test_result_fields`, plus new `test_created_at_contains_filed_at` verifying the value matches the stored `filed_at` metadata

## Test plan
- [x] `test_result_fields` — verifies `created_at` key exists in search results
- [x] `test_created_at_contains_filed_at` — verifies value matches the seeded `filed_at` metadata (`2026-01-01T00:00:00`)
- [x] Full test suite: 865 passed, 84% coverage
- [x] Lint clean (`ruff check .`, `ruff format --check .` with CI-pinned ruff >=0.4.0,<0.5)

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -397,7 +397,7 @@ def search_memories(
             "wing": meta.get("wing", "unknown"),
             "room": meta.get("room", "unknown"),
             "source_file": Path(source).name if source else "?",
-            "created_at": meta.get("filed_at"),
+            "created_at": meta.get("filed_at", "unknown"),
             "similarity": round(max(0.0, 1 - effective_dist), 3),
             "distance": round(dist, 4),
             "effective_distance": round(effective_dist, 4),
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -397,6 +397,7 @@ def search_memories(
             "wing": meta.get("wing", "unknown"),
             "room": meta.get("room", "unknown"),
             "source_file": Path(source).name if source else "?",
+            "created_at": meta.get("filed_at"),
             "similarity": round(max(0.0, 1 - effective_dist), 3),
             "distance": round(dist, 4),
             "effective_distance": round(effective_dist, 4),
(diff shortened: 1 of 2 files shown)


mempalace/searcher.py
```diff
@@ -397,6 +397,7 @@ def search_memories(
             "wing": meta.get("wing", "unknown"),
             "room": meta.get("room", "unknown"),
             "source_file": Path(source).name if source else "?",
+            "created_at": meta.get("filed_at"),
```
kbtale: Wouldn't an explicit fallback value be useful here? all the other fields that pull from external data seem to have a fallback.

kbtale (changes requested):

bensig (approved):

PR #857, opened 2026-04-14
Improve init command: create directory if missing and normalize path

iron-prog:
## What does this PR do?
This PR improves the mempalace init <dir> command by making it more user-friendly and robust.

## How to test

•	Automatically creates the target directory if it does not exist
•	Normalizes the provided path using Path.resolve() to ensure consistency across the application

Previously, the command assumed the directory already existed, which could lead to errors or confusion during initial setup.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`)
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`)

This issue was encountered during local setup, where init failed if the directory didn’t exist. This change aligns the behavior with common CLI tools and improves the developer experience.

PR #862, opened 2026-04-14
docs(cli): clarify that 'mempalace init' requires <dir> (#210)

### LINKED ISSUE #210: Init in documentation has no dir, but is required
mempalace init in documentation says it will create ~/.mempalace

instead it fails with "mempalace init: error: the following arguments are required: dir"

need to fix init or docs.

arnoldwender:
## What and Why

Closes #210. The reporter ran `mempalace init` with no argument and hit
`error: the following arguments are required: dir`, because the docs led them
to believe `init` just sets up `~/.mempalace/`.

## Root Cause

Two doc locations de-emphasized the required positional `<dir>`:

- `website/reference/cli.md` — the "What it does" list made "Creates `~/.mempalace/` config directory" sound like the primary effect of `init`, without reinforcing that `<dir>` is required.
- `website/guide/getting-started.md` — the Quick Start walked straight into `mempalace init ~/projects/myapp` without noting that a directory argument is mandatory.

## Change Summary

- `website/reference/cli.md` — mark `<dir>` as **Required**, add `mempalace init .` as the current-directory usage, re-order the "What it does" list so scanning `<dir>` is listed first, and add a short note about the argparse error users see when they forget the argument.
- `website/guide/getting-started.md` — add a one-line note that `<dir>` is required, show `mempalace init .` as an alternative, and soften "Sets up" to "Ensures … exists" to avoid implying init's only job is creating the global dir.

No CLI/code changes — this is a documentation-only PR and does not conflict with #857 (init directory creation/normalization).

## Test Plan

- [x] Manually re-read both docs to confirm `<dir>` is now unambiguously required
- [x] Verified the CLI still rejects `mempalace init` with no arg (expected behavior kept; docs now match)

Closes #210

### DIFF
--- a/website/guide/getting-started.md
+++ b/website/guide/getting-started.md
@@ -34,14 +34,20 @@ Three steps: **init**, **mine**, **search**.
 
 ### 1. Initialize Your Palace
 
+`mempalace init` requires a project directory to scan. Pass a path,
+or `.` to use the current directory.
+
 ```bash
 mempalace init ~/projects/myapp
+# or, from inside the project:
+mempalace init .
 ```
 
 This scans your project directory and:
+
 - Detects people and projects from file content
 - Creates rooms from your folder structure
-- Sets up `~/.mempalace/` config directory
+- Ensures the `~/.mempalace/` config directory exists
 
 ### 2. Mine Your Data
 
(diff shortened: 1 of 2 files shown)


bensig (approved):

PR #863, opened 2026-04-14
fix(hooks): stop precompact hook from blocking compaction (#856, #858)

### LINKED ISSUE #856: preCompact hook cancels compaction instead of deferring — `/compact` silently returns without compacting
## Problem

When the `preCompact` hook is installed, running `/compact` in Claude Code returns immediately without compacting. The hook fires, blocks with a save instruction, but Claude Code treats the block as a **cancellation** rather than a pause-then-proceed. Compaction never resumes.

The result is the opposite of the hook's intent: instead of preventing data loss before compaction, it makes compaction impossible — guaranteeing eventual data loss as context fills up.

## Root cause

Claude Code's PreCompact hook type has no retry/resume mechanism. A `block` decision means **cancel**, not **defer**. The current hook design assumes Claude Code will retry compaction after Claude acts on the save instruction — but it doesn't.

## Why a shell-side fix is non-trivial

The obvious workaround — have the hook save to MemPalace directly and `exit 0` — is blocked by a transport limitation: the MCP server uses **stdio transport** exclusively (no REST/HTTP API). Shell scripts cannot call MCP tools; those are LLM-only. The hook can only use the `mempalace` CLI.

## Proposed fix

Rewrite the `preCompact` hook to:

1. Call `mempalace mine <transcript_path> --mode convos --since <last_save_offset>` using the transcript path from the hook's stdin JSON
2. Exit `0` — compaction proceeds normally, no Claude involvement required

This requires a new `--since` flag on `mempalace mine --mode convos` to support incremental ingestion (avoiding re-processing the entire transcript on every compaction). The Stop hook already tracks save state per session in `~/.mempalace/hook_state/` — the same mechanism could be used here.

## Alternatives considered

- **Remove `preCompact` hook entirely** — the Stop hook fires every 15 messages and already provides protection against context loss. This is the minimal viable fix if `--since` is too much scope.
- **Add HTTP/REST transport to MCP server** — would allow hooks to call MCP tools directly; likely out of scope for this issue.

## Related

- #622 — Stop hook saves to Claude Code's native auto-memory instead of MemPalace MCP tools

## Environment

- mempalace 3.2.0 (installed via uv)
- macOS

### LINKED ISSUE #858: PreCompact hook always blocks /compact — no way to acknowledge/bypass
## Problem

`hooks/mempal-precompact-hook.sh` (via `mempalace hook run --hook precompact`) **unconditionally** returns:

```json
{"decision": "block", "reason": "COMPACTION IMMINENT. Save ALL topics..."}
```

This means `/compact` in Claude Code is permanently blocked as long as the plugin is installed. There is no mechanism for the model (or user) to signal "I saved what I needed, please proceed."

Observed in mempalace 3.0.14, Claude Code on macOS.

## Reproduction

1. Install mempalace plugin.
2. Run `/compact` in any Claude Code session.
3. Claude Code reports:

   > Error: Compaction blocked by PreCompact hook: [bash \${CLAUDE_PLUGIN_ROOT}/hooks/mempal-precompact-hook.sh]: COMPACTION IMMINENT. Save ALL topics, decisions, quotes, code, and important context from this session to your memory system...

4. The model saves drawers via `mempalace_add_drawer`.
5. `/compact` is retried — still blocked. The hook never inspects whether any save actually happened.

## Impact

- Users cannot compact sessions without manually patching / removing the hook.
- The "save before compact" nudge is still useful, but a hard block that can never be cleared is a footgun — many users will end up disabling the whole plugin.

## Suggested fix

One of:

1. **Stateful check** — track timestamp of last `mempalace_add_drawer` (or any write) and approve if a save happened since the last `/compact` / session-start.
2. **Two-phase hook** — first call returns `block` with instructions; subsequent call within N seconds returns `approve` (gives the model one turn to save, then lets compaction through).
3. **Warn-only mode** — return `approve` but include the reason text so the model still sees the nudge; document an opt-in `MEMPALACE_BLOCK_COMPACT=1` env var for users who want a hard block.

Option 1 matches user expectations best.

## Workaround (for now)

Replace `hooks/mempal-precompact-hook.sh` with:

```bash
#!/bin/bash
cat >/dev/null
echo '{"decision":"approve"}'
exit 0
```

…or delete the `PreCompact` block from `hooks/hooks.json`. Both get wiped on plugin update.

mvalentsev:
## Problem

The precompact hook in `hooks_cli.py` unconditionally returns `{"decision": "block"}`. Claude Code treats `block` on a PreCompact hook as a cancellation with no retry -- so `/compact` is permanently broken for anyone using the plugin.

The standalone bash hooks in `hooks/` were updated separately (#885), but `hooks_cli.py` (called by the plugin wrapper) still returned `block`.

Closes #856, closes #858.

## Changes

- `hook_precompact()` now mines the transcript directory synchronously (so data lands before compaction proceeds), then returns `{}` (empty JSON, not blocking compaction)
- Extracted `_get_mine_dir(transcript_path)` helper that resolves the mine directory from `MEMPAL_DIR` env var or the transcript file location
- Added `_mine_sync()` for synchronous mining with a 60-second timeout
- Extended `_maybe_auto_ingest()` to also accept `transcript_path` (falls back to `MEMPAL_DIR`)
- Stop hook behavior is intentionally unchanged -- #673 handles the full silent save path for that

## Test plan

- [x] `pytest tests/test_hooks_cli.py -v` -- 46 passed
- [x] `pytest tests/ --ignore=tests/benchmarks` -- 869 passed (1 pre-existing flaky unrelated)
- [x] Real hook invocation: `echo '{"session_id":"x"}' | python -m mempalace hook run --hook precompact --harness claude-code` returns `{}`
- [x] Stop hook still returns `block` at interval (unchanged)

## Scope note

This PR intentionally only fixes the precompact hook. The stop hook has several open PRs (#673, #711, #804) working on the broader silent save / configurable interval redesign. Kept this focused on the critical bug -- compaction should not be permanently broken.

---

p.s. Sorry for the PR volume from my side lately -- trying to keep each one small and focused rather than bundling unrelated fixes.

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -36,6 +36,11 @@ PRECOMPACT_BLOCK_REASON = (
     "Save everything to MemPalace, then allow compaction to proceed."
 )
 
+PRECOMPACT_ALLOW_REASON = (
+    "MemPalace pre-compaction save. Your conversation has been saved "
+    "in the background. Compaction can proceed safely."
+)
+
 
 def _sanitize_session_id(session_id: str) -> str:
     """Only allow alnum, dash, underscore to prevent path traversal."""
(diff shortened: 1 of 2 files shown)


bensig (approved):

PR #864, opened 2026-04-14
fix(mcp): redirect stdout to stderr during import to protect JSON-RPC channel (#225)

### LINKED ISSUE #225: [bug] mempalace mcp writes startup text to stdout instead of stderr, breaking Claude Desktop JSON parsing

arnoldwender:
## What and Why

Closes #225. On Windows, `mempalace mcp` writes startup banners and warnings to **stdout**, which corrupts the MCP JSON-RPC stream. Claude Desktop logs:

```
MCP mempalace: Unexpected token '*', "**********"... is not valid JSON
```

## Root Cause

`mempalace/mcp_server.py` imports `chromadb` (which loads `onnxruntime` and `posthog`). Those dependencies print banners and warnings directly to stdout — sometimes at the C level — before the MCP protocol loop starts. Stdout is the JSON-RPC transport, so any non-JSON byte breaks the parser. Existing `logging.basicConfig(stream=sys.stderr)` only handles Python `logging`, not `print()` and not C-level writes.

## Change Summary

- `mempalace/mcp_server.py:23-43` — at module load, before any heavy import, save the real stdout and redirect both `sys.stdout` (Python level) and fd 1 → fd 2 (`os.dup2(2, 1)`, OS level). The fd-level redirect catches C-level prints from onnxruntime/posthog. Wrapped in `try/except (OSError, AttributeError)` for embedded-interpreter and unusual test environments.
- `mempalace/mcp_server.py:1666-1678` — new `_restore_stdout()` helper restores the saved fd to fd 1 and re-points `sys.stdout` at the real stream. Idempotent.
- `mempalace/mcp_server.py:1681` — `main()` calls `_restore_stdout()` immediately, so JSON-RPC responses written via `sys.stdout.write(...)` reach Claude Desktop unchanged.
- `tests/test_mcp_stdio_protection.py` — three regression tests verifying redirect at import, restore via helper, and a clean-exit subprocess invocation that asserts stdout is empty.

## Test Plan

- [x] `pytest tests/test_mcp_stdio_protection.py -v` — all 3 pass on this branch
- [x] Same tests fail on `upstream/develop` baseline (verified via `git stash` swap), proving they actually catch the regression
- [x] `ruff check` and `ruff format --check` clean
- [x] Imports are intentionally below the redirect block; each carries `# noqa: E402` with a comment explaining why

## Notes for Reviewer

- The fd-level redirect is the canonical pattern recommended for stdio-based MCP servers (also used by the official MCP Python SDK examples). Pure `sys.stdout = sys.stderr` would not catch onnxruntime's C-level prints, which are the loudest source of corruption on Windows.
- No behavior change for existing JSON-RPC responses — they continue to go to the real stdout via `sys.stdout.write(...)` in `main()`.

Closes #225

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -20,22 +20,47 @@ Tools (maintenance):
   mempalace_reconnect       — force cache invalidation and reconnect after external writes
 """
 
-import argparse
 import os
 import sys
-import json
-import logging
-import hashlib
-import time
-from datetime import datetime
-from pathlib import Path
 
-from .config import MempalaceConfig, sanitize_kg_value, sanitize_name, sanitize_content
-from .version import __version__
-from .backends.chroma import ChromaBackend, ChromaCollection
-from .query_sanitizer import sanitize_query
-from .searcher import search_memories
-from .palace_graph import (
+# --- MCP stdio protection (issue #225) -----------------------------------
+# The MCP protocol multiplexes JSON-RPC over stdio: stdout MUST carry only
+# valid JSON-RPC messages, stderr is for human-readable logs. Some
+# transitive dependencies (chromadb → onnxruntime, posthog telemetry) print
+# banners and error messages directly to stdout — sometimes at C level —
+# which breaks Claude Desktop's JSON parser. Redirect stdout → stderr at
+# both the Python and file-descriptor level before heavy imports, then
+# restore the real stdout in main() before entering the protocol loop.
+_REAL_STDOUT = sys.stdout
+_REAL_STDOUT_FD = None
+try:
+    _REAL_STDOUT_FD = os.dup(1)
+    os.dup2(2, 1)
+except (OSError, AttributeError):
+    # Environments without fd-level stdio (embedded interpreters, some test
+    # harnesses). The Python-level redirect below still applies.
+    pass
+sys.stdout = sys.stderr
+
+import argparse  # noqa: E402  (deferred until after stdio protection above)
+import json  # noqa: E402
+import logging  # noqa: E402
+import hashlib  # noqa: E402
+import time  # noqa: E402
+from datetime import datetime  # noqa: E402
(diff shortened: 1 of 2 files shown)


arnoldwender:
Rebased on latest develop to resolve a small conflict in `mcp_server.py` introduced by 79c9c0e (the `sanitize_kg_value` import addition). No logical changes — kept the stdio protection block intact and merged the new import into the same `from .config import ...` group. Tests and lint still green.

bensig (approved):

Issue #185, opened 2026-04-08
mempalace init writes entities.json and mempalace.yaml to project directory instead of ~/.mempalace/


fubak:
## Issue

`mempalace init <dir>` writes two config files into the target project directory:

- `<dir>/entities.json`
- `<dir>/mempalace.yaml`

These land alongside the user's project files rather than in `~/.mempalace/` where `config.json` lives. This creates clutter, risks accidental commits, and splits mempalace config across two locations with no clear rationale.

## Steps to Reproduce

```bash
mempalace init /home/fubak/projects/myapp
ls /home/fubak/projects/myapp/
# → entities.json  mempalace.yaml  (plus all existing project files)
ls ~/.mempalace/
# → config.json  palace/
```

## Expected

All mempalace config files should live in `~/.mempalace/`, or at minimum be namespaced by wing to avoid collisions across projects:

```
~/.mempalace/
  config.json
  palace/
  wings/
    myapp/
      entities.json
      mempalace.yaml
```

## Impact

- Files in project root risk being committed to git (no default `.gitignore` entry added)
- Running `mempalace init` on multiple projects scatters config files with identical names across the filesystem
- Users initializing on a root directory (e.g. `/home/fubak`) get config files in their home dir root

## Environment

- mempalace 3.0.0

ttquoccuong:
i already have PR for this.
look at https://github.com/MemPalace/mempalace/pull/933

mindfullsilence:
FWIW, not a fan of this side-effect as a solution. I'd rather a software not modify my project files before I've told it to. Pretty sure this should just be a section added to the docs instructing how/what to add files to your global gitignore as is the case with pretty much any other software. 

Perhaps submitting a new entry to GitHub's own templates as well for completeness:
https://github.com/github/gitignore/tree/main/Global

PR #866, opened 2026-04-14
fix(init): auto-add per-project files to .gitignore in git repos (#185)

arnoldwender:
## What and Why

Partially addresses #185.

`mempalace init <dir>` writes `mempalace.yaml` and `entities.json` into the project root. When `<dir>` is a git repository, those files have **no default protection** against being staged into git — the loudest concern in the original report:

> Files in project root risk being committed to git (no default `.gitignore` entry added)

## Scope and Tradeoffs

The full proposal in #185 is to relocate the files to `~/.mempalace/wings/<wing>/`. That is a behavioral change touching the miner, the config loader, every example in the docs, and existing user setups, and it warrants its own design discussion (likely with #857 / PR #857 in mind). I deliberately scoped this PR down to the **gitignore safeguard** so the immediate risk is removed without breaking any existing flow. Happy to follow up with the relocation if the maintainers want to take that route.

## Change Summary

- `mempalace/cli.py` — new `_ensure_mempalace_files_gitignored(project_dir)` helper. Runs as the last step of `cmd_init()`. Conservative: only runs when `<dir>/.git` exists (no-op otherwise), skips entries already present, preserves existing `.gitignore` content, and handles files without trailing newlines. Adds a clearly-marked block referencing #185 so future readers know why the entries are there.
- `tests/test_init_gitignore_protection.py` — 5 cases: no git repo (no-op), fresh `.gitignore` creation, partial append (only missing entries), idempotency (no duplicates), trailing-newline edge case.

## Test Plan

- [x] `pytest tests/test_init_gitignore_protection.py -v` — all 5 pass
- [x] `ruff check` and `ruff format --check` clean
- [x] No changes to existing init flow when run outside a git repo

Refs #185 (this PR addresses the gitignore concern; the file-relocation question remains open)

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -36,6 +36,37 @@ from pathlib import Path
 from .config import MempalaceConfig
 
 
+_MEMPALACE_PROJECT_FILES = ("mempalace.yaml", "entities.json")
+
+
+def _ensure_mempalace_files_gitignored(project_dir) -> bool:
+    """If project_dir is a git repo, ensure MemPalace's per-project files
+    are listed in .gitignore so they don't get committed by accident.
+
+    Returns True if .gitignore was updated, False otherwise. Issue #185:
+    `mempalace init` writes mempalace.yaml + entities.json into the
+    project root, where they previously had no protection against being
+    staged into git.
+    """
+    from pathlib import Path
+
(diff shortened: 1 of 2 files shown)


bensig (approved):

PR #851, opened 2026-04-14
fix(status): paginate metadata fetch to support large palaces

vnguyen-lexipol:
## Summary

- Paginates `col.get()` in `miner.py:status()` using 5k batches with `offset` instead of a single `col.get(limit=total)` call
- Uses `col.count()` for the header total instead of `len(metas)`
- Aggregates wing/room counts incrementally per batch (no full metadata list in memory)

## Problem

On palaces with >10k drawers:
- **v3.0.x** (`limit=10000`): silently displays "10000 drawers" regardless of actual count (#850)
- **develop** (`limit=total`): crashes with SQLite "too many SQL variables" (#802)

## Fix

```python
total = col.count()
batch_size = 5000
offset = 0
while offset < total:
    r = col.get(limit=batch_size, offset=offset, include=["metadatas"])
    batch = r["metadatas"]
    if not batch:
        break
    for m in batch:
        wing_rooms[m.get("wing", "?")][m.get("room", "?")] += 1
    offset += len(batch)
```

## Testing

Tested on a 122,686-drawer palace — correctly displays full count and complete wing/room breakdown with no SQLite errors.

Fixes #850
Related: #802, #723

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #892, opened 2026-04-14
Fix stale status display by paginating drawer metadata

ldtalexthip:
## Summary
- `mempalace status` hard-caps `col.get(limit=10000, ...)` which makes the header literally display `MemPalace Status — 10000 drawers` once a palace exceeds 10k drawers.
- Room totals freeze at whatever fell inside the first 10k page, so long-lived palaces silently drift out of sync with Chroma.
- This patch paginates in 5k chunks up to `col.count()` so the reported totals match what's actually stored.

Before:
```
MemPalace Status — 10000 drawers
  WING: migration_ops
    ROOM: technical   5401 drawers
    ROOM: general        7 drawers  (actual: 19,475)
```

After:
```
MemPalace Status — 45743 drawers
  WING: migration_ops
    ROOM: general    19475 drawers
    ROOM: technical   5401 drawers
    ...
```

## Test plan
- [x] Run `mempalace status` against a palace with >10k drawers — totals now match `col.count()`
- [x] Run against an empty palace — no divide-by-zero / empty-list issues
- [ ] Run against a palace at exactly 10k drawers (boundary) — should loop once with `offset=0`, then exit

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -653,9 +653,15 @@ def status(palace_path: str):
         print("  Run: mempalace init <dir> then mempalace mine <dir>")
         return
 
-    # Count by wing and room
-    r = col.get(limit=10000, include=["metadatas"])
-    metas = r["metadatas"]
+    # Count by wing and room — paginate so we don't cap at 10k.
+    total = col.count()
+    metas: list = []
+    offset = 0
+    page = 5000
+    while offset < total:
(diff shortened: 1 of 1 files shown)


bensig:
closing in favor of #851 which covers the same status pagination fix and is mergeable. thanks @ldtalexthip!

PR #907, opened 2026-04-15
feat: add Italian language support

Archetipo95:
## Summary
- adds `it.json` to the i18n dictionaries
- covers Italian in dialect/i18n smoke tests
- leaves runtime language selection and spellcheck behavior unchanged


igorls:
Thanks for this, @Archetipo95! The `it.json` content itself is clean — schema matches `en.json` exactly, all interpolation variables (`{path}`, `{count}`, `{closets}`, `{drawers}`, `{fixed}`, `{query}`) are correct, `topic_pattern` and `action_pattern` properly localized with `À-ÿ` for accents.

One small ask before merge: please drop the edit to `mempalace/i18n/test_i18n.py`. That file moves to `tests/test_i18n.py` via #758 (which will merge ahead of this one), and once it lands your addition there will either conflict or be stranded. The test suite in the new location auto-discovers every `*.json` in `mempalace/i18n/`, so the `"it"` sample is unnecessary — `it.json` will be exercised automatically by `test_all_languages_load` and `test_interpolation`.

Also, would you mind triggering CI so we can confirm green? (No checks have run yet on this branch.)

Optionally — not blocking merge — if you want Italian to also power entity detection (name extraction from `.md`/`.txt` prose with accents like `Gianluca`, `Niccolò`, `Università`), #911 just landed the infra for that: you can add an `entity` section to `it.json` with the pt-br-style patterns. Totally optional; the CLI/AAAK translation is a standalone win.

Archetipo95:
@igorls all done!

I noticed CI didn’t trigger on this fork PR. From the GitHub side it looks like no check suite was created at all, so this may be waiting on maintainer approval for forked workflows? Or how should I trigger it?

I re-ran the relevant checks locally and they all passed:
- `uv run python -m pytest tests/test_i18n.py -v`
- `uv run python -m pytest tests/test_entity_detector.py -v`
- `uv run python -m pytest tests/test_readme_claims.py -v`

Issue #756, opened 2026-04-13
Add MemPalace OpenAI 3072-dim embedding support or something similar opensource.


marcuss-grindr:
**What problem does this solve?**
I've been working in something similar from a few months now, and have a RAG with lance DB, and my breakthrough recently has been that the OpenAi  3072-dim embeddings makes everything work a 1000 times better.

In fact I'm open this issue because I expected MemPalace to beat one of my plugin skill for code review by a lot, but it did, it failed and the reason for it at least as identified by my agent is :

  

> Mempalace missed it because ChromaDB's local embeddings (all-MiniLM-L6-v2, 384-dim) didn't score the query "TEXT primary key natural key surrogate integer" close enough to that entry. The smaller embedding model has less semantic resolution  for database schema terminology.

>   Mempalace just does raw cosine similarity with no reranking — so lower-quality embeddings + no metadata boosting = missed finding.

**What's the proposed solution?**
Also anecdotic-ally I stopped using chroma I now use lanceDB and get better results.

**Alternatives considered:**

codysnider:
Try https://github.com/codysnider/tagmem

It performs better (benchmark and inference timing) than mempalace and supports any OpenAI-compatible embedding backend.

NickShtefan:
Quick clarification on what's possible today vs what's blocked on a different change: PR #442 lets you swap the local embedding model to anything sentence-transformers can load — `intfloat/multilingual-e5-large-instruct` (1024-dim), `BAAI/bge-large-en-v1.5` (1024-dim), or `Qwen/Qwen3-Embedding-0.6B` (1024-dim) all work, and they push semantic resolution well above MiniLM-L6 without the OpenAI dependency.

3072-dim OpenAI specifically would need an OpenAI provider in the embedding-function chain (#442 plumbs the provider, but only sentence-transformers + ChromaDB-default are wired). The LanceDB switch you mentioned is independent — that's #1196's territory (SqliteVec backend in flight). Worth benchmarking a large local model first before committing to the OpenAI dependency: cost-per-query stays at zero and there's no rate limit during a re-mine.

PR #912, opened 2026-04-15
fix: centralize embedding model config to prevent query/ingest mismatch

### LINKED ISSUE #903: bug: embedding model mismatch — MCP server uses MiniLM (384-dim) while ingest can use mpnet (768-dim)

OmkarKirpan:
## Summary

Closes #903

Adds centralized embedding model configuration so the MCP server, CLI search, and all ingest paths use the same model — fixing silent query failures when models mismatch.

### Design Decisions

1. **Storage approach:** Embedding model name stored in ChromaDB collection metadata (not a separate file). Atomic with the collection, can't desync. Absence of the key = legacy palace.

2. **Default for new palaces:** `all-mpnet-base-v2` (768-dim) — better search quality (+3.5pp on LoCoMo R@10 benchmarks over MiniLM).

3. **Default for existing palaces:** `all-MiniLM-L6-v2` (384-dim) — backwards compatible, no re-mining required. Detected by absence of `embedding_model` key in collection metadata.

4. **Resolution chain:** Collection metadata (authoritative) > config file / env var (new palaces only) > built-in default. This means once a palace is created, its model is locked in and self-describing.

5. **No migration tool in this PR:** Re-embedding existing palaces from MiniLM to mpnet is a separate concern. This PR prevents the mismatch; migrating existing palaces is a follow-up.

6. **All create paths stamp the model:** Repair, rebuild, and migrate operations preserve the original model through the delete/recreate cycle.

### Resolution chain
```
1. Collection metadata "embedding_model" key (authoritative, stamped at build)
2. If absent → legacy palace → all-MiniLM-L6-v2
3. config.json "embedding_model" or MEMPALACE_EMBEDDING_MODEL env var → new palace creation only
```

### Files changed
- **New:** `mempalace/embedding.py` — model registry, resolution, embedding function factory
- **Modified:** `mempalace/config.py` — `embedding_model` property on MempalaceConfig
- **Modified:** `mempalace/backends/chroma.py` — `get_collection()`, `get_or_create_collection()`, `create_collection()` accept `embedding_function` + `embedding_model_name`
- **Modified:** `mempalace/palace.py` — resolves model from metadata on read, stamps on create
- **Modified:** `mempalace/mcp_server.py` — `_get_collection()` uses correct embedding function, `tool_status()` reports active model
- **Modified:** `mempalace/cli.py`, `mempalace/repair.py`, `mempalace/migrate.py` — all collection-create paths now stamp embedding model

### Known follow-ups (not in scope)
- Performance: double `get_collection` call on read paths for metadata resolution (~ms overhead, within budget)
- Migration tool to re-embed existing MiniLM palaces with mpnet
- OpenAI/external embedding support (#756)
- GPU acceleration (#515)

## Test plan
- [x] 936/936 tests pass locally
- [x] Legacy palace (no metadata key) resolves to MiniLM
- [x] New palace creation stamps mpnet in collection metadata
- [x] Config override and env var override work
- [x] Repair/migrate preserve embedding model through rebuild
- [x] `mempalace_status` reports active embedding model
- [x] `ruff check` and `ruff format` clean

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -164,10 +164,19 @@ class ChromaBackend:
         self._client(palace_path).delete_collection(collection_name)
 
     def create_collection(
-        self, palace_path: str, collection_name: str, hnsw_space: str = "cosine"
+        self,
+        palace_path: str,
+        collection_name: str,
+        hnsw_space: str = "cosine",
+        embedding_function=None,
+        embedding_model_name: str = None,
     ) -> "ChromaCollection":
-        """Create (not get-or-create) *collection_name* with cosine HNSW space."""
+        """Create (not get-or-create) *collection_name* with HNSW space and optional embedding config."""
+        metadata = {"hnsw:space": hnsw_space}
+        if embedding_model_name:
+            metadata["embedding_model"] = embedding_model_name
+        ef_kwargs = {"embedding_function": embedding_function} if embedding_function else {}
         collection = self._client(palace_path).create_collection(
-            collection_name, metadata={"hnsw:space": hnsw_space}
+            collection_name, metadata=metadata, **ef_kwargs
         )
         return ChromaCollection(collection)
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -112,7 +112,14 @@ class ChromaBackend:
     # Collection lifecycle
     # ------------------------------------------------------------------
 
-    def get_collection(self, palace_path: str, collection_name: str, create: bool = False):
+    def get_collection(
+        self,
+        palace_path: str,
+        collection_name: str,
+        create: bool = False,
+        embedding_function=None,
+        embedding_model_name: str = None,
+    ):
         if not create and not os.path.isdir(palace_path):
             raise FileNotFoundError(palace_path)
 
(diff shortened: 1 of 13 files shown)


NickShtefan:
@OmkarKirpan — heads-up that #442 just got rebased onto `develop` (mergeStateStatus moved from DIRTY → UNSTABLE; CI running). My fixes overlap on the `_get_collection` mismatch bug (#903) but the scopes differ:

| | #912 | #442 |
|--|--|--|
| MCP `_get_collection` reads model from collection metadata | ✓ | ✓ |
| `EmbeddingModelMismatchError` | — | ✓ (raised + propagated through `handle_request`) |
| `mempalace init --model <name>` | — | ✓ (model stamped at palace creation) |
| `mempalace re-mine --model <new>` | — | ✓ (safe migration) |
| Auto-detect device (cuda > mps > cpu) | — | ✓ |
| `[multilingual]` extra | — | ✓ |
| Default for new palaces | mpnet-base-v2 (768) | configurable via `--model` |
| Backwards compat for legacy palaces | ✓ (MiniLM fallback) | ✓ (auto-stamped as `chromadb-default`) |

#442 effectively contains your fix as a subset. If maintainers prefer the broader scope, #912 would be superseded. If they prefer the narrower minimal fix first, I can wait for #912 to merge and rebase #442 on top — happy to coordinate either way.

@igorls @bensig — would appreciate guidance on which sequencing you want.

igorls:

OmkarKirpan:
Closing in favor of #442.

The narrow `_get_collection` bug this PR targeted has since been fixed on `develop` via #1303 (merged 2026-05-01) using a different mechanism — passing `embedding_function=` on collection reopen so the EF identity always matches.

The remaining concern (per-palace embedding model so the MCP path uses the same model ingest used, closes #903) is better served by @NickShtefan's #442, which contains this PR's intent as a subset and adds init-time model binding, mismatch detection, and a recovery path. No point shipping a narrower competing version.

Thanks @bensig @igorls @milla-jovovich for the time on this thread.

Issue #872, opened 2026-04-14
bug: hooks use invalid decision value "allow" — should be "approve"


sha2fiddy:
PR #786 changed the save and precompact hooks from `"decision": "block"` to `"decision": "allow"` to stop the agent from writing in chat. The intent was right but `"allow"` isn't a valid value for the top-level `decision` field.

~~The fix is `"allow"` → `"approve"`.~~

**Update:** `"approve"` isn't valid here either — it's a deprecated PreToolUse-only alias (maps to `"allow"` in `hookSpecificOutput.permissionDecision`). The top-level `decision` field only recognizes `"block"`. To not block, you return empty JSON (`{}`).

**Affected files:**
- `hooks/mempal_save_hook.sh` (line 159)
- `hooks/mempal_precompact_hook.sh` (line 74)

**Fix:** Replace the full JSON output block with `{}` when not blocking. Only use `{"decision": "block", "reason": "..."}` when the hook actually needs to prevent the action.

**Note:** PR #863 already caught this — mvalentsev replaced the invalid `{"decision": "allow", ...}` with `{}` in `hooks_cli.py`. The standalone bash hooks still need the same treatment.

Related: #856, #858 (report compaction symptoms that may stem from this)

mvalentsev:
Good catch, thanks. Our #863 was indeed using the wrong value.

One correction though: `"approve"` isn't valid here either. It's a deprecated alias for `"allow"` in PreToolUse `permissionDecision`, not a top-level `decision` value. The top-level `decision` field only recognizes `"block"` -- to not block, you return empty JSON (`{}`).

Just pushed a fix to #863: `{"decision": "allow", ...}` replaced with `{}`.

The standalone bash hooks in `hooks/` have the same problem (`"decision": "allow"` on lines 74 and 159). Those need a separate fix since they're outside our PR scope.

sha2fiddy:
You're right, thanks for the correction. I double-checked the hooks spec — top-level `decision` only supports `"block"`, and non-blocking hooks should just return `{}`. Updated the issue description.

PR #915, opened 2026-04-15
docs(bugfixes): PreCompact deadlock writeup for #863

Robins163:
Follow-up to #867 / #863 as discussed in https://github.com/MemPalace/mempalace/pull/867#issuecomment-4245178126.

Standalone writeup of the PreCompact deadlock — symptom, root cause, and the fix that shipped in #863 (@mvalentsev). Net-new file; \`docs/bugfixes/\` does not exist on develop yet.

## Contents
- Session-log repro (8 fires in 15 minutes from session \`082d4cc3…\`)
- Re-fire timeline showing the block → cancel → save → re-fire loop
- Root cause: \`hooks_cli.hook_precompact()\` returning \`{"decision":"block"}\` unconditionally while Claude Code cancels compaction and re-fires
- The \`decision: "allow"\` correction from #872 (only \`"block"\` is a recognized top-level decision; anything else is a pass-through by accident — \`{}\` is the documented no-block shape)
- Rationale for removing the block entirely vs layering a stateful guard on top
- Escape hatch for anyone stuck on an unpatched version

Branch rebased on \`upstream/develop\`. Tests aren't included — #863's \`test_precompact_allows\` already covers the shape.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,6 +27,12 @@ These are non-negotiable. Every PR, every feature, every refactor must honor the
 - **Privacy by architecture** — The system physically cannot send your data because it never leaves your machine. No telemetry, no phone-home, no external service dependencies for core operations.
 - **Background everything** — Filing, indexing, timestamps, and pipeline work happen via hooks in the background. Nothing interrupts the user's conversation. Zero tokens spent on bookkeeping in the chat window.
 
+## Hook Return Contract
+
+Claude Code recognizes exactly one top-level `decision` value: `"block"`. Anything else is a pass-through by accident — `{"decision": "allow"}` is NOT a recognized value (see [#872](https://github.com/MemPalace/mempalace/issues/872)). To not block, return `{}`.
+
+**PreCompact hooks must never block**: Claude Code cancels compaction and re-fires the hook, causing a deadlock (see [#863](https://github.com/MemPalace/mempalace/issues/863), [#856](https://github.com/MemPalace/mempalace/issues/856), [#858](https://github.com/MemPalace/mempalace/issues/858)). Mine synchronously, then return `{}`.
+
 ## Contributing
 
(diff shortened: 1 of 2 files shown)


mvalentsev:
Been thinking about this. The realistic repeat scenario here is an LLM agent making the same mistake, not a human dev. The `"allow"` value from #872 was exactly that: an agent wrote the return value without checking the hook spec. Agents load `CLAUDE.md` before working on project code but won't discover files in `docs/bugfixes/` on their own. The key rule ("PreCompact hooks must not block, return `{}` not `"allow"`) probably belongs somewhere agents will actually see it.

Robins163:
Fully agree — that's the right framing. The "allow" mistake in #872 was an agent writing a return value without checking the spec, and the fix for *that* class of mistake has to live where agents actually load context, not in `docs/bugfixes/` which an agent won't discover unprompted.

I checked `develop`: neither `CLAUDE.md` nor `AGENTS.md` currently state the hook return contract. The rule has no agent-facing home right now.

Proposing to add a short **Hook return contract** section to `CLAUDE.md` (and mirror the key line in `AGENTS.md`), as an additional commit on this same branch. Draft:

> **Hook return contract.** Claude Code recognizes exactly one top-level `decision` value: `"block"`. Anything else is a pass-through by accident — `{"decision": "allow"}` is NOT a recognized value (see #872). To not block, return `{}`. **PreCompact hooks must never block**: Claude Code cancels compaction and re-fires the hook, causing a deadlock (see #863, #856, #858). Mine synchronously, then return `{}`.

The `docs/bugfixes/precompact-deadlock.md` writeup stays as the long-form reference with the session-log repro and re-fire timeline; CLAUDE.md/AGENTS.md carries the one-paragraph rule that agents actually see.

Does that scope extension work for you on this PR, or would you prefer I split it into a separate PR so #915 stays docs-only?

PR #928, opened 2026-04-15
fix(i18n): resolve language codes case-insensitively (#927)

### LINKED ISSUE #927: fix(i18n): language code lookup is case-sensitive — '--lang PT-BR' silently falls back to English
## Bug

Language file lookup in `mempalace/i18n/__init__.py` is case-sensitive and the locale files use inconsistent casing for the region subtag:

```
mempalace/i18n/de.json
mempalace/i18n/en.json
mempalace/i18n/es.json
mempalace/i18n/fr.json
mempalace/i18n/it.json
mempalace/i18n/ja.json
mempalace/i18n/ko.json
mempalace/i18n/pt-br.json   ← lowercase region
mempalace/i18n/ru.json
mempalace/i18n/zh-CN.json   ← uppercase region
mempalace/i18n/zh-TW.json   ← uppercase region
```

On case-sensitive filesystems (Linux, default APFS-CS on macOS, Windows when Python's path comparison is strict), passing `--lang PT-BR`, `--lang Pt-Br`, `--lang zh-cn`, or `--lang ZH-TW` silently misses the file. `_load_entity_section()` returns `{}`, the merge fallback fires English, and entity detection runs in the wrong language with **no warning**.

The cache key in `get_entity_patterns()` is `tuple(languages)` raw, so `("PT-BR",)` ≠ `("pt-br",)` — even mixing case across calls bypasses the cache and re-runs the wrong fallback every time. `cli.py:81` also persists the wrong-case value to the user's config via `cfg.set_entity_languages(languages)`.

## Reproduction

```bash
mempalace mine ~/docs --lang PT-BR
# Expected: Brazilian Portuguese entity patterns merged
# Actual: silently runs English-only entity detection; config persists "PT-BR"
```

## Why it matters

- BCP 47 language tags are case-insensitive by spec (RFC 5646 §2.1.1).
- The inconsistent file naming (`pt-br` vs `zh-CN`) is a footgun for users who copy from documentation that uses the canonical capitalization.
- The silent English fallback on a multilingual user's content means missed entities and broken cross-room search — the symptom is invisible.

## Suggested fix

Resolve language codes via a case-insensitive lookup against the actual filenames in `_LANG_DIR`. Normalize the cache key the same way so callers using different casing share the same merged dict. Apply the same normalization in `load_lang()`.

Happy to send a PR — already drafted on `fix/i18n-lang-case-insensitive`.

arnoldwender:
## What and Why

BCP 47 language tags are case-insensitive (RFC 5646 §2.1.1), but the locale files in `mempalace/i18n/` mix conventions:

```
pt-br.json   ← lowercase region
zh-CN.json   ← uppercase region
zh-TW.json   ← uppercase region
```

On case-sensitive filesystems (Linux, default APFS-CS macOS, strict Windows), `--lang PT-BR`, `--lang zh-cn`, or `--lang ZH-TW` silently missed the file. `_load_entity_section()` returned `{}`, the merge loop's `found_any` stayed False, and entity detection ran in **English** with no warning.

The cache key in `get_entity_patterns()` was the raw input tuple, so `("PT-BR",)` and `("pt-br",)` produced two distinct entries — both wrong.

## Reproduction

```bash
mempalace mine ~/docs --lang PT-BR
# Expected: Brazilian Portuguese entity patterns merged
# Actual:   silently runs English-only entity detection
```

## Change Summary

- `mempalace/i18n/__init__.py` — add `_canonical_lang(lang)` that resolves any casing to the on-disk filename stem via lowercase comparison. Route `load_lang`, `_load_entity_section`, and the `get_entity_patterns` cache key through it. Behaviour for known locales is unchanged; only the previously-broken case-mismatched paths now succeed.
- `tests/test_i18n_lang_case.py` — 8 regression tests: canonical resolution (lower/upper/unknown/empty), `load_lang` case insensitivity, entity-section parity across cases, cache deduplication across cases, English fallback for genuinely unknown codes still works.

## Test Plan

- [x] `ruff check .` and `ruff format --check .` — clean
- [x] Full suite: `pytest tests/ --ignore=tests/benchmarks` — **953 passed**
- [x] New regression suite proves the bug existed: `_load_entity_section('PT-BR')` returned `{}` on `develop` and now returns the same dict as `'pt-br'`
- [x] Confirmed cache deduplication: 3 calls with different casings of `zh-CN` produce **1** cache entry, not 3

Closes #927

### DIFF
--- a/mempalace/i18n/__init__.py
+++ b/mempalace/i18n/__init__.py
@@ -15,6 +15,7 @@ and the README section "Adding a new language" for the schema.
 
 import json
 from pathlib import Path
+from typing import Optional
 
 _LANG_DIR = Path(__file__).parent
 _strings: dict = {}
@@ -24,7 +25,7 @@ _current_lang: str = "en"
 _entity_cache: dict = {}
 
 
-def _canonical_lang(lang: str) -> str | None:
+def _canonical_lang(lang: str) -> Optional[str]:
     """Resolve a language code to its on-disk canonical filename stem.
 
     BCP 47 tags are case-insensitive (RFC 5646 §2.1.1), and the locale
--- a/mempalace/i18n/__init__.py
+++ b/mempalace/i18n/__init__.py
@@ -24,6 +24,23 @@ _current_lang: str = "en"
 _entity_cache: dict = {}
 
 
+def _canonical_lang(lang: str) -> str | None:
+    """Resolve a language code to its on-disk canonical filename stem.
+
+    BCP 47 tags are case-insensitive (RFC 5646 §2.1.1), and the locale
+    files mix conventions (``pt-br.json`` vs ``zh-CN.json``). Match on
+    lowercase so callers can pass ``PT-BR``, ``zh-cn``, ``Pt-Br``, etc.
+    Returns ``None`` if no file matches.
+    """
+    if not lang:
+        return None
+    target = lang.strip().lower()
+    for path in _LANG_DIR.glob("*.json"):
+        if path.stem.lower() == target:
+            return path.stem
+    return None
+
+
 def available_languages() -> list[str]:
     """Return list of available language codes."""
     return sorted(p.stem for p in _LANG_DIR.glob("*.json"))
@@ -32,12 +49,12 @@ def available_languages() -> list[str]:
 def load_lang(lang: str = "en") -> dict:
     """Load a language dictionary. Falls back to English if not found."""
     global _strings, _current_lang
-    lang_file = _LANG_DIR / f"{lang}.json"
-    if not lang_file.exists():
-        lang_file = _LANG_DIR / "en.json"
-        lang = "en"
+    canonical = _canonical_lang(lang)
+    if canonical is None:
+        canonical = "en"
+    lang_file = _LANG_DIR / f"{canonical}.json"
     _strings = json.loads(lang_file.read_text(encoding="utf-8"))
-    _current_lang = lang
+    _current_lang = canonical
     return _strings
 
 
@@ -81,9 +98,10 @@ def get_regex() -> dict:
 
 def _load_entity_section(lang: str) -> dict:
     """Load the raw entity section for one language. Returns {} if missing."""
-    lang_file = _LANG_DIR / f"{lang}.json"
-    if not lang_file.exists():
+    canonical = _canonical_lang(lang)
+    if canonical is None:
         return {}
+    lang_file = _LANG_DIR / f"{canonical}.json"
     try:
         data = json.loads(lang_file.read_text(encoding="utf-8"))
     except (json.JSONDecodeError, OSError):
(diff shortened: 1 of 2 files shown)


arnoldwender:
CI failure was my mistake — used PEP 604 `str | None` which requires Python 3.10+. Fixed in [6caac50](https://github.com/MemPalace/mempalace/pull/928/commits/6caac50) by switching to `typing.Optional[str]`. Local 3.9 + 3.13 verified clean.

mvalentsev (commented):
Nice catch on the BCP 47 case sensitivity. The bug is real and your fix handles it correctly.

One alternative worth considering: the root cause is that locale filenames mix conventions (`pt-br.json` lowercase vs `zh-CN.json` mixed-case). If we rename the files to all-lowercase (`zh-cn.json`, `zh-tw.json`), the resolution becomes trivial:

```python
lang_file = _LANG_DIR / f"{lang.strip().lower()}.json"
```

No glob scan, no lookup table, no new function. `pt-br.json` already follows this convention, so the inconsistency is only in the Chinese locale filenames.

The tradeoff is that renaming existing files is a (small) breaking change for anyone referencing them by path, while your approach works with any naming convention. Both are valid, curious what maintainers prefer.

arnoldwender:
Good point on the naming inconsistency. I'd lean toward my current approach for two reasons:

1. **Robustness to user input, not just shipped files.** Even with locale files normalized to lowercase, callers still pass codes from browser `Accept-Language` headers, config files, CLI args — all of which can arrive in any case (`zh-CN`, `ZH-cn`, `zh_cn`, etc.). The resolver handles that regardless of how we name the files.

2. **Rename is a breaking change for downstream consumers.** Anyone importing locale files by path (themes, plugins, third-party tooling) would need coordinated updates.

That said, the two aren't mutually exclusive — renaming files to lowercase *and* keeping the case-insensitive resolver gives us both the simpler internal code path and the external robustness. Happy to add the rename to this PR if maintainers prefer, or split it into a follow-up. Your call.

mvalentsev:
Fair point on needing normalization regardless of file naming. But `.lower()` at the API boundary *is* the resolver -- one line instead of a function that globs the filesystem on every call:

```python
lang = lang.strip().lower()
lang_file = _LANG_DIR / f"{lang}.json"
```

The locale JSON files are an internal implementation detail, not a public API. Downstream consumers interact through `load_lang()` / `get_entity_patterns()`, not by opening files by path. Renaming two files (`zh-CN` / `zh-TW` to lowercase) is a safe internal change.

Both approaches work. Leaving it to maintainers to pick whichever they prefer.

arnoldwender:
Fair — you're right that `load_lang()` is the only supported entry point, so the "public API" concern is weaker than I framed it. I'm happy either way. If maintainers prefer the rename-plus-`.lower()` approach, say the word and I'll fold it into this PR.

Issue #934, opened 2026-04-16
bug: 7 issues in v3.0.0→v3.3.0 — data loss in repair/migrate, missing input validation, retry gaps


shaun0927:
## Context

While reviewing the v3.0.0→v3.3.0 diff I found 7 issues that are not covered by any existing issue or PR. I checked #883, #809, #893, #851, #852, and #661 before filing this.

All code references are against the current `develop` branch (commit 4aa7e1e, v3.3.0 tag).

---

## Data Integrity

These three violate CLAUDE.md's design principle: *"Incremental only — A crash mid-operation must leave the existing palace untouched."*

### 1. `repair.py:265-275` — rebuild_index deletes collection before upsert completes

`rebuild_index` calls `backend.delete_collection()` at line 265, then upserts drawers in a loop (lines 269-275) with no try/except. If the upsert crashes partway through (OOM, ChromaDB error), the remaining drawers are lost. The backup copy is made before deletion (line 260), but the code never restores from it on failure — the user must do it manually.

```python
# Current (line 265-275)
backend.delete_collection(palace_path, COLLECTION_NAME)
new_col = backend.create_collection(palace_path, COLLECTION_NAME)
for i in range(0, len(all_ids), batch_size):
    # no try/except — crash here = partial data loss
    new_col.upsert(...)
```

**Fix:** Wrap the upsert loop in try/except and restore `chroma.sqlite3` from the backup file on failure.

### 2. `migrate.py:234-235` — non-atomic palace swap

```python
shutil.rmtree(palace_path)           # line 234 — old palace deleted
shutil.move(temp_palace, palace_path) # line 235 — new palace moved in
```

If the process dies between lines 234 and 235, both the old palace (deleted) and the new palace (not yet moved) are gone. The backup at line 204 mitigates this but recovery is manual.

**Fix:** Use `os.rename()` to swap atomically: rename old to `.old`, rename temp to palace, then remove `.old`.

### 3. `cli.py:276` — cmd_repair extraction uses fixed offset increment

```python
offset += batch_size  # always 5000, regardless of actual batch size
```

If a batch returns fewer than `batch_size` items (e.g. due to concurrent modification or the last page), subsequent batches skip `batch_size - actual_count` drawers. The extracted data is then used to rebuild the collection, so any skipped drawers are permanently lost. Compare with `exporter.py:70` which correctly uses `if not batch["ids"]: break`.

[...]

PR #239, opened 2026-04-08
fix: safer repair command that survives HNSW index corruption

rusel95:
## Summary

The `repair` command previously deleted and recreated the ChromaDB collection in-place. When the HNSW index is corrupted, even the `delete_collection()` call touches the broken segment and can segfault — making repair impossible on the very palaces that need it most.

This rewrites repair to use a **migrate-to-fresh-palace** strategy: read all data via `get()` (which bypasses HNSW entirely), write to a brand-new palace at a temporary path, verify counts match, then atomically swap directories. The original palace is always preserved as a `.backup`.

**Relates to:** #211 (scaling fixes for large vaults with 40k+ drawers — our repair approach was validated on a 61K-drawer palace)

## Changes

- Read existing drawers via `get()` — bypasses HNSW, works on corrupted indices
- Write to fresh palace at `_rebuild` path — never mutate the original
- Smaller batch sizes (read: 500, write: 100) to prevent OOM on large palaces
- Verify rebuilt count matches before swapping
- Error handling: partial rebuild preserved at `_rebuild` path, original untouched
- Atomic swap: original → `.backup`, rebuild → palace (with crash recovery notes)
- Infinite loop guard: break if `get()` returns empty batch
- Progress reporting every 5000 reads / 2000 writes

## Test plan

- [x] Run `mempalace repair` on a healthy palace (50 drawers) — rebuilt without data loss
- [x] Run `mempalace repair` on a palace with corrupted HNSW index — all 30 drawers recovered
- [x] Verify `.backup` directory contains original palace
- [x] Verify rebuilt palace supports both `add()` and `query()` operations
- [x] All 116 existing tests pass (0 regressions)

```
$ python -m pytest tests/ -v
========================== 116 passed in 62s ==========================
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #935, opened 2026-04-16
fix: guard against data loss in repair, migrate, and CLI rebuild

shaun0927:
## Summary

Three data-integrity fixes honoring CLAUDE.md's design principle: *"A crash mid-operation must leave the existing palace untouched."*

Refs: #934 (parent) — specifically #1151, #1152, #1153.

### `mempalace/repair.py` — restore from backup on rebuild failure (#1151)

`rebuild_index` calls `delete_collection` then upserts drawers in a loop. If any upsert fails partway (OOM, ChromaDB error), the rest of the drawers are gone. The backup copy exists but is never consulted.

```python
try:
    for i in range(0, len(all_ids), batch_size):
        new_col.upsert(...)
except Exception:
    shutil.copy2(backup_path, sqlite_path)  # restore pre-repair state
    raise
```

### `mempalace/migrate.py` — atomic palace swap (#1152)

`shutil.rmtree(palace_path)` followed by `shutil.move(...)` has a window where both copies are gone. Replaced with a rename-aside pattern. The second commit in this PR (659cb81) hardens the cross-device fallback per @qodo-ai-reviewer's feedback:

- Primary swap uses `os.replace` (same-filesystem atomic)
- `errno.EXDEV` gates the `shutil.move` fallback so unrelated `OSError`s surface instead of getting masked
- `_restore_stale_palace` clears any partial destination before the rollback `os.replace`, and logs both paths if restore itself fails

### `mempalace/cli.py` — defensive batch extraction offset (#1153)

`cmd_repair` extraction used `offset += batch_size` (fixed 5000) regardless of actual batch length. Changed to `offset += len(batch["ids"])` with an empty-batch guard, matching the pattern at `exporter.py:70`.

## Test plan

- [x] `ruff check mempalace/` — passed
- [x] `ruff format --check mempalace/` — passed
- [x] `pytest tests/test_migrate.py` — 5 passed (including 3 new regression tests covering the rollback hardening)

## Overlap with open PRs

Per @mvalentsev's review:
- **#239** (migrate-to-fresh repair rewrite) — if it lands first, the `repair.py` and `cli.py` hunks here become moot. Until then, they prevent data loss on the current code path.
- **#890** (Windows handle cleanup in swap path) — orthogonal; either order works without conflict.
- **#632** (nuke-rebuild repair) — same as #239.

## Split question

@bensig asked on #934 that each finding get its own PR. This bundles three under the "crash-safety" theme in three independent files. Happy to split into three per-issue PRs if preferred.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -270,10 +270,12 @@ def cmd_repair(args):
     offset = 0
     while offset < total:
         batch = col.get(limit=batch_size, offset=offset, include=["documents", "metadatas"])
+        if not batch["ids"]:
+            break
         all_ids.extend(batch["ids"])
         all_docs.extend(batch["documents"])
         all_metas.extend(batch["metadatas"])
-        offset += batch_size
+        offset += len(batch["ids"])
     print(f"  Extracted {len(all_ids)} drawers")
 
     # Backup and rebuild
--- a/mempalace/migrate.py
+++ b/mempalace/migrate.py
@@ -239,8 +239,11 @@ def migrate(palace_path: str, dry_run: bool = False, confirm: bool = False):
     try:
         os.rename(temp_palace, palace_path)
     except OSError:
-        # os.rename fails across filesystems; fall back to move
-        shutil.move(temp_palace, palace_path)
+        try:
+            shutil.move(temp_palace, palace_path)
+        except Exception:
+            os.rename(stale_path, palace_path)
+            raise
     shutil.rmtree(stale_path, ignore_errors=True)
 
     print("\n  Migration complete.")
--- a/mempalace/migrate.py
+++ b/mempalace/migrate.py
@@ -229,10 +229,19 @@ def migrate(palace_path: str, dry_run: bool = False, confirm: bool = False):
     del col
     del fresh_backend
 
-    # Swap: remove old palace, move new one into place
+    # Swap: rename old palace aside, then move new one into place.
+    # This avoids a window where both old and new are missing.
     print("  Swapping old palace for migrated version...")
-    shutil.rmtree(palace_path)
-    shutil.move(temp_palace, palace_path)
+    stale_path = palace_path + ".old"
+    if os.path.exists(stale_path):
+        shutil.rmtree(stale_path)
+    os.rename(palace_path, stale_path)
+    try:
+        os.rename(temp_palace, palace_path)
+    except OSError:
+        # os.rename fails across filesystems; fall back to move
+        shutil.move(temp_palace, palace_path)
+    shutil.rmtree(stale_path, ignore_errors=True)
 
     print("\n  Migration complete.")
     print(f"  Drawers migrated: {final_count}")
--- a/mempalace/migrate.py
+++ b/mempalace/migrate.py
@@ -16,11 +16,32 @@ Usage:
     mempalace migrate --dry-run                # show what would be migrated
 """
 
+import errno
 import os
 import shutil
 import sqlite3
 from collections import defaultdict
 from datetime import datetime
+
+
+def _restore_stale_palace(palace_path: str, stale_path: str) -> None:
+    """Roll back a failed swap.
+
+    shutil.move() can partially create palace_path before raising, which
+    would make a bare os.replace(stale_path, palace_path) fail (dest exists).
+    Clear any partial destination first, then restore. Best-effort: if the
+    restore itself fails, log both paths so the operator can recover by hand.
+    """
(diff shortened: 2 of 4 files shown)


mvalentsev (commented):
Finding 1 (repair crash safety) overlaps with open PR #239 (rusel95), which rewrites repair to a migrate-to-fresh-palace strategy: reads all data via get(), writes to a temp palace, verifies counts match, then atomically swaps directories. Original palace preserved as .backup with auto-restore on failure.

Findings 2 and 3 look clean to me.

mvalentsev:
Correction to my earlier review:

Finding 2 (migrate swap): PR #890 touches the same swap code path in migrate.py, adding _replace_palace_dir() with retry logic. Different focus (Windows handle cleanup) but related.

Finding 3 (cli.py offset): PRs #239 and #632 both rewrite the repair code path entirely (migrate-to-fresh strategy and nuke-rebuild respectively). If either lands, the extraction loop with the offset bug no longer exists.

Qodo-Free-For-OSS:
Hi, In migrate(), if the cross-device fallback (shutil.move) partially creates the destination directory and then fails, the rollback os.rename(stale_path, palace_path) will also fail because palace_path already exists, leaving a partial migrated palace at the target and the original only at .old.

**How to fix:** Harden rollback and cleanup

> ### Issue description
> `mempalace.migrate.migrate()` can leave the palace in an inconsistent state if the cross-device swap fallback (`shutil.move`) fails after partially creating `palace_path`. The rollback then tries `os.rename(stale_path, palace_path)` which will fail if `palace_path` already exists.
> 
> ### Issue Context
> The new logic correctly avoids the “both missing” window, but needs to handle the case where `shutil.move(temp_palace, palace_path)` partially copies and creates `palace_path` before raising.
> 
> ### Fix Focus Areas
> - mempalace/migrate.py[235-246]
> 
> ### What to change
> - In the `except Exception:` rollback path:
>   - If `palace_path` exists (directory or file), remove it first (e.g., `shutil.rmtree(palace_path, ignore_errors=False)` / `os.remove` for file), then restore `stale_path` into place.
>   - Consider wrapping rollback in its own try/except to emit a clear recovery message that names both paths (`stale_path` and partially-copied `palace_path`).
> - Optionally, use `os.replace` for same-filesystem renames, and distinguish cross-device errors (EXDEV) from other OSErrors to avoid masking real rename failures.

---

shaun0927:
Pushed 659cb81 addressing both pieces of review feedback.

**@qodo-ai-reviewer** — rollback hardening:
- Primary swap moved to `os.replace` so same-filesystem is atomic
- `errno.EXDEV` branch gates the `shutil.move` fallback — any non-EXDEV OSError now surfaces instead of getting masked by a slow copy
- Rollback extracted into `_restore_stale_palace` which clears the partial destination before `os.replace(stale, palace)`, and logs both paths if the restore itself fails
- Three regression tests in `tests/test_migrate.py` (clean rollback, partial-copy cleanup, failure-path logging)

**@mvalentsev** — overlap notes:
- Finding 1 vs #239: agree it'll be moot if the migrate-to-fresh rewrite lands. Happy to close the repair.py hunk if a maintainer wants to prioritize #239; otherwise it's a strict improvement over today's `rebuild_index`.
- Finding 2 vs #890: the Windows handle cleanup in `_replace_palace_dir()` is orthogonal to the atomicity fix here — if #890 goes in first, the rename pattern here can be dropped into it; if this goes in first, #890 slots on top without conflict.
- Finding 3 vs #239/#632: same situation — harmless if they land, prevents data loss on the current extraction loop until they do.

Also split the parent #934 into six child issues (#1151–#1156; finding 6 withdrawn). Filed corresponding child references: #1151 / #1152 / #1153.

**Question for maintainers (@bensig):** bensig asked on #934 that each finding get its own PR. This PR bundles three fixes under the "crash-safety" theme in three independent files. Happy to split into three per-child-issue PRs if preferred — say the word and I'll carve this up.

igorls:
Thanks for the careful split-up on #934 and for addressing the Qodo rollback feedback in 659cb81 — the rename-aside + EXDEV branch + partial-destination cleanup looks right, and the three regression tests cover the hazards we care about.

Let's keep this bundled. The three hunks live in independent files under one coherent crash-safety theme, so a split would just multiply review cycles for the same changes.

Could you rebase on `develop` when you get a chance? The branch is ~200 commits behind, so we'd like CI to run against current `develop` before merge. Once that's pushed I'll approve the workflow run.

PR #937, opened 2026-04-16
fix: LLM retry on JSONDecodeError + exporter symlink resolution

shaun0927:
## Summary

Two independent resilience fixes. Finding 6 (metadata cache reset in `tool_reconnect`) was dropped after @mvalentsev's call-chain analysis showed `_get_client` already clears the cache via the inode/mtime mismatch branch.

Refs: #934 — specifically #1155 (finding 5) and #1156 (finding 7).

### `mempalace/closet_llm.py` — retry on JSONDecodeError (#1155)

`_call_llm` runs a 3-attempt retry loop. HTTP 429/503 correctly `continue`, but `json.JSONDecodeError` from malformed LLM output immediately returned `(None, None)`, exiting the loop on first bad response. LLM output is nondeterministic — a single malformed response should not permanently skip the source file.

```python
# Before
except json.JSONDecodeError:
    return None, None    # exits the loop

# After
except json.JSONDecodeError:
    continue              # LLM output is nondeterministic; retry
```

### `mempalace/exporter.py` — resolve symlinks on output path (#1156)

`os.makedirs(output_dir)` follows symlinks. The miner already refuses symlinked inputs (`miner.py`, `convo_miner.py`); the exporter should apply the same rule to outputs. Defense-in-depth.

```python
# Before
os.makedirs(output_dir, exist_ok=True)

# After
output_dir = os.path.realpath(output_dir)
os.makedirs(output_dir, exist_ok=True)
```

## Dropped: finding 6 (metadata cache reset)

Dropped in d7e3197. The "fix" was redundant — `tool_reconnect` already clears `_metadata_cache` indirectly through the `_get_client` mismatch path.

## Checklist

- [x] `ruff check` — passed
- [x] `ruff format --check` — passed

### DIFF
--- a/mempalace/closet_llm.py
+++ b/mempalace/closet_llm.py
@@ -160,7 +160,7 @@ def _call_llm(cfg: LLMConfig, source_file: str, wing: str, room: str, content: s
             parsed = json.loads(text)
             return parsed, payload.get("usage")
         except json.JSONDecodeError:
-            return None, None
+            continue  # LLM output is nondeterministic; retry
         except urllib.error.HTTPError as e:
             # 429 / 503 = retry with backoff
             if e.code in (429, 503) and attempt < 2:
--- a/mempalace/exporter.py
+++ b/mempalace/exporter.py
@@ -48,6 +48,7 @@ def export_palace(palace_path: str, output_dir: str, format: str = "markdown") -
         print("  Palace is empty — nothing to export.")
         return {"wings": 0, "rooms": 0, "drawers": 0}
 
+    output_dir = os.path.realpath(output_dir)
     os.makedirs(output_dir, exist_ok=True)
     try:
         os.chmod(output_dir, 0o700)
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1118,10 +1118,17 @@ def tool_reconnect():
     Use after external scripts or CLI commands modify the palace database
     directly, which can leave the in-memory HNSW index stale.
     """
-    global _collection_cache, _palace_db_inode, _palace_db_mtime
+    global \
+        _collection_cache, \
+        _palace_db_inode, \
+        _palace_db_mtime, \
+        _metadata_cache, \
+        _metadata_cache_time
     _collection_cache = None
     _palace_db_inode = 0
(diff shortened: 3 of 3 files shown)


mvalentsev (commented):
Finding 6 (metadata cache in tool_reconnect) is already handled on current develop. The call chain:

1. `tool_reconnect` zeroes `_palace_db_inode = 0` and `_palace_db_mtime = 0.0` (lines 1122-1124)
2. Calls `_get_collection()` (line 1126) which calls `_get_client()` (line 218)
3. `_get_client()` compares zeroed values against the real file's inode/mtime (line 201), detects mismatch, enters the rebuild branch (line 204)
4. Resets `_metadata_cache = None` and `_metadata_cache_time = 0` at lines 207-208
5. Back in `_get_collection()`, `_collection_cache is None` triggers the elif (line 227), which resets metadata again at lines 229-230

The metadata cache is cleared before `tool_reconnect` returns. Adding an explicit reset in `tool_reconnect` itself is harmless but redundant.

shaun0927:
Pushed d7e3197 dropping the metadata-cache reset (finding 6).

@mvalentsev — you're right, the call-chain argument holds. `tool_reconnect` already clears `_metadata_cache` indirectly via `_get_client`'s inode/mtime mismatch branch, so the explicit reset was fixing nothing. Kept it out rather than shipping a "harmless redundancy" with a commit message that pretends it's a fix.

This PR now covers just the two real bugs:
- finding 5 (#1155): `_call_llm` retry on `JSONDecodeError`
- finding 7 (#1156): `exporter.py` output path follows symlinks

Updating the PR body to match. If @bensig wants these split into two PRs per the one-bug-one-PR rule on #934, happy to carve.

igorls:
Bumping — both findings verified REAL on develop:

- **#1155**: `mempalace/closet_llm.py:162-163` still `return None, None` on `JSONDecodeError` instead of continuing through retries.
- **#1156**: `mempalace/exporter.py:51,92` still call `os.makedirs(...)` without realpath / containment check.

Both pulled into v3.3.5. Tiny diff (+2/-1) suggests this might just need a CI nudge — please confirm green on latest commit.

PR #277, opened 2026-04-08
fix: improve Windows test compatibility

vanachterjacob:
﻿## What does this PR do?

- adds a lightweight Windows test job to CI
- explicitly closes Chroma clients in test teardown to avoid Windows file-lock cleanup failures
- keeps the change scoped to tests/CI only

Closes #275.

## How to test

- `python -m pytest tests/ -v`
- On Windows, specifically verify:
  - `python -m pytest tests/test_convo_miner.py tests/test_miner.py -v`

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`)
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`)

PR #939, opened 2026-04-16
ci: add pip caching and bump Python on macOS/Windows

mvalentsev:
## What

- Enable `setup-python`'s built-in `cache: 'pip'` on all four CI jobs (test-linux, test-windows, test-macos, lint). Avoids re-downloading ~300 MB of dependencies (chromadb, onnxruntime, hnswlib) on every run.
- Bump macOS and Windows from Python 3.9 to 3.11. Linux matrix already covers 3.9/3.11/3.13, so cross-platform runs on 3.9 are redundant. Python 3.11 is meaningfully faster and has better ARM support on macOS.

## Why

test-windows typically takes 3m30s–5m vs ~1m30s on Linux. Dependency installation accounts for ~95 MB of wheel downloads per push, which `cache: 'pip'` avoids on subsequent runs.

## Related

#277 also proposes bumping Windows to 3.11 (as part of a broader CI restructure). This PR is a smaller, standalone change that can land independently.

## Test plan

- [x] CI passes on all platforms (Linux 3.9/3.11/3.13, Windows 3.11, macOS 3.11, lint)
- [x] `cache: 'pip'` active on all four jobs (visible in the `Set up Python` step)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'pip'
       - run: pip install -e ".[dev]"
       - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10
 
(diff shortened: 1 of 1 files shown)


mvalentsev:
@igorls would appreciate your eyes on this when you get a chance.

PR #924, opened 2026-04-15
fix(mine): log warning when files exceed MAX_FILE_SIZE (#923)

mvalentsev:
## Summary

`mempalace mine` (both default and `--mode convos`) silently drops files at
the 500 MB `MAX_FILE_SIZE` threshold. No log, no counter, exit code 0. The
output is indistinguishable from a directory with no mineable files (the
original #923 complaint).

This PR logs a SKIP line on stderr per dropped file, matching the existing
`SKIP: <rel> (symlink)` stderr convention already present in the same
`scan_project` and `scan_convos` functions.

The `except OSError: continue` arm right below the size check is the same
bug class. A file whose `stat()` raises (permission denied, racing delete,
broken symlink that survived the earlier `is_symlink()` check) also drops
silently. This PR logs that case too.

## Sites covered

| File:line | Skip cause | Stream |
|---|---|---|
| `mempalace/miner.py:1160` | `st_size > MAX_FILE_SIZE` | stderr |
| `mempalace/miner.py:1167` | `stat()` OSError | stderr |
| `mempalace/convo_miner.py:372` | `st_size > MAX_FILE_SIZE` | stderr |
| `mempalace/convo_miner.py:379` | `stat()` OSError | stderr |

Sister site `mempalace/format_miner.py:344` (added in 3.3.6 via #1555)
already logs the same event via `logger.info("skip:too_large ...")` and is
unchanged here. Logging-style consolidation across the three miners is out
of scope for this PR.

Files that survive the walk and skip later in the processing loop (OSError
on `read_text`, content shorter than `MIN_CHUNK_SIZE`) are already counted
in the `Files skipped (already filed or other)` summary bucket. Those are
not the `#923` invisible-drop case. Bucketing the skip reasons by cause is
the same refactor as reporter spec item 2 below and is left out of this PR.

## Reporter spec coverage

#923 listed three asks; this PR addresses the first:

1. Warning per skipped file. Covered (stderr).
2. Distinct counter in the `Done.` summary. Not in this PR. `scan_project`
   filters oversized files out before they reach `mine()`'s
   `files_skipped` accounting; adding a counter would require changing
   the scan return shape (touched by 3 callers). Left for maintainer's
   judgment on whether the warning alone suffices.
3. `--max-file-size` CLI flag. Not in this PR. Reporter listed it as
   "ideally". Same maintainer-judgment note.

## Test plan

- [x] `pytest tests/test_miner.py::test_scan_project_skips_oversized_files`
[...]

PR #940, opened 2026-04-16
fix: handle ChatGPT bulk export list format and raise file size limit

Deen-Wong:
## What this fixes

Two bugs that prevent ChatGPT bulk exports from being mined correctly.

### Bug 1: File size limit too low (convo_miner.py)
MAX_FILE_SIZE was set to 10MB. ChatGPT exports split conversation history across multiple files, many of which exceed 10MB. Files over the limit are silently skipped with no warning to the user.

Fix: raised to 200MB.

### Bug 2: ChatGPT list format not handled (normalize.py)
_try_chatgpt_json() only handled a single conversation object (dict with mapping key). ChatGPT bulk exports are lists of conversation objects. When a list was passed, the function returned None and fell through to _try_slack_json(), producing 0 or 1 drawers instead of hundreds.

Fix: added list handling to _try_chatgpt_json() and extracted _try_chatgpt_single() as the per-conversation parser.

## How to test
Export your ChatGPT history and run:

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -55,7 +55,7 @@ CONVO_EXTENSIONS = {
 
 MIN_CHUNK_SIZE = 30
 CHUNK_SIZE = 800  # chars per drawer — align with miner.py
-MAX_FILE_SIZE = 200 * 1024 * 1024  # 10 MB — skip files larger than this
+MAX_FILE_SIZE = 50 * 1024 * 1024  # 50 MB — skip files larger than this
 
 
 def _register_file(collection, source_file: str, wing: str, agent: str):
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -55,7 +55,7 @@ CONVO_EXTENSIONS = {
 
 MIN_CHUNK_SIZE = 30
 CHUNK_SIZE = 800  # chars per drawer — align with miner.py
-MAX_FILE_SIZE = 10 * 1024 * 1024  # 10 MB — skip files larger than this
+MAX_FILE_SIZE = 200 * 1024 * 1024  # 10 MB — skip files larger than this
 
 
 def _register_file(collection, source_file: str, wing: str, agent: str):
--- a/mempalace/normalize.py
+++ b/mempalace/normalize.py
@@ -16,97 +16,8 @@ No API key. No internet. Everything local.
 
 import json
 import os
-import re
 from pathlib import Path
 from typing import Optional
-
-# Provenance footer appended to Slack transcript output so downstream consumers
-# know the speaker roles are positionally assigned, not verified.
-_SLACK_PROVENANCE_FOOTER = (
-    "\n[source: slack-export | multi-party chat — speaker roles are positional, not verified]"
-)
-
-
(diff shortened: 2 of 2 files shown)


mvalentsev:
The ChatGPT list-of-conversations handling looks like a valid addition. A couple of concerns with the rest of the diff though:

**strip_noise() removal** -- this was deliberately added across three commits (9b99c13, ca2598a, 7e5eeda) and has a `NORMALIZE_VERSION` schema gate in `palace.py` so existing drawers get silently rebuilt. Removing it means Claude Code system tags, hook output, and UI chrome all end up in drawers again and pollute search results.

**Slack sanitization removal** -- the `re.sub` on user IDs guards against chunk-boundary injection via crafted exports, and the `[{user_id}]` prefix preserves who said what in multi-party chats. Dropping both is a security/data regression.

**MAX_FILE_SIZE 200 MB** -- #396 was specifically about OOM on large transcript files. 20x the current limit risks reintroducing that. The comment on line 58 still says "10 MB" too. #924 already adds SKIP logging so users know when files are skipped.

Would it make sense to split the ChatGPT list handling into its own focused PR? The normalize.py changes unrelated to that feature seem risky to land together.

Deen-Wong:
Thanks for the detailed review — these are valid concerns.

You're right that copying the full normalize.py from my local venv 
inadvertently included changes beyond the ChatGPT list fix. The 
strip_noise() removal and Slack sanitization changes were not 
intentional — I should have done a surgical diff instead.

On MAX_FILE_SIZE: fair point on the OOM risk. Would 50MB be a 
reasonable middle ground, or is there a better approach given #924 
adds SKIP logging?

I'll revert to a targeted change — only the ChatGPT list handling 
in normalize.py and the file size adjustment in convo_miner.py, 
leaving strip_noise() and Slack sanitization intact.

Deen-Wong:
Thanks for the detailed review — these are valid concerns.
The strip_noise() removal and Slack sanitization changes were unintentional — I copied the full normalize.py from my local patched venv instead of doing a surgical diff. I've reverted those and the new commit contains only the ChatGPT list handling.
On MAX_FILE_SIZE: raised to 50MB as a middle ground rather than 200MB. Happy to defer to whatever the maintainers prefer, given the OOM history in #396 — the comment on line 58 is also updated to match.
The diff is now focused: only _try_chatgpt_json and _try_chatgpt_single in normalize.py, and the file size constant in convo_miner.py.

igorls:

Deen-Wong:
Rebased onto develop — branch is up to date, no conflicts. Ready for review.

> Hi, thanks for the contribution.
> 
> This PR has merge conflicts with `develop`, and the branch has not been updated in over 7 days, which puts it before our most recent release. The conflicts are likely against work that landed in that release.
> 
> Could you rebase onto `develop` so we can take another look?
> 
> If this change is no longer relevant, feel free to close the PR.
> 
> (This message is part of a periodic backlog pass, sent to all open PRs that match this state.)

Rebased onto develop — branch is up to date, no conflicts. Ready for review.

Deen-Wong:
Closing due to inactivity. The fix is documented and available for anyone who needs it. Happy to reopen if there's interest in merging.

PR #942, opened 2026-04-16
fix(hooks): resolve Claude plugin hook runner cross-platform

### LINKED ISSUE #753: Windows: python3 command not found in hook scripts
## Bug Description

MemPalace hook scripts (`mempal-stop-hook.sh`, `mempal-precompact-hook.sh`) fail on Windows because they use `python3` which doesn't exist on Windows. Python is registered as `python` on Windows.

## Error Message

```
/c/Users/USER/.claude/plugins/cache/mempalace/mempalace/3.0.14/hooks/mempal-stop-hook.sh: line 5: python3: command not found
```

## Affected Files

- `hooks/mempal-stop-hook.sh` (line 5)
- `hooks/mempal-precompact-hook.sh` (line 5)

## Current Code

```bash
echo "$INPUT" | python3 -m mempalace hook run --hook stop --harness claude-code
```

## Suggested Fix

Use a cross-platform Python command resolution. For example:

```bash
PYTHON_CMD="${PYTHON_CMD:-$(command -v python3 2>/dev/null || command -v python 2>/dev/null)}"
echo "$INPUT" | "$PYTHON_CMD" -m mempalace hook run --hook stop --harness claude-code
```

Or simply use `python` which works on both platforms (Python 3 is now the default `python` on most systems).

## Environment

- OS: Windows 11 Pro
- Shell: Git Bash (bash)
- Python: 3.x (available as `python`, not `python3`)
- MemPalace version: 3.0.14
- Harness: Claude Code

fatkobra:
## What does this PR do?
Closes #753

Summary
- update `.claude-plugin/hooks/mempal-stop-hook.sh`
- update `.claude-plugin/hooks/mempal-precompact-hook.sh`
- prefer the installed `mempalace` CLI
- fall back to importable `python3` / `python`
- add execution tests for the wrapper scripts

Scope
This PR intentionally fixes the Claude plugin shell hook wrappers only.
It does not change `.claude-plugin/plugin.json` or `.claude-plugin/.mcp.json`,
which still need a separate supported mechanism for MCP command resolution.

## How to test
Manual verification
- before: both wrappers failed with `python3: command not found`
- after: both wrappers returned `{}` with exit code 0 in a PATH containing `python` but no `python3`

Validation
- `bash -n .claude-plugin/hooks/mempal-stop-hook.sh`
- `bash -n .claude-plugin/hooks/mempal-precompact-hook.sh`
- `ruff format tests/test_claude_plugin_hook_wrappers.py`
- `ruff check tests/test_claude_plugin_hook_wrappers.py`
- `ruff format --check tests/test_claude_plugin_hook_wrappers.py`
- `python -m pytest tests/test_claude_plugin_hook_wrappers.py -v`
- `python -m pytest tests/test_hooks_cli.py tests/test_claude_plugin_hook_wrappers.py -v`

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/.claude-plugin/hooks/mempal-precompact-hook.sh
+++ b/.claude-plugin/hooks/mempal-precompact-hook.sh
@@ -1,5 +1,24 @@
 #!/bin/bash
 # MemPalace PreCompact Hook — thin wrapper calling Python CLI
 # All logic lives in mempalace.hooks_cli for cross-harness extensibility
-INPUT=$(cat)
-echo "$INPUT" | python3 -m mempalace hook run --hook precompact --harness claude-code
+run_mempalace_hook() {
+  if command -v mempalace >/dev/null 2>&1; then
+    mempalace hook run "$@"
+    return $?
+  fi
+
+  if command -v python3 >/dev/null 2>&1 && python3 -c "import mempalace" >/dev/null 2>&1; then
+    python3 -m mempalace hook run "$@"
+    return $?
+  fi
+
+  if command -v python >/dev/null 2>&1 && python -c "import mempalace" >/dev/null 2>&1; then
+    python -m mempalace hook run "$@"
+    return $?
+  fi
+
+  echo "MemPalace hook error: could not find a runnable mempalace command or module" >&2
+  return 1
+}
+
+run_mempalace_hook --hook precompact --harness claude-code
--- a/.claude-plugin/hooks/mempal-stop-hook.sh
+++ b/.claude-plugin/hooks/mempal-stop-hook.sh
@@ -1,5 +1,24 @@
 #!/bin/bash
 # MemPalace Stop Hook — thin wrapper calling Python CLI
 # All logic lives in mempalace.hooks_cli for cross-harness extensibility
-INPUT=$(cat)
-echo "$INPUT" | python3 -m mempalace hook run --hook stop --harness claude-code
+run_mempalace_hook() {
+  if command -v mempalace >/dev/null 2>&1; then
+    mempalace hook run "$@"
+    return $?
+  fi
+
+  if command -v python3 >/dev/null 2>&1 && python3 -c "import mempalace" >/dev/null 2>&1; then
+    python3 -m mempalace hook run "$@"
+    return $?
+  fi
+
+  if command -v python >/dev/null 2>&1 && python -c "import mempalace" >/dev/null 2>&1; then
+    python -m mempalace hook run "$@"
+    return $?
+  fi
+
+  echo "MemPalace hook error: could not find a runnable mempalace command or module" >&2
+  return 1
+}
+
+run_mempalace_hook --hook stop --harness claude-code
--- a/tests/test_claude_plugin_hook_wrappers.py
+++ b/tests/test_claude_plugin_hook_wrappers.py
@@ -1,6 +1,7 @@
 """Execution tests for Claude plugin hook wrapper scripts."""
 
 import os
+import shutil
 import subprocess
 from pathlib import Path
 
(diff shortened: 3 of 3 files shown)


mvalentsev:
~#833 by igorls already covers this (same issue #753) -- real python resolution with MEMPAL_PYTHON override, all CI green.~

~Also #670, #325, #384, #410 target the same problem.~

fatkobra:
Thanks for the pointer @mvalentsev .

I checked #833, and it looks complementary rather than identical:
- #833 updates the legacy shell hooks in `hooks/mempal_save_hook.sh` and `hooks/mempal_precompact_hook.sh`
- this PR updates the Claude plugin wrapper scripts in `.claude-plugin/hooks/mempal-stop-hook.sh` and `.claude-plugin/hooks/mempal-precompact-hook.sh`

Issue #753’s current code example is the wrapper path (`python3 -m mempalace hook run --hook stop --harness claude-code`), which is what this PR fixes.

If maintainers would prefer a single combined solution, I’m happy to close this and rework it on top of #833, but as written it looks like #942 fixes the plugin wrapper layer while #833 fixes the legacy save/precompact shell hooks.

mvalentsev:
You're right, my mistake -- #833 targets hooks/ (legacy scripts) while this PR targets .claude-plugin/hooks/ (plugin wrappers). Different files, complementary fixes. Thanks for checking.

mvalentsev:
btw, `.codex-plugin/hooks/mempal-hook.sh` has the same hardcoded `python3` on line 6. Might be worth including in this PR since it's the same class of fix.

mvalentsev:
Also worth noting: the plugin JSON configs (.claude-plugin/plugin.json, .mcp.json, .codex-plugin/plugin.json) also hardcode `python3` as the MCP server command. Those need an entry point fix rather than shell fallbacks -- #805 and #340 both address that side of things.

fatkobra:
Thanks — agreed on both points.

`.codex-plugin/hooks/mempal-hook.sh` does look like the same wrapper-layer issue, so I’m happy to add that here if maintainers want this PR to cover all plugin hook wrappers.

However, I think it is better to keep the MCP config files (`.claude-plugin/plugin.json`, `.claude-plugin/.mcp.json`, `.codex-plugin/plugin.json`) out of scope for this PR, since those seem to need the separate entry-point/config approach already being discussed in #805 / #340.

fatkobra:
I have a tested Codex-wrapper follow-up patch ready locally as well `(.codex-plugin/hooks/mempal-hook.sh)` if maintainers @igorls  would prefer this PR to cover all plugin hook wrappers.

Qodo-Free-For-OSS:
Hi, The wrapper execution tests pass a Windows-formatted path (backslashes/drive letters) directly to `bash`, which commonly fails under Git Bash/MSYS path semantics. This can break the Windows CI job whenever `bash` exists on PATH (tests are only skipped when bash is missing).

**How to fix:** Use POSIX path for bash

> ### Issue description
> `tests/test_claude_plugin_hook_wrappers.py` invokes `bash` with a script path built via `str(Path(...))`. On Windows this produces backslashes, which Git Bash/MSYS often cannot resolve.
> 
> ### Issue Context
> The tests are skipped only when `bash` is missing. Windows CI (`windows-latest`) frequently has Git Bash available, so these tests will run and can fail.
> 
> ### Fix Focus Areas
> - tests/test_claude_plugin_hook_wrappers.py[25-69]
>   - Convert the script path passed to `bash` using the existing `_shell_path()` helper (or otherwise normalize to POSIX/MSYS-acceptable form).

---

fatkobra:
Thanks — fixed. `_run_hook()` now passes the wrapper script path to `bash` through `_shell_path(...)` instead of `str(Path(...))`, so the tests use POSIX-style paths on Windows/Git Bash too.

Happy to look at the other findings too if you can share the specific issues or exact lines involved. I’d prefer to keep follow-up changes concrete and in-scope for this PR. @qodo-ai-reviewer

PR #948, opened 2026-04-16
fix: harden KnowledgeGraph and MCP server for multi-process access

felipecpaiva:
## Summary

When mempalace runs behind `mcp-proxy` (SSE mode) with multiple concurrent clients, two classes of concurrency bugs surface:

1. **SQLite "database is locked"** in KnowledgeGraph — `busy_timeout` was only 10s with no application-level retry. Concurrent writers from separate mcp-proxy processes exhaust the timeout.

2. **ChromaDB client cache thrashing** — every write changes `chroma.sqlite3` mtime. Other processes detect this and recreate `PersistentClient`, reloading the full HNSW index from disk on every operation.

### KnowledgeGraph fixes
- `busy_timeout` 10s → 60s
- `_sqlite_retry` decorator: exponential backoff with jitter (5 retries, only for "locked"/"busy" errors)
- `BEGIN IMMEDIATE` for write transactions (detect contention at start, not mid-transaction)
- `PRAGMA wal_autocheckpoint=1000` + `journal_size_limit=64MB` (manage WAL growth)
- `atexit.register(self.close)` for clean WAL checkpointing on shutdown

### MCP server fixes
- Rate-limit `chroma.sqlite3` stat/mtime checks to 5-second intervals
- `_refresh_db_mtime()` after writes prevents self-inflicted client recreation
- Bypass cooldown for safety-critical DB disappearance detection (rebuild scenarios)
- `tool_reconnect` now fully clears client + mtime state

### Tests
- `TestSQLiteRetryDecorator` — 5 cases: retry success, exhaustion, non-lock errors, busy variant
- `TestConnectionPragmas` — 3 cases: autocheckpoint, journal_size_limit, WAL mode
- `TestMultiProcessLocking` — 4 processes × 20 triples to same DB file, zero failures

## Test plan
- [x] `python -m pytest tests/ -v --ignore=tests/benchmarks` — 958 passed
- [x] `ruff check` + `ruff format --check` — clean
- [ ] Manual: connect 2+ Claude Code sessions via SSE, trigger concurrent kg_add calls
- [ ] Manual: concurrent add_drawer calls from multiple clients — no cache thrashing


Qodo-Free-For-OSS:
Hi, KnowledgeGraph.__init__ registers a bound method with atexit, which holds a strong reference to the instance and prevents it from being garbage-collected; repeated short-lived KnowledgeGraph creations will accumulate open SQLite connections until process exit.

**How to fix:** Avoid atexit holding self

> ### Issue description
> `KnowledgeGraph.__init__` registers `self.close` with `atexit`, which strongly retains the instance and leaks SQLite connections in any code path that constructs `KnowledgeGraph` without explicitly calling `close()`.
> 
> ### Issue Context
> This is triggered in `fact_checker._check_kg_contradictions()` which creates a new `KnowledgeGraph` per call and does not close it.
> 
> ### Fix Focus Areas
> - mempalace/knowledge_graph.py[109-113]
> - mempalace/fact_checker.py[179-217]
> 
> ### Suggested fix approaches
> - Remove the per-instance `atexit.register(self.close)` and rely on explicit `close()` in long-lived owners (MCP server already has a global KG) and/or context management.
> - Alternatively, register an atexit handler that does not capture `self` strongly (e.g., store weakref to the instance and close if still alive).
> - Ensure `fact_checker` closes the KG (e.g., `try/finally: kg.close()`), which is the most direct leak stopper regardless of atexit strategy.

We noticed a couple of other issues in this PR as well — happy to share if helpful.

---

felipecpaiva:
Thanks for flagging this — good catch. Addressed in 2feaa71.

**What changed**

- `knowledge_graph.py`: removed `atexit.register(self.close)` from `__init__` (and the now-unused `import atexit`). The bound-method registration was keeping a strong reference to every instance in `atexit`'s registry, so short-lived KGs could never be GC'd.
- `fact_checker.py::_check_kg_contradictions`: wraps the query loop in `try/finally: kg.close()` so the per-call KG always releases its SQLite connection, regardless of atexit strategy.
- `mcp_server.py`: moved the atexit registration to module scope on the `_kg` singleton (`atexit.register(_kg.close)`). The clean WAL-checkpoint-on-shutdown benefit is preserved for the long-lived owner, and there's no retention issue because the module already holds `_kg` for the process lifetime.
- Regression test in `tests/test_knowledge_graph.py::TestGarbageCollection`: create a KG, take a `weakref`, close + `del` + `gc.collect()`, assert the weakref resolves to `None`. This is the exact property that was broken before.

**Why this shape**

You listed three options:
1. Remove the per-instance atexit and rely on explicit `close()` in long-lived owners.
2. Use a `weakref`-based atexit handler.
3. Explicit `close()` in `fact_checker`.

I took **(1) + (3)**. The weakref variant (2) would work, but option (1) is simpler and makes ownership explicit: the MCP server owns the process-wide singleton, short-lived constructors are responsible for their own cleanup. A grep for `KnowledgeGraph(` confirmed `fact_checker._check_kg_contradictions` is the only production short-lived caller; tests already call `kg.close()` explicitly.

Suite is green (`pytest tests/ --ignore=tests/benchmarks` → 959 passed), lint/format clean.

Happy to hear the other issues you noticed — please share.

felipecpaiva:
Hi maintainers — both workflow runs on this PR are sitting in `action_required` state (first-time-contributor gate on Actions), so `Tests` hasn't actually executed yet. Could one of you click "Approve and run workflows" so reviewers have a green check to anchor on?

Locally on `2feaa71`:
- `pytest tests/ --ignore=tests/benchmarks` → 959 passed
- `ruff check` + `ruff format --check` → clean

Happy to rebase onto latest `develop` if preferred (`mergeable: true` as of now; PR #1014's schema columns don't overlap our modified regions).

felipecpaiva:
cc @bensig @igorls @milla-jovovich — friendly ping on the CI approval in the comment above. The `Tests` workflow is blocked on the first-time-contributor gate, so there's no green check for you to anchor on yet. Tests pass locally (959/959, ruff clean). Happy to address anything you'd want adjusted before this can move.

bensig:
@felipecpaiva thx for this - small lint issue `tests/test_knowledge_graph.py`

felipecpaiva:
@bensig Yeah i missed that due to  a ruff version difference.
Fixed and pushed.

felipecpaiva:
Also @bensig had a small conflict fixed on the branch.

felipecpaiva:
cc @bensig @igorls @milla-jovovich — friendly ping on the CI approval in the comment above. The Tests workflow is blocked on the first-time-contributor gate, so there's no green check for you to anchor on yet. 
Tests pass locally (959/959, ruff clean). 
Happy to address anything you'd want adjusted before this can move.

felipecpaiva:
Hi @bensig @igorls — just pushed a fresh merge with current `develop` (commit `0fb8a3f`), conflicts resolved across `knowledge_graph.py` and `mcp_server.py`. PR is now `MERGEABLE` / `CLEAN`.

Since I'm still a first-time contributor, the `Tests` workflow is sitting in `action_required` — could one of you click **Approve and run workflows** when you have a moment so CI can run against the new merge? Happy to address anything that surfaces. Thanks!

igorls:
CI status summary after latest run (post-0fb8a3f):
- `ruff format --check .` fails in `tests/test_knowledge_graph.py`.
- `tests/test_mcp_server.py::TestStructuredErrors::test_get_client_rearms_quarantine_on_reconnect` is failing on Linux/macOS/Windows:
  `AssertionError: _get_client should call _prepare_palace_for_open on reconnect`.

I’d hold merge until lint + this reconnect behavior assertion are fixed.

igorls:

felipecpaiva:
@igorls thanks for the nudge, and no apology needed. Rebased onto current `develop` — force-pushed, so the branch is now three commits on top of `develop` and merges cleanly as a fast-forward.

Two things changed beyond a plain rebase, and I want to flag both rather than have you find them in review.

**1. I dropped the `_get_client` mtime cooldown, and that is what was failing CI.**

The original PR added a 5-second cooldown (`_MTIME_CHECK_INTERVAL`) plus `_refresh_db_mtime()` calls in the write tools. Both are now actively wrong, because #2002 landed in `_get_client` in the meantime. Your own comment there says it:

> a peer process changed chroma.sqlite3 on disk. chromadb caches its System (and the live HNSW segment) keyed by path, so make_client() below would hand back the STALE segment, which then persists its outdated index over the peer's writes, driving the persisted count backwards.

The cooldown suppressed exactly that reconnect for up to 5 seconds, and `_refresh_db_mtime` stat'd the file *after* our own write, so a peer write landing in that window got stamped as ours and was never picked up. Both were performance optimisations; neither belongs in a PR about multi-process safety.

This is also the root cause of `test_get_client_rearms_quarantine_on_reconnect` failing on all three platforms in your 2026-06-10 run: the cooldown's early `return _client_cache` sat above the reconnect path, so `_prepare_palace_for_open` was never called. The test is asserting correct behaviour, so I removed our code rather than touching the test. It passes unmodified now.

**2. `BEGIN IMMEDIATE` is now a `_write_txn()` context manager**, covering all five write paths including `supersede` and `apply_row`, which you added after the PR was opened.

Reviewing that manager turned up a third commit's worth of bug in my own code, so it is in here too: `COMMIT` was outside the `try`. `COMMIT` can raise in its own right — SQLITE_BUSY during a WAL checkpoint — and a raising commit leaves the transaction open. The next `_sqlite_retry` attempt then hits `cannot start a transaction within a transaction`, which is not a lock error, so the retry gives up and reports *that* instead of the contention it exists to absorb. `rollback()` is a no-op when nothing is open, so moving `COMMIT` inside the `try` is safe on every path. Guarded by `test_failed_commit_does_not_leave_the_transaction_open`, which asserts on `in_transaction` rather than on a message so it survives a reworded error.

Worth recording that the obvious shortcut is wrong: `isolation_level = "IMMEDIATE"` does *not* work, because sqlite3 only opens a transaction on the first DML statement. In `invalidate` — which SELECTs `valid_from` to reject an inverted interval, then UPDATEs — the BEGIN still landed after the SELECT:

```
0 "SELECT id, valid_from FROM triples WHERE subject='max' ..."
1 'BEGIN IMMEDIATE'
2 "UPDATE triples SET valid_to='2026-02-15' ..."
```

so the validating read was outside the write lock and a peer could open a fact between the check and the UPDATE, producing `valid_to < valid_from`. The explicit BEGIN closes it. `add_triple` and `supersede` are write-first and were already atomic; `invalidate` was not.

**Also worth knowing:** the original marked `test_concurrent_process_writes` `@pytest.mark.slow`, and `addopts` excludes `slow` — so the test for the headline fix never actually ran in CI. Marker removed (it takes 0.27s), and it now uses `multiprocessing.get_context("spawn")` to match the convention in `test_closets.py`, so macOS and Windows exercise the same path your failures came from.

**Review tip:** read the diff with `git diff -w`. It is +583/-208 raw but +407/-32 ignoring whitespace — `fact_checker.py` is 4 real lines, the rest is reindentation from moving a loop under a `with` and five write blocks losing a nesting level.

**Verification, all local:** `ruff check` and `ruff format --check` clean on 0.16.1 to match the lint job's pin; full suite 4877 passed / 0 failed / 53 skipped; coverage 81.83% against the 80% gate. I also mutation-tested each new guard — removing the explicit `BEGIN IMMEDIATE` fails exactly `test_writes_open_an_immediate_transaction` and `test_validating_read_happens_inside_the_write_lock`, and moving `COMMIT` back outside the `try` fails exactly `test_failed_commit_does_not_leave_the_transaction_open`.

Two honest gaps: I only ran Linux / Python 3.14, so the other five matrix legs are on CI, and `test_concurrent_process_writes` passes even on unhardened code — 80 small WAL writes are not enough contention. It is a smoke test against gross regressions like a cross-process deadlock, not a guard for this fix; the three statement-order and pragma tests are the real guards.

**One thing I cannot do myself:** both workflow runs on this push came back `action_required` — the first-time-contributor gate again, same as in April. So there is no green check on the PR for you to anchor on until someone approves the run. `git merge-base --is-ancestor develop HEAD` passes and GitHub reports the PR `MERGEABLE`, so the "merges cleanly" half of your condition is met; the CI half needs one click. Whenever you get a moment.

Happy to split any of this out if you would rather review the transaction change separately from the connection-pragma change.

mempalace/knowledge_graph.py
```diff
@@ -208,8 +260,23 @@ def _migrate_schema(self, conn):
 
     def _conn(self):
         if self._connection is None:
-            self._connection = sqlite3.connect(self.db_path, timeout=10, check_same_thread=False)
+            # timeout= sets SQLite's busy_timeout: how long a writer waits for
+            # a peer process to release the write lock before raising
+            # "database is locked". 10s is not enough under a concurrent CLI
+            # mine; 60s absorbs a long WAL checkpoint.
+            self._connection = sqlite3.connect(self.db_path, timeout=60, check_same_thread=False)
```
igorls, line 267: [P1] Do not cache the connection until initialization succeeds

`_connection` becomes non-null before the WAL pragmas and `row_factory` assignment. If the initial WAL pragma raises a lock error, the new `_init_db` retry calls `_conn()` again and immediately returns this partially configured connection. Schema migration then indexes tuple rows as dictionaries and raises the exact CI TypeError. Reproduced with a real connection subclass failing only its first WAL pragma. Initialize a local connection completely before assigning this field, or close/reset it on failure, and pin the retry with a deterministic regression.

igorls (changes requested):
Requesting changes on initialization retry. I reproduced a specific path to the same tuple-index TypeError seen in CI.

A real `sqlite3.Connection` subclass that raises `OperationalError("database is locked")` only on its first `PRAGMA journal_mode=WAL` is enough: the new `_init_db` retry reuses the published connection, skips the remaining setup, and `_migrate_schema` fails at `row["name"]` because `row_factory` was never assigned.

Please fully configure a local connection before publishing it as `_connection`, or close/reset it if setup fails, and add that deterministic first-WAL-pragma-failure test. The full knowledge-graph module passed 61 tests in my WSL run, but the injected contention case reliably reproduces the failure; the concurrency test alone does not reliably exercise it.

Issue #955, opened 2026-04-16
Compaction Blocking


joshdeleenheer:
# Issue: PreCompact hook creates unrecoverable deadlock at context exhaustion

## Title (for GitHub)

`PreCompact hook hard-blocks /compact at context exhaustion, creating an unrecoverable deadlock`

## Summary

`hook_precompact` in `mempalace/hooks_cli.py` unconditionally emits `{"decision": "block", ...}` every time `/compact` is attempted, demanding the assistant save context to MemPalace before compaction proceeds. When the session has already reached context exhaustion — which is the exact situation where `/compact` is most needed — the assistant has no remaining context budget to perform the save. The hook then blocks indefinitely, and the only escape is to manually neutralize the hook scripts on disk from a separate process.

This has now happened to me twice. Both times required SSHing into the filesystem from a separate shell to `mv` the precompact hook script to a `.bak` and replace it with an `exit 0` stub so the stuck sessions could compact. Once compaction completed, I restored the originals.

## Reproduction

1. Run a long `claude` session until context usage approaches the limit (e.g., a multi-phase coordinator run dispatching many sub-agents).
2. At some point the assistant returns `Context limit reached · /compact or /clear to continue`.
3. Run `/compact`.
4. PreCompact hook returns `{"decision": "block", "reason": PRECOMPACT_BLOCK_REASON}`.
5. Assistant attempts to respond with a save action — also blocked by context exhaustion (`Context limit reached`).
6. Session is now unrecoverable through any in-session command. `/clear` works but discards everything; `/compact` is permanently blocked.

## Root cause

`mempalace/hooks_cli.py:189-212`:

```python
def hook_precompact(data: dict, harness: str):
    """Precompact hook: always block with comprehensive save instruction."""
    ...
    # Always block -- compaction = save everything
    _output({"decision": "block", "reason": PRECOMPACT_BLOCK_REASON})
```

The function has no conditional path. It blocks 100% of the time regardless of whether the session has the context budget to act on the block directive.

Notably, the same function **already performs auto-ingest via `mempalace mine`** (lines 196–209) when `MEMPAL_DIR` is set. So when the auto-ingest path is configured, the structured-save block becomes belt-and-suspenders — and the suspenders turn into a hangman's noose when context is exhausted.

## Impact

[...]

Issue #965, opened 2026-04-17
Critical Data write loop, HNSW compaction bug


raizenica:
┌───────────┬──────────────┬─────────────┐
  │  Metric   │    Before    │    After    │
  ├───────────┼──────────────┼─────────────┤
  │ Used      │ 1.7 T (100%) │ 211 G (13%) │
  ├───────────┼──────────────┼─────────────┤
  │ Available │ 13 G         │ 1.5 T       │
  └───────────┴──────────────┴─────────────┘

  Root cause: my mempalace mine on the auto-memory dir hit a chromadb 1.x HNSW compaction bug that wrote a 10.2 TB sparse link_lists.bin (≈1.5 TB on disk) before crashing. The
  crash didn't release the file handle, and the MCP server processes pinned the deleted inodes. Kill of PIDs 1581835 / 1581850 freed everything.

**Environment:**
- OS: Nobara
- MemPal version: 3.1.0

PR #346, opened 2026-04-09
fix: prevent HNSW index bloat from resize+persist cycles

yoshi280:
## Summary

- Set `hnsw:batch_size` and `hnsw:sync_threshold` to 50,000 on ChromaDB collection creation to defer index persistence until large batches complete
- Batch per-file drawer upserts in `process_file()` instead of inserting one drawer at a time via `add_drawer()`
- Apply the same HNSW metadata in `mcp_server.py` for collections created via the MCP server

Fixes #344

## Problem

Mining ~56,000 drawers (from a GSD project database export) caused `link_lists.bin` to grow to **441GB apparent / 168GB on disk** instead of the expected ~8MB. This caused two sequential system crashes from disk/memory exhaustion before the problem was identified. The corrupted index made `mempalace status` segfault (exit 139) and rendered the palace unusable.

**Root cause**: With default HNSW params (`batch_size=100`, `sync_threshold=1000`, initial capacity 1000), inserting 56K drawers one at a time triggers ~30 index resizes and ~560 `persistDirty()` calls. The `persistDirty()` function uses relative seek positioning in `link_lists.bin`, and after many resize cycles the accumulated seek positions drift, causing the OS to extend the sparse file with zero-filled regions. Each cycle compounds the problem.

**Numbers**:
- 56,387 drawers from 2,395 GSD project files
- `link_lists.bin`: 473,786,104,508 bytes apparent (441GB), 168GB actual disk
- `data_level0.bin`: 64MB (correctly sized)
- Expected `link_lists.bin` for 56K vectors at M=16: ~8.4MB
- Ratio: 56,083x expected size

## Changes

**`miner.py`**:
- Added `_HNSW_METADATA` constant with tuned batch/sync thresholds
- `get_collection()` passes metadata on `create_collection()`
- `process_file()` now collects all drawer IDs, documents, and metadata into lists and calls `collection.upsert()` once per file instead of once per chunk

**`mcp_server.py`**:
- `_get_collection(create=True)` passes the same HNSW metadata to `get_or_create_collection()`

## Test plan

- [ ] Mine a project with <1K drawers -- verify behavior unchanged
- [ ] Mine a project with 10K+ drawers -- verify `link_lists.bin` stays under 1MB
- [ ] Mine a project with 50K+ drawers -- verify no bloat (was 441GB, should be <50MB)
- [ ] Verify `mempalace status` and `mempalace search` work after large mines
- [ ] Verify MCP server creates collections with correct metadata

PR #976, opened 2026-04-17
fix: HNSW graph corruption, PreCompact deadlock, mine fan-out (closes #974, #965, #955)

### LINKED ISSUE #974: SIGSEGV in HNSW parallel inserts — missing hnsw:num_threads on collection creation
**What happened?**

The MCP server crashes with a SIGSEGV (segmentation fault 11) during HNSW index operations. The crash occurs in `repairConnectionsForUpdate` / `addPoint` when ChromaDB's default multi-threaded `ParallelFor` triggers a race condition in the HNSW graph mutation path. This reproduces on both chromadb <1.0 (`hnswlib.cpython-*-darwin.so`) and chromadb 1.5.x (`chromadb_rust_bindings.abi3.so`).
The root cause is `mempalace/backends/chroma.py` not setting `hnsw:num_threads` on `get_or_create_collection`, so ChromaDB defaults to multi-threaded parallel inserts.

Related: #965

**What did you expect?**

The MCP server should not crash during normal add_drawer / auto-mining operations.
**How to reproduce:**
1. Start the mempalace MCP server
2. Trigger auto-mining on a workspace, or issue multiple `add_drawer` calls
3. Server segfaults intermittently during HNSW index insertion

**Environment:**

- OS: macOS 26.3.1 (ARM64, M4 Pro)
- Python version: 3.9.6 and 3.14
- MemPal version: 3.3.0 (chromadb 1.5.8, also reproduced with 0.6.3)

**Suggested fix:**

```diff
# mempalace/backends/chroma.py
             collection = client.get_or_create_collection(
-                collection_name, metadata={"hnsw:space": "cosine"}
+                collection_name, metadata={"hnsw:space": "cosine", "hnsw:num_threads": 1}
             )
```

hnsw:num_threads only affects the batch insert path. Mempalace inserts drawers individually so there is zero performance impact. Queries are unaffected.

**Update:** The fix needs to cover three code paths:

1. **New collections** (`get_or_create_collection` / `create_collection`) — add `"hnsw:num_threads": 1` to metadata
2. **Existing collections** (`get_collection`) — call `collection.modify(metadata={"hnsw:space": "cosine", "hnsw:num_threads": 1})` so palaces created before the fix also get single-threaded inserts
3. **v3.1.0 `palace.py`** — `create_collection` has no metadata at all (no `hnsw:space` either)

Without item 2, any palace created before the fix will keep crashing until the user deletes and recreates their palace.

felipetruman:
## Summary

Three related problems — same failure mode (runaway HNSW writes + deadlocked session), different trigger paths. Observed in the wild this week on a Nobara install that wrote ~1.5 TB of sparse `link_lists.bin` before crashing and on a macOS M4 Pro install that segfaulted intermittently during `add_drawer`.

1. **`backends/chroma.py`** — pin `hnsw:num_threads: 1` on collection creation. Disables ChromaDB's multi-threaded `ParallelFor` insert path, which has a race in `repairConnectionsForUpdate` / `addPoint` that is the root cause of both #974 (SIGSEGV) and #965 (sparse link_lists.bin blowup). MemPalace ingests drawers one at a time anyway, so this path was never a throughput win — only a latent correctness hazard.
2. **`palace.py` + `miner.py`** — add `mine_global_lock()`, a non-blocking process-wide lock wrapping `miner.mine`. The existing `mine_lock` is per source-file (sha256 of path); it does not prevent N copies of `mempalace mine <dir>` from running concurrently, which is the fan-out pattern that feeds problem (1). The global lock uses `fcntl.LOCK_EX | LOCK_NB` on Unix and `msvcrt.LK_NBLCK` on Windows — if the lock is held, raise `MineAlreadyRunning`, print a short notice, exit cleanly. Concurrent spawns collapse to a single runner instead of queuing as waiters.
3. **`hooks_cli.py`** — cap `hook_precompact` block attempts per session (`MAX_PRECOMPACT_BLOCK_ATTEMPTS = 2`, tracked in `STATE_DIR/{session_id}_precompact_attempts`). Unconditional blocking creates the #955 deadlock at context exhaustion: the hook demands a save the assistant has no budget to perform. After the cap the hook passes through (`_output({})`) so `/compact` can make progress. Per-session state means a fresh session gets a fresh budget.

## Closes

- #974 — SIGSEGV in HNSW parallel inserts
- #965 — Critical data write loop, HNSW compaction bug
- #955 — Compaction blocking

## Test plan

- [x] `python3 -c \"import ast; ast.parse(open(f).read())\"` on all four modified files
- [x] Import smoke test: `from mempalace.palace import mine_global_lock, MineAlreadyRunning` + `from mempalace.hooks_cli import MAX_PRECOMPACT_BLOCK_ATTEMPTS`
- [x] Functional test of `mine_global_lock`:
  - first acquire succeeds
  - second concurrent acquire raises `MineAlreadyRunning`
  - lock is reusable after release
- [x] `pytest tests/test_backends.py tests/test_cli.py` → **49 passed, 0 failed**
- [ ] Would welcome reviewers running the full suite + a targeted test for `hook_precompact` attempt capping and a concurrency test that spawns two `mempalace mine` subprocesses against the same palace (happy to add if desired).

## Notes

- `dry_run` bypasses `mine_global_lock` (no writes, no point serializing).
- `MineAlreadyRunning` inherits from `RuntimeError` so it is a caught error, not a typing surprise.
- The existing per-file `mine_lock` is preserved — it still guards delete+insert interleave *within* a single mine run.
- Commit is a single logical change spanning four files; splitting would obscure the fact that the three fixes are interlocked (without the `hnsw:num_threads=1` fix, even a single mine can corrupt the graph; without the global lock, the correctness fix is leaky under hook fan-out).

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -153,9 +153,7 @@ def _pin_hnsw_threads(collection) -> None:
         return
     try:
         collection.modify(
-            configuration=UpdateCollectionConfiguration(
-                hnsw=UpdateHNSWConfiguration(num_threads=1)
-            )
+            configuration=UpdateCollectionConfiguration(hnsw=UpdateHNSWConfiguration(num_threads=1))
         )
     except Exception:
         logger.debug("_pin_hnsw_threads modify failed", exc_info=True)
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -566,7 +566,9 @@ class ChromaBackend(BaseBackend):
 
         if create:
             collection = client.get_or_create_collection(
-                collection_name, metadata={"hnsw:space": hnsw_space}, **ef_kwargs
+                collection_name,
+                metadata={"hnsw:space": hnsw_space, "hnsw:num_threads": 1},
+                **ef_kwargs,
             )
         else:
             collection = client.get_collection(collection_name, **ef_kwargs)
@@ -613,7 +615,9 @@ class ChromaBackend(BaseBackend):
         ef = self._resolve_embedding_function()
         ef_kwargs = {"embedding_function": ef} if ef is not None else {}
         collection = self._client(palace_path).create_collection(
-            collection_name, metadata={"hnsw:space": hnsw_space}, **ef_kwargs
+            collection_name,
+            metadata={"hnsw:space": hnsw_space, "hnsw:num_threads": 1},
+            **ef_kwargs,
         )
         return ChromaCollection(collection)
 
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -130,6 +130,37 @@ def quarantine_stale_hnsw(palace_path: str, stale_seconds: float = 3600.0) -> li
     return moved
 
 
+def _pin_hnsw_threads(collection) -> None:
+    """Best-effort retrofit: pin ``hnsw:num_threads=1`` on an existing collection.
+
+    Fresh collections set this via ``metadata=`` at creation. Legacy palaces
+    built before that change keep the default (parallel insert) and can hit
+    the HNSW race described in #974/#965. ChromaDB's
+    ``collection.modify(configuration=...)`` lets us re-apply ``num_threads=1``
+    in memory at load time so every new process is protected.
+
+    Note: in chromadb 1.5.x the modified ``configuration_json["hnsw"]`` does
+    not persist to disk across ``PersistentClient`` reopens, so this must
+    run on every ``get_collection`` call, not just once.
+    """
+    try:
+        from chromadb.api.collection_configuration import (
+            UpdateCollectionConfiguration,
+            UpdateHNSWConfiguration,
+        )
+    except ImportError:
+        logger.debug("_pin_hnsw_threads skipped: chromadb too old", exc_info=True)
+        return
+    try:
+        collection.modify(
+            configuration=UpdateCollectionConfiguration(
+                hnsw=UpdateHNSWConfiguration(num_threads=1)
+            )
+        )
+    except Exception:
+        logger.debug("_pin_hnsw_threads modify failed", exc_info=True)
+
+
 def _fix_blob_seq_ids(palace_path: str) -> None:
     """Fix ChromaDB 0.6.x -> 1.5.x migration bug: BLOB seq_ids -> INTEGER.
 
@@ -572,6 +603,7 @@ class ChromaBackend(BaseBackend):
             )
         else:
             collection = client.get_collection(collection_name, **ef_kwargs)
+        _pin_hnsw_threads(collection)
         return ChromaCollection(collection)
 
     def close_palace(self, palace) -> None:
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -643,6 +643,9 @@ def hook_session_start(data: dict, harness: str):
     _output({})
 
 
+MAX_PRECOMPACT_BLOCK_ATTEMPTS = 2
+
+
 def hook_precompact(data: dict, harness: str):
     """Precompact hook: mine transcript synchronously, then allow compaction."""
     parsed = _parse_harness_input(data, harness)
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -643,9 +643,6 @@ def hook_session_start(data: dict, harness: str):
     _output({})
 
 
-MAX_PRECOMPACT_BLOCK_ATTEMPTS = 2
-
-
 def hook_precompact(data: dict, harness: str):
     """Precompact hook: mine transcript synchronously, then allow compaction."""
     parsed = _parse_harness_input(data, harness)
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -217,9 +217,16 @@ def _get_collection(create=False):
     try:
         client = _get_client()
         if create:
+            # hnsw:num_threads=1 disables ChromaDB's multi-threaded ParallelFor
+            # HNSW insert path, which has a race in repairConnectionsForUpdate /
+            # addPoint (see issues #974, #965). The setting is only honored at
+            # collection creation time — pre-existing palaces created before
+            # this fix keep the unsafe default; users must `mempalace nuke` +
+            # re-mine to get the protection on legacy palaces.
             _collection_cache = ChromaCollection(
                 client.get_or_create_collection(
-                    _config.collection_name, metadata={"hnsw:space": "cosine"}
+                    _config.collection_name,
+                    metadata={"hnsw:space": "cosine", "hnsw:num_threads": 1},
                 )
             )
             _metadata_cache = None
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -57,7 +57,7 @@ from .config import (  # noqa: E402
     sanitize_content,
 )
 from .version import __version__  # noqa: E402
-from .backends.chroma import ChromaBackend, ChromaCollection  # noqa: E402
+from .backends.chroma import ChromaBackend, ChromaCollection, _pin_hnsw_threads  # noqa: E402
 from .query_sanitizer import sanitize_query  # noqa: E402
 from .searcher import search_memories  # noqa: E402
 from .palace_graph import (  # noqa: E402
@@ -219,20 +219,23 @@ def _get_collection(create=False):
         if create:
             # hnsw:num_threads=1 disables ChromaDB's multi-threaded ParallelFor
             # HNSW insert path, which has a race in repairConnectionsForUpdate /
-            # addPoint (see issues #974, #965). The setting is only honored at
-            # collection creation time — pre-existing palaces created before
-            # this fix keep the unsafe default; users must `mempalace nuke` +
-            # re-mine to get the protection on legacy palaces.
-            _collection_cache = ChromaCollection(
-                client.get_or_create_collection(
-                    _config.collection_name,
-                    metadata={"hnsw:space": "cosine", "hnsw:num_threads": 1},
-                )
+            # addPoint (see issues #974, #965). Set via metadata on fresh
+            # collections and re-applied via _pin_hnsw_threads() for legacy
+            # palaces whose collections were created before this fix (the
+            # runtime config does not persist cross-process in chromadb 1.5.x,
+            # so the retrofit runs every time _get_collection opens a cache).
+            raw = client.get_or_create_collection(
+                _config.collection_name,
+                metadata={"hnsw:space": "cosine", "hnsw:num_threads": 1},
             )
+            _pin_hnsw_threads(raw)
+            _collection_cache = ChromaCollection(raw)
             _metadata_cache = None
             _metadata_cache_time = 0
         elif _collection_cache is None:
-            _collection_cache = ChromaCollection(client.get_collection(_config.collection_name))
+            raw = client.get_collection(_config.collection_name)
+            _pin_hnsw_threads(raw)
+            _collection_cache = ChromaCollection(raw)
             _metadata_cache = None
             _metadata_cache_time = 0
         return _collection_cache
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -1005,6 +1005,7 @@ def mine(
             dry_run=dry_run,
             respect_gitignore=respect_gitignore,
             include_ignored=include_ignored,
+            files=files,
         )
 
     try:
@@ -1018,6 +1019,7 @@ def mine(
                 dry_run=dry_run,
                 respect_gitignore=respect_gitignore,
                 include_ignored=include_ignored,
+                files=files,
             )
     except MineAlreadyRunning:
         print(
@@ -1037,6 +1039,7 @@ def _mine_impl(
     dry_run: bool = False,
     respect_gitignore: bool = True,
     include_ignored: list = None,
+    files: list = None,
 ):
     project_path = Path(project_dir).expanduser().resolve()
     config = load_config(project_dir)
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -20,10 +20,12 @@ from typing import Optional
 from .palace import (
     NORMALIZE_VERSION,
     SKIP_DIRS,
+    MineAlreadyRunning,
     build_closet_lines,
     file_already_mined,
     get_closets_collection,
     get_collection,
+    mine_global_lock,
     mine_lock,
     purge_file_closets,
     upsert_closet_lines,
@@ -993,6 +995,48 @@ def mine(
     ``mine`` walks the tree itself just like before.
     """
 
+    if dry_run:
+        return _mine_impl(
+            project_dir,
+            palace_path,
+            wing_override=wing_override,
+            agent=agent,
+            limit=limit,
+            dry_run=dry_run,
+            respect_gitignore=respect_gitignore,
+            include_ignored=include_ignored,
+        )
+
+    try:
+        with mine_global_lock():
+            return _mine_impl(
+                project_dir,
+                palace_path,
+                wing_override=wing_override,
+                agent=agent,
+                limit=limit,
+                dry_run=dry_run,
+                respect_gitignore=respect_gitignore,
+                include_ignored=include_ignored,
+            )
+    except MineAlreadyRunning:
+        print(
+            "mempalace: another `mine` is already running — exiting cleanly.",
+            file=sys.stderr,
+        )
+        return
+
+
+def _mine_impl(
+    project_dir: str,
+    palace_path: str,
+    wing_override: str = None,
+    agent: str = "mempalace",
+    limit: int = 0,
+    dry_run: bool = False,
+    respect_gitignore: bool = True,
+    include_ignored: list = None,
+):
     project_path = Path(project_dir).expanduser().resolve()
     config = load_config(project_dir)
 
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -25,8 +25,8 @@ from .palace import (
     file_already_mined,
     get_closets_collection,
     get_collection,
-    mine_global_lock,
     mine_lock,
+    mine_palace_lock,
     purge_file_closets,
     upsert_closet_lines,
 )
@@ -1008,7 +1008,7 @@ def mine(
         )
 
     try:
-        with mine_global_lock():
+        with mine_palace_lock(palace_path):
             return _mine_impl(
                 project_dir,
                 palace_path,
@@ -1021,7 +1021,8 @@ def mine(
             )
     except MineAlreadyRunning:
         print(
-            "mempalace: another `mine` is already running — exiting cleanly.",
+            f"mempalace: another `mine` is already running against "
+            f"{palace_path} — exiting cleanly.",
             file=sys.stderr,
         )
         return
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -329,14 +329,21 @@ def mine_palace_lock(palace_path: str):
     *different* palaces can still run in parallel — we only serialize
     writes into the same palace, which is the correctness boundary.
 
+    The key is derived from a fully normalized form of the path:
+    `realpath` resolves symlinks and `..` segments, and `normcase` folds
+    case on Windows (which has a case-insensitive filesystem). Without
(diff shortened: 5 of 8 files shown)


felipetruman:
Thanks for the thorough review — all six comments were actionable. Pushed follow-up commit **4976e71**.

| # | Concern | Resolution |
|---|---|---|
| **#6** | MCP server `_get_collection` bypasses `ChromaBackend`, missing the `hnsw:num_threads=1` fix on the primary crash surface | **Fixed.** `hnsw:num_threads=1` now passed at the `mcp_server.py` `_get_collection` create site as well. |
| **#3** | `mine_global_lock` over-serializes mines across unrelated palaces | **Fixed.** Replaced with `mine_palace_lock(palace_path)` keyed by `sha256(os.path.abspath(palace_path))`. Same palace still serialized; different palaces run in parallel. `mine_global_lock` kept as a back-compat alias. |
| **#1** | `except OSError` around `subprocess.run(timeout=60)` misses `subprocess.TimeoutExpired` | **Fixed.** Now catches `(OSError, subprocess.TimeoutExpired)` so the hook always emits a JSON decision even when mining exceeds 60s. |
| **#2** | No test for the per-session precompact attempt cap | **Fixed.** Added `test_precompact_first_two_attempts_block`, `test_precompact_passes_through_after_cap`, `test_precompact_counter_is_per_session` in `tests/test_hooks_cli.py`. |
| **#4** | No test for `mine_palace_lock` non-blocking / reuse behavior | **Fixed.** New `tests/test_palace_locks.py` covers single-acquire, reuse-after-release, cross-process serialization on same palace, non-interference across different palaces, path normalization, and back-compat alias. |
| **#5** | `get_or_create_collection(metadata=...)` doesn't update existing collections; suggest `collection.modify(metadata=...)` | **Documented as a known limitation; not auto-fixed.** I verified against chromadb 1.5.7: `modify(metadata=...)` **replaces** metadata rather than merging, and re-passing `hnsw:space="cosine"` raises `ValueError: Changing the distance function of a collection once it is created is not supported currently.` The HNSW runtime config (`configuration_json`) also doesn't expose `num_threads` in chromadb 1.5.x, so the flag is read-only post-creation. Rather than paper over this with a `modify` that silently drops `hnsw:space`, I documented in a code comment at the mcp_server create site that pre-existing palaces need a `mempalace nuke` + re-mine to gain the protection. Fresh palaces are always protected. Open to adding a one-shot CLI helper (`mempalace repair-hnsw-threading` or similar) that detects this and prints a warning — happy to take a follow-up issue for that if you'd prefer not to leave it as silent behavior. |

### Testing

- `pytest tests/test_palace_locks.py tests/test_hooks_cli.py tests/test_backends.py tests/test_cli.py` → **98 passed, 0 failed**.
- Runtime validation with two concurrent `mempalace mine` calls:
  - **Different palaces** → both complete in parallel ✓
  - **Same palace** → one completes, the other exits with `mempalace: another \`mine\` is already running against <palace> — exiting cleanly.` ✓

mvalentsev:
Heads-up on the merge conflict. It's specifically in `hooks_cli.hook_precompact`, and the reason is that #863 landed on develop on 2026-04-15 with a different approach to #955.

#863 removed the unconditional block entirely. In current develop, `hook_precompact` runs the transcript mine synchronously and then emits `_output({})`, so there is no `decision: block` left to cap attempts against. Docstring on develop is `"Precompact hook: mine transcript synchronously, then allow compaction."`

After you rebase, the `MAX_PRECOMPACT_BLOCK_ATTEMPTS` hunk becomes unreachable. Two options:

1. Drop that hunk: #955 is already resolved on develop.
2. Keep it as defense-in-depth for harnesses that somehow re-introduce the block, but it would be dead code against the current hook body.

Separately on #965, there is an existing open PR #346 (yoshi280) that attributes the sparse `link_lists.bin` bloat to `persistDirty()` seek drift on default `batch_size=100 / sync_threshold=1000` rather than to the ParallelFor race. It sets `batch_size=50000, sync_threshold=50000` on collection creation and batches per-file upserts. Different root-cause than yours, so the two fixes are orthogonal and could land together: pin threads to cover the SIGSEGV path, raise thresholds to cover the resize-drift path. #346 is dirty against develop (bensig asked for a rebase on 04-11) but not closed.

The `mine_global_lock` piece looks independent. No overlap there.

felipetruman:
Thanks — pushed **d8a26cd** addressing both actionable points from the 2nd review. Response to all four comments:

| # | Concern | Resolution |
|---|---|---|
| **#7, #8** | `multiprocessing.get_context("fork")` unavailable on Windows CI | **Fixed.** Added `_get_mp_context()` helper that returns `"spawn"` on Windows (`os.name == "nt"`) and `"fork"` elsewhere. Both `test_same_palace_serializes_across_processes` and `test_different_palaces_dont_conflict` now use it. Spawn inherits `os.environ` from the parent (so `monkeypatch.setenv("HOME", ...)` still propagates), which is all these tests need. |
| **#10** | `palace_key` not normalized → Windows case-insensitive bypass | **Fixed.** Switched from `os.path.abspath(os.path.expanduser(palace_path))` to `os.path.normcase(os.path.realpath(os.path.expanduser(palace_path)))`. `realpath` canonicalises symlinks and `..`; `normcase` folds case on Windows (no-op on POSIX). `C:\Palace` and `c:\palace` now hash to the same `palace_key`. |
| **#9** | Auto-migrate existing palaces via `collection.modify(...)` | **Deferred to follow-up**, with an honest note on the tradeoff. I re-verified against chromadb 1.5.7: the `modify(configuration=UpdateCollectionConfiguration(hnsw=UpdateHNSWConfiguration(num_threads=1)))` path *does* add `num_threads=1` to the in-memory collection config without raising (unlike the `modify(metadata=...)` path, which is what my original reply was about). So a per-session retrofit at every `ChromaBackend.get_collection()` / MCP `_get_collection()` startup **is** technically possible. What I still observe is that the setting does not persist to disk across `PersistentClient` reopens in 1.5.7 — `configuration_json["hnsw"].get("num_threads")` returns `None` after the next session. So the retrofit would have to be re-applied on every process start rather than being a one-shot migration. Given that fresh palaces are already protected and existing users only lose parallelism (not correctness) on the first insert path before the retrofit runs, I'd rather land this PR as the correctness-boundary fix and open a follow-up issue for "per-session HNSW threading retrofit for legacy collections" so it gets its own review. Happy to push it into this PR instead if you'd prefer it all in one — just say the word. |

## Testing

`pytest tests/test_palace_locks.py tests/test_hooks_cli.py tests/test_backends.py tests/test_cli.py` → **98 passed, 0 failed** on Linux/Python 3.13. Windows CI will exercise the `spawn` path on push.

felipetruman:
## Follow-up: addressing remaining Copilot review items

### Copilot #1 — `subprocess.TimeoutExpired` não capturado em `hooks_cli.py`

**Resolvido** pela resolução do merge conflict com develop. `hook_precompact` agora delega para `_mine_sync()`, que já captura `(OSError, subprocess.TimeoutExpired)`. O teste `test_precompact_with_timeout` (que verifica o comportamento silencioso em caso de timeout) está verde.

### Copilot #2 — Teste de regressão para o attempt-cap

**Não aplicável após rebase.** PR #863 em develop removeu o comportamento de bloqueio do `hook_precompact` (agora minera sincronamente e passa direto). O attempt-cap da nossa branch seria código morto após o rebase. Os testes `test_precompact_first_two_attempts_block` e `test_precompact_counter_is_per_session` foram removidos por serem incompatíveis com a abordagem do develop.

### Copilot #4 — Testes para `mine_global_lock` (agora `mine_palace_lock`)

**Resolvido.** Adicionado `tests/test_palace_locks.py` com 6 testes cobrindo:
- Acquire simples sem erro
- Mesmo palace serializa entre processos (`MineAlreadyRunning` no 2º acquire)
- Palaces diferentes não conflitam (paralelismo preservado)
- Lock reutilizável após release
- Normalização de path (relativo e absoluto mapeiam para o mesmo lock)
- Alias `mine_global_lock` para retrocompatibilidade

Suite completa: **52 testes passando**.

felipetruman:
## Follow-up: merge-conflict cleanup + Copilot Review #3

Pushed commit **ffa9e8e** addressing the two remaining items.

### @mvalentsev — dead `MAX_PRECOMPACT_BLOCK_ATTEMPTS` constant

You were right. After rebasing on develop (#863 landed), `hook_precompact` now delegates to `_mine_sync` and never emits `decision: block`, so the attempt-cap constant was orphaned. I had already removed the counter logic in `404d73f`, but the bare constant `MAX_PRECOMPACT_BLOCK_ATTEMPTS = 2` survived. **Removed** — grep confirms 0 usages.

Also noted your pointer to PR #346 (yoshi280): agreed it's orthogonal — thread-pin covers the SIGSEGV path, the `batch_size`/`sync_threshold` bump covers the resize-drift path. Happy to help coordinate if #346 rebases.

### Copilot Review #3 (low-confidence) — `chroma.py:134` retrofit for legacy palaces

> `get_collection(create=False)` returns an existing Chroma collection without ensuring `hnsw:num_threads` is pinned to 1

**Fixed with a best-effort retrofit.** Added `_pin_hnsw_threads(collection)` that calls `collection.modify(configuration=UpdateCollectionConfiguration(hnsw=UpdateHNSWConfiguration(num_threads=1)))`. Applied on every `get_collection` code path — both `ChromaBackend.get_collection()` (create=False and create=True) and `mcp_server._get_collection()`.

**Caveat I verified empirically against chromadb 1.5.7** (repro below):

```python
coll = c.create_collection('legacy', metadata={'hnsw:space': 'cosine'})
# config before: {..., no num_threads}
coll.modify(configuration=UpdateCollectionConfiguration(
    hnsw=UpdateHNSWConfiguration(num_threads=1)))
# config after modify: {..., 'num_threads': 1}  ← applied
# reopen client, get_collection(...)
# config after reopen: {..., no num_threads}  ← NOT persisted
```

So the retrofit does **not** persist to disk across `PersistentClient` reopens in 1.5.x — it runs in memory each process start. Net effect:

- Fresh palaces → `metadata={"hnsw:num_threads": 1}` at creation (persisted, primary defense).
- Legacy palaces → per-session retrofit at `get_collection` (non-persisted, applies every process start).

Both paths are protected during any active session, which is when the SIGSEGV actually fires. Legacy users no longer need `mempalace nuke` + re-mine to get the fix. Updated the inline comment in `mcp_server.py` accordingly.

The retrofit is wrapped in `try/except` that swallows all errors (including `ImportError` on pre-1.5 chromadb without `UpdateCollectionConfiguration`), so it's strictly additive.

### Testing

`pytest tests/test_backends.py tests/test_palace_locks.py tests/test_hooks_cli.py` → **62 passed**.

New tests in `tests/test_backends.py`:
- `test_pin_hnsw_threads_retrofits_legacy_collection` — legacy collection (no `num_threads`) → retrofit → `num_threads == 1`.
- `test_pin_hnsw_threads_swallows_all_errors` — retrofit on a collection whose `modify()` raises `RuntimeError` — helper must not propagate.
- `test_get_collection_applies_retrofit_on_existing_palace` — bootstrap a legacy palace with plain `metadata={"hnsw:space":"cosine"}`, reopen via `ChromaBackend.get_collection(create=False)`, assert `num_threads == 1` on the returned wrapper.

One pre-existing failure in `tests/test_mcp_stdio_protection.py::test_module_import_redirects_stdout_to_stderr` (missing `overrides` module in my linuxbrew python3.14 sandbox) — reproducible on `develop` clean, unrelated to this PR.

bensig:
@felipetruman pls rebase

bensig:
### Code review

Found 3 issues:

1. **Duplicate of open PR #991.** The `hnsw:num_threads=1` change on `backends/chroma.py` is a strict subset of this PR; #991 is still OPEN with the same primary fix on the same file. A reviewer on #991 already flagged that this PR's chroma.py hunk supersedes it. Before merging either, one should be closed/redirected — shipping both invites a merge-order footgun where the narrower PR lands first, then this one's broader retrofit (`_pin_hnsw_threads` at `mcp_server._get_collection`, legacy-palace coverage) silently rebases out the duplicate hunks. For a v3.3.2 release cut, pick one path and close the other explicitly.

https://github.com/MemPalace/mempalace/blob/ffa9e8ebf92d9366384254acd910b1a168e1fe9f/mempalace/backends/chroma.py#L158-L170

2. **PR title/body still claims `closes #955` but the compaction fix was removed on rebase.** Per felipetruman's own follow-up: #863 already landed in develop and removed `hook_precompact`'s blocking behavior, so the attempt-cap logic (and the `MAX_PRECOMPACT_BLOCK_ATTEMPTS` constant) were deleted from this branch. `hooks_cli.py` no longer appears in the PR's changed files. The PR title and Summary still advertise a three-for-one fix including #955 — a reviewer triaging for v3.3.2 will expect compaction-deadlock coverage that isn't here. Update the title/body to reflect the actual scope (`closes #974, #965` only) and note that #955 is resolved independently by #863, so the issue should be closed via #863, not this PR.

https://github.com/MemPalace/mempalace/blob/ffa9e8ebf92d9366384254acd910b1a168e1fe9f/mempalace/palace.py#L313-L392

3. **Merge state is DIRTY (`needs-rebase`) against develop.** `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`; maintainer has explicitly requested a rebase. The PR cannot be evaluated against a stable base in its current state — the last substantive refactor in develop (#863) already invalidated one of the three advertised fixes, so a second rebase pass is likely to surface further scope trimming. Rebase before this lands or before a deeper review round.

https://github.com/MemPalace/mempalace/blob/ffa9e8ebf92d9366384254acd910b1a168e1fe9f/mempalace/miner.py#L756-L801

Qodo-Free-For-OSS:
Hi, New tests assert that `_pin_hnsw_threads()` updates `configuration_json`, but with the repo’s pinned chromadb==0.6.3 the function returns early (ImportError) and no retrofit occurs, causing the assertions to fail and blocking CI.

**How to fix:** Skip tests for chromadb<1.5

> ## Issue description
> `tests/test_backends.py` includes new tests that require ChromaDB 1.5+ collection configuration update APIs and expect `configuration_json["hnsw"]["num_threads"]` to be set after calling `_pin_hnsw_threads()`. The repository lockfile pins `chromadb==0.6.3`, where `_pin_hnsw_threads()` returns early due to ImportError, so these tests will fail in CI.
> 
> ## Issue Context
> `_pin_hnsw_threads()` is intentionally best-effort and version-gated; tests must reflect that behavior under the currently pinned dependency set.
> 
> ## Fix Focus Areas
> - tests/test_backends.py[150-197]
> - mempalace/backends/chroma.py[14-35]
> - uv.lock[342-345]
> 
> ## Suggested fix
> Choose one:
> 1) Add a version guard in the new tests (e.g., `pytest.importorskip`/`skipif`) so they only run when `chromadb.api.collection_configuration` is available.
> 2) Alternatively, update the pinned ChromaDB version used in CI/dev lockfile to a compatible 1.5+ version if that’s now the intended baseline.
> 3) Or mock the 1.5+ Update* classes in tests so the behavior can be validated without changing the pinned dependency.

---

bensig:
@felipetruman heads up — this branch has merge conflicts against develop and needs a rebase before it can ship.

bensig (approved):

jphein:
A narrower slice of the mine fan-out fix (hook-triggered only) landed on develop yesterday as #1023 — `_mine_already_running()` with a PID file guard in `hooks_cli.py`. Your broader `mine_global_lock()` still adds real value by catching direct-CLI fan-out too (hooks aren't the only way `mine` gets spawned twice). The hook-path hunk should conflict-resolve cleanly against #1023's version on rebase.

Separately: the `hnsw:num_threads: 1` pin is the right conservative default given ChromaDB ingests one drawer at a time. Worth landing on its own if the combined PR stays hard to review.

bensig:
Hey @felipetruman — quick ping. We're cutting **v3.3.3** in the next few hours to unbreak [#1093](https://github.com/MemPalace/mempalace/issues/1093) (v3.3.2 ships a plugin.json referencing a `mempalace-mcp` binary that `pyproject.toml` never declared — fresh `pip install` is broken for everyone).

If you can **rebase this against `develop`** and drop the reference to [#955](https://github.com/MemPalace/mempalace/issues/955) (already fixed by [#863](https://github.com/MemPalace/mempalace/pull/863)), we'd include this in v3.3.3 and close [#974](https://github.com/MemPalace/mempalace/issues/974) (HNSW SIGSEGV) in the same release.

If you can't get to it today — no pressure at all, it rolls to v3.3.4. Thanks for the work on this one.

jphein:
Corroborating data from a separate install.

Reproduced the exact failure mode on my fork (jphein/mempalace main tracking upstream develop, currently v3.3.4) today. Details:

- **Palace scale:** 135K drawers, ChromaDB 1.5.x (currently 1.5.7)
- **Concurrency pattern:** Claude Code spawns one `mcp_server.py` per open terminal + stop hooks spawn `mempalace mine` subprocesses — typically 4+ concurrent writers against a single palace directory
- **Failure signature:** matches this PR's description exactly — runaway writes to the HNSW segment
- **Magnitude:** quarantined a **437 GB** bloated segment (the whole palace dir had ballooned to 485 GB; deleting the two drift quarantines reclaimed 472 GB)
- **User-visible symptom:** fresh-process SIGSEGV in `chromadb_rust_bindings.abi3.so` on stop-hook fire, repeatably

@bensig / @rboarescu: scale reference — Nobara report quotes ~1.5 TB on an unknown-size palace; my 437 GB was on 135K drawers. Same mechanism, similar ratio.

I landed three defense-in-depth PRs in my fork today that catch the downstream effects of this race ([#1171](https://github.com/milla-jovovich/mempalace/pull/1171) backend-seam flock, [#1173](https://github.com/milla-jovovich/mempalace/pull/1173) quarantine on every make_client, [#1177](https://github.com/milla-jovovich/mempalace/pull/1177) `.blob_seq_ids_migrated` marker guard). None of them fix the race itself — that's this PR. My fork's approach serializes writes with `fcntl.flock` so only one client writes at a time, which reduces but doesn't eliminate the corruption because ChromaDB's internal `ParallelFor` can still race *within* a single client's write.

The `hnsw:num_threads: 1` pin in this PR is the actual root-cause fix. Happy to help with the rebase against current `develop` (it's `CONFLICTING` with today's entity-detection merges) if that would be useful — I've been working in this area all day and can put in a few hours on Friday if you want an extra pair of hands.

Thanks for chasing this one down — it's been a quietly expensive bug.

igorls:
Salve @felipetruman LGTM! Obrigado pelo PR

Seph396:
Reproduced this exact corruption pattern on `mempalace 3.3.2` running under Claude Desktop, and confirming the fix in this PR resolves it.

**Trigger pattern observed:** Claude Desktop's MCP loader spawns a new `mempalace.mcp_server` process before the previous one finishes its shutdown handshake. Two concurrent server processes operating against the same Chroma collection caused the multi-threaded HNSW writer race this PR fixes (`hnsw:num_threads=1`).

**Disk impact in our case:** `~/.mempalace/palace/<drawers-uuid>/link_lists.bin` bloated to 55 GB (1900× expected size for our ~50K vectors), which triggered a kernel-level watchdog timeout on a 1 TB Mac when free disk dropped below ~6 GB. Recovery required surgical removal of the corrupted collection only — `chroma.sqlite3` and the second collection's HNSW segment were intact, so no source data was lost.

**Verification of the fix:** Pinned to commit `0d9929c0dd31f5f8368d55300c2e71d8eb0f4e46` via `uv tool install --reinstall 'mempalace @ git+https://github.com/MemPalace/mempalace@0d9929c0dd31f5f8368d55300c2e71d8eb0f4e46'`. Confirmed the changes at `mcp_server.py:229` (collection config sets `hnsw:num_threads=1`) and `backends/chroma.py:134` (auto-retrofit for existing collections). 24+ hours of mixed concurrent use across Claude Desktop, Claude Code, and Cowork since the pin — no recurrence.

**Full sanitized transport log + analysis:** https://gist.github.com/Seph396/d8f724e58f066201b3cb527d0c7ffcc0

A separate observation: the concurrent-spawn behavior on the Claude Desktop client side is what makes this race triggerable in the first place. The `num_threads=1` fix in this PR closes the race window structurally (no concurrency in the writer), but a single-instance lock or PID file inside `mempalace.mcp_server` would also defend against any future client that exhibits the same spawn pattern. Filing that as a separate issue: https://github.com/MemPalace/mempalace/issues/1229

Looking forward to v3.3.4. Happy to provide additional logs or repro steps if useful.

PR #982, opened 2026-04-17
feat(embeddings): opt-in Ollama embedding backend for GPU acceleration

felipetruman:
## Summary

Adds an opt-in `OllamaEmbeddingFunction` path to the ChromaDB backend. When `EMBEDDING_PROVIDER=ollama` is set, MemPalace routes all embedding generation through a local Ollama server — typically GPU-accelerated on machines with a supported GPU (AMD via ROCm, NVIDIA via CUDA, Apple Silicon).

Default behavior is unchanged: without the env var, ChromaDB's bundled `DefaultEmbeddingFunction` (ONNX MiniLM, 384-dim, CPU) is used.

## Why

On a machine with a local GPU, the default ONNX MiniLM running on CPU becomes the bottleneck for `mempalace mine` on large projects — on my workstation (AMD Radeon RX 9060 XT, 16 GB VRAM), mining `~/freedomdigitalhub` with MiniLM/CPU was still running after 2 hours. A persistent Ollama+nomic-embed-text server on the same GPU generates embeddings at ~14 ms/embed warm, turning hour-scale mines into minute-scale ones.

## What changed

- New `get_embedding_function()` in `mempalace/backends/chroma.py` — pure env-var gated factory.
- `ChromaBackend.get_collection`, `get_or_create_collection`, and `create_collection` now forward the embedding function to ChromaDB.
- `mcp_server._get_collection` (which bypasses ChromaBackend for its own inode-based cache) also forwards the EF, keeping writes and reads consistent.

### Environment variables (all optional)

| Variable | Default | Meaning |
|---|---|---|
| `EMBEDDING_PROVIDER` | (unset) | Set to \`ollama\` to enable |
| `OLLAMA_URL` | \`http://localhost:11434\` | Base URL of the Ollama server |
| `OLLAMA_EMBED_MODEL` | \`nomic-embed-text\` | Model tag for \`/api/embeddings\` |
| `OLLAMA_EMBED_TIMEOUT` | \`60\` | HTTP timeout in seconds |

## Compatibility

This is **opt-in**. Existing palaces are not touched unless the user sets `EMBEDDING_PROVIDER=ollama`.

Because `nomic-embed-text` produces **768-dim** vectors and MiniLM produces 384-dim, switching providers on an existing palace requires `mempalace nuke` + remine. The docs should make this explicit — happy to add a note in a follow-up commit or in this PR, whichever reviewers prefer.

## Verification

End-to-end on the reporter's workstation (Python 3.14, chromadb 1.5.7, ROCm, AMD RX 9060 XT):

- \`get_embedding_function()\` returns an \`OllamaEmbeddingFunction\` with the expected base URL / model / timeout
- Fresh \`mempalace init\` + \`mempalace mine --limit 20\` produced **425 drawers**, all in 768-dim collections (verified: \`SELECT dimension FROM collections\` on \`chroma.sqlite3\` → \`768\` for both \`mempalace_drawers\` and \`mempalace_closets\`)
- \`mempalace search\` round-trips correctly and returns semantically ranked matches
- VRAM during mine: ~2.33 GB (nomic-embed-text loaded)

## Test plan

- [x] Default path still works (no env → Chroma default, 384-dim)
- [x] Opt-in path works (env set → Ollama, 768-dim)
- [x] Write + read use same EF (search works)
- [ ] CI (reviewers: please run the existing test suite)

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -3,12 +3,31 @@
 import logging
 import os
 import sqlite3
+from typing import Any, Dict
 
 import chromadb
 
 from .base import BaseCollection
 
 logger = logging.getLogger(__name__)
+
+
+_DEFAULT_OLLAMA_URL = "http://localhost:11434"
+_DEFAULT_OLLAMA_MODEL = "nomic-embed-text"
+_DEFAULT_OLLAMA_TIMEOUT = 60
+
+
+def _parse_timeout(raw):
+    """Parse OLLAMA_EMBED_TIMEOUT env with a clear error on non-integers."""
+    if raw is None or str(raw).strip() == "":
+        return _DEFAULT_OLLAMA_TIMEOUT
+    try:
+        return int(str(raw).strip())
+    except ValueError as exc:
+        raise ValueError(
+            f"OLLAMA_EMBED_TIMEOUT must be an integer number of seconds, "
+            f"got {raw!r}"
+        ) from exc
 
 
 def get_embedding_function():
@@ -24,16 +43,29 @@ def get_embedding_function():
     Returning ``None`` keeps ChromaDB's ``DefaultEmbeddingFunction`` (ONNX
     MiniLM, 384 dims, CPU) for backward compatibility with existing palaces.
     """
-    provider = os.environ.get("EMBEDDING_PROVIDER", "").lower()
+    provider = os.environ.get("EMBEDDING_PROVIDER", "").strip().lower()
     if provider != "ollama":
         return None
     from chromadb.utils.embedding_functions import OllamaEmbeddingFunction
 
     return OllamaEmbeddingFunction(
-        url=os.environ.get("OLLAMA_URL") or "http://localhost:11434",
-        model_name=os.environ.get("OLLAMA_EMBED_MODEL") or "nomic-embed-text",
-        timeout=int(os.environ.get("OLLAMA_EMBED_TIMEOUT") or "60"),
+        url=(os.environ.get("OLLAMA_URL") or _DEFAULT_OLLAMA_URL).strip(),
+        model_name=(os.environ.get("OLLAMA_EMBED_MODEL") or _DEFAULT_OLLAMA_MODEL).strip(),
+        timeout=_parse_timeout(os.environ.get("OLLAMA_EMBED_TIMEOUT")),
     )
+
+
+def _ef_kwargs() -> Dict[str, Any]:
+    """Return ``{"embedding_function": ef}`` when an EF is configured, else ``{}``.
+
+    Spreading this dict into ChromaDB's ``get_or_create_collection`` / ``get_collection``
+    / ``create_collection`` calls keeps the default-embedding-function path completely
+    untouched (no ``embedding_function=None`` kwarg is passed) and keeps the Ollama
+    path explicit, eliminating both the typing mismatch and any risk of passing
+    ``None`` into a positional that older chromadb releases may reject.
+    """
+    ef = get_embedding_function()
+    return {"embedding_function": ef} if ef is not None else {}
 
 
 def _fix_blob_seq_ids(palace_path: str):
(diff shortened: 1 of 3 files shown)


felipetruman:
Thanks for the review @copilot-pull-request-reviewer. All 7 comments addressed in 06956aa. Item-by-item:

| # | File:Line | Comment | Resolution |
|---|---|---|---|
| 1 | `mcp_server.py:226` | `# type: ignore[arg-type]` masking `None` mismatch | Removed. New `_ef_kwargs()` helper returns `{}` when EF is not configured; call sites spread it via `**` so `embedding_function` is never passed as `None`. |
| 2 | `mcp_server.py:220` | Bare `except` swallows EF misconfig | Now `logger.debug("_get_collection failed", exc_info=True)` before returning `None`. Kept broad on purpose — palace-not-found must stay non-fatal — but misconfigs are diagnosable now. |
| 3 | `chroma.py:162` | `embedding_function=ef` can be `None` | Same `**_ef_kwargs()` pattern applied to `get_collection` / `get_or_create_collection`. |
| 4 | `chroma.py:183` | `create_collection` same issue | Same pattern applied; also keeps older chromadb releases that reject `embedding_function=None` happy. |
| 5 | `chroma.py:36` | No unit tests | New `tests/test_embeddings_ollama.py` — 25 tests, all passing, no Ollama server required. Covers provider-disabled / enabled / normalization / env overrides / bad-timeout / `_parse_timeout` valid+invalid / `_ef_kwargs` empty+populated / spread invariant / no cached state. |
| 6 | `chroma.py:36` | `int(...)` on `OLLAMA_EMBED_TIMEOUT` raises opaque `ValueError` | Extracted to `_parse_timeout()` which names the env var in the error (`"OLLAMA_EMBED_TIMEOUT must be an integer number of seconds, got 'foo'"`) and falls back to 60 on empty/whitespace. |
| 7 | `chroma.py:27` | `EMBEDDING_PROVIDER` not stripped | Now `.strip().lower()`. Also applied `.strip()` to `OLLAMA_URL` / `OLLAMA_EMBED_MODEL` values for consistency. |

Classification: all 7 fell into "quality / design" — no blocking critical bugs. One design pattern change (the `_ef_kwargs()` helper) made the same improvement apply to all 3 collection accessor call sites + the MCP server bypass, instead of four separate fixes.

CI: no checks report back on this PR — let me know if there's something I should wire up locally.

Qodo-Free-For-OSS:
Hi, get_embedding_function() selects the env var value before stripping, so whitespace-only OLLAMA_URL or OLLAMA_EMBED_MODEL becomes an empty string and is passed into OllamaEmbeddingFunction, causing embedding requests to fail at runtime.

**How to fix:** Validate stripped URL/model strings

> ### Issue description
> `get_embedding_function()` currently computes `url` and `model_name` as `(env or default).strip()`. If the env var is present but whitespace-only, the value becomes an empty string and gets passed to `OllamaEmbeddingFunction`, causing runtime failures.
> 
> ### Issue Context
> This occurs only when `EMBEDDING_PROVIDER=ollama`.
> 
> ### Fix Focus Areas
> - mempalace/backends/chroma.py[33-55]
> 
> ### Suggested fix
> - Read env vars, strip them, then if the result is empty, fall back to defaults (or raise a clear ValueError).
>   - Example approach:
>     - `raw_url = os.environ.get("OLLAMA_URL")`
>     - `url = (raw_url or "").strip() or _DEFAULT_OLLAMA_URL`
>     - similarly for `model_name`
> - Consider adding a unit test for whitespace-only env values to ensure defaults are used (or error is raised).

---

igorls:

Motokiyo:
**+1 from a heavy multilingual user** — landing this would close a real gap.

I've been monkey-patching the same hook locally since v3.0.0 (FR/EN corpus, 68k drawers). The architectural cleanup in 3.3.4 (`embedding.py` factory, `backends/chroma.py` wrapper) made my patch trivial — 17 lines, one file. This PR is essentially the same idea applied at the backend layer.

One suggestion: consider documenting `bge-m3` as the recommended model for non-English corpora, or even shipping it as an alternate default. `nomic-embed-text` (768d, English bias) gives near-random semantic ranking on French / Korean / Japanese content. `bge-m3` (1024d, 100+ langs, BAAI) restores meaningful similarity scores — I see queries like *"détection de chute pour personnes âgées EHPAD"* correctly returning `fall_detector.py` as the top hit, which MiniLM and `nomic-embed-text` both rank deep in the noise.

Tested today against v3.3.4 with this PR's approach (re-implemented locally as a `mempalace/embedding.py` hook, since I can't easily run a development checkout side-by-side with my pip install). Hybrid BM25 + cosine search (also new in 3.3.0, great work!) further compensates for embedding model choice — but exact-name retrieval alone isn't a substitute for proper multilingual semantic similarity.

Happy to share concrete metrics on a 68k-drawer FR/EN palace (latency, recall@k, hybrid score distributions) if it would help reviewers, just ping me.

PR #985, opened 2026-04-17
feat(hooks): tune stop-hook cadence via env vars (MEMPAL_SAVE_INTERVAL / MEMPAL_STOP_HOOK_DISABLE)

felipetruman:
## Summary

The stop hook currently blocks every **15 human messages** with the \"AUTO-SAVE checkpoint\" prompt, hard-coded via the module-level \`SAVE_INTERVAL\` constant in \`mempalace/hooks_cli.py\`. Two new env vars let users tune that without reinstalling or editing plugin files:

| Variable | Effect |
|---|---|
| \`MEMPAL_SAVE_INTERVAL=N\` | Override the cadence (clamped to \`max(1, N)\`; invalid/empty falls back to 15 so a typo can't collapse the interval to 0) |
| \`MEMPAL_STOP_HOOK_DISABLE=1\` | Pass-through (also accepts \`true\`/\`yes\`/\`on\`, case-insensitive, whitespace trimmed). State still tracked, no blocking. |

Both are read on every hook invocation, so users can change behavior mid-session without touching the install.

## Why

Two concrete cases I hit on my workstation today:

1. **Long Ollama-GPU embedding migration session** — a session that does real work (editing the backend, reviewing a PR, adding tests) easily crosses 15 messages multiple times, and each block interrupts flow to repeat the same 3 MCP calls. A higher per-session interval (e.g. \`MEMPAL_SAVE_INTERVAL=50\`) keeps the auto-save intent while drastically cutting noise.
2. **Plugin installed but MCP tools not registered for the current project** — the block message instructs the model to call \`mempalace_diary_write\` / \`mempalace_add_drawer\` / \`mempalace_kg_add\`, but if the MemPalace MCP server isn't exposed in the active Claude session those tools aren't callable. The model can't satisfy the block, the user sees a loop of error messages. \`MEMPAL_STOP_HOOK_DISABLE=1\` gives a clean opt-out.

## Behavior preserved

- Default cadence is still 15 (existing users see no change).
- \`SAVE_INTERVAL\` constant is still exported from the module (set to \`DEFAULT_SAVE_INTERVAL\`), so any external code / tests that read it keep working.
- The existing stop_hook_active infinite-loop guard and MEMPAL_DIR auto-ingest paths are untouched.

## Test plan

Added \`tests/test_hooks_cli_tuning.py\` — **22 unit tests**, all passing:

- Default when unset.
- Integer overrides with whitespace trim (\`\"  100  \"\` → \`100\`).
- Empty / whitespace values fall back to default.
- Invalid strings (\`\"abc\"\`, \`\"3.5\"\`, \`\"--\"\`) fall back to default rather than raising or silently zeroing.
- Non-positive values (\`\"0\"\`, \`\"-1\"\`) clamped to 1.
- Disable flag truthy variants (\`1\`, \`true\`, \`TRUE\`, \`Yes\`, \`on\`, \`\"  true  \"\`) all return True.
- Disable flag falsy variants (\`0\`, \`false\`, \`no\`, \`off\`, \`\"\"\`, \`\"maybe\"\`) all return False.
- Backwards-compat constant export still equals 15.

\`\`\`
$ pytest tests/test_hooks_cli_tuning.py -v
22 passed in 0.06s
\`\`\`

## Files

- \`mempalace/hooks_cli.py\` — new \`_get_save_interval()\` / \`_stop_hook_disabled()\` helpers + wired into \`hook_stop\`.
- \`tests/test_hooks_cli_tuning.py\` — new test file.

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -14,7 +14,45 @@ import sys
 from datetime import datetime
 from pathlib import Path
 
-SAVE_INTERVAL = 15
+DEFAULT_SAVE_INTERVAL = 15
+
+
+def _get_save_interval() -> int:
+    """Return the configured save interval, clamped to a sane minimum.
+
+    Reads ``MEMPAL_SAVE_INTERVAL`` from the environment on every hook
+    invocation so users can re-tune the cadence without reinstalling.
+    Invalid / non-positive values silently fall back to the default so a
+    typo can never make the hook block on every single turn.
+    """
+    raw = os.environ.get("MEMPAL_SAVE_INTERVAL", "").strip()
+    if not raw:
+        return DEFAULT_SAVE_INTERVAL
+    try:
+        value = int(raw)
+    except ValueError:
+        return DEFAULT_SAVE_INTERVAL
+    return max(1, value)
+
+
+def _stop_hook_disabled() -> bool:
+    """True when ``MEMPAL_STOP_HOOK_DISABLE`` is set to a truthy value.
+
+    Lets users opt out of the auto-save nag entirely (e.g. on machines
+    where MemPalace MCP tools are not registered in the session) without
+    having to uninstall the plugin.
+    """
+    return os.environ.get("MEMPAL_STOP_HOOK_DISABLE", "").strip().lower() in (
+        "1",
+        "true",
+        "yes",
+        "on",
+    )
+
+
+# Backwards-compatible module-level constant (re-exported for tests that
+# assert the default value; runtime always calls _get_save_interval()).
+SAVE_INTERVAL = DEFAULT_SAVE_INTERVAL
 STATE_DIR = Path.home() / ".mempalace" / "hook_state"
 
 STOP_BLOCK_REASON = (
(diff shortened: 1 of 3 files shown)


felipetruman:
Thanks for the review. All 3 comments addressed in **ba07cb6**.

| # | File:Line | Comment | Resolution |
|---|---|---|---|
| **#1** | `hooks_cli.py:25-26` | Docstring said "fall back to default" but code clamps to 1 via `max(1, value)` | **Rewrote the docstring** to document actual behavior: missing/non-integer → default, parsed `< 1` → clamped to 1 (blocks every turn, intentional). Added pointer to `MEMPAL_STOP_HOOK_DISABLE` for users who wanted "no block" rather than "block every turn". |
| **#2** | `hooks_cli.py:260` | `MEMPAL_STOP_HOOK_DISABLE` bypassed state tracking — toggling the flag off mid-session would include every accumulated message in `since_last` and trigger an immediate block | **Fixed.** Disabled path now advances the `*_last_save` watermark to the current `exchange_count` before returning `{}`. State stays synced with activity while disabled, so re-enabling resumes the normal cadence from the current position. No block is ever emitted while disabled. |
| **#3** | `hooks_cli.py:249` | Env-var behavior only unit-tested at helper level; existing `hook_stop` tests don't clear env vars | **Added 5 integration tests** in `test_hooks_cli_tuning.py` that run `hook_stop` end-to-end: custom interval triggers block at the override; custom interval passes through below override; `STOP_HOOK_DISABLE=1` passes through even far above interval; disabled path advances the watermark (proves #2 fix); `stop_hook_active` short-circuit does not touch state. Also added an **autouse fixture** in `test_hooks_cli.py` that clears `MEMPAL_SAVE_INTERVAL` / `MEMPAL_STOP_HOOK_DISABLE` so a developer's shell config can't make these tests flaky. |

## Testing

`pytest tests/test_hooks_cli.py tests/test_hooks_cli_tuning.py` → **79 passed** (46 existing + 33 tuning, 5 of which are new integration tests).

Classification: #1 quality (docs), #2 critical (bug — disabled flag actually broke on toggle-off), #3 quality (test coverage). The state-tracking fix in #2 is the load-bearing one — the docstring and test improvements make the feature safe to recommend.

igorls:

PR #987, opened 2026-04-18
fix(mcp): handle null JSON-RPC request payloads safely

alpiua:
## What does this PR do?
When an MCP client sends a malformed or `null` top-level request, `handle_request(request)` crashes with an `AttributeError` because it attempts to call `.get()` on a non-dictionary object. 

This PR adds explicit dictionary type validation at the top of the handler. Instead of crashing the server and disconnecting the client, it now gracefully returns a standard JSON-RPC Error `-32600` (Invalid Request), adhering to robust null-handling practices for JSON-RPC parameters.

## How to test
1. Run the MemPalace MCP server locally.
2. Send a malformed payload (e.g., `null`, `[]`, or a plain string) to the standard input.
3. Observe that instead of throwing an exception (which forces a restart), the server correctly replies with:
`{"jsonrpc": "2.0", "error": {"code": -32600, "message": "Invalid Request"}}`

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1969,7 +1969,11 @@ SUPPORTED_PROTOCOL_VERSIONS = [
 
 def handle_request(request):
     if not isinstance(request, dict):
-        return {"jsonrpc": "2.0", "id": None, "error": {"code": -32600, "message": "Invalid Request"}}
+        return {
+            "jsonrpc": "2.0",
+            "id": None,
+            "error": {"code": -32600, "message": "Invalid Request"},
+        }
     method = request.get("method") or ""
     params = request.get("params") or {}
     req_id = request.get("id")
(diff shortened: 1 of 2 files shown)


mvalentsev:
Good catch on the `AttributeError`. Two small nits before this lands:

**1. Missing `"id": null` per JSON-RPC 2.0 §5.1**

Other error branches in `handle_request` include `"id": req_id` (tool errors, unknown method). Spec says: *"If there was an error in detecting the id in the Request object ... it MUST be Null"*. Strict clients correlate request-response via `id`; without it they may ignore the response. Suggested:

```python
return {"jsonrpc": "2.0", "id": None, "error": {"code": -32600, "message": "Invalid Request"}}
```

**2. No regression test**

A parametrised test covering `None`, `[]`, `"str"`, `42`, `True` payloads would lock the fix in. Something like:

```python
@pytest.mark.parametrize("payload", [None, [], "plain", 42, True])
def test_handle_request_invalid_payload_returns_jsonrpc_error(payload):
    resp = handle_request(payload)
    assert resp == {"jsonrpc": "2.0", "id": None, "error": {"code": -32600, "message": "Invalid Request"}}
```

Otherwise a narrow fix on a real bug.

alpiua:
@mvalentsev faced the issue when agent just refused to send a proper request.

Mitigated this by adding `test_tools_call_missing_params` and the corresponding fix. 
So there was a gap at defense layers. Now:
1. The previous test (`-32600 Invalid Request`) ensures the entire payload is a valid JSON dictionary before routing.
2. This new test (`-32602 Invalid params`) ensures that once the request reaches `tools/call`, it explicitly contains the mandatory `params` object and `name` field, preventing an `AttributeError` crash on the server end.

Issue #989, opened 2026-04-18
RFC: Source adapter plugin specification


bensig:
## Context

The storage backend seam (#413, formalized in #743) solved pluggable *writes*. The mirror problem exists on the *read* side: source adapters that mine content into the palace.

Six community PRs are now building source-specific ingesters on ad-hoc surfaces:

- #274 Cursor SQLite (`workspaceStorage/*.vscdb`)
- #23 OpenCode SQLite
- #169 Pi agent JSONL
- #232 Cursor JSONL (earlier variant)
- #567 / #98 git-mine (commits, PRs, review threads, decision-signal regex)
- #591 / #592 Delphi Oracle real-time signals
- #702 Cursor + factory.ai combined

Plus three ingesters already grafted into core without a shared contract:

- \`mempalace/miner.py\` — filesystem miner
- \`mempalace/convo_miner.py\` — conversation miner
- \`mempalace/normalize.py\` — format detection for four chat-export shapes

Plus one open proposal for a different ingest semantic:

- #981 — path-level descriptions (metadata-as-content instead of raw bytes)

Each PR reinvents source discovery, item identity, incremental-ingest bookkeeping, metadata shape, and chunking strategy. We need a formal plugin specification so adapter authors can build to a stable contract, matching what RFC 001 does for storage backends.

## What the spec should define

1. **Adapter interface** — unified \`ingest()\` yielding \`SourceItemMetadata | DrawerRecord\`, \`describe_schema()\`, optional \`is_current()\` for incremental
2. **Declared transformations** — every adapter publishes the set of transformations it applies to source bytes; \`byte_preserving\` adapters declare the empty set. Replaces the informal \"verbatim always\" promise with a verifiable one (current \`convo_miner.py\` + \`normalize.py\` pipeline is extensively transformed; this makes it honest)
3. **Registration** — entry-point group \`mempalace.sources\` (third-party packages ship as \`pip install mempalace-source-<name>\`)
4. **Metadata schema** — universal required fields (no renames to existing drawer fields) + per-adapter declared schema; flat values only per ChromaDB constraints
5. **Privacy class** — declarable per adapter; enforced per palace via \`privacy_floor\`; enables regulated-domain use cases
6. **Incremental ingest** — palace IS the cursor via \`is_current(item, existing_metadata)\`; no sidecar
7. **Closet integration** — core builds closets post-step; adapters MAY emit flat \`closet_hints\`; closes current gap where conversation drawers get no closets
[...]

Issue #981, opened 2026-04-17
feat: path-level mining descriptions — mine metadata instead of content for matched paths


roip:
## Problem

MemPalace currently has two modes for any given file: mine everything (chunk all content into drawers), or exclude entirely (via .gitignore / SKIP_DIRS). There's no middle ground.

Real projects have directories where the palace needs to know **what files are and what they do**, but embedding every line of content is wasteful or harmful:

- **Experiment result dumps** — hundreds of KB of JSON numbers. The palace should know 'this is centroid tracking data from Phase 2' not embed 2000 chunks of float arrays.
- **Test directories** — the palace should know 'pytest suite for the miner and MCP server', not embed every assert statement.
- **Data/content folders** — static assets, CSV datasets, model checkpoints. Descriptions matter, content doesn't.
- **Generated output** — build artifacts, compiled files, log directories.

Today the only workaround is: .gitignore the heavy files, then manually write a separate markdown doc describing them and hope it gets mined. This is fragile and doesn't scale.

## Proposed solution

Add a **descriptions** section to \mempalace.yaml\:

\\\yaml
wing: myproject
rooms:
  - name: experiments
    keywords: [data, results]

descriptions:
  'data/results/*.json': 'Experiment result JSONs from task-geometry Phase 2 — centroid tracking, behavioral validation, spectral features'
  'data/checkpoints/': 'Model checkpoints saved during DPO fine-tuning runs'
  'tests/': 'Pytest test suite covering mining, MCP server, room detection, and ignore logic'
  '**/*_responses.jsonl': 'Raw model response dumps from steering/calibration experiments'
\\\

### Mining behavior

When \scan_project\ encounters a file matching a description pattern:
1. **Skip** normal content chunking (no 800-char chunks of raw data)
2. **Create one drawer** with the description text as content
3. Set \source_file\ metadata to the matched path (so search can surface it)
4. Respect mtime — only re-create the drawer if the description changed

This gives the palace semantic knowledge about what's in a path without the noise of embedding raw content.

### Pattern matching

Use the same glob syntax as .gitignore — users already know it. Patterns are matched against the project-relative path.

## Use cases

- Any ML project with large result files, activation dumps, or model outputs
- Web projects with asset directories, uploaded content, or build output
- Monorepos where some subdirectories should be described, not indexed
- Projects with test suites that are useful to reference but not embed line-by-line

Issue #274, opened 2026-04-08
feat: Native Cursor SQLite Ingestion Support


Perseusxrltd:
# Feature Request: Native Cursor Ingestion Support

## 🚀 The Problem
Currently, MemPalace relies on standard text/JSON exports or manual mining of directories. However, **Cursor** (the AI code editor) stores its most valuable conversational "DNA" (including agent reasoning and composer history) inside internal SQLite databases (`state.vscdb`) located in obfuscated workspace storage directories. Users cannot easily "mine" this memory into MemPalace without manual extraction.

## 💡 The Proposal
I propose adding a native Cursor ingestion mode to the `mempalace mine` command. This would allow MemPalace to automatically:
1.  Locate Cursor's `workspaceStorage` directory (platform-specific).
2.  Iterate through all workspace SQLite databases.
3.  Extract conversational data from the `ItemTable` (specifically the `composer.composerData` key).
4.  File these sessions verbatim into appropriate Rooms/Wings in the Memory Palace.

## 🛠️ Technical Details
I have already developed a working prototype script that achieves this on Windows:
*   **Path:** `%APPDATA%\Cursor\User\workspaceStorage`
*   **Query:** `SELECT value FROM ItemTable WHERE key = 'composer.composerData'`
*   **Data Format:** The value is a JSON string containing an `allComposers` array with full chat transcripts.

By integrating this logic into `mempalace/importers/cursor.py`, we can provide a zero-config way for Cursor users to back up their entire coding history into MemPalace.

## ✅ Alignment with MemPalace Principles
*   **Verbatim First:** Extracts raw transcripts exactly as they appear in the Cursor DB.
*   **Local First:** No cloud API or export required; works entirely on the local filesystem.
*   **Zero API:** Uses standard Python `sqlite3` and `json` modules.

## 🧪 Proof of Concept
I successfully mined **160 sessions** from a local Cursor installation using this method today. 

I am happy to submit a PR if this approach aligns with the roadmap!

nanoscopic:
I was also wondering about integration with Cursor. Cursor does not have hooks for receiving prompts and responses live as it is running.

I did RE Cursor package itself and modify it to have hooks, so it could be done, but I dislike having to do that to get things integrated to it.

Mining its storage is fine, but I'm unclear what the utility of that would be if it doesn't pull context into a session. How would you use this mined information? I haven't yet tried it but is the idea to use MCP integration? Would that even work?

PR #990, opened 2026-04-18
docs: RFC 002 — Source adapter plugin specification

bensig:
## Summary

- Drafts a formal plugin specification for source adapters (ingest side) that mirrors RFC 001's role for storage backends (write side). Third parties ship `pip install mempalace-source-<name>` packages via entry points.
- Replaces the informal "verbatim always" promise with a verifiable **declared-transformation** model so every adapter publishes what it does to source bytes and the conformance suite verifies it.
- Consolidates six in-flight community ingesters (#274, #23, #169, #232, #567/#98, #702) plus #981's metadata-only mode under one contract.

Tracking issue: #989

## Key decisions in the draft

- **Single `ingest()` method** — lazy adapters yield `SourceItemMetadata` ahead of drawers; eager adapters interleave. Avoids the complexity of a `discover` / `extract` split for v1 with no clear laziness benefit in current ingesters.
- **Declared transformations** (§1.4) — `declared_transformations: ClassVar[frozenset[str]]` with 13 reserved names (`utf8_replace_invalid`, `line_join_spaces`, `strip_tool_chrome`, `spellcheck_user`, …). `byte_preserving` adapters declare the empty set; `declared_lossy` adapters enumerate. Existing `miner.py` and the `convo_miner.py` + `normalize.py` pipeline map cleanly onto two declared sets (§1.4 existing-code mapping table). The current pipeline is *not* byte-preserving — it strips tool chrome, collapses blank lines, joins AI response lines with spaces, applies spellcheck to user turns, truncates Bash output, inserts synthesized markers. This spec makes that honest.
- **Palace-is-the-cursor incremental** — adapter implements `is_current(item, existing_metadata)` instead of a persisted sidecar cursor blob. Matches existing `file_already_mined()` semantics in `palace.py:313`.
- **Adapter-owned routing** (§2.5) — `detect_room`/`detect_hall` move into the filesystem adapter. Three-tier precedence: CLI flags → config match → adapter fallback.
- **Flat metadata** — matches ChromaDB's `str | int | float | bool` constraint (RFC 001 §1.4). Entity hints go into a `json_string` field alongside the existing flat `entities`. KG triples route to the SQLite knowledge graph via `KnowledgeGraph.add_triple()`, not drawer metadata.
- **Closets stay core-built** (§1.7) — adapters yield drawers only; core runs `build_closet_lines` post-step. Adapters MAY hint via a flat `closet_hints` string. Closes an existing gap where `convo_miner.py` skips closet building entirely.
- **Privacy class** (§6) — declarable per adapter, enforced per-palace via `privacy_floor`. Default floor: none (single-user frictionless); enterprise deployments set explicitly.
- **No per-drawer field renames** — §5.1 preserves every existing field (`source_file`, `filed_at`, `source_mtime`, `added_by`, `normalize_version`, `entities`, `ingest_mode`). Spec adds three: `adapter_name`, `adapter_version`, `privacy_class`. Existing queries keep working.

## Impact on in-flight PRs

Each of #274, #23, #169, #232, #567/#98, #702 is called out in §10 with the specific alignment work required. #567 (git-mine) is closest to what the spec envisions; formally it becomes the reference first-party adapter for structured extraction. #981 (path-level descriptions) is absorbed as the `metadata_only` ingest mode. #591/#592 (Delphi Oracle, live-stream) deferred to v1.1.

## Gating cleanup (not in this PR)

§9 enumerates the cleanup PR that must land before enforcement:

- `mempalace/sources/` module with `BaseSourceAdapter`, typed records, registry
- `mempalace/sources/transforms.py` with reference implementations of every reserved transformation
- `miner.py` → `mempalace/sources/filesystem.py` (behavior preserved, `READABLE_EXTENSIONS` + `detect_room`/`detect_hall` move with it)
- `convo_miner.py` + `normalize.py` → `mempalace/sources/conversations.py` (format detection becomes per-format plugins, eliminating the `if source_type` chain)
- `PalaceContext` facade exposed by `palace.py` so adapters do not import palace internals
- `KnowledgeGraph.add_triple()` gains optional `source_drawer_id` + `adapter_name` params (backwards-compatible)
- `--mode {projects,convos}` becomes a deprecated alias for `--source {filesystem,conversations}`

## Test plan

- [ ] In-flight adapter authors review for show-stoppers against their work: @Perseusxrltd (#274), @JakobSachs (#23), @adv3nt3 (#169), @zendesk-thittesdorf (#567), @mfhens (#702), @MrDys (#98), @roip (#981)
- [ ] Agreement on the seven open questions in §11 (cross-adapter dedup, live-stream deferral, LLM-assisted extraction, adapter-vs-format split, default privacy_floor, `canonical_source_bytes` for API-backed adapters, `adapter_version` bump semantics)
- [ ] Declared-transformation mapping for `convo_miner.py` + `normalize.py` confirmed by a code read — the §1.4 existing-code table is my read; second pair of eyes welcome
- [ ] Spec-ready before the cleanup PR lands

### DIFF
--- a/docs/rfcs/002-source-adapter-plugin-spec.md
+++ b/docs/rfcs/002-source-adapter-plugin-spec.md
@@ -1,0 +1,768 @@
+# RFC 002 — Source Adapter Plugin Specification
+
+- **Status:** Draft
+- **Tracking issue:** [#989](https://github.com/MemPalace/mempalace/issues/989)
+- **Related:** [#274](https://github.com/MemPalace/mempalace/issues/274), [#23](https://github.com/MemPalace/mempalace/pull/23), [#169](https://github.com/MemPalace/mempalace/pull/169), [#232](https://github.com/MemPalace/mempalace/pull/232), [#567](https://github.com/MemPalace/mempalace/pull/567), [#98](https://github.com/MemPalace/mempalace/pull/98), [#591](https://github.com/MemPalace/mempalace/pull/591), [#592](https://github.com/MemPalace/mempalace/pull/592), [#702](https://github.com/MemPalace/mempalace/pull/702), [#981](https://github.com/MemPalace/mempalace/issues/981), [#244](https://github.com/MemPalace/mempalace/pull/244), [#419](https://github.com/MemPalace/mempalace/pull/419), [#300](https://github.com/MemPalace/mempalace/pull/300), [#952](https://github.com/MemPalace/mempalace/pull/952), [#389](https://github.com/MemPalace/mempalace/pull/389), [#434](https://github.com/MemPalace/mempalace/pull/434)
+- **Sibling spec:** [RFC 001 — Storage Backend Plugin Specification](001-storage-backend-plugin-spec.md)
+- **Spec version:** `1.0`
+
+## Summary
+
+A formal contract for MemPalace source adapters so third parties can ship `pip install mempalace-source-<name>` packages (Cursor, OpenCode, git, Slack, Notion, email, calendar, Whisper transcripts, …) that drop into `mempalace mine` without patching core. The spec defines the adapter interface, record shape, metadata schema contract, privacy class, entry-point registration, incremental-ingest semantics, closet integration, a declared-transformation model that replaces the informal "verbatim" promise with a verifiable one, conformance tests, and the refactor of the existing file and conversation miners into first-party adapters on the same contract.
+
+RFC 001 formalized the write side (where drawers are stored). This RFC formalizes the read side (where content comes from). Both are required for MemPalace to function as a durable daemon managing heterogeneous palaces across many source types.
+
+## Motivation
+
+Six source ingesters are currently in flight, each solving the same problem a different way:
+
+| PR / Issue | Source | Mechanism |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


roip:
Thanks for the thorough spec, Ben. 
LGTM, in #981 my requirements were simple: the system should allow meta data level granularity instead of full content, formalize declarative ignore away from .gitignore. Support path matching patterns for these feature. 
 
Two things to surface for the spec:

level: ignore alongside metadata_only: in PR #986 I implemented three tiers — ignore (skip entirely), describe (one drawer), full (normal). The ignore tier is useful for lock files, node_modules artifacts, etc. In adapter terms: adapter yields zero items for that path. Worth making explicit in the spec so adapters have a clean "I've seen this file, deliberately producing nothing" signal vs "I failed to process it." the overlap with .gitignore is not complete and hence this is needed. 
If the intent is to use an ignore adapter, recommending to ship it with the feature as built in. 

Config-driven routing: Our paths: section in mempalace.yaml uses fnmatch glob patterns with per-pattern level + description + optional room override. This fits §2.5's "config match" tier. a proposed schema is available in PR #986.

PR #991, opened 2026-04-18
fix: set hnsw:num_threads to 1 for thread-safe HNSW inserts

bobo-xxx:
Fixes SIGSEGV in HNSW parallel inserts by limiting HNSW to single-threaded inserts.

## Summary
When multiple threads insert in parallel to an HNSW index, the Rust compactor crashes with SIGSEGV. This fix adds `hnsw:num_threads: 1` to the collection metadata, ensuring thread-safe HNSW operations.

## Changes
- `mempalace/backends/chroma.py`: Added `"hnsw:num_threads": "1"` to metadata in:
  - `get_or_create_collection()` - line 129
  - `create_collection()` - line 150

## Testing
This fix prevents the SIGSEGV crash when running parallel insert workloads against the HNSW index.

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -126,7 +126,7 @@ class ChromaBackend:
         client = self._client(palace_path)
         if create:
             collection = client.get_or_create_collection(
-                collection_name, metadata={"hnsw:space": "cosine"}
+                collection_name, metadata={"hnsw:space": "cosine", "hnsw:num_threads": "1"}
             )
         else:
             collection = client.get_collection(collection_name)
@@ -147,6 +147,6 @@ class ChromaBackend:
     ) -> "ChromaCollection":
         """Create (not get-or-create) *collection_name* with cosine HNSW space."""
         collection = self._client(palace_path).create_collection(
-            collection_name, metadata={"hnsw:space": hnsw_space}
+            collection_name, metadata={"hnsw:space": hnsw_space, "hnsw:num_threads": "1"}
         )
         return ChromaCollection(collection)


mvalentsev:
Same `hnsw:num_threads=1` change on the same file is already in #976 (felipetruman, 2026-04-17, closes #974 / #965 / #955). #976 pins the flag at three sites (`get_or_create_collection`, `create_collection`, `mcp_server._get_collection`) and adds a best-effort retrofit `_pin_hnsw_threads` for pre-existing palaces via `collection.modify(configuration=UpdateCollectionConfiguration(hnsw=UpdateHNSWConfiguration(num_threads=1)))`. Third Copilot review round has been addressed.

Separately, #346 (yoshi280) attributes the sparse `link_lists.bin` bloat in #965 to `persistDirty()` seek drift on the default `batch_size=100` / `sync_threshold=1000`, and raises both thresholds on collection creation. Different root cause from the thread race, orthogonal mitigation. Both paths could land together to cover the full failure surface.

This PR's chroma.py hunk is a narrow subset of #976.

bensig:
### Code review

Found 2 issues:

1. `hnsw:num_threads` is set to the string `"1"`, but ChromaDB validates it as `int`. Collection creation raises on both supported chromadb versions, so this change turns a SIGSEGV into an immediate hard failure on every fresh palace. Empirically reproduced: chromadb 0.6.3 → `ValueError: Invalid value for HNSW parameter: hnsw:num_threads = 1`; chromadb 1.5.8 → `InvalidArgumentError: Failed to parse hnsw parameters from segment metadata`. Issue #974's own diff uses int `1`, and the parallel PR #976 uses int `1` at the same sites. Must be `1`, not `"1"`.

https://github.com/MemPalace/mempalace/blob/19ad6dfeeb1d48d9c37470299f40c04e8d2807ad/mempalace/backends/chroma.py#L127-L132

https://github.com/MemPalace/mempalace/blob/19ad6dfeeb1d48d9c37470299f40c04e8d2807ad/mempalace/backends/chroma.py#L145-L152

2. The primary crash surface from #974 is the MCP server's `_get_collection` in `mempalace/mcp_server.py`, which bypasses `ChromaBackend` and calls `client.get_or_create_collection` directly with only `{"hnsw:space": "cosine"}` — it is untouched by this PR. A user running the MCP server will still hit the SIGSEGV even after merging, because every MCP read/write path goes through this call. PR #976 pins the flag here as well (Copilot review round 1, item #6) after the same hole was flagged; this PR leaves it open, so it does not actually close #974 for the reporter's reproducer (MCP `add_drawer` / auto-mining).

https://github.com/MemPalace/mempalace/blob/19ad6dfeeb1d48d9c37470299f40c04e8d2807ad/mempalace/mcp_server.py#L214-L233

igorls:
Closing as superseded by #976, which just merged into develop and includes the same `hnsw:num_threads=1` change at both `get_or_create_collection()` and `create_collection()` call sites in `mempalace/backends/chroma.py` — plus a `_pin_hnsw_threads()` retrofit that re-applies the pin on existing legacy collections that were created before the fix landed.

Thanks for filing this — your PR was the clearest articulation of the SIGSEGV root cause, and the diff in #976 is a strict superset of yours. If you have palaces created on earlier MemPalace versions, the retrofit will pick them up automatically on the next `get_collection` call.

Issue #195, opened 2026-04-08
IndexError when ChromaDB query returns empty results


wanikua:
## Bug

`searcher.search()`, `searcher.search_memories()`, and `Layer3.search()` all access `results["documents"][0]` without checking if the outer list is empty first.

When a query matches zero documents — e.g. a freshly initialized palace before any mining, or a wing/room filter that excludes everything — ChromaDB returns `{"documents": [], ...}` and the `[0]` indexing crashes with `IndexError`.

### Affected locations

| File | Function | Line |
|------|----------|------|
| `searcher.py` | `search()` | ~58 |
| `searcher.py` | `search_memories()` | ~132 |
| `layers.py` | `Layer3.search()` | ~289 |

### Steps to reproduce

```python
import chromadb

client = chromadb.PersistentClient(path="/tmp/test_palace")
col = client.get_or_create_collection("mempalace_drawers")

# Query an empty collection
results = col.query(query_texts=["test"], n_results=5, include=["documents", "metadatas", "distances"])
print(results["documents"])  # [] — empty list
print(results["documents"][0])  # IndexError: list index out of range
```

### Expected behavior

Should return "No results found" gracefully instead of crashing.

### Fix

Guard the outer list before indexing: `if not results["documents"] or not results["documents"][0]`.

PR #995, opened 2026-04-18
refactor(backends): RFC 001 §10 cleanup — typed results, PalaceRef, registry

igorls:
## Summary

Advances the RFC 001 §10 cleanup prerequisite so the six in-flight backend PRs (#574 LanceDB, #665 Postgres, #700 Qdrant, #697 hosted Chroma, #643 PalaceStore, #381 Qdrant) have a stable target to align against. Resolves @skuznetsov's open coordination question on #743 — the spec surface this PR lands is what #665 can rebase onto.

Sibling spec [RFC 002 #990](https://github.com/MemPalace/mempalace/pull/990) will get its parallel `mempalace/sources/` cleanup in a follow-up PR once this lands.

## What's in this PR

- **Typed results replace the Chroma dict shape at the `BaseCollection` boundary** (§1.3). `QueryResult` and `GetResult` are frozen dataclasses with the outer/inner list semantics the spec prescribes. `QueryResult.empty(num_queries=N)` constructs the spec-correct empty shape.
- **Transitional `_DictCompatMixin`** on both result types so `result["ids"]` / `result.get("documents")` keep working at roughly 60 existing call sites while the attribute-access migration proceeds file-by-file in follow-ups. Scheduled for removal once all callers migrate.
- **Kwargs-only `BaseCollection`** across `add/upsert/query/get/delete/update/count` (§1.1). ABC defaults for `estimated_count`, `close`, `health`, plus a non-atomic default `update()` that does get+merge+upsert (§1.2). Backends advertising `supports_update` override.
- **`PalaceRef` replaces raw path strings** at the backend boundary (§2.2). `id` is always present (cache key), `local_path` for filesystem-rooted palaces, `namespace` for server-mode tenant routing.
- **`BaseBackend` ABC** with `get_collection(palace=PalaceRef, …)`, `close_palace`, `close`, `health`, and an optional `detect(path) -> bool` classmethod for the migration-path auto-detection rule (§3.3 (4)).
- **Entry-point registry**: `[project.entry-points."mempalace.backends"]` wired in `pyproject.toml`, discovery via `importlib.metadata.entry_points(group="mempalace.backends")`, explicit `register()` for tests wins on conflict (§3.1–3.2). `resolve_backend_for_palace()` implements the §3.3 priority order.
- **`ChromaCollection` normalizes** chroma's dict returns into typed results with spec-correct empty-outer-shape preservation (fixes issue #195 at the type layer). Unknown where-clause operators raise `UnsupportedFilterError` — silent dropping is forbidden (§1.4).
- **`ChromaBackend` absorbs the inode/mtime client-cache freshness check** previously duplicated in [mcp_server._get_client()](mempalace/mcp_server.py#L162) (§10 + PR #757). `close_palace(ref)` now flushes a single palace's handles.
- **`searcher.py` migrated to attribute access** as the reference call site; `_first_or_empty` is polymorphic over `QueryResult` and the legacy dict shape so test mocks and mid-migration callers keep passing.

## Explicitly out of scope (follow-up PRs)

These are called out in the spec and deferred so this PR stays reviewable:

- Full caller migration off the dict-compat shim: `palace.py`, `mcp_server.py`, `miner.py`, `convo_miner.py`, `dedup.py`, `repair.py`, `exporter.py`, `palace_graph.py`, `cli.py`, `closet_llm.py`.
- Embedder injection + three-state `EmbedderIdentityMismatchError` check (§1.5).
- `maintenance_state()` / `run_maintenance()` benchmark hooks (§7.3).
- Full `AbstractBackendContractSuite` conformance coverage (§7.1–7.2).
- `mempalace migrate` / `mempalace verify` rewrites through `BaseCollection` (§8).
- `mcp_server._get_client()` removal (its `_metadata_cache` invalidation is tangled with the client cache in ways worth disentangling separately).

## Test plan

- [x] `uv run python -m pytest tests/ --ignore=tests/benchmarks` — **970 passed** (up from 967 on develop).
- [x] `uv run ruff check .` — clean.
- [x] `uv run ruff format --check .` — clean.
- [x] New test coverage in `tests/test_backends.py`:
  - Typed `QueryResult` / `GetResult` return from `ChromaCollection`
  - Empty-result outer-shape preservation (`QueryResult.empty(num_queries=2)`)
  - Dict-compat shim (`result["ids"]`, `result.get(...)`, `in` operator)
  - `UnsupportedFilterError` on unknown where operator (`$regex`)
  - Registry lookup and unknown-backend `KeyError`
  - `resolve_backend_for_palace` priority order
  - `ChromaBackend.detect()`
  - `PalaceRef` kwarg path on `ChromaBackend.get_collection`

## Coordination

Refs #743 (RFC 001), #990 (RFC 002), #989 (RFC 002 tracking issue).

cc @skuznetsov @dekoza @cschnatz @jphein — this is the §10 spec surface I mentioned. #665 / #574 / #697 can now rebase against these interfaces. Happy to help review the alignment deltas.

### DIFF
--- a/mempalace/backends/__init__.py
+++ b/mempalace/backends/__init__.py
@@ -1,6 +1,64 @@
-"""Storage backend implementations for MemPalace."""
+"""Storage backend implementations for MemPalace (RFC 001).
 
-from .base import BaseCollection
+Public surface:
+
+* :class:`BaseCollection` — per-collection read/write contract.
+* :class:`BaseBackend` — per-palace factory contract.
+* :class:`PalaceRef` — value object identifying a palace for a backend.
+* :class:`QueryResult` / :class:`GetResult` — typed read returns.
+* Error classes: :class:`PalaceNotFoundError`, :class:`BackendClosedError`,
+  :class:`UnsupportedFilterError`, :class:`DimensionMismatchError`,
+  :class:`EmbedderIdentityMismatchError`.
+* Registry: :func:`get_backend`, :func:`register`, :func:`available_backends`,
+  :func:`resolve_backend_for_palace`.
+* In-tree Chroma default: :class:`ChromaBackend`, :class:`ChromaCollection`.
+"""
+
+from .base import (
+    BackendClosedError,
+    BackendError,
+    BaseBackend,
+    BaseCollection,
+    DimensionMismatchError,
+    EmbedderIdentityMismatchError,
+    GetResult,
+    HealthStatus,
+    PalaceNotFoundError,
+    PalaceRef,
+    QueryResult,
+    UnsupportedFilterError,
+)
 from .chroma import ChromaBackend, ChromaCollection
+from .registry import (
+    available_backends,
+    get_backend,
+    get_backend_class,
+    register,
+    reset_backends,
+    resolve_backend_for_palace,
+    unregister,
+)
 
-__all__ = ["BaseCollection", "ChromaBackend", "ChromaCollection"]
+__all__ = [
+    "BackendClosedError",
+    "BackendError",
+    "BaseBackend",
+    "BaseCollection",
+    "ChromaBackend",
+    "ChromaCollection",
+    "DimensionMismatchError",
+    "EmbedderIdentityMismatchError",
+    "GetResult",
+    "HealthStatus",
+    "PalaceNotFoundError",
+    "PalaceRef",
+    "QueryResult",
+    "UnsupportedFilterError",
+    "available_backends",
+    "get_backend",
+    "get_backend_class",
+    "register",
+    "reset_backends",
+    "resolve_backend_for_palace",
+    "unregister",
+]
--- a/mempalace/backends/base.py
+++ b/mempalace/backends/base.py
@@ -1,49 +1,354 @@
-"""Abstract collection interface for MemPalace storage backends."""
+"""Storage backend contract for MemPalace (RFC 001).
+
+This module defines the surface every storage backend must implement:
+
+* ``BaseCollection`` — the per-collection read/write interface, kwargs-only.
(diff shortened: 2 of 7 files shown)


igorls:
Thanks @copilot — all four review items addressed in 42b940d. Recap:

1. **`BaseCollection.update()` default — length validation** ([base.py](mempalace/backends/base.py)). `documents` / `metadatas` / `embeddings` lengths are now checked against `ids` up front; mismatched lists raise `ValueError` with a clear "{field} length N does not match ids length M" message instead of `IndexError` or silent misalignment.
2. **`ChromaCollection.query()` — input validation** ([chroma.py](mempalace/backends/chroma.py)). Exactly one of `query_texts` / `query_embeddings` is required; both-provided and empty-list cases now raise `ValueError` before delegating to chromadb (whose own errors are less obvious).
3. **`QueryResult.empty()` — embeddings outer shape when requested** ([base.py](mempalace/backends/base.py)). Added `embeddings_requested: bool = False` parameter per your suggestion. When requested, `embeddings` preserves the outer query dimension with empty inner lists; otherwise stays `None`. `ChromaCollection.query()` threads this through on the empty-result path.
4. **`ChromaBackend._client` — cache freshness edge cases** ([chroma.py](mempalace/backends/chroma.py)). Now matches `mcp_server._get_client` semantics (merged via #757) on three cases: (a) DB file disappears while a cached client is held → invalidate; (b) 0 → nonzero stat transition (DB created after cache) → rebuild; (c) re-stat *after* `PersistentClient` constructs the DB lazily so freshness reflects the post-creation state, not the pre-creation `(0, 0.0)`.

Tests: 978 passed (up from 970), +8 targeted tests. Lint + format clean. CI is running — ping when green if you want me to mark ready.

PR #999, opened 2026-04-18
fix(searcher): guard against None metadata in CLI print path

jphein:
## Summary

`col.query(...)` can return `None` entries in the inner `metadatas` list for drawers whose metadata was never set — older palaces, or rows written outside the normal mining path. When that happens, the CLI `search()` function renders earlier results successfully, then crashes mid-loop with:

```
```

at `searcher.py:286` (`meta.get("source_file", "?")`). The user sees partial output followed by a traceback, with no indication of which drawers rendered and which were skipped.

## Fix

One-line guard inside the loop: `meta = meta or {}` so entries with missing metadata fall back to the existing `"?"` defaults. Matches the way `search_memories()` assembles hit dicts against the same data (`meta.get("wing", "unknown")` etc.).

## Test

Added a regression test (`test_search_handles_none_metadata_without_crash`) that mocks a ChromaDB result with a `None` entry in the middle of the inner metadatas list and asserts both result blocks render. Verified the test fails without the guard (reproduces the crash) and passes with it.

## Scope

Independent of — but complementary to — #951 which handles a different search failure mode (ChromaDB filter planner returning `Error finding id` on migrated palaces). #951 wraps the query call; this one wraps the result rendering.

## Testing

- `pytest tests/test_searcher.py` — 20 passed (up from 19)
- Regression-test verification: temp-reverted the one-line guard, ran just the new test, confirmed `AttributeError`
- `ruff check` / `ruff format --check` — clean

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -315,6 +315,7 @@ def tool_status():
     try:
         all_meta = _get_cached_metadata(col)
         for m in all_meta:
+            m = m or {}
             w = m.get("wing", "unknown")
             r = m.get("room", "unknown")
             wings[w] = wings.get(w, 0) + 1
@@ -368,6 +369,7 @@ def tool_list_wings():
     try:
         all_meta = _get_cached_metadata(col)
         for m in all_meta:
+            m = m or {}
             w = m.get("wing", "unknown")
             wings[w] = wings.get(w, 0) + 1
     except Exception as e:
@@ -391,6 +393,7 @@ def tool_list_rooms(wing: str = None):
         where = {"wing": wing} if wing else None
         all_meta = _fetch_all_metadata(col, where=where)
         for m in all_meta:
+            m = m or {}
             r = m.get("room", "unknown")
             rooms[r] = rooms.get(r, 0) + 1
     except Exception as e:
@@ -409,6 +412,7 @@ def tool_get_taxonomy():
     try:
         all_meta = _get_cached_metadata(col)
         for m in all_meta:
+            m = m or {}
             w = m.get("wing", "unknown")
             r = m.get("room", "unknown")
             if w not in taxonomy:
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -854,6 +854,7 @@ def status(palace_path: str):
 
     wing_rooms = defaultdict(lambda: defaultdict(int))
     for m in metas:
+        m = m or {}
         wing_rooms[m.get("wing", "?")][m.get("room", "?")] += 1
 
     print(f"\n{'=' * 55}")
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -283,6 +283,7 @@ def search(query: str, palace_path: str, wing: str = None, room: str = None, n_r
 
     for i, (doc, meta, dist) in enumerate(zip(docs, metas, dists), 1):
         similarity = round(max(0.0, 1 - dist), 3)
+        meta = meta or {}
         source = Path(meta.get("source_file", "?")).name
         wing_name = meta.get("wing", "?")
         room_name = meta.get("room", "?")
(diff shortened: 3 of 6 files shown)


jphein:
Expanded scope in feba7e8 — same `AttributeError: 'NoneType' object has no attribute 'get'` hit again today in `miner.status()` on a production palace with 2 `None`-metadata drawers. Added the matching `m = m or {}` guard at the histogram loop and a regression test that feeds status a fake collection whose `get()` returns a `None` in the metadatas list.

Both spots are the same class of bug — ChromaDB can surface `None` entries in the `metadatas` inner list for drawers whose metadata was never stored, and the two read paths (searcher print, status histogram) both ran `.get()` on the entry unconditionally. Happy to split into two PRs if you'd prefer one-concern-per-PR.

jphein:
Thanks @copilot-pull-request-reviewer — good catch. Addressed in 7690574: added the same `meta = meta or {}` / `cmeta = cmeta or {}` guards to the two call sites in `search_memories()` (the closet boost lookup at the `cmeta.get` line and the drawer scoring loop at the `meta.get` line). Added an API-path regression test (`test_search_memories_handles_none_metadata`) that mocks a drawer result with a `None` metadata entry and asserts both hits render, with the None-metadata hit using the existing `"unknown"` sentinels. Verified it fails without the guard.

jphein:
Heads-up on scope drift — this PR now covers 8 spots across 3 files:

- `mempalace/searcher.py` — CLI print loop + `search_memories()` closet-boost loop + drawer-scoring loop
- `mempalace/miner.py` — `status()` wing/room histogram loop
- `mempalace/mcp_server.py` — `tool_status`, `tool_list_wings`, `tool_list_rooms`, `tool_get_taxonomy`

All the same `meta = meta or {}` one-liner, all catching ChromaDB `get()` / `query()` returning `None` entries in the inner `metadatas` list for drawers with no stored metadata.

Whack-a-mole across 8 call sites suggests the cleaner fix probably lives one layer down — either:

1. `_get_cached_metadata` / `_fetch_all_metadata` (in `mcp_server.py`) strip `None` entries as they build the cache, OR
2. `ChromaCollection.get()` / `ChromaCollection.query()` in `backends/chroma.py` coerce `None` metadatas to `{}` at the adapter boundary.

Option 2 is more general — every future call site would get the guard for free, and it matches @bensig's observation in #1013 that `Layer3.search_raw` also needs this (a fourth file we don't touch here). Option 1 only covers the two cache helpers but keeps the backend adapter faithful to what Chroma actually returns.

Happy to convert this to the adapter-level fix if maintainers prefer. Current state (per-site guards) is defensive and ships today; the adapter-level consolidation would be a follow-up that deletes all the individual guards.

jphein:
Surfacing a parallel effort for maintainer awareness: @cantenesse's #1019 just filed the same `meta or {}` / `doc or ""` guard pattern for `searcher.py`'s three loops. Their scope is narrower (just `searcher.py`); this PR additionally covers `miner.status()` and four `mcp_server.py` handlers (`tool_status`, `tool_list_wings`, `tool_list_rooms`, `tool_get_taxonomy`) that hit the same class of bug. If #1019 merges first I'm happy to rebase this down to just the miner + mcp_server sites.

Dialectician (commented):
Read the diff. Defensive `meta or {}` coercion pattern applied at each of the call sites that could crash on a `None` metadata entry, and the new tests cover each guarded path (`tool_status`, miner.status, searcher, and the MCP list handlers per #1005-era conversation). Additive only, no deletions, no behavior change for the happy path.

The failure mode this guards against is real (ChromaDB `query()` / `get()` returning `None` for drawers with no stored metadata, crashing mid-tally on `AttributeError`) and the class-of-fix matches what is running in production at 137K drawers in `jphein/mempalace`. Straightforward and well-scoped.

One observation for a follow-up, not for this PR: 8 guard sites is a strong signal that adapter-level coercion (`ChromaCollection.get/query` returning `meta or {}`) would be more DRY than whack-a-mole at call sites. You flagged this yourself in the fork CLAUDE.md as a proposed follow-up — agreed.

Peace B.Sweet

PR #1000, opened 2026-04-18
feat(backends): quarantine_stale_hnsw — recover from HNSW/sqlite drift (closes #823)

### LINKED ISSUE #823: ChromaDB: HNSW persistence vs add_drawer — default sync_threshold can leave semantic search stale
## Summary

When MemPalace adds drawers via `add_drawer` (and similar paths), ChromaDB may persist **metadata** to `chroma.sqlite3` while the **HNSW vector index** on disk (`data_level0.bin`, etc.) lags behind or never reaches parity with the metadata segment under default `hnsw:sync_threshold` (often **1000**). Users then observe:

- `collection.get()` / listing shows new documents
- `collection.query()` / semantic search returns **no** or **stale** matches for those documents (e.g. similarity `0.0` or unrelated hits)
- On-disk HNSW file timestamps stay old while `chroma.sqlite3` updates

This is consistent with ChromaDB’s design: vectors are buffered and the on-disk HNSW segment is flushed only after enough **new** vectors accumulate (see [Chroma configuration](https://cookbook.chromadb.dev/core/configuration/) — `sync_threshold`, `batch_size`).

## Expected behavior

After a successful `add_drawer`, **semantic search** should reliably find the new content without requiring manual maintenance, or MemPalace should **document** that operators must run `chromadb-ops` / adjust HNSW settings.

## Suggested fixes (any one or combination)

1. **Set a lower default `sync_threshold`** (e.g. **100**) when creating or opening the `mempalace_drawers` collection, so HNSW is flushed to disk more often for typical incremental adds.
2. **Expose optional config** (env var or `config.json`) for `hnsw:sync_threshold` / `hnsw:batch_size` so power users can tune persistence vs performance.
3. After bulk adds, optionally call into Chroma’s maintenance path (e.g. WAL commit / HNSW rebuild via supported APIs) where applicable for the pinned `chromadb` version.
4. **Document** the interaction between SQLite metadata and HNSW persistence, and point to `chromadb-ops` (`chops db info`, `chops hnsw rebuild`) for recovery if metadata and vector segments diverge.

## Environment (example)

- ChromaDB **0.6.x** / **1.x** (depending on MemPalace release)
- `PersistentClient` against `~/.mempalace/palace/`
- Windows + Linux (devcontainer bind mount) — issue reproduced when metadata updated but HNSW files were stale

## Additional context

We recovered by exporting all drawers, recreating the collection with `sync_threshold=100`, and re-importing. A first-class MemPalace-side default or documented knob would prevent that class of failure.

Thank you for MemPalace — happy to help test a PR if useful.

jphein:
## Summary

Adds a `quarantine_stale_hnsw(palace_path, stale_seconds=3600)` helper that renames HNSW segment directories whose `data_level0.bin` is significantly older than `chroma.sqlite3`. Drift between the on-disk HNSW graph and the live `embeddings` table in sqlite is the root cause of the bug #823 describes — and, under heavier load or after an interrupted write, the same drift escalates from "silent stale results" to a SIGSEGV where the Rust graph-walk dereferences dangling neighbor pointers for entries that exist in the metadata segment but not in the HNSW index.

Closes #823.

## The drift, in data

On one 135K-drawer palace where the bug reproduced, `index_metadata.pickle` claimed `total_elements_added: 137813` against `SELECT COUNT(*) FROM embeddings = 135464` — a 2,349-entry drift. Fresh Python processes crashed in `col.count()` on 17/20 attempts:

```python
import chromadb
c = chromadb.PersistentClient(path=\"/palace\")
col = c.get_collection(\"mempalace_drawers\")
col.count()  # SIGSEGV, background thread, chromadb_rust_bindings.abi3.so
```

After renaming the segment dir out of the way and letting ChromaDB rebuild lazily, the same 20-run check went to **0 crashes** and queries returned reasonable distances against the same underlying drawers.

Same failure mode is tracked publicly at [neo-cortex-mcp#2](https://github.com/LadislavSopko/neo-cortex-mcp/issues/2) (chromadb 1.5.5, Python 3.12, SIGSEGV on `.count()` with the deep-recursion stack signature) and acknowledged at [chroma-core/chroma#2594](https://github.com/chroma-core/chroma/issues/2594) (HNSW drift is by-design under default `sync_threshold`).

## Why not #823's suggested recovery

#823 suggests \"export all drawers, recreate the collection with `sync_threshold=100`, re-import.\" That works, but it re-embeds every drawer — expensive on a 100K+ palace and requires the collection to be readable in the first place, which a drifted palace sometimes isn't. This helper is lighter:

- Rename the drifted segment dir so ChromaDB reopens without it.
- ChromaDB writes a fresh minimal segment on next use; the underlying embeddings in `chroma.sqlite3` are untouched.
- The renamed directory is **not deleted** — it's moved to `<uuid>.drift-YYYYMMDD-HHMMSS` so operators can recover if the heuristic misfires.

## Heuristic

If `chroma.sqlite3` is more than `stale_seconds` (default **3600**) newer than the segment's `data_level0.bin`, the segment is considered suspect. One hour is deliberately conservative — normal HNSW flush cadence is seconds to minutes, so an hour of drift implies a crashed mid-write, not routine lag.

## Scope

- **Additive only.** Exposes `quarantine_stale_hnsw()` as a helper. Intentionally not wired into `_client()` / startup on this PR — the goal is to land the primitive first so operators and higher layers can opt in. A natural follow-up is to call it automatically on palace open behind an env var or a config flag; happy to do that in a second PR once this shape is agreed.
- No conflict with #976 (INSERT-path HNSW race, `hnsw:num_threads=1`) or #951 (graceful fallback when filtered queries fail) — those are sibling failure modes. This PR addresses the READ-path crash on drifted segments, which neither of them touches.
- Relates to #823 as the recovery primitive. Tuning `sync_threshold` (the other half of #823) can still be done separately and remains a reasonable addition.

## Tests

Four new tests in `tests/test_backends.py`:
- `test_quarantine_stale_hnsw_renames_drifted_segment` — 2h drift: renamed, original files preserved under `.drift-TS` suffix.
- `test_quarantine_stale_hnsw_leaves_fresh_segment_alone` — 10s drift: not touched.
- `test_quarantine_stale_hnsw_no_palace` — missing palace / missing `chroma.sqlite3`: returns `[]`, no raise.
- `test_quarantine_stale_hnsw_skips_already_quarantined` — directories already carrying the `.drift-` suffix are never re-renamed.

`pytest tests/test_backends.py` → **11 passed**. `ruff check` / `ruff format --check` — clean.

## Credit

Diagnosis and fix worked out in [jphein/mempalace@fdfacdf](https://github.com/jphein/mempalace/commit/fdfacdf) while chasing the read-path SIGSEGV on a production palace. The mtime heuristic is a deliberate compromise (no `pickle.load` on `index_metadata.pickle` — linting policy in the downstream didn't allow it); happy to make it authoritative (read `total_elements_added` vs sqlite count) on the upstream path if you'd prefer.

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -1,5 +1,6 @@
 """ChromaDB-backed MemPalace storage backend (RFC 001 reference implementation)."""
 
+import datetime as _dt
 import logging
 import os
 import sqlite3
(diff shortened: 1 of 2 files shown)


jphein:
Thanks @copilot-pull-request-reviewer — addressed in c6413fd: tightened the signature to `-> list[str]` and the local to `moved: list[str] = []` to match PEP 585 usage elsewhere in the repo (e.g. `exporter.py`, `sweeper.py`).

jphein:
Rebased onto `develop` after #995 merged (055386e). Integration was clean — RFC 001's `_validate_where` + `_fix_blob_seq_ids` kept from upstream, `quarantine_stale_hnsw()` added alongside with `list[str]` type hints matching the new typed-results style. Dropped my earlier one-line style commit since the type hint was already right in the rebased placement. Tests still pass (29/29 on `test_backends.py`).

Dialectician (commented):
Read the diff. Logic is right: compare `data_level0.bin` mtime to `chroma.sqlite3` mtime, rename the segment directory if the gap exceeds threshold. Rename-not-delete is the right call so recovery is reversible by hand if the heuristic ever misfires. Default `stale_seconds=3600.0` is conservative enough that a quick write-then-read cycle will not misclassify a fresh segment.

Tests cover the two main branches (drifted segment renamed, fresh segment untouched) with clean fixtures. Closes #823 credibly.

Two small questions, both follow-ups not blockers:

1. Multi-segment palaces: the loop iterates all subdirs with `data_level0.bin`, but if two segments drift, the second rename could collide with a name from the first. Probably safe because `quarantine-<ts>` is unique per call, worth a quick verification.
2. Logging: `logger.exception` on rename failure is right. Might also log the sqlite/hnsw mtime delta when quarantining, so postmortem grep tells you how stale the segment was.

Peace B.Sweet

jphein:
Addressed Copilot type hint nit: `-> list` → `-> list[str]` and `moved: list = []` → `moved: list[str] = []` to match the repo's use of built-in generics elsewhere.

jphein:
Thanks for reading the diff carefully — the rename-not-delete design was exactly the choice I sweated over most, so good to have it independently confirmed. The 1h default being conservative-by-design is also appreciated; a false positive here (quarantining a healthy segment) would be annoying to recover from, so erring cautious is the right call.

Fixed the `list[str]` nit.

bensig:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Issue #487, opened 2026-04-10
chromadb<0.7 pin breaks MCP server on Python 3.14


director-LAC:
## Summary

`mempalace 3.1.0` pins `chromadb>=0.5.0,<0.7`, but chromadb 0.6.x uses Pydantic V1 internally, which is incompatible with Python 3.14. This causes the MCP server to crash on startup.

## Environment

- Python 3.14.3 (Windows 11)
- mempalace 3.1.0
- Upgraded from mempalace 3.0.0 (which had been working with chromadb 1.5.6)

## Steps to Reproduce

1. Install mempalace 3.1.0 on Python 3.14
2. pip resolves chromadb to 0.6.3 (satisfies `>=0.5.0,<0.7`)
3. Start MCP server: `python -m mempalace.mcp_server`

## Error

File "...\chromadb\config.py", line 21, in <module>
from pydantic.v1 import BaseSettings
...
File "...\chromadb\config.py", line 101, in <module>
class Settings(BaseSettings):
...
File "...\pydantic\v1\main.py", line 221, in new
inferred = ModelField.infer(...)

The root cause is `chromadb 0.6.x` depending on `pydantic.v1`, which is not compatible with Python 3.14+.

## Workaround

Force-install chromadb 1.5.6, which works on Python 3.14:

```bash
pip install chromadb==1.5.6

pip warns about the version conflict (mempalace wants <0.7, you have 1.5.6), but the server starts and runs correctly.

Suggested Fix
Either:

Widen the chromadb pin to allow >=0.5.0 (dropping the <0.7 ceiling), or
Add a Python 3.14+ conditional that requires chromadb>=1.0, or
Document the incompatibility and pin guidance for Python 3.14 users

web3guru888:
We've been tracking ChromaDB version fragmentation as a recurring pain point across the community (#426, #445, #457, #458) — it's probably the single most common friction point in fresh installations right now. The `<0.7` ceiling was added defensively to avoid the 0.7.x API breaking changes, but it inadvertently created a trap for Python 3.14 users: chromadb 0.6.x pulls in pydantic.v1, which simply doesn't work on 3.14.

director-LAC's workaround of force-installing chromadb 1.5.6 is actually solid — 1.x dropped the pydantic.v1 dependency entirely and its API has stabilized. We've been running with a `>=0.6.0,<0.7.0` pin ourselves to avoid the 0.7.x instability, but that's clearly not viable long-term either.

[...]

PR #1008, opened 2026-04-18
fix: pin chromadb to <1.0 to avoid 1.5.x queue-stall regression (#1006)

### LINKED ISSUE #1006: bug: fresh `pip install mempalace` silently broken on chromadb 1.5.x — writes queued but never flush, search returns ghosts
## Severity: critical

Every new user who runs \`pip install mempalace\` today hits this silently. The mine output claims success ("Drawers filed: 231"), but \`mempalace status\` shows 0 user drawers and \`mempalace search\` either returns no results or crashes on \`NoneType\` metadata (separate issue #IS2).

Root cause: \`pyproject.toml\` pins \`chromadb>=0.5.0\` with no upper bound, so fresh installs pull \`chromadb==1.5.8\` (latest) which has a regression where the \`embeddings_queue\` → \`embeddings\` + \`embedding_metadata\` processor never runs.

## Reproduction

```bash
python3.11 -m venv .venv
source .venv/bin/activate
pip install mempalace                          # pulls chromadb 1.5.8 (latest)

mempalace mine ~/.claude/projects/<session-dir> --mode convos --wing test
#   ✓ [   1/6] ...jsonl  +191
#   Done.
#   Drawers filed: 231

mempalace status
#   MemPalace Status — 0 drawers   (or only ancient test fixtures if present)

mempalace search "anything"
#   AttributeError: 'NoneType' object has no attribute 'get'
#   (see sibling issue for searcher.py defensive fix)
```

Reproduced on Python 3.11.11 and 3.14.2 — same result. Not Python-version-dependent.

## Evidence

Direct SQLite probe of \`~/.mempalace/palace/chroma.sqlite3\` after a mine that reported 231 drawers filed:

| Table | Row count |
|---|---|
| \`embeddings\` | 8 |
| \`embeddings_queue\` | 652 |
| \`embedding_metadata\` | 56 |
| \`col.count()\` | 4 |

Every upsert added rows to \`embeddings_queue\` (op=2) but \`embeddings\` / \`embedding_metadata\` never grew. The segment materialization step didn't run — not on \`PersistentClient\` open, not after 10+ seconds of dwell time, not after a probe \`col.modify()\` call, not after an additional upsert.

The HNSW vector index *does* get populated directly (sync path) — so \`col.query()\` returns the correct IDs and distances, but with \`None\` for every \`documents\` and \`metadatas\` entry:

```python
res = col.query(query_texts=['RFC 002 source adapter'], n_results=5,
                include=['documents','metadatas','distances'])

res['ids']        # ['drawer_mempalace_technical_15fc11...', ...]  ← real queued ids
res['distances']  # [0.79, 0.89, 0.92, 1.01, 1.03]                  ← real scores
res['documents']  # [None, None, None, None, None]                  ← never flushed
res['metadatas']  # [None, None, None, None, None]                  ← never flushed
```

\`col.get(ids=<queued-ids>)\` returns empty — they're unreachable via the normal read path.

## Tested versions

| \`chromadb\` | Result |
|---|---|
| 1.5.8 (current default on fresh install) | Broken — queue stalls |
| 1.5.6 | Broken — same stall |
| 0.6.3 | **Works immediately** — 231 drawers visible, search returns real content |

Verified by installing \`chromadb==0.6.3\`, wiping \`~/.mempalace/palace\`, re-running the same mine. Status instantly shows 231 drawers, search returns our conversation content with correct similarity scores.

## Hypothesis on root cause

The 1.x architecture introduced \`segments\` + an async \`embeddings_queue\` as a WAL-like write path. The segment materialization step (queue → embeddings + embedding_metadata tables) isn't running from \`PersistentClient\` alone. Possibly:

- A background-thread startup regressed in the 1.5.x series
- An initialization/compaction call that \`PersistentClient\` no longer triggers
- A separate server/worker process that \`HttpClient\` would start but \`PersistentClient\` doesn't

Not investigated upstream — ChromaDB issue search left for the fix PR.

## Proposed fix

Two-line change in \`pyproject.toml\`:

```diff
- "chromadb>=0.5.0",
+ "chromadb>=0.5.0,<1.0",
```

Pins to the last-known-good major. Existing installs that pulled 1.5.x can recover by \`pip install "chromadb<1.0"\` + re-mine (deterministic IDs, safe upsert).

Long-term: investigate upstream, contribute a fix to ChromaDB, or raise the ceiling once a 1.x version ships with the queue processor restored.

## Related

- #664 — auto-repair BLOB seq_ids from chromadb 0.6→1.5 migration
- #457 — 3.1.0 upgrade silently breaks MCP servers due to chromadb version downgrade (1.5.6 → 0.6.3)
- #426 — chore(deps): pin chromadb to >=1.0,<2 — open PR going the opposite direction; **this bug suggests that pin would make things worse, not better**

## Companion issue

The \`searcher.py\` crash on \`None\` metadata is a separate defensive-coding gap that should be fixed regardless — partial-state results can occur during interrupted mines or schema boundaries even on known-good versions.

bensig:
## Summary

Fixes #1006.

`chromadb>=0.5.0` with no upper bound lets fresh `pip install mempalace` pull the latest `chromadb==1.5.8`, which has a queue-processor regression that makes writes invisible. Mine reports success; `mempalace status` / `search` show 0 drawers. `chromadb==0.6.3` verified working.

## Change

One line in `pyproject.toml`:

```diff
- "chromadb>=0.5.0",
+ "chromadb>=0.5.0,<1.0",
```

## Test plan

- [ ] Fresh venv + `pip install` from this branch — verify chromadb resolves below 1.0
- [ ] `mempalace mine <dir> --mode convos` reports filed drawers
- [ ] `mempalace status` shows the filed drawers (not 0)
- [ ] `mempalace search` returns real content with non-None metadata

## Root cause

The 1.x architecture introduced async segment materialization via `embeddings_queue`. In 1.5.x that processor isn't running from `PersistentClient` alone — writes durably queue but never reach the `embeddings` / `embedding_metadata` tables. SQLite-level evidence in #1006.

Upstream fix not investigated in this PR. Long-term: contribute to ChromaDB or raise the ceiling once a 1.x version ships with the processor restored.

## Note for affected users

Users who already pulled 1.5.x can recover without losing data:

```bash
pip install "chromadb<1.0"
# re-mine — deterministic IDs, safe upsert
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ classifiers = [
     "Topic :: Utilities",
 ]
 dependencies = [
-    "chromadb>=0.5.0",
+    "chromadb>=0.5.0,<1.0",
     "pyyaml>=6.0,<7",
 ]
 


bensig:
### Code review

Found 1 issue:

1. Pinning `chromadb<1.0` forces installation back to the 0.6.x line, which imports from `pydantic.v1` — this fails to import on Python 3.13+ and is documented as a hard regression in [#487](https://github.com/MemPalace/mempalace/issues/487). PR [#302](https://github.com/MemPalace/mempalace/pull/302) was specifically written to widen the chromadb ceiling so that mempalace works on modern Python; this PR reverses that fix. Two related PRs are also going in the opposite direction for the same reason: [#426](https://github.com/MemPalace/mempalace/pull/426) (`>=1.0,<2`) and [#581](https://github.com/MemPalace/mempalace/pull/581) (`>=1.5.4,<2`, already APPROVED with 50K-drawer migration validation per jphein's testing). Consider a narrower workaround for the 1.5.6/1.5.8 queue-stall instead — e.g., excluding the specific broken versions (`!=1.5.6,!=1.5.8`) or raising the floor to a known-good 1.x release — so modern-Python users aren't hard-blocked.

https://github.com/MemPalace/mempalace/blob/8ad57728da85abc1f02477d3ddd61054ccfcaa4d/pyproject.toml#L26-L31

bensig:
Closing in favor of #581 (`fix: upgrade chromadb to >=1.5.4 for python 3.13/3.14 compatibility`).

Per the review on this PR, the `<1.0` cap reintroduces a documented breaking incompatibility on Python 3.13+: chromadb 0.6.x imports from \`pydantic.v1\`, which fails on modern Python — see #487 for the exact ConfigError. This reverses the intent of #302, which deliberately widened the chromadb ceiling to fix the same regression.

#581 is already APPROVED with a 50K-drawer auto-migration validation (per @jphein's testing — 648 tests pass, palace migrates 0.6.x → 1.5.x cleanly). It keeps mempalace on a supported-Python path and plausibly excludes the broken 1.5.x versions via its `>=1.5.4` floor.

Issue #1006 stays open. The queue-stall regression is real; the right fix is a narrow floor in the 1.5.x line, not a blanket cap below 1.0. If #581 merges and 1.5.4 itself exhibits the queue-stall, a follow-up bump to the floor is simpler than capping below the whole major.

PR #1009, opened 2026-04-18
fix: guard searcher loop against None doc/meta from ChromaDB (#1007)

### LINKED ISSUE #1007: bug: `mempalace search` crashes with `AttributeError: 'NoneType' object has no attribute 'get'` on partial-state results (searcher.py:286)
## Problem

\`mempalace search <query>\` raises \`AttributeError: 'NoneType' object has no attribute 'get'\` whenever a ChromaDB query result contains a \`None\` metadata entry. The reader assumes \`meta\` is always a dict; it isn't.

## Traceback

```
  File "mempalace/cli.py", line 681, in main
    dispatch[args.command](args)
  File "mempalace/cli.py", line 153, in cmd_search
    search(query=args.query, ..., n_results=args.results)
  File "mempalace/searcher.py", line 286, in search
    source = Path(meta.get("source_file", "?")).name
             ^^^^^^^^
```

## When this happens

Any ChromaDB state where \`query()\` returns \`[None]\` for one or more results:

- **Partial-flush states** — HNSW vector index has a drawer but \`embedding_metadata\` hasn't been materialized yet. Hit reliably today on \`chromadb==1.5.8\` (see companion issue #1006).
- **Mid-delete** — drawer's metadata rows purged before HNSW entry removed.
- **Upgrade boundaries** — palace written by one schema, read by another.
- **Interrupted mines** — process killed between segment writes.

Even after #1006's chromadb pin lands, these edge cases remain. Defensive reads are cheap.

## Current code

\`mempalace/searcher.py:284-288\`:

```python
    similarity = round(max(0.0, 1 - dist), 3)
    source = Path(meta.get("source_file", "?")).name
    wing_name = meta.get("wing", "?")
    room_name = meta.get("room", "?")
```

Line 295 further down does \`doc.strip()\` which has the same \`None\` vulnerability.

## Proposed fix

```diff
 for i, (doc, meta, dist) in enumerate(zip(docs, metas, dists), 1):
     similarity = round(max(0.0, 1 - dist), 3)
+    meta = meta or {}
+    doc = doc or ""
     source = Path(meta.get("source_file", "?")).name
     wing_name = meta.get("wing", "?")
     room_name = meta.get("room", "?")
```

Two-line defensive guard. Degrades gracefully: source shows \`"?"\`, wing/room show \`"?"\`, the document body prints empty. The hit still appears in results (with distance/index), which is more useful than crashing.

Worth considering: should a \`None\`-metadata result be filtered out entirely rather than shown with placeholders? Argument for showing: distances are real, user may want to see that results exist even if storage is partial. Argument for filtering: empty rows look broken. Either decision is fine — crashing is not.

## Test idea

Add a regression test that mocks a ChromaDB query returning \`{"documents": [[None]], "metadatas": [[None]], "distances": [[0.5]], "ids": [["x"]]}\` and confirms \`search()\` completes without raising. Simple, targeted, cheap.

## Good first issue

Small scope, clear fix, obvious test.

## Related

- #1006 — root cause for why \`None\` metadata appears today (chromadb 1.5.x queue stall). Fixing the pin makes this bug less *frequent* but doesn't eliminate it.

bensig:
## Summary

Fixes #1007.

\`mempalace search\` crashes with \`AttributeError: 'NoneType' object has no attribute 'get'\` when any result has \`None\` metadata. Happens routinely in partial-flush states (very common on chromadb 1.5.x per #1006, but also possible during interrupted mines or schema upgrades even on fixed versions).

## Change

Two-line defensive coercion at the top of the results loop in \`mempalace/searcher.py\`:

\`\`\`diff
+    meta = meta or {}
+    doc = doc or ""
     similarity = round(max(0.0, 1 - dist), 3)
     source = Path(meta.get("source_file", "?")).name
\`\`\`

Result: the hit still shows with its real distance, source/wing/room fall back to \`"?"\`, the document body prints empty. User sees that a hit exists even if its content row is missing — strictly more useful than crashing.

## Test plan

- [ ] Reproduce the crash on a palace with pending \`embeddings_queue\` entries (any chromadb 1.5.x install with a recent mine — see #1006)
- [ ] With this patch, \`mempalace search "anything"\` completes and prints hit placeholders
- [ ] Existing searches on fully-flushed palaces return unchanged output

## Related

- #1006 — root cause for why \`None\` metadata appears routinely today
- \`searcher.py:295\` \`doc.strip()\` also needs the \`doc\` guard (covered by this patch)

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -282,6 +282,13 @@ def search(query: str, palace_path: str, wing: str = None, room: str = None, n_r
     print(f"{'=' * 60}\n")
 
     for i, (doc, meta, dist) in enumerate(zip(docs, metas, dists), 1):
+        # ChromaDB may return None for doc/meta when a drawer's HNSW entry
+        # exists but its metadata/document rows haven't been materialized
+        # (partial-flush states, mid-delete, schema upgrade boundaries).
+        # Degrade gracefully rather than crash — the hit still appears with
+        # real distance; storage fields show "?" where content is missing.
+        meta = meta or {}
+        doc = doc or ""
         similarity = round(max(0.0, 1 - dist), 3)
         source = Path(meta.get("source_file", "?")).name
         wing_name = meta.get("wing", "?")


bensig:
### Code review

Found 1 issue:

1. A more comprehensive predecessor is already open: PR [#999](https://github.com/MemPalace/mempalace/pull/999) (`fix(searcher): guard against None metadata in CLI print path`), opened ~72 minutes before this PR. #999 applies the same `meta = meta or {}` guard in `search()` plus extends coverage to `search_memories()` (the MCP-facing path) and `miner.status()`, and adds regression tests for both. A [Copilot inline comment on #999 line 288](https://github.com/MemPalace/mempalace/pull/999) explicitly flagged the `search_memories()` gap, which #999 addresses but this PR does not. Since `search_memories()` is the production MCP path most likely to be hit, one of these PRs should be closed as superseded in favor of the other — the narrower scope here means the same `AttributeError` will still reach MCP tool consumers even after this PR merges.

https://github.com/MemPalace/mempalace/blob/80ba1eaeed61831199d9765a6377f4e282ff5ac8/mempalace/searcher.py#L283-L291

bensig:
Closing in favor of #999 (opened ~72 min earlier and more comprehensive).

#999 covers the same `meta = meta or {}` guard in `search()`, extends it to `search_memories()` (the MCP-facing path), and also guards `miner.status()`, with regression tests for both paths. A [Copilot inline comment on #999 line 288](https://github.com/MemPalace/mempalace/pull/999) explicitly flagged the `search_memories()` gap — #999 addresses it; this narrower PR does not.

Since `search_memories()` is the production MCP path, leaving it unguarded would mean the same `AttributeError` is still reachable from MCP tool consumers even after this PR merges. No reason to duplicate the diff.

Issue #1007 remains open and is covered by #999.

PR #1010, opened 2026-04-18
fix: upgrade chromadb to >=1.5.4 for Python 3.13/3.14 compatibility + fix 1.5.x queue-stall (closes #1006)

### LINKED ISSUE #1006: bug: fresh `pip install mempalace` silently broken on chromadb 1.5.x — writes queued but never flush, search returns ghosts

bensig:
## Summary

Fixes #1006 (chromadb 1.5.x queue-stall regression) and #487 (chromadb 0.6.x fails on Python 3.13/3.14 due to Pydantic V1 incompatibility) via a single pin change:

\`\`\`diff
- "chromadb>=0.5.0",
+ "chromadb>=1.5.4,<2",
\`\`\`

Plus a regenerated \`uv.lock\` and two docstring/comment updates in \`mempalace/__init__.py\` and \`mempalace/migrate.py\` that reflect the new behavior (chromadb 1.x auto-migrates 0.4.1+ palaces; posthog telemetry is now a no-op stub).

## Provenance

This is a maintainer-sponsored integration of @Legion345's work from #581. The two commits below carry Legion345 as the original git author, preserving their authorship unchanged. I'm landing this on a parallel branch to move the fix through the maintainer path while the #581 thread sits on community re-review after its Apr 17 rebase.

Credits:
- **@Legion345** — authored both commits; diagnosed the Pydantic V1 / Python 3.13/3.14 incompatibility and chose 1.5.4 as the correct floor (commits preserved via cherry-pick)
- **@web3guru888** — approved in [#581](https://github.com/MemPalace/mempalace/pull/581) on Apr 11 after substantive review: *\"The fix is correct for Python 3.13/3.14, and the diagnosis is right — Pydantic V1 is the blocker, and 1.5.4 is where chromadb fully dropped it.\"*
- **@jphein** — production-validated the 0.6.3 → 1.5.7 auto-migration on a 50K-drawer palace: 648 tests pass, 20% test-suite speedup, zero API breakage

## Why this instead of the closed #1008

#1008 took the opposite direction — \`chromadb<1.0\`. A code review on that PR flagged that the \`<1.0\` cap reintroduces a documented breaking incompatibility with Python 3.13+ (see #487 for the exact Pydantic V1 \`ConfigError\`), reversing the intent of #302 which specifically widened chromadb to fix that.

Raising the floor to 1.5.4 is the correct direction: it excludes the broken 1.5.3-and-below versions (covering the known Pydantic V1 cutoff), keeps Python 3.13+ working, and adds a \`<2\` safety cap to guard future chromadb major breakage per @web3guru888's feedback thread.

## Open question on the queue-stall floor

The #1006 reproduction tested chromadb 1.5.6 and 1.5.8 — both stall in the same way (embeddings queued, never materialized; search returns ghosts). I have NOT tested 1.5.4 against that reproduction.

If 1.5.4 is clean on the queue-stall path, this pin is sufficient as-is. If 1.5.4 also stalls, a follow-up tightens the floor (e.g., to the specific patch where the processor was fixed) on top of this PR — strictly narrower than the blanket \`<1.0\` approach, and no regression for Python 3.13+ users.

## Relationship to #581

Happy to close this in favor of #581 if @Legion345 or the maintainers prefer — the diff is Legion345's verbatim. This PR exists as a sponsored merge path, not competing work.

## Supersedes / closes

- Closes #1008 (the \`<1.0\` approach — already closed)
- Fixes #1006 (queue-stall), #487 (Pydantic V1 on Py 3.13/3.14)
- Parallel to #581 (same diff, different merge path)

## Test plan

- [ ] Fresh venv on Python 3.11.x + \`pip install -e .\` — chromadb resolves within \`>=1.5.4,<2\`
- [ ] Fresh venv on Python 3.13.x + \`pip install -e .\` — import succeeds (no Pydantic V1 ConfigError per #487)
- [ ] \`mempalace mine <dir> --mode convos\` — drawers land AND flush (verifies #1006 fix: \`embeddings_queue\` drains, \`col.count()\` matches filed count)
- [ ] \`mempalace status\` — shows filed drawers
- [ ] \`mempalace search "query"\` — returns real content with non-None metadata
- [ ] Existing palace on chromadb 0.6.x opens cleanly under 1.5.4+ (auto-migration per @jphein's validation)

### DIFF
--- a/mempalace/__init__.py
+++ b/mempalace/__init__.py
@@ -4,10 +4,10 @@ import logging
 
 from .version import __version__  # noqa: E402
 
-# ChromaDB 0.6.x ships a Posthog telemetry client whose capture() signature is
-# incompatible with the bundled posthog library, producing noisy stderr warnings
-# on every client operation ("Failed to send telemetry event … capture() takes
-# 1 positional argument but 3 were given").  Silence just that logger.
+# chromadb telemetry: posthog capture() was broken in 0.6.x causing noisy stderr
+# warnings ("capture() takes 1 positional argument but 3 were given"). In 1.x the
+# posthog client is a no-op stub, so this is now harmless — kept as a guard in
+# case future chromadb versions re-introduce real telemetry calls.
 logging.getLogger("chromadb.telemetry.product.posthog").setLevel(logging.CRITICAL)
 
 # NOTE: the previous block set ``ORT_DISABLE_COREML=1`` on macOS arm64 as a
--- a/mempalace/migrate.py
+++ b/mempalace/migrate.py
@@ -7,8 +7,10 @@ Reads documents and metadata directly from the palace's SQLite database
 then re-imports everything into a fresh palace using the currently installed
 ChromaDB version.
 
-This fixes the 3.0.0 → 3.1.0 upgrade path where chromadb was downgraded
-from 1.5.x to 0.6.x, breaking the on-disk storage format.
+Since mempalace 3.2.0 (chromadb>=1.5.4), chromadb automatically migrates
+0.4.1+ databases on first open — no manual migration needed for upgrades.
+Use this command only when downgrading chromadb (e.g. rolling back to an
+older mempalace release) or if automatic migration fails.
 
 Usage:
     mempalace migrate                          # migrate default palace
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,11 +21,13 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
     "Topic :: Scientific/Engineering :: Artificial Intelligence",
     "Topic :: Utilities",
 ]
 dependencies = [
-    "chromadb>=0.5.0",
+    "chromadb>=1.5.4",
     "pyyaml>=6.0,<7",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
     "Topic :: Utilities",
 ]
 dependencies = [
-    "chromadb>=1.5.4",
+    "chromadb>=1.5.4,<2",
     "pyyaml>=6.0,<7",
 ]
 
--- a/uv.lock
+++ b/uv.lock
@@ -65,15 +65,12 @@ wheels = [
 ]
 
 [[package]]
-name = "asgiref"
-version = "3.11.1"
+name = "attrs"
+version = "26.1.0"
 source = { registry = "https://pypi.org/simple" }
-dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
-]
-sdist = { url = "https://files.pythonhosted.org/packages/63/40/f03da1264ae8f7cfdbf9146542e5e7e8100a4c66ab48e791df9a03d3f6c0/asgiref-3.11.1.tar.gz", hash = "sha256:5f184dc43b7e763efe848065441eac62229c9f7b0475f41f80e207a114eda4ce", size = 38550, upload-time = "2026-02-03T13:30:14.33Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/9a/8e/82a0fe20a541c03148528be8cac2408564a6c9a0cc7e9171802bc1d26985/attrs-26.1.0.tar.gz", hash = "sha256:d03ceb89cb322a8fd706d4fb91940737b6642aa36998fe130a9bc96c985eff32", size = 952055, upload-time = "2026-03-19T14:22:25.026Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/5c/0a/a72d10ed65068e115044937873362e6e32fab1b7dce0046aeb224682c989/asgiref-3.11.1-py3-none-any.whl", hash = "sha256:e8667a091e69529631969fd45dc268fa79b99c92c5fcdda727757e52146ec133", size = 24345, upload-time = "2026-02-03T13:30:13.039Z" },
+    { url = "https://files.pythonhosted.org/packages/64/b4/17d4b0b2a2dc85a6df63d1157e028ed19f90d4cd97c36717afef2bc2f395/attrs-26.1.0-py3-none-any.whl", hash = "sha256:c647aa4a12dfbad9333ca4e71fe62ddc36f4e63b2d260a37a8b83d2f043ac309", size = 67548, upload-time = "2026-03-19T14:22:23.645Z" },
 ]
 
 [[package]]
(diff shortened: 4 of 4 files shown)


bensig:
Rebased onto current develop (tip `2b9f17c`) to pick up the RFC 001 backend entry-points block (`[project.entry-points."mempalace.backends"]`) from PR #995 which merged after this branch was originally based. Branch was ~4 commits behind.

Both Legion345 commits preserved with original authorship through the rebase — only the SHAs changed (`8445b95` → `d0c8ecd`, `ffe8e9a` → `fa7fe1d`). No diff conflicts.

igorls (commented):
Tested locally on a clean pr-1010 worktree. **Open question in the PR body — whether chromadb 1.5.4 itself is clean on the #1006 queue-stall path — is now answered: yes.**

## What I ran

**Python 3.14.4 venv, chromadb pinned to exactly `==1.5.4`:**
- `import chromadb` succeeds — no Pydantic V1 `ConfigError` (issue #487).
- End-to-end mine → `col.count()` → search, using a synthetic Claude-format JSONL session:
  ```
  Drawers filed:                2
  collection count after mine:  2   ← matches, no #1006 queue stall
  search hits (first query):    4   ← hybrid expansion, all non-None meta
  ```
- Full test suite: **991 passed**, 2 warnings, 85s.

**Python 3.14.4 venv, uv's resolved `==1.5.7` (same as lockfile):**
- Same e2e: count matches, search works.
- Full test suite: **991 passed**, 82s.

So `>=1.5.4` is the right floor — it's where the queue-stall is already fixed, not just where Pydantic V1 was dropped. No need to tighten to 1.5.7+.

## Logic-diff review (ignoring uv.lock)

Three-line production change is clean:

- `pyproject.toml`: `chromadb>=0.5.0` → `chromadb>=1.5.4,<2`, plus `Python 3.13`/`3.14` classifiers. The `<2` cap is correct belt-and-suspenders per @web3guru888's feedback thread.
- `mempalace/__init__.py`: posthog telemetry-silencer comment updated to reflect that the filter is now a guard against regressions (chromadb 1.x makes posthog a no-op stub). The `logging.getLogger(...).setLevel(CRITICAL)` call is kept — good, harmless as a guard.
- `mempalace/migrate.py`: docstring updated to note chromadb 1.x auto-migrates 0.4.1+ palaces, and this command is now only for downgrades / auto-migrate failures. Accurate.

## What I didn't test

- **Existing-palace 0.6.x → 1.5.x auto-migration.** The PR relies on @jphein's 50K-drawer validation in #581. Relying on upstream validation here is fine; I didn't spin up a 0.6.x palace to re-verify.
- Windows.

## Attribution

Cherry-picked commits preserve @Legion345 as author per the PR body — confirmed via `git log`. Credit chain (Legion345 / web3guru888 / jphein) is spelled out clearly.

LGTM — happy to merge once CI is green. The open question about 1.5.4 is resolved; the pin is correct as written.

PR #1013, opened 2026-04-18
fix: guard Layer3.search_raw against None doc/meta from ChromaDB (#1011)

### LINKED ISSUE #1011: bug: `Layer3.search_raw()` in layers.py crashes on None metadata (same pattern as #1007)
## Problem

\`Layer3.search_raw()\` in \`mempalace/layers.py\` has the same unguarded \`meta.get(...)\` pattern that #1007 / PR #999 fix in \`searcher.py\`. When ChromaDB's \`query()\` returns \`None\` for any metadata entry (partial-flush, mid-delete, schema upgrade boundaries), this path crashes with:

\`\`\`
\`\`\`

## Where

\`mempalace/layers.py\`, around lines 300-341 — the \`Layer3.search_raw\` method iterates directly over \`col.query(**kwargs)\` results and calls \`meta.get(\"wing\", \"unknown\")\`, \`meta.get(\"room\", \"unknown\")\`, and \`meta.get(\"source_file\", \"?\")\` without any null safety.

## Root cause

Same as #1007: ChromaDB \`query()\` can return \`None\` in the \`metadatas\` list when a drawer's HNSW vector entry exists but its metadata row hasn't been materialized yet. #1006 documented this happens routinely on chromadb 1.5.x; even after the chromadb pin lands, partial-state results remain possible during interrupted mines or schema upgrades.

## Proposed fix

Same two-line defensive coercion as #999 applies in \`searcher.py\`:

\`\`\`diff
+    meta = meta or {}
+    doc = doc or ""
     # ... meta.get(...) calls that follow
\`\`\`

Apply at the top of the results loop in \`Layer3.search_raw()\` so every downstream \`meta.get()\` and \`doc.strip()\` (or equivalent) is safe.

## Relationship to other PRs

- #1007 + PR #999 — same class of bug, covers \`searcher.py\` \`search()\` + \`search_memories()\` + \`miner.status()\`. **Does NOT cover \`Layer3.search_raw()\`** (different file).
- #1006 — root cause (chromadb 1.5.x queue-stall makes partial-flush states the common case today)
- #1010 — upgrades chromadb floor to 1.5.4; reduces frequency of None metadata but doesn't eliminate it

## Good first issue

Small scope, established pattern (can copy from #999 / PR #1009 approach), self-contained in one file.

## Why separate from #1007

Kept separate because \`layers.py\` is a different file with different test fixtures. PR #999 scoped itself to \`searcher.py\` + \`miner.status()\`; this is the third sibling that should land as a standalone PR to keep diffs reviewable.

bensig:
## Summary

Fixes #1011.

`Layer3.search_raw()` in `mempalace/layers.py` has the same unguarded `meta.get(...)` pattern that #1007 / PR #999 fix in `searcher.py`. Applies the same two-line defensive coercion at the top of the results loop.

## Change

One file, defensive guard added at the top of the `search_raw` results loop:

```diff
 for ...
+    meta = meta or {}
+    doc = doc or ""
     # ... downstream meta.get() / doc.* calls
```

## Test plan

- [ ] Reproduce on a palace with `None` metadata in ChromaDB query results (any 1.5.x install per #1006)
- [ ] `Layer3.search_raw("query")` completes and returns results
- [ ] Existing callers unaffected on fully-flushed palaces

## Related

- #1007, PR #999 — same class of bug in `searcher.py`
- #1006, #1010 — chromadb 1.5.x queue-stall root cause and fix

### DIFF
--- a/mempalace/layers.py
+++ b/mempalace/layers.py
@@ -281,6 +281,8 @@ class Layer3:
 
         lines = [f'## L3 — SEARCH RESULTS for "{query}"']
         for i, (doc, meta, dist) in enumerate(zip(docs, metas, dists), 1):
+            meta = meta or {}
+            doc = doc or ""
             similarity = round(1 - dist, 3)
             wing_name = meta.get("wing", "?")
             room_name = meta.get("room", "?")
@@ -327,6 +329,13 @@ class Layer3:
             _first_or_empty(results, "metadatas"),
             _first_or_empty(results, "distances"),
         ):
+            # ChromaDB may return None for doc/meta when a drawer's HNSW entry
+            # exists but its metadata/document rows haven't been materialized
+            # (partial-flush states, mid-delete, schema upgrade boundaries).
+            # Degrade gracefully — the hit still appears with real distance;
+            # storage fields show their fallback where content is missing.
+            meta = meta or {}
+            doc = doc or ""
             hits.append(
                 {
                     "text": doc,


igorls (commented):
Tested locally (py 3.12, clean worktree on pr-1013):

**Reproduction confirms fix.** Wrote a small script that patches `_get_collection` to return the partial-flush shape ChromaDB emits (valid hit + `None` meta + `None` doc):

```python
from unittest.mock import patch
from mempalace.layers import Layer3

class FakeCol:
    def query(self, **_):
        return {
            "documents": [["real doc text", None]],
            "metadatas": [[{"wing": "w1", "room": "r1", "source_file": "/a/b.md"}, None]],
            "distances": [[0.1, 0.42]],
        }

with patch("mempalace.layers._get_collection", return_value=FakeCol()):
    hits = Layer3(palace).search_raw("anything", n_results=2)
```

- **Without the fix** (checked out `develop`'s `layers.py`): `AttributeError: 'NoneType' object has no attribute 'get'` at `layers.py:333` — exactly matches issue #1011.
- **With the fix** (pr-1013 `layers.py`): 2 hits, second falls through cleanly with `wing="unknown"`, `room="unknown"`, `source_file="?"`, `text=""`.

Full test suite: 973 passed. The two-line guard exactly mirrors #999's pattern in `searcher.py`, so the code-review surface is trivial.

Worth a small follow-up (not this PR): a unit test that pins the None-meta behavior so neither this nor the `searcher.py` sibling regresses. The reproduction above is essentially that test.

LGTM.

bensig:
### Code review

Found 2 issues:

1. The sibling method `Layer3.search` has the identical unguarded `meta.get(...)` pattern and is NOT covered by this PR. It iterates over the same ChromaDB \`zip(docs, metas, dists)\` shape and calls \`meta.get(\"wing\", \"?\")\`, \`meta.get(\"room\", \"?\")\`, \`meta.get(\"source_file\", ...)\` without a preceding \`meta = meta or {}\` guard — so the exact partial-flush / mid-delete failure mode this PR fixes for \`search_raw\` still raises \`AttributeError\` on \`Layer3.search\`. Ideally this PR extends the guard to both methods, or a follow-up does.

https://github.com/MemPalace/mempalace/blob/66f33a640f8f6b2889c4f90e43f8c784842f3041/mempalace/layers.py#L280-L288

2. This is the third per-site guard for the same bug class (after #999 for `searcher.py` and this one for `layers.py`). Reviewer @jphein on [#999](https://github.com/MemPalace/mempalace/pull/999) explicitly proposed consolidating the coercion at the ChromaDB adapter boundary (`backends/chroma.py` — coerce `None` metadatas to `{}` at the \`ChromaCollection.query()\` / \`.get()\` layer), calling the per-site approach \"whack-a-mole across 8 call sites\" and noting adapter-level consolidation would let every future caller inherit the guard for free. Worth a maintainer decision before this lands: either accept per-site guards as policy, or block in favor of an adapter-level fix that would render #999, #1013, and any future sibling PR redundant.

https://github.com/MemPalace/mempalace/blob/66f33a640f8f6b2889c4f90e43f8c784842f3041/mempalace/backends/chroma.py#L1-L20

bensig:
Amended to also guard `Layer3.search` (sibling method with the identical pattern). Scorer confirmed it had the same unguarded `meta.get()` calls that would raise `AttributeError` under the same partial-flush conditions this PR addresses for `search_raw`. Two-line coercion mirrors the existing guard — no new comment block since the one above `search_raw`'s guard already documents the pattern.

Remaining finding (adapter-level consolidation per @jphein on #999) is a maintainer design call — not pulling that into this PR.

PR #1019, opened 2026-04-18
fix(searcher): guard against None metadata/doc in search result loops

### LINKED ISSUE #1007: bug: `mempalace search` crashes with `AttributeError: 'NoneType' object has no attribute 'get'` on partial-state results (searcher.py:286)

cantenesse:
## Problem

`mempalace search` crashes with `AttributeError: 'NoneType' object has no attribute 'get'` when ChromaDB returns `None` for a metadata or document entry. This happens in several edge-case states:

- **Partial-flush** — HNSW index has an entry but `embedding_metadata` hasn't materialized yet
- **Mid-delete** — metadata rows purged before HNSW entry removed
- **Upgrade boundaries** — palace written by one schema, read by another
- **Interrupted mines** — process killed between segment writes

Fixes #1007, #1011

## Fix

Add `meta = meta or {}` and `doc = doc or ""` defensive guards at the top of the three result-iteration loops in `searcher.py`:

1. **`search()` display loop** (line ~284) — guards `meta.get()` and `doc.strip()` calls
2. **`search_memories()` closet hybrid loop** (line ~368) — guards `cmeta.get()`
3. **`search_memories()` drawer scored loop** (line ~388) — guards `meta.get()`

Results with `None` metadata degrade gracefully (showing `?` placeholders) rather than crashing.

## Testing

The fix is a two-line defensive guard per loop — verifiable by mocking a ChromaDB query response with `{"documents": [[None]], "metadatas": [[None]], "distances": [[0.5]], "ids": [["x"]]}` and confirming `search()` completes without raising.

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -340,7 +340,7 @@ def search(query: str, palace_path: str, wing: str = None, room: str = None, n_r
     # `_hybrid_rank`; do the same here so CLI results match what agents
     # see via `mempalace_search`.
     hits = [
-        {"text": doc, "distance": float(dist), "metadata": meta or {}}
+        {"text": doc or "", "distance": float(dist), "metadata": meta or {}}
         for doc, meta, dist in zip(docs, metas, dists)
     ]
     hits = _hybrid_rank(hits, query)
@@ -809,6 +809,8 @@ def search_memories(
         _first_or_empty(drawer_results, "metadatas"),
         _first_or_empty(drawer_results, "distances"),
     ):
+        meta = meta or {}
+        doc = doc or ""
         # Filter on raw distance before rounding to avoid precision loss.
         if max_distance > 0.0 and dist > max_distance:
             continue


igorls:
Rebased onto develop (post-#1029 + #1030 hybrid-rank refactor and clamp). Conflict in `searcher.py`: the metadata-None guard from this PR landed upstream as part of the hits-builder list comp (`meta or {}` on the `hits=[...]` line). Final delta from this PR is now narrowly:

1. Add `doc or ""` alongside `meta or {}` at the same hits-builder line
2. Keep the `meta = meta or {}` / `doc = doc or ""` guards in `search_memories` (separate loop, not refactored)

`pytest tests/test_searcher.py`: 27 passed locally. Lint/format clean.

PR #1021, opened 2026-04-18
fix(hooks): restore silent-save visibility on Claude Code 2.1.114

jphein:
Two related silent-save fixes exposed by Claude Code 2.1.114's plugin-scope Stop hook dispatch. Both are producer-side — getting clean JSON with `systemMessage` onto fd 1 where Claude Code reads it. The separate UI-side regression that's preventing the `<Line>` from persisting on screen is filed at [anthropics/claude-code#50542](https://github.com/anthropics/claude-code/issues/50542).

## Fix 1 — honor `silent_save` when `stop_hook_active: true`

Claude Code 2.1.114 sets `stop_hook_active: true` on every Stop fire after the first in a session (at least for plugin-dispatched hooks). The legacy guard at the top of `hook_stop` was written for block-mode — when the hook returned `{"decision":"block"}` to ask Claude to save, a re-fire with the flag set meant "you already blocked, don't block again" so we output `{}` and bailed.

Silent-save mode (default since #673) never blocks — it saves directly and returns. The flag is meaningless in that mode, but the old guard was still suppressing every auto-save after the first one in a Claude Code session:

- First fire: `stop_hook_active: false` → saves fine, `systemMessage` output
- Every subsequent fire: `stop_hook_active: true` → legacy guard returns `{}`, silent no-op, log stays empty, marker stuck

The fix only applies the guard when the user is actually in block mode:

```python
if str(stop_hook_active).lower() in ("true", "1", "yes"):
    try:
        silent_guard = MempalaceConfig().hook_silent_save
    except Exception:
        silent_guard = True
    if not silent_guard:
        _output({})
        return
```

## Fix 2 — write hook JSON to real stdout

`mempalace.mcp_server` redirects stdout → stderr at module-level import (Python-level `sys.stdout = sys.stderr` and fd-level `os.dup2(2, 1)`) to protect the MCP JSON-RPC channel from ChromaDB's C-level noise. That redirect was added in #864 and has lived peacefully alongside silent-save for months.

Silent-save imports `mcp_server` transitively via `_save_diary_direct` → `from .mcp_server import tool_diary_write`. By the time `_output()` runs, `sys.stdout` is stderr, and `print(json.dumps(...))` lands on fd 2. Claude Code reads the hook's stdout (fd 1), finds it empty, and never sees the `systemMessage` payload.

Historically this didn't bite because… well, I'm honestly not sure. The redirect's been in place since early March, silent-save's been in place since early April, and the `systemMessage` line was rendering through at least 2026-04-17. Something in 2.1.114's dispatch path seems stricter — possibly a shift from reading both streams and parsing the first JSON-like chunk to strictly parsing stdout. Either way, writing to the real fd is the correct long-term fix.

`_output()` now writes to `_REAL_STDOUT_FD` (saved by `mcp_server` before the redirect, exported as a module attribute) via `os.write()`, falling back to `sys.stdout` only when unavailable (e.g., hooks_cli used standalone without `mcp_server`). The full schema including `continue: true, suppressOutput: false` is emitted for completeness:

```json
{
  "continue": true,
  "suppressOutput": false,
  "systemMessage": "✦ 30 memories woven into the palace — themes"
}
```

## Test

- Existing suite passes (58/58 in `tests/test_hooks_cli.py`).
- Manual verification — `bash hooks/mempal-stop-hook.sh 2>/dev/null` now shows only the JSON on stdout, `2>&1 >/dev/null` shows only the log line on stderr. Cleanly separated streams.

## UI-side status

The corresponding `<Line>` rendering regression in Claude Code 2.1.114 is tracked at [anthropics/claude-code#50542](https://github.com/anthropics/claude-code/issues/50542). That one's not fixable from our side; this PR ensures we're producing correct output once the client fix lands.

## Test plan

- [ ] `python -m pytest tests/test_hooks_cli.py -q` — 58 tests pass
- [ ] Install the plugin; run Claude Code 2.1.114; confirm `diary_*` drawers land at the 15-exchange mark
- [ ] Inspect hook script output directly: clean JSON on stdout, log noise on stderr

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -255,9 +255,7 @@ def hook_stop(data: dict, harness: str):
             try:
                 silent_guard = MempalaceConfig().hook_silent_save
             except AttributeError as exc:
-                _log(
-                    f"WARNING: could not read hook_silent_save: {exc}; defaulting to silent mode"
-                )
+                _log(f"WARNING: could not read hook_silent_save: {exc}; defaulting to silent mode")
         if not silent_guard:
             _output({})
             return
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -209,10 +209,19 @@ def hook_stop(data: dict, harness: str):
     stop_hook_active = parsed["stop_hook_active"]
     transcript_path = parsed["transcript_path"]
 
-    # If already in a save cycle, let through (infinite-loop prevention)
+    # If already in a block-mode save cycle, let through (infinite-loop prevention).
+    # Silent mode saves directly without returning {"decision":"block"}, so there's
+    # no loop to prevent — and Claude Code's plugin dispatch sets this flag on every
+    # fire after the first, which would otherwise suppress all subsequent auto-saves.
     if str(stop_hook_active).lower() in ("true", "1", "yes"):
-        _output({})
-        return
+        try:
+            from .config import MempalaceConfig
+            silent_guard = MempalaceConfig().hook_silent_save
+        except Exception:
+            silent_guard = True
+        if not silent_guard:
+            _output({})
+            return
 
     # Count human messages
     exchange_count = _count_human_messages(transcript_path)
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -241,19 +241,22 @@ def hook_stop(data: dict, harness: str):
     # no loop to prevent — and Claude Code's plugin dispatch sets this flag on every
     # fire after the first, which would otherwise suppress all subsequent auto-saves.
     if str(stop_hook_active).lower() in ("true", "1", "yes"):
-        silent_guard = False
+        # Safe default: assume silent mode on any config-read failure so saves
+        # proceed rather than being silently dropped. Silent mode is the default
+        # (v3.3.0+), so if we can't read config, behave as if it's still on.
+        silent_guard = True
         try:
             from .config import MempalaceConfig
         except ImportError as exc:
             _log(
-                f"WARNING: could not import MempalaceConfig for stop guard: {exc}; preserving block-mode guard"
+                f"WARNING: could not import MempalaceConfig for stop guard: {exc}; defaulting to silent mode"
             )
         else:
             try:
                 silent_guard = MempalaceConfig().hook_silent_save
             except AttributeError as exc:
                 _log(
-                    f"WARNING: could not read hook_silent_save: {exc}; preserving block-mode guard"
+                    f"WARNING: could not read hook_silent_save: {exc}; defaulting to silent mode"
                 )
         if not silent_guard:
             _output({})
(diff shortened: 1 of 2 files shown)


jphein:
Addressed all five Copilot review points:

- **Unconditional import**: `_output()` now uses `sys.modules.get()` to check if `mcp_server` is already loaded rather than importing it — avoids triggering the stdout redirect as a side effect in any process that calls `_output()` cold.
- **Partial writes**: added a write-all loop around `os.write()`, retrying on `InterruptedError` and falling back to `sys.stdout.buffer` on `OSError`.
- **Docstring**: removed the inaccurate `_save_diary_direct` reference; the doc now just describes what the function actually does.
- **Broad except in guard**: narrowed to `ImportError`/`AttributeError`, default `silent_guard=False` (safe: preserves block-mode loop prevention when config can't be loaded), logs a warning instead of silently changing behavior.
- **Regression test**: two new tests — one asserts `_output()` writes to the `_REAL_STDOUT_FD` pipe when `mcp_server` is in `sys.modules`, one asserts it falls back to fd 1 when it isn't. Both verify the JSON payload round-trips correctly.

bensig:
### Code review

Found 1 issue:

1. Guard fallback defaults to wrong state, re-enabling the bug this PR fixes. `silent_guard` is initialized to `False` (line 244). When `MempalaceConfig` import fails (`ImportError`) or property access fails (`AttributeError`), neither exception handler reassigns `silent_guard` -- it stays `False`. Then `if not silent_guard:` evaluates to `True`, firing the guard and suppressing the save. This is the exact opposite of the intended behavior: in silent mode (the default), saves should proceed when config is unavailable. The safe default should be `silent_guard = True` so that on any error, the guard is skipped and saves are not suppressed.

https://github.com/MemPalace/mempalace/blob/2183d866f3477c620b16cc34155cd5c5cf288c60/mempalace/hooks_cli.py#L243-L260

bensig:
@jphein pls see comment above

bensig (approved):

jphein:
Good catch, thanks. Fixed in [`2629ae5`](https://github.com/jphein/mempalace/commit/2629ae5): `silent_guard` now defaults to `True`, and the warning log text was updated to say "defaulting to silent mode" so the behaviour in the logs matches the new default. Tests still green on all 3 platforms + both Python versions.

jphein:
Small correction on my earlier comment — the CI matrix is 3 Python versions (3.9, 3.11, 3.13) × 3 platforms, not 2. All still green.

jphein:
One more correction on the same thread — the "tests still green" claim I made was premature: the CI run for `2629ae5` was actually still in flight when I posted, and it failed shortly after on a `ruff format` rule (one log call needed to be on a single line). Fix pushed in [`d657626`](https://github.com/jphein/mempalace/commit/d657626); that run is green across all 3 platforms × 3 Python versions. Apologies for the sloppy narration.

PR #1023, opened 2026-04-19
fix(hooks): PID file guard prevents stacking mine processes

jphein:
## Summary

Claude Code stop hooks fire every 15 user messages. Each fire calls `_maybe_auto_ingest()`, which spawns `mempalace mine` via `subprocess.Popen`. `Popen` returns immediately, so a long-running mine never blocks the next hook fire. Over a long session this stacks mine processes — **4 concurrent mines pinning ~770% CPU** observed in production on my workstation before this fix.

## Root cause

`_maybe_auto_ingest` has no dedup. `Popen(...)` returns before the child has finished; the next stop-hook fire has no way to see that a previous background mine is still running, so it spawns another. Each mine competes for the same ChromaDB writer and HNSW index — result is both high CPU and contention on the palace.

## Fix

Two new helpers in `mempalace/hooks_cli.py`:

- `_mine_already_running()` — reads `$STATE_DIR/mine.pid`, parses it, and calls `os.kill(pid, 0)`. Signal 0 is the documented POSIX idiom for "does this PID exist, and may I signal it?" — no signal delivered. Swallows `FileNotFoundError`, `ValueError`, `ProcessLookupError`, and `PermissionError` and returns `False` — any of those means "no live mine from us."
- `_spawn_mine(cmd)` — performs the old `Popen` call plus `_MINE_PID_FILE.write_text(str(proc.pid))` immediately after. Centralizes the spawn so the PID-file write can't drift out of sync with the process.

`_maybe_auto_ingest` checks the guard up front; if True, it logs `"Skipping auto-ingest: mine already running"` and returns. Otherwise it delegates to `_spawn_mine`.

The PID file lives at `$STATE_DIR/mine.pid` (default `~/.mempalace/hook_state/mine.pid`) next to the existing `hook.log`. No new directories.

## Why `os.kill(pid, 0)` is safe

- Signal `0` sends no signal — pure existence probe.
- `ProcessLookupError` → PID is dead → guard returns `False`, new mine spawned.
- `PermissionError` → PID belongs to another user → treated as `False`; can't be our mine.
- PID reuse is a theoretical concern but not practical at the seconds-to-minutes timescales between hook fires. Palace is single-user in typical use.

## Scope

- Only touches the background `mempalace mine` spawn path in `_maybe_auto_ingest`. `_mine_sync` (precompact) is left alone — it's synchronous and already self-serializing.
- No behavior change when no mine is running: identical spawn, just with a PID write afterward.
- No new configuration flags or env vars.

## Test coverage

5 new tests in `tests/test_hooks_cli.py`:

- `test_mine_already_running_no_file` — False when `mine.pid` doesn't exist
- `test_mine_already_running_dead_pid` — False when PID file contains `999999999`
- `test_mine_already_running_live_pid` — True when PID file contains `os.getpid()`
- `test_mine_already_running_corrupt_file` — False on a non-integer PID file
- `test_maybe_auto_ingest_skips_when_mine_running` — asserts `subprocess.Popen` is NOT called when `_mine_already_running` is patched to True

Three existing tests are patched to redirect `_MINE_PID_FILE` into `tmp_path` so the suite never writes to the real `~/.mempalace/hook_state/mine.pid`.

```
pytest tests/test_hooks_cli.py -q
51 passed in 0.17s
```

## Test plan

- [ ] `pytest tests/test_hooks_cli.py -q` — all green (51 tests)
- [ ] Fire the stop hook twice in rapid succession within the mine duration; confirm only one `mempalace mine` process appears in `ps`
- [ ] `grep "Skipping auto-ingest" ~/.mempalace/hook_state/hook.log` — guard log line on the second fire
- [ ] Wait for first mine to exit, fire hook again, confirm fresh mine spawns (guard doesn't stick)

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -159,7 +159,10 @@ def _mine_already_running() -> bool:
         pid = int(_MINE_PID_FILE.read_text().strip())
         os.kill(pid, 0)  # signal 0 = existence check, no actual signal sent
         return True
-    except (FileNotFoundError, ValueError, ProcessLookupError, PermissionError):
+    except (OSError, ValueError):
+        # OSError covers: FileNotFoundError (no pid file), ProcessLookupError
+        # (dead PID on POSIX), PermissionError (not our process), and
+        # WinError 87 / "invalid parameter" (dead or unknown PID on Windows).
         return False
 
 
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -150,20 +150,38 @@ def _get_mine_dir(transcript_path: str = "") -> str:
     return ""
 
 
+_MINE_PID_FILE = STATE_DIR / "mine.pid"
+
+
+def _mine_already_running() -> bool:
+    """Return True if a background mine process from a previous hook fire is still alive."""
+    try:
+        pid = int(_MINE_PID_FILE.read_text().strip())
+        os.kill(pid, 0)  # signal 0 = existence check, no actual signal sent
+        return True
+    except (FileNotFoundError, ValueError, ProcessLookupError, PermissionError):
+        return False
+
+
+def _spawn_mine(cmd: list) -> None:
+    """Spawn a mine subprocess, write its PID to the lock file, log to hook.log."""
+    STATE_DIR.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


bensig:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

bensig (approved):

PR #1027, opened 2026-04-19
ci: retry pip installs on transient PyPI/TLS failures

vnguyen-lexipol:
Observed `SSL: DECRYPTION_FAILED_OR_BAD_RECORD_MAC` during `pip install` on [run 24379493758](https://github.com/MemPalace/mempalace/actions/runs/24379493758). Plain `pip install` has no retries, so a transient TLS glitch fails the whole job.

This PR:
- Upgrades pip before install
- Adds `--retries 10 --timeout 120` to pip install
- Wraps install in a 5-attempt retry loop with backoff (15s / 30s / 45s / 60s)
- Applies to all 4 install steps (ubuntu / windows / macos + lint)

No behavior change on the happy path — retries only kick in when pip fails transiently.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,22 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
-      - run: pip install -e ".[dev]"
+      - name: Install package (retry on transient PyPI/TLS errors)
+        shell: bash
+        run: |
+          set -e
+          python -m pip install --upgrade pip
+          for attempt in 1 2 3 4 5; do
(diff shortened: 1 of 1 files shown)


igorls:
Closing in favor of #939, which just merged and addresses the same root cause from a different angle: enabling `cache: 'pip'` on `actions/setup-python` removes most PyPI traffic from CI, which is the more reliable mitigation for transient network/TLS errors than retry logic.

Thanks for the contribution. If transient install failures persist after caching is in place, a follow-up that wraps installs in something like `nick-fields/retry@v3` (rather than four copy-pasted shell loops) would be welcome — but let's see whether caching alone is enough first.

PR #1031, opened 2026-04-19
feat(i18n): add Hebrew language support

shaibachar:
## What does this PR do?

  Adds Hebrew language support to the i18n/entity-detection layer.

  - Adds a new Hebrew locale file at /abs/path/mempalace/i18n/he.json
  - Extends i18n coverage tests to include Hebrew sample text in /abs/path/mempalace/tests/test_i18n.py
  - Adds Hebrew-specific entity detection tests in /abs/path/mempalace/tests/test_entity_detector.py

  ## How to test

  Run:

  python -m pytest tests/test_i18n.py tests/test_entity_detector.py -v
  python -m pytest tests/ -v

  Expected:

  - Hebrew locale loads successfully
  - Hebrew sample compression test passes
  - Hebrew entity candidate extraction and person-verb scoring tests pass

  ## Checklist

  - [x] No hardcoded paths
  - [ ] Linter passes (ruff check .)


mvalentsev:
A couple of things I noticed while reading the diff:

**Scope**: this PR is bundling a release bump with the locale addition. The Hebrew scope is `he.json` plus the two test files, but the diff also touches `pyproject.toml`, `mempalace/version.py`, `uv.lock`, the `README.md` badge, all three plugin manifests (`.claude-plugin/marketplace.json`, `.claude-plugin/plugin.json`, `.codex-plugin/plugin.json`), and adds a full `[3.3.1]` CHANGELOG block that attributes 20+ other merged PRs (#156, #760, #907, #773, #778, #911, #932, #928, #931, #946, #758, #876, ...). Those version bumps and release notes are usually maintainer work, and keeping them here will conflict with whatever 3.3.1 looks like when it is actually cut. Reverting the non-Hebrew files would leave a cleaner PR to review.

**`regex.stop_words` catches domain nouns**: the current list filters `ארמון` (palace), `אגף` (wing), `ארון` (closet), `מגירה` (drawer), plus generic terms like `קובץ`, `קוד`, `בדיקה`, `פרויקט`, `עבודה`. BM25 will strip those from Hebrew queries and documents, so searches that use the palace vocabulary or common project nouns will miss their natural matches. Tightening the list to function words (prepositions, pronouns, copulas) matches what the other locales ship.

**Duplicate `את` entry**: `"את"` appears twice in the `entity.stopwords` array (as the 2nd item and again as the 22nd). The same token also repeats inside the space-separated `regex.stop_words` string. Stopword lookup dedupes on match, so the repeats are harmless but likely unintended.

**`boundary_chars`**: Hebrew has the same `\b` problem as Devanagari and Arabic that #932 introduced `boundary_chars` for. Without it, `\b{name}\b` person-verb patterns will not fire reliably when Hebrew names adjoin Hebrew text (`ן`, `ם`, `י` endings in particular). Adding a `boundary_chars` field with the Hebrew block would let the loader expand `\b` correctly.

shaibachar:
i will do cleanup and apply comments

PR #1047, opened 2026-04-20
fix(search): tolerate HNSW orphans and surface real MCP errors

vilosource:
## Summary

- **Fix crash when ChromaDB returns HNSW orphan vectors.** If an embedding exists in the HNSW index but its main-table row is gone (partial writes / interrupted deletes), `col.query()` surfaces `metadata=None, document=None`. `searcher.search_memories` and the CLI `search()` both called `meta.get("wing", ...)` unconditionally and crashed with `AttributeError`. Both now skip `None` rows and log a warning with the distance so orphans can be identified and pruned.
- **Stop masking real errors in the MCP dispatcher.** `mcp_server.py:1380` wrapped every tool handler in a bare `except` and returned a generic `{"code": -32000, "message": "Internal tool error"}`. The payload now includes the exception class name and message (e.g. `"Internal tool error: KeyError: 'missing-key'"`) so clients can diagnose without log access.

## Context

Encountered in the wild on a 126-drawer palace that had accumulated 47 HNSW orphans. `mempalace_search` returned `-32000` for any query where an orphan fell into the top-N by cosine distance — deterministic per query, invisible without reading server code. Orphans were cleaned via `col.delete(ids=[...])` (count stayed unchanged — HNSW-only entries aren't counted); root cause of their creation is unclear (interrupted `add_drawer`, `dedup.py`, or earlier delete-path bug — deferred since this change makes it non-fatal).

`repair.py scan` does **not** detect this class of corruption — it iterates `col.get()` IDs and probes them, so HNSW-only orphans aren't in its input set. Potential follow-up: probe the HNSW directly in `scan` to surface this case.

## Changes

- `mempalace/searcher.py`
  - `search_memories`: skip rows where `meta is None` or `doc is None`; log warning with distance.
  - CLI `search`: same skip + warning.
- `mempalace/mcp_server.py` — include `type(exc).__name__` and message in `-32000` response.
- `tests/test_searcher.py` — 4 new cases (API + CLI × null meta + null doc).
- `tests/test_mcp_server.py` — 1 new case asserting exception class reaches the client.

## Test plan

- [x] `pytest tests/test_searcher.py` — 20 passed (4 new)
- [x] `pytest tests/test_mcp_server.py` — 50 passed (1 new)
- [x] Full suite (excluding benchmarks): **657 passed**
- [x] Verified against live palace: after cleanup, all previously-failing queries (`todo`, `current work`, etc.) return results cleanly.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1377,12 +1377,15 @@ def handle_request(request):
                 "id": req_id,
                 "result": {"content": [{"type": "text", "text": json.dumps(result, indent=2)}]},
             }
-        except Exception:
+        except Exception as exc:
             logger.exception(f"Tool error in {tool_name}")
             return {
                 "jsonrpc": "2.0",
                 "id": req_id,
-                "error": {"code": -32000, "message": "Internal tool error"},
+                "error": {
+                    "code": -32000,
+                    "message": f"Internal tool error: {type(exc).__name__}: {exc}",
+                },
             }
 
     # Notifications (missing id) must never get a response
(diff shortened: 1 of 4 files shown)


vilosource:
Closing — opening on the fork instead.

PR #893, opened 2026-04-14
fix: harden hooks, MCP server, and config (security audit)

brodsky754:
## Summary
- **hooks_cli**: Add `_is_safe_transcript_path()` to reject `..` traversal in transcript reads; add `_is_valid_mempal_dir()` to validate MEMPAL_DIR contains project markers before subprocess spawn; canonicalize all paths via `os.path.realpath()`
- **mcp_server**: Replace `str(e)` with `"Internal error"` in 12 `except Exception` handlers to prevent leaking file paths/internal state to MCP clients (keeps `str(e)` in `ValueError` handlers for user-facing validation)
- **config**: Create `config.json` atomically with `os.open(O_CREAT|O_EXCL, 0o600)` to eliminate TOCTOU permission window; write `people_map.json` with `0o600` permissions (contains PII)

## Context
Full security audit of the codebase found 0 Critical, 3 High, 6 Medium, 6 Low findings. This PR addresses all 3 High and 3 trivial Medium fixes. Remaining Medium/Low findings are documented in the audit report (informational only).

## Test plan
- [x] All 864 baseline tests pass (0 regressions)
- [x] Updated 4 test fixtures to create `.git` marker dirs for new `_is_valid_mempal_dir()` validation
- [x] Verified via background agent that all fixes are correct and complete

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1053, opened 2026-04-20
fix(hooks): auto-mine transcripts in convos mode with stable wing

kasparas-anusauskas:
## Problem

The plugin's Stop/PreCompact hooks auto-mine Claude Code (and Codex) session transcripts on every save trigger, but the invocation is hardcoded as `mempalace mine <dir>` with **no `--mode`** flag. The CLI default is `projects` mode, which treats the JSONL as a source file.

**The primary defect is chunk quality.** Projects mode runs the project-file chunker over the JSONL, producing 800-char sliding windows over raw `{"role": "…", "content": "…"}` records. Drawers cut mid-JSON and mid-sentence:

    }, {"role": "user", "content": "Can you help me understand how\nthe system works?"}, {"role": "assistant", "content": "S

These drawers are BM25-findable but their embeddings represent structural boilerplate, and when they surface in search results the returned text is unreadable — users can't reconstruct what was actually said. That defeats the point of storing the transcript.

Convos mode chunks via [`chunk_exchanges`](https://github.com/MemPalace/mempalace/blob/develop/mempalace/convo_miner.py#L98) after normalizing the JSONL into `>`-prefixed turn markers, emitting one semantically whole Q+A exchange per drawer. **That's the chunk-level quality difference this PR delivers.**

**Secondary cleanup** also fixed:
- **Wing name.** With no `mempalace.yaml` in the transcript folder, projects mode falls back to the encoded-path directory name — producing wings like `-Users-you-Projects-my_app`. This PR defaults to the stable `conversations` wing.
- **Room classification.** Projects mode without a yaml puts every drawer in `room:general`. Convos mode's `detect_convo_room` picks a topical room per session from `{technical, architecture, planning, decisions, problems}`. **Note on room-per-session semantics:** `detect_convo_room` is called **once per JSONL file** ([convo_miner.py:442-446](https://github.com/MemPalace/mempalace/blob/develop/mempalace/convo_miner.py#L442-L446)), so all drawers from a single session share one room by design. Per-chunk room routing requires `--extract general` (exposed here via the `MEMPAL_EXTRACT` env var).

### Evidence (from a user's `~/.mempalace/hook_state/hook.log` on current `develop` @ 32ec74d)

    [19:28:59] TRIGGERING SAVE at exchange 80
      No mempalace.yaml found in /Users/…/.claude/projects/-Users-…-Projects-mempalace
      — using auto-detected defaults (wing='-Users-…-Projects-mempalace').

      MemPalace Mine
      Wing:    -Users-kasparasanusauskas-Projects-mempalace
      Rooms:   general
      Drawers filed: 1322
      By room: general 3 files

Four independent signals confirm projects-mode fired: the `mempalace.yaml not found` warning only fires in projects mode, `Rooms: general` is the projects-mode single-room fallback, the source files are `.jsonl` session transcripts, and the drawer count matches the orphan wing exactly.

## Prior art

PR #633 (*"feat: two-layer hook capture, auto-mine transcripts"*, @jphein) proposed the same fix, but for the pre-migration standalone bash hooks (`hooks/mempal_save_hook.sh`). It was closed on 2026-04-12 as *"superseded by recently merged PRs to develop"* — but the supersession was architectural only. PRs #840 and #863 moved hook logic from bash into `mempalace/hooks_cli.py` and carried the *infrastructure* forward but not the *correctness fixes*:

- No `--mode convos` flag on the auto-mine command anywhere in the plugin codepath on `develop`.
- No stable-wing default (`"conversations"` in #633) was carried forward.
- `grep "wing.*conversations\|chunk_exchanges\|PYMINE" hooks/mempal_save_hook.sh` on `develop` returns zero matches.

This PR completes that migration — brings the correctness payload onto the `hooks_cli.py` codepath that the installed plugin actually uses.

### `mempalace/hooks_cli.py`

- **`_get_mine_target(transcript_path)`** returns `(mine_dir, source)` where `source` is `"mempal_dir"`, `"transcript"`, or `""`. Callers now know whether the mine dir came from a user-configured `MEMPAL_DIR` or was derived from the harness's transcript path.
- **`_get_mine_dir()`** kept as a thin wrapper over `_get_mine_target` for backward compatibility (tests and external callers unaffected).
- **`_build_mine_cmd(mine_dir, source)`** builds the `mempalace mine` argv, driven by three new optional env vars:

  | Env var | Values | Default | Effect |
  |---|---|---|---|
  | `MEMPAL_MODE` | `projects` / `convos` / `auto` | `auto` | Auto picks `convos` for transcript-sourced mines, `projects` for `MEMPAL_DIR`-sourced. |
  | `MEMPAL_WING` | any string | `conversations` (transcript source only) | Explicit wing override. |
  | `MEMPAL_EXTRACT` | `exchange` / `general` | miner default (`exchange`) | `exchange` → one room per file via `detect_convo_room`; `general` → per-chunk rooms via `general_extractor.memory_type`. |

  Auto-detect rule (default):
  - `source=transcript` → `--mode convos --wing conversations` (exchange-pair chunking, stable wing)
  - `source=mempal_dir` → `--mode projects` (wing left to the miner, preserving existing behavior)

- **Unknown values for `MEMPAL_MODE` / `MEMPAL_EXTRACT` fall back to auto-detect / miner-default** with a warning in `hook.log`. They are never silently dropped — silent-dropping `--mode` would let the CLI's `projects` default reassert and reintroduce the original bug.

- Both auto-mine spawn sites (`_maybe_auto_ingest`, `_mine_sync`) now use `_get_mine_target` + `_build_mine_cmd`.

### `mempalace/cli.py`

- **Shared constants `MINE_MODES` and `MINE_EXTRACTS`** added at module level as the single source of truth. `add_argument` now uses `choices=list(MINE_MODES)` / `list(MINE_EXTRACTS)`. `hooks_cli._build_mine_cmd` imports the same tuples. Adding a new mode to the CLI propagates automatically to the hook validator.

### `tests/test_hooks_cli.py`

- 12 new tests covering `_get_mine_target` (all four branches) and `_build_mine_cmd` (auto-detect for each source, every env-var override, invalid-value fallback + log-warning emission, extract-only-in-convos).
- One existing test (`test_precompact_mines_transcript_dir`) updated: it asserted `mine_dir` was the last argv entry, which no longer holds now that the command grows flags. It now locates `mine_dir` by position after `"mine"`.

## Compatibility

- **`MEMPAL_DIR` behavior unchanged by default.** Users who point `MEMPAL_DIR` at a project directory keep projects-mode and existing wing-derivation rules.
- **`_get_mine_dir()` signature preserved**, so PRs in flight that import it (e.g. #893) continue to compile.
- **Transcript auto-mines on upgrade:** the first post-upgrade mine of a live session's `.jsonl` will delete-and-reinsert its drawers under the new `conversations` wing via the existing `file_already_mined` + mtime-rebuild path ([convo_miner.py:320-333](https://github.com/MemPalace/mempalace/blob/develop/mempalace/convo_miner.py#L320-L333)). Pre-existing drawers under ugly wings on static/old sessions remain in place — documented behavior of the purge-on-change pattern, not new to this PR.

## Relationship to other PRs

- **Complementary with #673** (*"deterministic hook saves"*, open) — that PR adds a parallel `_ingest_transcript` Python-API path and does not modify `_maybe_auto_ingest` / `_spawn_mine`. This PR fixes the existing subprocess path; the two can land in either order without conflict.
- **Does not conflict with #893** (hook path-traversal hardening) — that PR operates on `_is_safe_transcript_path` and leaves the spawn sites alone.
- **Follows #985's pattern** — env-var-driven hook configuration.

### Primary benefit — chunk quality

**Before (projects mode on JSONL, 800-char windows over structural noise):**

    }, {"role": "user", "content": "Can you help me understand how\nthe system works?"}, {"role": "assistant", "content": "S

**After (convos mode with exchange-pair chunking):**

    > Can you help me understand how the system works?

    Sure, here's how it works: [full coherent response preserved in one drawer]

### Secondary — wing and room

**Before (current `develop`):**

    Wing:    -Users-kasparasanusauskas-Projects-mempalace
    Rooms:   general

**After (this branch, default `--extract exchange`):**

    Wing:    conversations
    Room:    one of { technical | architecture | planning | decisions | problems }
             — detect_convo_room is called once per JSONL (convo_miner.py:442-446),
               so all drawers from a single session share one room. This is by
               design for the default extractor. Set MEMPAL_EXTRACT=general for
               per-chunk rooms (decision / preference / milestone / problem / emotional).
    Chunks:  exchange pairs (one > turn + AI response per drawer) instead of raw
             800-char windows over JSONL structural noise.

### Test + lint

Tests: `pytest tests/ --ignore=tests/benchmarks -q` → **1047 passed**.
Lint: `ruff format --check` + `ruff check` → clean.

## Credit

Approach (convos mode + stable `conversations` wing default) follows @jphein's design in #633. This PR brings it onto the current `hooks_cli.py` architecture that #840 / #863 established.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -37,6 +37,11 @@ from .config import MempalaceConfig
 from .version import __version__
 
 
+# Valid values for `mempalace mine --mode` and `--extract`. Shared with
+# hooks_cli.py so the auto-mine env-var validation tracks the CLI surface.
+MINE_MODES = ("projects", "convos")
+MINE_EXTRACTS = ("exchange", "general")
+
 _MEMPALACE_PROJECT_FILES = ("mempalace.yaml", "entities.json")
 
 
@@ -557,7 +562,7 @@ def main():
     p_mine.add_argument("dir", help="Directory to mine")
     p_mine.add_argument(
         "--mode",
-        choices=["projects", "convos"],
+        choices=list(MINE_MODES),
         default="projects",
         help="Ingest mode: 'projects' for code/docs (default), 'convos' for chat exports",
     )
@@ -584,7 +589,7 @@ def main():
     )
     p_mine.add_argument(
         "--extract",
-        choices=["exchange", "general"],
+        choices=list(MINE_EXTRACTS),
         default="exchange",
         help="Extraction strategy for convos mode: 'exchange' (default) or 'general' (5 memory types)",
     )
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -138,16 +138,71 @@ def _output(data: dict):
     print(json.dumps(data, indent=2, ensure_ascii=False))
 
 
-def _get_mine_dir(transcript_path: str = "") -> str:
-    """Determine directory to mine from MEMPAL_DIR or transcript path."""
+# Default wing for transcript-sourced auto-mines. Overridable via MEMPAL_WING.
+_TRANSCRIPT_DEFAULT_WING = "conversations"
+_AUTO_MODE = "auto"
+
+
+def _get_mine_target(transcript_path: str = "") -> tuple:
+    """Return (mine_dir, source) where source is 'mempal_dir', 'transcript', or ''."""
     mempal_dir = os.environ.get("MEMPAL_DIR", "")
     if mempal_dir and os.path.isdir(mempal_dir):
-        return mempal_dir
+        return mempal_dir, "mempal_dir"
     if transcript_path:
         path = Path(transcript_path).expanduser()
         if path.is_file():
-            return str(path.parent)
-    return ""
+            return str(path.parent), "transcript"
+    return "", ""
+
+
+def _get_mine_dir(transcript_path: str = "") -> str:
+    """Legacy wrapper — returns only the mine dir. Prefer _get_mine_target."""
+    mine_dir, _ = _get_mine_target(transcript_path)
+    return mine_dir
+
+
+def _build_mine_cmd(mine_dir: str, source: str) -> list:
+    """Build `mempalace mine` argv, configurable via env vars.
+
+    MEMPAL_MODE    — one of MINE_MODES, or 'auto' (default). Auto picks convos
+                     for transcript-sourced mines, projects otherwise. Unknown
+                     values fall back to auto (with a warning in hook.log).
+    MEMPAL_WING    — explicit wing name. Transcript-sourced mines default to
+                     _TRANSCRIPT_DEFAULT_WING when unset; MEMPAL_DIR-sourced
+                     mines leave --wing out so the miner derives it.
+    MEMPAL_EXTRACT — one of MINE_EXTRACTS. Applied only in convos mode.
+                     Unknown values are dropped with a warning.
+    """
+    # Local import: cli.py imports run_hook from this module.
+    from .cli import MINE_EXTRACTS, MINE_MODES
+
+    cmd = [sys.executable, "-m", "mempalace", "mine", mine_dir]
+
+    mode = os.environ.get("MEMPAL_MODE", _AUTO_MODE).strip().lower() or _AUTO_MODE
+    if mode != _AUTO_MODE and mode not in MINE_MODES:
+        _log(
+            f"MEMPAL_MODE={mode!r} not in {MINE_MODES + (_AUTO_MODE,)}; falling back to auto-detect"
+        )
+        mode = _AUTO_MODE
+    if mode == _AUTO_MODE:
+        mode = "convos" if source == "transcript" else "projects"
+    cmd.extend(["--mode", mode])
+
+    wing = os.environ.get("MEMPAL_WING", "").strip()
+    if not wing and source == "transcript":
+        wing = _TRANSCRIPT_DEFAULT_WING
+    if wing:
+        cmd.extend(["--wing", wing])
+
+    if mode == "convos":
+        extract = os.environ.get("MEMPAL_EXTRACT", "").strip().lower()
+        if extract and extract not in MINE_EXTRACTS:
+            _log(f"MEMPAL_EXTRACT={extract!r} not in {MINE_EXTRACTS}; using miner default")
+            extract = ""
+        if extract:
+            cmd.extend(["--extract", extract])
+
+    return cmd
 
 
 _MINE_PID_FILE = STATE_DIR / "mine.pid"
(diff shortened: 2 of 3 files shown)


igorls:

jphein:
Operator follow-up — we landed an architecturally adjacent change in our [techempower-org/palace-daemon](https://github.com/techempower-org/palace-daemon) fork yesterday (2026-05-14) that supersedes the hook side of what this PR is fixing. Sharing in case the design point is useful.

## What we did

Same problem class — hooks were producing low-quality drawers because of how they wrote — but we took it a step further by **collapsing the two parallel write paths into one**.

Before:
- Hook calls `mempalace_diary_write` MCP tool → 1 summary drawer per session
- Hook ALSO calls `mempalace mine <dir>` (no --mode) → N chunked drawers per file with projects-mode chunking

Both paths emit drawers, with different wing/room/ID conventions, different metadata fields, different chunking behavior.

After (our Phase 1D refactor):
- Hook is **trigger-only** — it computes cadence, picks a wing, POSTs `/mine` to palace-daemon with `--mode convos`
- Miner is the **sole writer** — all drawer-shape decisions live there
- `mempalace_diary_write` becomes a deprecated legacy shim calling miner internally

So all your fixes here (`--mode convos`, stable wing, room classification) become invariants enforced by the single write path rather than per-hook configuration. Hook hosts that haven't been updated automatically get the right behavior because they don't decide anything — the miner does.

## Wing derivation specifically

We also tightened wing derivation. Our [hook.py `_project_wing`](https://github.com/techempower-org/palace-daemon/blob/main/clients/hook.py) does:
1. Explicit `--wing` flag
2. `mempalace.yaml` in project dir
3. Claude Code transcript path: strip `~/.claude/projects/`, strip leading `-`, replace `-` → `/`, take final basename → slug-normalize. This handles the hyphenated-project-name case from #1424 generally rather than as a special case.
4. cwd basename → slug-normalize
5. fallback: `"unknown"` (not `conversations` — `conversations` collides with real wings on multi-machine setups)

## Trade-offs

The "single write path" approach is more invasive than just adding `--mode convos` to the hook. We accepted that cost because we were also adding canonical-room taxonomy enforcement at the same time, which similarly needs a single point of validation. If your design is "hook + miner stay separate", the change here is the right shape for your codebase.

Happy to share the full refactor diff if useful. Operator details in our changelog at [familiar.realm.watch CHANGELOG.md](https://github.com/jphein/familiar.realm.watch/blob/main/CHANGELOG.md#unreleased--2026-05-14-later--hybrid-search--taxonomy--unified-write-path).

PR #1054, opened 2026-04-20
fix(db): paginate collection fetches to avoid SQL variable limit

airmonitor:
- Replace single-shot col.get(limit=total) calls with paginated requests
- Batch size is set to 1000 to stay safely below ChromaDB's SQLite bound-variable cap
- Add unit test to verify pagination logic and ensure complete data retrieval at scale

## Checklist
- [ ] No hardcoded paths

### DIFF
--- a/mempalace/closet_llm.py
+++ b/mempalace/closet_llm.py
@@ -221,7 +221,18 @@ def regenerate_closets(
         print("No drawers in palace.")
         return {"processed": 0}
 
-    all_data = drawers_col.get(limit=total, include=["documents", "metadatas"])
+    # Paginate: one-shot get(limit=total) exceeds chroma's SQLite bound-variable
+    # cap at large palace sizes (reproduced at ~12k drawers on chroma 1.5.8).
+    all_data: dict = {"ids": [], "documents": [], "metadatas": []}
+    offset = 0
+    while offset < total:
+        batch = drawers_col.get(limit=1000, offset=offset, include=["documents", "metadatas"])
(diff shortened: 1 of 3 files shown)


Qodo-Free-For-OSS:
Hi, The new pagination test doesn’t assert that `status()` passes increasing `offset=` values to `col.get()`, so a regression that drops the offset kwarg could repeatedly fetch the first page and still pass while printing “12000 drawers”.

**Severity:** remediation recommended | **Category:** maintainability

**How to fix:** Assert offsets advance in test

> ### Issue description
> `test_status_paginates_instead_of_one_shot_fetch` records `(limit, offset)` but doesn’t assert offset progression. This allows a regression where `status()` forgets to pass `offset=...` to still pass the test while repeatedly fetching the first page.
> 
> ### Issue Context
> The fake collection defaults `offset` to 0 when not provided. Because `status()` increments its local `offset` by the *returned* page size, the loop will still terminate and print the expected drawer count even if every `get()` call used `offset=0`.
> 
> ### Fix Focus Areas
> - tests/test_miner.py[376-420]
> 
> ### What to change
> - Add assertions that:
>   - `fake.get_calls[0][1] == 0`
>   - offsets are strictly increasing (or at least not all equal)
>   - offsets match expected page boundaries for the chosen batch size (e.g., `[0, 1000, 2000, ...]` for total 12_000)
> - Optionally also assert the final offset reaches `>= total` or that the last call returns the tail page size.

---

PR #1016, opened 2026-04-18
fix(status): paginate col.get() to stay under SQLite variable limit

eldar702:
Fixes #1015.

## Summary

`mempalace status` crashes on any palace with more than ~32 766 drawers with:

```
chromadb.errors.InternalError: Error executing plan:
  Internal error: error returned from database:
  (code: 1) too many SQL variables
```

Root cause: `miner.py::status()` calls `col.get(limit=total)` unbounded. Chroma issues one SQL statement whose host-parameter count grows with the limit, and once `total > 32 766` (SQLite's default `SQLITE_MAX_VARIABLE_NUMBER`) SQLite rejects it.

## Change

Paginate `col.get()` in 10 000-drawer batches and aggregate `metadatas`. No behavior change for smaller palaces.

```diff
-    # Count by wing and room
+    # Count by wing and room. Paginate col.get() to stay under SQLite's
+    # SQLITE_MAX_VARIABLE_NUMBER (default 32766). A single col.get(limit=total)
+    # raises `too many SQL variables` once a palace exceeds that many drawers.
     total = col.count()
-    r = col.get(limit=total, include=["metadatas"]) if total else {"metadatas": []}
-    metas = r["metadatas"]
+    metas: list = []
+    if total:
+        BATCH = 10000
+        for offset in range(0, total, BATCH):
+            r = col.get(limit=BATCH, offset=offset, include=["metadatas"])
+            metas.extend(r.get("metadatas", []) or [])
```

One file touched, 9 added / 3 removed.

## Reproducer

On any palace with > 32 766 drawers:

```
mempalace status    # crashes with the traceback above
```

Minimal programmatic repro:

```python
from mempalace.config import MempalaceConfig
from mempalace.palace import get_collection
col = get_collection(MempalaceConfig().palace_path)
col.get(limit=col.count(), include=["metadatas"])
# → InternalError: too many SQL variables
```

Exact inflection point:
```
limit=32766  → ok (32766 rows)
limit=33000  → FAIL: too many SQL variables
```

## Test plan

- [x] `mempalace status` succeeds on a 43 852-drawer palace after the patch (produced the normal Wing/Room breakdown)
- [x] `mempalace status` succeeds on an empty palace
- [x] No change in output format (verified by diffing against a < 32 K drawer palace before / after)
- [ ] Existing test suite (`pytest`) — not run locally; maintainers please verify in CI

## Why 10 000

[...]

PR #1081, opened 2026-04-21
fix(search): hint at `mempalace repair` when filtered query hits HNSW index mismatch (#1035)

### LINKED ISSUE #1035: bug: existing 0.6.x palaces need `mempalace repair` after #1010 — filtered `query(where=...)` fails with 'Error finding id' until HNSW index is rebuilt
## Summary

After upgrading an existing 0.6.x palace via the `chromadb>=1.5.4,<2` pin landed in #1010, `mempalace search` (and any code path that calls `col.query(query_texts=..., where={...})`) fails with:

```
chromadb.errors.InternalError: Error executing plan: Internal error: Error finding id
```

…until `mempalace repair` is run against the upgraded palace. Unfiltered queries (`col.query(query_texts=...)` with no `where`) work fine even before the repair.

Because `mempalace search --wing X` and `mempalace search --wing X --room Y` always build a `where` filter (`searcher.py:251` → `build_where_filter`), essentially every upgrading user will hit this on their first filtered search. The CLI reports the error as "Search error: …" and returns non-zero, with no hint about running repair.

## Repro

Starting state: a palace built over a long period on `chromadb==0.6.3`. Upgrade in place:

```bash
pipx inject --force mempalace 'chromadb>=1.5.4,<2'
mempalace migrate --dry-run
# → "Palace is already readable by chromadb 1.5.8. N drawers found. No migration needed."

mempalace search "encryption" --wing my_project
#   Search error: Error executing plan: Internal error: Error finding id

mempalace search "encryption"                 # no filter
#   Returns results correctly.

mempalace repair --yes
#   Re-files all N drawers, backs up to palace.backup, swaps in place.

mempalace search "encryption" --wing my_project
#   Works.
```

### Direct chromadb probe (confirms it's the filter path, not the palace contents)

```python
import chromadb
c = chromadb.PersistentClient(path='/Users/.../palace')   # palace built on 0.6.3
col = c.get_collection('mempalace_drawers')

col.count()                                                # 36621 ✅
col.get(limit=5, include=['metadatas'])                    # ✅
col.query(query_texts=['encryption'], n_results=3)         # ✅
col.query(query_texts=['encryption'], n_results=3,
          where={'wing': 'my_project'})
# → InternalError: Error executing plan: Internal error: Error finding id
```

The SQLite source of truth (`chroma.sqlite3`) is complete (74,911 embedding rows, 38,762 unique embedding_ids, 36,621 live drawers reported by `col.count()`). It's the old 0.6.x HNSW index that chromadb 1.5.x's Rust query path can't combine with a metadata filter.

`mempalace repair` fixes it because it rebuilds the vector index from sqlite:

```
Drawers found: 36621
Extracting drawers...
Extracted 36621 drawers
Backing up to /Users/.../palace.backup...
Rebuilding collection...
Re-filed 5000/36621 drawers...
...
Repair complete. 36621 drawers rebuilt.
```

## Why this wasn't caught in #1010's validation

The 50K-drawer migration test referenced in #1010 validated `count()` and unfiltered reads, which matches the maintainer claim of "zero API breakage". It looks like no filtered `query(where={...})` was exercised against a 0.6.x-built HNSW index. Fresh palaces created under 1.5.x don't have this problem — it only affects users with pre-existing 0.6.x palaces.

## Suggested fix

One of:

1. **Auto-repair on version mismatch.** `ChromaBackend._client()` already runs `_fix_blob_seq_ids` preemptively (`backends/chroma.py:14`). A similar "if we detect a 0.6.x-built HNSW in the palace dir and chromadb is 1.5.x, run repair before returning the client" check would make the upgrade seamless.
2. **Release-note warning.** Release notes for the version that ships the 1.5.x pin should tell existing users to run `mempalace repair` once after upgrade. Ideally the CLI also detects the mismatch and prints a hint on the first filtered-query failure.
3. **Call it from `mempalace migrate` when source_version is 0.6.x and target is 1.5+.** Today `migrate` early-returns when chromadb can read the palace ("already readable") — but readable ≠ filter-queryable against the old index. Extending `migrate` to run the repair pass in that case would keep a single user-facing upgrade command.

Option 1 is the most user-friendly; option 3 is the most consistent with the existing `migrate` flow.

## Environment

- mempalace: 3.3.1 (pipx)
- chromadb: post-upgrade 1.5.8 (source: 0.6.3)
- Python: 3.12.7
- macOS arm64 (Darwin 25.3.0)
- Palace size: 36,621 drawers, 525 MB `chroma.sqlite3`

## Happy to help

Can put up a PR for option 1 or option 3 if the maintainers have a preferred direction.

potterdigital:
## Summary

Fixes #1035. When a palace's HNSW index predates 1.0 (i.e. built under
`chromadb` 0.6.x), a filtered search on 1.5.x raises an opaque
`ValueError: Error finding id ...`. The repair command already handles
the actual upgrade (`mempalace repair`), but users upgrading in place
have no way to discover that from the error alone.

This PR catches the exception in both query paths in `searcher.py`,
detects the specific signature (a `where` filter was in play **and**
the exception message contains `"Error finding id"`), and surfaces a
hint pointing at `mempalace repair` before re-raising (CLI) or
returning the error dict (API / MCP). The no-palace path already pairs
`error` + `hint` in the returned dict, so the API branch mirrors that
shape.

## Changes

- `mempalace/searcher.py` — detect-and-hint inside the existing
  `except` blocks in `search()` (CLI) and `search_memories()` (API),
  guarded by `where and "Error finding id" in str(e)`
- `CHANGELOG.md` — one-line entry under `### Bug Fixes` in the
  `[Unreleased] — v3.3.0` section
- `tests/test_searcher.py::TestHNSWRepairHint` — three focused tests:
  CLI positive, API positive, and a negative-guard test to ensure
  unrelated failures don't get a misleading repair suggestion

All 1060 tests in the full suite pass. `ruff check` and
`ruff format --check` both clean on the touched files.

## Scope note for the maintainer

In the issue thread, #851 and #1016 were mentioned as related —
happy to be corrected, but my read is those address SQLite parameter
limits in `col.get()` pagination, which is a different code path from
the HNSW `col.query()` failure this PR addresses. If you'd like this
to also cover the pagination path, I'm glad to extend the PR; I kept
it narrow since your earlier reply greenlit a ~5-line detect-and-hint
specifically against the 0.6.x→1.5.x HNSW failure.

## Known related gap

The `mempalace repair` command this hint points at currently
segfaults on a sufficiently corrupt HNSW index — `repair` calls
`col.count()` before doing any repair work, and that's the call that
crashes on drift. That's tracked separately in #1108 (wiring
`quarantine_stale_hnsw()` from #1000 into the backend open path). On
palaces where the index is merely stale (not corrupt enough to crash
`count()`), `repair` works as intended and this hint is actionable;
#1108 closes the remaining case.

## Test plan

- [x] `pytest tests/test_searcher.py -v` — 24 passed (3 new)
- [x] `pytest -x` — 1060 passed
- [x] `ruff check mempalace/searcher.py tests/test_searcher.py`
- [x] `ruff format --check mempalace/searcher.py tests/test_searcher.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Sync `version.py` to match `pyproject.toml` (#820)
 - Remove unused `main` import from `mempalace/__init__.py` (#827)
 - README audit — fix 7 stale claims (tool count, version badge, wake-up token cost, `dialect.py` lossless disclaimer, `pyproject.toml` version) with 42 regression-guard tests (#835)
+- Surface a `mempalace repair` hint when a filtered search hits an HNSW index from a pre-1.0 palace, instead of raising an opaque ChromaDB `Error finding id` (#1035)
 
 ### Improvements
 - Optimize entity detection with regex caching and pre-compilation (#828)
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -262,6 +262,11 @@ def search(query: str, palace_path: str, wing: str = None, room: str = None, n_r
         results = col.query(**kwargs)
 
     except Exception as e:
+        if where and "Error finding id" in str(e):
+            print(
+                "\n  Hint: this palace's HNSW index predates 1.0 and fails on "
+                "filtered queries.\n  Run: mempalace repair <palace_path>"
+            )
         print(f"\n  Search error: {e}")
         raise SearchError(f"Search error: {e}") from e
 
@@ -352,6 +357,11 @@ def search_memories(
             dkwargs["where"] = where
         drawer_results = drawers_col.query(**dkwargs)
     except Exception as e:
+        if where and "Error finding id" in str(e):
+            return {
+                "error": f"Search error: {e}",
+                "hint": "This palace's HNSW index predates 1.0. Run: mempalace repair <palace_path>",
+            }
         return {"error": f"Search error: {e}"}
 
     # Gather closet hits (best-per-source) to build a boost lookup.
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -265,7 +265,7 @@ def search(query: str, palace_path: str, wing: str = None, room: str = None, n_r
         if where and "Error finding id" in str(e):
             print(
                 "\n  Hint: this palace's HNSW index predates 1.0 and fails on "
-                "filtered queries.\n  Run: mempalace repair <palace_path>"
+                f"filtered queries.\n  Run: mempalace repair {palace_path}"
             )
         print(f"\n  Search error: {e}")
         raise SearchError(f"Search error: {e}") from e
@@ -360,7 +360,7 @@ def search_memories(
         if where and "Error finding id" in str(e):
             return {
                 "error": f"Search error: {e}",
-                "hint": "This palace's HNSW index predates 1.0. Run: mempalace repair <palace_path>",
+                "hint": f"This palace's HNSW index predates 1.0. Run: mempalace repair {palace_path}",
             }
         return {"error": f"Search error: {e}"}
 
--- a/tests/test_searcher.py
+++ b/tests/test_searcher.py
@@ -191,3 +191,51 @@ class TestSearchCLI:
         assert "[2]" in captured.out
         # Second result renders with fallback '?' values instead of crashing
         assert "second doc" in captured.out
+
+
+# ── HNSW / pre-1.0 index-mismatch repair hint (#1035) ─────────────────
+
+
+class TestHNSWRepairHint:
+    """When a filtered query hits a pre-1.0 palace, ChromaDB raises
+    `Error finding id ...`. Surface a `mempalace repair` hint instead
+    of leaking the opaque error alone.
+    """
+
+    def test_cli_filtered_query_hints_repair(self, capsys):
+        """CLI path: filtered search + 'Error finding id' → hint printed."""
+        mock_col = MagicMock()
+        mock_col.query.side_effect = RuntimeError("Error finding id 1234")
+
+        with patch("mempalace.searcher.get_collection", return_value=mock_col):
+            with pytest.raises(SearchError, match="Search error"):
+                search("test", "/fake/path", wing="project")
+
+        captured = capsys.readouterr()
+        assert "mempalace repair" in captured.out
+
+    def test_api_filtered_query_returns_hint(self):
+        """API path: filtered search + 'Error finding id' → dict has hint."""
+        mock_col = MagicMock()
+        mock_col.query.side_effect = RuntimeError("Error finding id 1234")
+
+        with patch("mempalace.searcher.get_collection", return_value=mock_col):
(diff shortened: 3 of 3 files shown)


jphein (approved):
Reviewed on `review/1081` branch against `upstream/develop`:

- All three `TestHNSWRepairHint` tests pass locally (`pytest tests/test_searcher.py::TestHNSWRepairHint -v`).
- Full `tests/test_searcher.py` suite: 24 passed.
- Diff reads clean — guard `where and "Error finding id" in str(e)` is correctly narrow; CLI prints hint before re-raising, API path returns it as a first-class `hint` key so MCP-server consumers don't have to parse stderr.

On the live-data smoke: my own palace (1.5.x, 165K drawers) doesn't reproduce the stale-HNSW failure mode the hint targets, so the hint is unreachable on my corpus by design — I can't add a "confirmed against real palace" data point the way I had hoped. Left the review as tests-and-diff-only for that reason; no changes requested.

Qodo-Free-For-OSS:
Hi, Both the CLI and API/MCP hint text tells users to run `mempalace repair <palace_path>` but never substitutes the actual `palace_path`, reducing the hint’s usefulness even though the value is available in scope.

**How to fix:** Interpolate palace_path in hint

> ### Issue description
> The repair hint currently includes a literal `<palace_path>` placeholder in both CLI output and the API return dict, even though `palace_path` is available.
> 
> ### Issue Context
> This hint is shown specifically to help users upgrade pre-1.0 palaces. Providing the concrete path makes the suggested command directly copy/pastable.
> 
> ### Fix Focus Areas
> - mempalace/searcher.py[264-271]
> - mempalace/searcher.py[359-365]
> 
> ### Suggested change
> - Update the printed/returned hint to use an f-string, e.g. `mempalace repair {palace_path}` (optionally `os.path.expanduser(palace_path)` if you want it normalized).

---

potterdigital:
Fixed in 54029c7 — both hint strings now use an f-string to interpolate `palace_path`, so the suggested command is directly copy-pasteable. Both positive tests in `TestHNSWRepairHint` were strengthened to assert the interpolated path appears **and** that the literal `<palace_path>` placeholder does not, guarding against the same regression in future edits.

> We noticed a couple of other issues in this PR as well - happy to share if helpful.

@qodo-ai-reviewer yes please — go ahead and surface the others.

jphein (approved):
Re-reviewed `54029c7` against the 2026-04-21 approval baseline.

The change is small (2 lines in `searcher.py`, strengthened tests in `test_searcher.py`): replaces the literal `<palace_path>` placeholder in both hint sites with f-string interpolation, so the suggested command is directly copy-pasteable. Both positive tests in `TestHNSWRepairHint` now assert both (a) the interpolated path appears and (b) the literal `<palace_path>` doesn't leak — good regression guard for future edits to the hint text.

No changes to the guard logic (`where and "Error finding id" in str(e)`) I approved earlier. Re-approving.

igorls:
Thanks! develop already surfaces a 'mempalace repair' hint on HNSW divergence (searcher.py, via #1222). Closing as overlapping — if there is a path that still does not hint, please reopen with that case.

PR #1085, opened 2026-04-21
perf: batch ChromaDB inserts in miner - 10-30x faster mining

midweste:
### Problem

`process_file()` called `add_drawer()` once per chunk, resulting in:
- **N separate ONNX embedding passes** per file (one per chunk)
- **N separate ChromaDB upsert calls** per file
- **N redundant `datetime.now()` and `os.path.getmtime()` syscalls** per file

On large files (1000+ chunks), this made mining slow.

### Solution

**1. DRY metadata construction via `_build_drawer()` helper**

Extracted the drawer ID generation and metadata assembly into a shared `_build_drawer()` function used by both:
- `add_drawer()` — single-insert path (unchanged contract, used by MCP tools)
- `add_drawers()` — new batch-insert path (used by `process_file()`)

**2. Batched upserts with sub-batching**

`add_drawers()` collects all chunks into batch lists and upserts in groups of `CHROMA_BATCH_LIMIT` (5000) to stay under ChromaDB's hard cap of 5,461.

**3. Hoisted syscalls**

`datetime.now()` and `os.path.getmtime()` are called once per file and shared across all chunks via `_build_drawer(now=..., source_mtime=...)`.

### Results

| Metric | Before | After |
|--------|--------|-------|
| ONNX passes per file | N (per chunk) | ⌈N/5000⌉ |
| DB writes per file | N | ⌈N/5000⌉ |
| Syscalls per file | 2N | 2 |
| Metadata construction | duplicated in `add_drawer` + `process_file` | single `_build_drawer` helper |

Tested against a 499-file project (~70k drawers) with zero errors, including an 18,803-chunk benchmark file.

### Breaking Changes

None. `add_drawer()` retains its existing signature and behavior. `add_drawers()` is additive.

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -1134,14 +1134,28 @@ class ChromaBackend(BaseBackend):
         return ChromaCollection(collection)
 
     def close_palace(self, palace) -> None:
-        """Drop cached handles for ``palace``. Accepts ``PalaceRef`` or legacy path str."""
+        """Drop cached handles for ``palace``. Accepts ``PalaceRef`` or legacy path str.
+
+        Calls ``client.close()`` first to flush ChromaDB's background
+        compactor — failing to do so corrupts the HNSW index on exit.
+        """
         path = palace.local_path if isinstance(palace, PalaceRef) else palace
         if path is None:
             return
-        self._clients.pop(path, None)
+        client = self._clients.pop(path, None)
+        if client is not None:
+            try:
+                client.close()
+            except Exception:
+                logger.warning("Failed to close ChromaDB client for %s", path, exc_info=True)
         self._freshness.pop(path, None)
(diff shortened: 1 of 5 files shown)


midweste:
Closing this PR — the core optimization (batched ChromaDB upserts) was independently implemented in #1185 (perf/batched-upsert-gpu), which expanded on the same approach with GPU acceleration support and hardware-aware embedding functions.

All key improvements from this PR are now covered in develop:

DRY metadata builder (_build_drawer_metadata)
Batched collection.upsert() per file instead of per-chunk
Hoisted os.path.getmtime() out of the per-chunk loop
Sub-batching via DRAWER_UPSERT_BATCH_SIZE to stay under ChromaDB's hard cap
Thanks for landing the expanded version! 🎉

PR #1097, opened 2026-04-22
fix: treat empty string as no filter in mempalace_search wing/room

### LINKED ISSUE #1084: bug: mempalace_search rejects empty-string wing/room with "must be a non-empty string" even though schema marks them optional
## Summary

The \`mempalace_search\` tool marks \`wing\` and \`room\` as optional in its schema, but the handler rejects empty-string values with \`{"error": "wing must be a non-empty string"}\`. This is confusing and trips up LLM agents that default to filling every declared parameter.

## Reproduction

```bash
# Call via MCP stdio (or from Claude Code / OpenCode / any MCP client)
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "tools/call",
  "params": {
    "name": "mempalace_search",
    "arguments": {
      "query": "some topic",
      "wing": "",
      "room": ""
    }
  }
}
```

Response:

```json
{"error": "wing must be a non-empty string"}
```

Fixing `wing` and retrying with `room: ""` returns the equivalent error for `room`.

## Schema context

From `mempalace_search`'s tool definition (trimmed):

```json
{
  "name": "mempalace_search",
  "inputSchema": {
    "type": "object",
    "properties": {
      "query":  {"type": "string", ...},
      "wing":   {"type": "string", "description": "Filter by wing (optional)"},
      "room":   {"type": "string", "description": "Filter by room (optional)"},
      "limit":  {"type": "integer", ...},
      ...
    },
    "required": ["query"]
  }
}
```

The tool advertises `wing` and `room` as optional (`"(optional)"` in description, not in `required` array), so a caller reasonably expects either omission or empty value to mean "don't filter."

## Why it matters (real-world agent trace)

In an OpenCode session, we give our agent an instructions file containing:

> Before answering... call `mempalace_search` with just the query text (no wing/room filter).

The agent faithfully tried to comply but interpreted "no filter" as "empty string":

```
Call 1: {"query": "TrainTrac Express API Feed Management relationship", "wing": "", "room": ""}
        → error: wing must be a non-empty string

Call 2: {"query": "TrainTrac Feed domain relationship", "wing": "all", "room": ""}
        → error: room must be a non-empty string
```

Agent then gave up on `mempalace_search` and fell back to grep. Two otherwise-productive queries wasted.

This pattern will hit anyone whose agent instructions tell it to skip the filter — models often default to filling every declared field, especially when schemas mark optionality with description text rather than JSON Schema structure.

## Proposed fix

Two options, listed in order of preference:

**A) Treat empty string as "no filter"** — easiest for callers, matches the "optional" intent:

```python
if wing in (None, ""):
    wing = None  # -> no filter
```

**B) Improve the error message** — less breaking, still helpful:

```json
{"error": "wing is empty — omit the 'wing' field from the request to skip filtering, or pass an actual wing name"}
```

Either works. (A) means agents never fail on this, and existing working callers are unaffected.

## Workarounds I used

For now I've pushed a rule tweak in our devbox repo instructing the agent explicitly to omit `wing` / `room` rather than pass empty strings — visible to future readers at [MILCGroup/milc-devbox#130](https://github.com/MILCGroup/milc-devbox/pull/130). Happy to open a PR here too if option (A) looks good.

## Environment

- mempalace 3.3.1 (from `pip install mempalace`, installed ~2 days ago)
- Called via MCP stdio from OpenCode 0.x
- Python 3.10 on Ubuntu 22.04

KunalG67:
Fixes #1084

## What does this PR do?
Treats empty string as `None` (no filter) in `_sanitize_optional_name`. 
Previously, `mempalace_search` rejected empty-string `wing`/`room` values 
with "must be a non-empty string" even though the schema marks them optional.
This broke LLM agents that default to filling every declared parameter with "".

## How to test
Call `mempalace_search` with `wing: ""` and `room: ""` - should return 
results instead of an error. Quick unit check:

from mempalace.mcp_server import _sanitize_optional_name
assert _sanitize_optional_name("") is None
assert _sanitize_optional_name(None) is None
assert _sanitize_optional_name("main") == "main"

## Checklist
- [ ] No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -285,7 +285,7 @@ def _get_cached_metadata(col, where=None):
 
 def _sanitize_optional_name(value: str = None, field_name: str = "name") -> str:
     """Validate optional wing/room-style filters."""
-    if value is None:
+    if value is None or value == "":
         return None
     return sanitize_name(value, field_name)
 
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -285,7 +285,7 @@ def _get_cached_metadata(col, where=None):
 
 def _sanitize_optional_name(value: str = None, field_name: str = "name") -> str:
     """Validate optional wing/room-style filters."""
-    if value is None or value == "":
+    if value is None or not value.strip():
         return None
     return sanitize_name(value, field_name)
 


sha2fiddy:
Hey @KunalG67, your one-liner handles the exact reproduction case from #1084.

Heads-up on one edge case: whitespace-only strings (`"   "`, `"\t"`, `"\n"`) still raise `ValueError` because `sanitize_name` at `mempalace/config.py:27` rejects any string where `not value.strip()`. Agents trip on this the same way they produce `""` (trailing space, stray tab). I've seen it in practice.

Changing the guard to `value is None or not value.strip()` covers both.

I've got four regression tests for this sitting on a local branch (empty + whitespace on `list_rooms`, `list_drawers`, `search`). Want me to paste the diff so you can fold them in, or send as a follow-up after yours lands?

KunalG67:
@sha2fiddy Thanks! Updated to not value.strip() to cover whitespace only strings too. Feel free to paste the test diff ,happy to fold them in

sha2fiddy:
Thanks for jumping on the whitespace case.

The 4 tests below drop into `tests/test_mcp_server.py` right before `test_wal_redacts_sensitive_fields`. They reuse the `_patch_mcp_server` helper already in that file. I ran them locally against your `e27967b` head, all 4 pass.

```python
    # ── #1084: empty/whitespace filters normalize to None ───────────────

    def test_search_accepts_empty_wing_as_no_filter(self, monkeypatch, config, kg):
        """Empty-string wing must be treated as 'no filter', not a validation error.

        LLM agents that eagerly fill every declared parameter pass ``""`` to
        mean "no filter". Regression guard for #1084.
        """
        _patch_mcp_server(monkeypatch, config, kg)
        from mempalace import mcp_server

        captured = {}

        def _fake_search(query, palace_path, wing, room, n_results, max_distance):
            captured["wing"] = wing
            captured["room"] = room
            return {"results": []}

        monkeypatch.setattr(mcp_server, "search_memories", _fake_search)

        result = mcp_server.tool_search(query="JWT", wing="")
        assert "error" not in result
        assert captured["wing"] is None

    def test_search_accepts_whitespace_wing_as_no_filter(self, monkeypatch, config, kg):
        """Whitespace-only wing collapses to None — same rationale as empty string."""
        _patch_mcp_server(monkeypatch, config, kg)
        from mempalace import mcp_server

        captured = {}

        def _fake_search(query, palace_path, wing, room, n_results, max_distance):
            captured["wing"] = wing
            captured["room"] = room
            return {"results": []}

        monkeypatch.setattr(mcp_server, "search_memories", _fake_search)

        result = mcp_server.tool_search(query="JWT", wing="   ", room="\t")
        assert "error" not in result
        assert captured["wing"] is None
        assert captured["room"] is None

    def test_list_drawers_accepts_empty_room_as_no_filter(
        self, monkeypatch, config, palace_path, seeded_collection, kg
    ):
        """Empty-string room on list_drawers must not raise — regression for #1084."""
        _patch_mcp_server(monkeypatch, config, kg)
        from mempalace.mcp_server import tool_list_drawers

        result = tool_list_drawers(wing="", room="")
        assert "error" not in result
        assert "drawers" in result

    def test_list_rooms_accepts_empty_wing_as_no_filter(
        self, monkeypatch, config, palace_path, seeded_collection, kg
    ):
        """Empty-string wing on list_rooms must not raise — regression for #1084."""
        _patch_mcp_server(monkeypatch, config, kg)
        from mempalace.mcp_server import tool_list_rooms

        result = tool_list_rooms(wing="")
        assert "error" not in result
        assert result["wing"] == "all"
```

Rename or trim whatever doesn't fit your style.

Issue #1069, opened 2026-04-21
Consolidate `hooks/mempal_*_hook.sh` into thin wrappers delegating to `mempalace hook run`


sha2fiddy:
After #833 and #942 (both merged 2026-04-21), the repo carries two Stop/PreCompact hook implementations doing the same job with diverging architectures:

| Path | Shape | Stdin parsing | mempalace call |
|---|---|---|---|
| `hooks/mempal_*_hook.sh` (legacy) | Fat bash, ~120 LoC | Inline `python3 -c "import json,sys; ..."` + JSONL heredoc, gated on `MEMPAL_PYTHON` (#833) | `"$MEMPALACE_BIN" mine` (#340) |
| `.claude-plugin/hooks/mempal-*-hook.sh` (plugin) | Thin wrapper, ~23 LoC | Delegated to `mempalace/hooks_cli.py` | `mempalace hook run --hook {stop,precompact} --harness claude-code` with 3-tier fallback (#942) |

Both are user-facing — `mempalace init` wires one or the other depending on install flavor.

**Cost of the split:**

1. Every new knob gets written twice or applies on only one path. `MEMPAL_PYTHON` is inert on the plugin path; `hooks.auto_save` (#711) is only honored wherever the PR author remembers.
2. Open fixes entrench the duplication: #670 threads a `find_python()` into both paths rather than collapsing one; #805 explicitly leaves the legacy `python3 -c` calls alone.
3. Forked test surface — `tests/test_hooks_shell.py`, `tests/test_claude_plugin_hook_wrappers.py`, `tests/test_hooks_cli.py` each cover a different slice.

**Proposal:**

Reshape `hooks/mempal_*_hook.sh` to match the plugin wrappers — a 3-tier `run_mempalace_hook()` function that calls `mempalace hook run --hook {stop|precompact} --harness claude-code`. All real logic (JSON parsing, exchange counting, 15-message block decision, background mine) already lives in `mempalace/hooks_cli.py` for the plugin path; any gaps get ported once.

**Result:** inline `python3 -c` disappears from bash (killing the three failure modes discussed on #833 for any future mempalace-importing addition), `MEMPAL_PYTHON` becomes obsolete, and `MEMPAL_VERBOSE` / `hooks.auto_save` / future toggles live in one place. Pure refactor — same exchange counts, same block text, same `~/.mempalace/hook_state/hook.log`.

**Related:** #340, #513, #670, #805, #833, #942.

Issue #1068, opened 2026-04-21
Background `mempalace mine` pins 400–500 % CPU — ORT intra_op pool ignores OMP env vars


sha2fiddy:
**What happened?**

On a multi-core host, every `mempalace mine` run pins 4–5 cores (peak sustained **372–463 %** in `ps`). Stacked Stop-hook fires from concurrent Claude Code sessions make the machine unusable.

The cost isn't in mempalace's Python — `miner.py` and `convo_miner.py` are sequential `for` loops (no multiprocessing, no torch, no manual `hnswlib` calls). The CPU sits inside a single `collection.upsert(...)`.

ChromaDB's default `ONNXMiniLM_L6_V2` builds its ONNX Runtime `InferenceSession` with no `SessionOptions`, so ORT's intra_op pool defaults to ≈physical-core-count workers. **`OMP_NUM_THREADS` / `MKL_NUM_THREADS` / `OPENBLAS_NUM_THREADS` are inert against this pool** — ORT owns its own. Same class of silent no-op as `ORT_DISABLE_COREML` (#397 / #653).

**What did you expect?**

A user-settable cap on the threads a background mine is allowed to use, with a safe default so a fresh install on a 10-core host doesn't turn every Stop-hook fire into a thermal event.

**How to reproduce:**

1. mempalace 3.3.0 (any 3.x with chromadb 1.5.x default embedder) on a multi-core host.
2. Mine ≥500 JSONL transcripts (~1360 drawers):
   ```bash
   nice -n 10 ~/.local/bin/mempalace mine <project-dir>
   ```
3. Sample CPU + threads:
   ```bash
   ps -o pid,%cpu,command -p "$(pgrep -f 'mempalace mine')"
   ps -M "$(pgrep -f 'mempalace mine')"   # count onnxruntime::ThreadPo workers
   ```
4. Setting `OMP_NUM_THREADS=2 MKL_NUM_THREADS=2 OPENBLAS_NUM_THREADS=2` before step 2 changes nothing.

**Benchmarks (M-series Mac, 10 cores, 500 files / 1360 drawers):**

| Run                                                                 | Peak %CPU           | Hot ORT workers | Wall |
| ------------------------------------------------------------------- | ------------------- | --------------- | ---- |
| Uncapped (stock)                                                    | **372–463**        | 8               | 54 s |
| Isolated `upsert()` + `SessionOptions(intra_op=2, inter_op=1)`      | **62.9**           | 1               | 47.5 s |
[...]

Issue #1083, opened 2026-04-21
Stop + PreCompact hooks auto-run `mempalace mine` on chat transcript parent with default flags → polluted mega-wing, no opt-out


raphaelsamy:
## Summary

The MemPalace Claude Code plugin's Stop hook (and, identically, its PreCompact hook) auto-runs `python3 -m mempalace mine <transcript_path.parent>` with **default flags** every ~15 human messages. Because Claude Code stores transcripts under `~/.claude/projects/<url-encoded-cwd>/<session>.jsonl`, the hook ends up mining the **chat transcript folder** (not the user's project dir) with:

- `--mode projects` (default — **wrong** for JSONL conversation files)
- wing derived from the folder's basename (a URL-encoded path like `-Users-<user>-<encoded>-<cwd>`)
- single `general` room

The result is a polluted mega-wing containing hundreds-to-thousands of drawers of raw JSONL text chunked as if it were project code. There is no working opt-out — setting `MEMPAL_DIR=""` falls through to the transcript path; there is no other env var, plugin setting, or config key that gates `_maybe_auto_ingest`. Worse, this auto-mine is entangled with the (useful) save-reminder JSON in the same `hook_stop` function, so users who want the save-reminder have no way to keep it while disabling the broken auto-mine.

## Environment

- **mempalace**: 3.3.2 (from PyPI)
- **Install command**: `uv pip install --python $(pyenv which python3) mempalace`
- **Python**: 3.12.11 via pyenv
- **OS**: macOS, Darwin 25.3 (macOS 15, arm64)
- **Claude Code plugin**: `MemPalace/mempalace` marketplace branch `develop`, plugin version 3.3.2
- **Plugin-shipped hooks** (at `.claude-plugin/hooks/`):
  - `mempal-stop-hook.sh` → `python3 -m mempalace hook run --hook stop --harness claude-code`
  - `mempal-precompact-hook.sh` → `python3 -m mempalace hook run --hook precompact --harness claude-code`
  - Registered for `Stop` and `PreCompact` events in `.claude-plugin/hooks/hooks.json`

[...]

PR #1110, opened 2026-04-22
feat: split hooks.auto_save and hooks.auto_mine for independent control of diary vs mine

sha2fiddy:
## Problem

The Stop and PreCompact hooks currently do three things with no way to enable them independently:

1. `_save_diary_direct` is cheap, a few hundred ms. Writes the distilled memory entry.
2. `_ingest_transcript` is medium cost. Syncs the current session's `.jsonl`.
3. `_maybe_auto_ingest` / `_mine_sync` is expensive. Walks the whole `~/.claude/projects/<cwd>/` tree and can peg a CPU core for hours on large archives.

So users are stuck picking between "no hook-driven memory" and "burn a core every 15 exchanges." Some people want the fast diary write but not the recursive mine, especially on machines with thousands of accumulated transcripts. #1068 and #1069 have context.

## Solution

Adds a new config flag `hooks.auto_mine` (default `true`) that gates the two mine paths in both hooks. The diary save stays unconditional because it's cheap and it's the thing users actually want.

With `auto_mine: false`, hook-driven diary writes keep working but the recursive mine is skipped. "Combined mode" is just leaving both `silent_save` and `auto_mine` on.

### Flags

| Flag | Default | Gates |
|---|---|---|
| `hooks.silent_save` | `true` | Silent Python diary write vs legacy `{"decision": "block"}` AI-driven save |
| **`hooks.auto_mine` *(new)*** | `true` | Single-file ingest + recursive mine after the diary save |

Override precedence follows the existing pattern in `config.py`: env var (`MEMPALACE_HOOKS_AUTO_MINE`, legacy alias `MEMPAL_HOOKS_AUTO_MINE`) > `~/.mempalace/config.json` > default `True`.

### Usage

Via the existing MCP tool:

```
mempalace_hook_settings(auto_mine=False)
```

Via config file:

```json
{
  "hooks": {
    "auto_mine": false
  }
}
```

Via env var:

```bash
export MEMPALACE_HOOKS_AUTO_MINE=false
```

Falsey values: `0`, `false`, `no`, `off` (case-insensitive). Anything else is truthy. Empty-string env var falls through to the config file, matching the pattern used by `MEMPALACE_PALACE_PATH` and `MEMPALACE_ENTITY_LANGUAGES`.

## Scope note

This PR doesn't touch `hooks.silent_save` or any other existing flag. It's independent of #711 (`hooks.auto_save`, the outer hook enable/disable). Either can land first; the other rebases cleanly. I kept them as separate PRs.

## Test plan

- [x] `tests/test_config.py`: 5 new tests plus 7 parametrize cases covering default, file override, env override, and the legacy env prefix.
- [x] `tests/test_hooks_cli.py`: 3 new tests for the stop-hook silent branch (auto_mine on and off) and precompact with auto_mine off.
- [x] Full non-benchmark suite 1080/1080 locally on Python 3.14.
- [x] `ruff check .` passes.
- [x] `ruff format --check` passes on touched files under both current ruff and CI-pinned `ruff>=0.4.0,<0.5`.
- [x] CI: 6/6 green (lint, test-linux 3.9/3.11/3.13, test-macos, test-windows).

## Backwards compatibility

`hook_auto_mine` defaults to `True`. Users who don't set the flag see no behavior change.

## Files touched

- `mempalace/config.py`: new `hook_auto_mine` property
- `mempalace/hooks_cli.py`: gate mine calls in `hook_stop` (both branches) and `hook_precompact`
- `mempalace/mcp_server.py`: `auto_mine` parameter on `mempalace_hook_settings`
- `tests/test_config.py`: config coverage
- `tests/test_hooks_cli.py`: hook coverage (helper `_capture_hook_output` gains an `auto_mine=True` default)
- `website/reference/mcp-tools.md`: doc update
- `CHANGELOG.md`: `[Unreleased]` entry

## Related

- Informed by #1068 (ORT CPU pin diagnosis) and #1069 (hooks consolidation refactor)
- Independent of #711 (`hooks.auto_save`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -322,6 +322,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ### Improvements
 - Optimize entity detection with regex caching and pre-compilation (#828)
 - Extract locked filing block into helper to keep `mine_convos` under C901 complexity
+- `hooks.auto_mine` config flag decouples the background mine from the diary save in stop/precompact hooks — set `false` (via `mempalace_hook_settings` MCP tool, `config.json`, or `MEMPALACE_HOOKS_AUTO_MINE` env var) to keep fast diary writes without the recursive `~/.claude/projects/<cwd>/` walk that can peg CPU on large archives; defaults to `true` to preserve existing behavior
 
 ### Documentation
 - Add `docs/CLOSETS.md` — closet layer overview
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -602,14 +602,18 @@ class MempalaceConfig:
 
         Env var ``MEMPALACE_HOOKS_AUTO_MINE`` (or legacy ``MEMPAL_HOOKS_AUTO_MINE``)
         overrides the config file. Accepts ``0``/``false``/``no``/``off`` as
-        falsey; anything else is truthy. Default: ``True`` (preserves existing
+        falsey; anything else (including ``true``/``1``/``yes``/``on``) is
+        truthy. Empty / whitespace-only values are treated as unset and fall
+        back to the config file. Default: ``True`` (preserves existing
         behavior for users who don't opt out).
         """
-        env_val = os.environ.get("MEMPALACE_HOOKS_AUTO_MINE") or os.environ.get(
-            "MEMPAL_HOOKS_AUTO_MINE"
-        )
-        if env_val:
-            return env_val.strip().lower() not in ("0", "false", "no", "off")
+        env_val = os.environ.get("MEMPALACE_HOOKS_AUTO_MINE")
+        if env_val is None:
+            env_val = os.environ.get("MEMPAL_HOOKS_AUTO_MINE")
+        if env_val is not None:
+            stripped = env_val.strip().lower()
+            if stripped:
+                return stripped not in ("0", "false", "no", "off")
         return self._file_config.get("hooks", {}).get("auto_mine", True)
 
     @property
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -592,6 +592,27 @@ class MempalaceConfig:
         return self._file_config.get("hooks", {}).get("silent_save", True)
 
     @property
+    def hook_auto_mine(self):
+        """Whether the stop/precompact hooks run the background mine after the diary save.
+
+        Set to ``False`` to keep the fast diary write but skip the
+        ``_ingest_transcript`` + ``_maybe_auto_ingest`` calls that walk the
+        ``~/.claude/projects/<cwd>/`` tree and can peg a CPU core for hours on
+        large archives. Diary writes (``_save_diary_direct``) still fire.
+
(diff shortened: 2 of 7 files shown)


jphein (commented):
Pulled and verified on Python 3.12:

- **1080/1080** non-benchmark tests pass (matches your claim).
- **Gate placement checks out**: `_save_diary_direct` stays unconditional; `_ingest_transcript` + `_maybe_auto_ingest` both guarded in silent AND legacy-block branches; precompact early-returns cleanly when `auto_mine=false`. Exception fallback defaults to `True`, so a broken config never accidentally disables mining — same defensive posture as `silent_save`.
- **Env-var override** (`MEMPALACE_HOOKS_AUTO_MINE` + legacy alias) works as documented; falsey parsing matches the `MEMPALACE_PALACE_PATH` precedent.

Design aligns with what I proposed in #1083 — thanks for picking this up. `_mine_sync`'s recursive walk has been our production's biggest CPU headache on a 165K-drawer palace; first-class opt-out is much cleaner than the wrapper workarounds I was queueing in a fork-ahead PR. I'll drop that row from our tracker in favor of yours.

Nit, not a blocker: `hook_precompact` re-instantiates `MempalaceConfig()` on its own rather than reusing the instance loaded elsewhere in the module. Fires once per session so probably not worth a re-roll — just flagging.

Qodo-Free-For-OSS:
Hi, `MempalaceConfig.hook_auto_mine` only applies the env override when the env var is truthy, so `MEMPALACE_HOOKS_AUTO_MINE=""` falls back to config/default instead of overriding. Whitespace values (e.g. " ") are treated as truthy and end up enabling mining, which is inconsistent with the intent of using env vars to disable background mining.

**How to fix:** Handle empty env values

> ### Issue description
> `hook_auto_mine` treats `MEMPALACE_HOOKS_AUTO_MINE=""` as “not set” and treats whitespace-only as truthy (enables mining). This makes env-based disabling unreliable in environments that inject empty values.
> 
> ### Issue Context
> The code currently checks `if env_val:` and then interprets any non-listed value as truthy.
> 
> ### Fix Focus Areas
> - mempalace/config.py[298-316]
> 
> ### Suggested change
> - Change the guard to distinguish between unset vs set-to-empty, e.g. `if env_val is not None:`.
> - Decide semantics for empty/whitespace explicitly (common options: treat as `False`, or treat as “unset” consistently after `strip()`).
> - Add unit tests for `""` and `"   "` env values to lock the behavior.

---

sha2fiddy:
Fixed in ec33d18. Changed the guard from `if env_val:` to `if env_val is not None:` and only treat the env var as authoritative when the stripped value is non-empty — so `""` and whitespace-only fall back to config (matching the unset case), and `0`/`false`/`no`/`off` correctly disable mining. Also separated the primary/legacy env var lookup so an empty primary doesn't silently fall through to the legacy var.

Added 5 parametrized tests in `tests/test_config.py` covering `""`, `"   "`, `"\t"`, `"\n"`, plus the empty-primary-doesn't-leak-to-legacy case. 4 of them fail against the prior code; all pass after the fix.

igorls:

Issue #1093, opened 2026-04-22
v3.3.2 release defect: `mempalace-mcp` entry point missing from `pyproject.toml` despite `plugin.json` requiring it — MCP server fails to start on fresh install


jphein:
## Summary

The v3.3.2 release artifact on PyPI ships a `.claude-plugin/plugin.json` that declares `"command": "mempalace-mcp"`, but the same artifact's `pyproject.toml` only declares the `mempalace` console script — not `mempalace-mcp`. Net effect: a fresh `pip install mempalace` yields a Claude Code plugin config that points to a binary the package didn't install. MCP server fails to launch with "command not found."

First surfaced by @raphaelsamy in [#1049#issuecomment-4292661...](https://github.com/MemPalace/mempalace/issues/1049) — filing as a dedicated issue so it has a tracking home.

## Reproducer

```bash
python3 -m venv /tmp/mpv && source /tmp/mpv/bin/activate
pip install "mempalace==3.3.2"

# The binary that plugin.json calls doesn't exist:
which mempalace        # /tmp/mpv/bin/mempalace   ← present
which mempalace-mcp    # (nothing)                 ← missing

# The plugin config tries to invoke it anyway:
cat $(python3 -c "import mempalace, os, pathlib; print(pathlib.Path(mempalace.__file__).parent.parent / '.claude-plugin' / 'plugin.json')") | grep command
#   "command": "mempalace-mcp"
```

Then opening Claude Code with the MemPalace plugin installed: MCP server startup fails silently or with `exec: "mempalace-mcp": executable file not found in $PATH`.

## Root cause

Two changes were supposed to ship together but landed on different branches relative to the v3.3.2 cut:

| Change | Where it lives on develop | Present in v3.3.2 tag? |
|---|---|---|
| `plugin.json` → `"command": "mempalace-mcp"` | `.claude-plugin/plugin.json` | ✅ |
| `pyproject.toml` → add `mempalace-mcp = "mempalace.mcp_server:main"` entry point (from [#340](https://github.com/MemPalace/mempalace/pull/340) / [#805](https://github.com/MemPalace/mempalace/pull/805)) | `pyproject.toml` | ❌ |

The `release/3.3.2` branch appears to have been cut before #340/#805 made it into the release cherry-pick set, but *after* the plugin.json update landed. The two changes are independently correct; together on v3.3.2 they aren't compatible.

## Who's affected

[...]

Issue #1049, opened 2026-04-20
Hooks and MCP server break in projects with Python venv (hardcoded `python3`)


sergesha:
## Problem

All hook scripts in `.claude-plugin/hooks/` and the MCP server definition in `plugin.json` call bare `python3`:

**Hooks (`mempal-stop-hook.sh`, `mempal-precompact-hook.sh`):**
```bash
echo "$INPUT" | python3 -m mempalace hook run --hook stop --harness claude-code
```

**MCP server (`plugin.json`):**
```json
"mcpServers": {
  "mempalace": {
    "command": "python3",
    "args": ["-m", "mempalace.mcp_server"]
  }
}
```

In projects with an active Python `.venv`, `python3` resolves to `.venv/bin/python3`, which doesn't have `mempalace` installed. This causes the stop hook to fail on every session exit:

```
.venv/bin/python3: No module named mempalace
```

The MCP server also fails to start for the same reason.

## Environment

- mempalace 3.3.1 (pip) / 3.3.0 (plugin)
- Python installed via mise, mempalace installed in mise python
- Project has its own `.venv` which shadows `python3` in PATH

## Proposed fix

Support a `MEMPALACE_PYTHON` environment variable with fallback to `python3`:

**Hooks:**
```bash
MEMPALACE_PYTHON="${MEMPALACE_PYTHON:-python3}"
echo "$INPUT" | "$MEMPALACE_PYTHON" -m mempalace hook run --hook stop --harness claude-code
```

**plugin.json:** Harder to fix since JSON is static. Options:
1. A wrapper script (`mempalace-run`) that reads `MEMPALACE_PYTHON`
2. Document that users with venvs need to set `MEMPALACE_PYTHON` in Claude Code `settings.json` → `env`

This lets users set `MEMPALACE_PYTHON` in Claude Code's `settings.json` → `env` to point to the python where mempalace is installed, without affecting other hooks or tools.

mvalentsev:
This is already tracked in a cluster of issues: #378 is the broader canonical tracker, and #408 / #650 / #753 cover the same pattern from different angles. A few PRs are open against exactly these files:

- `.claude-plugin/hooks/*.sh`: #942 (fatkobra) uses an installed `mempalace` CLI first and falls back to `python3` / `python`
[...]

PR #1115, opened 2026-04-22
fix(plugin): restore venv-aware Python resolution for hooks + MCP

jphein:
## Summary

Three recent `.claude-plugin/` changes regressed hook and MCP invocation for installs that depend on a local venv:

- `5fe0c1c` — `mempal-stop-hook.sh` → PATH-only (`command -v mempalace` / bare `python3 -m mempalace`)
- `be9214a` — `mempal-precompact-hook.sh` → same regression
- `9f5b8f5` — `.mcp.json` → bare `"mempalace-mcp"` command

All three assume a system-wide install via pipx/uv/pip-global — fine for that path, but they break:

1. **Editable dev installs** (`pip install -e .` inside a repo venv). `mempalace` / `mempalace-mcp` and the module only live in the repo venv, not on PATH. The stop/precompact hooks fall through all three checks and print:

   ```
   MemPalace hook error: could not find a runnable mempalace command or module
   ```

   MCP fails to spawn because bare `mempalace-mcp` isn't on PATH.

2. **Claude Code plugin installs that create a per-plugin venv** at `${CLAUDE_PLUGIN_ROOT}/venv/`. Neither the hooks nor `.mcp.json` look there, so anything that isn't also system-installed breaks.

## Fix

Restore the three-tier resolution:

1. `MEMPALACE_PYTHON` env var (user override)
2. `$PLUGIN_ROOT/venv/bin/python3` (covers dev installs AND Claude Code's per-plugin venv)
3. System `python3` (pip --user / pipx / uv as system install — preserves the original intent of the regressing commits)

`.mcp.json` goes back to `${CLAUDE_PLUGIN_ROOT}/venv/bin/python -m mempalace.mcp_server`. `${CLAUDE_PLUGIN_ROOT}` is expanded by Claude Code before spawn, so this works for dev installs (via venv symlink) and packaged plugin installs alike.

## Test plan

- [x] Editable dev install (`pip install -e .` in repo venv): both hooks return `{}` from a neutral cwd; MCP `initialize` handshake returns `serverInfo = {name: mempalace, version: 3.3.1}`; `tools/list` returns all 29 tools.
- [x] Fallback to system `python3`: confirmed by inspection — if neither `MEMPALACE_PYTHON` nor `$PLUGIN_ROOT/venv/bin/python3` resolves, the hook executes the same `python3 -m mempalace hook run ...` invocation the regressing commits used.
- [ ] `${CLAUDE_PLUGIN_ROOT}/venv/` packaged install path — works by construction (Claude Code's plugin manager creates the venv at that exact location), but not re-tested in this PR.

### DIFF
--- a/.claude-plugin/.mcp.json
+++ b/.claude-plugin/.mcp.json
@@ -1,5 +1,9 @@
 {
   "mempalace": {
-    "command": "mempalace-mcp"
+    "command": "${CLAUDE_PLUGIN_ROOT}/venv/bin/python",
+    "args": [
+      "-m",
+      "mempalace.mcp_server"
+    ]
   }
 }
--- a/.claude-plugin/hooks/mempal-precompact-hook.sh
+++ b/.claude-plugin/hooks/mempal-precompact-hook.sh
@@ -1,24 +1,20 @@
 #!/bin/bash
 # MemPalace PreCompact Hook — thin wrapper calling Python CLI
 # All logic lives in mempalace.hooks_cli for cross-harness extensibility
-run_mempalace_hook() {
-  if command -v mempalace >/dev/null 2>&1; then
-    mempalace hook run "$@"
-    return $?
-  fi
+#
+# Python resolution order:
+#   1. MEMPALACE_PYTHON env var (user override)
+#   2. Plugin root's venv (development installs; Claude Code also creates one per plugin)
+#   3. System python3 (pip install --user / pipx)
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PLUGIN_ROOT="$(dirname "$SCRIPT_DIR")"
 
-  if command -v python3 >/dev/null 2>&1 && python3 -c "import mempalace" >/dev/null 2>&1; then
-    python3 -m mempalace hook run "$@"
-    return $?
-  fi
-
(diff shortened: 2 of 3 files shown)


jphein:
Closing this as premature on my end — I had explicitly deferred filing in my [#1049 comment on 2026-04-21](https://github.com/MemPalace/mempalace/issues/1049#issuecomment-4292554161) pending @bensig's arbitration between this autodetect pattern and @sha2fiddy's broader consolidation in #1069. I jumped the gun.

The failing tests in `tests/test_claude_plugin_hook_wrappers.py` are correctly enforcing the contract #942 established — PATH-only resolution, explicit *"could not find a runnable mempalace command or module"* error, `python` fallback after `python3`. My change reintroduced `dirname`/`cat` and dropped those guarantees, which the tests caught in the first run. Working as designed.

Separating the two motivations cleanly:

- **The `.mcp.json` piece** (bare `mempalace-mcp` breaks fresh v3.3.2 installs) gets fixed by the v3.3.3 release @bensig mentioned on #976 earlier today, via #1093 → `pyproject.toml` entry point. No longer needs this PR.
- **The hook-script piece** (venv detection for editable dev installs) properly belongs to whatever shape #1069 settles on. Re-opening that ahead of arbitration would just repeat this mistake.

Thanks Copilot for the clear inline review — the test-as-spec read was exactly right.

Issue #1098, opened 2026-04-22
status command crashes with "too many SQL variables" on large palaces (129K drawers)


bloodcarter:
## Version
- mempalace 3.3.2
- chromadb 1.5.8
- Python 3.13
- macOS (Apple Silicon)

## Summary
`mempalace status` crashes with `chromadb.errors.InternalError: too many SQL variables` on a palace with 129,382 drawers. Looks like a regression of the fix from #66 ("Batch ChromaDB reads to avoid SQLite variable limit") that landed in v3.1.0.

## Stack trace
```
File "/Users/vlad/.mempalace-venv/lib/python3.13/site-packages/mempalace/miner.py", line 852, in status
    r = col.get(limit=total, include=["metadatas"]) if total else {"metadatas": []}
File "/Users/vlad/.mempalace-venv/lib/python3.13/site-packages/mempalace/backends/chroma.py", line 341, in get
    raw = self._collection.get(**kwargs)
File "/Users/vlad/.mempalace-venv/lib/python3.13/site-packages/chromadb/api/models/Collection.py", line 161, in get
File "/Users/vlad/.mempalace-venv/lib/python3.13/site-packages/chromadb/api/rust.py", line 440, in _get
chromadb.errors.InternalError: Error executing plan: Internal error: error returned from database: (code: 1) too many SQL variables
```

## Root cause
`miner.py:852` passes `limit=total` to `col.get(...)` without batching. When `total` exceeds SQLite's bind-parameter ceiling (default 999, sometimes 32766), the single query fails.

## Repro
1. Mine enough content to exceed ~10-30K drawers
2. `mempalace status`

## Expected behavior
Status should paginate the metadata read using `offset`/`limit` batching, similar to the approach that #66 applied elsewhere. A batch size of 5000–10000 works reliably.

## Workaround
```python
import chromadb
from collections import defaultdict
client = chromadb.PersistentClient(path="~/.mempalace/palace")
col = client.get_collection("mempalace_drawers")
total = col.count()
wing_room = defaultdict(lambda: defaultdict(int))
offset = 0
while offset < total:
    r = col.get(include=["metadatas"], limit=5000, offset=offset)
    for m in r["metadatas"]:
        wing_room[m.get("wing","?")][m.get("room","?")] += 1
    offset += 5000
```

## Same pattern likely affects
[...]

PR #1124, opened 2026-04-23
fix: harden Chroma status path and Windows CLI output

### LINKED ISSUE #1121: Chroma client can crash opening palaces with stale HNSW segment directories
## Summary
Opening an existing palace can fail hard when `chroma.sqlite3` is much newer than the on-disk HNSW segment directories. In my local repro on Windows, the process previously aborted while opening the real palace; after quarantining stale segment dirs before `PersistentClient(...)`, the same palace opens normally.

## What I observed
- Palace path: existing real user palace with ~45k drawers
- `chroma.sqlite3` mtime was about 43-45 hours newer than both segment directories
- Before the fix, commands that open the collection (`status` was the easiest repro) could abort during Chroma open / early collection access
- After renaming stale HNSW dirs out of the way, Chroma rebuilt lazily and the palace opened again

## Likely cause
`mempalace.backends.chroma` already has `quarantine_stale_hnsw(...)`, but `ChromaBackend._client()` does not call it before constructing a fresh `chromadb.PersistentClient`. That leaves the crash-prone stale-index path reachable on real palaces.

## Proposed fix
Call `quarantine_stale_hnsw(palace_path)` immediately before `chromadb.PersistentClient(path=palace_path)` when `_client()` is creating or refreshing the cached client.

## Validation
I tested the fix locally against the real palace and it changed the behavior from a hard failure to a normal recoverable runtime error later in `status`, which exposed a second independent bug (`too many SQL variables`) that can then be fixed separately.

I also added a regression test that verifies stale segment directories are quarantined before the backend opens the client.

## Local patch reference
Branch: `codex/mempalace-fix-windows-help-and-status`
Commit: `dda82d1`

### LINKED ISSUE #1122: Windows default console encoding can crash CLI help output
## Summary
On Windows with the default `cp1252` console encoding, CLI help output can crash with `UnicodeEncodeError` because user-facing help strings include non-ASCII punctuation.

## Repro
On Windows PowerShell, running:

```bash
python -m mempalace.cli --help
```

raised:

```text
UnicodeEncodeError: 'charmap' codec can't encode character '\u2192' ...
```

The most direct trigger in my repro was the Unicode arrow in the `migrate` subcommand help text (`3.0.0 ? 3.1.0`). There were also several em dashes in help / status strings.

## Impact
This blocks normal CLI discovery on a standard Windows console unless the user explicitly forces UTF-8 output, e.g. by setting `PYTHONIOENCODING=utf-8`.

## Proposed fix
Use ASCII-safe user-facing help/status strings in the CLI path, e.g.:
- replace `?` with `->`
- replace em dashes in printed/help text with `-`

## Validation
After normalizing the CLI/help strings locally, all of these worked without forcing UTF-8:
- `python -m mempalace.cli --help`
- `python -m mempalace.cli init --help`
- `python -m mempalace.cli status`

## Local patch reference
Branch: `codex/mempalace-fix-windows-help-and-status`
Commit: `dda82d1`

### LINKED ISSUE #1123: mempalace status fetches large palaces in one query and can hit SQLite variable limits
## Summary
`mempalace status` currently does `col.get(limit=total, include=["metadatas"])`. On larger palaces this can exceed SQLite/Chroma parameter limits and fail even though the data itself is healthy.

## Repro
On a real palace with about 45k drawers, after getting past a separate stale-HNSW issue, this command failed:

```bash
python -m mempalace.cli status
```

with:

```text
chromadb.errors.InternalError: Error executing plan: Internal error: error returned from database: (code: 1) too many SQL variables
```

## Why this happens
`status()` asks Chroma for the entire palace in one `get()` call. Other parts of the codebase already paginate reads specifically to stay below Chroma/SQLite limits.

## Proposed fix
Page through metadata in batches, for example:
- `limit=500`
- `offset += len(batch_metas)`
- stop when a batch is empty or shorter than the batch size

That fixes the failure on the large palace while preserving the same output.

## Validation
After switching `status()` to paginated reads locally, `python -m mempalace.cli status` succeeded against the same real palace and reported `45,436` drawers.

I also added a regression test that exercises a fake 1200-drawer collection and verifies `status()` paginates instead of doing a single giant read.

## Local patch reference
Branch: `codex/mempalace-fix-windows-help-and-status`
Commit: `dda82d1`

vanachterjacob:
## Summary
This PR fixes three user-facing failures I hit while verifying a local MemPalace install on Windows:

- quarantine stale Chroma HNSW segment directories before opening a refreshed client
- paginate mempalace status metadata reads so large palaces do not hit SQLite variable limits
- make CLI help/status text ASCII-safe so default Windows cp1252 consoles do not crash on --help

## Validation
- python -m pytest tests/test_backends.py tests/test_cli.py tests/test_miner.py -q
- python -m mempalace.cli --help
- python -m mempalace.cli init --help
- python -m mempalace.cli status

Fixes #1121
Fixes #1122
Fixes #1123

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -466,6 +466,7 @@ class ChromaBackend(BaseBackend):
 
         if cached is None or inode_changed or mtime_changed or mtime_appeared:
             _fix_blob_seq_ids(palace_path)
+            quarantine_stale_hnsw(palace_path)
             cached = chromadb.PersistentClient(path=palace_path)
             self._clients[palace_path] = cached
             # Re-stat after the client constructor runs: chromadb creates
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
-MemPalace — Give your AI a memory. No API key required.
+MemPalace - Give your AI a memory. No API key required.
 
 Two ways to ingest:
   Projects:      mempalace mine ~/projects/my_app          (code, docs, notes)
@@ -103,7 +103,7 @@ def cmd_init(args):
                     json.dump(confirmed, f, indent=2)
                 print(f"  Entities saved: {entities_path}")
         else:
-            print("  No entities detected — proceeding with directory-based rooms.")
+            print("  No entities detected - proceeding with directory-based rooms.")
 
     # Pass 2: detect rooms from folder structure
     detect_rooms_local(project_dir=args.dir, yes=getattr(args, "yes", False))
@@ -151,7 +151,7 @@ def cmd_sweep(args):
 
     The sweeper deduplicates against its own prior writes via
     deterministic drawer IDs + a timestamp cursor. It does NOT currently
-    coordinate with the file-level miners (miner.py / convo_miner.py) —
+    coordinate with the file-level miners (miner.py / convo_miner.py) -
     those produce char-chunked drawers without compatible message
     metadata, so running both miners may store overlapping content under
     different IDs.
@@ -205,7 +205,7 @@ def cmd_search(args):
 
 
 def cmd_wakeup(args):
-    """Show L0 (identity) + L1 (essential story) — the wake-up context."""
+    """Show L0 (identity) + L1 (essential story) - the wake-up context."""
     from .layers import MemoryStack
 
     palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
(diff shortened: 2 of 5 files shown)


igorls:
Hi @aaron-trinity — this PR is currently DRAFT, +3180/-3104, CONFLICTING, and bundles three unrelated fixes (HNSW segment quarantine, status pagination, Windows CLI ASCII).

The status pagination part is **already shipped on develop** via `mcp_server.py:_fetch_all_metadata` and `miner.py:status` pagination (issues #1098/#1073/#1123 closed as fixed). The Windows ASCII piece is now being handled by #1282.

The remaining valuable piece is the HNSW segment quarantine for `_client()` — that's the only path not covered by the merged #1173 (which only wired `make_client()`). It maps to issue #1121 (now in v3.3.5).

Could you split the quarantine fix into its own focused PR against current `develop`? Or, if you'd rather hand it off, comment here and we'll write the minimal patch ourselves and credit you. The remaining diff for that piece should be ~5-10 lines (extend `_client()` to call `quarantine_stale_hnsw` like `make_client()` already does).

PR #1135, opened 2026-04-23
fix: narrow `_fix_blob_seq_ids` + add `repair --mode max-seq-id`

### LINKED ISSUE #1134: `_fix_blob_seq_ids` misreads chromadb 1.5.x `max_seq_id` BLOBs, suppressing new writes
## What's broken

`mempalace/backends/chroma.py::_fix_blob_seq_ids` (added in #664) runs `int.from_bytes(blob, 'big')` on every BLOB row in `max_seq_id`. But chromadb 1.5.x writes its own BLOB format there — `b'\x11\x11'` followed by six ASCII digits — and the shim reads that as a big-endian u64. The result is a ~1.23×10^18 integer sitting in the column.

chromadb's `embeddings_queue` filter is `seq_id > start`, and `start` is read from `max_seq_id.seq_id` at segment open. New rows' seq_ids are small (<1e10 even on large palaces), so nothing ever crosses the poisoned threshold. Writes queue and never flush. Old drawers still read fine, which is why the bug is silent until you try to fetch something written after the poisoning.

## Repro

Any palace with BLOB-typed `max_seq_id.seq_id` rows the first time it opens under chromadb 1.5.x with the #664 shim active. Mine got hit on 2026-04-20:

| segment | scope | clean | poisoned |
|---|---|---|---|
| `b8e762ef-92b8-4b67-8bd0-ae8ba4622826` | closets VECTOR | 481194 | 1229821589196978484 |
| `c9bf2ac4-2557-4540-b5f8-76f0f75c71be` | closets METADATA | 501418 | 1229822654349062456 |
| `5ef0f58c-dea2-4247-9a0a-fa3c256f0a4c` | drawers VECTOR | 502498 | 1229822654365841720 |
| `b7b6cc41-a631-4d0c-803e-5256793be770` | drawers METADATA | 502607 | 1229822654365970487 |

Clean values came from a `chroma.sqlite3.pre-sysdb10fix-2026-04-20` backup. 465k+ existing drawers still read; every write after 2026-04-20 silently disappeared.

## Fix

Shim: drop `max_seq_id` from the loop — chromadb owns that format. The `embeddings` branch stays for real 0.6.x palaces but skips any BLOB starting with `b'\x11\x11'` as belt-and-braces.

New subcommand `mempalace repair --mode max-seq-id` restores poisoned rows (detects `seq_id > 2^53`). Two clean-value sources: `--from-sidecar <path>` copies exact values out of a pre-corruption `chroma.sqlite3` backup, or the default heuristic uses `MAX(embeddings.seq_id)` over the owning collection. Heuristic matches METADATA max exactly; VECTOR segments land a few seq_ids ahead, which the queue treats as already consumed — skipping a tiny window of already-indexed embeddings.

PR to follow.

## Possibly related

- #664 — introduced the shim. This is a regression in its loop.
- #1006 — closed by pinning chromadb, attributed to "segment materialization never runs." Same symptom on 1.5.x; could be this.
- #1099 — silent no-op deletes after 0.6.3 → 1.5.8. Same migration boundary.
- #538, #813 — silent write loss reports. Overlap not verified; linking for triage.

## Out of scope

- chromadb's choice of `max_seq_id` BLOB format.
- A `status --health` command that would have caught this earlier.

sha2fiddy:
Closes #1134.

## What

Shim (`mempalace/backends/chroma.py`):
- Drop `max_seq_id` from `_fix_blob_seq_ids`. chromadb 1.5.x writes its own BLOB format there (`b'\x11\x11'` + 6 ASCII digits). Misreading it as a legacy 0.6.x big-endian u64 produces a ~1.23e18 integer that stalls every subsequent write.
- `embeddings` branch stays for real 0.6.x palaces but skips any BLOB starting with `b'\x11\x11'`.

Recovery (`mempalace/repair.py`, `mempalace/cli.py`):
- `mempalace repair --mode max-seq-id [--segment <uuid>] [--from-sidecar <path>] [--dry-run] [--yes] [--no-backup]`
- Detection: `seq_id > 2^53`.
- Default heuristic: `MAX(embeddings.seq_id)` over the owning collection. Exact for the METADATA segment; for the sibling VECTOR segment it can land a few seq_ids ahead of the true max, which the queue treats as already consumed. Better than resetting to 0.
- `--from-sidecar` copies clean values from a pre-corruption `chroma.sqlite3` backup.
- Backs up the SQLite file, closes chroma handles, runs atomic UPDATEs, and verifies after. `MaxSeqIdVerificationError` fires if any row is still above the threshold.

## Test plan

- [x] 8 new tests in `tests/test_repair.py` (detect, heuristic, sidecar, dry-run, segment filter, no-op, backup, rollback-on-verify-failure)
- [x] 3 new tests in `tests/test_backends.py` (`max_seq_id` untouched, sysdb-10 prefix skipped in `embeddings`, legacy big-endian u64 still converts)
- [x] `pytest tests/ --ignore=tests/benchmarks` -> 1077 passed
- [x] `ruff check .` clean (pinned to `>=0.4.0,<0.5`)
- [x] `ruff format --check` clean on touched files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -177,6 +177,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Hall detection — routes drawer content to `emotions` / `technical` / `family` / `memory` / `identity` / `consciousness` / `creative` halls, enabling hall-based graph connectivity within wings (#835)
 
 ### Bug Fixes
+- Repair `max_seq_id` corruption caused by `_fix_blob_seq_ids` misinterpreting chromadb 1.5.x's sysdb-10 BLOB format (`b'\x11\x11'` + ASCII digits) as legacy 0.6.x big-endian BLOBs. The shim now skips the `max_seq_id` table entirely and guards the `embeddings` branch with a prefix check. New subcommand `mempalace repair --mode max-seq-id [--from-sidecar <path>]` restores affected palaces. Fixes silent drawer-write drops that began after chromadb 1.5.x upgrades on palaces that still had BLOB-typed `max_seq_id` rows at migration time.
 - Set `hnsw:space=cosine` metadata on all collection creation sites — fixes broken similarity scoring under ChromaDB's default L2 distance (#807, #218)
 - File-level locking prevents duplicate drawers when agents mine the same file concurrently (#784, #826)
 - Hybrid closet+drawer retrieval — closets boost ranking, never gate results (#795)
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -561,12 +561,13 @@ def _fix_blob_seq_ids(palace_path: str) -> None:
                     "Skipped %d sysdb-10-format BLOB seq_id(s) in embeddings (not converting)",
                     skipped,
                 )
-            if not safe_rows:
-                return
-            updates = [(int.from_bytes(blob, byteorder="big"), rowid) for rowid, blob in safe_rows]
-            conn.executemany("UPDATE embeddings SET seq_id = ? WHERE rowid = ?", updates)
-            logger.info("Fixed %d BLOB seq_ids in embeddings", len(updates))
-            conn.commit()
+            if safe_rows:
+                updates = [
+                    (int.from_bytes(blob, byteorder="big"), rowid) for rowid, blob in safe_rows
+                ]
+                conn.executemany("UPDATE embeddings SET seq_id = ? WHERE rowid = ?", updates)
+                logger.info("Fixed %d BLOB seq_ids in embeddings", len(updates))
+                conn.commit()
     except Exception:
         logger.exception("Could not fix BLOB seq_ids in %s", db_path)
         return
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -517,6 +517,21 @@ def _fix_blob_seq_ids(palace_path: str) -> None:
     the Rust compactor to crash with "mismatched types; Rust type u64 (as SQL
     type INTEGER) is not compatible with SQL type BLOB".
 
+    Scoped to the ``embeddings`` table only. The ``max_seq_id`` table used
+    to be included in this loop, but chromadb 1.5.x writes its own BLOB
+    format there (``b'\\x11\\x11'`` + 6 ASCII digits). Misinterpreting that
+    format via ``int.from_bytes(..., 'big')`` yields a ~1.23e18 integer
+    that silently suppresses every subsequent write for the affected
+    segment (``embeddings_queue`` filters on ``seq_id > start``). chromadb
+    owns the ``max_seq_id`` column — we leave it alone. Palaces already
+    poisoned by the old behaviour can be repaired via
+    ``mempalace repair --mode max-seq-id``.
+
+    Defense-in-depth: rows with the sysdb-10 ``b'\\x11\\x11'`` prefix in
+    ``embeddings`` are skipped rather than converted. Real 0.6.x BLOBs are
+    pure big-endian u64 with no text prefix, so the prefix check is a
+    no-op for genuine legacy data.
+
     Must run BEFORE PersistentClient is created (the compactor fires on init).
 
     Opening a Python sqlite3 connection against a ChromaDB 1.5.x WAL-mode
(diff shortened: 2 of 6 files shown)


bensig (approved):
**Approve.** This is the deeper, correct fix for #1134. The shim narrowing is exactly right and the recovery sub-command is well-designed.

Full pytest on this branch: **1077 passed** (matches the PR body claim).

## Shim change — the right shape

Dropping \`max_seq_id\` from \`_fix_blob_seq_ids\` is the principled fix: chromadb 1.5.x writes its own BLOB format there (\`b'\\x11\\x11'\` + 6 ASCII digits) and we have no business converting it. The prefix check on the \`embeddings\` branch is good defense-in-depth — real 0.6.x BLOBs are pure big-endian u64 with no text prefix, so the check is a no-op for genuine legacy data. Worth the explicit log-on-skip so operators see when the guard fires.

Comment in \`_fix_blob_seq_ids\` is the right level of detail — names the symptom (\`embeddings_queue\` filters on \`seq_id > start\`, so a poisoned ~1.23e18 value silently suppresses every subsequent write). Future maintainer won't accidentally re-add \`max_seq_id\` to the loop.

## Recovery sub-mode — well-designed

\`mempalace repair --mode max-seq-id\` fits cleanly into the existing repair entry. Specific things I checked:

- **Detection threshold = 2^53.** Defensible — that's JS Number.MAX_SAFE_INTEGER, the practical correctness bound for any consumer that JSON-encodes the value. The actual corruption value (~1.23e18) is well above this; the threshold catches anything pathological without false-positiving on real high-throughput palaces.
- **Heuristic via \`MAX(embeddings.seq_id)\` over the owning collection.** PR body is honest about the trade-off: "for the sibling VECTOR segment it can land a few seq_ids ahead of the true max, which the queue treats as already consumed." That means a few legit-but-pending writes could be marked stale. Reasonable for an emergency recovery tool — better than resetting to 0 (which would re-process everything) or refusing to recover at all.
- **\`--from-sidecar\` for the careful path.** Letting users point at a pre-corruption \`chroma.sqlite3\` backup and copy clean values verbatim is the correct ergonomic for the case where they have one.
- **Backup before mutation, atomic UPDATEs, post-repair verification with \`MaxSeqIdVerificationError\` on detection-still-positive.** Standard data-integrity discipline applied where it should be.
- **\`_close_chroma_handles\` does \`gc.collect()\` to release mmap handles.** Brittle across chromadb versions but necessary — the current chromadb pattern doesn't expose an explicit handle-close. Worth a comment noting this is what we expect to need a touch-up if chromadb 1.6 or 2.x changes their cache structure, but fine as-is.

## Tests

- 8 new tests in \`test_repair.py\` covering: detect, heuristic, sidecar, dry-run, segment filter, no-op (no poisoned rows), backup, rollback-on-verify-failure.
- 3 new tests in \`test_backends.py\`: \`max_seq_id\` left untouched, sysdb-10 prefix skipped in \`embeddings\`, legacy big-endian u64 still converts.
- The "rollback on verify failure" test is the one I'd most want to see, and it's there.

## CHANGELOG

Already updated with a precise description of the failure mode and the recovery command. Nothing to add.

This PR + #1177 (skip _fix_blob_seq_ids on already-migrated palaces) are complementary — #1177's marker file means the migration loop runs at most once per palace, this PR makes that one run safe. They should land together. Recommend merging this one first since it's the substantive fix.

Closes #1134. Ship it.

igorls:
Rebased onto current `develop` (04c48dd) and pushed.

**Conflicts resolved:**
- `mempalace/repair.py` — additive: kept this PR's `repair_max_seq_id` + helpers alongside develop's new `status()` from #1227 (capacity check). Both are independent functions in the same module.
- `tests/test_backends.py` — kept all five tests covering both axes: this PR's three narrowing-behaviour tests (`does_not_touch_max_seq_id`, `skips_sysdb10_prefix_in_embeddings`, `still_converts_legacy_blobs_in_embeddings`) **and** develop's three marker-contract tests from #1177 (`writes_marker_after_blob_path`, `writes_marker_when_already_integer`, `skips_sqlite_when_marker_present`).
- `mempalace/backends/chroma.py` — kept develop's reworded `quarantine_stale_hnsw` log message ("integrity check failed" was a real semantic addition, not just line layout). The PR's "style revert" commit became a no-op against develop's reworded message and was dropped during rebase — that's correct.

**One follow-up commit** (`04c48dd`): the narrowed `_fix_blob_seq_ids` was returning early when `safe_rows` was empty, which broke #1177's contract that the marker is written on every successful pass (even no-op). This PR predates #1177's marker landing on develop, so the interaction wasn't visible until rebase. Restructured the conditional so the marker write is unconditional after a successful sqlite scan.

**Tests:** 1429 passed, 1 skipped on the rebased branch. `ruff check` + format clean.

Author preserved on the original commit; committer is mine (force-pushed via maintainer-edit access). Thanks for the deep dig on the sysdb-10 incident — the threshold + heuristic-vs-sidecar split is exactly the right shape.

Issue #1020, opened 2026-04-18
refactor: coerce None metadatas at ChromaDB backend boundary (supersedes per-site guards from #999, #1013)


bensig:
## Context

Two PRs recently merged to guard against \`None\` metadata / document entries returned by ChromaDB \`query()\`:

- **#999** — guards in \`searcher.py\` (\`search\`, \`search_memories\`), \`miner.py\` (\`status\`), \`mcp_server.py\` (4 sites) + regression tests. Merged at \`1b89b49\`.
- **#1013** — guards in \`layers.py\` (\`Layer3.search\`, \`Layer3.search_raw\`). Merged at \`6426669\`.

Both fix real user-visible crashes (\`AttributeError: 'NoneType' object has no attribute 'get'\`) that hit frequently on chromadb 1.5.x via the queue-stall in #1006.

Combined, these introduce ~9 per-site \`meta = meta or {}\` / \`doc = doc or \"\"\` guards across 5 files. During review of #999, @jphein proposed centralizing the coercion at the adapter boundary:

> \"Whack-a-mole across 8 call sites suggests the cleaner fix probably lives one layer down — either:
> 1. \`_get_cached_metadata\` / \`_fetch_all_metadata\` (in \`mcp_server.py\`) strip \`None\` entries as they build the cache, OR
> 2. \`ChromaCollection.get()\` / \`ChromaCollection.query()\` in \`backends/chroma.py\` coerce \`None\` metadatas to \`{}\` at the adapter boundary.
>
> Option 2 is more general — every future call site would get the guard for free... Happy to convert this to the adapter-level fix if maintainers prefer.\"

This issue tracks the adapter-level consolidation jphein offered.

## Proposal

Normalize \`None\` → \`{}\` (for metadatas) and \`None\` → \`\"\"\` (for documents) inside \`mempalace/backends/chroma.py\`'s \`ChromaCollection.query()\` and \`.get()\` return paths. Every consumer inherits the guard; every per-site \`meta or {}\` / \`doc or \"\"\` guard introduced by #999 and #1013 gets deleted.

## Alignment with RFC 001

This fits RFC 001's direction of formalizing backend return shapes (typed \`QueryResult\` / \`GetResult\`, normalized at the boundary). Arguably the coercion belongs in whatever the RFC 001 cleanup PR lands — merge coordination needed.

## Scope

- [ ] Add \`None\` → \`{}\` / \`\"\"\` coercion in \`ChromaCollection.query()\` return path
[...]

PR #1138, opened 2026-04-23
fix: reject non-http(s) LLM endpoints + clear ruff bugbear/silent-except findings

anthonyonazure:
## Summary

Two unrelated but small cleanup items surfaced by a ruff audit against the current codebase:

1. **Privacy hardening in `closet_llm.py`** — `LLMConfig` passes `LLM_ENDPOINT` straight to `urllib.request.urlopen` with no scheme check, so a misconfigured endpoint such as `file:///etc/passwd` would be read and forwarded as a \"Chat Completions\" request. Given the project's *privacy by architecture* guarantee, the constructor now validates the scheme and raises `ValueError` for anything other than `http`/`https`. All existing `LLMConfig` tests use `http://` endpoints, so they still pass.
2. **Ruff bugbear + silent-except cleanup** — fixes every finding under `ruff --select B,S110,S112` inside the `mempalace/` package:
   - `B904` exception chaining in `normalize.py` (2×) and `searcher.py` (1×).
   - `B007` unused loop variables renamed to `_name` in `dedup`, `dialect`, `layers`, `room_detector_local`.
   - `S110`/`S112` bare `try/except/pass` and `try/except/continue` replaced with `logger.debug(..., exc_info=True)` in `mcp_server`, `searcher`, `palace`, `palace_graph`, `miner`, `convo_miner`, `fact_checker`. Five of those files didn't previously have a module-level logger, so one was added using the existing `\"mempalace_mcp\"` logger name that `mcp_server.py` and `searcher.py` already use.

No behaviour change — failures that were already silent still don't propagate; they're just visible at `DEBUG` level now. Configured `ruff check` (E/F/W/C901) continues to pass.

## Test plan

- [x] `ruff check .` — all checks pass.
- [x] `ruff check --select B,S110,S112 mempalace/` — all checks pass.
- [x] `python -c \"import ast, pathlib; [ast.parse(p.read_text()) for p in pathlib.Path('mempalace').rglob('*.py')]\"` — every file parses.
- [ ] `pytest tests/` — not run locally (env without chromadb). CI will cover.

## Notes for reviewers

- Happy to split the endpoint-scheme commit into its own PR if that's preferred for release-note hygiene — it's the only functional change.
- Did **not** run `ruff format` because the local ruff is `0.15.11` and `.pre-commit-config.yaml` pins `0.4.10`; formatter output differs between those versions.
- Did **not** touch the `UP*` findings — `target-version = \"py39\"` in `pyproject.toml` blocks the `X | None` rewrites anyway.

### DIFF
--- a/mempalace/closet_llm.py
+++ b/mempalace/closet_llm.py
@@ -40,6 +40,7 @@ import json
 import os
 import re
 import time
+import urllib.parse
 import urllib.request
 import urllib.error
 from datetime import datetime
@@ -101,6 +102,14 @@ class LLMConfig:
         self.endpoint = (endpoint or os.environ.get("LLM_ENDPOINT", "")).rstrip("/")
         self.key = key or os.environ.get("LLM_KEY", "")
         self.model = model or os.environ.get("LLM_MODEL", "")
+        if self.endpoint:
+            # Privacy-by-architecture: reject file:// and other non-HTTP schemes
+            # so a misconfigured endpoint cannot exfiltrate local files.
+            scheme = urllib.parse.urlparse(self.endpoint).scheme.lower()
+            if scheme not in ("http", "https"):
+                raise ValueError(
+                    f"LLM_ENDPOINT must use http:// or https:// (got scheme {scheme!r})"
+                )
 
     def missing(self) -> list:
         missing = []
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -11,6 +11,7 @@ Same palace as project mining. Different ingest strategy.
 import os
 import sys
 import hashlib
+import logging
 from pathlib import Path
 from datetime import datetime
 from collections import defaultdict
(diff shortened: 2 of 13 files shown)


igorls:
Rebased onto develop. Conflict in `mempalace/palace_graph.py` was the logger-name change in this PR (`mempalace_mcp`) vs. the same-line edit in #1194 / #1020 follow-ups (`mempalace_graph`). Kept HEAD's `mempalace_graph` since that was a deliberate rename for the `palace_graph` module's own logger and #1194 added `_normalize_wing` adjacent to it; the original PR's logger rename to `mempalace_mcp` looked accidental. If that was intentional, please re-apply on top.

Verified locally on the rebased branch:
- `uvx ruff@0.4 check .` and `ruff format --check .` both clean
- Full test suite: **1552 passed, 1 skipped** (`pytest tests/ --ignore=tests/benchmarks`)

Net diff is 13 files / +32/-17 — small per-file ruff bugbear and silent-except cleanups, plus the `closet_llm` endpoint-scheme guard that's the headline privacy fix.

Issue #1145, opened 2026-04-24
followup(3.3.4): three small bugs found during v3.3.3 smoke test


igorls:
Three independent observations surfaced while smoke-testing v3.3.3 against real Claude Code transcripts. None block the 3.3.3 cut; filing for 3.3.4 triage.

---

## 1. `_wing_from_transcript_path` regex is macOS-centric

**File:** `mempalace/hooks_cli.py:490-505`

The regex only matches `-Projects-<project>` in the transcript path:

```python
match = re.search(r"-Projects-([^/]+?)(?:/|$)", normalized)
```

Claude Code encodes the full source directory into the project folder name. Users whose code lives outside `~/Projects/` — e.g. `~/dev/`, `~/src/`, `~/code/`, or any Linux convention — get transcript paths like `~/.claude/projects/-home-user-dev-MyProject-myproject/session.jsonl` with no `-Projects-` segment, so they fall through to the `wing_sessions` fallback and lose the per-project diary scoping that #659 was meant to deliver.

**Repro:** Stop-hook invocation with `transcript_path=~/.claude/projects/-home-<user>-dev-<Parent>-<project>/*.jsonl` → diary checkpoint lands in `wing_sessions` instead of `wing_<project>`.

**Suggested fix:** broaden the heuristic to also match the last path segment of the encoded project folder (after the final `-`), not just the `-Projects-<x>` slice.

---

## 2. `mempalace_diary_read` doesn't normalize empty-string `wing`

**File:** `mempalace/mcp_server.py` — `tool_diary_read` handler

`#1097` fixed `mempalace_search` to treat `wing=""` / `room=""` as "no filter" (LLM agents frequently default optional string parameters to `""`). The same fix isn't applied to `mempalace_diary_read` — passing `wing=""` returns `{"entries": [], "message": "No diary entries yet."}` even when entries exist under named wings.

**Repro:**

```python
diary_write(agent_name="x", entry="foo", wing="mempalace-dev")  # lands in mempalace-dev
diary_read(agent_name="x", wing="mempalace-dev")                # finds it
diary_read(agent_name="x", wing="")                             # returns empty, should return all
```

**Suggested fix:** apply the same empty-string → `None` normalization from #1097 to `diary_read` (and audit other tools with optional wing/room kwargs).

---

## 3. Stop-hook auto-mine subprocess ignores `MEMPALACE_PALACE_PATH` env

[...]

PR #1146, opened 2026-04-24
fix: diary wing routing for non-Projects layouts + empty-wing reads (#1145 bugs 1 & 2)

jphein:
Addresses two of the three findings @igorls flagged on [#1145](https://github.com/MemPalace/mempalace/issues/1145) from the v3.3.3 smoke test. Batched into one PR because both are in the same file class and share the "LLM agent / non-`~/Projects/` user defaults to empty string" failure mode. Bug 3 (Stop-hook auto-mine `MEMPALACE_PALACE_PATH` env propagation) lives in a different code path and is deferred to a follow-up PR pending local repro.

## Bug 1 — `_wing_from_transcript_path` is `~/Projects/`-centric

The heuristic only matched `-Projects-<project>` in the encoded transcript folder name. Users whose source lives outside `~/Projects/` — `~/dev/`, `~/src/`, `~/code/`, or any other Linux convention — got transcript paths like `~/.claude/projects/-home-<user>-dev-<Parent>-<project>/*.jsonl` with no `-Projects-` segment, fell through to the `wing_sessions` fallback, and lost the per-project diary scoping #659 was meant to deliver.

**Fix:** keep the explicit `-Projects-<x>` match first (preserves existing behavior / test coverage), then fall back to the last `-`-delimited segment of the encoded folder name. That segment is the actual project directory name regardless of source layout — Claude Code's encoding puts it there for any path, not just under `~/Projects/`.

## Bug 2 — `tool_diary_read` didn't adopt #1097's empty-string pattern

[#1097](https://github.com/MemPalace/mempalace/pull/1097) fixed `mempalace_search` to treat `wing=""` / `room=""` as "no filter" (LLM agents frequently default optional string parameters to `""`). `tool_diary_read` still defaulted an empty `wing` to `wing_<agent_name>`, silently hiding entries written to named wings — e.g., the per-project wings from #659 that the stop-hook now derives.

**Fix:** route `wing` through the `_sanitize_optional_name()` helper `#1097` introduced so empty / whitespace / `None` all coerce to `None` and drop out of the query. An explicit non-empty wing still scopes to that wing. Build the `where` clause from a list of required conditions (`room=diary`, `agent=<name>`) and conditionally append the wing filter — matches the pattern `tool_list_drawers` already uses.

## Tests

Six new tests, all green. Full suite: 1067 passed (this branch, from `upstream/develop`).

**`tests/test_hooks_cli.py`** — four `_wing_from_transcript_path` cases:
- `test_wing_from_transcript_path_non_projects_layout` — igor's specific scenario (`~/dev/<Parent>/<project>`)
- `test_wing_from_transcript_path_src_layout` — `~/src/<app>`
- `test_wing_from_transcript_path_code_layout` — `~/code/<app>`
- `test_wing_from_transcript_path_bare_folder_no_delimiters_falls_back` — genuinely unstructured parent → `wing_sessions`

**`tests/test_mcp_server.py::TestDiaryTools`** — two `tool_diary_read` cases:
- `test_diary_read_empty_wing_returns_across_wings` — `wing=""` returns entries from any wing (the bug as reported)
- `test_diary_read_explicit_wing_still_filters` — `wing="wing_a"` still scopes correctly; regression guard against overcorrecting the empty-string fix

## Not in this PR

Bug 3 from #1145 (Stop-hook auto-mine subprocess ignores `MEMPALACE_PALACE_PATH` env) — env-propagation bug in a different subprocess-launch site, same file but different narrative/tests. Deferred to a follow-up PR once I've reproduced locally and confirmed root cause. Happy to bundle if reviewers prefer one PR; say the word.

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -490,18 +490,44 @@ def _parse_harness_input(data: dict, harness: str) -> dict:
 def _wing_from_transcript_path(transcript_path: str) -> str:
     """Derive a project wing name from a Claude Code transcript path.
 
-    Claude Code stores transcripts at:
-        ~/.claude/projects/-home-<user>-Projects-<project>/session.jsonl
-    We extract <project> and return ``wing_<project>`` to match the
-    AAAK_SPEC convention (``wing_user``, ``wing_agent``, ``wing_code``,
-    ``wing_<project>``…). Falls back to ``wing_sessions``.
+    Claude Code encodes the full source directory path into the project
+    folder name, replacing ``/`` with ``-``. Transcripts live at:
+        ~/.claude/projects/<encoded-folder>/session.jsonl
+
+    We try two strategies in order:
+
+    1. Match ``-Projects-<project>`` if the source lives under
+       ``~/Projects/`` (the most common layout on our fork maintainer's
+       machine — preserving exact existing behavior).
+    2. Fall back to the last ``-``-delimited segment of the encoded
+       folder name — works for any layout (``~/dev/``, ``~/src/``,
+       ``~/code/``, etc.) because that segment is the actual project
+       directory name. Resolves #1145 bug 1.
+
+    Returns ``wing_<project>`` to match the AAAK_SPEC convention; falls
+    back to ``wing_sessions`` when the path doesn't match either pattern.
     """
     # Normalize path separators for cross-platform (Windows backslashes)
     normalized = transcript_path.replace("\\", "/")
+
+    # Strategy 1: explicit `-Projects-<project>` slice (preserves behavior
+    # for code living under ~/Projects/, which is what the existing tests
+    # assert against).
     match = re.search(r"-Projects-([^/]+?)(?:/|$)", normalized)
     if match:
         project = match.group(1).lower().replace(" ", "_")
         return f"wing_{project}"
+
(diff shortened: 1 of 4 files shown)


jphein:
Closing as duplicate — we filed #1146 and @igorls filed [#1147](https://github.com/MemPalace/mempalace/pull/1147) within four minutes of each other covering the same two bugs. Their PR is the right vehicle for 3.3.3:

- **Cleaner primary regex** — `/\.claude/projects/-([^/]+)` matches Claude Code's actual encoded-folder location as the primary strategy, with `-Projects-<x>` as the legacy fallback. That reflects the encoding rules more directly than my "`-Projects-` first, last-segment fallback" ordering.
- **CHANGELOG updated** — the 3.3.3 followup context is preserved in the release notes where it belongs.
- **Docstring update on `tool_diary_read`** is clearer about the "empty = no filter" contract.
- Tests cover the same ground; theirs have the macOS `/Users/` layout case which mine didn't.

Thanks @igorls for the smoke test + quick turnaround on both the findings and the fix. Re the Bug 3 retraction — confirmed on my side: `subprocess.Popen` without `env=` inherits parent env, `MempalaceConfig.palace_path` reads `MEMPALACE_PALACE_PATH` / `MEMPAL_PALACE_PATH` before falling back to config-file default. No propagation gap. Good catch on the measurement artifact.

No action needed on my end.

PR #1147, opened 2026-04-24
fix(3.3.3): two followups from #1145 before tag cut

igorls:
Two follow-up fixes surfaced by the v3.3.3 smoke test, landing before the `v3.3.3` tag is cut so they ship in this release rather than 3.3.4.

Filed as #1145. The env-propagation observation in that issue was a measurement artifact from concurrent background mines on different palaces — verified `MEMPALACE_PALACE_PATH` is honored by `subprocess.Popen` (inherits parent env by default). No code change needed for that one.

### 1. `_wing_from_transcript_path` — broaden beyond macOS `~/Projects/`

`mempalace/hooks_cli.py:490` only matched `-Projects-<name>` in encoded transcript folders, so Linux users with code under `~/dev/`, `~/code/`, `~/src/` fell through to the `wing_sessions` fallback and lost the per-project diary scoping that #659 was meant to deliver.

Fix derives the project name from the final dash-separated token of the encoded folder under `.claude/projects/` — this covers macOS `/Users/<user>/…/<project>`, Linux `/home/<user>/dev/<parent>/<project>`, and nested layouts while preserving the legacy `-Projects-` match as a secondary fallback for transcripts outside the standard Claude Code path.

### 2. `mempalace_diary_read(wing="")` — return entries across all wings

After #659, hook-driven diary writes land in `wing_<project>` instead of the agent's default `wing_<agent>`. But `diary_read` with `wing=""` defaulted to `wing_<agent>`, invisible-ing those hook-written entries.

Fix applies the #1097 "empty-string as no filter" pattern: when `wing` is empty, filter by `agent + room=diary` only. Explicit `wing=<name>` continues to scope to that wing.

## Verification

- `uv run pytest tests/` — **1075 passed** (includes 4 new tests: 3 wing-derivation layouts, 1 cross-wing diary read)
- `uvx --from 'ruff>=0.4.0,<0.5' ruff check` — **All checks passed** (CI parity)
- `uvx --from 'ruff>=0.4.0,<0.5' ruff format --check` — **4 files already formatted**

## Changelog

Entries appended to the existing `[3.3.3]` Bug Fixes section — no new version section, since these ship as part of 3.3.3.

## Sequencing

The `v3.3.3` tag hasn't been pushed yet (confirmed against `git ls-remote --tags origin`). Merging this before `git tag -a v3.3.3` folds both fixes into the release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Real `python3` resolution for `.sh` hooks with a `MEMPAL_PYTHON` override path. (#833)
 - Add optional `wing` parameter to `tool_diary_write` / `tool_diary_read` and derive per-project wing from the Claude Code transcript path when writing from the stop hook — diary entries from different projects no longer collapse into a shared default wing. (#659)
 - Treat empty string as "no filter" in `mempalace_search` `wing`/`room`; LLM agents that default to filling every optional parameter with `""` no longer get bounced with `must be a non-empty string`. (#1097, #1084)
+- Broaden `_wing_from_transcript_path` to handle Claude Code project folders without a `-Projects-` segment (e.g. `~/dev/<parent>/<project>`, `~/code/<project>`). The project name is now derived from the final dash-separated token of the encoded folder, so Linux users with code outside `~/Projects/` get per-project diary scoping instead of falling through to `wing_sessions`. (#1145, follow-up to #659)
+- `mempalace_diary_read(wing="")` now returns diary entries from every wing this agent has written to, matching the #1097 "empty-string as no filter" pattern. Previously defaulted to `wing_<agent>`, siloing entries that hooks wrote to project-derived wings. (#1145)
 
 ### Improvements
 
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -490,14 +490,29 @@ def _parse_harness_input(data: dict, harness: str) -> dict:
 def _wing_from_transcript_path(transcript_path: str) -> str:
     """Derive a project wing name from a Claude Code transcript path.
 
-    Claude Code stores transcripts at:
+    Claude Code encodes the project's source directory by replacing path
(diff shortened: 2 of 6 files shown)


bensig (approved):
Read + ran locally against \`fix/3.3.3-followups\` (commits d158375, 1fd16da, 6fcfd34). **LGTM, ship it.**

## Diff review

**Bug 1 — \`_wing_from_transcript_path\` broadening (hooks_cli.py)**
- Primary path: match \`/\.claude/projects/-<encoded>\` and take the final dash-separated token. Covers macOS \`~/Projects/\`, Linux \`~/dev/\`, \`~/code/\`, \`~/src/\`, and nested layouts. Correct shape.
- Legacy \`-Projects-<name>\` regex retained as secondary fallback for transcripts outside the standard Claude Code projects dir. Defensive, doesn't interfere with the primary.
- Three new unit tests cover non-Projects Linux, macOS users-folder, and nested-deep. Good coverage.

**Bug 2 — \`tool_diary_read(wing='')\` empty-string normalization (mcp_server.py)**
- Removes the \`wing_<agent>\` auto-fallback; builds a \`conditions\` list that always scopes by \`room=diary\` + \`agent\` and only adds \`wing\` when truthy.
- Matches the #1097 pattern applied to \`mempalace_search\`. Semantically consistent.
- Subtle behavior change worth noting in docstring (which you did): old default was "this agent's default wing"; new default is "all wings this agent has written to." The new default is clearly right given hooks write to project-derived wings via #659 — entries would otherwise be invisible.
- Test covers both the empty-wing broadening case and the explicit-wing still-scoped case. Good.

## Local verification

\`\`\`bash
$ pytest tests/ --ignore=tests/benchmarks --cov=mempalace --cov-fail-under=80 -q
1075 passed, 1 warning in 76.62s
Required test coverage of 80% reached. Total coverage: 83.35%
\`\`\`

All green. 4 new tests land cleanly (3 for bug 1, 1 for bug 2).

## CHANGELOG

Already updated in 6fcfd34 with accurate bullets in the \`[3.3.3]\` section. Nothing to add.

Ready for tag as soon as this merges. Thanks for the thorough smoke test.

PR #1148, opened 2026-04-24
feat(init): scan manifests and git authors for real entity signal (v1)

igorls:
## Summary

`mempalace init` now gets project names from **package manifests** (`package.json`, `pyproject.toml`, `Cargo.toml`, `go.mod`) and people from **git commit authors**, rather than inferring both from regex over prose. The regex detector stays as a fallback for prose-only folders.

This is **version 1** — it fixes the codebase/project case. The prose-only case (diaries, transcripts, research notes with no git and no manifests) still depends on the regex detector and remains limited; phase 2 is an LLM-assisted entity pass for that use case.

## Motivation

Running `mempalace init` on a codebase produced systematic false positives from the regex-only path. Examples from a local sanity run:

- "Projects" included common English words caught by the old `versioned/hyphenated` regex on any hyphenated compound (e.g. a prefix like `multi` matching `multi-word`, a descriptor like `hybrid` matching `hybrid-search`).
- "Uncertain" was a dumping ground for capitalized sentence-starts (`Never`, `Before`, past-tense verbs like `Created`, `Extracted`, `Processed`).
- CamelCase project names were invisible because the candidate regex `[A-Z][a-z]{1,19}` requires a lowercase run after one capital and fragmented `FooBar` into `Foo` and `Bar` separately.
- The regex path produced zero people for code repositories — there is no `said`/`laughed`/`asked` verb pattern in commit messages or READMEs.

After this PR, a codebase init pulls the canonical project name from its manifest and ranks contributors by commit volume, with the regex path filling in extra names only mentioned in prose.

## What's in this PR

**New module `mempalace/project_scanner.py`:**
- Parses manifests for canonical project names
- Walks for `.git` directories to find repo boundaries
- Collects authors via `git log`, filters bots by name pattern (`[bot]`, `dependabot`, `renovate`, `-autoroll`, trailing `bot`, …)
- Keeps `@users.noreply.github.com` identities — that's GitHub's privacy-protected format for real humans, not a bot marker
- Union-find dedup: commits sharing a name OR an email collapse into one person, so the same contributor with multiple display-name / email aliases appears once
- Marks "mine" projects: current git user is top-5 committer OR has ≥10% of commits OR ≥20 commits
- Public entry point `discover_entities()` merges scanner results with the regex detector case-insensitively (so `foo` from a manifest absorbs `Foo` from docs) and suppresses the regex `uncertain` bucket when real signal is already present

**Regex detector improvements (apply to the prose-only fallback):**
- CamelCase extraction pattern — multi-capital names like `FooBar`, `FooBAR`, `FooAPI` are visible
- Dialogue `^NAME:\s` requires ≥2 matches to count. A single metadata line like `Created: 2026-04-21` was scoring as dialogue and classifying the field name as a person.
- Versioned/hyphenated pattern tightened to `\b{name}[-_]v?\d+(?:\.\d+)*\b` (version-only). The previous `\b{name}[-v]\w+` matched every hyphenated compound.
- Skip LICENSE/COPYING/NOTICE/AUTHORS/PATENTS files (pure-prose noise)
- Extra SKIP_DIRS: `.terraform`, `vendor`, `target`
- Stopword expansion: capitalized participles/descriptors common at sentence start (`created`, `updated`, `extracted`, `processed`, `total`, `auto`, `multi`, `hybrid`, `context`, `batch`, `never`, `before`, `after`, …)
- `classify_entity`: high-pronoun single-category signal now classifies as person. The previous "require two signal categories" rule demoted a character referenced mostly with pronouns (the common case in narrative prose) to uncertain despite overwhelming evidence. New gate: `pronoun_hits >= 5 AND ratio >= 0.2` so common sentence-start words with incidental proximity stay uncertain.

**Integration:**
- `cli.py cmd_init` switches to `discover_entities()`; same output shape so `confirm_entities` is unchanged.

**Tests:**
- 39 new tests (`tests/test_project_scanner.py`) — manifest parsing, bot filtering, union-find dedup, git repo discovery, scan + merge + fallback
- 2 new regression tests for `classify_entity`
- All existing tests still pass; ruff clean

## Known limitations (addressed in phase 2)

1. **Prose-only folders still limited.** A folder with no git and no manifests falls back to the regex detector. Common-word noise is reduced by the stopword expansion but fundamentally unsolvable without semantic understanding — a short capitalized token can be a project name, a city, or a character.
2. **Sample size.** The regex detector reads 10 files × 5KB each. Fine for sampling, but misses the tail in large prose corpora. Scanner path has no such limit for git/manifests.
3. **No NER / no LLM.** Either would properly disambiguate ambiguous tokens.

Phase 2 plan: optional local-model or user-configured endpoint at init time to classify the top capitalized terms in a sampled text window. Consistent with MemPalace's local-first / zero-API-required default — opt-in, not required.

## Test plan

- [x] `uv run pytest tests/ --ignore=tests/benchmarks` — full suite passes
- [x] `ruff check mempalace/ tests/` — clean
- [x] `ruff format --check mempalace/ tests/` — clean
- [x] `python -m mempalace.project_scanner .` on this repo — canonical project name from `pyproject.toml` plus contributor list from `git log`
- [ ] Test on a prose-only folder (diary/transcript) to surface where the regex fallback still falls short — feeds phase 2 scope

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -71,7 +71,8 @@ def _ensure_mempalace_files_gitignored(project_dir) -> bool:
 def cmd_init(args):
     import json
     from pathlib import Path
-    from .entity_detector import scan_for_detection, detect_entities, confirm_entities
+    from .entity_detector import confirm_entities
+    from .project_scanner import discover_entities
     from .room_detector_local import detect_rooms_local
 
     cfg = MempalaceConfig()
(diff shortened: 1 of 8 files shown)


igorls:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/MemPalace/mempalace/pull/1148#pullrequestreview-4167648201)

PR #1157, opened 2026-04-24
feat(init): wire confirmed entities into the miner's known-entities registry

igorls:
## Summary

The init step's entity output was a dead file. `miner.py` has always read `~/.mempalace/known_entities.json` to tag drawer metadata with recognized names, but nothing ever wrote it — so every improvement to init detection (manifest/git/regex/LLM) stopped at `<project>/entities.json` and never reached the path that actually uses it.

This wires init → registry. Per-project file is kept as an audit trail.

## Measured value

On a representative prose snippet (eight sentences mentioning six real people and four real projects):

| Registry state | Entities recognized |
|---|---|
| Empty (current) | 0 |
| Populated by init | 12, all correct |

Multi-word names (`Alice Example`, `Bob Sample`) fail the frequency-threshold fallback because each word only appears once. Lowercase / hyphenated project names (`my-lib`, `foo-bar`) don't match the CamelCase regex. Both categories were completely invisible to the miner until now. Every recognized name becomes a semicolon-separated tag on the drawer, which ChromaDB uses for entity-filtered search.

## Why stacked on #1150

Logically independent from the LLM refinement, but each earlier PR in the stack improves the *input* to the registry: #1148 added manifest/git authors, #1150 adds LLM-classified topics/people. This PR ensures all of that work reaches the one code path that uses it. Merging order is linear: #1148 → #1150 → this.

## Implementation

**`miner.add_to_known_entities({category: [names]}) -> str`** (new):
- Reads the existing registry, unions each category, writes back.
- Case-insensitive dedup, preserves first-seen casing.
- Tolerant of both on-disk shapes the miner already supports: list of names, or dict mapping `name → code` (dialect-style). In the dict case, new names are added as keys with `None` values so existing codes aren't overwritten.
- Preserves untouched categories — merging `{people: [...]}` does not clobber an existing `places` or `projects` category.
- Invalidates the in-process mtime cache so `cmd_init` → `cmd_mine` in one run sees the write immediately.
- Writes with `ensure_ascii=False` so non-ASCII names stay readable.
- `chmod 0o600` — the registry mirrors the user's confirm-step PII.

**`cmd_init`** now calls it at the end of the confirm-entities step, after the per-project `entities.json` is written.

## Tests

17 new tests (`tests/test_known_entities_registry.py`), all offline:
- Fresh-file creation
- List-category union + case-insensitive dedup
- Preservation of untouched categories
- Dict-format registries (new keys added with None; existing codes preserved)
- Malformed / non-dict file recovery (starts fresh)
- Cache invalidation (same-process read reflects the write)
- Unicode round-trip
- End-to-end verification that `_extract_entities_for_metadata` picks up every registered name

Full suite: 1221 passed, ruff clean.

## Test plan

- [x] `uv run pytest tests/ --ignore=tests/benchmarks` — full suite passes
- [x] `ruff check mempalace/ tests/` — clean
- [x] `ruff format --check mempalace/ tests/` — clean
- [x] Measured: 0 → 12 entity recognition delta on representative prose sample
- [ ] Reviewer verification: run `mempalace init <repo>` then `mempalace mine <repo>`, confirm drawer metadata contains the registered names

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -125,7 +125,7 @@ def cmd_init(args):
         # global registry the miner reads at mine time.
         if confirmed["people"] or confirmed["projects"]:
             entities_path = Path(args.dir).expanduser().resolve() / "entities.json"
-            with open(entities_path, "w") as f:
+            with open(entities_path, "w", encoding="utf-8") as f:
                 json.dump(confirmed, f, indent=2, ensure_ascii=False)
             print(f"  Entities saved: {entities_path}")
 
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -120,12 +120,19 @@ def cmd_init(args):
     total = len(detected["people"]) + len(detected["projects"]) + len(detected["uncertain"])
     if total > 0:
         confirmed = confirm_entities(detected, yes=getattr(args, "yes", False))
-        # Save confirmed entities to <project>/entities.json for the miner
+        # Save confirmed entities to <project>/entities.json (per-project
+        # audit trail — user can inspect or hand-edit) AND merge into the
+        # global registry the miner reads at mine time.
         if confirmed["people"] or confirmed["projects"]:
             entities_path = Path(args.dir).expanduser().resolve() / "entities.json"
             with open(entities_path, "w") as f:
-                json.dump(confirmed, f, indent=2)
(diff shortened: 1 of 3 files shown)


igorls:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/MemPalace/mempalace/pull/1157#pullrequestreview-4168006968)

PR #167, opened 2026-04-07
fix: defer KnowledgeGraph init to first use in MCP server

adv3nt3:
## Summary
`mcp_server.py` line 34 runs `_kg = KnowledgeGraph()` at **module level**. `KnowledgeGraph.__init__` creates the parent directory and opens/creates a SQLite database. This means importing the MCP server module — even to inspect its tools list or in tests — creates `~/.mempalace/knowledge_graph.sqlite3` as a side effect.

## Fix
Replace `_kg = KnowledgeGraph()` with `_kg = None` and a lazy `_get_kg()` getter that creates the instance on first call. All 5 call sites updated from `_kg.method()` to `_get_kg().method()`.

```python
_kg = None

def _get_kg():
    global _kg
    if _kg is None:
        _kg = KnowledgeGraph()
    return _kg
```

## Changes
1 file changed (`mempalace/mcp_server.py`), 14 insertions, 6 deletions.

## Test plan
- [x] `ruff check` + `ruff format --check` pass
- [x] `python3 -m py_compile` compiles OK
- [x] Pyright reports 0 new diagnostics from this change
- [x] All 5 `_kg.` call sites updated to `_get_kg().`
- [x] Lazy init pattern verified via Context7 CPython docs

Refs: #159 (point 8)

PR #1160, opened 2026-04-24
fix(mcp): lazy per-path KnowledgeGraph cache (#1136)

### LINKED ISSUE #1136: KnowledgeGraph module-level singleton prevents safe multi-tenant use
## Summary
`mempalace.mcp_server._kg` is constructed once at module import, so any deployment that wants to route the same Python process for multiple isolated palaces (multi-user hosted setups, per-user background workers, per-project process pools) can only target one SQLite file for the lifetime of the process. `MEMPALACE_PALACE_PATH` env var changes between calls are silently ignored by the KG (they still work for Chroma because `get_collection` reads env vars per-call).

## Reproduction (v3.3.0)
```python
import mempalace.mcp_server as mcp
import os

# Pretend we're a multi-tenant dispatcher:
os.environ["MEMPALACE_PALACE_PATH"] = "/tmp/tenant_a"
mcp.tool_kg_add(subject="alice_secret", predicate="owns", object="repo_a")
os.environ["MEMPALACE_PALACE_PATH"] = "/tmp/tenant_b"
facts = mcp.tool_kg_query(entity="alice_secret")
# Expected for tenant B: empty
# Actual: returns tenant A's fact — both calls hit /tmp/.mempalace or whatever path was resolved at import
```

## Root cause (code path)
`mcp_server.py` around line 76:
```python
_kg = KnowledgeGraph(db_path=os.path.join(_config.palace_path, "knowledge_graph.sqlite3"))
```
- `_config` is a module-level `MempalaceConfig()` built at import
- `_config.palace_path` is a property that reads env, but this read happens once at `_kg` construction
- After that, `_kg.db_path` is frozen; env var changes do not rebind it

The rest of the library is already per-call safe:
- `_get_collection` reads `_config.collection_prefix` as a property → ChromaDB routing follows env changes correctly
- Only the KG singleton is constructed eagerly

## Impact
For self-hosted single-user setups (the target user in `README.md`) this is benign — there is only one palace anyway.

For any deployment that wants to route the same process to multiple isolated palaces, the singleton forces all callers into the same SQLite file, so per-palace isolation via env vars is structurally impossible. Reviewing your own issue history it looks like multi-palace support is on the roadmap anyway; this is one of the concrete blockers.

A local workaround exists (swapping `mcp._kg` per call under a dispatch lock and restoring it in `finally`), but it monkey-patches library internals and is fragile against upstream refactors. Happy to send a PR that implements one of the upstream-friendly options below if you confirm the direction.

### Option A — Lazy per-call property
Make `_kg` a module-level property that reads the current palace_path on access:

```python
# mcp_server.py
_kg_by_path: dict[str, KnowledgeGraph] = {}

def _get_kg() -> KnowledgeGraph:
    path = os.path.join(_config.palace_path, "knowledge_graph.sqlite3")
    if path not in _kg_by_path:
        _kg_by_path[path] = KnowledgeGraph(db_path=path)
    return _kg_by_path[path]
```
And replace `_kg.x(...)` call sites with `_get_kg().x(...)`. Caches per path so repeated calls stay cheap.

Pro: minimal behavior change for single-user, works for multi-palace out of the box.
Con: 5 call-site replacements in `mcp_server.py`. Need to audit for thread-safety of the cache dict.

### Option B — Path parameter on every tool function
```python
def tool_kg_add(..., db_path: str = None):
    kg = KnowledgeGraph(db_path=db_path) if db_path else _get_default_kg()
    kg.add_triple(...)
```
Pro: fully explicit, no globals.
Con: API change, every caller has to pass the path.

### Option C — Thread-local `_kg`
Stash `_kg` in a `threading.local()` and set it at request boundaries.
Pro: backwards compatible.
Con: not reentrant across asyncio.to_thread, easier to misuse.

## Preferred direction
Option A. Minimal API change, same ergonomics for single-user, correct for multi-palace. Happy to PR if you agree.

## Environment
- mempalace v3.3.0
- Python 3.14

mvalentsev:
## Summary

Replaces the module-level `_kg = KnowledgeGraph(...)` singleton in
mcp_server.py with a lazy, per-path cache. The KG path is now resolved
on each tool invocation, so a multi-tenant host can rotate
`MEMPALACE_PALACE_PATH` between calls and have each call address the
correct sqlite file. The ChromaDB side of the server was already
per-call via `_get_client()`; this brings the KG side in line.

Closes #1136.

## Before / After

Before (module scope, import-time):

```python
if _args.palace:
    _kg = KnowledgeGraph(db_path=...)
else:
    _kg = KnowledgeGraph()
# tool call: _kg.add_triple(...)
```

After:

```python
_kg_by_path: dict[str, KnowledgeGraph] = {}
_kg_cache_lock = threading.Lock()

def _get_kg() -> KnowledgeGraph:
    path = os.path.abspath(_resolve_kg_path())
    kg = _kg_by_path.get(path)       # lock-free hot path
    if kg is not None:
        return kg
    with _kg_cache_lock:              # cold path, double-checked
        kg = _kg_by_path.get(path)
        if kg is None:
            kg = KnowledgeGraph(db_path=path)
            _kg_by_path[path] = kg
    return kg
# tool call: _get_kg().add_triple(...)
```

Follows Option A from the reporter in #1136.

## Context: earlier attempt

A previous attempt in #167 introduced a lazy single-instance init. It
was closed on 2026-04-12 with a note saying it had been reworked in
#647 and #667, but neither of those PRs actually touched the singleton
pattern. That is why #1136 landed on current develop with the eager
init still in place. Easy to verify against `upstream/develop` at
`mempalace/mcp_server.py:101` before this PR.

## Scope notes

- Strict backward-compat for the default path. When the server starts
  without `--palace`, KG keeps writing to
  `~/.mempalace/knowledge_graph.sqlite3` regardless of
  `MEMPALACE_PALACE_PATH`. Multi-tenant path rotation applies only
  when `--palace` was given at startup. The "no --palace but env var
  set" case is the subject of #540 and is explicitly out of scope here.
- No atexit cleanup of KG connections; same posture as before this PR.
- No eviction policy on the cache. Per-process palace count is small;
  each KnowledgeGraph holds one sqlite connection plus a lock.

## Test plan

- [x] `_patch_mcp_server` helper patches `_get_kg` instead of `_kg`
- [x] `test_lazy_init_no_import_side_effect`: fresh subprocess import
      under `HOME=tmp` does not create the sqlite file
- [x] `test_get_kg_returns_same_instance`: cache hits return `is`-equal
- [x] `test_get_kg_different_paths_different_instances`: rotating env
      gives distinct KGs
- [x] `test_multi_tenant_env_switch`: reporter's scenario (write under
      A, query under B is empty, query under A again returns the fact)
- [x] `test_cache_thread_safe`: 16 concurrent `_get_kg()` end up with
      one shared instance and one cache entry
- [x] Full `pytest tests/ --ignore=tests/benchmarks`: 1080 passed

## Breaking changes (internal)

Direct attribute access to `mempalace.mcp_server._kg` is removed. This
is an underscore-prefixed private attribute; the only in-tree callers
were three test files, all migrated in this PR.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1418,10 +1418,11 @@ def tool_memories_filed_away():
 
 
 def tool_reconnect():
-    """Force the MCP server to drop the cached ChromaDB collection and reconnect.
+    """Force the MCP server to drop cached ChromaDB + KnowledgeGraph state.
 
     Use after external scripts or CLI commands modify the palace database
-    directly, which can leave the in-memory HNSW index stale.
+    or replace ``knowledge_graph.sqlite3`` directly, which can leave the
+    in-memory HNSW index stale or pin a closed-on-disk SQLite connection.
     """
     global \
         _client_cache, \
@@ -1439,6 +1440,15 @@ def tool_reconnect():
     # still applies after the reconnect.
     _vector_disabled = False
     _vector_disabled_reason = ""
+    # Drain the per-path KnowledgeGraph cache so a replaced sqlite file is
+    # reopened on the next tool call rather than served from a stale handle.
+    with _kg_cache_lock:
+        for kg in _kg_by_path.values():
+            try:
+                kg.close()
+            except Exception:
+                pass
+        _kg_by_path.clear()
     try:
         col = _get_collection()
         if col is None:
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -105,18 +105,8 @@ if _args.palace:
 
 _config = MempalaceConfig()
 
-# Lazy per-path KG cache. Import no longer creates the sqlite file as a side
-# effect (see issue #1136). The path is resolved on each tool call so that a
-# multi-tenant host rotating MEMPALACE_PALACE_PATH between calls routes each
-# call to the correct KG file, matching the per-call behavior of _get_client()
-# on the ChromaDB side.
 _kg_by_path: dict[str, KnowledgeGraph] = {}
 _kg_cache_lock = threading.Lock()
-
-# Whether --palace was given at startup. Controls default-path resolution:
-# with the flag, KG follows _config.palace_path per call; without it, KG stays
-# on DEFAULT_KG_PATH regardless of env var (issue #540's territory, out of
-# scope here).
 _palace_flag_given: bool = bool(_args.palace)
 
 
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -46,6 +46,7 @@ import argparse  # noqa: E402  (deferred until after stdio protection above)
 import json  # noqa: E402
 import logging  # noqa: E402
 import hashlib  # noqa: E402
+import sqlite3  # noqa: E402
 import threading  # noqa: E402
 import time  # noqa: E402
 from datetime import date, datetime  # noqa: E402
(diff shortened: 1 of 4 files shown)


Qodo-Free-For-OSS:
Hi, tool_reconnect() clears only the ChromaDB client/collection caches and leaves the per-path KnowledgeGraph cache intact, so after external edits/replacement of knowledge_graph.sqlite3 the server can keep using an old open SQLite connection.

**How to fix:** Invalidate KG cache on reconnect

> ### Issue description
> `tool_reconnect()` currently resets only ChromaDB caches. With the new per-path KG cache, `KnowledgeGraph` instances can remain open and be reused after an external KG file replacement/rebuild, leading to stale results and lingering handles.
> 
> ### Issue Context
> `_get_kg()` caches `KnowledgeGraph(db_path=path)` objects in `_kg_by_path`. `KnowledgeGraph` holds an open `sqlite3.Connection` until `close()` is called.
> 
> ### Fix Focus Areas
> - mempalace/mcp_server.py[108-129]
> - mempalace/mcp_server.py[1342-1381]
> 
> ### Suggested change
> In `tool_reconnect()`, iterate over `_kg_by_path.values()` and call `close()` (best-effort), then clear `_kg_by_path` (optionally under `_kg_cache_lock`) so the next tool call reopens the KG from disk.

---

mvalentsev:
Heads-up: `b8816e0` adds a `_call_kg(op)` retry wrapper for the concurrent-close race between a KG handler and `tool_reconnect`. All five KG handlers (`tool_kg_query` / `add` / `invalidate` / `timeline` / `stats`) now route through `_call_kg(lambda kg: kg.method(...))`. On `sqlite3.ProgrammingError: Cannot operate on a closed database` the wrapper evicts the stale entry from `_kg_by_path` (only if the slot still points at the closed instance -- another thread may have already replaced it) and retries once with a fresh `_get_kg()`. Beyond one retry the error propagates -- a sustained close stream is a different bug and a hung loop is worse than a clear failure surface.

Adds `import sqlite3` plus three tests in `tests/test_mcp_server.py::TestKGLazyCache`:

- `test_call_kg_retries_after_concurrent_close` -- closed KG -> retry returns from a fresh KG.
- `test_call_kg_does_not_retry_on_other_errors` -- non-`ProgrammingError` exceptions propagate without retry; we don't want the guard masking real bugs.
- `test_call_kg_gives_up_after_one_retry` -- the loop is bounded; sustained closes raise.

The body's `_get_kg()` call-site example is still the right way to access the KG; this is the supplementary safety layer between `_get_kg()` and the actual sqlite3 call.

Issue #1161, opened 2026-04-24
HNSW sparse bloat + chromadb segfaults persist even after hnsw:num_threads metadata is set (config not persisted by chromadb 1.5.x)


imtylervo:
## Symptom

On a freshly patched install where `hnsw:num_threads: 1` is passed via `metadata=` to `get_or_create_collection` / `create_collection` (the fix in #991 and part of #976), the guard still catches sparse `link_lists.bin` bloat within an hour, and the kernel log records segfaults in `chromadb_rust_bindings.abi3.so` whenever multiple processes touch the palace concurrently:

```
kernel: tokio-rt-worker[229244]: segfault at 88 ip ... in chromadb_rust_bindings.abi3.so[32d4...]
kernel: tokio-rt-worker[230410]: segfault at 88 ip ... in chromadb_rust_bindings.abi3.so[32d4...]
```

```
mempalace-disk-monitor.sh: HNSW sparse bloat: .../link_lists.bin claims 1337GB. Truncating (chroma rebuilds).
```

## Environment

- OS: Ubuntu 24.04 (GCP VM)
- Python: 3.12
- chromadb: 1.5.8
- mempalace: 3.3.2 (pip `--user --break-system-packages`)
- Used concurrently by: Claude Code Stop/PreCompact hooks (per turn), a cron health-check (every 30 min), an hourly disk-guard timer, and the MCP server.

## Reproduction

Restore a palace from backup SQLite, then run these concurrently (e.g. via `threading.Thread` spawning `subprocess.run`):

```python
subprocess.run(["python3", "-m", "mempalace", "mine", small_dir_A])
subprocess.run(["python3", "-m", "mempalace", "mine", small_dir_B])
subprocess.run(["python3", "-m", "mempalace", "mine", small_dir_C])
```

Observe:
- `link_lists.bin` grows to sparse >100GB apparent size within a few minutes
- `journalctl -k | grep segfault.*chromadb` shows crashes
- Each subsequent mine may fail with `InternalError: Error constructing hnsw segment reader: Error deserializing pickle file: eval error at offset 0: EOF while parsing` (truncated by the guard)

## Root causes

[...]

PR #1162, opened 2026-04-24
fix: serialize ChromaCollection writes through palace lock

imtylervo:
#976 protects `mempalace mine`, but MCP/direct backend writers still call `ChromaCollection.add/upsert/update/delete` without the palace lock. This PR moves the lock boundary to the Chroma backend seam so all Chroma writes share the same palace-level serialization, with a re-entrant guard for miner paths that already hold the lock.

## Why this matters

`mine_palace_lock(palace_path)` closes the mine fan-out hole, but its current coverage is the `miner.mine()` pipeline. Other writers can still reach Chroma directly through the backend adapter, including MCP write tools and third-party callers that use `ChromaCollection`.

When one of those writers runs concurrently with `mempalace mine` or another direct writer, it can still enter ChromaDB's HNSW write path concurrently. That is the gap this PR closes.

## What changed

- `ChromaCollection` now accepts an optional `palace_path`.
- `ChromaCollection.add/upsert/update/delete` acquire `mine_palace_lock(palace_path)` when a palace path is available.
- `ChromaBackend.get_collection/create_collection` and MCP `_get_collection()` pass the palace path through.
- `mine_palace_lock` is now re-entrant for the same thread and palace key, so `miner.mine()` can hold the outer lock while internal collection writes pass through safely.
- The re-entrant holder state is tagged with the current PID so forked children do not inherit lock-holder credit from the parent.

`palace_path=None` preserves the previous no-lock behavior for legacy/tests/direct construction.

## Production evidence

An equivalent local patch on a large production palace caught the gap under live load:

- A `mempalace mine` subprocess held the palace lock while ingesting project transcripts.
- A sibling MCP `diary_write` attempted to write during that window and was blocked by the patched ChromaCollection lock instead of entering Chroma concurrently.
- Before this lock layer, the same workload had produced HNSW index bloat in `link_lists.bin` on the order of tens of GB.
- After the patch, the palace stayed healthy under the same workload: no new Chroma/Rust segfaults, no new auto-quarantine batch, and `link_lists.bin` remained in the KB range.

The intended trade-off is: fail/skip a concurrent write loudly rather than corrupting the HNSW index.

## Tests

```console
$ pytest tests/test_chroma_collection_lock.py tests/test_palace_locks.py tests/test_backends.py -q
....................................................                     [100%]
52 passed in 1.09s
```

Coverage added:

- `palace_path=None` skips locking for backward compatibility.
- Write methods raise `MineAlreadyRunning` while another process holds the palace lock, without calling the underlying Chroma collection.
- Re-entrant writes inside an existing `mine_palace_lock` do not deadlock.
- Concurrent direct writers serialize through the palace lock.
- Read paths do not acquire the write lock.

## Scope

This PR intentionally only covers the correctness layer: serializing Chroma writes through the existing palace lock.

Deferred follow-up: a durable pending-write queue for MCP/autosave writes that encounter a busy palace lock, so agent autosaves can be replayed later instead of being dropped.

Refs #1161.

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -1,5 +1,6 @@
 """ChromaDB-backed MemPalace storage backend (RFC 001 reference implementation)."""
 
+import contextlib
 import datetime as _dt
 import logging
 import os
@@ -573,10 +574,43 @@ def _as_list(v: Any) -> list:
 
 
 class ChromaCollection(BaseCollection):
-    """Thin adapter translating ChromaDB dict returns into typed results."""
+    """Thin adapter translating ChromaDB dict returns into typed results.
 
-    def __init__(self, collection):
+    When ``palace_path`` is set, all write methods (``add``, ``upsert``,
+    ``update``, ``delete``) acquire ``mine_palace_lock(palace_path)`` for the
+    duration of the underlying chromadb call. This serializes MCP and other
+    direct-backend writers against ``mempalace mine`` and against each other,
+    closing the race between concurrent writers that triggers ChromaDB's
+    multi-threaded HNSW corruption (#974/#965).
+
+    The lock is the same primitive used by ``miner.mine()`` so re-entrant
+    acquisition from inside the mine pipeline (mine -> _mine_body ->
+    collection.upsert) is short-circuited by the per-thread guard inside
+    ``mine_palace_lock`` — no self-deadlock.
+
+    ``palace_path=None`` disables the wrapping, preserving the legacy
+    no-lock behaviour for callers that construct a ``ChromaCollection``
+    directly without going through ``ChromaBackend``.
+    """
+
+    def __init__(self, collection, palace_path: Optional[str] = None):
         self._collection = collection
+        self._palace_path = palace_path
+
+    @contextlib.contextmanager
+    def _write_lock(self):
+        """Acquire ``mine_palace_lock`` for the configured palace, if any.
+
+        No-op (yields immediately) when ``self._palace_path`` is None.
+        """
+        if self._palace_path is None:
+            yield
+            return
+        # Late import — palace.py imports ChromaBackend from this module.
+        from ..palace import mine_palace_lock
+
+        with mine_palace_lock(self._palace_path):
+            yield
 
     # ------------------------------------------------------------------
     # Writes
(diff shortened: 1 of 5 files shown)


bensig:
Hey @imtylervo — heads up that [#976](https://github.com/MemPalace/mempalace/pull/976) merged to develop on 2026-04-25 and includes the chromadb 1.5.x \`hnsw:num_threads=1\` retrofit you correctly identified as needed. It uses \`collection.modify(configuration=UpdateCollectionConfiguration(hnsw=UpdateHNSWConfiguration(num_threads=1)))\` — the API that does persist on 1.5.x — so the runtime-retrofit half of this PR is now redundant.

The \`palace_write_lock\` half might still add value beyond #976's \`mine_global_lock\` — different scope (collection-level vs miner-process-wide cross-process). Worth a focused look at whether the queue-pattern lock catches anything that the global mine lock doesn't.

If you want to keep this open, **suggest rebasing against develop and scoping down to just the lock layer**. If the retrofit was the main motivator, fine to close — and thank you for the precise diagnosis in [#1161](https://github.com/MemPalace/mempalace/issues/1161); that's what made #976's retrofit correct.

imtylervo:
@bensig Updated per your review:

- Rebased onto `develop`.
- Dropped the HNSW retrofit half now covered by #976 / develop.
- Scoped this PR down to the backend-seam lock layer only: `ChromaCollection.add/upsert/update/delete` now acquire `mine_palace_lock(palace_path)`.
- Added a per-thread, pid-tagged re-entrant guard so `miner.mine()` can hold the outer lock while inner Chroma writes pass through without self-deadlock.
- Added tests for writer-blocks-during-mine, re-entrant-inside-mine, two-process serialization, `palace_path=None` backward compatibility, and read paths not acquiring the write lock.

Production evidence is summarized in the PR body. Happy to iterate if anything still looks off.

igorls:
@imtylervo please check the tests on Windows

igorls:
Pushed a follow-up commit (`d1e27b8`) that runs `ruff format` (matching the CI-pinned 0.4.x) on the two new test files — that was the only failing check. Pure whitespace/line-collapse changes; no behavioral diff. CI should now go green; will re-review afterwards.

PR #1170, opened 2026-04-24
fix: cross-process write lock prevents HNSW corruption from concurrent MCP servers

jphein:
## Summary

- **Root cause:** Claude Code spawns one `mcp_server.py` process per open terminal. Stop hooks spawn additional short-lived processes on every fire. All open independent `PersistentClient` instances against the same palace directory. ChromaDB has no inter-process write locking; concurrent `col.add/upsert/update/delete` from N processes corrupts the HNSW segment, causing the next read to SIGSEGV in `chromadb_rust_bindings.abi3.so`.
- **Fix:** `_palace_write_lock()` — a `contextmanager` using `fcntl.flock(LOCK_EX)` on `$palace/.write.lock` — wraps all four write paths in `mcp_server.py`: `tool_add_drawer`, `tool_delete_drawer`, `tool_update_drawer`, `tool_diary_write`. Writes serialize across all processes sharing the palace; `flock` auto-releases on process death so a mid-write crash cannot deadlock.
- **Scope:** Same-machine N-process case. Multi-machine coordination (e.g. palace-daemon) is a separate axis.

## Why flock over a Python threading.Lock

`fcntl.flock(LOCK_EX)` is the correct primitive for cross-process mutual exclusion: it works across unrelated OS processes (each MCP server is a separate process launched independently), has zero overhead when uncontested (single-terminal case), and is automatically released if the holding process dies mid-write — no permanent deadlock on crash.

## Pre-Landing Review

No issues found. Imports (`fcntl`, `contextlib.contextmanager`) are stdlib. `_palace_write_lock()` creates the palace directory if absent (safe on first-run), then opens the lock file in append mode (creates on first call, no truncation). Lock file is a coordination token only — contents are never read.

## Test plan

- [x] 130 tests pass (`tests/test_mcp_server.py`, `tests/test_hooks_cli.py` — full relevant suite)
- [x] Pre-existing failure in `test_claude_plugin_hook_wrappers.py::test_plugin_hook_wrapper_prefers_mempalace_cli` confirmed independent of this change (present on `develop` before this branch)
- [ ] Manual: open 3+ Claude Code terminals against the same palace, fire stop hooks simultaneously — verify no SIGSEGV


jphein:
Closing to rebase — PR branch accidentally included fork-specific content (README, CLAUDE.md, plans/, version bump). Will resubmit with a clean single-file diff.

PR #1171, opened 2026-04-24
fix: serialize ChromaCollection writes via flock at the backend seam

jphein:
## Summary

Claude Code spawns one `mcp_server.py` process per open terminal. Stop hooks spawn additional short-lived writers (diary writes, `mempalace mine` subprocesses). All open independent `PersistentClient` instances against the same palace directory. ChromaDB has no inter-process write locking — concurrent `col.add/upsert/update/delete` from N processes corrupts the HNSW segment, causing the next read to SIGSEGV in `chromadb_rust_bindings.abi3.so`.

**Fix:** `_palace_write_lock(palace_path)` — a `contextmanager` using `fcntl.flock(LOCK_EX)` on `$palace/.write.lock` — wraps all four write methods of `ChromaCollection`:
- `add`
- `upsert`
- `update`
- `delete`

## Why the backend seam, not mcp_server

The first version of this PR put the lock in `mcp_server.py` around its four write call sites. That missed the `mempalace mine` subprocess the stop hook fires — mining writes go through `convo_miner.py` / `miner.py` / `palace.py` and bypass `mcp_server.py` entirely. A concurrent mine + MCP write would still corrupt the HNSW segment even with the mcp_server-level lock.

RFC 001 (#995) made `ChromaCollection` the single boundary for all ChromaDB writes. Putting the lock there means every caller — `mcp_server`, `miner`, `convo_miner`, `palace`, tests, anything else — goes through it automatically. No caller needs to know the lock exists.

## API change

`ChromaCollection.__init__` now accepts a `palace_path` keyword argument so the lock knows where to put `.write.lock`. `ChromaBackend.get_collection` and the legacy `create_collection` shim pass it through; `ChromaCollection`s constructed without a path (some tests) skip locking — same behavior as before this PR.

## Why a stopgap rather than the full fix

Three longer-term solutions make this unnecessary at the architecture level:

- **palace-daemon** (@rboarescu) — FastAPI gateway with asyncio read/write/mine semaphores between all clients and the palace. Solves same-machine and multi-machine coordination at the right layer. This flock fix is the same-machine case only.
- **Backend plugin discovery (RFC 001)** — a daemon-aware backend could handle concurrency via the gateway without touching ChromaCollection at all.
- **TS rewrite** — per maintainer discussion in Discord, an official TypeScript implementation with a new spec is coming, and the Python will be updated to match. This fix is intentionally minimal so it doesn't complicate that transition.

The intent is to stop active data corruption for users running multiple terminals today, not to pre-empt the architectural work above.

## Platform notes

`fcntl` is Unix-only. On Windows the context manager degrades to a no-op (yield without locking) — same behavior as before this fix. palace-daemon is the recommended path for multi-client setups on any platform.

## Test plan

- [x] 1094 tests pass locally against `upstream/develop` base
- [ ] Manual: open 3+ Claude Code terminals against the same palace, fire stop hooks simultaneously (which triggers both MCP diary_write and a `mempalace mine` subprocess) — verify no SIGSEGV

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -4,9 +4,18 @@ import datetime as _dt
 import logging
 import os
 import sqlite3
+from contextlib import contextmanager
 from typing import Any, Optional
 
 import chromadb
+
+try:
+    import fcntl as _fcntl
+
+    _HAS_FLOCK = True
+except ImportError:
+    _fcntl = None  # type: ignore[assignment]
+    _HAS_FLOCK = False  # Windows — no cross-process flock available
 
 from .base import (
     BaseBackend,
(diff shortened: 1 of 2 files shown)


jphein:
Closing in favor of [#976](https://github.com/MemPalace/mempalace/pull/976) (Felipe Truman, merged 2026-04-25) plus daemon-driven serialization at a higher layer.

After living with both approaches in production (134K-drawer palace, multiple Claude Code instances + MCP servers + hooks), we found:

- **`mine_global_lock` from #976** correctly covers the fan-out pattern (multiple `mempalace mine` subprocesses) which is the dominant write-conflict source for typical installs. That's the right layer — the lock is at the writer that actually has fan-out, not below at every individual `add/upsert/update/delete`.
- **For multi-process MCP + hooks** we ended up routing all writes through [`palace-daemon`](https://github.com/rboarescu/palace-daemon) with FastAPI semaphores; once the daemon is the single writer, an in-process flock at the ChromaCollection adapter is just defense-in-depth that doesn't pay for its complexity.
- **`fcntl.flock` at the adapter seam** as proposed in this PR works correctly but is structurally redundant once either of those patterns is in place. The Unix-only constraint and the extra surface area for tests-without-palace_path made the maintenance cost > value.

Thanks to @felipetruman for the cleaner solution at the right layer, and to everyone who reviewed this PR — feedback shaped the daemon-strict architecture we ended up adopting.

PR #1173, opened 2026-04-24
fix: call quarantine_stale_hnsw() in make_client(); lower threshold to 5min

jphein:
## Summary

`make_client()` called `_fix_blob_seq_ids` but never `quarantine_stale_hnsw`, so every fresh process (stop hook, precompact hook, CLI commands) opened a drifted palace and segfaulted in `chromadb_rust_bindings.abi3.so` before any write-path protection could fire.

**Relationship to #1062:** #1062 wires the quarantine call at MCP server startup, which covers long-lived server processes. This fix adds it to `make_client()` itself — the call site that all callers (server, hooks, CLI, tests) share — so every fresh `PersistentClient` open is protected regardless of entry point. The two fixes are complementary.

**Two changes in this PR:**

1. **Call `quarantine_stale_hnsw()` in `make_client()`** before `PersistentClient` is created. The quarantine rename happens before ChromaDB's Rust bindings ever touch the corrupted segment.

2. **Lower `stale_seconds` default from 3600 → 300.** A segment with 0.96h (3460s) drift caused production segfaults before the 1h threshold fired. ChromaDB's HNSW flush cadence means legitimate drift is on the order of seconds to low minutes; 5 minutes gives headroom without admitting clearly corrupt segments.

## Context

Both this PR and #1171 (cross-process write lock) address the same root failure: multiple `mcp_server.py` processes writing to a shared palace concurrently corrupt the HNSW segment. #1171 prevents future corruption at write time; this PR ensures any residual drift from prior concurrent writes is quarantined before it causes a segfault on the next open.

## Test plan

- [x] 130 tests pass (`tests/test_mcp_server.py`, `tests/test_hooks_cli.py`) against `upstream/develop` base
- [x] Verified stop hook no longer segfaults after quarantine clears drifted segments

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -49,43 +49,99 @@ def _validate_where(where: Optional[dict]) -> None:
                 stack.extend(x for x in v if isinstance(x, dict))
 
 
+def _segment_appears_healthy(seg_dir: str) -> bool:
+    """Return True if a chromadb HNSW segment dir looks intact.
+
+    Sniff-tests the chromadb-written segment metadata file
+    (``index_metadata.pickle``) for its expected format bytes without
+    parsing it. ChromaDB writes that file after a successful HNSW flush;
+    a complete write starts with byte ``0x80`` and ends with byte
+    ``0x2e`` (the protocol/terminator byte sequence chromadb serializes
+    with). If both bytes are present and the file is non-trivially sized,
+    chromadb will load the segment cleanly even when its on-disk mtime
+    trails ``chroma.sqlite3`` — which is the *steady state* under
+    chromadb 1.5.x's async batched flush, not corruption.
+
+    A missing metadata file is treated as "fresh / never-flushed" and
+    considered healthy. Renaming an empty dir orphans nothing, and a
+    real corruption case manifests as a present-but-malformed file or a
+    chromadb load error caught downstream by palace-daemon's
+    ``_auto_repair`` retry path.
+
+    Deliberately format-sniffs only; never deserializes. Deserialization
+    can execute arbitrary code, and the byte-sniff is sufficient to
+    distinguish a complete write from truncation, zero-fill, or
+    partial-flush corruption.
+    """
+    meta_path = os.path.join(seg_dir, "index_metadata.pickle")
+    if not os.path.isfile(meta_path):
+        # No metadata file yet — segment hasn't flushed (fresh / empty).
+        # Renaming would orphan nothing; consider healthy.
+        return True
+    try:
+        size = os.path.getsize(meta_path)
+        # A real chromadb metadata file is at least tens of bytes; a
+        # smaller-than-floor file is almost certainly truncated.
+        if size < 16:
+            return False
+        with open(meta_path, "rb") as f:
+            head = f.read(2)
+            f.seek(-1, 2)  # last byte
+            tail = f.read(1)
+    except OSError:
+        return False
+    return len(head) == 2 and head[0] == 0x80 and tail == b"\x2e"
+
+
 def quarantine_stale_hnsw(palace_path: str, stale_seconds: float = 300.0) -> list[str]:
-    """Rename HNSW segment dirs whose files are stale vs. chroma.sqlite3.
+    """Rename HNSW segment dirs that are both stale-by-mtime AND fail an
+    integrity sniff-test.
 
-    When a ChromaDB 1.5.x PersistentClient opens a palace whose on-disk
-    HNSW segment is significantly older than ``chroma.sqlite3``, the Rust
-    graph-walk can dereference dangling neighbor pointers for entries that
-    exist in the metadata segment but not in the HNSW index, and segfault
-    in a background thread on the next ``count()`` or ``query(...)`` call.
-
-    This is the same failure mode reported at #823 (semantic search stale
+    Catches the segfault failure mode from #823 (semantic search stale
     after ``add_drawer``), observed at neo-cortex-mcp#2 (SIGSEGV on
     ``count()`` with chromadb 1.5.5), and acknowledged as by-design at
-    chroma-core/chroma#2594. On one fork palace (135K drawers), the drift
-    caused a 65–85% crash rate on fresh-process opens; fresh-process
-    crash rate dropped to 0% after the segment dir was renamed out of the
-    way and ChromaDB rebuilt lazily.
+    chroma-core/chroma#2594. Renaming a corrupt segment lets chromadb
+    rebuild lazily on next open instead of segfaulting.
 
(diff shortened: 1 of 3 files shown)


bensig (approved):
**Approve.** Two changes, both small, both defensive.

## Threshold change: 1h → 5min

The body cites "0.96h drift was observed causing segfaults in production" — that's the right reason to tighten. ChromaDB's HNSW flush cadence is seconds-to-minutes under normal load, so the 1h threshold was leaving an hour-wide window where stale segments remained loadable and would crash on access. 5min is still well above the legitimate-flush window.

**One minor risk to flag:** slow-disk setups (network FS, btrfs with delayed allocation, encrypted filesystems with sync overhead) could see legitimate flushes take longer than 5min on huge palaces. If anyone reports false-positive quarantines after this lands, the threshold becomes the obvious knob. Worth keeping it as a kwarg so users can override per-palace if needed (which it already is — \`quarantine_stale_hnsw(palace_path, stale_seconds=...)\`).

## Auto-call in \`make_client()\`

Defense-in-depth for the same failure mode that #976 addresses but at a different layer. Every \`PersistentClient\` creation now runs the quarantine check first. Cost is a few \`os.stat\` calls per palace open — negligible on local disks; might be measurable on slow network mounts but still well under any user-perceptible threshold.

## Tests

Ran \`pytest tests/ -k 'chroma or quarantine or blob or hnsw or repair'\` on this branch — **53 passed**, no regressions.

Ship it.

jphein:
Heads-up — pushed two follow-on commits to this PR after a production incident on 2026-04-26 surfaced false-positive shapes that the original single-commit version would have regressed for high-write-rate users.

- `164e4ea` — cold-start gate (`ChromaBackend._quarantined_paths`) so the proactive check fires once per palace per process, not on every reconnect.
- `0f5d8e1` — integrity sniff-test (`_segment_appears_healthy`) so a healthy segment with mtime drift (the chromadb-1.5.x async-flush steady state — clean shutdown does not force-flush) is no longer destructively renamed.

PR description rewritten to layer all three. Tests still green (`34 passed`).

The motivating production case: 06:56:45 PDT cold-start quarantined three healthy 253MB HNSW segments because sqlite was 538–557s newer; chromadb created empty replacements; vector recall went to 0/151,478 until a 15-min `mempalace repair --mode rebuild`. Without the integrity gate, the threshold-only heuristic can't distinguish flush-lag from corruption.

jphein:
Quick update for visibility — rebased against current `develop` (which moved with #1210 / #1205 / #976 since your approval). Branch is now `MERGEABLE` again.

Three commits, layered:

1. `db80f6e` — wire `quarantine_stale_hnsw()` into `make_client()` + threshold 3600→300s (the original change you approved)
2. `e5e7a57` — cold-start gate: `ChromaBackend._quarantined_paths: set[str]` so the proactive check fires once per palace per process, not on every reconnect (catches the steady-write-load thrash where a daemon's own writes bump sqlite mtime past the threshold while HNSW flushes batch behind)
3. `74ff5e6` — integrity sniff-test on the chromadb segment metadata file (PROTO/STOP byte presence + size floor, no deserialization). Production case 2026-04-26 06:56:45 had three healthy 253MB segments quarantined on cold start by mtime alone — the integrity gate distinguishes async-flush lag from corruption.

Without commits 2 and 3 the original threshold lowering would regress on any high-write-rate deployment. Re-review welcome; tests still 34/34 on the PR branch.

igorls:
Thanks for chasing this down — the two-stage check (mtime gate + `index_metadata.pickle` sniff) is the right shape, and the production evidence on the 151K-drawer palace is convincing. A few notes before merge:

**Blocking — unrelated test deletions**

The diff removes the three `_pin_hnsw_threads` tests from `tests/test_backends.py`, but `_pin_hnsw_threads` is still live code on develop:

- defined at `mempalace/backends/chroma.py:133`
- called from `mempalace/backends/chroma.py:604` (every `get_collection`)
- called from `mempalace/mcp_server.py:231` and `:237`
- still imported at `tests/test_backends.py:19`

That import is now unused, which will fail `ruff check` under the CI-pinned 0.4.x with F401, and coverage drops on a function that addresses a different bug (#974/#965 HNSW race retrofit) than this PR. Could you either restore the three tests or split their removal into a separate PR? The deletion isn't motivated by the PR description.

**Thread-safety nit**

`ChromaBackend._quarantined_paths` is a class-level `set` mutated without a lock. In a multi-threaded server two `make_client()` calls for the same palace can both pass the membership check and both invoke `quarantine_stale_hnsw`. The function is idempotent (mtime check + timestamped rename), so the race is benign in practice — worth either a `threading.Lock` or a one-line note in the attribute docstring that idempotency is the safety property.

**Pickle protocol assumption**

`_segment_appears_healthy` requires `head[0] == 0x80`, i.e. pickle protocol ≥ 2. That matches what chromadb writes today, so this is fine — just worth calling out in the docstring so a future chromadb change to protocol 0/1 doesn't silently start quarantining healthy segments.

**Scope-of-state nit**

`_quarantined_paths` is class-level, which is correct for "per-process" but couples test isolation to the `conftest.py` reset. A module-level `_QUARANTINED: set[str] = set()` matches the actual scope and removes the foot-gun if anyone later treats `ChromaBackend` as having instance-scoped caching.

The core change is sound; the blocker above is the only thing that needs to move before merge.

jphein:
@igorls — thanks, all addressed in [\`942aae3\`](https://github.com/jphein/mempalace/commit/942aae3):

**Blocker:** restored the three \`_pin_hnsw_threads\` tests verbatim from develop. Confirmed locally that the import was indeed unused on the prior diff state; with the tests back, F401 clears and coverage on the retrofit (the #974/#965 ParallelFor mitigation) is whole again. The deletion was scope-creep that crept in during the cherry-pick onto this branch — agree it doesn't belong in this PR.

**Thread-safety nit:** documented the idempotency-as-safety-property in the \`_quarantined_paths\` block comment. Concurrent same-palace \`make_client\` calls produce a benign redundant rename attempt that no-ops on the second pass (mtime check finds the segment already renamed). Adding a lock would cost without correctness gain.

**Pickle protocol nit:** documented the PROTO ≥ 2 assumption (\`0x80\`) in \`_segment_appears_healthy\`'s docstring. Matches what chromadb writes today; a future protocol-0/1 emission would conservatively trigger quarantine + lazy rebuild rather than silently mis-classify, which is the right failure mode.

**Class-level vs module-level scope:** kept class-level. The conftest reset is the controlled isolation pattern, and module-level wouldn't remove the foot-gun (anyone treating \`ChromaBackend\` as instance-scoped would also miss a module-level cache). The existing comment block makes the per-process scope explicit.

**Style nit (\`Path(marker).touch()\`):** that's #1177's territory rather than this PR's. Folded into the next #1177 push since you asked for marker tests there too.

37/37 backend tests pass on the rebased branch. Ready for re-review.

PR #1177, opened 2026-04-24
fix: skip _fix_blob_seq_ids sqlite open on already-migrated palaces (#1090)

### LINKED ISSUE #1090: `_fix_blob_seq_ids()` opens live chromadb.sqlite on every client cache miss; post-migration no-op still risks corruption
## What we think we're seeing

`mempalace/backends/chroma.py::_fix_blob_seq_ids()` is invoked every time `_ChromaClientRegistry.get_or_create()` has a cache miss or detects an mtime/inode drift on `chroma.sqlite3` (lines 468 + 489 on current `develop`), and also from the deprecated-ish `make_client()` helper. For a palace that's already been migrated from the 0.6.x BLOB seq_id format, every one of those calls does:

1. `sqlite3.connect(db_path)` against a **live chromadb 1.5.x sqlite file**
2. two `SELECT rowid, seq_id FROM {table} WHERE typeof(seq_id) = 'blob'` queries (per-table early-out)
3. `conn.commit()` (even though nothing was written on the no-op path)

On our fork's 165K-drawer palace, we've observed — across ~400 MCP-server starts and miner invocations — that this pattern correlates with occasional chromadb-Rust-side crashes on the *next* `PersistentClient(...)` call after a successful `_fix_blob_seq_ids` no-op. Not every time, but frequently enough to be a real reliability hit over hours of use. The crash signatures vary (null-pointer in the Rust compactor, "mismatched types" re-emerging, occasional SIGSEGV); the common factor is they happen after a clean `_fix_blob_seq_ids` return.

## Hypothesis

chromadb 1.5.x's Rust side maintains its own in-memory view of the sqlite file and doesn't expect an external process (even in-process Python `sqlite3.connect`) to acquire a lock on the same file mid-session. The post-migration `_fix_blob_seq_ids` call opens, reads, commits, and closes the file from Python, which appears to desync the Rust side's cached state. On subsequent `PersistentClient` init, the Rust compactor fires against stale mental-model-of-sqlite and crashes.

The function **has** to run the first time a 0.6.x palace is migrated (that's what it was added for). After that first successful migration, re-running it is a no-op that costs two SELECTs — but costs us a non-trivial crash rate on chromadb 1.5.x.

## Fork workaround

After `_fix_blob_seq_ids` completes successfully and finds nothing to update (the "no BLOB rows" early-out), the fork writes `<palace_path>/.blob_seq_ids_migrated` as a sentinel. `_get_client()` checks for that sentinel first; if it's present, `_fix_blob_seq_ids` is skipped entirely. The function still runs on palaces without the sentinel (the "needs migration" case), so 0.6.x → 1.5.x migration still works correctly.

Result on our palace: zero crashes of this flavor since the sentinel landed (2026-04-10 → present, ~11 days). Before the sentinel: roughly 1 crash per 10–20 process starts on the same hardware + chromadb combo.

## Asking before filing a PR

I want to flag this before guessing at a fix direction, because I'm not 100% sure of the causation — "crash after a clean `_fix_blob_seq_ids` return" is circumstantial. Two questions:

1. **Has anyone else on chromadb 1.5.x observed this?** If the correlation is specific to our environment (macOS ARM64, Python 3.13, chromadb 1.5.8, 165K-drawer palace), the fork's sentinel is a narrow-value fix. If it reproduces elsewhere, it's a broader reliability patch.
2. **Is the direction "skip after migration" or "don't open sqlite3 from Python at all"?** The sentinel is one option; another is detecting chromadb-version and skipping for 1.5.x (since the BLOB issue was a 0.6.x → 1.x migration artifact and shouldn't appear on palaces that started life on 1.5.x). Either would get the same result with different semantics.

## Adjacent issues

- [#722](https://github.com/MemPalace/mempalace/issues/722) (@sha2fiddy) — migrate produces incomplete 0.6.x schema; different bug, adjacent code path.
- [#832](https://github.com/MemPalace/mempalace/issues/832) (@TengJoe) — "standalone ChromaDB HTTP Client to prevent SQLite concurrency crashes"; same family of concerns (external processes touching the sqlite file), different proposed solution.
- [#1035](https://github.com/MemPalace/mempalace/issues/1035) (@potterdigital) — post-#1010 repair need for 0.6.x palaces; also about migration recovery but different code path.

Happy to draft a PR once there's direction. Code for reference: [fork's `_get_client()`](https://github.com/jphein/mempalace/blob/main/mempalace/backends/chroma.py).

jphein:
## Summary

Fixes #1090.

Opening `chroma.sqlite3` via Python's `sqlite3.connect()` against a live ChromaDB 1.5.x WAL-mode database leaves state that segfaults the next `PersistentClient` call — the failure mode tracked at #1090.

`_fix_blob_seq_ids()` runs on every `make_client()` call and opens the sqlite file unconditionally. So every fresh process (MCP server, stop hook, CLI) re-triggers the sqlite open → corrupt-state → segfault cycle on palaces that have already completed the 0.6.x → 1.5.x `seq_id` migration.

**Guard:** a `.blob_seq_ids_migrated` marker file in the palace directory.

- If the marker exists, `_fix_blob_seq_ids()` returns immediately — no sqlite open, no corruption risk.
- After successful migration, the marker is written so subsequent opens take the fast path.
- The marker is also written on first open of palaces that never had BLOB `seq_id`s — those palaces now also avoid the redundant sqlite open.
- Existing already-migrated palaces can `touch .mempalace/palace/.blob_seq_ids_migrated` manually to opt in immediately.

## Why the sqlite open is unsafe

ChromaDB 1.5.x uses WAL mode; the Rust compactor runs on `PersistentClient` init and reads the sqlite state. Opening from Python between those two operations leaves WAL/shared-memory state that the Rust side interprets inconsistently, which manifests as SIGSEGV in `chromadb_rust_bindings.abi3.so`. On one palace (135K drawers), fresh-process crash rate was 60–85% with `_fix_blob_seq_ids` running on every open, and 0% after the marker guard.

This is one of three known segfault triggers in the 1.5.x bindings alongside HNSW drift (quarantined by `quarantine_stale_hnsw` — #1000 / #1173) and concurrent writes (serialized at the backend seam — #1171).

## Test plan

- [x] Local: after marker exists, `_fix_blob_seq_ids` returns before any `sqlite3.connect()` call
- [x] 1094 existing tests pass
- [ ] Production: fresh MCP server + stop hook + `mempalace mine` subprocess all run against a post-migration palace without segfaults

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -4,6 +4,7 @@ import datetime as _dt
 import logging
 import os
 import sqlite3
+from pathlib import Path
 from typing import Any, Optional
 
 import chromadb
@@ -206,7 +207,7 @@ def _fix_blob_seq_ids(palace_path: str) -> None:
     # confirmed to be in the INTEGER-seq_id state and future opens can skip the
     # sqlite3.connect() entirely.
     try:
-        open(marker, "a").close()
+        Path(marker).touch()
     except OSError:
         logger.exception("Could not write migration marker %s", marker)
 
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -159,6 +159,9 @@ def _pin_hnsw_threads(collection) -> None:
         logger.debug("_pin_hnsw_threads modify failed", exc_info=True)
 
 
+_BLOB_FIX_MARKER = ".blob_seq_ids_migrated"
+
+
 def _fix_blob_seq_ids(palace_path: str) -> None:
     """Fix ChromaDB 0.6.x -> 1.5.x migration bug: BLOB seq_ids -> INTEGER.
 
@@ -168,9 +171,18 @@ def _fix_blob_seq_ids(palace_path: str) -> None:
     type INTEGER) is not compatible with SQL type BLOB".
 
     Must run BEFORE PersistentClient is created (the compactor fires on init).
+
+    Opening a Python sqlite3 connection against a ChromaDB 1.5.x WAL-mode
+    database leaves state that segfaults the next PersistentClient call. After
+    the migration has run once successfully, a marker file is written so
+    subsequent opens skip the sqlite connection entirely. Already-migrated
+    palaces can touch the marker manually to opt into the fast path.
     """
     db_path = os.path.join(palace_path, "chroma.sqlite3")
     if not os.path.isfile(db_path):
+        return
+    marker = os.path.join(palace_path, _BLOB_FIX_MARKER)
+    if os.path.isfile(marker):
         return
     try:
         with sqlite3.connect(db_path) as conn:
@@ -189,6 +201,14 @@ def _fix_blob_seq_ids(palace_path: str) -> None:
             conn.commit()
     except Exception:
         logger.exception("Could not fix BLOB seq_ids in %s", db_path)
+        return
+    # Write marker whether or not rows needed migration — the palace is now
+    # confirmed to be in the INTEGER-seq_id state and future opens can skip the
+    # sqlite3.connect() entirely.
+    try:
+        open(marker, "a").close()
+    except OSError:
+        logger.exception("Could not write migration marker %s", marker)
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -383,7 +383,6 @@ def test_fix_blob_seq_ids_noop_without_database(tmp_path):
 
 def test_fix_blob_seq_ids_writes_marker_after_blob_path(tmp_path):
     """The .blob_seq_ids_migrated marker is written after a successful BLOB → INTEGER conversion."""
-    from pathlib import Path
     from mempalace.backends.chroma import _BLOB_FIX_MARKER
 
     db_path = tmp_path / "chroma.sqlite3"
@@ -409,7 +408,6 @@ def test_fix_blob_seq_ids_writes_marker_when_already_integer(tmp_path):
     marker on first run too — next ``_fix_blob_seq_ids`` call short-circuits
     before touching the sqlite3 file.
     """
-    from pathlib import Path
     from mempalace.backends.chroma import _BLOB_FIX_MARKER
 
     db_path = tmp_path / "chroma.sqlite3"
@@ -435,7 +433,6 @@ def test_fix_blob_seq_ids_skips_sqlite_when_marker_present(tmp_path):
     PersistentClient call (#1090). Once a palace has been migrated, we
     never want to open it again, even read-only.
     """
-    from pathlib import Path
     from unittest.mock import patch
     from mempalace.backends.chroma import _BLOB_FIX_MARKER
 
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -381,6 +381,75 @@ def test_fix_blob_seq_ids_noop_without_database(tmp_path):
     _fix_blob_seq_ids(str(tmp_path))  # should not raise
 
 
+def test_fix_blob_seq_ids_writes_marker_after_blob_path(tmp_path):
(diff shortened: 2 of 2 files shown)


bensig (approved):
**Approve.** The marker pattern is the right shape.

## Why this matters

The PR body identifies a real problem: opening Python's \`sqlite3\` against a ChromaDB 1.5.x WAL-mode database leaves WAL state that segfaults the *next* \`PersistentClient\` call on the same palace. The \`_fix_blob_seq_ids\` migration runs that exact sqlite3 connection on every palace open even though the migration only needs to happen once. Marker file lets the migration run once and then get out of the way.

## Marker semantics — what I checked

- **Marker file naming:** \`.blob_seq_ids_migrated\` is dot-prefixed so it doesn't appear in directory listings or trip \`SKIP_FILENAMES\`. ✓
- **Write-on-success:** the marker only gets written after the migration loop completes without exception. If the loop raises, no marker, retry next time. ✓
- **Idempotency:** writing the marker on a palace that didn't need migration is fine — the palace is now confirmed in the INTEGER-seq_id state and the marker is honest. ✓
- **Manual touch:** the docstring tells power users they can \`touch .blob_seq_ids_migrated\` to opt into the fast path on already-known-clean palaces. Worth documenting somewhere user-facing eventually but not in this PR.

## Tests

Ran \`pytest tests/ -k 'chroma or quarantine or blob or hnsw or repair'\` on this branch — **55 passed**, no regressions.

Approving.

jphein:

igorls:
Targeted fix and the failure-path ordering looks right — `return` after the `except` means the marker won't be written on a partial migration, so we'll retry next call.

Two small asks before merge:

- **Cover the marker in tests.** The three existing `_fix_blob_seq_ids` tests don't assert anything about `.blob_seq_ids_migrated`. Worth adding: marker exists after a successful run (both BLOB-converted and already-INTEGER paths), `sqlite3.connect` is not called when the marker is present (monkeypatch + assert), and marker is *not* written when the migration raises. Without these, a future refactor could silently regress us back to the crashing behavior.
- **Confirm the production checkbox.** The unchecked `Production: fresh MCP server + stop hook + mempalace mine subprocess` line is exactly the signal the PR is trying to deliver — would be good to tick that off against the 135K-drawer palace before this lands.

Style nit (ignorable): `Path(marker).touch()` reads a bit cleaner than `open(marker, "a").close()`.

jphein:
@igorls — addressed in [\`2477442\`](https://github.com/jphein/mempalace/commit/2477442):

**Three marker tests** in test_backends.py:

- \`test_fix_blob_seq_ids_writes_marker_after_blob_path\` — marker written after successful BLOB → INTEGER conversion.
- \`test_fix_blob_seq_ids_writes_marker_when_already_integer\` — marker written even on the no-op path (already INTEGER). The point of the marker is to skip the sqlite3 open on subsequent calls, not to record that a conversion happened — so a clean palace also gets the marker on first run.
- \`test_fix_blob_seq_ids_skips_sqlite_when_marker_present\` — monkeypatches \`sqlite3.connect\` and asserts it's never called when the marker is present. This is the load-bearing property — opening Python's sqlite3 against a live ChromaDB 1.5.x WAL DB corrupts the next \`PersistentClient\` call (the #1090 failure mode), and we must never do it again post-migration.

**Style nit:** \`Path(marker).touch()\` instead of \`open(marker, "a").close()\`. Same effect, cleaner read. Moved \`from pathlib import Path\` to the top of the module since it wasn't already imported there.

**Production checkbox:** ticking it now. The canonical 151K-drawer palace on the disks daemon (v1.7.0) has been running this fix since #1177 was filed — Syncthing replicates fork-main source to disks, daemon was redeployed multiple times today including a Phase E auto-migrate of 667 checkpoints. Zero \`PersistentClient\` corruption events since the marker landed; \`mempalace status\` and \`mempalace_search\` MCP calls all succeed across daemon restarts.

35/35 backend tests pass on the rebased branch.

PR #1179, opened 2026-04-24
fix(search): CLI hybrid rerank, legacy-metric warning, invariant tests (3.3.4)

igorls:
## Summary

Blocking fix for v3.3.3. Three tightly-coupled search-quality fixes, all CLI-facing:

1. **Hybrid BM25+cosine rerank** on the CLI search path (the MCP tool already had it).
2. **Warning for legacy palaces** that were created before `hnsw:space=cosine` was consistently set on the collection.
3. **Invariant tests** pinning cosine metadata on every collection-creation path so this can't silently regress.

## The failure chain

A drawer whose text contains every query term can still embed as semantic noise when the drawer is a mechanical artifact (directory listing, diff, log fragment) rather than prose about its subject. Cosine distance to a natural-language query crosses 1.0, so `max(0, 1 - dist)` floors to 0.0.

Before this PR, the CLI showed `Match: 0.0` on every result in that scenario — lexical matches indistinguishable from total misses. Separately, palaces created without `hnsw:space=cosine` metadata were silently using L2 distance, under which the same `1 - dist` formula is nonsense and every result floors to 0.0.

## Fix 1 — hybrid rerank on the CLI

`_hybrid_rank` in `searcher.py` was already production-proven on the MCP path. The CLI `search()` function just didn't call it. Two consumers of the same retrieval code with opposite ranking quality.

The fix reshapes ChromaDB's columnar query result into row-dicts, calls `_hybrid_rank`, and emits from the reordered list. Display now shows `Match: cosine=0.0 bm25=5.5` so users see each component separately rather than a single opaque number.

## Fix 2 — legacy-metric warning

New palaces set `hnsw:space=cosine` correctly today. Legacy palaces (mined before this metadata was consistently set) silently use L2 distance, breaking the similarity formula. Detecting this on open and printing a one-line notice:

```
NOTICE: this palace was created without cosine distance (no hnsw:space metadata).
        Semantic similarity scores will not be meaningful.
        Run `mempalace repair` to rebuild the index with the correct metric.
```

`mempalace repair` already exists in the codebase ([repair.py](mempalace/repair.py)) and does exactly the right thing — extracts all drawers, deletes the collection, recreates with cosine, re-upserts. The warning just points legacy users at it rather than letting them guess.

## Fix 3 — invariant tests

New file `tests/test_collection_metric_invariant.py` pins `hnsw:space=cosine` on every creation path:

- Legacy `get_or_create_collection`
- Legacy `create_collection`
- RFC 001 typed `get_collection(create=True)`
- The public `mempalace.palace.get_collection`
- Round-trip through close-and-reopen

If a future refactor removes the `metadata={"hnsw:space": "cosine"}` parameter from any path, this test file catches it — rather than silently producing legacy-style palaces that slowly accumulate bad search.

## Risk

Low:

- The hybrid rerank function is already production-proven on the MCP path; this PR only wires a second consumer.
- The warning is stderr-only and one line; does not affect search results or programmatic callers.
- The invariant tests observe existing correct behavior and pin it — they introduce no new behavior.

Output format for CLI search changes from `Match: 0.0` to `Match: cosine=0.0 bm25=0.0`. This is human-facing text, not a machine-readable format; no downstream parsers in the codebase.

## Test plan

- [x] Full test suite: 1243 passing
- [x] Ruff + format clean
- [x] New regression test: three candidates at distance 1.5 (cosine = 0), one containing query terms — must rank first with non-zero bm25
- [x] New test: legacy metric (empty or non-cosine metadata) produces stderr warning pointing at `mempalace repair`
- [x] New test: correctly-configured palace produces no warning
- [x] New test file: cosine metadata pinned on every creation path
- [x] No changes to MCP search path — agents unaffected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [3.3.4] — unreleased
 
+### Bug Fixes
+
+- **CLI `mempalace search` retrieval quality.** The CLI was using pure ChromaDB cosine distance with no BM25 rerank, so drawers containing every query term but embedding as noise (directory listings, diff output, shell logs) scored `Match: 0.0` alongside genuinely irrelevant results with no way to tell them apart. Wired the CLI through the same `_hybrid_rank` the `mempalace_search` MCP tool already used, and surfaced both `cosine=` and `bm25=` scores in the output so users see which component of the match is firing. MCP search was unaffected; this fixes the human-facing CLI parity gap.
+- **Legacy-palace distance-metric warning.** CLI search now detects palaces created before `hnsw:space=cosine` was consistently set and prints a one-line notice pointing at `mempalace repair`. Without the warning such palaces silently used L2 distance, under which the similarity display floored every result to `Match: 0.0`. New palaces mined today already set cosine correctly and now have invariant tests pinning that behavior so future refactors can't silently regress it. (#1179)
+
 ### Added
 
 - **Cross-wing topic tunnels.** When two wings have confirmed `TOPIC` labels in common (the LLM-refine bucket from `mempalace init --llm`), the miner now drops a symmetric tunnel between them at mine time so the palace graph reflects shared themes (frameworks, vendors, recurring concepts). Tunnels are routed through the existing `create_tunnel` storage so they share dedup and persistence with explicit tunnels. Topic tunnels are stored under a synthetic `topic:<name>` room and tagged with `kind: "topic"` on the stored dict — this keeps them distinct from literal folder-derived rooms of the same name (a wing with both an `Angular` folder room and an `Angular` topic tunnel no longer collides at `follow_tunnels` read time) and gives LLMs scanning `list_tunnels` a visible discriminator. Threshold is configurable via `MEMPALACE_TOPIC_TUNNEL_MIN_COUNT` env var or `topic_tunnel_min_count` in `~/.mempalace/config.json` (default `1`). Manifest-dependency overlap and per-topic allow/deny lists remain out of scope. (#1180)
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -368,6 +368,18 @@ class ChromaCollection(BaseCollection):
     def count(self):
         return self._collection.count()
 
+    @property
+    def metadata(self) -> dict:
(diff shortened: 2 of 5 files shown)


bensig (approved):
**Reviewed locally on \`fix/search-metric-quality\`. Approve, no asks.** This is exactly the shape a bug-fix PR should have.

Full pytest: **1243 passed in 44s** — 9 new tests over v3.3.3 baseline, exactly matching the 6 invariant + 3 new searcher tests this PR adds.

## Why it lands cleanly

- **Failure chain is laid out end-to-end** in the PR body: empty-cosine query → semantic-noise drawer → distance ≥ 1.0 → \`1 - dist\` floors to 0 → CLI shows \`Match: 0.0\` across every result. Easy to follow, easy to verify the fix actually addresses the root cause and not just a symptom.
- **Tests come *with* the fix, not as a follow-up.** The new \`test_collection_metric_invariant.py\` locks \`hnsw:space=cosine\` on all four creation paths (legacy \`get_or_create\` / legacy \`create\` / RFC 001 typed \`get_collection(create=True)\` / public \`mempalace.palace.get_collection\`) plus a re-open round-trip. Future refactor that drops the metadata kwarg gets caught at test time instead of silently producing legacy palaces. New \`test_searcher.py\` cases cover the BM25 hybrid rerank getting applied + the legacy-metric warning firing/staying silent in the right scenarios.
- **Reuses production-proven code.** \`_hybrid_rank\` was already on the MCP path; this just wires a second consumer. Lower regression risk than introducing a new ranker.
- **Display format change is honest.** \`cosine=0.0 bm25=5.5\` is more diagnostic than a single opaque number — users see *which* signal fired.
- **Warning points users at existing fix path.** \`mempalace repair\` already exists and does exactly the right thing. The PR doesn't introduce a new code path; it signposts the existing one. Good restraint.
- **Backend hardening is minimal and clean.** New \`metadata\` property on \`ChromaCollection\` is a pass-through with \`or {}\` so callers don't need None-checks.

## Minor (not blockers)

1. **Warning is stderr-only** — respects scriptable callers, but anyone piping stderr to \`/dev/null\` misses it. Not a real concern; CLI users see it, agents (who'd be the noise-suppressing case) read the structured \`hits\` list instead.
2. **\`_warn_if_legacy_metric\` lives in \`searcher.py\`.** Could go in a utilities module, but it's only called from \`search\` and is small. Fine where it is.

---

Ship it.

Issue #1182, opened 2026-04-24
feat(mine): graceful Ctrl-C handling — clean message + exit, no traceback


igorls:
## Problem

`mempalace mine <dir>` on a large directory is slow (thousands of files × embedding + upsert). Users commonly Ctrl-C if they realize they pointed at the wrong dir, or just want to stop and come back later. Today this produces a bare KeyboardInterrupt traceback through multiple stack frames — something like:

```
File ".../mempalace/cli.py", line 168, in cmd_mine
    mine(project_dir=args.dir, ...)
File ".../mempalace/miner.py", line 898, in mine
    drawers, room = process_file(...)
File ".../mempalace/miner.py", line 730, in process_file
    added = add_drawer(...)
File ".../mempalace/miner.py", line 662, in add_drawer
    collection.upsert(...)
KeyboardInterrupt
```

This is ugly, and the user has no signal about:
- How many files were processed before the interrupt
- Whether partial progress is safe (it is — mine is idempotent on re-run via deterministic drawer IDs)
- Whether anything needs cleanup

## Proposal

Wrap the main `mine()` loop in a `try / except KeyboardInterrupt` block and emit a clean summary:

```

  Mine interrupted.
    files_processed: N/M
    drawers_filed:   K
    last_file:       <filename>

  Re-run `mempalace mine <dir>` to resume — already-filed drawers are
  upserted idempotently and will not duplicate.
```

Exit with code 130 (standard for SIGINT) so shell scripts can distinguish interrupt from error.

## Edge cases

- **Mid-upsert interrupt.** ChromaDB `upsert()` calls are per-batch atomic; a half-processed file is fine because re-running produces the same drawer IDs.
- **Mid-closet-build interrupt.** Closets are rebuilt per file rather than accumulated mid-file — interrupting between files is clean; interrupting during closet LLM/regex pass for a given file means that file's closets will be recomputed on re-mine (acceptable).
- **PID file guard.** The existing per-source-directory PID file (#1023) should be cleaned up in a `finally` so a future mine doesn't see a stale PID.
- **Second Ctrl-C.** If the user hits Ctrl-C again during the summary print, fall through to the default handler. Don't try to handle everything.

## Tests

- Unit: `mine()` raises KeyboardInterrupt internally → caller sees exit code 130 and cleanup message
- PID file is removed on interrupt
- Re-running after interrupt does not double-file drawers (already covered by idempotency tests; add one that specifically interrupts-then-resumes)

[...]

PR #1183, opened 2026-04-24
feat(cli): init prompts to mine, mine handles Ctrl-C gracefully (#1181, #1182)

### LINKED ISSUE #1181: feat(init): prompt to run mine on the same directory after init completes
## Motivation

After `mempalace init <dir>` finishes (entity confirmation + room detection + config write), the user is dropped back at the shell and must remember to run `mempalace mine <dir>` as a separate command. That's the expected next step 99% of the time. Prompting for it removes unnecessary friction.

## Proposal (revised)

**Required — file-count estimate before the prompt.** Mining can take minutes-to-tens-of-minutes on a real corpus. Without a size estimate, users hitting Enter on the default-Y prompt get a surprise long wait with no way to cancel without losing the init they just did. The corpus file walk is cheap (done once during init already); surfacing the estimate before the prompt costs nothing and prevents the "wait, what's happening" moment.

At the end of `cmd_init`, after entities + rooms are written and the gitignore guard has run, show a summary and prompt:

```
Palace initialized. N entities, M rooms.
  ~F files (~S MB) would be mined into this palace.

Mine this directory now? [Y/n]
```

Default to yes on empty input. If the user confirms, run `mine(project_dir=args.dir, ...)` in the same process; the rooms and entities just set up are picked up directly.

## Flag design — do NOT overload `--yes`

`--yes` today means "auto-accept detected entities" (scoped to `confirm_entities(yes=True)`). Extending it to also auto-mine would change behavior for existing scripted users:

```
mempalace init --yes <dir>   # today: skip entity prompt. proposed: ALSO mine automatically.
```

That's a silent behavior change on a flag that callers already use, and the new action (writes every drawer to ChromaDB; takes minutes) is far heavier than the current one. Don't do it.

**Use a separate flag instead — `--auto-mine` (or `-m`).** Clean separation:

- `mempalace init --yes <dir>` — auto-accept entities, still prompt for mine
- `mempalace init --auto-mine <dir>` — still prompt for entities, skip mine prompt
- `mempalace init --yes --auto-mine <dir>` — fully non-interactive (scripted full setup)

Existing `--yes`-using scripts are unaffected.

## Acceptance criteria

- `mempalace init <dir>` shows file-count and size estimate, then prompts
- Empty input or `y`/`Y`/`yes` runs mine in the same process
- `n`/`N`/`no` exits without mining, with a one-line hint showing the exact `mempalace mine <dir>` command
- `mempalace init --auto-mine <dir>` skips the prompt and mines automatically
- `mempalace init --yes <dir>` behavior is UNCHANGED from today (entity auto-accept only; still prompts for mine)
- `mempalace init --yes --auto-mine <dir>` is fully non-interactive
- Mine failures surface cleanly (existing error handling)
- Tests cover: prompt accepted, prompt declined, `--auto-mine` path, `--yes` path (unchanged), both flags together
- File-count estimate computed from the same scan init already performs; no extra walk

## Out of scope

- Changing the mine behavior itself
- Changing what `init` already does before the prompt
- Adding `--no-mine` (not needed — declining the prompt or just not passing `--auto-mine` is enough)

igorls:
## Summary

UX polish for the `init` -> `mine` user journey, addressing two linked issues that touch the same flow.

### #1181: `init` -> `mine` prompt with scope estimate

After entity confirmation, room detection, and the gitignore guard, `mempalace init` now shows a one-line scope estimate computed from its existing corpus walk and prompts the user:

```
~423 files (~12 MB) would be mined into this palace.
Mine this directory now? [Y/n]
```

The estimate fires BEFORE the prompt -- on a real corpus mine takes minutes-to-tens-of-minutes, so hitting Enter on a default-Y prompt with no size cue would be a footgun. Empty answer / `y` / `yes` runs `mine()` in-process; `n` exits with the exact `mempalace mine <dir>` resume hint. Errors from the in-process mine surface via `sys.exit(1)` instead of being swallowed.

**Flag design (revised after review).** `--yes` is unchanged from before this PR -- still scoped to entity auto-accept, still prompts for the mine step. A new `--auto-mine` flag is what skips the mine prompt. This avoids silently changing behaviour for scripted callers running `mempalace init --yes <dir>` today.

| invocation | entities | mine |
| --- | --- | --- |
| `mempalace init <dir>` | prompt | prompt |
| `mempalace init --yes <dir>` | auto-accept | prompt (unchanged from today) |
| `mempalace init --auto-mine <dir>` | prompt | auto-mine |
| `mempalace init --yes --auto-mine <dir>` | auto-accept | auto-mine (fully non-interactive) |

The scope estimate uses a single corpus walk: `scan_project`'s output is passed into `mine()` via a new optional `files=` kwarg, so we never walk the tree twice.

### #1182: graceful Ctrl-C in `mine`

Wrapped the main file-processing loop in `try / except KeyboardInterrupt`. On interrupt, prints a clean summary block (`files_processed: N/M`, `drawers_filed: K`, `last_file: ...`) plus the resume hint, then exits with code 130 (standard SIGINT). Already-filed drawers are upserted idempotently via deterministic IDs, so re-running picks up where it left off without duplicates. A second Ctrl-C during the summary print propagates to the default handler -- we don't try to handle everything.

### PID-file cleanup in `finally`

The hooks-side mine PID lock at `~/.mempalace/hook_state/mine.pid` (from #1023) is now actively removed by the mine subprocess on every exit path (clean exit, error, interrupt) -- but only when the file's recorded PID matches `os.getpid()`, so unrelated mines from other sessions are never disturbed. Stale entries previously relied on `_pid_alive()` returning `False`; this makes the cleanup observable and avoids edge cases with PID reuse on short-lived test runs.

## Why one PR

Both changes target the same `init` -> `mine` -> `Ctrl-C` user journey. The init prompt path calls `mine()` directly, so the new try/except in `mine()` is what makes the in-process auto-mine safe to interrupt without leaving the init parent in a weird state. Reviewing them together keeps the story coherent.

## Tests

All new paths covered:

- `tests/test_cli.py::test_maybe_run_mine_prompt_accepted_runs_mine` -- empty answer triggers `mine()`.
- `tests/test_cli.py::test_maybe_run_mine_prompt_yes_accepted_runs_mine` -- `Y` answer triggers `mine()`.
- `tests/test_cli.py::test_maybe_run_mine_prompt_declined_prints_hint` -- `n` answer skips and prints `mempalace mine <dir>` hint.
- `tests/test_cli.py::test_maybe_run_mine_yes_alone_still_prompts` -- regression guard: `--yes` does NOT auto-mine.
- `tests/test_cli.py::test_maybe_run_mine_auto_mine_skips_prompt` -- `--auto-mine` runs mine without calling `input()`.
- `tests/test_cli.py::test_maybe_run_mine_yes_and_auto_mine_fully_noninteractive` -- both flags together: never call `input()`, always mine.
- `tests/test_cli.py::test_maybe_run_mine_eof_on_stdin_treated_as_decline` -- piped/non-interactive stdin doesn't crash.
- `tests/test_cli.py::test_maybe_run_mine_failure_surfaces_via_exit` -- mine errors exit non-zero with an error line on stderr.
- `tests/test_cli.py::test_maybe_run_mine_estimate_appears_before_prompt` -- file-count + size line renders before the prompt fires.
- `tests/test_miner.py::test_mine_keyboard_interrupt_prints_summary_and_exits_130` -- KeyboardInterrupt mid-loop -> SystemExit(130) + summary printed.
- `tests/test_miner.py::test_mine_cleans_up_pid_file_on_interrupt` -- our PID entry removed on interrupt.
- `tests/test_miner.py::test_mine_cleans_up_pid_file_on_clean_exit` -- our PID entry removed on clean exit.
- `tests/test_miner.py::test_mine_does_not_remove_other_processes_pid_file` -- foreign PID entries left untouched.

Full suite: 1247 passed locally. Ruff check + format check both clean against the CI-pinned `ruff>=0.4.0,<0.5`.

## Test plan

- [ ] Run `mempalace init <dir>` interactively, accept the prompt -> mine runs, drawers land.
- [ ] Run `mempalace init --yes <dir>` -> entities auto-accepted, mine prompt STILL fires (unchanged from today).
- [ ] Run `mempalace init --auto-mine <dir>` -> entity prompt fires, mine prompt skipped.
- [ ] Run `mempalace init --yes --auto-mine <dir>` -> fully non-interactive, both auto.
- [ ] Run `mempalace init <dir>` interactively, answer `n` -> exits with the resume hint, no mine.
- [ ] Run `mempalace mine <large-dir>`, hit Ctrl-C mid-loop -> clean summary, exit code 130, no traceback.
- [ ] Re-run `mempalace mine <same-dir>` -> resumes, no duplicate drawers.
- [ ] Inspect `~/.mempalace/hook_state/mine.pid` before/after a mine run -> cleared after exit.

Closes #1181, #1182.


bensig (approved):
**Reviewed locally on \`feat/init-mine-ux\`. Approve, no asks.** Tight scope, well-tested, backward-compat preserved by deliberate design.

Full pytest: **1247 passed in 48s** — +13 from v3.3.3 baseline, mapping to the new test_cli/test_miner cases this PR adds.

## Strengths

- **Backward-compat flag matrix is the right call.** \`--yes\` stays scoped to entity auto-accept; the new \`--auto-mine\` flag is the mine-skip toggle. Anyone running \`mempalace init --yes <dir>\` in a script today gets exactly the same behaviour after this lands. The matrix in the PR body is explicit, and the inline comment in \`cmd_init\` (\"\`--yes\` is SCOPED to entity auto-accept and does NOT imply mining\") makes the design defendable in code review six months from now.
- **Single corpus walk.** \`scan_project\`'s output feeds both the user-visible scope estimate AND \`mine()\` via the new \`files=\` kwarg. No double-walk. The \`mine()\` signature change preserves backward-compat (\`files=None\` falls through to the existing \`scan_project\` call).
- **Footgun mitigation is honest.** Scope estimate prints BEFORE the prompt and the comment explicitly names the failure mode (\"hitting Enter on a default-Y prompt with no size cue\"). \`<1 MB\` floor on \`_format_size_mb\` so small projects never see a misleading \`0 MB\`.
- **\`EOFError\` handling for piped stdin** — non-interactive callers get \"decline\" instead of a \`raise EOFError\`. Won't block scripts that pipe \`init\` from somewhere.
- **Errors surfaced, not swallowed.** Mine failure → \`sys.exit(1)\` so downstream scripts see the non-zero status.

### Ctrl-C handling

This is exactly the right level of mature:

- \`try/except KeyboardInterrupt\` around the file loop with an inner \`try\` capturing \`last_file\` for the summary.
- Outer \`except\` prints the clean \`files_processed: N/M\` / \`drawers_filed: K\` / \`last_file: …\` block.
- \`sys.exit(130)\` for standard SIGINT semantics — shells and CI runners read this correctly as user-interrupted.
- \`finally\` still runs (PID cleanup happens), and the comment explicitly calls out that a second Ctrl-C during the summary print propagates to the default handler. Knowing what you're *not* trying to catch is half the discipline.
- Resume guidance points at the idempotent re-mine path with deterministic IDs — already-filed drawers upsert to the same row, no duplicates. Honest description of why partial progress is safe to leave in place.

### PID-file cleanup

\`_cleanup_mine_pid_file\` is its own function with a long docstring explaining the cross-session reasoning. The \"only remove if the file claims our own PID\" check is the part that prevents the bug where a separate concurrent mine gets its lock file yanked. Defensive in exactly the right way.

## Minor observations (not blockers)

1. \`last_file\` is updated inside both the inner exception handler and the post-\`process_file\` line — slightly redundant, but the comment explains why. Cleaner alternative would be to set \`last_file\` before \`process_file()\` so the variable always reflects \"the file we attempted last\", but the current pattern works.
2. \`_format_size_mb\` and \`_maybe_run_mine_after_init\` extracted for unit-testability — exactly the right move and a contrast with the (still-good) PR #1185 where I had to ask for tests. Nothing to do here.

---

Ship it. Same review-quality bar as #1179.

Issue #1180, opened 2026-04-24
feat(graph): cross-wing tunnels by shared topics


igorls:
## Motivation

Wings currently connect only through explicit relationships. Projects that use the same tech stack, framework, vendor, or concept often have meaningfully related memories across wings, but today there is no way to surface that.

Example: if two wings both discuss a specific framework (Angular, OpenAPI, a given cloud vendor), the user searching for notes about that framework in one wing may reasonably want to see what the other wing captured about it.

The LLM refine step introduced in #1150 already emits a `TOPIC` label for framework/vendor/product references the user confirms. That is the right signal source for this feature.

## Proposal

After confirmation, when two wings share one or more TOPICs above a weight threshold, automatically drop a tunnel between them in `palace_graph.py`. Tunnels are labeled with the shared topic and its co-occurrence strength.

## Design tradeoffs worth discussing before building

- **Weight threshold.** Too loose and every wing links to every other via broadly-used topics (Python, Docker). Too tight and real connections are missed. Needs a sensible default plus a config knob.
- **Signal source.** Topics-from-LLM only, or also entity co-occurrence and manifest-dependency overlap (e.g. two `package.json` files declaring the same non-trivial dependency)? Manifest overlap is harder signal but narrower scope.
- **When to compute.** Mine-time (slow first mine, nothing extra at search), search-time (slow search, always fresh), or nightly rebuild. Mine-time is probably right since the graph changes infrequently.
- **User control.** Some shared topics make useful tunnels, others are noise. Consider: a confirm step analogous to `confirm_entities`, or a per-topic allowlist/denylist the user maintains once.
- **Display.** How are these tunnels surfaced in search results? A separate "related wings" section, or inline as breadcrumbs on each drawer?

## Acceptance criteria (draft)

- `palace_graph.py` can add topic-weighted tunnels between two wings given a shared TOPIC set
- Mine-time pass (or explicit `mempalace rebuild-tunnels` command) populates these from confirmed entity data
- Search results can traverse tunnels to surface related drawers from adjacent wings
- Config knob for minimum weight threshold
- Tests covering: topic-only link, manifest-only link (if included), threshold honored, no link when below threshold

## Out of scope for this issue

- Automatic tunnel naming / semantic labels beyond the shared topic
- Cross-palace tunnels
- Dependency graph ingestion from lockfiles (separate idea)

PR #1184, opened 2026-04-24
feat(graph): cross-wing tunnels by shared topics (#1180)

igorls:
## Summary

Implements #1180: when two wings share one or more confirmed `TOPIC` labels (the LLM-refine bucket from `mempalace init --llm`), the miner automatically drops a symmetric tunnel between them at mine time. This surfaces meaningful cross-wing connections via the palace graph for projects that share frameworks, vendors, or recurring concepts.

## Design decisions for the tradeoffs called out in the issue

- **Signal source: TOPIC labels only.** Manifest-dependency overlap is intentionally out of scope for v1 — it's harder signal with a narrower scope, and the LLM-refined topic set already captures the framework / vendor case the issue motivates. Worth a separate PR.
- **When to compute: mine-time, post-file-loop.** Cheaper than search-time (graph changes infrequently) and avoids the freshness lag of a nightly job. Tunnel creation runs after every file in the current wing has been processed and degrades quietly on failure (warning printed, mine still succeeds).
- **Threshold: simple per-pair shared-topic count.** Default `1` (any single shared topic is enough). Configurable via `MEMPALACE_TOPIC_TUNNEL_MIN_COUNT` env var or `topic_tunnel_min_count` in `~/.mempalace/config.json`. Bump to `2+` for users whose projects share lots of common-tech labels (Python, Docker, Git) and only want meaningfully overlapping wings to link.
- **Persistence: extend `~/.mempalace/known_entities.json`.** Topics are persisted under a new `topics_by_wing` map keyed by wing name; the per-project `entities.json` gains a `topics` list as audit trail. The miner's `SKIP_FILENAMES` already excludes `entities.json`, so no churn there. Wing names in `topics_by_wing` are explicitly *not* surfaced through the flat known-name set (`_load_known_entities`) used by drawer-tagging.
- **User control: out of scope for v1.** Per-topic allow/deny lists are a future idea — adding one wrong topic at worst creates a low-traffic tunnel and never alters drawer storage, so the cost of a bad classification is small.
- **Display: not in this PR.** Tunnels being present in the graph is enough — they're retrievable via the existing `list_tunnels` / `follow_tunnels` API. Surfacing them in search results is a follow-up.

## Behavior changes

- `mempalace init --llm`: confirmed topics are now persisted (previously TOPIC classifications collapsed into the `uncertain` bucket and were dropped when the user accepted the defaults).
- `mempalace mine <project>`: after the file loop, computes topic tunnels for the freshly-mined wing against every other wing recorded in `topics_by_wing`. Prints `Topic tunnels: +N cross-wing link(s)` when tunnels are created. Without `--llm` confirmed topics, this is a no-op.
- `palace_graph`: gains `compute_topic_tunnels(topics_by_wing, min_count, label_prefix)` and `topic_tunnels_for_wing(wing, ...)`. Both route through the existing `create_tunnel` API so storage, dedup, and the `list_tunnels` / `follow_tunnels` / `delete_tunnel` paths just work.

## Test plan

- [x] `tests/test_palace_graph_tunnels.py::TestTopicTunnels` — 9 unit tests covering shared-topic link, threshold honored, threshold blocks single overlap, case-insensitive overlap, multi-topic per-pair links, three-wings pairwise expansion, `topic_tunnels_for_wing` only links the named wing, dedup on recompute, empty-input no-op.
- [x] `tests/test_known_entities_registry.py` — 7 new tests covering `topics_by_wing` persistence, replace-on-reinit, multi-wing coexistence, case-insensitive dedup, `get_topics_by_wing` reader, no-leak of wing names into the flat known-set, and the wing-less code path.
- [x] `tests/test_miner.py` — 3 end-to-end tests exercising the full mine-time flow: tunnel created when shared topic exists, no tunnel when threshold blocks the overlap, no tunnel when only one wing has topics.
- [x] `tests/test_llm_refine.py` — updated `TOPIC` routing test to assert the new `topics` bucket (separate from `uncertain`) and added an `AMBIGUOUS` assertion to lock in the still-uncertain behavior for that label.
- [x] Full suite passes locally: `1255 passed`.
- [x] `ruff check` + `ruff format --check` clean against the CI-pinned `ruff>=0.4.0,<0.5`.
- [x] No real names, real paths, or private data introduced — all test fixtures use generic placeholders (`wing_alpha`, `wing_beta`, `Alice`, `Bob`, `foo`, `bar`, `Angular`, `OpenAPI`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Added
 
-- **Cross-wing topic tunnels.** When two wings have confirmed `TOPIC` labels in common (the LLM-refine bucket from `mempalace init --llm`), the miner now drops a symmetric tunnel between them at mine time so the palace graph reflects shared themes (frameworks, vendors, recurring concepts). Tunnels are routed through the existing `create_tunnel` storage so they share dedup and persistence with explicit tunnels. Threshold is configurable via `MEMPALACE_TOPIC_TUNNEL_MIN_COUNT` env var or `topic_tunnel_min_count` in `~/.mempalace/config.json` (default `1`). Manifest-dependency overlap and per-topic allow/deny lists remain out of scope. (#1180)
+- **Cross-wing topic tunnels.** When two wings have confirmed `TOPIC` labels in common (the LLM-refine bucket from `mempalace init --llm`), the miner now drops a symmetric tunnel between them at mine time so the palace graph reflects shared themes (frameworks, vendors, recurring concepts). Tunnels are routed through the existing `create_tunnel` storage so they share dedup and persistence with explicit tunnels. Topic tunnels are stored under a synthetic `topic:<name>` room and tagged with `kind: "topic"` on the stored dict — this keeps them distinct from literal folder-derived rooms of the same name (a wing with both an `Angular` folder room and an `Angular` topic tunnel no longer collides at `follow_tunnels` read time) and gives LLMs scanning `list_tunnels` a visible discriminator. Threshold is configurable via `MEMPALACE_TOPIC_TUNNEL_MIN_COUNT` env var or `topic_tunnel_min_count` in `~/.mempalace/config.json` (default `1`). Manifest-dependency overlap and per-topic allow/deny lists remain out of scope. (#1180)
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ---
 
+## [3.3.4] — unreleased
+
+### Added
+
+- **Cross-wing topic tunnels.** When two wings have confirmed `TOPIC` labels in common (the LLM-refine bucket from `mempalace init --llm`), the miner now drops a symmetric tunnel between them at mine time so the palace graph reflects shared themes (frameworks, vendors, recurring concepts). Tunnels are routed through the existing `create_tunnel` storage so they share dedup and persistence with explicit tunnels. Threshold is configurable via `MEMPALACE_TOPIC_TUNNEL_MIN_COUNT` env var or `topic_tunnel_min_count` in `~/.mempalace/config.json` (default `1`). Manifest-dependency overlap and per-topic allow/deny lists remain out of scope. (#1180)
+
+---
+
 ## [3.3.3] — 2026-04-23
 
 ### Bug Fixes
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -117,21 +117,34 @@ def cmd_init(args):
     if languages_tuple != ("en",):
         print(f"  Languages: {', '.join(languages_tuple)}")
     detected = discover_entities(args.dir, languages=languages_tuple, llm_provider=llm_provider)
-    total = len(detected["people"]) + len(detected["projects"]) + len(detected["uncertain"])
+    total = (
+        len(detected["people"])
+        + len(detected["projects"])
+        + len(detected.get("topics", []))
+        + len(detected["uncertain"])
+    )
     if total > 0:
         confirmed = confirm_entities(detected, yes=getattr(args, "yes", False))
         # Save confirmed entities to <project>/entities.json (per-project
         # audit trail — user can inspect or hand-edit) AND merge into the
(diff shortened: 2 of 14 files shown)


bensig (approved):
**Reviewed locally on \`feat/cross-wing-topic-tunnels\`. Approve, with one mechanical ask + one design question.**

Full pytest on this branch: **1255 passed in 42s** — +8 over #1183 baseline, mapping to the new tunnel + registry tests this PR adds.

## Mechanical ask: rebase

The PR is showing CONFLICTING. Real conflict surface is small — one file:

- \`mempalace/config.py\` — both this PR and #1185 added new properties to \`MempalaceConfig\` (your \`embedding_device\` from #1185 vs \`topic_tunnel_min_count\` here). Trivial resolution; both properties just need to coexist.

\`tests/test_miner.py\` and \`mempalace/miner.py\` both auto-merged cleanly when I tried locally, so the actual rebase is one config.py merge.

## Design question (not a blocker)

**Topic-name-as-room semantics.** \`compute_topic_tunnels\` uses the topic itself (e.g. \`\"Angular\"\`) as the \`room\` field on both ends of the tunnel:

\`\`\`python
tunnel = create_tunnel(
    source_wing=wa, source_room=room,    # room = topic name
    target_wing=wb, target_room=room,
    label=f\"{label_prefix}: {room}\",
)
\`\`\`

If a project happens to have a literal \`Angular\` room from autodetected folder structure, the tunnel target collides with that real room. Is the intent that \`follow_tunnels\` / \`list_tunnels\` callers can transparently traverse from drawers in a real \"Angular\" room into topic-tunnel-linked wings, or are these meant as virtual graph-only connections? Either is fine — just want to confirm you considered the case where topic strings collide with structural room names.

If they're meant as virtual: a \`label_prefix\`-style namespace on the room field (e.g. \`\"topic:Angular\"\`) would make the distinction visible at \`list_tunnels\` read time. If they're meant to merge with literal rooms: the current shape is exactly right and ignore me.

## Strengths

- **Out-of-scope decisions are explicit and defensible.** Manifest-dep overlap deferred, per-topic allow/deny lists deferred, search-result surfacing deferred. Each labeled as future work with the rationale (\"a wrong topic at worst creates a low-traffic tunnel and never alters drawer storage\").
- **Mine-time, not search-time.** Cheaper, no freshness lag, degrades quietly on failure (\`try/except\` prints warning, mine still succeeds with the right exit code).
- **Defensive type-checking everywhere.** \`isinstance\` guards on wing names, topic lists, the \`topics_by_wing\` structure. Won't crash on malformed registry data — important because users can hand-edit \`known_entities.json\`.
- **Reuses existing tunnel storage.** \`create_tunnel\` / \`list_tunnels\` / \`follow_tunnels\` / \`delete_tunnel\` paths all just work. No new graph code to maintain past this PR.
- **\`topics_by_wing\` correctly namespaced from flat known_entities.** The \`_refresh_known_entities_cache\` change skips the wing-keyed map when building the flat name set — wing names don't leak into drawer-tagging. The inline comment makes the intent obvious to the next reader.
- **Replace-not-union semantics for topics.** Re-running \`init\` reflects the user's latest confirmation rather than accumulating stale labels. Right call; users would otherwise get a one-way ratchet of topic counts.
- **Threshold knob.** \`topic_tunnel_min_count\` defaults to 1, configurable via env or config. Users with lots of common-tech labels (Python, Docker, Git) can bump to 2+ for meaningful overlaps only. Documented in the property docstring.
- **Casing preserved from \"first seen\"** while overlap detection is case-insensitive. Won't show jarring case mismatches in tunnel labels (\`\"angular\"\` from one wing meeting \`\"Angular\"\` from another renders consistently).
- **Test coverage is comprehensive.** 9 tunnel tests + 7 registry tests + new test_miner cases. The cases I'd most want covered (dedup on recompute, threshold blocks single overlap, three-wing pairwise, case-insensitive overlap, empty-input no-op) are all there.

## Minor (not blockers)

1. \`_compute_topic_tunnels_for_wing\` reads \`MempalaceConfig()\` every mine call. Cheap (file read with caching), but technically re-reads if mine is called multiple times in one session. Negligible.
2. The mine-time tunnel print is a single line, only fires when tunnels were actually created. Quiet by default — good.

---

Once rebased, ship it. The design question above is genuinely a question, not a blocker — your answer determines whether to refactor the room field, but either answer is defensible.

igorls:
@bensig went with the namespaced topic: room + kind field (option A in your review); see 865a36b.

PR #1185, opened 2026-04-24
perf(mining): batch per-chunk upserts + optional GPU acceleration

igorls:
## Summary

`mempalace mine` was slow because every drawer was upserted in its own `collection.upsert(...)` call (one tokenizer + ONNX forward pass per chunk), and because no `EmbeddingFunction` was ever wired through the backend (CPU-only, by omission).

This PR fixes both, independently:

1. **Batched upsert** in [mempalace/miner.py](mempalace/miner.py) and [mempalace/convo_miner.py](mempalace/convo_miner.py) — one `collection.upsert(...)` per file instead of per chunk.
2. **Hardware-accelerated embedding function** — new [mempalace/embedding.py](mempalace/embedding.py) wraps ChromaDB's `ONNXMiniLM_L6_V2` with configurable `preferred_providers`. Selectable via `MEMPALACE_EMBEDDING_DEVICE` (`auto` / `cpu` / `cuda` / `coreml` / `dml`) or `embedding_device` in `~/.mempalace/config.json`. Wired through [mempalace/backends/chroma.py](mempalace/backends/chroma.py) so miner writes and searcher reads always agree.

Same `all-MiniLM-L6-v2` model, same 384-dim vectors — **existing palaces need no rebuild** when turning GPU on. The factory subclasses `ONNXMiniLM_L6_V2` and renames it `"default"` so ChromaDB's persisted-EF check matches palaces created with the bare library default.

## Benchmark

Hardware: Intel Core i9-12900KF (24 threads) · NVIDIA RTX 3090 · Linux 6.8 · Python 3.13.12
Software: `chromadb 1.5.7` · `onnxruntime 1.24.4` (CPU) / `onnxruntime-gpu 1.25.0` (CUDA)
Method: seeded synthetic corpora, fresh palace per run, model warm-loaded before the first scenario.

| Scenario | Files | Drawers | Per-chunk + CPU <br>(old baseline) | Batched + CPU <br>(this PR, no GPU) | Batched + CUDA <br>(this PR, `mempalace[gpu]`) | End-to-end |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Small (~50 paragraphs) | 10 | 70 | 3.36 s · 21 drw/s | 2.08 s · 34 drw/s | **0.49 s · 143 drw/s** | **6.9×** |
| Medium (~200 paragraphs) | 20 | 568 | 19.77 s · 29 drw/s | 8.07 s · 70 drw/s | **2.15 s · 264 drw/s** | **9.2×** |
| Large (~500 paragraphs) | 10 | 709 | 21.13 s · 34 drw/s | 8.95 s · 79 drw/s | **2.13 s · 334 drw/s** | **9.9×** |

Isolating the two wins (medium scenario, same numbers sliced):

|  | CPU | CUDA | GPU-over-CPU |
| --- | ---: | ---: | ---: |
| Per-chunk upsert (old path) | 29 drw/s | 47 drw/s | 1.6× |
| Batched upsert (new path) | 70 drw/s | **264 drw/s** | 3.8× |
| **Batching-over-per-chunk** | **2.4×** | **5.6×** | — |

Key observation: **GPU alone is not enough.** Per-chunk + CUDA (47 drw/s) is only modestly better than per-chunk + CPU (29 drw/s) — per-call launch overhead dominates. Batching is what unlocks the accelerator.

Reproduce via [`benchmarks/mine_bench.py`](benchmarks/mine_bench.py):

```bash
# CPU:
uv run python benchmarks/mine_bench.py

# GPU:
uv pip install -e '.[gpu]'
export MEMPALACE_EMBEDDING_DEVICE=cuda
python benchmarks/mine_bench.py --device cuda
```

## Install paths

```bash
pip install mempalace[gpu]      # NVIDIA CUDA (onnxruntime-gpu)
pip install mempalace[dml]      # DirectML (Windows AMD/Intel/NVIDIA)
pip install mempalace[coreml]   # macOS Neural Engine
```

`mempalace mine` header now prints `Device: cpu|cuda|...` so users can confirm the accelerator actually engaged.

### Known ergonomics gap (follow-up)

`pip install mempalace[gpu]` alone is not sufficient on Linux today — current `onnxruntime-gpu` wheels do not bundle cuDNN/cuBLAS/cudart, and the `nvidia-*-cu12` runtime wheels require `LD_LIBRARY_PATH` to be set before Python starts. The failure mode is a silent fallback to CPU. A follow-up PR should: (a) extend `[gpu]` to also pull `nvidia-cudnn-cu12 nvidia-cuda-runtime-cu12 nvidia-cublas-cu12`, (b) `ctypes.CDLL`-preload their `.so` files from `site-packages/nvidia/*/lib` at import time (same trick PyTorch uses), and (c) turn the silent fallback into a loud, actionable warning. Not scoped into this PR — this one is about the core speedup.

## Principles respected

- **Local-first, zero API** — all acceleration is local ONNX execution, no cloud dependency.
- **Verbatim always** — batching only changes upsert granularity, not chunking or content.
- **Privacy by architecture** — no telemetry, no new network surface.

## Test plan

- [x] `uv run pytest tests/ --ignore=tests/benchmarks` — 1234 passed locally
- [x] `uvx --from 'ruff>=0.4.0,<0.5' ruff check` + `format --check` clean
- [x] `pre-commit` hooks (ruff + ruff-format) pass
- [x] CPU benchmark reproduces the published numbers from the final script location
- [x] CUDA benchmark reproduces the published numbers from the final script location
- [x] `MEMPALACE_EMBEDDING_DEVICE=cuda` on a host *without* CUDA logs a clear fallback warning instead of crashing
- [x] Existing CPU-created palaces continue to work with the new EF (name-spoofing check)

### DIFF
--- a/benchmarks/mine_bench.py
+++ b/benchmarks/mine_bench.py
@@ -1,0 +1,301 @@
+"""Mining throughput benchmark: per-chunk vs batched upsert, CPU vs GPU.
+
+Compares the legacy per-chunk ``add_drawer`` loop against the batched
+``collection.upsert`` path introduced in the "batched upsert + GPU" PR.
+Runs both paths on an identical seeded synthetic corpus, reports
+wall-clock time + drawers/sec, and prints a markdown table suitable
+for pasting into a PR description.
+
+Usage
+-----
+
+    # CPU (whatever onnxruntime is installed — CPU if you don't have
+    # onnxruntime-gpu):
+    uv run python benchmarks/mine_bench.py
+
+    # GPU (NVIDIA):
+    uv venv /tmp/gpu && source /tmp/gpu/bin/activate
+    uv pip install -e '.[gpu]' 'nvidia-cudnn-cu12>=9,<10' \\
+        'nvidia-cuda-runtime-cu12' 'nvidia-cublas-cu12'
+    export LD_LIBRARY_PATH=$(python -c "import nvidia.cudnn, os; \\
+        print(os.path.dirname(nvidia.cudnn.__file__)+'/lib')"):$LD_LIBRARY_PATH
+    MEMPALACE_EMBEDDING_DEVICE=cuda python benchmarks/mine_bench.py
+
+Flags
+-----
+
+    --device cpu|cuda|coreml|dml|auto   Override MEMPALACE_EMBEDDING_DEVICE
+    --scenarios small,medium,large      Which scenarios to run
+    --seed 42                           RNG seed for reproducibility
+"""
+
+from __future__ import annotations
+
+import argparse
+import hashlib
+import os
+import random
+import shutil
+import string
+import sys
+import tempfile
+import time
+from datetime import datetime
+from pathlib import Path
+
+
+def build_corpus(dest: Path, n_files: int, paragraphs_per_file: int, seed: int) -> None:
+    """Generate ``n_files`` markdown files of random words under ``dest``."""
+    rng = random.Random(seed)
+    dest.mkdir(parents=True, exist_ok=True)
+    for i in range(n_files):
+        paragraphs = []
+        for _ in range(paragraphs_per_file):
+            words = [
+                "".join(rng.choices(string.ascii_lowercase, k=rng.randint(3, 10)))
+                for _ in range(12)
+            ]
+            paragraphs.append(" ".join(words))
+        (dest / f"doc_{i:03d}.md").write_text("\n\n".join(paragraphs))
+    (dest / "mempalace.yaml").write_text(
+        "wing: bench\n"
+        "rooms:\n"
+        "  - name: general\n"
+        "    description: all\n"
+        "    keywords: [general]\n"
+    )
+
+
+def _process_file_unbatched(filepath, project_path, collection, wing, rooms, agent, closets_col):
+    """Legacy per-chunk upsert path (pre-batching).
+
+    Reproduces the exact loop shape the miner used before this PR so the
+    comparison is apples-to-apples; only the upsert granularity differs.
+    """
+    from mempalace import miner
+    from mempalace.palace import (
+        build_closet_lines,
+        file_already_mined,
+        mine_lock,
+        purge_file_closets,
+        upsert_closet_lines,
+    )
+
+    source_file = str(filepath)
+    if file_already_mined(collection, source_file, check_mtime=True):
+        return 0, "general"
+    try:
+        content = filepath.read_text(encoding="utf-8", errors="replace")
+    except OSError:
+        return 0, "general"
+    content = content.strip()
+    if len(content) < miner.MIN_CHUNK_SIZE:
+        return 0, "general"
+    room = miner.detect_room(filepath, content, rooms, project_path)
+    chunks = miner.chunk_text(content, source_file)
+
+    with mine_lock(source_file):
+        if file_already_mined(collection, source_file, check_mtime=True):
+            return 0, room
+        try:
+            collection.delete(where={"source_file": source_file})
+        except Exception:
+            pass
+        drawers_added = 0
+        for chunk in chunks:
+            miner.add_drawer(
+                collection=collection,
+                wing=wing,
+                room=room,
+                content=chunk["content"],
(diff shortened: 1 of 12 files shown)


bensig (approved):
**Reviewed locally on \`perf/batched-upsert-gpu\`.** Quality is high — clean design, honest fallbacks, well-documented EF workaround, real benchmarks. **Approving** with one ask + a few minor notes worth a follow-up commit if you want.

Full pytest matches v3.3.3: **1234 passed in 35s** on this branch.

## Strengths

- **Two wins kept independent.** Batching gives ~2.4× CPU; GPU adds another 3.8× on top. Either alone is useful, the combination is multiplicative.
- **EF identity rename to \`\"default\"\`** is the high-value detail. The persisted-EF check in ChromaDB 1.x would otherwise force a palace rebuild on the cutover. Lets the new GPU-capable EF read palaces created by the bare library default and vice versa. The inline comment in \`_build_ef_class\` documents the why clearly enough that the next reader won't accidentally undo it.
- **Fallback chain is solid.** Unknown device → CPU+warn, missing accelerator → CPU+warn, EF construction error → ChromaDB's own default. The \"still works on a laptop\" goal is honored.
- **One-shot warning via \`_WARNED\`** — won't spam logs across a 10k-file mine. Process-global EF cache keyed by provider tuple — one model load per device per process.
- **\`Device:\` line in the mine banner** — small UX win, users see whether GPU actually engaged.
- **Optional deps split into 3 extras** (\`gpu\` / \`dml\` / \`coreml\`) — no over-pulling for users who only have one acceleration path available.

## Ask: unit tests for \`mempalace/embedding.py\`

The module has zero tests today. \`_resolve_providers\` has 5+ branches all driven by a string and all mockable without a GPU:

\`\`\`python
@patch(\"onnxruntime.get_available_providers\")
def test_auto_picks_cuda(mock):
    mock.return_value = [\"CUDAExecutionProvider\", \"CPUExecutionProvider\"]
    assert _resolve_providers(\"auto\") == ([\"CUDAExecutionProvider\", \"CPUExecutionProvider\"], \"cuda\")

def test_auto_falls_to_cpu(mock): ...
def test_cuda_missing_warns(...)
def test_unknown_device_warns(...)
def test_onnxruntime_import_error(...)
\`\`\`

This is exactly the kind of code that breaks silently in three months when ChromaDB or onnxruntime renames a provider string and we never notice because nothing exercises the resolver. Cheap to add, high coverage win.

## Minor notes (not blockers)

1. **Atomicity trade-off is unmentioned in the diff.** Per-chunk → per-file upsert means one bad chunk takes the whole file's chunks down. Almost certainly the right call, but worth a one-line comment in \`process_file\` so the next reader doesn't accidentally \"fix\" it back to per-chunk thinking they're improving error isolation.
2. **\`filed_at\` is now per-file** in \`convo_miner\` (was per-chunk). Practically fine, arguably more correct semantically — flagging only because behavior changed and isn't called out.
3. **\`tomli\` dep + \`add_to_known_entities\` helper** look unrelated to the PR title — likely develop-merge overlap. Worth a quick check with #1183 to make sure we're not duplicating work.
4. **Subclassing \`chromadb.utils.embedding_functions.ONNXMiniLM_L6_V2\`** is an internal-API dependency on ChromaDB. Calculated risk for the persistence-compat win, but worth a comment so a future reader knows it's deliberate, not accidental.
5. **CUDA-installed-but-no-driver edge case**: \`onnxruntime-gpu\` may report \`CUDAExecutionProvider\` as available on a machine without a CUDA driver — session creation then fails at runtime. The \`[CUDAExecutionProvider, CPUExecutionProvider]\` cascade may save us, but a brief test or note would confirm.

---

None of the minor notes block — they're \"the next person reading this code in six months will thank you\" material. The unit tests are the one I'd push for before merge.

Approving on the strength of the design, test passes, and clean fallback story.

igorls:
@copilot+gpt-5.5 apply changes based on the comments in [this thread](https://github.com/MemPalace/mempalace/pull/1185#pullrequestreview-4173831183)

also read the comment from ben

PR #1191, opened 2026-04-25
fix: prevent HNSW index bloat from resize+persist cycles

### LINKED ISSUE #344: HNSW index bloat: link_lists.bin grows to 441GB when mining >10K drawers
## Summary

When mining a project with ~56K drawers, ChromaDB's HNSW index file `link_lists.bin` grows to **441GB apparent size (168GB on disk)** instead of the expected ~8MB. This caused two sequential system crashes due to memory and disk exhaustion before the problem was identified. The `mempalace status` command segfaults (exit 139) when attempting to load the corrupted index.

## Environment

- mempalace v3.0.0
- chromadb 1.5.7
- macOS Darwin 25.3.0

## Steps to Reproduce

1. Have a project with ~50K+ text files (or fewer files that chunk into 50K+ drawers)
2. Run `mempalace mine <project_dir>`
3. Observe `link_lists.bin` growing to hundreds of GB
4. `mempalace status` segfaults

Smaller mines (~10K drawers) work fine. The issue appears at scale.

## Root Cause

ChromaDB creates the HNSW collection with default parameters:
- `hnsw:batch_size = 100`
- `hnsw:sync_threshold = 1000`
- Initial capacity = 1000
- `resize_factor = 1.2`

When mempalace inserts drawers one at a time via `collection.upsert()`, each 1000 inserts trigger a `persistDirty()` call, and the index resizes ~30 times to accommodate 56K elements (1000 -> 1200 -> 1440 -> ... -> 56000+). The `persistDirty()` function uses relative seek positioning in `link_lists.bin`, and after many resize cycles the seek positions drift, causing macOS to extend the sparse file with zero-filled regions. Each resize+persist cycle compounds the problem.

Relevant upstream issues:
- chroma-core/chroma#6621 -- proposes `hnsw:initial_capacity` (not yet merged)
- nmslib/hnswlib#633 -- chunked array allocation (not yet merged)
- chroma-core/chroma#5925 -- user with 240K vectors reports failure loading HNSW index

## Impact

- **Disk exhaustion**: 441GB file on a laptop with limited SSD space
- **System crashes**: Two sequential crashes from memory/disk pressure before the cause was identified
- **Data loss risk**: The HNSW index stores the actual embedding vectors; when corrupted, they cannot be recovered from ChromaDB's SQLite (which only stores IDs, metadata, and documents)
- **Segfaults**: `mempalace status` and `mempalace search` crash when trying to load the bloated index

## Workaround

Delete the corrupted index and re-mine:
```bash
rm -rf ~/.mempalace/palace/*/
mempalace mine <project_dir>
```

## Fix

PR incoming -- two changes:
1. Pass `hnsw:batch_size=50000` and `hnsw:sync_threshold=50000` on collection creation to defer persistence until large batches complete
2. Batch per-file drawer upserts instead of inserting one drawer at a time, reducing the number of resize operations

funguf:
## Summary

Rebases #346 onto current `develop` and verifies the fix on a fresh 39,792-drawer workload. The original PR has been stalled awaiting rebase since 2026-04-11 (@bensig requested rebase, @robot-rocket-science agreed but went silent).

Fixes #344. Supersedes #346 — credit retained via `Co-authored-by:` trailer.

## What changed

**`mempalace/backends/chroma.py`** — adds `_HNSW_BLOAT_GUARD = {"hnsw:batch_size": 50_000, "hnsw:sync_threshold": 50_000}` and merges it into the collection metadata in both `get_collection(create=True)` and the legacy `create_collection`, alongside the existing `hnsw:space`.

**`mempalace/miner.py` `process_file`** — replaces the per-chunk `add_drawer()` loop with a single batched `collection.upsert(documents=..., ids=..., metadatas=...)` per file. The batched path recomputes the same deterministic drawer id (`drawer_{wing}_{room}_{sha256(source_file + chunk_index)[:24]}`) and preserves every existing metadata field (`wing`, `room`, `source_file`, `chunk_index`, `added_by`, `filed_at`, `normalize_version`, `source_mtime`, `hall`, `entities`) so re-mines remain on-disk-compatible with previously-filed drawers. The surrounding `mine_lock`, `file_already_mined` re-check, `collection.delete(where=...)` purge, and closet-building are untouched. `add_drawer()` is kept for any external callers.

## Empirical verification

Rebuilt a palace from scratch on **39,792 drawers across 5 wings** with this fix applied — that's ~4× past the threshold where the bug was reproducible.

| Metric | Before fix (~15K drawers) | After fix (39,792 drawers) |
|---|---|---|
| `du -sh palace/` | 30 GB on disk | **376 MB** |
| `link_lists.bin` apparent | 614 GB sparse | **0 bytes** |
| `mempalace status` | Segfault (exit 139) | Returns instantly |
| `mempalace search` | Segfault | Returns ranked results |
| `mempalace repair` | Segfault | n/a |

Both Chroma collection dirs in the post-fix palace have `link_lists.bin` at literal 0 bytes (zero, not zero-filled-sparse). The metadata-persistence concern raised in #1161 was reproduced as a 5-line test against chromadb 1.5.8 and **does not affect** the keys this PR sets — `hnsw:batch_size` and `hnsw:sync_threshold` survive `PersistentClient` reopen cleanly.

## Migration note

The HNSW config is only honoured at collection-create time. **Users on already-bloated palaces still need to nuke and re-mine** — there's no in-place fix because Chroma treats HNSW metadata as immutable post-creation. Worth a release-note line; happy to add a one-time warning in `_get_collection()` if maintainers want one.

## Test plan

- [x] Mine `<1K` drawer wing — behaviour unchanged (esp32-tuya-gw, 717 drawers)
- [x] Mine `~1K` drawer wing — behaviour unchanged (chirpa, 1367 drawers)
- [x] Mine `~10K` drawer wing — no bloat (home, 9,712 drawers)
- [x] Mine `~28K` drawer wing — no bloat (git-home, 27,862 drawers)
- [x] Sequential mine 5 wings → 39,792 drawers total → palace 376 MB, link_lists.bin 0 bytes
- [x] `mempalace status` and `mempalace search` no segfault on post-fix palace
- [x] Drawer id formula and metadata fields verified byte-identical against pre-fix `add_drawer()` for re-mine compatibility
- [x] chromadb 1.5.8 persists batch_size + sync_threshold through reopen (verified separately, addresses #1161 concern)

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -28,6 +28,31 @@ _REQUIRED_OPERATORS = frozenset({"$eq", "$ne", "$in", "$nin", "$and", "$or", "$c
 _OPTIONAL_OPERATORS = frozenset({"$gt", "$gte", "$lt", "$lte"})
 _SUPPORTED_OPERATORS = _REQUIRED_OPERATORS | _OPTIONAL_OPERATORS
 
+# HNSW tuning to prevent link_lists.bin bloat on large mines (#344).
+#
+# With default params (batch_size=100, sync_threshold=1000, initial capacity
+# 1000), inserting tens of thousands of drawers triggers ~30 index resizes
+# and hundreds of persistDirty() calls. persistDirty uses relative seek
+# positioning in link_lists.bin; accumulated seek drift across resize cycles
+# causes the OS to extend the sparse file with zero-filled regions, each
+# cycle compounding the next. Result: link_lists.bin grows into hundreds of
+# GB sparse, after which `status`/`search`/`repair` segfault.
+#
+# Setting large batch and sync thresholds at collection creation defers
+# persistence until a single large batch completes, breaking the resize+
+# persist feedback loop. Empirically validated on a 39,792-drawer rebuild
+# (palace 376 MB, link_lists.bin 0 bytes, no segfault) in 2026-04.
+#
+# Note: chromadb 1.5.x exposes a `collection.modify(configuration={"hnsw":
+# {"batch_size": ..., "sync_threshold": ...}})` retrofit path for already-
+# created collections (`UpdateHNSWConfiguration` in chromadb's API), but
+# this PR doesn't pursue that — once link_lists.bin has bloated, the index
+# is already corrupt and the only known recovery is a fresh mine.
+_HNSW_BLOAT_GUARD = {
+    "hnsw:batch_size": 50_000,
+    "hnsw:sync_threshold": 50_000,
+}
+
 
 def _validate_where(where: Optional[dict]) -> None:
     """Scan a where-clause for unknown operators and raise ``UnsupportedFilterError``.
@@ -992,7 +1017,11 @@ class ChromaBackend(BaseBackend):
         if create:
             collection = client.get_or_create_collection(
                 collection_name,
-                metadata={"hnsw:space": hnsw_space, "hnsw:num_threads": 1},
+                metadata={
+                    "hnsw:space": hnsw_space,
+                    "hnsw:num_threads": 1,
+                    **_HNSW_BLOAT_GUARD,
+                },
                 **ef_kwargs,
             )
         else:
@@ -1042,7 +1071,11 @@ class ChromaBackend(BaseBackend):
         ef_kwargs = {"embedding_function": ef} if ef is not None else {}
         collection = self._client(palace_path).create_collection(
             collection_name,
-            metadata={"hnsw:space": hnsw_space, "hnsw:num_threads": 1},
+            metadata={
+                "hnsw:space": hnsw_space,
+                "hnsw:num_threads": 1,
+                **_HNSW_BLOAT_GUARD,
+            },
             **ef_kwargs,
         )
         return ChromaCollection(collection)
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -60,6 +60,7 @@ from .version import __version__  # noqa: E402
 from .backends.chroma import (  # noqa: E402
     ChromaBackend,
     ChromaCollection,
+    _HNSW_BLOAT_GUARD,
     _pin_hnsw_threads,
     hnsw_capacity_status,
 )
@@ -285,7 +286,11 @@ def _get_collection(create=False):
             # so the retrofit runs every time _get_collection opens a cache).
             raw = client.get_or_create_collection(
                 _config.collection_name,
-                metadata={"hnsw:space": "cosine", "hnsw:num_threads": 1},
+                metadata={
+                    "hnsw:space": "cosine",
+                    "hnsw:num_threads": 1,
+                    **_HNSW_BLOAT_GUARD,
+                },
             )
             _pin_hnsw_threads(raw)
             _collection_cache = ChromaCollection(raw)
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -336,6 +336,42 @@ def test_chroma_backend_creates_collection_with_cosine_distance(tmp_path):
     assert col.metadata.get("hnsw:space") == "cosine"
 
 
+def test_chroma_backend_sets_hnsw_bloat_guard_on_creation(tmp_path):
+    """The HNSW guard from #344 must land on freshly-created collection metadata.
+
+    Without batch_size + sync_threshold, mining ~10K+ drawers triggers the
+    resize+persist drift that bloats link_lists.bin into hundreds of GB sparse
+    and segfaults `status` / `search` / `repair`. The guard belongs at
+    collection-creation time so every fresh palace gets it without needing
+    a runtime retrofit. Asserting both keys land on the persisted metadata
+    also covers the #1161 "config silently dropped" concern at CI time.
+    """
+    palace_path = tmp_path / "palace"
+
+    ChromaBackend().get_collection(
+        str(palace_path),
+        collection_name="mempalace_drawers",
+        create=True,
+    )
+
+    client = chromadb.PersistentClient(path=str(palace_path))
+    col = client.get_collection("mempalace_drawers")
+    assert col.metadata.get("hnsw:batch_size") == 50_000
+    assert col.metadata.get("hnsw:sync_threshold") == 50_000
+
+
+def test_chroma_backend_create_collection_sets_hnsw_bloat_guard(tmp_path):
+    """Same guard must apply via the legacy create_collection() path."""
+    palace_path = tmp_path / "palace"
+
+    ChromaBackend().create_collection(str(palace_path), "mempalace_drawers")
+
+    client = chromadb.PersistentClient(path=str(palace_path))
+    col = client.get_collection("mempalace_drawers")
+    assert col.metadata.get("hnsw:batch_size") == 50_000
+    assert col.metadata.get("hnsw:sync_threshold") == 50_000
+
+
 def test_fix_blob_seq_ids_converts_blobs_to_integers(tmp_path):
     """Simulate a ChromaDB 0.6.x database with BLOB seq_ids and verify repair."""
     db_path = tmp_path / "chroma.sqlite3"


igorls:
## Review

`mergeable: dirty` against `develop` — rebase onto current `develop` will not be a quiet refresh, the same area was substantially restructured on 2026-04-24 by `a4868a3` ("perf(mining): batch per-chunk upserts and add optional GPU acceleration").

### chroma.py — valuable, but needs adjustment on rebase

The HNSW guard is the real, novel fix `develop` doesn't have. The diff still bases on the older signature though — `develop`'s `get_collection` and `create_collection` now pass `**ef_kwargs` to plumb the GPU/CoreML/DML embedding function. A naïve rebase would silently drop that. The merged form should be:

```python
client.get_or_create_collection(
    collection_name,
    metadata={"hnsw:space": hnsw_space, **_HNSW_BLOAT_GUARD},
    **ef_kwargs,
)
```

…at both call sites (`get_collection` and `create_collection`).

### miner.py — superseded, please drop the hunk

`a4868a3` already batches per-file upserts. The current shape on `develop` is materially better than this PR's shape:

- A `_build_drawer_metadata()` helper factors the metadata construction. The PR re-inlines it, duplicating logic.
- The on-`develop` loop uses bounded sub-batches of `DRAWER_UPSERT_BATCH_SIZE = 1000`. The PR collapses to a single unbounded upsert per file, regressing the deliberate trade-off (a bad chunk fails its sub-batch instead of the whole file).

The HNSW-bloat root cause is resize/persist drift driven by the default thresholds — that's fixed by the chroma.py metadata, not by switching from "loop calling upsert" to "one upsert" (both forms hit the same hnswlib internals once batch/sync are large).

### Other notes

- The new try-block reduces to:
  ```python
  try:
      collection.upsert(...)
      drawers_added = len(drawer_ids)
  except Exception:
      raise
  ```
  That's a no-op handler and should be removed (moot if the whole hunk is dropped).
- The drawer-id formula and the metadata-field set are byte-identical to `add_drawer()` — the re-mine compatibility claim holds.
- A small regression test asserting the two HNSW keys land on a freshly-created collection's metadata would lock the contract and cover the #1161 "config not persisted" concern at CI time:
  ```python
  col = backend.get_collection(palace_ref, "drawers", create=True)
  assert col._raw.metadata["hnsw:batch_size"] == 50_000
  assert col._raw.metadata["hnsw:sync_threshold"] == 50_000
  ```
- Migration note is correct — HNSW metadata is immutable post-creation, so already-bloated palaces need a fresh mine. A one-time warning in `_get_collection()` when the keys are absent would be friendly but isn't a blocker.

### Suggested next steps

1. Rebase onto current `develop`.
2. Drop the `miner.py` hunk — the batched-upsert path already exists upstream with bounded sub-batches and a shared metadata helper.
3. Preserve `**ef_kwargs` when adding `_HNSW_BLOAT_GUARD` to the two `metadata={...}` call sites in `chroma.py`.
4. Add a test asserting the two HNSW keys land on the created collection's metadata.

The chroma.py change is the load-bearing fix and should ship.

funguf:
@igorls thank you for the careful review — agreed on every point, force-pushed accordingly:

1. **Rebased** onto current `develop` (`0d9929c`). PR is now `MERGEABLE`.
2. **Dropped the `miner.py` hunk entirely.** You're right that `a4868a3`'s bounded `DRAWER_UPSERT_BATCH_SIZE=1000` sub-batches with the shared `_build_drawer_metadata()` helper is strictly better than my single-unbounded-upsert. Re-inlining the metadata construction would have been a regression and the failure-isolation argument is convincing.
3. **`_HNSW_BLOAT_GUARD` merged into the metadata dict at both call sites**, alongside the existing `hnsw:space`, `hnsw:num_threads=1`, and `**ef_kwargs`. No silent drop of either the race fix or the embedding-function plumbing.
4. **Added two regression tests** (`test_chroma_backend_sets_hnsw_bloat_guard_on_creation` for the RFC 001 path, `test_chroma_backend_create_collection_sets_hnsw_bloat_guard` for the legacy `create_collection` path). Both assert `hnsw:batch_size` and `hnsw:sync_threshold` land on the persisted collection metadata, which doubles as a CI-time guard against the #1161 silent-drop concern. Full `tests/test_backends.py` suite still passes (34/34).
5. **No-op `try/except: raise` removed** — moot now that the miner.py hunk is gone.

Happy to add the one-time warning in `_get_collection()` for pre-fix palaces (when the keys are absent in already-existing collection metadata) as a follow-up; let me know if you'd rather it land here or as a separate PR.

funguf:
Self-review caught two more nits before this gets re-reviewed; folded them in (force-push, single commit `9a1e929`):

1. **`mempalace/mcp_server.py:227` was a third collection-creation call site bypassing `ChromaBackend`** and would have created palaces without the guard if the MCP server ran before any CLI mine. Imported `_HNSW_BLOAT_GUARD` and merged it into that `metadata=` dict alongside the existing `hnsw:space` and `hnsw:num_threads=1`. Verified all 67 tests in `tests/test_mcp_server.py` + `tests/test_mcp_stdio_protection.py` still pass, and added a smoke test confirming the metadata round-trips through `PersistentClient` reopen.

2. **The "immutable post-creation" migration claim was wrong on chromadb 1.5.x.** Verified empirically that `UpdateHNSWConfiguration` (defined at `chromadb/api/collection_configuration.py:460-466`) exposes `batch_size`, `sync_threshold`, `num_threads`, `ef_search`, `resize_factor` as updatable via `collection.modify(configuration={"hnsw":{...}})`. I'd cargo-culted the claim from #344's discussion. Reworded the PR body migration note and the `_HNSW_BLOAT_GUARD` docstring to acknowledge the retrofit path exists but isn't useful here (by the time `link_lists.bin` has bloated, the index is already corrupt and live retrofit won't recover it). Test docstring also softened.

Diff is small (`mempalace/backends/chroma.py | 8`, `mempalace/mcp_server.py | 13`, `tests/test_backends.py | 8`).

bensig (approved):
**Approve.** The empirical verification table makes this an easy yes — 30 GB → 376 MB on the same drawer count is the kind of result that's hard to argue with.

## What I checked

- **Diff is tight.** Three call sites get \`**_HNSW_BLOAT_GUARD\` merged into their collection-create metadata: \`ChromaBackend.get_collection(create=True)\`, \`ChromaBackend.create_collection\`, and \`mcp_server._get_collection\`. No drift between the three.
- **chromadb 1.5.x persistence.** The body claims \`hnsw:batch_size\` + \`hnsw:sync_threshold\` survive \`PersistentClient\` reopen (verified separately against #1161's concern). That's the right thing to verify; \`hnsw:num_threads\` was the one that didn't persist on 1.5.x and #976 had to add the explicit retrofit. Worth a comment in the constant block noting which keys are persistence-safe vs which need the retrofit.
- **Migration honesty.** The PR body says "users on already-bloated palaces still need to nuke and re-mine — there's no in-place fix." That's right: chromadb's HNSW config is immutable post-creation, and the rebuild path (#935 + #1210) handles the recovery. Worth a release-note line so users with pre-existing bloat understand why repair alone won't reclaim disk.

## Tests

\`pytest tests/test_backends.py tests/test_mcp_server.py -k "hnsw or bloat or guard or batch_size"\` — **8 passed** on this branch. The three new metadata-assertion tests (\`get_collection_includes_hnsw_bloat_guard\`, \`create_collection_includes_hnsw_bloat_guard\`, \`get_collection_persists_hnsw_metadata_through_reopen\`) cover the surface this PR changes.

## Two minor observations

1. **PR body has stale scope.** It mentions \`process_file\` batching changes, but those landed via [#1185](https://github.com/MemPalace/mempalace/pull/1185) and aren't in this diff. Worth trimming the body before merge so a reader doesn't search for the change and wonder why it isn't there.
2. **Two-key vs three-key set.** This PR's guard is \`{batch_size: 50_000, sync_threshold: 50_000}\`. #976's existing pin is \`{num_threads: 1}\`. Both end up in the same metadata dict via \`**_HNSW_BLOAT_GUARD\` and the explicit \`hnsw:num_threads: 1\` line. They don't conflict, but the dual sources for HNSW config are a small maintenance trap if someone later assumes everything HNSW lives in one constant. Could promote \`num_threads\` into \`_HNSW_BLOAT_GUARD\` as a follow-up.

Closes #344. Approving.

igorls:
Rebased onto current `develop` (88a53b2) and pushed. The conflict was in the `from .backends.chroma import` block in `mcp_server.py` — `_HNSW_BLOAT_GUARD` from this PR alongside `hnsw_capacity_status` from #1227 (merged 2026-04-26). Both symbols are now imported and used; no behaviour change.

Author preserved on the commit; committer is mine (force-pushed via maintainer-edit access). Tests: 1420 passed, 1 skipped on the rebased branch; `ruff check` + format clean.

Thanks for picking up this stalled rebase from #346 — the empirical 30 GB → 376 MB result is the kind of evidence that closes the loop on the bug class.

PR #1195, opened 2026-04-25
fix(graph): normalize wing slug at init so topic tunnels fire for hyphenated dirs (#1194)

bensig:
Fixes #1194.

## Problem

`mempalace init` recorded the `topics_by_wing` registry key under the **raw** directory name (e.g. `mempalace-public`), while `mempalace.yaml`'s `wing` field used the lower-cased + separator-collapsed slug (`mempalace_public`). At mine time the miner read the slug from the yaml and queried the registry under the wrong key, so `_compute_topic_tunnels_for_wing` returned `0` silently. Every project whose dirname contained a `-` or space lost every topic tunnel — the common shape in the wild.

Confirmed in a 5-project end-to-end repro (see #1194 for the table): expected 7 tunnels, got 2 (only the no-separator pair). Direct repro of the lookup miss:

\`\`\`python
>>> from mempalace.miner import get_topics_by_wing
>>> from mempalace.palace_graph import topic_tunnels_for_wing
>>> m = get_topics_by_wing()
>>> len(topic_tunnels_for_wing('mempal_private', m, min_count=1))   # what mine actually calls
0
>>> len(topic_tunnels_for_wing('mempal-private', m, min_count=1))   # what's in topics_by_wing
5
\`\`\`

## Fix

Extracted the normalization rule into `config.normalize_wing_name()` and routed both call sites through it:
- `mempalace/cli.py:146` — registry write at init time
- `mempalace/room_detector_local.py:307` — `mempalace.yaml` write

The two paths now produce the same key by construction.

## Tests

- `tests/test_config.py` — 4 unit tests for `normalize_wing_name` (hyphen, space, already-clean, mixed).
- `tests/test_cli.py::test_cmd_init_normalizes_wing_name_for_topics_registry` — regression test asserting `add_to_known_entities` is called with the normalized slug when the dirname is hyphenated. This test fails on develop without the fix.
- Full suite: `1310 passed` (+5 new).
- `ruff check` + `ruff format --check` clean for the changed files.

## Why the original tests missed it

`tests/test_palace_graph_tunnels.py` and `tests/test_miner.py` topic-tunnel fixtures use `wing_alpha`, `wing_beta`, `foo`, `bar` — names that look identical before and after normalization. Hyphenated names were never exercised at the init→mine seam.

cc @igorls

PR #1197, opened 2026-04-25
fix(tunnels): normalize wing names in topic tunnel lookup for hyphenated dirs

wahajahmed010:
### Root Cause
`init` stores wing names with hyphens/spaces replaced by underscores in `mempalace.yaml` (e.g. `mempalace_public`). But tunnel lookup functions (`find_tunnels`, `list_tunnels`, `follow_tunnels`) used the raw directory name (e.g. `mempalace-public`) as the lookup key, causing silent misses.

### Fix
- Added `_normalize_wing()` helper that lowercases and replaces hyphens/spaces with underscores, aligning lookup keys with stored metadata
- Applied normalization in `find_tunnels()`, `create_tunnel()`, `list_tunnels()`, and `follow_tunnels()`
- Added `logging.warning()` calls when tunnel lookups return empty results for filtered queries — no more silent failures
- Backward compatible: single-token wing names (no hyphens) are unaffected

### Testing
- 4 new tests in `TestHyphenatedWingNormalization`:
  - `test_list_tunnels_filters_hyphenated_wing` — verifies list_tunnels matches hyphenated input to stored underscore names
  - `test_follow_tunnels_matches_hyphenated_wing` — verifies follow_tunnels normalizes correctly
  - `test_create_tunnel_normalizes_wing_names` — verifies created tunnels use normalized names
  - `test_find_tunnels_warns_on_empty_result` — verifies warning log on empty results
- All 13 tests pass

### DIFF
--- a/mempalace/palace_graph.py
+++ b/mempalace/palace_graph.py
@@ -15,6 +15,11 @@ Enables queries like:
 No external graph DB needed — built from ChromaDB metadata.
 """
 
+# PEP 604 (``str | None``) needs 3.10+ at runtime; the project still
+# supports 3.9, so defer annotation evaluation to keep the union syntax
+# working on the older interpreter.
+from __future__ import annotations
+
 import hashlib
 import json
 import logging
--- a/mempalace/palace_graph.py
+++ b/mempalace/palace_graph.py
@@ -17,6 +17,7 @@ No external graph DB needed — built from ChromaDB metadata.
 
 import hashlib
 import json
+import logging
 import os
 import threading
 import time
@@ -26,6 +27,22 @@ from datetime import datetime, timezone
 from .config import MempalaceConfig
 from .palace import get_collection as _get_palace_collection
 from .palace import mine_lock
+
+logger = logging.getLogger("mempalace_graph")
+
+
+def _normalize_wing(wing: str | None) -> str | None:
+    """Normalize a wing name for consistent lookup.
+
+    ``init`` stores wing names with hyphens and spaces replaced by underscores
+    (e.g. ``mempalace_public``).  Callers that pass the raw directory name
+    (``mempalace-public``) would silently miss.  This helper aligns the lookup
+    key with the stored metadata.
+    """
+    if wing is None:
+        return None
+    return wing.lower().replace(" ", "_").replace("-", "_")
+
 
 # Module-level graph cache with TTL and write-invalidation.
 # Warm cache serves build_graph() in O(1); invalidate_graph_cache() clears on writes.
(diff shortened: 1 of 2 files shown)


bensig (approved):
**Approve.** Right shape for the lookup-side defense — small, focused, doesn't conflict with the init-side fix in [#1195](https://github.com/MemPalace/mempalace/pull/1195).

\`pytest tests/test_palace_graph_tunnels.py\` — **13 passed** on this branch.

## What landed

- \`_normalize_wing()\` helper in \`palace_graph.py\`
- \`find_tunnels()\` normalizes both query keys before lookup
- \`create_tunnel()\` normalizes both store keys on creation — important so all *new* writes go in normalized regardless of caller discipline
- Logger warning when a filter returns empty — actionable signal vs silent zero

## One coordination ask before / after merge

[#1195](https://github.com/MemPalace/mempalace/pull/1195) (mine, on the init side) introduces \`normalize_wing_name()\` in \`mempalace/config.py\` as the canonical implementation. The two helpers are byte-identical (\`name.lower().replace(" ", "_").replace("-", "_")\`).

If both PRs land, the cleanup is one line: replace this PR's \`_normalize_wing\` with \`from .config import normalize_wing_name\`. Same logic, single source of truth. Either:

- **Option A (preferred):** rebase this PR after #1195 merges, swap to importing the helper. Minor touch-up.
- **Option B:** land both as-is; I'll open a one-commit follow-up that consolidates the two helpers.

Either works — flagging so we don't drift if someone later changes one normalizer and not the other.

## Why both PRs should still land

The init-side fix prevents *new* bad keys from being written. The lookup-side fix here recovers gracefully from the bad keys that *already exist* in production palaces. Together they're complete defense; either one alone leaves a gap.

Ship it.

wahajahmed010:
Thanks for the thorough review! Going with **Option A** — I'll rebase after #1195 merges and swap `_normalize_wing` for `from .config import normalize_wing_name`. Makes sense to have a single source of truth for the normalizer. Appreciate the coordination heads-up.

igorls:
Rebased onto current `develop` (3474641) and pushed.

**Conflict resolved** in `tests/test_palace_graph_tunnels.py` — additive: kept this PR's `TestHyphenatedWingNormalization` class alongside develop's new `TestTopicTunnels` class from #1184. Both are independent and exercise different surfaces (lookup-side normalization vs. topic-tunnel computation), so they coexist with no logic interaction.

**Note for follow-up (after #1195 merges):** Igor's review comment on #1195 suggested folding this PR's `_normalize_wing` helper through `from .config import normalize_wing_name` once Ben's PR lands — single source of truth for the lower/replace rule across all four wing-slug producers. That swap is mechanical and a one-line change; can either land here as a follow-up commit after #1195 merges, or as a separate cleanup PR. Left as-is for now since `normalize_wing_name` doesn't exist on develop yet.

**Tests:** 1422 passed, 1 skipped on the rebased branch. `ruff check` + format clean. Author preserved on the original commit; committer is mine (force-pushed via maintainer-edit access).

Thanks for the lookup-side fix — together with #1195's producer-side normalization, this closes the hyphenated-wing tunnel loss for both new palaces (Ben's PR) and palaces with bad keys already on disk (this PR).

PR #1198, opened 2026-04-25
fix(searcher): tolerate None documents in BM25 reranker

jphein:
## Problem

`searcher._tokenize` calls `text.lower()` unconditionally. When ChromaDB returns a drawer with `documents` containing `None`, the hybrid-rerank path raises `AttributeError` mid-search, dropping the entire query response.

Observed in production on 2026-04-24 21:07:05:

```
File "mempalace/searcher.py", line 81, in _bm25_scores
    tokenized = [_tokenize(d) for d in documents]
File "mempalace/searcher.py", line 52, in _tokenize
    return _TOKEN_RE.findall(text.lower())
AttributeError: 'NoneType' object has no attribute 'lower'
```

This closes the gap left by #999's None-metadata audit, which fixed metadata read loops but did not extend to BM25 reranker helpers.

## Fix

`_tokenize` short-circuits to `[]` when `text` is falsy (`None` or `""`). A `None` doc gets BM25 score 0.0; the rest of the corpus reranks normally.

## Tests

Three regression tests in `TestBM25NoneSafety`:

1. `test_tokenize_handles_none` — direct unit guard
2. `test_tokenize_handles_empty_string` — covers the empty-string case at the same boundary
3. `test_bm25_scores_does_not_crash_on_none_documents` — load-bearing: mixed corpus with a `None` mid-list, asserts score 0.0 for that doc and finite scores for the rest. This is the exact production failure shape.

All existing searcher and hybrid_search tests still pass (`pytest tests/test_searcher.py tests/test_hybrid_search.py` → 30/30 against develop after cherry-pick).

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -46,7 +46,14 @@ def _first_or_empty(results, key: str) -> list:
 
 
 def _tokenize(text: str) -> list:
-    """Lowercase + strip to alphanumeric tokens of length ≥ 2."""
+    """Lowercase + strip to alphanumeric tokens of length ≥ 2.
+
+    Tolerates ``None`` documents — Chroma can return ``None`` in the
+    ``documents`` field for drawers without text content, which would
+    otherwise raise ``AttributeError`` mid-rerank.
+    """
+    if not text:
+        return []
     return _TOKEN_RE.findall(text.lower())
 
 
--- a/tests/test_searcher.py
+++ b/tests/test_searcher.py
@@ -121,6 +121,39 @@ class TestSearchMemories:
         assert none_hit["room"] == "unknown"
 
 
+# ── BM25 internals: None / empty document safety ─────────────────────
+
+
+class TestBM25NoneSafety:
+    """Regression tests for the AttributeError observed in production when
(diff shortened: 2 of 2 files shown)


bensig (approved):
**Approve, no asks.** Defensive single-line fix with a real production trace driving it.

## What landed

\`_tokenize\` now returns \`[]\` for falsy input instead of raising \`AttributeError\` on \`None.lower()\`. The trace in the test docstring is genuine (daemon log 2026-04-24 21:07:05) — Chroma was returning \`None\` for drawers without text content during a hybrid-rerank pass, and the BM25 path crashed mid-search.

## Test coverage

The new \`TestBM25NoneSafety\` class covers both the \`None\` and empty-string paths and includes the failure trace in the docstring so a future reader knows exactly what symptom this prevents. \`pytest tests/test_searcher.py\` — **27 passed** on this branch.

## Minor

Could also have hardened \`_bm25_scores\` upstream of \`_tokenize\` to filter \`None\` documents before iterating. The chosen approach (defend at the leaf function) is more defensive and catches any other call site that might pass \`None\` — better choice for a small surface like this.

Ship it.

jphein:

igorls (approved):
LGTM

PR #1199, opened 2026-04-25
fix(hooks): drop transcript-parent fallback that caused unbounded ingest

rmdes:
## Summary

The silent stop hook fires **two** mine subprocesses on every save trigger, and the second one re-ingests the entire growing Claude Code session JSONL on every fire. In a long-running project the resulting palace can balloon by orders of magnitude — observed in the wild at ~800 GB from a single project after a year of sessions, which filled a WSL VHD.

`_get_mine_dir` (used by both `_maybe_auto_ingest` and `_mine_sync`) fell back to the transcript's parent directory when `MEMPAL_DIR` was unset. For Claude Code that parent is `~/.claude/projects/-<project>/`, a folder of growing JSONL session files. Reading the hook log on a real session shows the redundancy clearly:

```
14:04 — first save  → +309 drawers (regular mine of ONE jsonl)
14:15 — second save → 0 drawers   (convo mode, correctly skipped)
14:15 — second save → +478 drawers (regular mine, RE-INGESTED same jsonl)
```

The second mine treats the JSONL as a project file, has no dedup against transcripts, and re-chunks the whole conversation history on every hook fire. As the session grows, each fire ingests more.

## Fix

`_get_mine_dir` now returns `""` unless `MEMPAL_DIR` is explicitly set to a real directory. `_maybe_auto_ingest` keeps its intended purpose — mining a separate project codebase the user has opted into via `MEMPAL_DIR`. Conversation mining still happens unchanged via `_ingest_transcript` (which uses `--mode convos` and dedups already-filed transcripts).

## Tests

- `test_get_mine_dir_no_transcript_fallback` — replaces `test_get_mine_dir_transcript_fallback`; asserts the new contract.
- `test_maybe_auto_ingest_no_mempal_dir_does_not_spawn` — replaces `test_maybe_auto_ingest_with_transcript`; asserts no subprocess spawn without `MEMPAL_DIR`.
- `test_precompact_no_sync_project_mine_without_mempal_dir` — replaces `test_precompact_mines_transcript_dir`; same fix path applied to the precompact hook.
- `test_stop_hook_does_not_double_mine_transcript` — new regression test at the `hook_stop` level. Asserts exactly one `subprocess.Popen` call (the convo-mode mine), and that the spawned command carries `--mode convos`.

## Test plan

- [x] `python -m pytest tests/ --ignore=tests/benchmarks` — 1306 passed
- [x] `ruff check mempalace/hooks_cli.py tests/test_hooks_cli.py` — clean
- [x] `ruff format --check` — clean
- [x] Verified the failing test path exactly reproduces the original bug (two Popen calls before fix → one after)
- [ ] CI green

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -198,14 +198,20 @@ def _output(data: dict):
 
 
 def _get_mine_dir(transcript_path: str = "") -> str:
-    """Determine directory to mine from MEMPAL_DIR or transcript path."""
+    """Resolve MEMPAL_DIR to a directory the project miner should ingest.
+
+    Returns the value of ``MEMPAL_DIR`` if it is set and points at an
+    existing directory, otherwise the empty string. ``transcript_path``
+    is accepted for signature stability but no longer used as a fallback:
+    the previous behaviour of mining the transcript's parent directory
+    (``~/.claude/projects/-<project>/``) caused the regular project
+    miner to re-chunk every JSONL session file on every hook fire, with
+    no dedup, since :func:`_ingest_transcript` already covers transcripts
+    in conversation mode. See PR #<this> for the incident.
+    """
     mempal_dir = os.environ.get("MEMPAL_DIR", "")
     if mempal_dir and os.path.isdir(mempal_dir):
         return mempal_dir
-    if transcript_path:
-        path = Path(transcript_path).expanduser()
-        if path.is_file():
-            return str(path.parent)
     return ""
 
 
--- a/tests/test_hooks_cli.py
+++ b/tests/test_hooks_cli.py
@@ -445,8 +445,14 @@ def test_maybe_auto_ingest_with_env(tmp_path):
                     mock_popen.assert_called_once()
 
 
-def test_maybe_auto_ingest_with_transcript(tmp_path):
-    """Falls back to transcript directory when MEMPAL_DIR is not set."""
+def test_maybe_auto_ingest_no_mempal_dir_does_not_spawn(tmp_path):
+    """Without MEMPAL_DIR set, _maybe_auto_ingest must NOT mine the transcript dir.
+
+    Regression: the previous transcript-parent fallback caused the regular
+    project miner to re-chunk Claude Code session JSONLs on every hook fire,
+    duplicating work already done by _ingest_transcript in convo mode and
+    producing unbounded palace growth.
+    """
     transcript = tmp_path / "t.jsonl"
     transcript.write_text("")
     with patch.dict("os.environ", {}, clear=True):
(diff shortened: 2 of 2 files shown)


jphein (commented):
Pulled the branch and ran `pytest tests/test_hooks_cli.py` — 67/67 pass on Python 3.12 + the fork's `chromadb 1.5.x` stack. The actual fix in `ae7b65e` is +16/-(small) lines in `hooks_cli.py` and is correct and well-scoped.

(The diff against current `develop` looks much larger because the branch is behind upstream's [#1210](https://github.com/MemPalace/mempalace/pull/1210) truncation-guard merge from yesterday — it'll narrow back to the hook fix once rebased.)

This composes cleanly with the fork's recent PreCompact incorporation (jphein/mempalace [`42817d7`](https://github.com/jphein/mempalace/commit/42817d7) — `hook_precompact` writes a session-recovery checkpoint via `_save_diary_direct` before transcript ingest + compaction). With your fix landed:

- `_save_diary_direct` keeps writing the checkpoint diary entry (fork-only routing to a dedicated `mempalace_session_recovery` collection for now)
- `_ingest_transcript` keeps mining transcripts in `--mode convos` with dedup — unchanged
- `_maybe_auto_ingest` and `_mine_sync` correctly become no-ops without `MEMPAL_DIR` — *your fix*

So the polluted-mega-wing failure mode from [#1083](https://github.com/MemPalace/mempalace/issues/1083) (raphaelsamy) and the unbounded-ingest from your bug report are both fully eliminated. We'd been working around #1083 with `PALACE_DAEMON_STRICT=1` (route hook writes through palace-daemon, skip local mines), but that's a heavier hammer than this single function-level change.

Worth rebasing against current develop so the diff reads as the +16-line fix it actually is. After that, +1 from this side.

Closes #1083 (would be worth referencing in the body alongside the unbounded-ingest case you describe).

igorls:
Thanks @rmdes! Closing as superseded — the transcript-parent fallback this PR removes was already eliminated on `develop` by:

- `eb4de04` fix(hooks): always mine the active transcript as convos, additive to MEMPAL_DIR
- `fe56797` fix(hooks): consolidate transcript ingest (#1231 review)

Current `mempalace/hooks_cli.py` uses `_get_mine_targets()` (~L200) which only returns MEMPAL_DIR-derived paths — no transcript-parent fallback. Both `_maybe_auto_ingest()` and `_mine_sync()` gate on this and become no-ops when MEMPAL_DIR is unset. Transcripts go exclusively through `_ingest_transcript()` in `--mode convos` with dedup. The mega-wing pollution failure mode is gone.

Appreciate the early diagnosis on this one — it informed the eventual fix in #1231.

Issue #1149, opened 2026-04-24
feat(init): optional LLM-assisted entity classification (phase 2)


igorls:
## Context

PR #1148 landed v1 of init entity detection: deterministic signal from package manifests (`package.json`, `pyproject.toml`, `Cargo.toml`, `go.mod`) + git authors, with an improved regex fallback for prose. That path is strong for codebases. It's still limited for prose-only corpora (notes, diaries, chat transcripts) where a short capitalized token can be a person, a place, a fictional character, or a common English word — regex alone cannot disambiguate.

This issue tracks phase 2: an **opt-in** LLM refinement step that takes the candidate set produced by phase 1 and reclassifies it with semantic understanding.

## Goals

- Ship as **opt-in only** — default remains zero-API / local-first.
- Default provider is **local** (Ollama), so the feature can work fully offline on a consumer machine.
- Pluggable provider interface so a user with an API key can swap in a hosted model.
- Run **interactively** during `mempalace init` with a visible progress indicator and clean cancellation (Ctrl-C → return partial results and proceed).
- Don't feed raw corpora to the model; feed regex-detected candidates + sampled context.
- Separately: deterministic parsing of `~/.claude/projects/-slug` directory names as an extra project-name source (no LLM required, but useful for conversation-heavy users).

## CLI shape

```
mempalace init <dir>                                        # unchanged, no LLM
mempalace init <dir> --llm                                  # opt-in, use configured default
mempalace init <dir> --llm-provider ollama --llm-model gemma3:4b
mempalace init <dir> --llm-provider openai-compat \
                     --llm-endpoint http://localhost:1234/v1 \
                     --llm-model <any>
mempalace init <dir> --llm-provider anthropic --llm-model claude-haiku-4-5
```

Persist defaults via `mempalace config set llm.provider ollama` / `llm.model ...`. `MEMPALACE_LLM_*` env vars override.

## Architecture

New `discover_entities(dir, llm=None)` layering:

```
1. project_scanner         — manifests + git (fast, deterministic)
2. convo_scanner (new)     — parse .claude/projects/-slug dirnames
3. regex detector          — prose fallback (fast, noisy)
4. llm_refine (if --llm)   — sample corpus, classify candidates, merge
```

LLM step specifically:
- Take top N regex-detected candidates (with their current type + confidence).
- For each, collect 3–5 sample context lines from the source corpus.
- Send as one batch per 30–50 candidates to stay within small-model context.
[...]

PR #1200, opened 2026-04-25
feat(llm): add claude-code provider (#1193)

### LINKED ISSUE #1193: feat(llm): add claude-code provider for Claude Pro/Max subscription users
## Summary

Add `claude-code` as an LLM provider option, allowing users with Claude Pro/Max subscriptions to use their subscription for LLM-assisted features (rerank, entity refinement) without requiring a separate API key.

## Motivation

Currently, MemPalace supports three LLM providers (`llm_client.py`):
- `ollama` — local models (default)
- `openai-compat` — OpenAI-compatible endpoints
- `anthropic` — requires API key

Users who already pay for Claude Pro/Max subscriptions cannot leverage their existing authentication for MemPalace's LLM features. They must either:
1. Set up Ollama locally
2. Pay for a separate Anthropic API key
3. Use a third-party OpenAI-compatible service

## Proposed Solution

Add a `claude-code` provider that uses the **Claude Agent SDK** (`claude-agent-sdk`) to make LLM calls through the user's existing Claude CLI authentication (`claude auth login`).

## Reference Implementation

**Hindsight** has already implemented this pattern successfully:
- **File**: [`hindsight-api-slim/hindsight_api/engine/providers/claude_code_llm.py`](https://github.com/vectorize-io/hindsight/blob/main/hindsight-api-slim/hindsight_api/engine/providers/claude_code_llm.py)
- **Approach**: Uses `claude_agent_sdk.query()` for simple calls and `ClaudeSDKClient` for tool calling
- **Auth**: Leverages existing `claude auth login` credentials — no API key needed
- **Fallback**: Clear error messages guide users to authenticate if not logged in

Key implementation details from Hindsight:
```python
from claude_agent_sdk import query, ClaudeAgentOptions

# Simple call
async for message in query(prompt=user_content, options=options):
    # process response

# With tools
async with ClaudeSDKClient(options=options) as client:
    await client.query(user_content)
    async for message in client.receive_response():
        # process response
```

## Use Cases

1. **Rerank (Hybrid + LLM)**: Users can get ≥99% R@5 using their Claude subscription instead of 98.4% with Hybrid v4 alone
2. **Entity refinement** (`mempalace init --llm`): Better entity classification without Ollama setup
3. **Future LLM features**: Any upcoming LLM-assisted functionality

## Implementation Scope

1. Add `ClaudeCodeProvider` class to `mempalace/llm_client.py`
2. Add `claude-code` to provider selection logic
3. Add dependency: `claude-agent-sdk` (optional, only needed if using this provider)
4. Update CLI flags documentation

## Additional Context

- Claude Agent SDK: `pip install claude-agent-sdk`
- Requires user to run `claude auth login` once
- No API key management needed
- Works with Claude Pro ($20/mo) and Max ($100/mo) subscriptions

mvalentsev:
## Summary

Adds a `claude-code` LLM provider in `mempalace/llm_client.py` that routes through the local `claude` CLI binary using the user's Claude Pro/Max subscription via `claude auth login`. Mirrors the existing `OllamaProvider` / `OpenAICompatProvider` / `AnthropicProvider` shape so `mempalace init --llm --llm-provider claude-code --llm-model claude-haiku-4-5` works with no API key.

Closes #1193.

## How it works

```
claude -p \
    --no-session-persistence \
    --output-format json \
    --model <model>
```

System prompt + user prompt go through **stdin**, framed with XML-like tags:

```
<system>
{system_prompt + JSON-only directive when json_mode=True}
</system>
<user>
{user_prompt}
</user>
```

`cwd=tempfile.gettempdir()` so claude does not pick up a project-level `CLAUDE.md`. Auth flows through `claude auth login` (OAuth / keychain). The resolved binary path comes from `shutil.which("claude")` (not the bare `"claude"` literal) so a PATH change between `check_available()` and `classify()` does not introduce a TOCTOU.

`check_available()` runs `claude auth status --text` and surfaces a friendly error pointing at `claude auth login` if not authenticated.

## Hardening

- **System prompt off argv.** The first version passed the system prompt via `--system-prompt <text>`. argv is visible to other local users via `ps` / `/proc/*/cmdline`, and the prompt can carry sensitive context (entity names, project paths). Moving system content to stdin keeps that surface zero. The `<system>/<user>` XML framing replaces literal `SYSTEM:` / `USER:` markers a malicious drawer could spoof to inject a second system prompt.
- **Credential env scrub.** `_subprocess_env()` strips `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN` before spawning `claude -p`. If the user has either exported in their shell, the CLI may fall back to API-key auth and bill the API account instead of the subscription this provider is built around. Removing the credentials forces OAuth / keychain auth, which is the documented path. Configuration vars (`ANTHROPIC_BASE_URL` for corporate-proxy / internal-gateway routing, etc.) pass through so users behind a custom endpoint keep working.
- **External-service flag.** `is_external_service` is overridden to `True`. The base class default is URL-based and would treat this provider as local because `endpoint=None`, but every classify call routes user content to Anthropic-hosted models. The override makes the privacy-warning gate (#1224) fire for `claude-code` the same way it does for the `anthropic` provider.
- **Friendly UX hint** when `--llm-provider claude-code` is paired with the Ollama-shaped default `--llm-model gemma4:e4b`. `cmd_init` prints a one-line stderr nudge pointing at `claude-haiku-4-5` and continues -- the hint is informational, not blocking.

## Why subprocess and not `claude-agent-sdk`

The Anthropic-published Python SDK was the obvious alternative; subprocess won on every axis for our use case:

| Criterion | `subprocess.run(['claude', '-p', ...])` | `claude-agent-sdk` |
|---|---|---|
| Pip dependency | none (stdlib) | `claude-agent-sdk` + `anyio` |
| Python floor | matches current `>=3.9` | bumps to `>=3.10` |
| API surface | sync, matches existing providers | async-only, needs asyncio bridge |
| Auth path | `claude auth login` (CLI keychain) | same (SDK delegates to the CLI) |
| Maintenance | pin claude CLI flags | pin SDK API + CLI flags |

The SDK is a thin wrapper around the same `claude` binary the user already has. Going direct keeps `llm_client.py`'s zero-SDK style intact and does not raise the Python floor.

## Why `--bare` is NOT used

`claude --bare` would skip hooks, plugins, and CLAUDE.md auto-discovery for clean isolation. From `claude --help`:

> `--bare`: ... Anthropic auth is strictly ANTHROPIC_API_KEY or apiKeyHelper via --settings (OAuth and keychain are never read).

That defeats the point of a subscription provider. We omit it and reduce ambient noise via `cwd=tempfile.gettempdir()` and `--no-session-persistence` instead.

## Subscription policy fragility

This provider is fully opt-in: `--llm` is opt-in, and within that `--llm-provider claude-code` is opt-in. Default `init` path remains zero-API.

Anthropic blocked OAuth-token-replay through third-party harnesses on April 4, 2026; `claude -p` invocation from third-party tools was subsequently sanctioned for first-party CLI binaries. That sanction may change. If it does, `check_available()` will return `(False, ...)` from the post-policy `claude auth status` failure, surfacing a clear error before any classify call. Existing `llm_refine.py` callers can fall through to a different provider.

Documented in the provider docstring so future readers know this path is best-effort.

## Failure modes

All raise `LLMError` like the other providers:

- `claude` binary missing -> `check_available()` returns `(False, "not found in PATH")`; `classify()` raises `LLMError("`claude` CLI not found in PATH")` if the binary disappears between provider construction and classify
- Not logged in -> `check_available()` returns `(False, "Run `claude auth login`...")`
- `claude -p` timeout -> `LLMError("`claude -p` timed out after Ns")`
- Spawn failure (`OSError`) -> `LLMError("failed to spawn")`
- Non-zero exit -> `LLMError` with `stderr[:500]`
- Malformed JSON envelope -> `LLMError("non-JSON envelope")` with `stdout[:200]` excerpt for debugging
- Empty `result` field -> `LLMError("empty result")`

## Tests

16 new tests in `tests/test_llm_client.py`: 1 factory-dispatch test, 13 unit tests covering `check_available()` and `classify()` paths (mocking `subprocess.run` and/or `shutil.which`) plus the hardening hooks (env strip, `is_external_service`, binary missing), and 1 gated integration test `test_claude_code_real_invocation` that runs a live `claude -p` round-trip when `MEMPAL_TEST_CLAUDE_CLI=1` is set (skipped by default; CI has no authenticated user). `tests/test_cli.py` adds 2 tests pinning the `gemma4:e4b`-with-claude-code stderr hint behaviour. Local pytest: 108 passed + 1 skipped, ruff clean.

## Out of scope

- Benchmark scripts (`benchmarks/longmemeval_bench.py`, `benchmarks/locomo_bench.py`) `--llm-backend claude-code`. Benchmarks are excluded from the package tests and have their own argparse layer (`--llm-backend [anthropic, ollama]`); plumbing claude-code through the rerank/refine call sites is a parallel concern. Happy to do it as a follow-up if useful.
- `mempalace config set llm.provider ...` persistence (planned under #1149 follow-ups).
- Optimizing CLAUDE.md auto-discovery overhead. First call pays a one-time cache-miss; subsequent calls hit Anthropic's prompt cache and are cheap. If real users hit this on tiny corpora we can revisit.

---

_Body reflects the landed code: the system prompt goes to stdin (not `--system-prompt` argv), framed with `<system>/<user>` XML; `ANTHROPIC_*` credential env vars are stripped before spawning; `is_external_service` is overridden to `True` so the privacy gate fires; the env strip is narrowed to `ANTHROPIC_API_KEY` + `ANTHROPIC_AUTH_TOKEN` only (configuration vars like `ANTHROPIC_BASE_URL` pass through for corporate-proxy users); and `cmd_init` prints a hint when `claude-code` is paired with the Ollama default model._

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -308,6 +308,18 @@ def cmd_init(args):
     if not getattr(args, "no_llm", False):
         provider_name = getattr(args, "llm_provider", "ollama") or "ollama"
         provider_model = getattr(args, "llm_model", "gemma4:e4b") or "gemma4:e4b"
+        # Friendly hint when claude-code is paired with the Ollama-shaped
+        # default model. claude -p rejects "gemma4:e4b" as an unknown
+        # Anthropic model with a confusing error; surface what the user
+        # likely meant and continue (they may have set the same value
+        # explicitly on purpose).
+        if provider_name == "claude-code" and provider_model == "gemma4:e4b":
+            print(
+                "  Hint: --llm-model defaults to 'gemma4:e4b' (Ollama). "
+                "claude-code expects an Anthropic model name like "
+                "'claude-haiku-4-5'. Pass --llm-model claude-haiku-4-5 to silence.",
+                file=sys.stderr,
+            )
         try:
             candidate = get_provider(
                 name=provider_name,
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1804,20 +1804,30 @@ def main():
     p_init.add_argument(
         "--llm-provider",
         default="ollama",
-        choices=["ollama", "openai-compat", "anthropic"],
-        help="LLM provider (default: ollama). Pass --no-llm to disable LLM-assisted refinement entirely.",
+        choices=["ollama", "openai-compat", "anthropic", "claude-code"],
+        help=(
+            "LLM provider (default: ollama). Pass --no-llm to disable LLM-assisted "
+            "refinement entirely. claude-code routes through the local `claude` CLI "
+            "using your Claude Pro/Max subscription (run `claude auth login` first); "
+            "no API key needed."
+        ),
     )
     p_init.add_argument(
         "--llm-model",
         default="gemma4:e4b",
-        help="Model name for the chosen provider (default: gemma4:e4b for Ollama).",
+        help=(
+            "Model name for the chosen provider (default: gemma4:e4b for Ollama). "
+            "For claude-code, pass an Anthropic model name "
+            "such as claude-haiku-4-5."
+        ),
     )
     p_init.add_argument(
         "--llm-endpoint",
         default=None,
         help=(
             "Provider endpoint URL. Default for Ollama: http://localhost:11434. "
-            "Required for openai-compat."
+            "Required for openai-compat. Ignored for claude-code (auth comes "
+            "from the local CLI)."
         ),
     )
     p_init.add_argument(
@@ -1825,7 +1835,8 @@ def main():
         default=None,
         help=(
             "API key for the provider. For anthropic, defaults to $ANTHROPIC_API_KEY; "
-            "for openai-compat, defaults to $OPENAI_API_KEY."
+            "for openai-compat, defaults to $OPENAI_API_KEY. "
+            "Ignored for claude-code (auth comes from `claude auth login`)."
         ),
     )
     p_init.add_argument(
--- a/mempalace/llm_client.py
+++ b/mempalace/llm_client.py
@@ -485,13 +485,21 @@ class ClaudeCodeProvider(LLMProvider):
         # Override the URL-based default in the base class.
         return True
 
+    # Only the credential-bearing ANTHROPIC_* vars get stripped. Configuration
+    # vars like ANTHROPIC_BASE_URL (corporate proxy / custom endpoint) must
+    # pass through so users behind an internal gateway keep working. A broad
+    # prefix-strip would break those deployments without warning.
+    _SUBPROCESS_ENV_STRIP = ("ANTHROPIC_API_KEY", "ANTHROPIC_AUTH_TOKEN")
+
     def _subprocess_env(self) -> dict:
-        # Strip ANTHROPIC_* env vars before spawning `claude -p`. If the user
-        # has ANTHROPIC_API_KEY exported in their shell, the CLI may fall
-        # back to API-key auth and bill the API account instead of the
-        # subscription this provider is built around. Removing the vars
-        # forces OAuth / keychain auth, which is the documented path.
-        return {k: v for k, v in os.environ.items() if not k.upper().startswith("ANTHROPIC_")}
+        # Strip credential env vars before spawning `claude -p`. If the user
+        # has ANTHROPIC_API_KEY or ANTHROPIC_AUTH_TOKEN exported in their
+        # shell, the CLI may fall back to API-key auth and bill the API
+        # account instead of the subscription this provider is built
+        # around. Removing them forces OAuth / keychain auth, which is the
+        # documented path. Other ANTHROPIC_* (BASE_URL, etc.) are passed
+        # through.
+        return {k: v for k, v in os.environ.items() if k.upper() not in self._SUBPROCESS_ENV_STRIP}
 
     def check_available(self) -> tuple[bool, str]:
         binary = shutil.which("claude")
--- a/mempalace/llm_client.py
+++ b/mempalace/llm_client.py
@@ -1,7 +1,7 @@
 """
 llm_client.py — Minimal provider abstraction for LLM-assisted entity refinement.
 
-Three providers cover the useful space:
+Four providers cover the useful space:
 
 - ``ollama`` (default): local models via http://localhost:11434. Works fully
   offline. Honors MemPalace's "zero-API required" principle.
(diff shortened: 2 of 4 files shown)


igorls:
Thanks for the thorough write-up, @mvalentsev, but unfortunately we can't merge this yet. 

The current Claude Code legal page (https://code.claude.com/docs/en/legal-and-compliance) reads:

> OAuth authentication is intended exclusively for purchasers of Claude Free, Pro, Max, Team, and Enterprise subscription plans and is designed to support ordinary use of Claude Code and other native Anthropic applications.

> Developers building products or services that interact with Claude's capabilities, including those using the Agent SDK, should use API key authentication through Claude Console or a supported cloud provider. Anthropic does not permit third-party developers to offer Claude.ai login or to route requests through Free, Pro, or Max plan credentials on behalf of their users.

> Anthropic reserves the right to take measures to enforce these restrictions and may do so without prior notice.

`ClaudeCodeProvider` routes user classify calls through Pro/Max credentials via claude auth login — the subprocess wrapper around `claude -p` doesn't change the underlying pattern, and the Agent SDK (also a wrapper around the same binary) is named explicitly. If we ship this and Anthropic enforces, users following our README could have their subscriptions actioned, with MemPalace as the cause.

Users with an `ANTHROPIC_API_KEY` are already covered by the existing anthropic provider.

Leaving the PR open so the discussion stays visible. Happy to revisit if Anthropic publishes guidance that permits this.

mvalentsev:
@igorls fair concern. The frame I'm working from is OpenClaw's own published stance. Their Anthropic provider docs (https://docs.openclaw.ai/providers/anthropic) state, verbatim:

> Anthropic staff told us OpenClaw-style Claude CLI usage is allowed again, so OpenClaw treats Claude CLI reuse and `claude -p` usage as sanctioned unless Anthropic publishes a new policy.

That's the most-watched third-party Anthropic harness, post-April-4 block (openclaw/openclaw#63316), publishing direct guidance from Anthropic that user-local `claude -p` subprocess invocation is not the prohibited route. Their `claude-cli` provider has been in production since openclaw/openclaw#61160 and is actively maintained -- merged: openclaw#69179, #69211, #70902; open: #71332, #70863, #68682, #66819, #68388.

This PR's `ClaudeCodeProvider` is structurally identical -- spawn user's logged-in `claude` binary, no token extraction, no direct API replay. Under OpenClaw's reading of Anthropic guidance, the Agent SDK clause in the legal page is read narrowly: SDK is a wrapper around `claude` running under the user's own login, and the "route requests on behalf of their users" prohibition targets server-side third-party services, not user-local subprocess invocation.

OpenClaw's own docs do note that for long-lived gateway hosts, API keys remain "the clearest and most predictable production path" -- fair, and `mempalace init --llm` is a one-shot local invocation, not a long-lived gateway.

Your call.

igorls:
The usage from my perspective is different, agentic usage is sparse and with varied amount of tokens, which is in some ways similar to a human prompting an LLM, MemPalace using Haiku for example for entity extraction is a repeated mass automation usage. This is where I think the issue is.

Qodo-Free-For-OSS:
Hi, ClaudeCodeProvider.classify passes the full system prompt as a command-line argument (`--system-prompt <text>`), which exposes prompt content to local process listings and logs. This can leak sensitive instructions/context to other local users on the same machine.

**Severity:** action required | **Category:** security

**How to fix:** Move system prompt off argv

> ### Issue description
> `ClaudeCodeProvider.classify()` passes the full system prompt via the `--system-prompt` argv parameter, which can leak prompt contents via process listings. We need to avoid placing potentially sensitive prompt text in argv.
> 
> ### Issue Context
> - `user` is already passed via stdin (good), but `sys_prompt` is exposed in argv.
> - The `llm_refine` pipeline only needs the model to follow instructions; strict “system” separation is less important than preventing local leakage.
> 
> ### Fix Focus Areas
> - mempalace/llm_client.py[350-379]
> 
> ### Implementation notes
> - Prefer an approach that does not include the system prompt in argv:
>   - Option A: stop using `--system-prompt` and instead prepend the system instructions to stdin input (e.g., `input = f"SYSTEM:\n{sys_prompt}\n\nUSER:\n{user}"`).
>   - Option B: if the `claude` CLI supports reading system prompt from a file or stdin, use that mechanism (e.g., write to a temp file with restrictive permissions and pass only the filename in argv).
> - Add/adjust a unit test to assert `sys_prompt` is not present in `captured["cmd"]`.

---

mvalentsev:
Quick note on policy: Anthropic published updated guidance on Agent SDK / `claude -p` use with Claude subscriptions: https://support.claude.com/en/articles/15036540-use-the-claude-agent-sdk-with-your-claude-plan (effective 2026-06-15).

Two relevant points for this PR:

1. `claude -p` (non-interactive mode) is explicitly listed in the covered scope, alongside the Agent SDK and GitHub Actions integration.
2. Subscriptions get a separate per-user monthly Agent SDK credit ($20 Pro / $100 Max-5x / $200 Max-20x). Beyond the credit: optional overflow at standard API rates if extra usage is enabled, otherwise a hard stop.

The page steers shared production automation toward API keys; individual experimentation and automation stays in scope. The credit also creates a natural ceiling for high-volume use: a user running mining at scale hits the budget and switches to `--llm-provider anthropic` with an API key, which keeps the "API key for sustained automation" guidance intact.

Issue #1203, opened 2026-04-25
benchmarks: LongMemEval runner opens JSON with platform default encoding on Windows


mschultheiss83:
**What happened?**

Running `benchmarks/longmemeval_bench.py` on native Windows failed before benchmark execution unless UTF-8 mode was forced.

Reproduction on 2026-04-25:

- OS: Windows 11
- Python: 3.11.9
- Install: fresh venv, `pip install -e ".[dev]"`
- Dataset: `longmemeval_s_cleaned.json` downloaded from the benchmark docs
- Command: `python benchmarks/longmemeval_bench.py benchmarks-data\\longmemeval_s_cleaned.json --limit 20`

Observed failure:

```text
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 107102: character maps to <undefined>
```

The script opens the dataset with the platform default encoding (`with open(data_file) as f:`), which is `cp1252` on this machine. The run succeeds if UTF-8 mode is forced explicitly:

```powershell
$env:PYTHONUTF8='1'
python benchmarks\\longmemeval_bench.py benchmarks-data\\longmemeval_s_cleaned.json --limit 20
```

and the full 500-question run also completes successfully with the same workaround.

**What did you expect?**

Benchmark runners should read their JSON inputs with explicit UTF-8 so they work on Windows without requiring `PYTHONUTF8=1` or a UTF-8 system locale.

**How to reproduce**

1. Use native Windows with a non-UTF-8 default text encoding.
2. Download `longmemeval_s_cleaned.json` as documented.
3. Run `python benchmarks/longmemeval_bench.py <path-to-json> --limit 20`.
4. Observe `UnicodeDecodeError` before benchmark execution.

**Likely fix**

Audit benchmark runners and switch JSON/text file reads for benchmark datasets and caches to explicit `encoding="utf-8"` (with `errors="replace"` only where lossy behavior is acceptable).

mschultheiss83:
closing see https://github.com/MemPalace/mempalace/pull/1382

PR #1204, opened 2026-04-25
benchmarks: harden longmemeval runner for Windows encoding

mschultheiss83:
## Summary

Narrow Windows fix for the LongMemEval benchmark runner.

This PR addresses two reproducible failures in `benchmarks/longmemeval_bench.py` on native Windows:

- local benchmark files were opened with the platform default encoding instead of explicit UTF-8
- benchmark console output used non-ASCII separator characters that could raise `UnicodeEncodeError` on a default `cp1252` console

## Changes

- make LongMemEval dataset / split / cache / result file I/O explicit UTF-8
- replace non-ASCII console separator/output characters in the LongMemEval runner with ASCII-safe equivalents
- add a focused regression test for Windows default-encoding behavior in `tests/benchmarks/test_longmemeval_bench.py`

## Reproduction

Before this change on Windows:

```bash
python benchmarks/longmemeval_bench.py /path/to/longmemeval_s_cleaned.json --limit 20
```

could fail with a `UnicodeDecodeError` while reading the dataset, and after fixing file I/O the same runner could still fail with `UnicodeEncodeError` when printing separators to a non-UTF-8 console.

After this change, the runner completes on the same machine without forcing UTF-8 mode.

## Validation

Ran:

```bash
python -m pytest tests/benchmarks/test_longmemeval_bench.py -q --basetemp .pytest-tmp
```

Result:

```text
2 passed
```

Ran:

```bash
python benchmarks/longmemeval_bench.py benchmarks-data/longmemeval_s_cleaned.json --limit 1
```

Result:
- completed successfully on native Windows
- wrote a normal benchmark results file

## Scope / Notes

- intentionally scoped only to `benchmarks/longmemeval_bench.py`
- I have not claimed broader Windows benchmark hardening in other runners
- I have not re-run the full project test suite here because there are existing unrelated Windows failures outside this runner path
- related issue: #1203

### DIFF
--- a/benchmarks/longmemeval_bench.py
+++ b/benchmarks/longmemeval_bench.py
@@ -2917,7 +2917,7 @@ def _load_or_create_split(split_file: str, data: list, dev_size: int = 50, seed:
 
     split_path = Path(split_file)
     if split_path.exists():
-        with open(split_path) as f:
+        with open(split_path, encoding="utf-8") as f:
             return json.load(f)
 
     # Create new split
@@ -2927,9 +2927,9 @@ def _load_or_create_split(split_file: str, data: list, dev_size: int = 50, seed:
     dev_ids = all_ids[:dev_size]
     held_out_ids = all_ids[dev_size:]
     split = {"dev": dev_ids, "held_out": held_out_ids, "seed": seed, "dev_size": dev_size}
-    with open(split_path, "w") as f:
+    with open(split_path, "w", encoding="utf-8") as f:
         json.dump(split, f, indent=2)
-    print(f"  Created new split: {len(dev_ids)} dev / {len(held_out_ids)} held-out → {split_path}")
+    print(f"  Created new split: {len(dev_ids)} dev / {len(held_out_ids)} held-out -> {split_path}")
     return split
 
 
@@ -2957,7 +2957,7 @@ def run_benchmark(
     split_subset: "dev" (50 questions for tuning) or "held_out" (450 for final evaluation).
                   None = run all questions.
     """
-    with open(data_file) as f:
+    with open(data_file, encoding="utf-8") as f:
         data = json.load(f)
 
     # Apply train/test split filter before limit/skip
@@ -2966,7 +2966,7 @@ def run_benchmark(
         subset_ids = set(split[split_subset])
         before = len(data)
         data = [entry for entry in data if entry["question_id"] in subset_ids]
-        print(f"  Split filter ({split_subset}): {before} → {len(data)} questions")
+        print(f"  Split filter ({split_subset}): {before} -> {len(data)} questions")
 
     if limit > 0:
         data = data[:limit]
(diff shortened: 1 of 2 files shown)


mvalentsev:
Nice fix on both layers. The same dual pattern (bare `open(...)` plus non-ASCII chars in `print` / docstrings) shows up in the other three benchmark runners on develop:

- `benchmarks/locomo_bench.py`: 4 file-IO sites (lines 639, 664, 706, 980)
- `benchmarks/membench_bench.py`: 2 file-IO sites (207, 422)
- `benchmarks/convomem_bench.py`: 5 file-IO sites (73, 79, 92, 103, 305)

The issue framing ("audit benchmark runners", plural) puts them in scope, FWIW.

jphein:
Audited the other benchmark runners for the same `open(...)` pattern in case it's useful for scoping:

```
benchmarks/membench_bench.py:207   open(fpath)              # read
benchmarks/membench_bench.py:422   open(out_file, "w")      # write
benchmarks/convomem_bench.py:73    open(cache_path)         # read
benchmarks/convomem_bench.py:79    open(cache_path)         # read
benchmarks/convomem_bench.py:92    open(cache_path)         # read
benchmarks/convomem_bench.py:103   open(cache_path, "w")    # write
benchmarks/convomem_bench.py:305   open(out_file, "w")      # write
```

Same failure shape on any non-ASCII content. Up to you whether to expand scope here or leave them for a follow-up — both are reasonable, and the LongMemEval fix is a clean unit on its own. Thanks for the explicit-UTF-8 + ASCII-console-separator hardening; that's the right pattern.

mschultheiss83:
Thanks, this is helpful.

I intentionally kept this PR narrow to the reproducible LongMemEval Windows failure from #1203 so the change stayed easy to review and validate.

I agree the same pattern exists in `locomo_bench.py`, `membench_bench.py`, and `convomem_bench.py`. Would you prefer that I expand this PR to cover the other benchmark runners as well, or keep this one scoped to LongMemEval and send a follow-up PR for the rest?

jphein:
Follow-up PR is the right call — keeps this one focused on the reproducible failure you opened it for, and the other three runners can land separately without complicating review here.

If it's useful, I can take a swing at the follow-up since the audit was already done — but happy to leave it to you if you'd rather keep ownership of the benchmark hardening thread. Either works.

mschultheiss83:
Thanks, @jphein — agree a follow-up PR is the right call.

If you’re up for it, please take a swing at the follow-up and own the changes for `benchmarks/locomo_bench.py`, `benchmarks/membench_bench.py`, and `benchmarks/convomem_bench.py` (same explicit `encoding="utf-8"` pattern + any non-ASCII-safe console output tweaks as needed). I’ll keep this PR focused on the reproducible LongMemEval Windows failure from #1203.

jphein:
Filed [#1382](https://github.com/MemPalace/mempalace/pull/1382) covering `locomo_bench.py`, `membench_bench.py`, `convomem_bench.py` per the audit. Same pattern: explicit `encoding="utf-8"` on every local `open(...)`, ASCII separators (`-` and `->`) in `print(...)` chrome. Total +16/-16 across the three files; comments/docstrings untouched. Thanks for the reminder — sorry it took a beat.

PR #1167, opened 2026-04-24
fix(kg): validate ISO-8601 date formats at MCP boundary

arnoldwender:
## What and Why

`tool_kg_query` (`as_of`), `tool_kg_add` (`valid_from`), and `tool_kg_invalidate` (`ended`) accepted any string and forwarded it to SQLite without format validation. Parameterized queries prevent SQL injection, but invalid date strings silently produce empty result sets — callers cannot distinguish "no fact at this time" from "your date format was unrecognized." This is especially painful for natural-language LLM callers that synthesize dates like `"March 2026"` or `"Jan 2025"`.

## Root Cause

`mempalace/mcp_server.py:843-897` — the three kg tool wrappers sanitize subject/predicate/object but pass temporal parameters straight through to `_kg`.

## Change Summary

- `mempalace/config.py` — new `sanitize_iso_date()` validator alongside the other input sanitizers. Accepts `YYYY`, `YYYY-MM`, and `YYYY-MM-DD`; passes through `None` / `""`; raises `ValueError` with a field-named message on anything else.
- `mempalace/mcp_server.py` — call `sanitize_iso_date(...)` in `tool_kg_query`, `tool_kg_add`, and `tool_kg_invalidate` before values reach the storage layer.
- `tests/test_config.py` — 13 unit tests for `sanitize_iso_date` (accepted forms, passthrough, whitespace, rejection of natural-language / US format / invalid month/day / non-string).
- `tests/test_mcp_server.py` — 4 integration tests at the MCP boundary: rejection of `"Jan 2025"` / `"March 2026"` / `"yesterday"`, and acceptance of partial ISO forms.

## Test Plan

- [x] `pytest tests/test_config.py tests/test_mcp_server.py::TestKGTools -v` — all green (9/9 kg + 13/13 date validator tests)
- [x] `ruff check` — clean
- [x] `ruff format --check` — formatted

Closes #1164

PR #1214, opened 2026-04-26
fix(kg): reject inverted intervals in add_triple (valid_to < valid_from)

arnoldwender:
## What and Why

A triple with `valid_to` before `valid_from` satisfies neither of the temporal filter clauses in `KnowledgeGraph.query_entity()`:

```sql
valid_from <= as_of AND valid_to >= as_of
```

so the triple is invisible to every query — silently corrupt. The data lives in SQLite forever but never surfaces, and the caller never sees a warning. This is a P0 data-integrity bug in a path adapters can hit easily (any caller that mixes up the two date params).

## Root Cause

`mempalace/knowledge_graph.py:149` — `add_triple()` accepts `valid_from` and `valid_to` as opaque strings, validates each format independently in upstream callers (and now PR #1167 at the MCP boundary), but never checks the relationship between them.

## Fix

Reject at write time with a clear ValueError naming both bounds. The guard fires only when both are set:

- Open intervals (only `valid_from`, or only `valid_to`) — accepted unchanged.
- Same-day intervals (`valid_from == valid_to`, point-in-time facts) — explicitly allowed (`<` not `<=`).
- String comparison is correct because `sanitize_iso_date` (PR #1167) ensures `YYYY-MM-DD` lex order matches calendar order.

## Test plan

- `test_add_triple_rejects_inverted_interval` — asserts ValueError with `'before valid_from'` match
- `test_add_triple_accepts_equal_dates` — point-in-time facts pass
- `test_add_triple_allows_only_one_bound` — open intervals (one bound) still pass
- All 24 `test_knowledge_graph.py` tests green
- Full suite: 1318 passed locally (2 pre-existing failures in `test_backends.py::test_pin_hnsw_threads*` unrelated — environmental ChromaDB `configuration_json["hnsw"]` schema drift, also fails on a clean `develop` checkout)

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -171,6 +171,15 @@ class KnowledgeGraph:
             add_triple("Max", "does", "swimming", valid_from="2025-01-01")
             add_triple("Alice", "worried_about", "Max injury", valid_from="2026-01", valid_to="2026-02")
         """
+        # Reject inverted intervals: a triple with valid_to < valid_from
+        # would never satisfy `valid_from <= as_of AND valid_to >= as_of`,
+        # so it would be invisible to every query — silently corrupt.
+        if valid_from is not None and valid_to is not None and valid_to < valid_from:
+            raise ValueError(
+                f"valid_to={valid_to!r} is before valid_from={valid_from!r}; "
+                "an inverted interval would be invisible to every KG query"
+            )
+
         sub_id = self._entity_id(subject)
         obj_id = self._entity_id(obj)
         pred = predicate.lower().replace(" ", "_")
--- a/tests/test_knowledge_graph.py
+++ b/tests/test_knowledge_graph.py
@@ -5,6 +5,8 @@ Covers: entity CRUD, triple CRUD, temporal queries, invalidation,
 timeline, stats, and edge cases (duplicate triples, ID collisions).
 """
 
+import pytest
+
 
 class TestEntityOperations:
     def test_add_entity(self, kg):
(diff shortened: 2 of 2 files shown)


jphein:
+1 to rejecting at write time — silently invisible triples are a hard failure mode to debug after the fact.

One coordination note worth surfacing: lex-string `valid_to < valid_from` is correct *only* once both sides are canonical `YYYY-MM-DD`. Counterexample without `sanitize_iso_date` on the boundary:

```python
"2026-12-01" < "2026-3-01"  # True — the guard would falsely reject Dec→Mar as inverted
```

The PR description names the dependency on #1167. Just noting that if #1167 stalls and #1214 lands first, MCP `tool_kg_add` callers passing non-zero-padded months would hit spurious `ValueError`. Tests in this PR all use canonical form so they don't exercise the gap.

Low-risk in practice (LLM-driven callers tend to emit ISO 8601), and the right fix shape is #1167 anyway. Not blocking — just worth tagging the dependency in case the merge order goes #1214 → #1167 rather than the other way.

arnoldwender:
@jphein Thanks — good call on tagging the merge-order gap explicitly. Concrete behavior if #1214 lands before #1167:

- MCP `tool_kg_add` callers passing non-zero-padded months like `"2026-3-01"` would hit `ValueError: valid_to < valid_from` even on valid intervals, because lex-string `"2026-12-01" < "2026-3-01"` is `True`.
- LLM-driven callers tend to emit canonical ISO 8601, so practical exposure is low.
- The right fix is #1167 (`sanitize_iso_date` at the MCP boundary), not duplicate validation here.

Happy with either merge order. If #1167 first, the gap closes naturally; if #1214 first, the limitation is documented and #1167 follows up.

jphein:
Thanks @arnoldwender — concrete on the merge-order behavior. Your read matches mine: practical exposure is small because LLM-driven callers tend to emit canonical ISO 8601, but the gap is worth the explicit docstring note in the meantime. Happy to defer to whichever of #1167 / #1214 lands first; the other will get a small rebase note if needed.

PR #1216, opened 2026-04-26
fix(migrate): close SQLite connection and clean temp palace on exception

arnoldwender:
## What and Why

Two resource leaks in `mempalace.migrate` that bite hardest on Windows:

### 1. SQLite connection leak in `extract_drawers_from_sqlite()`

`mempalace/migrate.py:55` opened a connection at the top of the function and only closed it on the happy path. Any exception during query/iteration leaked the handle. On Windows that holds a file lock on `chroma.sqlite3` — the rest of the migration cannot `shutil.copytree` or `os.replace` the palace directory until the process exits. Users see a confusing `PermissionError` after a partial extraction.

### 2. Orphan temp dir in `migrate()`

`mempalace/migrate.py:232` (now ~234) called `tempfile.mkdtemp(prefix="mempalace_migrate_")` with no `try/finally`. If anything between mkdtemp and the final `os.replace(temp_palace, palace_path)` raises (chromadb open failure, batch import failure, count mismatch, EXDEV fallback failure), the partial palace directory under the system temp root leaks forever. For a 50K-drawer palace this can be hundreds of MB the user has to find and clean by hand.

## Fix

1. `extract_drawers_from_sqlite` — wrap the connection in `contextlib.closing(sqlite3.connect(db_path))` so `close()` runs on every exit path. `detect_chromadb_version` already had a `try/finally` and is left alone.
2. `migrate` — wrap the temp-palace import-and-swap dance in `try/finally`. The cleanup checks `os.path.exists(temp_palace)` first; on the happy path `os.replace` consumes the temp dir so the guard makes cleanup a no-op.

## Test plan

- `test_extract_drawers_returns_drawers` — happy path still works after the `closing` refactor (real SQLite fixture with the schema `extract_drawers_from_sqlite` reads)
- `test_migrate_cleans_temp_palace_on_chromadb_failure` — `mkdtemp`'s directory is removed when `ChromaBackend.get_or_create_collection` raises, asserted via a `tracking_mkdtemp` that captures the path
- All 7 `test_migrate.py` tests pass
- Full suite: 1318 passed locally (2 pre-existing `test_backends.py::test_pin_hnsw_threads*` failures unrelated — chromadb `configuration_json["hnsw"]` schema drift, also fails on clean `develop`)

### DIFF
--- a/mempalace/migrate.py
+++ b/mempalace/migrate.py
@@ -22,8 +22,10 @@ import errno
 import os
 import shutil
 import sqlite3
+import tempfile
 import uuid
 from collections import defaultdict
+from contextlib import closing
 from datetime import datetime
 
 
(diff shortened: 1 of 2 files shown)


arnoldwender:
Rebased on `upstream/develop` (latest `1247e17`, post-3.3.5 release).

Conflict resolution:
- **`mempalace/migrate.py`** import block — both `tempfile` (this PR) and `uuid` (upstream) belong; kept both alphabetically.
- **`tests/test_migrate.py`** — three conflict regions covering the import line and two unrelated test additions:
  1. Imports unioned: `_restore_stale_palace, collection_write_roundtrip_works, extract_drawers_from_sqlite, migrate`
  2. Upstream's `_FakeGetResult` / `_WritableFakeCollection` test scaffolding for write-roundtrip probing was preserved alongside this PR's `_make_minimal_chromadb_sqlite` helper + `test_extract_drawers_returns_drawers`.
  3. Upstream's `test_migrate_dry_run_rebuilds_when_collection_is_readable_but_not_writable` and this PR's `test_migrate_cleans_temp_palace_on_chromadb_failure` were both kept; the `palace_dir` fixture setup was duplicated so each test owns its own.

Ran `uv run pytest tests/test_migrate.py`: **11 passed**. `ruff check` clean.

PR #1219, opened 2026-04-26
feat(search): expose drawer_id in mempalace_search results

### LINKED ISSUE #1026: Expose Drawer IDs in MemPalace Tools
## Summary

Add drawer ID visibility to MemPalace search and listing tools so agents can programmatically identify and delete specific entries without manual filesystem inspection.

## Problem

The current toolset (`mempalace_search`, `mempalace_list_wings`, `mempalace_list_rooms`, `mempalace_status`) returns content and metadata but **does not expose internal drawer IDs**. The `mempalace_delete_drawer` command requires a drawer ID to operate. This creates a gap: agents can find data but cannot act on it for deletion or targeted management.

### Real-World Impact

- Multi-agent setups (e.g. Hermes + Claude sharing a palace) cannot be cleanly separated without manual cleanup
- Stale or misattributed entries cannot be pruned automatically
- Any bulk cleanup operation requires dropping to the filesystem level

## Proposed Solution

**Option A — Augment existing tools (preferred, minimal change):**

Extend `mempalace_search` results to include a `drawer_id` field alongside each hit. Example:

```json
{
  "results": [
    {
      "drawer_id": "d-a1b2c3",
      "wing": "wing_claude",
      "room": "sessions",
      "summary": "Session log 2025-04-12",
      "relevance": 0.91
    }
  ]
}
```

**Option B — New dedicated tool:**

Add `mempalace_list_drawers` that accepts:

| Parameter | Required | Description |
|-----------|----------|-------------|
| `wing`    | Yes      | Target wing to list drawers from |
| `room`    | No       | Narrow to a specific room |
| `pattern` | No       | Filter drawer labels/content by substring or regex |
| `limit`   | No       | Max results (default 50) |

Returns an array of `{ drawer_id, label, created_at, wing, room }`.

**Option C — Both:**

Augment search results (Option A) *and* provide the dedicated listing tool (Option B) for maximum flexibility.

## Suggested Workflow After Implementation

```
1. mempalace_search query="_claude" → returns hits with drawer_ids
2. Agent reviews list, confirms targets
3. mempalace_delete_drawer id="d-a1b2c3" (per entry, or batch)
4. mempalace_status → verify cleanup
```

## Priority

**Medium-High** — This is a blocker for automated multi-agent palace separation and general housekeeping. Current workaround (manual filesystem access) is fragile and not agent-accessible.

## Notes

- Option A is the lowest-effort change and solves the most common case
- Option B adds power-user capability for bulk operations
- Consider also adding a `mempalace_delete_drawers` (batch) variant accepting an array of IDs

pepo72:
Closes #1026.

## Problem
`mempalace_search` returns `text`, `wing`, `room`, `source_file`, `similarity` etc. but not `drawer_id`. Any caller that wants to follow up with `mempalace_delete_drawer`, `mempalace_get_drawer`, or `mempalace_update_drawer` (all of which require `drawer_id`) has to drop into SQLite and reverse-lookup by content/metadata. This breaks the MCP abstraction and turns simple cleanup workflows into bespoke scripts.

## Fix
Chroma's `query()` already returns `ids` on every result. Thread it through `search_memories` (`searcher.py`) into each result entry as `drawer_id`. Three lines.

## Diff
```diff
     scored: list = []
-    for doc, meta, dist in zip(
+    for drawer_id, doc, meta, dist in zip(
+        _first_or_empty(drawer_results, "ids"),
         _first_or_empty(drawer_results, "documents"),
         _first_or_empty(drawer_results, "metadatas"),
         _first_or_empty(drawer_results, "distances"),
     ):
         ...
         entry = {
+            "drawer_id": drawer_id,
             "text": doc,
             ...
```

## Test
Smoke-tested against a real palace (~1100 drawers) on `develop`-equivalent v3.3.3:
```
{"drawer_id": "drawer_wing_peter_health_2c4fb820a4186c5c", "wing": "wing_peter", "room": "health", ...}
```

## Scope
Intentionally minimal — no rename, no schema, no behavior change beyond adding the field. Re-rank / hydration / closet-boost paths preserve the new key (they only `pop` `_sort_key` / `_source_file_full` / `_chunk_index`).

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -442,9 +442,13 @@ def search_memories(
     CLOSET_DISTANCE_CAP = 1.5  # cosine dist > 1.5 = too weak to use as signal
 
     scored: list = []
+    ids = _first_or_empty(drawer_results, "ids")
+    docs = _first_or_empty(drawer_results, "documents")
+    if not ids and docs:
+        ids = [None] * len(docs)
     for drawer_id, doc, meta, dist in zip(
-        _first_or_empty(drawer_results, "ids"),
-        _first_or_empty(drawer_results, "documents"),
+        ids,
+        docs,
         _first_or_empty(drawer_results, "metadatas"),
         _first_or_empty(drawer_results, "distances"),
     ):
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -442,7 +442,8 @@ def search_memories(
     CLOSET_DISTANCE_CAP = 1.5  # cosine dist > 1.5 = too weak to use as signal
 
     scored: list = []
-    for doc, meta, dist in zip(
+    for drawer_id, doc, meta, dist in zip(
+        _first_or_empty(drawer_results, "ids"),
         _first_or_empty(drawer_results, "documents"),
         _first_or_empty(drawer_results, "metadatas"),
         _first_or_empty(drawer_results, "distances"),
@@ -465,6 +466,7 @@ def search_memories(
 
         effective_dist = dist - boost
         entry = {
+            "drawer_id": drawer_id,
             "text": doc,
             "wing": meta.get("wing", "unknown"),
             "room": meta.get("room", "unknown"),


jphein:
Independent confirmation — landed the same change in [our fork](https://github.com/jphein/mempalace) at [`9a8bb77`](https://github.com/jphein/mempalace/commit/9a8bb77) yesterday. Same diff shape, +1 to the approach.

We extended it to two adjacent payloads with the same "consumer wants to follow up by ID" problem: `tool_diary_read` entries (for `mempalace_diary_read`) and `tool_session_recovery_read` entries (fork-only MCP tool for the dedicated session-recovery collection split). Happy to fold those into this PR, or land yours narrow and I'll file a follow-up — either works.

One tip from cherry-picking ours: a couple existing test mocks omit `ids` in `mock_col.query.return_value` (e.g. the BM25 hybrid rerank test in `tests/test_searcher.py`). Threading `_first_or_empty(drawer_results, "ids")` makes those fixtures collapse to zero hits because `zip` truncates to the shortest input. Defensive pad worked for us:

```python
ids = _first_or_empty(drawer_results, "ids")
if not ids and drawer_results.get("documents"):
    ids = [None] * len(_first_or_empty(drawer_results, "documents"))
```

Production chromadb always returns ids — purely a test-fixture compatibility shim, but might be worth including here to avoid mock churn during review.

Closes #1026 either way.

---

**Update 2026-04-27**: going with the follow-up path — your PR can land narrow, and I'll file the diary + recovery slices as a small separate PR after. No need to expand scope here.

pepo72:
Thanks for the review and the heads-up on the test mocks — added the defensive pad in 490e5b1. Happy to land narrow as suggested; looking forward to your follow-up PR for the diary and recovery slices.

PR #1227, opened 2026-04-26
fix(repair): detect HNSW capacity divergence and fall back to BM25 (#1222)

### LINKED ISSUE #1222: HNSW max_elements frozen at 16K while collection grows to 100K+ entries — MCP server segfaults on every tool call
**What happened?**

After organic palace growth through regular mining and session capture, the MCP server began crashing with a segfault (exit 139) on every tool call, including \`mempalace_status\`. The palace was completely inaccessible.

Diagnosis revealed a fundamental mismatch: the HNSW vector index (\`ba5ad160…\` VECTOR segment for \`mempalace_drawers\`) had \`max_elements=16384\` but \`chroma.sqlite3\` held **192,997 embeddings**. The HNSW resize mechanism stopped at some point, leaving SQLite and HNSW permanently diverged. Any operation that touched the HNSW triggered SIGSEGV in the Rust layer.

This is distinct from the link_lists.bin sparse-file bloat issues (#344, #1092). No bloated files here — just the SQLite entry count outgrowing \`max_elements\` with no guard or warning.

**What did you expect?**

Either ChromaDB auto-resizes the HNSW as the collection grows, or mempalace detects and warns about a SQLite/HNSW count mismatch before the server segfaults.

**How to reproduce:**

1. Allow a palace to grow organically past ~16K drawers via \`mine\` or session hooks
2. At some point (likely after an interrupted mine or concurrent write), the HNSW stops resizing
3. SQLite continues accumulating entries; HNSW stays at 16K max_elements
4. On next MCP server start, every tool call segfaults

The divergence is completely silent — no warning that HNSW is lagging behind SQLite.

**Workaround:**

```bash
# 1. Export current drawers from SQLite (safe, no HNSW touch)
python3 ~/.mempalace/recover.py export

# 2. Confirm the mismatch
python3 ~/.mempalace/recover.py status

# 3. Delete the undersized HNSW dir (segment ID = VECTOR segment for mempalace_drawers)
rm -rf ~/.mempalace/palace/<segment-id>/

# 4. ChromaDB auto-creates a fresh HNSW — BM25 text search keeps working for all entries
#    Semantic vector search degrades until new entries organically rebuild the index
```

Find the right segment ID:
```python
import sqlite3
conn = sqlite3.connect(os.path.expanduser('~/.mempalace/palace/chroma.sqlite3'))
c = conn.cursor()
c.execute("""SELECT s.id FROM segments s
             JOIN collections col ON s.collection = col.id
             WHERE col.name = 'mempalace_drawers' AND s.scope = 'VECTOR'""")
print(c.fetchone()[0])
```

Also worth knowing: \`recover.py test-fresh\` fails with "attempt to write a readonly database" if the export JSON retains the \`chroma:document\` reserved metadata key. Strip it before upsert: \`{k: v for k, v in meta.items() if k not in {'chroma:document', 'chroma:embedding'}}\`.

**Suggested fix:**

- \`recover.py status\` (and ideally the MCP server startup) should compare the SQLite embedding count vs HNSW \`cur_element_count\` and \`max_elements\`, and warn loudly when they diverge significantly
- A SIGSEGV in the HNSW layer shouldn't take the whole MCP server down silently — some form of guard or graceful fallback to BM25-only mode would prevent the "palace completely unreachable" scenario

**Environment:**
- OS: Linux 6.6.87 (WSL2 / Debian)
- Python: 3.12
- MemPalace: 3.3.3 (pip), 3.3.0 (plugin)
- chromadb: 1.5.8

igorls:
Closes #1222.

## Summary

When chromadb's HNSW segment freezes at a stale `max_elements` while sqlite keeps accumulating embeddings (the failure mode reported in #1222: 16 384 vs 192 997), the next segment load segfaults the MCP server on every tool call. This PR adds **detection** + **graceful degradation** so the palace stays reachable.

- **Detection** — `hnsw_capacity_status` (`mempalace/backends/chroma.py`) compares sqlite's embedding count against the HNSW pickle's `id_to_label` size. Pure-filesystem read: no chromadb client, no `hnswlib` import, no segment load.
- **Read-only CLI** — `mempalace repair-status` prints the comparison and recommends `repair rebuild` when divergence exceeds the flush-lag threshold (2× chromadb's `sync_threshold` of 1000).
- **MCP startup pre-flight** — `mcp_server` runs the probe at startup and on every reconnect. When divergent, it sets `_vector_disabled` and surfaces the state through `tool_status` / `tool_reconnect` responses.
- **BM25-only fallback** — `_bm25_only_via_sqlite` (`mempalace/searcher.py`) reads drawers via chromadb's own FTS5 trigram index and re-ranks with the existing Okapi-BM25, so search keeps working even when vector search is unsafe to invoke.

The pickle reader uses a tight-allowlist unpickler (`_SafePersistentDataUnpickler`) — only `chromadb.segment.impl.vector.local_persistent_hnsw.PersistentData` resolves; everything else, including `os.system`, is rejected before `find_class` returns.

## Threshold rationale

ChromaDB 1.5.x flushes HNSW asynchronously, so the index lags sqlite by up to `sync_threshold` (default 1000) records under steady write load. Two windows worth (2000 absolute, or 10% of sqlite_count, whichever is greater) is the floor for "real divergence." The #1222 case (176 613 missing of 192 997) blasts past either; a typical post-mine palace shows a few hundred missing.

## Test plan

- [x] 17 new tests in `tests/test_hnsw_capacity.py` covering segment-id lookup, the unpickler allowlist (incl. an `os.system` rejection probe), all four capacity-status branches, BM25 fallback shape/filtering, and the `repair-status` CLI output
- [x] Full pytest suite: 1408 passed
- [x] `ruff check` + `ruff format --check` (CI-pinned 0.4.x): clean
- [ ] Manual verification on a real diverged palace (cannot reproduce the 16 384/192 997 case locally; probe and fallback do work on a live 1798/1000 palace as a softer divergence)

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -211,6 +211,272 @@ def quarantine_stale_hnsw(palace_path: str, stale_seconds: float = 300.0) -> lis
     return moved
 
 
+def _vector_segment_id(palace_path: str, collection_name: str) -> Optional[str]:
+    """Return the VECTOR segment UUID for ``collection_name`` or ``None``.
+
+    Reads ``chroma.sqlite3`` directly so we never have to load a segment
+    that may segfault on open (#1222 is exactly this case).
+    """
+    db_path = os.path.join(palace_path, "chroma.sqlite3")
+    if not os.path.isfile(db_path):
+        return None
+    try:
+        conn = sqlite3.connect(f"file:{db_path}?mode=ro", uri=True)
+        try:
+            row = conn.execute(
+                """
+                SELECT s.id
+                FROM segments s
+                JOIN collections c ON s.collection = c.id
+                WHERE c.name = ? AND s.scope = 'VECTOR'
+                LIMIT 1
+                """,
+                (collection_name,),
+            ).fetchone()
+            return row[0] if row else None
+        finally:
+            conn.close()
+    except sqlite3.Error:
+        return None
+
+
+class _PersistentDataStub:
+    """Minimal stand-in for chromadb's ``PersistentData`` during safe unpickling.
+
+    Accepts any constructor args so pickle's REDUCE opcode succeeds,
+    captures ``__setstate__`` into ``__dict__``. Only used by
+    :func:`_hnsw_element_count` — never persisted, never re-pickled.
+    """
+
+    def __init__(self, *args, **kwargs):
+        # Some chromadb versions pickle PersistentData by passing init args
+        # positionally via REDUCE. We don't care about reconstructing the
+        # object faithfully — we only need the id_to_label dict — so swallow
+        # all positional args and re-expose the relevant attributes via
+        # __setstate__ or __dict__ population further down.
+        pass
+
+    def __setstate__(self, state):
+        if isinstance(state, dict):
+            self.__dict__.update(state)
+        elif isinstance(state, tuple) and len(state) == 2 and isinstance(state[1], dict):
+            # (slot_state, dict_state) two-tuple form — only the dict part has
+            # the named attributes we care about.
+            self.__dict__.update(state[1])
+
+
+class _SafePersistentDataUnpickler:
+    """Whitelist-only unpickler for ``index_metadata.pickle``.
+
+    Allows only ``PersistentData`` from chromadb's HNSW module; everything
+    else raises ``UnpicklingError``. Standard container types (dict, list,
+    tuple, str, int, float) are handled by the pickle machinery itself
+    and don't need allowlisting via ``find_class`` — only constructed
+    classes do.
+
+    This is the same trust model chromadb uses (it pickles its own files),
+    but with a tight class allowlist so a tampered file can't instantiate
+    arbitrary classes during deserialization.
+    """
+
+    _ALLOWED = frozenset(
+        {
+            (
+                "chromadb.segment.impl.vector.local_persistent_hnsw",
+                "PersistentData",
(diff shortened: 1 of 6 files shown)


igorls:
Addressed Copilot review + the Python 3.9 CI red in 57ac669:

- **Python 3.9 CI**: replaced the `dict | None` annotated assignment in `mcp_server.py` with a type-comment so module load doesn't evaluate PEP 604 syntax on 3.9.
- **Wrong CLI command in user-facing strings**: `mempalace repair rebuild` doesn't exist (only `mempalace repair` and `mempalace repair-status`). Fixed every message, docstring, and test assertion.
- **`tool_search` defeated its own fallback**: was calling `_get_client()` to trigger the probe, which constructs the very chromadb client the fallback is meant to avoid. Switched to `_refresh_vector_disabled_flag()` (pure sqlite/pickle, no chromadb).
- **`tool_status` opened the chromadb collection unconditionally**: added `_tool_status_via_sqlite` short-circuit that runs when `_vector_disabled` is set — wing/room counts come back via direct `embedding_metadata` reads so status stays reachable in the segfault scenario.
- **Recency-window query was unguarded**: wrapped in `sqlite3.Error` handling with an `id`-ordered fallback for legacy schemas missing `created_at`.

CI is green across lint + 3.9/3.11/3.13/macOS/Windows. 1409 tests pass locally.

PR #1231, opened 2026-04-27
fix(hooks): always mine the active transcript as convos, additive to MEMPAL_DIR

igorls:
## Summary

Follow-up to #1230. That PR fixed `--mode convos` for one specific case: a Python hook user with `MEMPAL_DIR` unset. Two configurations were still silently broken:

- **`MEMPAL_DIR` set to a project directory** — convos never mined; the hook ran the projects miner on `MEMPAL_DIR` and ignored the transcript path entirely.
- **`MEMPAL_DIR` set to a conversations directory** (per the old `hooks/README.md` recommendation) — the projects miner ran on JSONL, ingesting it as if it were source code.

Both shell hooks (`hooks/mempal_save_hook.sh`, `hooks/mempal_precompact_hook.sh`) had the same `MEMPAL_DIR`-overrides-`TRANSCRIPT_PATH` bug *and* were missing `--mode` on every spawned `mempalace mine`. The PreCompact shell hook didn't even read `transcript_path` from stdin.

This PR closes all of those.

## Approach

Make the auto-ingest **additive**, not "MEMPAL_DIR overrides transcript":

- `_get_mine_dir` is renamed to `_get_mine_targets` and now returns `list[tuple[str, str]]`.
- `MEMPAL_DIR` (when valid) contributes `(dir, "projects")`.
- A valid transcript JSONL contributes `(parent_dir, "convos")`.
- Both can appear together — the hook then spawns one ingest per target.

The shell save and precompact hooks get the same dual-target structure (each writes one independent `mempalace mine ... --mode <mode>` call). The precompact shell hook also gained `transcript_path` parsing so it can run the convos mine synchronously before context is compressed.

`hooks/README.md` updated to describe `MEMPAL_DIR` accurately: it is a **project files** target, never a convos target. The active conversation transcript is mined automatically regardless of `MEMPAL_DIR`.

## Why this exists

Milla flagged that users had been told their conversations were stored verbatim, but with the original `MEMPAL_DIR`-overrides-transcript semantics + the missing `--mode` flag, that was only true for one of four common configurations. After this PR, the auto-mine path produces verbatim convo drawers for every supported hook configuration.

## Stack

This branch was cut from `fix/hooks-convos-mode-1232` (#1230). The two are textually adjacent but logically independent — either can land first; if #1230 lands first, this rebases cleanly.

## Test plan

- [x] `_get_mine_targets` covers all four cases (none / env-only / transcript-only / both)
- [x] New `test_maybe_auto_ingest_with_both_set` asserts two `Popen` spawns with the right `--mode` per spawn
- [x] New `test_mine_sync_with_both_set` asserts the same for the precompact-sync path
- [x] `test_hook_mines_transcript_path` updated to assert `dirname "$TRANSCRIPT_PATH"` + `--mode convos` are present in the shell hook (the previous assertion required `MINE_DIR`, which no longer exists)
- [x] Full suite green: `1416 passed, 1 skipped`

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,10 @@ system prompt:
 Two Claude Code hooks save periodically and before context compression:
 [mempalaceofficial.com/guide/hooks](https://mempalaceofficial.com/guide/hooks.html).
 
+For per-message recall on top of the file-level chunks the hooks produce,
+run `mempalace sweep <transcript-dir>` periodically — it stores one
+verbatim drawer per user/assistant message, idempotent and resume-safe.
+
 ---
 
 ## Requirements
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -67,7 +67,7 @@ Edit `mempal_save_hook.sh` to change:
 
 - **`SAVE_INTERVAL=15`** — How many human messages between saves. Lower = more frequent saves, higher = less interruption.
 - **`STATE_DIR`** — Where hook state is stored (defaults to `~/.mempalace/hook_state/`)
-- **`MEMPAL_DIR`** — Optional. Set to a conversations directory to auto-run `mempalace mine <dir>` on each save trigger. Leave blank (default) to let the AI handle saving via the block reason message.
+- **`MEMPAL_DIR`** — Optional **project directory** (code, notes, docs) to also mine on each save trigger, with `--mode projects`. The hook ALWAYS mines the active conversation transcript automatically with `--mode convos` — `MEMPAL_DIR` is purely additive, never an override. Leave blank if you don't want to ingest project files.
 - **`MEMPALACE_PYTHON`** — Optional env var. Python interpreter with mempalace + chromadb installed. Auto-detects: `MEMPALACE_PYTHON` env var → repo `venv/bin/python3` → system `python3`. Set this if your venv is in a non-standard location.
 
 ### mempalace CLI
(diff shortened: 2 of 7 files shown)


igorls:
Pushed `fe56797` addressing the Copilot review:

- **Removed the duplicate transcript target on the Python side.** `_get_mine_targets` now returns only the `MEMPAL_DIR` projects target. Transcript ingestion stays in `_ingest_transcript`, which was already firing on every hook event. Side effect: the PID-file overwrite (Copilot #2) and the 60s+60s sync timeout stack (Copilot #4) are gone — Python path now spawns at most one `_spawn_mine` per call.
- **Switched the spawned `mempalace mine` interpreter to `_mempalace_python()`** in both `_maybe_auto_ingest` and `_mine_sync` (Copilot #3).
- **Replaced `eval $(...)` in both shell hooks** with a `mapfile` reader against one-value-per-line sanitized stdout (Copilot #6).
- **Mirrored `_validate_transcript_path` in the shell hooks** as `is_valid_transcript_path` and gated the convos mine on it (Copilot #7, #8).
- **Tightened the traversal test** to use a `.jsonl` path with `..` segments so the traversal branch is actually exercised (Copilot #5).

The shell-side additive structure (separate `MEMPAL_DIR` and transcript mines) is preserved — the shell hooks have no `_ingest_transcript` equivalent, so the additive shape is correct there.

Tests: `1418 passed, 1 skipped` (full suite minus benchmarks); ruff check + format clean.

Issue #59, opened 2026-04-07
feat: add import support for more AI tool session formats (Cursor, Copilot, Codex, Windsurf, Aider, etc.)


adv3nt3:
## Summary

MemPalace normalizes 6 chat export formats in `normalize.py`:
- Claude AI JSON
- ChatGPT `conversations.json`
- Claude Code JSONL
- OpenAI Codex CLI JSONL *(added recently via PR #61)*
- Slack JSON
- Plain text

PR #155 (in review) adds Gemini CLI JSON. PR #169 (in review) adds Pi agent JSONL. PR #172 (in review) adds Aider markdown. PR #23 adds OpenCode SQLite (in review). There's no support yet for several other widely-used AI coding tools.

## Candidate formats

| Tool | Export format | Status |
|------|-------------|--------|
| **Codex (OpenAI CLI)** | JSONL sessions (`~/.codex/sessions/`) | ✅ PR #61 merged |
| **Gemini CLI** | JSON sessions (`~/.gemini/tmp/{hash}/chats/session-*.json`) | 🔄 PR #155 in review |
| **Pi agent** | JSONL sessions (`~/.config/pi/agent/sessions/`) | 🔄 PR #169 in review |
| **Aider** | Markdown chat logs (`.aider.chat.history.md`) | 🔄 PR #172 in review (by @mvanhorn) |
| **OpenCode** | SQLite session DB | 🔄 PR #23 in review |
| **Cursor** | SQLite (`~/.cursor/`) or workspace state | — |
| **GitHub Copilot Chat** | VS Code extension state (JSON/SQLite) | — |
| **Windsurf (Codeium)** | Local session files | — |
| **Continue.dev** | JSON session logs (`~/.continue/sessions/`) | — |
| **Cline (prev. Claude Dev)** | VS Code extension state | — |

### Pi session format (from @tunnckoCore)
- **Path**: `~/.config/pi/agent/sessions/{encoded-cwd}/{timestamp}_{uuid}.jsonl`
- **Folder naming**: CWD path with `/` replaced by `-`, wrapped in `--`, e.g. `--home-arcka-openclaude--`
- **File naming**: `{ISO-timestamp}_{uuid}.jsonl`
- **JSONL types**: `session` (header), `model_change`, `thinking_level_change`, `message`, `toolCall`, `compaction`, `branch_summary`, `label`, `custom`
[...]

Issue #1194, opened 2026-04-25
Topic tunnels silently skipped for wings with hyphenated dir names


bensig:
## Summary

After #1184 (cross-wing topic tunnels), tunnels are silently never created for projects whose directory name contains a hyphen or space. The miner's tunnel-computation step looks the wing up in `topics_by_wing` under the **raw** dirname (`mempalace-public`), but `init` records it under that raw name while writing **`mempalace.yaml`** with the **normalized** name (`mempalace_public`). At mine time the yaml wins, the lookup misses, and `_compute_topic_tunnels_for_wing` returns 0 with no warning.

Real-world common case: any project dir like `my-app`, `mempalace-public`, `react-thing` is affected. Single-token dirs (`memorymark`, `desktop`) work by coincidence.

## Repro

Five sibling project dirs init+mine'd in order with `--llm --llm-provider anthropic`:

| Wing dir         | yaml `wing`       | `topics_by_wing` key | Stack overlap                              |
|------------------|-------------------|----------------------|--------------------------------------------|
| `mempalace-ts`   | `mempalace_ts`    | `mempalace-ts`       | shares Angular, Bun with memorymark        |
| `desktop`        | `desktop`         | `desktop`            | —                                          |
| `memorymark`     | `memorymark`      | `memorymark`         | shares Angular, Bun with mempalace-ts      |
| `mempalace-public` | `mempalace_public` | `mempalace-public` | shares 5 topics with mempal-private        |
| `mempal-private` | `mempal_private`  | `mempal-private`     | shares 5 topics with mempalace-public      |

Expected tunnels: 7 (2 ts↔memorymark + 5 public↔private).
Actual tunnels: 2 (only the no-hyphen pair `memorymark ↔ mempalace-ts`).

Direct repro of the lookup miss:

\`\`\`python
>>> from mempalace.miner import get_topics_by_wing
>>> from mempalace.palace_graph import topic_tunnels_for_wing
>>> m = get_topics_by_wing()
>>> list(m.keys())
['mempalace-ts', 'desktop', 'memorymark', 'mempalace-public', 'mempal-private']
>>> len(topic_tunnels_for_wing('mempal_private', m, min_count=1))   # what mine actually calls
0
>>> len(topic_tunnels_for_wing('mempal-private', m, min_count=1))   # what's in topics_by_wing
5
\`\`\`

## Root cause

- `mempalace/cli.py:146` (cmd_init): \`wing = project_path.name\` — raw dirname.
[...]

PR #1236, opened 2026-04-27
feat(convo_miner): auto-route AI tool sessions to wing_api

milla-jovovich:
When mempalace mine --mode convos is invoked against a directory inside a known AI-tool storage path (Claude Code, Codex CLI, Gemini CLI), the destination wing now auto-defaults to wing_api rather than the directory basename. Conversations from external API-keyed tools land grouped under a single dedicated wing for visibility.

Detected paths (exact-segment match — substrings like .gemini-backup or .codex-archive do NOT match):

  - any segment .codex (Codex CLI sessions / archives)
  - any segment .gemini (Gemini CLI sessions under ~/.gemini/tmp/...)
  - the consecutive segment pair .claude/projects (Claude Code). .claude alone is NOT matched - that is the settings/config dir, not a conversation source.

Wing-resolution precedence (first match wins):

  1. Explicit --wing argument from the user - always wins
  2. AI-tool path detection -> wing_api
  3. Basename fallback (existing behavior, unchanged)

Two new helpers split out of mine_convos for unit-test coverage:

  - _is_ai_tool_path(path: Path) -> bool
  - _resolve_wing(convo_path: Path, wing: Optional[str]) -> str

mine_convos now calls _resolve_wing in place of its inline basename logic. No other call sites or downstream consumers change.

Test coverage:

  - 15 unit tests covering positive matches (Claude Code subdir + root, Codex root + sessions, Gemini root + chats), negative cases (.claude alone is settings dir, unrelated paths, substring no-match on .gemini-backup / .codex-archive), explicit --wing override, auto-route trio, basename fallback, empty-string-as-no-wing.
  - End-to-end smoke test (manual): real-shape Claude Code JSONL fixture mined via the actual CLI; sqlite read-back of /tmp palace confirms drawers landed with wing='wing_api' and verbatim content preserved; mempalace search --wing wing_api returns expected content ranked.
  - Full pytest sweep: 1388 baseline + 15 new = 1403 passed, zero regressions.

Closes part of #59 for the auto-routing UX.

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -14,6 +14,7 @@ import hashlib
 from pathlib import Path
 from datetime import datetime
 from collections import defaultdict
+from typing import Optional
 
 from .normalize import normalize
 from .palace import (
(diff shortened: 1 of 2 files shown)


bensig (approved):
**Approve.** Clean separation of concerns, correct path-matching, full test coverage, and the wing-resolution precedence is exactly right.

Full pytest on this branch: **1456 passed, 1 skipped, 19 in \`test_convo_miner.py\`** (matches the +15 new the body promised).

## What I checked

### Path matching is right and defensive

- \`path.resolve().parts\` handles symlinks and relative paths correctly. A user who symlinks a Claude transcript dir somewhere else still gets routed to \`wing_api\` because resolve surfaces the original \`.claude/projects/\` path.
- \`try/except (OSError, RuntimeError)\` around \`resolve()\` catches the rare case of a broken symlink or path-too-long without crashing the mine.
- **Exact-segment match** is the key correctness detail. \`.gemini-backup\` and \`.codex-archive\` correctly do NOT match — that would have been an easy mistake. The negative tests cover both.
- \`.claude/projects\` requires the consecutive-segment pair, not bare \`.claude\` (which is the settings dir, not conversations). Also tested.

### Wing-resolution precedence is correct

1. **Explicit \`--wing\`** always wins. User intent sacrosanct.
2. **AI-tool path → \`wing_api\`** when no explicit wing.
3. **\`normalize_wing_name(basename)\` fallback** uses the shared helper from \`config.py\` — same source of truth as \`cmd_init\`, \`room_detector_local\`, and \`miner.load_config\`. Slots cleanly into the #1194 consolidation work that landed yesterday.

The empty-string handling (\`if wing:\` is falsy on \`""\`) matches the #1097 "empty-string as no filter" pattern that's now consistent across the codebase.

### Tests cover the right surface

The 15 new test cases hit:

- **Positive matches**: Claude Code subdir + root, Codex root + sessions, Gemini root + chats
- **Negative cases**: \`.claude\` alone (settings, NOT conversations), unrelated paths, substring no-match on \`.gemini-backup\` / \`.codex-archive\`
- **Override paths**: explicit \`--wing\` beats auto-route, basename fallback on non-AI paths, empty-string treated as no-wing

The negative cases are the ones that prove the matching is exact rather than fuzzy. Good discipline.

### Architecturally sound

Routing API-driven conversations to a dedicated \`wing_api\` (separate from project wings) is the right default. They're a different kind of content — general LLM exchanges that can span any topic — so segregating them into one wing makes search and graph traversal cleaner. Users who want them in a specific project wing pass \`--wing\`; users who do nothing get something semantically reasonable.

## Minor observations (not blockers)

1. **\`wing_api\` is hardcoded** in \`_resolve_wing\`. Could be promoted to a module-level constant (\`_AI_TOOL_DEFAULT_WING = "wing_api"\`) for visibility, but not material.

2. **The detection list is closed.** As more AI-tool ecosystems land (Cursor, Continue, Aider, Zed AI, etc.), this set needs extension. Could become config-driven later (env var or \`config.json\` key like \`ai_tool_path_segments\`). Out of scope for this PR; worth a follow-up issue if the list grows.

3. **Edge case worth knowing:** if a user mines \`~/.claude/projects/-Users-me-Projects-MyProject/\` and wants those Claude conversations *specifically* in \`wing_myproject\`, they need \`--wing myproject\`. The default (\`wing_api\`) is more useful for the majority case where Claude conversations are general-purpose, but worth a doc note that "explicit per-project routing of AI-tool conversations is one flag away."

Closes part of #59. Ship it.

alisacorporation:
@milla-jovovich

photo_2026-04-28_11-40-23

milla-jovovich:
approved.

Issue #284, opened 2026-04-08
pip install requires --break-system-packages


flundstrom2:
**What happened?**

I tried to install mempalace using the instructions, but on my computer, pip  chokes and requires --break-system-package to work

**What did you expect?**

**How to reproduce:**

```
fredrik@By-Tor:~$ pip install mempalace
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
fredrik@By-Tor:~$ python3 --version
Python 3.11.2
fredrik@By-Tor:~$ pip show mempalace
WARNING: Package(s) not found: mempalace
fredrik@By-Tor:~$ pip3 install mempalace
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
fredrik@By-Tor:~$ python -m pip install mempalace
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

[...]

PR #1237, opened 2026-04-27
docs(readme): recommend pipx for install (fixes PEP 668 + global dep conflicts)

OmkarKirpan:
## Summary

The `pip install mempalace` line in the README either fails on
PEP 668-managed Pythons (Debian/Ubuntu, Homebrew) or silently upgrades
`chromadb` / `numpy` / `grpcio` / `click` in the user's global
site-packages and breaks unrelated tools.

A real run on macOS today, against a Python that already had `numba`,
`litellm`, `tutor`, and the OpenTelemetry instrumentation packages
installed, produced this from a single `pip install mempalace`:

```
opentelemetry-instrumentation 0.62b0 requires
  opentelemetry-semantic-conventions==0.62b0, but you have 0.62b1
numba 0.61.2 requires numpy<2.3,>=1.24, but you have numpy 2.4.4
litellm 1.80.11 requires grpcio<1.68.0,>=1.62.3, but you have grpcio 1.80.0
tutor 20.0.2 requires click<8.3.0,>=8.0, but you have click 8.3.3
```

mempalace ships a CLI, so `pipx install mempalace` (or
`uv tool install mempalace`) is the right default — it isolates the
install and still puts `mempalace` on `PATH`. Plain `pip install` is
kept as the documented alternative for users who want `import
mempalace` inside their own activated venv.

Refs #284 (PEP 668 / `--break-system-packages`).

## What changed

- `README.md` "Install" section: lead with `pipx install mempalace`,
  mention `uv tool install` as an equivalent, briefly explain *why*
  (PEP 668 + global dep conflicts), and keep `pip install` for the
  in-venv case.
- No code changes.

## Test plan

- [x] `git diff README.md` — only the Install section changes
- [x] `pipx install mempalace` succeeds on a system Python where
      `pip install mempalace` would either trip PEP 668 or
      downgrade/upgrade global packages


igorls:
Looks good and it's the way it should be installed indeed. @bensig @milla-jovovich ok on this ?

OmkarKirpan:
Rebased onto latest develop (5c9a5e0) and force-pushed (4ed7eb7).

Develop independently landed a similar Install section recommending `uv tool install` — same intent, less context. Resolved the conflict additively rather than overwriting:

- **Kept** the maintainers' `uv tool install` as the primary recommendation (no flip-flopping the chosen tool)
- **Added** the WHY: explicit mention of PEP 668 errors on Debian/Ubuntu/Homebrew Pythons and dep conflicts with `chromadb` / `numpy` / `grpcio` in global site-packages — the original motivation that opened #284
- **Added** `pipx` as a one-line alternative for users who already have it but not `uv`
- **Added** the venv-for-import note so users who want `import mempalace` inside their own venv know which install path to take

Net diff: 1 file, +15 / −1 — same scope as before, now structured as additions on top of develop's wording.

@igorls — you indicated this approach was fine on the original revision. The merged text now also keeps your team's `uv tool` preference visible up top. cc @bensig @milla-jovovich.

PR #1259, opened 2026-04-28
fix: reconfigure stdio to UTF-8 on Windows

### LINKED ISSUE #1242: diary_write fails with Unicode characters on Windows (cp950 stdin   encoding)
## Environment
  - Windows 11 Home (Traditional Chinese)
  - Python 3.13
  - mempalace 3.3.3
  - Claude Code MCP client

  ## Bug
  Any non-ASCII Unicode characters in the `entry` parameter of
  `mempalace_diary_write` (e.g. `★`, `🎉`, emoji) cause an MCP error -32000.
   ASCII-only content works fine.

  This affects all CJK Windows users where the default stdin/stdout encoding
   is cp950 (Big5) or cp1252, not UTF-8.

  ## Root Cause
  The MCP server communicates via JSON-RPC over stdio. On Windows,
  `sys.stdin.encoding` defaults to the system codepage (cp950 for
  Traditional Chinese, cp1252 for Western, etc.). When the MCP client sends
  UTF-8 encoded JSON containing Unicode characters that don't exist in the
  system codepage, `sys.stdin.readline()` raises a `UnicodeDecodeError`.

  Verified:
  ```python
  # On affected Windows system
  import sys
  sys.stdin.encoding  # 'cp950' — not UTF-8

  Fix

  Add UTF-8 reconfiguration at the start of main() in mcp_server.py:

  def main():
      _restore_stdout()
      if sys.platform == "win32":
          sys.stdin.reconfigure(encoding="utf-8")
          sys.stdout.reconfigure(encoding="utf-8")
      logger.info("MemPalace MCP Server starting...")
      # ... rest of main()

  Steps to Reproduce

  1. Install mempalace on a Traditional Chinese Windows system
  2. Configure as MCP server in Claude Code
  3. Call mempalace_diary_write with entry containing ★ or any emoji
  4. Observe: MCP error -32000 Internal tool error

  Workaround

  Set environment variable before launching the MCP server:
  PYTHONUTF8=1 mempalace-mcp

oussamalembarki:
Fixes #1242.

This PR adds a platform check to reconfigure sys.stdin and sys.stdout to use UTF-8 encoding when running on Windows. This prevents UnicodeDecodeError when the MCP server receives non-ASCII characters (emojis, CJK, etc.) from clients.


igorls:
Thanks @oussamalembarki! Closing as duplicate — #1280 is a strict superset of this fix and is the canonical PR for the Windows UTF-8 stdio cluster. It covers everything here (mcp_server stdin/stdout reconfigure) plus:

- stderr coverage and `errors="replace"` so the JSON-RPC loop never crashes on undecodable bytes
- import-time `_force_utf8_stdio()` call that fixes a stdout-redirect race
- `hooks_cli.run_hook` reconfigure (this PR doesn't touch hooks)
- `ensure_ascii=False` in JSON-RPC responses so CJK round-trips without `\uXXXX` escapes
- a real cp936 subprocess regression test

Please follow #1280 (and #1282 for the CLI / fact-checker side, which neither this PR nor #1280 covers). Really appreciate the diagnosis — confirming the fix on multiple forks gives us high confidence it works.

PR #1262, opened 2026-04-28
fix(storage): stop ChromaDB from crashing when reopening an existing …

### LINKED ISSUE #1089: SIGSEGV on `get_or_create_collection` when call-site metadata differs from stored collection metadata (chromadb 1.5.x)
## Symptom

On chromadb 1.5.x, calling `client.get_or_create_collection(name, metadata={...})` against an existing collection whose stored metadata differs from the call-site `metadata` argument can produce a SIGSEGV in `chromadb_rust_bindings.abi3.so` rather than raising a Python-level exception. The crash is non-deterministic — it depends on what the stored metadata is and what the call site passes — but it's reliably reachable in production.

## Reproducer

Any code path that (1) created the `mempalace_drawers` collection once with one set of metadata (e.g. default `{"hnsw:space": "cosine"}`) and (2) later calls `get_or_create_collection(name="mempalace_drawers", metadata={"hnsw:space": "cosine", "hnsw:num_threads": 4})` can trigger this. In our fork this happened naturally as we experimented with different `hnsw:*` knobs during the chromadb 1.5.x migration — a palace created in one session would crash on open in the next if the metadata drifted.

No Python traceback. The process dies mid-`get_or_create_collection` call, typically with "zsh: segmentation fault" or `chromadb.errors.InternalError` before the frame unwinds.

## Proposed workaround

Split the call into `get_collection(name)` first, then fall back to `create_collection(name, metadata=...)` only on `chromadb.errors.InvalidCollectionException`:

```python
try:
    col = client.get_collection(name)
except InvalidCollectionException:
    col = client.create_collection(name, metadata=metadata)
```

This avoids the code path where chromadb's Rust side compares metadata and crashes. The tradeoff: if the user *wanted* new metadata applied to an existing collection, this pattern silently keeps the old metadata. In practice that's fine for mempalace because we only set `hnsw:space: cosine` once at collection birth and don't rely on mutating it later.

## Current fork behavior

**Update 2026-04-30:** correcting two things in the original paragraph below.

1. The guard actually landed on our fork at [`d3a2d22`](https://github.com/jphein/mempalace/commit/d3a2d22) on **2026-04-18**, not 2026-04-10 — apologies for the date error.
2. It was subsequently dropped during a develop merge-conflict resolution at [`7e18a70`](https://github.com/jphein/mempalace/commit/7e18a70) on **2026-04-25**, so the "~400 starts, zero crashes" framing only covers the 7-day window the guard was actually running.

Since 2026-04-25 the guard has been absent. Tracing the call graph (apologies — this paragraph has been edited twice; the prior versions overcorrected in two different directions): chromadb's `get_or_create_collection` is called from two places on current fork main — `backends/chroma.py:1061` and `mcp_server.py:295` — both with identical metadata (`hnsw:space`, `hnsw:num_threads: 1`, `_HNSW_BLOAT_GUARD`). The fork's 2-arg `ChromaBackend.get_or_create_collection(path, name)` legacy shim at `chroma.py:1101` (used by `migrate.py:237` and `cli.py:1002`) delegates through `get_collection(create=True)` to the same metadata-passing line, so metadata is consistent across opens. The residual exposure for the SIGSEGV is mostly legacy palaces created under older chromadb whose stored metadata differs from the current expanded set — a per-palace-history risk, not a structural one. I don't have a clean recurrence count to share.

Leaving the original text below for the historical record.

> ~~Running with this guard on our fork's `_get_client()` in `backends/chroma.py` since 2026-04-10, across ~400 MCP-server starts on a 165K-drawer palace. Zero crashes in this path since the guard landed. Without it we saw ~1 crash every 10-20 starts while iterating on `hnsw:*` metadata.~~

## Why an issue rather than a PR

This is a *workaround* for a chromadb bug, not a fix. Two reasons to ask direction first:

1. **Upstream-of-upstream report.** The cleanest fix is a bug report to chroma-core — `get_or_create_collection` should either raise a Python exception on metadata mismatch, or update metadata in place, never SIGSEGV. If the maintainer prefers that path over a mempalace-side workaround, I'd rather help file the chroma-core issue than land a patch that hides the underlying bug.
2. **Intersects with #1071's `hnsw:num_threads` path.** [#1071](https://github.com/MemPalace/mempalace/pull/1071) (MERGEABLE) writes `hnsw:num_threads` into collection metadata on create. Once that merges, every palace created post-#1071 will have extra metadata keys that pre-#1071 collections don't have — exactly the "metadata differs" condition that triggers this crash. A `get-then-create` guard reduces the blast radius of the upgrade, but the long-term fix is either a chroma-core bug report or a one-time migration that normalizes metadata across palaces.

## Adjacent issues

Not duplicates — different SIGSEGV triggers, different code paths:

- [#974](https://github.com/MemPalace/mempalace/issues/974) — SIGSEGV in HNSW parallel inserts via missing `hnsw:num_threads` (different cause, being addressed by [#1071](https://github.com/MemPalace/mempalace/pull/1071))
- [#521](https://github.com/MemPalace/mempalace/issues/521) — `EXC_BAD_ACCESS` in hnswlib `updatePoint` on modified-file re-mine (different code path, different root cause)
- [#74](https://github.com/MemPalace/mempalace/issues/74) (closed) — null-pointer crash at scale, resolved by adding `mempalace repair` in #119 (user-level escape hatch, not the fix for this specific crash)

## Happy to file a PR if...

- The maintainer wants the `get-then-create` guard as an interim in `backends/chroma.py::_get_client()` → I can file the narrow diff (~15 lines).
- The maintainer prefers the chroma-core route → I can help draft the bug report there.
- The maintainer sees this as fully covered by the #1071 + migration story → I'll close this issue.

No PR yet; flagging the scope first.

Legion345:
## What does this PR do?

  Fixes #1089.

 When opening an existing palace, `get_or_create_collection` was called with hnsw metadata on every open. In ChromaDB 1.5.x, if that metadata differs from what's already stored, the Rust bindings segfault — no traceback, just a core dump. This is what causes the stop hook to crash at session end.

 Fix: try `get_collection` first, fall back to `create_collection` only when the collection doesn't exist yet. Existing palaces open without touching their metadata; new ones are created with the full settings as before.

  Credit to @jphein who described this exact approach in #1089 and has been running it on their fork since April 10 with zero crashes across 400+ starts.

  ## How to test

  Open a palace twice in the same session:

  ```python
  import tempfile, os

  with tempfile.TemporaryDirectory() as d:
      palace = os.path.join(d, "palace")
      backend = ChromaBackend()
      backend.get_collection(palace, collection_name="mempalace_drawers", create=True)
      backend.get_collection(palace, collection_name="mempalace_drawers", create=True)
      print("no crash") 

```
Or run the new regression tests directly:
python -m pytest tests/test_backends.py -v -k "idempotent or preserves_existing"

  Checklist

  - No hardcoded paths

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Any, Optional
 
 import chromadb
+from chromadb.errors import NotFoundError as _ChromaNotFoundError
 
 from .base import (
     BaseBackend,
@@ -1037,15 +1038,18 @@ class ChromaBackend(BaseBackend):
         ef_kwargs = {"embedding_function": ef} if ef is not None else {}
 
         if create:
-            collection = client.get_or_create_collection(
-                collection_name,
-                metadata={
-                    "hnsw:space": hnsw_space,
-                    "hnsw:num_threads": 1,
-                    **_HNSW_BLOAT_GUARD,
-                },
-                **ef_kwargs,
-            )
+            try:
+                collection = client.get_collection(collection_name, **ef_kwargs)
+            except _ChromaNotFoundError:
+                collection = client.create_collection(
+                    collection_name,
+                    metadata={
+                        "hnsw:space": hnsw_space,
+                        "hnsw:num_threads": 1,
+                        **_HNSW_BLOAT_GUARD,
+                    },
+                    **ef_kwargs,
+                )
         else:
             collection = client.get_collection(collection_name, **ef_kwargs)
         _pin_hnsw_threads(collection)
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -372,6 +372,32 @@ def test_chroma_backend_create_collection_sets_hnsw_bloat_guard(tmp_path):
     assert col.metadata.get("hnsw:sync_threshold") == 50_000
 
 
+def test_get_collection_create_true_is_idempotent(tmp_path):
+    """Calling get_collection(create=True) twice on the same name must not crash.
+
+    ChromaDB 1.5.x's Rust bindings SIGSEGV when get_or_create_collection is
+    called with metadata that differs from the stored collection metadata. The
+    fix splits the call into get_collection -> fallback create_collection so the
+    metadata-comparison codepath in chromadb_rust_bindings is never reached for
+    existing collections. Regression guard for issue #1089.
+    """
+    palace = str(tmp_path / "palace")
+    backend = ChromaBackend()
+    backend.get_collection(palace, collection_name="mempalace_drawers", create=True)
+    col2 = backend.get_collection(palace, collection_name="mempalace_drawers", create=True)
+    assert isinstance(col2, ChromaCollection)
+
+
+def test_get_collection_create_true_preserves_existing_metadata(tmp_path):
+    """Existing collection metadata is not overwritten when reopened with create=True."""
+    palace = str(tmp_path / "palace")
+    backend = ChromaBackend()
+    backend.get_collection(palace, collection_name="mempalace_drawers", create=True)
+    col = backend.get_collection(palace, collection_name="mempalace_drawers", create=True)
+    assert col._collection.metadata["hnsw:space"] == "cosine"
+    assert col._collection.metadata.get("hnsw:batch_size") == 50_000
+
+
 def test_fix_blob_seq_ids_converts_blobs_to_integers(tmp_path):
     """Simulate a ChromaDB 0.6.x database with BLOB seq_ids and verify repair."""
     db_path = tmp_path / "chroma.sqlite3"


jphein:
Thanks for picking this up @Legion345 — and for the kind credit, that was generous.

A couple of corrections about the field data, since you anchored the PR description on it and I want to make sure the maintainer isn't reviewing under a wrong impression:

The "since April 10" / "400+ starts, zero crashes" line came from my [#1089](https://github.com/MemPalace/mempalace/issues/1089) issue body, and both numbers are off. The guard actually landed on our fork at [`d3a2d22`](https://github.com/jphein/mempalace/commit/d3a2d22) on **2026-04-18** (8 days later than I wrote there), and it was dropped during a develop merge-conflict resolution at [`7e18a70`](https://github.com/jphein/mempalace/commit/7e18a70) on **2026-04-25**. So the guard was alive in fork main for 7 days, not 20. I've edited #1089 with the correction so the record there is right.

What's actually true today (and apologies — this paragraph has been edited twice already; the prior versions overcorrected in two different directions). Tracing the call graph: chromadb's `get_or_create_collection` is called from two places on current fork main — `backends/chroma.py:1061` and `mcp_server.py:295` — both with identical metadata (`hnsw:space`, `hnsw:num_threads: 1`, plus the bloat-guard keys from #1191). The fork's 2-arg `ChromaBackend.get_or_create_collection(path, name)` legacy shim at `chroma.py:1101` (used by `migrate.py:237` and `cli.py:1002`) delegates through `get_collection(create=True)` to the same metadata-passing line, so metadata is consistent across opens. The residual exposure is legacy palaces created under older chromadb whose stored metadata differs from the current expanded set — a per-palace-history risk, not a structural one. I don't have a clean recurrence count to share. Net: this PR is still genuinely needed because the chromadb-side crash class isn't fixed, and the "validated by JP's fork" framing isn't load-bearing in either direction.

On the diff itself: the approach is right. One small thing I'd flag as a positive — you're catching `chromadb.errors.NotFoundError`, which is the 1.5.x-renamed exception. Our older code referenced `InvalidCollectionException`, and yours is the right one going forward. Both regression tests cover the case cleanly. Once this merges I'll pull it in via the next develop sync.

Thanks again for taking this on, and apologies for the back-and-forth on this comment — I shouldn't have edited my own correction with another wrong claim.

PR #1270, opened 2026-04-29
fix: persistent daemon to prevent zombie processes and enable multi-session access (closes #1229)

### LINKED ISSUE #1229: mempalace.mcp_server has no single-instance guard — concurrent client spawns race on the palace and trigger HNSW corruption (defense-in-depth for #976)
## Description

Claude Desktop's MCP loader can spawn a second `python -m mempalace.mcp_server` process against the same palace before the previous process has finished its shutdown handshake. Two writer processes then operate concurrently on `~/.mempalace/palace/<collection-uuid>/`, which on `mempalace 3.3.2` (i.e. without PR #976's `hnsw:num_threads=1` pin) reliably triggers the HNSW `link_lists.bin` corruption pattern from #974.

PR #976 is the correct ChromaDB-side fix and resolves the corruption itself. This issue is the complementary MemPalace-side defense: the server has no single-instance guard, so any client whose lifecycle misbehaves (Desktop today, potentially other harnesses tomorrow) can put two writers on the same palace. With #976 in place corruption is prevented, but two processes still race on the SQLite metadata layer, double-load the embedding model, and waste RAM/CPU on every overlap. A startup-time PID file + advisory lock would make MemPalace robust against any misbehaving client without requiring fixes upstream of MemPalace.

## How this is different from #976 / #974

- **#976 (merged to `develop`, awaiting v3.3.4):** fixes the ChromaDB-side data race by pinning `hnsw:num_threads=1` so a single-threaded HNSW writer can't corrupt `link_lists.bin` even under concurrent access. **Necessary, in.**
- **This issue:** addresses the MemPalace-side trigger condition — there is nothing in `mempalace.mcp_server` preventing two server processes from attaching to the same palace at the same time. With #976, those two processes won't corrupt the index, but they will still race on SQLite, double-load the embedding model, and waste a couple of GB of RAM each time the client misbehaves. **Defense in depth, not a duplicate.**

Cross-ref: this is the same family of "concurrent writer" bug as #1202 (Stop hook firing a second mine while one is already running). #1202 added a hook-side lock; this issue proposes the equivalent guard inside `mcp_server` itself, so any client — Desktop, Code, a third-party MCP host, a future hook variant — gets the same protection for free.

## Steps to reproduce

1. Install `mempalace 3.3.2` (the version captured in evidence; behavior is structurally present on all current versions since there's no PID guard in `mcp_server`).
2. Configure the MemPalace MCP server in `~/Library/Application Support/Claude/claude_desktop_config.json` per the standard install instructions.
3. Use Claude Desktop normally. The exact trigger from the client side is not yet pinned down — see "What I haven't been able to isolate" below — but in the captured 30-hour window the second-process spawn happens in two distinct patterns:
   - **Pattern A (no shutdown logged):** a second `Initializing server...` line appears ~1h17m after the first, with no `Shutting down server...` between them.
   - **Pattern B (rapid re-spawn after intentional shutdown):** an `intentional shutdown` is logged and then a new `Initializing server...` appears ~1m later, followed by another spawn ~18m after that, all without the original process being cleanly torn down.
4. On `3.3.2` (no `hnsw:num_threads` pin), this overlap is the reproducer for the #974/#976 corruption — `link_lists.bin` bloats unboundedly (peaked at 55 GB in our environment for ~50K vectors; expected size ~30 MB).

## Observed behavior — evidence from a 30-hour Desktop transport log

Concurrent-spawn instances captured in the sanitized Claude Desktop MCP transport log:

| First server `Initializing server...` | Second `Initializing server...` (overlapping) | First server final close |
|---|---|---|
| `2026-04-25T02:32:02.572Z` | `2026-04-25T03:49:08.358Z` (1h17m later, **no `Shutting down server...` between them**) | `2026-04-25T04:07:23.637Z` (`closed unexpectedly` — operator `pkill` during recovery) |
| `2026-04-26T06:35:37.955Z` | `2026-04-26T07:10:47.037Z`, then again `2026-04-26T07:29:08.614Z` | `2026-04-26T08:05:41.878Z` (`closed unexpectedly` — operator `pkill` during recovery) |

Full sanitized log (username/drawer-counts/AAAK content redacted, all timestamps and JSON-RPC events preserved): https://gist.github.com/Seph396/d8f724e58f066201b3cb527d0c7ffcc0

Important context for reading the log: the `closed unexpectedly` lines at `04-25 04:07:23` and `04-26 08:05:41` are not server-side crashes — they correspond to manual `pkill mempalace` invocations during recovery from the runaway `link_lists.bin` growth. The bug being reported here is the **concurrent spawn**, not the unexpected close.

## Why this is a MemPalace-side problem (and not just a Claude Desktop problem)

I want to be careful with the scope of the claim, since I haven't independently audited Claude Desktop's MCP loader.

**What I observed in this environment:** the Desktop transport log shows multiple `Initializing server...` events without intervening `Shutting down server...` events for the same server tag, which is sufficient evidence that two server processes were live concurrently against the same palace.

**What is publicly reported about Claude Desktop's MCP lifecycle:** there is an open upstream report ([anthropics/claude-code#53134](https://github.com/anthropics/claude-code/issues/53134)) describing two internal managers (`directMcpHost` and `LocalMcpServerManager`) spawning every configured MCP server twice on Windows MSIX builds without coordinating. The Cursor community ([forum thread](https://forum.cursor.com/t/mcp-server-race-condition-causes-infinite-process-spawning-on-windows-cursor-2-0-34/139610)) has reported a related "spawns happen faster than a PID lock can be established" race in their MCP loader. Independent fixes for this class of issue exist ([Cresnova/claude-desktop-mcp-fix](https://github.com/Cresnova/claude-desktop-mcp-fix)). My environment is macOS Tahoe, not Windows MSIX, so I'm not asserting it's the same exact bug — only that the pattern of MCP host loaders double-spawning servers is a known, documented class of behavior in the broader ecosystem.

**Why this still belongs in MemPalace:** even if every MCP host eventually fixes its own lifecycle, MemPalace today has no defense against a misbehaving client. A single-instance guard inside `mempalace.mcp_server` is the structural fix — any future client (Desktop today, Cursor, Continue, a third-party MCP host, a hook variant) that misbehaves can't damage the palace if the second process refuses to start.

## Expected behavior

When `python -m mempalace.mcp_server` is invoked while another `mcp_server` process is already attached to the same palace, the second invocation should:

1. Detect the existing process via PID file + advisory lock.
2. Refuse to start, emit a clear stderr message naming the holding PID and palace path, and exit non-zero.
3. The MCP host's transport log will then surface the failure cleanly (`closed unexpectedly` with a useful stderr trail) instead of silently double-attaching.

## Suggested fix

A startup guard inside `mempalace/mcp_server/__init__.py` (or wherever the entrypoint lives — happy to PR once direction is confirmed):

1. On startup, compute a palace-keyed lock path, e.g. `<palace_root>/.mcp_server.lock`.
2. Open the lock file and attempt `fcntl.flock(fd, LOCK_EX | LOCK_NB)` (POSIX) / `msvcrt.locking` (Windows).
3. If the lock is already held: log `"MemPalace MCP server already running for palace <path> (held by PID <pid>) — refusing to start"` to stderr and exit `1`.
4. If acquired: write the current PID into the file, register an `atexit` / signal handler to release on clean shutdown.
5. Stale lock recovery: if the PID in the file is no longer alive (kernel panic, `pkill -9`), reclaim the lock — `flock` releases automatically when the holder dies, but the PID-file content will be stale and the new process should overwrite it.

**One caveat from prior art:** the [Cursor forum report](https://forum.cursor.com/t/mcp-server-race-condition-causes-infinite-process-spawning-on-windows-cursor-2-0-34/139610) notes that some loaders can spawn processes "faster than a PID lock can be established." `flock`-based locking is the right primitive here precisely because the kernel atomically arbitrates the contending opens — this is hopefully a non-issue on macOS/Linux, but worth verifying on Windows if MemPalace supports it.

Deferring final implementation choice to maintainer judgment — the structural ask is "make `mcp_server` refuse to second-start against a palace it doesn't own."

## What I haven't been able to isolate

- **Exact client trigger.** I can confirm from the Desktop log that two server processes existed concurrently, but I haven't been able to pin down which user-side action causes the second spawn. It is not consistently correlated with Desktop restarts, sleep/wake, or specific MCP tool calls in the captured window. A maintainer fix doesn't depend on knowing the trigger — the lock guard is correct regardless — but I want to flag the gap honestly.
- **Whether this reproduces on Claude Code.** The captured log is Desktop-only (`~/Library/Logs/Claude/`); Claude Code's MCP traffic logs to `~/.claude/projects/.../*.jsonl` instead and I haven't examined those for the same pattern. The lock guard would protect both.
- **Behavior post-#976.** The captured log ends 5 minutes before I pinned the install to PR #976's commit `0d9929c0`, so I don't have empirical evidence of how a `num_threads=1` build behaves under the same concurrent-spawn condition. Theory says the corruption is prevented but the resource waste / SQLite contention remains.

## Related upstream

- #974 — original HNSW SIGSEGV bug (closed by #976)
- #976 — `hnsw:num_threads=1` fix (merged to `develop`, awaiting v3.3.4)
- #1202 — Stop-hook concurrent-mine race (same bug family — concurrent writers — different trigger; that issue's per-palace lock approach is the precedent for this proposal)
- #1218 — `mempalace migrate/status` SIGSEGV on chromadb version mismatch (unrelated root cause but adjacent failure mode)
- #1222 — HNSW `max_elements` capacity issue (separate from this)

## Environment

```
mempalace 3.3.2 (uv tool install, ~/.local/share/uv/tools/mempalace/)
chromadb 1.5.x (whatever 3.3.2 pins)
Python 3.11
MCP protocol 2025-11-25
Claude Desktop (claude-ai client v0.1.0 per protocol handshake)
macOS Tahoe 26.3.1
Apple M4 Mac
```

Sanitized 30-hour Desktop MCP transport log: https://gist.github.com/Seph396/d8f724e58f066201b3cb527d0c7ffcc0

Vasanth19:
## Problem

Closes #1229.

The current `mempalace-mcp` / `mempalace.mcp_server` model runs **one Python process per agent session**. In practice, developers run Claude Code, Codex, Gemini CLI, and GG simultaneously. This exposes two compounding failure modes.

### Failure 1 — Zombie processes after SIGKILL

MCP hosts sometimes force-quit sessions with `SIGKILL`. Python's `atexit` and `signal.signal(SIGTERM)` cleanup never fires on `SIGKILL`, so the process exits without releasing the PID file. The next session sees a stale PID file, decides another instance is running, and refuses to start. Every MCP tool call returns "Connection closed". Manual PID-file removal is the only recovery — until it happens again.

### Failure 2 — Concurrent ChromaDB writers corrupt HNSW

When multiple sessions each hold an open `PersistentClient` against the same `chroma.sqlite3` and simultaneously call `upsert()`, the writes interleave at the mmap level. The in-memory HNSW tree and on-disk sqlite metadata diverge — exactly the divergence that #1222 detects but cannot prevent. The only safe fix is a **single process owning the ChromaDB connection**.

## Solution — Daemon + Bridge Architecture

```
macOS LaunchAgent
  └── mempalace-daemon.py   (one process, holds ChromaDB, listens on ~/.mempalace/mcp.sock)
        ├── Claude Code  ←→  mempalace-bridge.py  ←→  socket
        ├── Codex        ←→  mempalace-bridge.py  ←→  socket
        ├── Gemini CLI   ←→  mempalace-bridge.py  ←→  socket
        └── GG           ←→  mempalace-bridge.py  ←→  socket
```

- **No zombie problem.** If a session is SIGKILL'd, only the bridge dies. The daemon keeps running; the next session connects within milliseconds.
- **No concurrent writer corruption.** All `tools/call` requests are serialized through a `threading.Lock` inside the daemon. Protocol messages (`initialize`, `tools/list`, `ping`) remain lock-free.
- **Auto-start on first use.** The bridge detects a missing socket and starts the daemon automatically.
- **LaunchAgent keeps it alive.** If the daemon crashes, launchd restarts it within `ThrottleInterval` seconds (default: 5).

## Changes

Four files added — **no changes to the core `mempalace` package**:

| File | Description |
|------|-------------|
| `examples/mempalace-daemon.py` | Persistent Unix socket MCP server; the LaunchAgent target |
| `examples/mempalace-bridge.py` | ~60-line stdio↔socket relay; this is the MCP command each session uses |
| `examples/com.mempalace.daemon.plist` | macOS LaunchAgent template (edit paths, then `launchctl load`) |
| `docs/multi-session-daemon.md` | Full explanation of the problem, architecture, install steps, and MCP config examples for Claude Code, Codex, Gemini CLI, and any stdio client |

## Test Plan

- [ ] Confirm `mempalace-daemon.py` starts and creates `~/.mempalace/mcp.sock`
- [ ] Confirm `mempalace-bridge.py` connects and relays `tools/list` / `mempalace_status` correctly
- [ ] SIGKILL a bridge process; confirm daemon keeps running and next session connects cleanly
- [ ] Run two bridge sessions concurrently; confirm both `mempalace_add_drawer` calls succeed without HNSW divergence
- [ ] Unload LaunchAgent; confirm daemon stops; confirm bridge auto-restarts it on next connection attempt
- [ ] Verify no changes to existing `mcp_server.py` behavior for users not using the daemon

## Tested on

- macOS 14 Sonoma / macOS 15 Sequoia
- Python 3.11, 3.12
- MemPalace 3.3.x (ChromaDB 0.6.x)
- Concurrent sessions: Claude Code + Codex + Gemini CLI + GG

### DIFF
--- a/docs/multi-session-daemon.md
+++ b/docs/multi-session-daemon.md
@@ -1,0 +1,218 @@
+# Multi-Session MCP Daemon — Persistent MemPalace for All Your AI Agents
+
+**Fixes:** #1229 — zombie MCP server processes blocking sessions and corrupting ChromaDB
+
+## The Problem
+
+The default `mempalace-mcp` command (which calls `mempalace.mcp_server`) is designed
+to run as one process per agent session. When you run Claude Code, Codex, Gemini CLI,
+and GG simultaneously, each session spawns its own Python process — and each process
+holds an open ChromaDB `PersistentClient`.
+
+This creates two compounding failure modes:
+
+### 1. Zombie processes after SIGKILL
+
+MCP host applications (Claude Desktop, VS Code, terminal multiplexers) sometimes
+force-quit sessions with `SIGKILL` rather than `SIGTERM`. Python's `atexit` handlers
+and `signal.signal(SIGTERM, ...)` cleanup traps never fire on `SIGKILL`. The result
+is a Python process that exits without releasing its file descriptor on
+`~/.mempalace/mcp-server.pid`.
+
+On the next session start, a PID-file guard sees a stale PID file, decides another
+instance is running, and refuses to start. The session connects to nothing, and every
+MCP tool call returns "Connection closed".
+
+Removing the stale PID file by hand is the only recovery — until it happens again.
+
+### 2. Concurrent ChromaDB writers corrupt HNSW
+
+ChromaDB's HNSW index uses memory-mapped segment files. When two processes both hold
+a `PersistentClient` against the same `chroma.sqlite3` and both call `upsert()`, the
+writes interleave at the mmap level. This causes the in-memory HNSW tree and the on-disk
+sqlite metadata to diverge — exactly the divergence issue #1222 documents and detects.
+
+The `hnsw_capacity_status` probe (introduced in #1222) can detect this after the fact,
+but it cannot prevent it. The only safe fix is ensuring **a single process owns the
+ChromaDB connection** at all times.
+
+## The Solution: Daemon + Bridge Architecture
+
+Instead of each session spawning its own `mcp_server.py` process, a single **daemon**
+process runs continuously (managed by macOS LaunchAgent), holds the ChromaDB connection,
+and serves all sessions over a Unix socket. Each agent session runs a tiny **bridge**
+script that relays its stdio to/from the daemon socket.
+
+```
+macOS LaunchAgent
+  └── mempalace-daemon.py    (one process, holds ChromaDB, listens on ~/.mempalace/mcp.sock)
+        ├── Claude Code session  ←→  mempalace-bridge.py  ←→  socket
+        ├── Codex session        ←→  mempalace-bridge.py  ←→  socket
+        ├── Gemini CLI session   ←→  mempalace-bridge.py  ←→  socket
+        └── GG session           ←→  mempalace-bridge.py  ←→  socket
+```
+
+**Benefits:**
+
+- **No zombie problem.** If a session is SIGKILL'd, only the bridge dies. The daemon
+  keeps running; the socket stays open; the next session connects immediately.
+- **No concurrent writer corruption.** All `tools/call` requests are serialised through
+  a single `threading.Lock` inside the daemon. Protocol messages (`initialize`,
+  `tools/list`, `ping`) are lock-free for speed.
+- **Auto-start on first use.** The bridge detects a missing socket and starts the
+  daemon automatically, so you don't have to think about it.
+- **LaunchAgent keeps it alive.** If the daemon crashes (e.g. OOM), launchd restarts
+  it within `ThrottleInterval` seconds (default: 5).
+
+## Files
+
+Three files are provided in the `examples/` directory:
+
+| File | Purpose |
+|------|---------|
+| `mempalace-daemon.py` | The persistent server process — run once via LaunchAgent |
+| `mempalace-bridge.py` | Per-session stdio relay — this is the MCP command you configure |
+| `com.mempalace.daemon.plist` | macOS LaunchAgent template — edit paths, then `launchctl load` |
+
+## Installation
+
+### Step 1 — Copy the scripts
+
+```bash
+# Copy to wherever you keep your local tooling.
+# The bridge looks for the daemon script relative to its own location,
+# so keep both files in the same directory.
+cp examples/mempalace-daemon.py ~/bin/mempalace-daemon.py
+cp examples/mempalace-bridge.py ~/bin/mempalace-bridge.py
+chmod +x ~/bin/mempalace-daemon.py ~/bin/mempalace-bridge.py
+```
+
+### Step 2 — Edit the LaunchAgent plist
+
+Open `examples/com.mempalace.daemon.plist` and replace:
+
+- `/path/to/mempalace/venv/bin/python` — the Python interpreter in your MemPalace
+  virtualenv (run `which python` inside the venv to get the path)
+- `/path/to/mempalace-daemon.py` — the absolute path where you copied the daemon script
+- `YOUR_USERNAME` — your macOS username (`echo $USER`)
+
+### Step 3 — Install and load the LaunchAgent
+
+```bash
+cp examples/com.mempalace.daemon.plist ~/Library/LaunchAgents/com.mempalace.daemon.plist
+launchctl load ~/Library/LaunchAgents/com.mempalace.daemon.plist
+```
+
+Verify it started:
+
+```bash
+# The socket should exist within a few seconds
+ls -la ~/.mempalace/mcp.sock
+
+# Check the log
+tail -f ~/.mempalace/daemon.log
+```
+
+### Step 4 — Configure each agent to use the bridge
+
+Replace every `mempalace-mcp` / `mempalace.mcp_server` invocation with the bridge.
+
+#### Claude Code (`~/.claude.json`)
+
+```json
+{
+  "mcpServers": {
+    "mempalace": {
+      "type": "stdio",
+      "command": "/absolute/path/to/python",
+      "args": ["/absolute/path/to/mempalace-bridge.py"]
+    }
+  }
+}
+```
+
+Or register via CLI:
+
+```bash
+claude mcp add mempalace -- /path/to/python /path/to/mempalace-bridge.py
+```
+
+#### Codex (`~/.codex/config.toml`)
+
+```toml
+[mcp_servers.mempalace]
+command = "/absolute/path/to/python"
+args    = ["/absolute/path/to/mempalace-bridge.py"]
+```
+
+#### Gemini CLI (`~/.gemini/settings.json`)
+
+```json
+{
+  "mcpServers": {
+    "mempalace": {
+      "command": "/absolute/path/to/python",
+      "args": ["/absolute/path/to/mempalace-bridge.py"]
+    }
+  }
+}
+```
+
+Or via CLI:
+
+```bash
+gemini mcp add mempalace /absolute/path/to/python /absolute/path/to/mempalace-bridge.py --scope user
+```
+
+#### Any other stdio MCP client
+
+The bridge is a generic stdio relay. Any client that accepts a `command` + `args`
+MCP configuration can use it:
+
+```
+command: /path/to/python
+args:    ["/path/to/mempalace-bridge.py"]
+```
+
+## Troubleshooting
+
+### "MemPalace daemon not reachable"
+
+The bridge tried 20 times (5 seconds total) to connect and failed.
+
+1. Check the daemon log: `tail ~/.mempalace/daemon.log`
+2. Check launchd status: `launchctl list | grep mempalace`
+3. Try starting the daemon manually to see startup errors:
+   ```bash
+   /path/to/python /path/to/mempalace-daemon.py
+   ```
+
+### LaunchAgent not starting
+
+- Verify plist syntax: `plutil ~/Library/LaunchAgents/com.mempalace.daemon.plist`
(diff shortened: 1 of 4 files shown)


igorls:
Moved to v3.4 milestone — this is a feature-scale change (new long-running daemon process model + multi-session access pattern), not a maintenance fix. v3.3.5 is scoped to stability/bug fixes only. Companion issue #1229 moved with it.

Continuing review here is welcome; we'll target landing this in the v3.4 cycle once v3.3.5 ships.

marcoaperez:
Real-world testimony — running these 4 files unmodified on macOS (M-series, bash 3.2) since 2026-05-05 (9 days) alongside `mempalace==3.3.4` from pipx. Setup: `com.mempalace.daemon.plist` keeps the daemon up; 2 concurrent Claude Code sessions plus their MCP servers hit it through the bridge.

`~/.mempalace/daemon.log` in that window:
- 15 clean daemon restarts, zero hung shutdowns
- 13 `flush-lag, not corruption` events correctly classified by the daemon's heuristic (segment metadata intact → no quarantine) — the exact false-positive class that would have triggered unnecessary repair pre-#1270
- 0 HNSW corruption, 0 SIGSEGV, 0 SQLite lock errors
- 15,544 drawers across 8 productive wings, growing daily via Stop/PreCompact hooks

The multi-session bridge is the load-bearing piece — without it, the two concurrent MCP server processes race exactly as #1229 describes.

Not asking to bump the milestone, just adding a field signal that the design works end-to-end on a non-trivial palace. Happy to share logs / plist / anything else if it helps review.

marcoaperez:
Heads-up for anyone following this PR: I've been running this exact daemon design in production (it's a great fix for the concurrent-writer HNSW corruption). But the peer-writer guard merged in 3.5.0 (#1818 / #1823) holds `mine_palace_lock` for the whole process lifetime, which collides with this topology — it locks out hook/manual `mempalace mine` runs and its re-entrancy is per-thread while a daemon serves clients on many threads. Details + suggestions in #1888. Would be great to land #1270 and the guard in a way that lets them coexist.

fatkobra:
Hi @Vasanth19 — I’ve reworked #1976 so it complements this PR rather than introducing a competing standalone daemon.

The revised implementation keeps #1270’s daemon + bridge architecture, but uses the existing package-level `mempalace.daemon` queue/server as the owner:

- `mempalace-mcp` becomes a stdio bridge into the existing daemon;
- the daemon exposes a token-authenticated `/mcp` endpoint;
- MCP `tools/call` requests and existing daemon queue jobs share one in-process writer gate;
- the raw stdio server remains available as `mempalace-mcp-stdio`;
- the low-level `mine_palace_lock()` fail-fast behavior remains unchanged;
- focused tests cover bridge forwarding, request serialization, identity mismatch handling, and daemon client behavior.

I removed the separate `mempalace.mcp_daemon` implementation, so this is now intended as package integration and automated-test coverage for the direction proposed here.

There is still rollout work before the full Tier 3 goal is complete: hook and CLI workflows need to consistently use daemon-backed execution rather than direct writer paths.

I’d appreciate your feedback on whether this division makes sense and whether any parts should be moved into, or coordinated more closely with, #1270.

fatkobra:
Follow-up to my earlier comment: the package-integration and writer-routing work discussed there has now been implemented as a staged PR stack.

Current pieces:

- #1976 — package-level MCP stdio bridge into the existing `mempalace.daemon`, including shared MCP/daemon writer serialization, authoritative backend/collection identity, and installed-command E2E coverage;
- #2027 — shared `direct` / `prefer` / `require` write-routing policy;
- #2030 — applies that policy to hook-triggered writes;
- #2033 — applies it to routine CLI writes such as `mine`, `sweep`, `sync`, and the optional post-`init` mine.

The routing PRs intentionally preserve the current `direct` default. They provide an explicit `require` mode that never silently falls back to a direct ChromaDB writer.

The stack is currently:

    #2027
      -> #2030
          -> #2033

All of these are intended to complement the daemon + bridge direction proposed in #1270, not to erase its design history or compete with it.

I’m leaving the branches unchanged while maintainers decide which architectural direction and merge sequence they prefer.

PR #1271, opened 2026-04-29
feat(backends): pickle-counts gate in quarantine_stale_hnsw

sha2fiddy:
## Problem

`quarantine_stale_hnsw` misses one corruption shape: `index_metadata.pickle` where `len(id_to_label) != total_elements_added`. Stage 2's byte-sniff sees the PROTO + STOP markers and returns healthy, so the segment stays in place. HNSW traversal eventually dereferences the unmapped label and SIGSEGVs.

Same shape as [chroma-core/chroma#6979](https://github.com/chroma-core/chroma/issues/6979). Hit on a real 522k-drawer palace 2026-04-28; recovered with #1126's `repair --mode hnsw`. This gate would have caught it before the crash.

## Fix

A stage 3 check runs only when stage 2 passes: deserialize `index_metadata.pickle` and compare `len(id_to_label) == total_elements_added`.

- If counts disagree, quarantine to `<uuid>.drift-<timestamp>` (same path as the stage 2 failure).
- If counts agree, leave in place.
- If the unpickler refuses, keys are missing, or types are unexpected, return undetermined and let stage 2's verdict stand. Older 0.6.x palaces with attr-object metadata are not quarantined by this gate alone.

`_BoundedUnpickler` refuses any non-builtin `find_class`, so a structurally-valid-but-malicious pickle cannot execute code at quarantine time.

## Tests

8 new in `tests/test_backends.py`:

- Unit (`_segment_pickle_counts_consistent`): matching, off-by-one, class-pickle (refused), missing keys, missing file.
- Integration (`quarantine_stale_hnsw`): quarantine on off-by-one, leave-alone on consistent counts, leave-alone on byte-sniff-pass + unparseable (the 0.6.x path).

53/53 in `test_backends.py` and 1469/1469 in the full suite pass. ruff clean. No dependency or lockfile changes.

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -517,7 +517,6 @@ class _SafePersistentDataUnpickler:
 
     @classmethod
     def load(cls, path: str):
-
         class _Restricted(pickle.Unpickler):
             def find_class(self, module: str, name: str):
                 if (module, name) in cls._ALLOWED:
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -219,23 +219,128 @@ def _segment_appears_healthy(seg_dir: str) -> bool:
     return len(head) == 2 and head[0] == 0x80 and tail == b"\x2e"
 
 
+class _BoundedUnpickler(pickle.Unpickler):
+    """Unpickler that refuses any class instantiation.
+
+    ``index_metadata.pickle`` written by chromadb 1.5.x is a plain dict of
+    builtins (``id_to_label`` is a ``dict``, ``total_elements_added`` is
+    an ``int``, etc.). Refusing ``find_class`` for anything else means a
+    structurally-valid-but-malicious pickle cannot execute arbitrary code
+    at quarantine-check time. Older 0.6.x palaces that stored the
+    metadata as a custom-class instance fall back to "unknown" — the
+    existing byte-sniff verdict stands.
+    """
+
+    _ALLOWED_BUILTINS = frozenset(
+        {"dict", "int", "str", "tuple", "list", "set", "frozenset", "bytes", "bool"}
+    )
+
+    def find_class(self, module: str, name: str) -> Any:
+        if module == "builtins" and name in self._ALLOWED_BUILTINS:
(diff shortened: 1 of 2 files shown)


igorls:
Moved to v3.4 milestone — the pickle-counts gate is a behavior enhancement to `quarantine_stale_hnsw` (new gating logic), not a bug fix. v3.3.5 stays scoped to stability/bug fixes. Will land in the v3.4 cycle.

igorls:
Good intent (guarding chroma-core/chroma#6979 off-by-one before a SIGSEGV) and it's nicely crash-safe (rename, not delete) — but the stage-3 gate is a no-op on real palaces:

**Blocking — `_BoundedUnpickler` refuses the actual on-disk object.** chromadb 1.5.7 pickles `index_metadata.pickle` as a `PersistentData` *class instance* (no `__reduce__`, not a dict), but `_BoundedUnpickler.find_class` only allows `module == "builtins"`, so it raises, `_segment_pickle_counts_consistent` returns `None` (undetermined) → no quarantine. Verified empirically: an off-by-one `PersistentData` instance → `None`; only a plain **dict** returns `False`. So on the real 1.5.x corruption you cite, stage-3 does nothing.

**Blocking — the tests mask it.** `_write_meta_pickle` pickles a plain dict, which is the one shape that never hits disk, so the off-by-one tests pass against a path that can't occur in production. `test_..._returns_none_for_class_pickle` actually uses the *real* shape and asserts `None` (no-op).

**Fix:** reuse the existing `_SafePersistentDataUnpickler` (it already allowlists `PersistentData`) + the shape-agnostic `_persisted_metadata_value` to read the counts, and add a test that builds a real `PersistentData`-instance pickle with an off-by-one and asserts quarantine. (FYI the docstring at `chroma.py:556-560` claiming "1.5.x writes a plain dict via `__reduce_ex__`" is inaccurate and probably what led here.)

To be fair, I checked the "this duplicates existing machinery" angle and it's actually **new** coverage: the existing `int(total) != label_count` check is gated behind `dimensionality is None`, so valid-dimensionality off-by-one segments aren't covered today. The feature is worth having — it just needs to read the real shape.

PR #1276, opened 2026-04-30
feat(repair): --mode reconcile to re-embed SQL-only orphan rows

sha2fiddy:
## What

`mempalace repair --mode reconcile --segment <uuid>` finds drawers that have rows in `embedding_metadata` but no HNSW label, re-embeds them, rebuilds the segment with old + new vectors, and atomic-swaps it in.

Hit my own 533k-drawer palace 2026-04-29: one 04-28 mining run left 10,404 SQL-only orphans. chromadb crashed mid-flush, SQL committed transactionally, HNSW additions lost. PR #1271's pickle-counts gate doesn't catch this because the pickle was internally consistent for what actually landed.

## Tradeoffs

- This stacks on #1126; it reuses `_extract_vectors`, `_build_persistent_index`, `_atomic_swap_segment`. Inlining would have been ~500 LOC duplication. Cost of stacking is that this can't merge until #1126 does.
- Rebuilds from scratch rather than appending to the live index. Same swap pattern #1126 used today on `bdc58f8f`. Modifying the live pickle in place is exactly the failure mode I'm trying not to repeat.
- Auto-detects the sibling METADATA segment, since each VECTOR collection always has exactly one METADATA sibling. The `--metadata-segment` flag handles the rare ambiguous case.

## Tests

6 new in `test_repair.py`: auto-detect, dry-run, happy-path append, no-orphans short-circuit, off-by-one pickle abort, missing-METADATA abort. 54/54 in `test_repair.py`, 1349/1349 full suite, ruff clean, no new deps.

Verified against my live 533,342-drawer palace. Strict zero-diff after the run, and `mempalace search` returns reconciled drawers as top hits.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -832,6 +832,23 @@ def cmd_repair(args):
         )
         return
 
+    if getattr(args, "mode", "legacy") == "reconcile":
+        if not getattr(args, "segment", None):
+            print("  --mode reconcile requires --segment <uuid>")
+            return
+        from .repair import reconcile_orphan_sql_rows
+
+        reconcile_orphan_sql_rows(
+            palace_path,
+            segment=args.segment,
+            metadata_segment=getattr(args, "metadata_segment", None),
+            max_elements=getattr(args, "max_elements", None),
+            backup=getattr(args, "backup", True),
+            dry_run=getattr(args, "dry_run", False),
+            assume_yes=getattr(args, "yes", False),
+        )
+        return
+
     if getattr(args, "mode", "legacy") == "from-sqlite":
         from .migrate import confirm_destructive_action
         from .repair import RebuildPartialError, rebuild_from_sqlite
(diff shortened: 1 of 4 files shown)


sha2fiddy:
Rebased on upstream/develop @fdfaf01, conflict resolved in `mempalace/cli.py` (argparse merge of `--mode reconcile` + `--metadata-segment` alongside upstream `--mode max-seq-id` + `--from-sidecar`). One ruff-format reflow folded into top commit. 120/120 targeted tests pass, ruff clean. Marked draft until CI runs green.

igorls:
Moved to v3.4 milestone — `--mode reconcile` is an additive feature (new repair mode that re-embeds SQL-only orphan rows). v3.3.5 is scoped to maintenance / bug fixes; new modes belong in the next feature release. Review continues — we'll merge in the v3.4 cycle.

igorls:

PR #1285, opened 2026-04-30
fix: harden Chroma repair preflight and rollback recovery

mjc:
## What does this PR do?

Hardens the Chroma repair path so a bad or partially corrupted palace does not get worse during startup or recovery.

This PR:
- validates and quarantines invalid HNSW metadata before reopening a client
- skips destructive quarantine for transient truncated pickle reads during reconnects
- rebuilds collections through a temporary staging collection so a failed rebuild does not touch the live index
- restores from backup if repair fails after the live collection has already been replaced
- makes the CLI exit non-zero when repair recovery fails
- preserves the original rebuild error if later cleanup also fails
- closes the active backend before rollback restore so file handles are released
- makes CLI repair restore use the live backend instance rather than a fresh helper instance

This follows up on PR #1124 by covering additional startup, reconnect, rebuild, and rollback failure modes found in practice, without pulling in that PR's unrelated Windows CLI encoding and status pagination changes.

## How to test

1. Start with a palace whose Chroma HNSW metadata is malformed or partially corrupted, then start MemPalace and confirm startup quarantines the invalid state instead of reopening it.
2. Simulate a transient truncated pickle read during reconnect and confirm it is treated as a transient read failure rather than triggering quarantine.
3. Force a rebuild failure while repairing a collection and confirm the live collection remains untouched because rebuild happens through a staging collection.
4. Force a failure after live replacement during repair and confirm rollback restores the backup successfully.
5. Run the CLI repair path through a recovery failure and confirm it exits non-zero instead of reporting success.


mjc:
Rebased onto current upstream/develop to pick up the recent HNSW threshold and Chroma reopen fixes. The only semantic conflict was the missing index_metadata.pickle path: I kept upstream’s dynamic sync_threshold scaling, but preserved this PR’s behavior that missing pickle metadata is inconclusive and should not disable vector search by itself.

igorls:
Maintainer-edit pushed to keep this on track for v3.3.5:

- Merged \`develop\` in. Two real conflicts: \`mempalace/backends/chroma.py\` (post-#1162 lock-plumbing era refactored the per-open pass into \`ChromaBackend._prepare_palace_for_open\`, which collided with this branch's inline-expansion form that added \`quarantine_invalid_hnsw_metadata\`) and a trivial import collision in \`tests/test_backends.py\`.
- Resolution: kept the \`_prepare_palace_for_open\` helper from develop and folded \`quarantine_invalid_hnsw_metadata\` into it, gated by the same \`_quarantined_paths\` set. Kept the \`if inode_changed: _quarantined_paths.discard\` guard from this branch in \`_client\` so a fresh inode re-runs quarantine.
- Linux full suite: 1579 passed, ruff lint+format clean against the 0.4.x CI pin.

Now also has \`quarantine_invalid_hnsw_metadata\` reachable from every open path (CLI mining, search, repair, status), not just \`make_client\`. Watching CI for the cross-platform jobs.

PR #1286, opened 2026-04-30
fix(mcp_server): log exception + retry once on _get_collection failure

jphein:
A small improvement to `_get_collection`: log the exception on failure (instead of returning `None` silently) and retry once after clearing the cached client and collection. Behaviour on the happy path is unchanged.

## Why

The current shape (`try: ... except Exception: return None`) has two costs in practice on a long-running deployment:

1. **Invisible incidents.** When a transient ChromaDB error poisons the cached `_client_cache` / `_collection_cache`, every subsequent call returns `None` but the operator never sees the underlying exception in the log. Debugging has to start from the symptoms — 404s, "no palace" responses — without the actual stack.
2. **No self-healing.** Once the cache is stale, the only way back to a working state is a process restart. A single retry that clears the caches first resolves the most common shape (a transient error left bad state behind) without operator intervention.

## What changes

```python
for attempt in range(2):
    try:
        client = _get_client()
        # ...same body as today...
        return _collection_cache
    except Exception as e:
        logger.error(
            "_get_collection attempt %d failed (palace=%s): %s",
            attempt + 1, _config.palace_path, e,
        )
        if attempt == 0:
            _client_cache = None
            _collection_cache = None
            _metadata_cache = None
            _metadata_cache_time = 0
return None
```

- One `logger.error(...)` per failed attempt, including the palace path so multi-palace deployments can tell which one is hurting.
- After the first failure, all four caches reset before the second attempt — `_client_cache` is now in the `global` declaration so the second attempt actually reopens the client from scratch.
- The function still returns `None` if both attempts fail. Existing callers that handle `None` (there are several throughout `mcp_server.py`) keep working unchanged.
- Happy path: zero log lines, no behaviour change. The first iteration returns immediately on success.

## What's intentionally not in this PR

- The HNSW `num_threads=1` retrofit and `_pin_hnsw_threads()` call inside the function are untouched. That work landed separately and is unrelated.
- No new dependencies, no new config keys, no test changes. The retry path is timing-dependent (a transient failure followed by a recovery) and hard to fake reliably; happy to add a unit test if there's a preferred shape (`mock.patch` on `_get_client` raising once?).

## Where it ran

This has been running as a local patch on top of the pipx install on a 151K-drawer palace fronting `palace-daemon` since 2026-04-25. The retry path fired twice during HNSW segment-quarantine windows after a daemon restart — both times the second attempt succeeded and traffic kept flowing without intervention. No false-positive retries observed (the happy path takes the `return _collection_cache` exit on the first iteration).

I've been carrying this as a sed-applied patch in `palace-daemon/patches/`; happy to retire that and just depend on a release line that has this in it once it lands. Equally happy to rework the shape (different log level, structured logging, no retry at all and just the log) — let me know what fits best.

Thanks for taking a look!


jphein:
Pushed [`757eb7a`](https://github.com/MemPalace/mempalace/pull/1286/commits/757eb7a) addressing both Copilot notes:

- Switched `logger.error("...", str(e))` to `logger.exception("...")` so the traceback lands in the daemon log alongside the message instead of being dropped. Same context lines, just with `exc_info` from `sys.exc_info()` automatically.
- Added `TestGetCollectionRetry` (`tests/test_mcp_server.py`) with three monkeypatch tests that pin the new semantics:
  - `test_retry_succeeds_on_second_attempt` — first `_get_client` raises, caches are cleared between attempts (verified by inspecting planted sentinels in `_client_cache` / `_collection_cache`), second attempt returns the wrapped collection.
  - `test_returns_none_when_both_attempts_fail` — two failures in a row return `None` and stop after exactly two attempts.
  - `test_happy_path_does_not_retry` — a successful first attempt returns immediately and never invokes `_get_client` a second time.

  All three use `monkeypatch` on `_get_client` / `ChromaCollection` / `_pin_hnsw_threads` so they don't need a real ChromaDB; they run in ~0.4s end-to-end and don't touch disk.

Also fixed a `ruff format` lint failure from the previous push (multi-arg `logger.exception` call needed each arg on its own line under the project's ruff config). Full CI matrix now green — test-linux 3.9/3.11/3.13, test-macos, test-windows, lint. Thanks for the careful review.

igorls:
Thanks @jphein — retry-once with cache-clear on `_get_collection` is correct and complementary to the recent #1303/#1299/#1289 work (those fixed correctness on first open; this adds resilience after a transient failure). Local tests for `TestGetCollectionRetry` pass.

One blocker: **rebase against `develop`** — the diff was authored against the pre-#1303 body of `_get_collection`. Develop's current implementation (mcp_server.py ~L276-338) does EF resolution via `ChromaBackend._resolve_embedding_function()` and a `get_collection`/`create_collection` split guarded by `_ChromaNotFoundError`. The retry loop needs to wrap *that* shape, not `get_or_create_collection`. CONFLICTING right now.

Once rebased and CI is green we can merge for v3.3.5.

jphein:
Rebased onto develop after #1289 and #1303 merged. The retry/log envelope is unchanged; the inner body now sits on top of the get-then-create + `embedding_function=` plumbing those PRs landed, so retries benefit from the corrected call shape rather than re-attempting a known-bad pattern. CI green locally, ready for another look when convenient.

igorls:
Hi @jphein — thanks for carrying this; the underlying retry+log idea is sound and composes cleanly with #1322 (clearing `_client_cache` on attempt 1 forces a fresh `make_client()` which re-runs `quarantine_stale_hnsw`).

I'd like to land that piece for v3.3.5 (due 2026-05-08), but this PR can't merge as-is — it carries fork-only files (`FORK_CHANGELOG.md`, `docs/fork-changes.yaml`, `docs/superpowers/**`, `scripts/{deploy,preflight,rebase-on-develop}.sh`), bumps `version.py` / `pyproject.toml` to 3.3.4 without the matching `marketplace.json` / `plugin.json` / `codex-plugin/plugin.json` (which is what `check-versions` is catching), and the `convo_miner.py` rewrite regresses `test_convo_mining` + `test_mine_convos_rebuilds_stale_drawers_after_schema_bump` on all three OSes (Linux/macOS/Windows go red on those two tests where `develop` is green).

I opened **#1377** with just the surgical `_get_collection` retry, credited as `Co-authored-by: Jeffrey Hein <jp@jphein.com>`. It includes two unit tests for the retry path and ships the cache reset between attempts so a stale handle self-heals on the second call.

Closing this one in favour of #1377 — sorry for the friction. The other genuinely-useful pieces buried in this PR are very welcome as separate follow-ups against current develop:

- **Session-recovery collection split** in `mcp_server.py` (routing diary checkpoints to a separate collection so they don't dominate search) — substantial enough to deserve its own design discussion / PR
- **`backends/chroma.py` None-metadata coercion** at the backend boundary (would generalise the per-site guards from #999 / #1013 / #1094)
- **Closet-boost ranking refactor in `searcher.py`** with hoisted tunables

If you'd rather I cherry-pick any of those into surgical PRs the way I did with the retry, ping me and I will. Either way thanks for the diagnosis on the underlying bug — easy to miss.

PR #1280, opened 2026-04-30
Fix/windows hook stdio utf8

yangshare:
这个修复把 #363/#400 的 Windows UTF-8 stdin 处理扩展到 Claude/Codex hook 入口，避免非 ASCII hook payload 在 json.load(sys.stdin) 前被系统 ANSI codepage
  解码坏。

PR #1291, opened 2026-05-01
fix: Windows 中文编码修复 - MCP server UTF-8 支持

hzx945627450-eng:
修复 Windows 环境下 MCP server 中文乱码问题。

问题根因：
- Windows 默认 stdin/stdout 使用 GBK/CP936 编码
- MCP 协议要求 UTF-8 传输
- 导致中文字符在 JSON-RPC 通信中损坏

修复方案（mempalace/mcp_server.py）：
1. 第 1889 行：工具返回结果添加 ensure_ascii=False
2. 第 1949 行：JSON-RPC 响应添加 ensure_ascii=False
3. main() 函数：Windows 平台强制包装 stdin/stdout 为 UTF-8

跨平台兼容：
- 使用 sys.platform == 'win32' 检测
- Linux/macOS 默认 UTF-8，无需包装
- ensure_ascii=False 对所有平台都有益

测试结果：
- ✅ Windows: 中文读写完全正常
- ✅ Linux/macOS: 不受影响，保持原有 UTF-8 行为

## Checklist
- [ ] No hardcoded paths

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -11,7 +11,7 @@ import os
 import re
 import subprocess
 import sys
-from datetime import datetime
+import datetime as _dt
 from pathlib import Path
 
 SAVE_INTERVAL = 15
@@ -153,7 +153,7 @@ def _log(message: str):
             _state_dir_initialized = True
         log_path = STATE_DIR / "hook.log"
         is_new = not log_path.exists()
-        timestamp = datetime.now().strftime("%H:%M:%S")
+        timestamp = _dt.datetime.now().strftime("%H:%M:%S")
         with open(log_path, "a") as f:
             f.write(f"[{timestamp}] {message}\n")
         if is_new:
@@ -658,7 +658,7 @@ def hook_stop(data: dict, harness: str):
 
 
 def hook_session_start(data: dict, harness: str):
-    """Session start hook: initialize session tracking state."""
+    """Session start hook: warm up MemPalace backend."""
     parsed = _parse_harness_input(data, harness)
     session_id = parsed["session_id"]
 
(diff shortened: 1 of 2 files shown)


igorls:
Thanks for this fix.

The stdio reconfigure portion of this PR overlaps with #1259, which addresses the same Windows encoding problem using `sys.stdin.reconfigure(encoding="utf-8")`. Both approaches work. The `reconfigure()` idiom is what the other Windows stdio PRs in flight (#1280, #1282) are using, so we are likely to take #1259 to keep the codebase consistent.

That said, the `ensure_ascii=False` additions on the two `json.dumps` calls in this PR are independently valuable and not present in #1259. Without them, non-ASCII content in tool results is emitted as `\uXXXX` escapes, which is verbose and can trip downstream MCP clients that do not unescape correctly.

Would you be willing to open a separate PR with just the `ensure_ascii=False` change, plus a small test that round-trips a Chinese string through `handle_request` so the fix has a regression guard? That change can land on its own merits and stay decoupled from the stdio-reconfigure work.

hzx945627450-eng:
> Thanks for this fix.
> 
> The stdio reconfigure portion of this PR overlaps with #1259, which addresses the same Windows encoding problem using `sys.stdin.reconfigure(encoding="utf-8")`. Both approaches work. The `reconfigure()` idiom is what the other Windows stdio PRs in flight (#1280, #1282) are using, so we are likely to take #1259 to keep the codebase consistent.
> 
> That said, the `ensure_ascii=False` additions on the two `json.dumps` calls in this PR are independently valuable and not present in #1259. Without them, non-ASCII content in tool results is emitted as `\uXXXX` escapes, which is verbose and can trip downstream MCP clients that do not unescape correctly.
> 
> Would you be willing to open a separate PR with just the `ensure_ascii=False` change, plus a small test that round-trips a Chinese string through `handle_request` so the fix has a regression guard? That change can land on its own merits and stay decoupled from the stdio-reconfigure work.

Thanks for the clear and helpful feedback! 

I've split the changes as you suggested into one separate PR:
- PR #1293: `ensure_ascii=False` for all platforms

I'll close this original PR in favor of the split one, to keep things clean and decoupled as you recommended.

Issue #1275, opened 2026-04-30
silent_save config flag is a no-op — hooks_cli.py never reads it


contabilidad-source:
## Summary

`mempalace_hook_settings(silent_save=True)` returns success and persists the value in config, but the stop and precompact hooks continue to emit `{"decision": "block", "reason": "AUTO-SAVE checkpoint..."}` exactly as before. The toggle is dead code in v3.3.x — the config getter exists but no hook handler consumes it.

## Repro

Tested on plugin v3.3.3 / pip package v3.1.0 (Python 3.14, Windows 11, Claude Code 2.x).

```
# 1. Confirm baseline behavior — stop hook nudges every 15 messages.
# 2. Call MCP tool:
mempalace_hook_settings(silent_save=true)
# Returns: {"success": true, "settings": {"silent_save": true, "desktop_toast": false}, "updated": ["silent_save → True"]}

# 3. Trigger another stop after 15 more human messages.
# Expected: hook returns {} (silent direct save).
# Actual: hook still returns {"decision": "block", "reason": STOP_BLOCK_REASON} — identical to silent_save=false.
```

## Root cause

`silent_save` is wired into config storage and the MCP setter/getter but never consulted by the actual hook implementations:

- **Config layer** — `mempalace/config.py:177` defines `Config.hook_silent_save` (defaults to `True`).
- **MCP layer** — `mempalace/mcp_server.py:1006-1041` exposes `tool_hook_settings()` which reads/writes the flag correctly.
- **Hook layer** — `mempalace/hooks_cli.py` (the script `bash mempal-stop-hook.sh` ultimately invokes via `mempalace hook run --hook stop --harness claude-code`) **never imports `Config` and never branches on `silent_save`**. The `hook_stop()` function (lines 133-176) emits `{"decision": "block", "reason": STOP_BLOCK_REASON}` whenever `since_last >= SAVE_INTERVAL`, regardless of any config. `hook_precompact()` (lines 193-216) is even stricter — always blocks.

`grep -rn "silent_save" mempalace/` returns hits only in `config.py` and `mcp_server.py` — zero references in `hooks_cli.py`.

## Suggested fix

Honor the flag in both hook handlers. Minimal patch sketch:

```python
# hooks_cli.py
from mempalace.config import Config  # or wherever it lives

def hook_stop(data, harness):
    ...
    if since_last >= SAVE_INTERVAL and exchange_count > 0:
        try:
[...]

Issue #1212, opened 2026-04-26
Stop hook spawns concurrent `mempalace mine` processes that survive across sessions and bypass PID guard


ttessarolo:
## Summary

  The Claude Code Stop hook (`hooks/mempal-stop-hook.sh` → `mempalace hook run --hook stop`) spawns `mempalace mine` subprocesses that:

  1. Run **concurrently** for the same source path, in spite of the PID-file guard added in #1023.
  2. **Survive across Claude Code sessions** as orphaned processes — they remain alive after the IDE that triggered them is closed and after the parent shell exits.
  3. **Re-spawn within seconds** of being killed, as long as the IDE is still running and the hook is still wired up — making manual cleanup ineffective without disabling the binary itself.

  Combined with #344 (HNSW link_lists.bin unbounded growth, fixed in #1191) and the absence of a "max concurrent mines" cap, this can fill a 1.8 TB disk in under two hours of normal IDE use.

  ## Environment

  - MemPalace: **3.3.3** (also reproducible on 3.3.2)
  - ChromaDB: **1.5.8**
  - Claude Code: **2.1.119**
  - Plugin: `~/.claude/plugins/cache/mempalace/mempalace/3.3.2` (hooks/hooks.json registers `Stop` and `PreCompact`)
  - OS: macOS 15 (Darwin 25.3.0), Apple Silicon
  - Source dir mined: `~/.claude/projects/-Users-...-recipe-as-graph` (~684 MB of jsonl session transcripts)

  ## Steps to Reproduce

  1. Use Claude Code with the MemPalace plugin enabled.
  2. Open a project that generates substantial transcript volume in `~/.claude/projects/<encoded-path>/`.
  3. Let the Stop hook fire normally over a long session.
  4. Run `pgrep -fl 'mempalace mine'` periodically.

  ## Observed Behavior

  After ~1 hour of normal use:

  $ pgrep -fl 'mempalace mine'
  1568  python -m mempalace mine /Users/.../-Users-...-recipe-as-graph --mode convos --wing sessions
  1569  python -m mempalace mine /Users/.../-Users-...-recipe-as-graph

  Two `mempalace mine` processes against the same source path, both alive at the same time, both writing to the same palace. `~/.mempalace/hook_state/mine.pid` contains the PID of one of them; the other started
  anyway.

  After `kill -9 1568 1569`:

  $ pgrep -fl 'mempalace mine'
  45810  python -m mempalace mine ... --mode convos --wing sessions
[...]

PR #1305, opened 2026-05-02
fix(hooks): treat absent ~/.mempalace as auto-save off

lcatlett:
## Summary

When the user removes `~/.mempalace/` (a strong "do not auto-capture" signal), the next hook fire silently recreates the entire dir and ingests existing transcripts. This PR makes the absence of `~/.mempalace/` a kill-switch across all hooks.

Caught in the wild: ~146 drawers materialized in <1s after a deliberate `rm -rf ~/.mempalace/`, into a session that was explicitly not meant to be captured.

## Root cause (three layers)

1. **`_log()` at `hooks_cli.py:148`** unconditionally calls `STATE_DIR.mkdir(parents=True, exist_ok=True)`. So writing the hook log line itself recreates `~/.mempalace/hook_state/`.
2. **With no config file present, hook_stop_auto_save / hook_precompact_auto_save default to `True`** (no override to read). The fail-open semantics treat "no config" as "save enabled" — exactly backwards when the entire dir was just deleted.
3. **`hook_session_start` has no toggle at all** and unconditionally `_log()`s + `STATE_DIR.mkdir()`s.

## Fix

All four entry points (`hook_stop`, `hook_precompact`, `hook_session_start`, and `_log` itself) now check a new `PALACE_ROOT = Path.home() / ".mempalace"` constant and short-circuit (returning `{}` on stdout, never logging) when the directory is absent. The user-removable directory is now a kill-switch.

`_output()` writes directly to stdout fd 1 — no disk side effect. So the short-circuit response can be sent without recreating anything.

## Tests

5 unit tests in `tests/test_hooks_cli.py`:
- `test_hook_stop_does_not_create_palace_dir_when_absent`
- `test_hook_precompact_does_not_create_palace_dir_when_absent`
- `test_hook_session_start_does_not_create_palace_dir_when_absent`
- `test_log_does_not_create_palace_dir_when_absent`
- `test_existing_dir_proceeds_normally` (regression)

Full suite: 1469 passed, 1 skipped → 1474 passed, 1 skipped (+5, 0 regressions).

## Related

- #1212 (concurrent mine processes) — same family of "hooks doing things users didn't authorize"
- #1083 (polluted mega-wing) — symptom-adjacent
- #1275 (silent_save flag is a no-op) — same pattern of config-not-respected

Not a duplicate of any of the above, but adjacent.

## Test plan

- [x] `python3 -m pytest tests/test_hooks_cli.py -k "palace_dir or proceeds_normally" -v`
- [x] Full suite: `python3 -m pytest tests/ --ignore=tests/benchmarks -q`
- [x] Live: `rm -rf ~/.mempalace`, fire all three hooks, confirm `~/.mempalace` stays absent


Qodo-Free-For-OSS:
Hi, The new kill-switch treats any existing `~/.mempalace` path (including a file) as “enabled”, so hooks can proceed and then crash on `STATE_DIR.mkdir()` because the parent is not a directory.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Use is_dir for PALACE_ROOT

> ### Issue description
> Hooks should only proceed when `~/.mempalace` is an actual directory. Using `Path.exists()` allows a file at that path to bypass the kill-switch and later crash when `STATE_DIR.mkdir()` runs.
> 
> ### Issue Context
> - `_palace_root_exists()` is the new gate for all hook entrypoints.
> - Several hook paths call `STATE_DIR.mkdir(...)` without catching `OSError`.
> 
> ### Fix Focus Areas
> - mempalace/hooks_cli.py[22-30]
> - mempalace/hooks_cli.py[155-179]
> - mempalace/hooks_cli.py[565-715]

---

igorls:
Thanks @lcatlett — small, correct, and the new tests look good. Two changes before merge:

**1. Rebase against `develop`** — currently DIRTY (CHANGELOG.md likely conflicts).

**2. Address the Qodo bot comment** — `Path.exists()` returns True if a file (not a directory) sits at `~/.mempalace`, which would let later `STATE_DIR.mkdir()` crash. One-line hardening:

```python
def _palace_root_exists() -> bool:
    return PALACE_ROOT.is_dir()
```

Once both are in, I'll authorize CI on the fork PR and we can land it for v3.3.5.

PR #1312, opened 2026-05-02
fix: use configured collection in repair recovery paths

mjc:
## What does this PR do?

This PR is intentionally stacked on #1285. The first five commits are from #1285 and are included only so this branch can be reviewed/tested against the repair staging and rollback implementation already queued for v3.3.5. They are not meant to be reviewed as new work here.

The new work in this PR is the final commit, `34282d0`, which fixes configured drawer collection handling in recovery paths.

- Uses `MempalaceConfig().collection_name` consistently across `palace`, MCP status/search fallback, repair, SQLite counts, truncation safety checks, and BM25 fallback paths.
- Drops cached MemPalace/Chroma handles on MCP reconnect and reports reset failures.
- Adds post-write readback validation for MCP drawer writes.
- Scopes SQLite BM25 candidates to the configured collection and filtered wing/room before limiting.

Without the collection-name fix, a user with a non-default drawer collection could diagnose one collection, extract from another, or pass truncation checks against the wrong SQLite rows during recovery.

Once #1285 lands, this PR should rebase down to the single follow-up commit.

## How to test

Focused regression coverage is in `tests/test_mcp_server.py`, `tests/test_searcher.py`, `tests/test_repair.py`, `tests/test_cli.py`, and `tests/test_backends.py`.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -663,8 +663,10 @@ def cmd_repair(args):
         check_extraction_safety,
     )
 
+    config = MempalaceConfig()
+    collection_name = config.collection_name
     palace_path = os.path.abspath(
-        os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+        os.path.expanduser(args.palace) if args.palace else config.palace_path
     )
 
     if getattr(args, "mode", "legacy") == "max-seq-id":
@@ -749,7 +751,7 @@ def cmd_repair(args):
 
     # Try to read existing drawers
     try:
-        col = backend.get_collection(palace_path, "mempalace_drawers")
+        col = backend.get_collection(palace_path, collection_name)
         total = col.count()
         print(f"  Drawers found: {total}")
     except Exception as e:
@@ -784,6 +786,7 @@ def cmd_repair(args):
             palace_path,
             len(all_ids),
             confirm_truncation_ok=getattr(args, "confirm_truncation_ok", False),
+            collection_name=collection_name,
         )
     except TruncationDetected as e:
         print(e.message)
@@ -810,6 +813,7 @@ def cmd_repair(args):
             all_docs,
             all_metas,
             batch_size,
+            collection_name=collection_name,
             progress=print,
         )
     except RebuildCollectionError as e:
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -7,6 +7,7 @@ Priority: env vars > config file (~/.mempalace/config.json) > defaults
 import json
 import os
 import re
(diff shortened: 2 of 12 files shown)


mjc:
Addressed both config lookup comments in `34282d0`.

`DEFAULT_COLLECTION_NAME` already lived in `mempalace.config`, so I added `get_configured_collection_name()` there as the shared cached resolver. `palace.get_collection()`, `repair._get_collection_name()`, and `_bm25_only_via_sqlite()` now use that helper instead of constructing `MempalaceConfig()` directly in fallback paths.

MCP still threads `_config.collection_name` explicitly where it already has the config object, so the cached helper is only for callers that do not already have one.

igorls:
Rebased onto develop tip post-#1285/#1310 merges (force-pushed with \`--force-with-lease\`).

Two non-trivial conflict resolutions, plus a format follow-up:

1. **\`mempalace/repair.py\`** — develop merged #1285 (which introduced \`_drawers_collection_name()\`) and #1310 (which introduced the \`rebuild_from_sqlite\` block). This branch's commit \`34282d0\` was authored against an earlier base and added a parallel helper \`_get_collection_name()\` that reads the same config through a different API. After rebase, both helpers existed in the working tree and the \`status()\` function appeared twice. Resolution:
   - Kept \`_drawers_collection_name()\` (the established name on develop) and removed the duplicate \`_get_collection_name()\`.
   - Replaced all \`_get_collection_name()\` call sites in this branch's added code with \`_drawers_collection_name()\`.
   - Replaced the hardcoded \`"mempalace_closets"\` literal in \`status()\` with \`CLOSETS_COLLECTION_NAME\` for consistency.
   - Updated \`tests/test_repair.py\` (4 tests) to patch \`_drawers_collection_name\` instead of the removed name.

2. **\`mempalace/searcher.py\` + \`mempalace/palace.py\`** — small import / signature collisions; both new keyword-only parameters retained, both new imports retained.

3. **\`tests/test_hnsw_capacity.py\`** — the PR's production change adds \`_config.collection_name\` access in \`_tool_status_via_sqlite\`, but the two pre-existing \`_Cfg\` mocks only set \`palace_path\`. Added \`collection_name = "mempalace_drawers"\` to both so the new dependency is exercised.

4. Final \`fix(tests): apply ruff format\` commit covers blocks the rebase resolution left unformatted (\`test_mcp_server.py\`, \`test_searcher.py\`).

Local: 1603 tests pass, ruff lint + format clean against the 0.4.x CI pin. Watching CI.

Issue #1073, opened 2026-04-21
mempalace status CLI fails on palaces with >32k drawers: too many SQL variables (miner.py:852)


sha2fiddy:
On a 465k-drawer palace, `mempalace status` dies:

```
File ".../mempalace/miner.py", line 852, in status
    r = col.get(limit=total, include=["metadatas"]) if total else {"metadatas": []}
File ".../chromadb/api/rust.py", line 440, in _get
chromadb.errors.InternalError: Error executing plan: Internal error: error returned from database: (code: 1) too many SQL variables
```

`col.get(limit=total)` becomes `WHERE id IN (?, ?, ...)` server-side and blows SQLite's 32,766 bound-parameter limit. Any palace past that size kills the CLI.

**Env:** mempalace develop @ `1b00f93`, chromadb 0.6.3, Python 3.12, macOS.

**Repro:**

```
mempalace --palace <palace-with->32k-drawers> status
```

**Fix:** chunk the get, e.g.

```python
BATCH = 10_000
metas = []
for offset in range(0, total, BATCH):
    metas.extend(col.get(limit=BATCH, offset=offset, include=["metadatas"])["metadatas"])
```

Or aggregate counts in the backend — the MCP `mempalace_status` tool already does this without fetching metadata and doesn't hit the limit. The two paths could share an implementation.

MCP tool unaffected. Low severity, but blocks the CLI once a palace reaches active-use scale (~32k drawers = a few months of mining).

Can open a PR with the chunked-get fix if that's the preferred direction.

jfcampbl:
Reproduced on v3.3.2 with a palace of **114,232 drawers** across 11 wings. Same trace as the issue body, failure at `mempalace/miner.py:852` on the unbatched `col.get(limit=total, ...)`.

PR #1016 fixes this with the standard offset-batching pattern already used in `cli.py:309-319` and `repair.py:240-246`. Applied it locally — `mempalace status` now runs clean in ~5.6s, all wings visible, correct total.

Since PRs #1016, #851, #770, and #380 all propose the same pagination fix, merging any one of them (#1016 is the freshest and smallest at +9/-3) would close this. Left a reproducer on #1016 as well.

sha2fiddy:
[...]

PR #1316, opened 2026-05-02
fix: resolve 10K drawer pagination bug in status/taxonomy tools

brianmagicmoney:
# MemPalace 10K Drawer Bug — Root Cause Analysis

## Bug
`mempalace status` reports exactly 10,000 drawers. Real count: **24,775**.

## Root Cause
ChromaDB's `collection.get()` defaults to `limit=10000` when no limit is specified, and the code explicitly passes `limit=10000` in several places. Data IS being written correctly — it's a retrieval/display bug, not a write bug.

## Evidence
```
sqlite3 ~/.mempalace/palace/chroma.sqlite3 "SELECT COUNT(*) FROM embeddings;"
→ 24775

Wing breakdown (from raw SQLite):
  development         12122
  claude-project-memory 5033
  memory               3502
  claude-user-memory   2413
  mempalace            1076
  claude-sessions       609
  + 6 small wings
```

### 1. `mcp_server.py:71` — `tool_status()` (**REMAINING BUG**)
```python
all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
```
`col.count()` returns the true 24,775, but the wing/room dicts are built from only the first 10K rows. The MCP `mempalace_status` tool returns correct `total_drawers` but incomplete `wings`/`rooms`.

### 2. `miner.py:657` — CLI `status()` (**REMAINING BUG**)
```python
r = col.get(limit=10000, include=["metadatas"])
metas = r["metadatas"]
# ...
print(f"  MemPalace Status — {len(metas)} drawers")  # Shows 10000!
```
Uses `len(metas)` for display instead of `col.count()`. This is the source of the "exactly 10,000 drawers" report.

### 3. `mcp_server.py:420` — `tool_diary_read()` (low risk)
```python
limit=10000
```
Unlikely to hit 10K diary entries per agent, but should be fixed for consistency.

### 4. `mcp_server.py:122-130` — Already fixed (uncommitted)
`tool_list_wings`, `tool_list_rooms`, `tool_get_taxonomy` — already patched to use `_TAXONOMY_AGGREGATE_LIMIT = 1_000_000`. This fix is in the working tree but not committed.

## Fix
- Site 1: Change `tool_status()` to use `_TAXONOMY_AGGREGATE_LIMIT` like the other tools.
- Site 2: Change CLI `status()` to use `col.count()` for the header AND paginate the metadata fetch (same pattern as `cmd_repair` already uses).
- Site 3: Change `tool_diary_read()` to use a reasonable limit (100K) or paginate.

## Confidence
**HIGH** — verified by direct SQLite count (24,775) vs displayed count (10,000). The ChromaDB `limit=10000` default is documented behavior. The partial fix for list_wings/list_rooms/get_taxonomy in the uncommitted diff confirms this was already identified for those functions but `tool_status` and `miner.status` were missed.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -68,7 +68,7 @@ def tool_status():
     wings = {}
     rooms = {}
     try:
-        all_meta = col.get(include=["metadatas"], limit=10000)["metadatas"]
+        all_meta = col.get(include=["metadatas"], limit=_TAXONOMY_AGGREGATE_LIMIT)["metadatas"]
         for m in all_meta:
             w = m.get("wing", "unknown")
             r = m.get("room", "unknown")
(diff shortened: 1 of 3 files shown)


igorls:
Thanks for tackling this @brianmagicmoney! Closing as superseded — `develop` already ships a paginated fix for #1098 / #1073 / #1123:

- `mempalace/mcp_server.py` introduces `_fetch_all_metadata()` (true offset-based pagination in 1000-row batches) plus a 5s-TTL `_get_cached_metadata()` wrapper, and uses them across `tool_status`, `tool_list_wings`, `tool_list_rooms`, and `tool_get_taxonomy` with structured `partial=True` / `error` responses.
- `mempalace/miner.py` `status()` (around L1240) already paginates with `batch_size=5000` (refs #802/#850).

The approach in this PR (raising `_TAXONOMY_AGGREGATE_LIMIT` to 1,000,000) doesn't actually paginate — it just lifts the cap, so on large palaces it can still trip SQLite's 32766-variable ceiling. Merging it would regress the existing offset-batched implementation.

If `scripts/launch-mcp-persistent.sh` is still useful to you, please open a separate PR against `develop` and we'll review it on its own merits.

Really appreciate the diagnosis in #1098 — it's what made the develop fix possible.

PR #1327, opened 2026-05-03
feat: SessionStart context injection & Knowledge Graph summary

zhapostolski:
Automates palace 'wake-up' for all agents (Gemini, Claude, etc.) by injecting a summary of recent work and current Knowledge Graph relationships at session start.


zhapostolski:
Superseded by #1332 which includes all commits from this branch plus KG integration and protocol nudge.

Issue #1217, opened 2026-04-26
mine --mode convos should exclude subagent transcripts by default — 98.8% noise on Claude Code sessions


mirkobozzetto:
## Description

When mining Claude Code sessions with `--mode convos`, the miner processes all `.jsonl` files including those in `subagents/` directories. On real-world Claude Code usage, subagent transcripts represent **98.8% of files** but carry near-zero value for persistent memory.

## Data from real usage (2 months, 49 projects)

| | Files | Size | Signal value |
|---|---|---|---|
| Main sessions (`SESSION_ID.jsonl`) | 57 (1.1%) | 288 MB | High — real conversations, decisions, architecture choices |
| Subagents (`subagents/agent-*.jsonl`) | 4,928 (98.8%) | 1.2 GB | Low — ephemeral Explore/Grep/Plan agents, typically 2-5 turns |
| **Total** | **4,985** | **1.5 GB** | |

Mining all 4,985 files required **2h30+ of 100% CPU** on an Apple M5 Pro (embedding generation via `all-MiniLM-L6-v2`). Mining only the 57 main sessions would take a few minutes.

## What subagent transcripts typically contain

Most `subagents/agent-*.jsonl` files are 2-5 turn exchanges:
- Explore agent: "find files matching X" → glob result → summary
- Plan agent: reads 3 files → produces a plan
- Code-reviewer agent: reads diff → returns review table

These are ephemeral, disposable, and semantically redundant with the main session that spawned them. The parent session already contains the subagent's summary/result.

## Current behavior

`convo_miner.py` / `scan_convos()` picks up all `.jsonl` files recursively. Issue #111 fixed skipping `tool-results/` and `.meta.json`, but explicitly kept `subagents/*.jsonl` as "valuable" — this assessment doesn't hold at scale.

## Proposed solutions

**Option A — Skip `subagents/` by default, opt-in flag:**

```bash
# Default: skip subagents/
mempalace mine ~/.claude/projects/ --mode convos

# Opt-in: include subagents if desired
mempalace mine ~/.claude/projects/ --mode convos --include-subagents
```

**Option B — Heuristic: skip files with fewer than N exchanges:**

Most subagents are 2-5 turns. Main sessions are 20-200+ turns. A threshold of 10 turns would cleanly separate the two populations.

**Option C — Add `subagents/` to default `SKIP_DIRS` for convos mode:**

Similar to how `tool-results/` was added in #111.

## Workaround

```bash
[...]

Issue #587, opened 2026-04-10
feat: default-exclude runtime-state files + per-file drawer cap to prevent ingestion noise


mssteuer:
## Summary

When mining a directory of mixed "real content" and "machine-written runtime state," a single large JSON/cache file can dominate the palace with thousands of low-value, semantically near-identical drawers, crowding out search recall for the *actual* knowledge. There's no filename denylist, per-file drawer cap, or warning — the miner just happily files 2,479 drawers from a single cache blob.

This is related to the existing feature requests #56 (external exclude list, closed) and #233 (`.gitignore` support, closed — `.gitignore` IS honored now, nice). But neither of those catches the case where the noisy file lives *inside* a non-gitignored path: runtime state files that are supposed to be there but that a human reading the repo would never treat as "knowledge."

## Environment

- Ubuntu 24.04, Python 3.12
- `mempalace` from PyPI, mining `~/.hermes` (a Hermes agent home directory)
- Miner options: `--wing hermes --limit 30` (no `--no-gitignore`, no extra excludes)

## What happened

Mining 30 files from `~/.hermes` produced 2,619 drawers. Breakdown on a single file:

```
✓ [ 23/30] models_dev_cache.json                           +2479
```

That's **2,479 drawers from one cache file**, vs 140 drawers combined from 29 legitimate content files in the same run (SKILL.md files, SOUL.md, IDENTITY.md, USER.md, AGENTS.md, etc.). After completing a wider mine, the `cache` room had 1,965 drawers — 20% of the entire palace — all from files that are re-generated on every Hermes run and carry no durable knowledge.

`models_dev_cache.json` is exactly what it sounds like: a cache of the models.dev registry, structured like:

```json
{
  "anthropic/claude-opus-4.6": {
    "id": "anthropic/claude-opus-4.6",
    "context_length": 200000,
    "input_cost": 15.0,
    "output_cost": 75.0,
    ...
  },
  "openai/gpt-5": { ... },
  ...
}
```

Every model entry ends up in its own 800-char chunk, and they're all semantically near-identical, so they dilute the embedding space *and* crowd out relevant results for queries about models, pricing, etc.

## Why `.gitignore` doesn't help here

Hermes keeps `~/.hermes/cache/` *outside* of any git checkout — it's runtime state in the user's home directory. There's no `.gitignore` to opt into. The `~/.hermes` tree is a mixture of:

- real content (skills, profiles, config, docs)
[...]

PR #1330, opened 2026-05-03
fix(convo-miner): skip Claude Code subagent transcripts by default (#1217)

mvalentsev:
## Problem

`mempalace mine --mode convos` walks `~/.claude/projects/<slug>/` recursively
and picks up every `.jsonl` file, including those inside `<session-uuid>/subagents/`
where Claude Code records Explore/Plan/Grep subagent transcripts.

Reporter data on issue #1217 (49 projects, 2 months of usage):
- 57 main session files (1.1%, 288 MB) -- the high-signal content
- 4,928 subagent files (98.8%, 1.2 GB) -- short ephemeral exchanges already
  summarized in the parent session
- Mining took 2h30+ at 100% CPU on an Apple M5 Pro

Smoke-tested against a real Claude Code project directory: 21 main + 92
subagent jsonl files (5x ratio), default-skip drops 38% of mined files with
no signal loss.

## Fix

`scan_convos()` now skips any directory named `subagents` by default. New
`--include-subagents` flag re-enables them for users who want full history.

The filter is applied inside `convo_miner.scan_convos`'s directory
pruning (the same place `SKIP_DIRS` is consulted). Pruning governs a
directory walk only, so a transcript named directly -- `mine
<session>/subagents/agent-x.jsonl` -- is still mined without
`--include-subagents`. That split is deliberate: the flag decides what a
walk sweeps up, while naming a path stays an explicit request.

The shared `SKIP_DIRS` set in `palace.py` is left untouched, so project
mining (`miner.scan_project`) still descends into user-created
`subagents/` Python modules in regular code projects.

## Files

- `mempalace/convo_miner.py` -- `scan_convos(include_subagents=False)`
  parameter, threaded from `mine_convos()`
- `mempalace/cli.py` -- new `--include-subagents` flag on `mine`,
  forwarded through `cmd_mine` to `mine_convos`
- `tests/test_convo_miner_unit.py` -- 6 new tests (default skip, opt-in,
  any-depth match, suffix-named directories left alone, case-insensitive
  match, and an explicitly named file inside `subagents/`)
- `tests/test_cli.py` -- existing `test_cmd_mine_convos_mode` updated
  for the new kwarg, new test pinning the flag-threading path

## Verification

End-to-end smoke test against a real Claude Code project directory:

```
$ python3 -m mempalace --palace /tmp/a mine <claude-projects-dir> --mode convos --dry-run
  Files:   150
$ python3 -m mempalace --palace /tmp/b mine <claude-projects-dir> --mode convos --dry-run --include-subagents
  Files:   242
```

Difference = 92, matches `find <claude-projects-dir> -path '*/subagents/*.jsonl' | wc -l`.

## Out of scope

- `tool-results/` directories -- already covered by the existing
  `.meta.json` filename filter (no `.jsonl` files there).
- A turn-count heuristic (Option B in the issue body) -- more fragile
  and a larger surface; default-skip + opt-in flag is the cleaner UX.
- `.mempalaceignore` (#587) -- orthogonal: a future ignore-file
  mechanism would let users override the default if needed, but the
  default-safe path here is additive and doesn't conflict.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -596,6 +596,7 @@ def cmd_mine(args):
                 limit=args.limit,
                 dry_run=args.dry_run,
                 extract_mode=args.extract,
+                include_subagents=getattr(args, "include_subagents", False),
             )
         elif args.mode == "extract":
             from .format_miner import mine_formats
@@ -1918,6 +1919,16 @@ def main():
             f"summary counter. Default {_CLI_MAX_CHUNKS_PER_FILE_DEFAULT} "
             f"(or MEMPALACE_MAX_CHUNKS_PER_FILE). Set 0 to disable. Lower this on "
             f"Windows if you hit ONNX bad_alloc (#1455)."
+        ),
+    )
+    p_mine.add_argument(
+        "--include-subagents",
+        action="store_true",
+        help=(
+            "Also mine Claude Code subagent transcripts (subagents/ dirs). "
+            "Excluded by default — these are short ephemeral exchanges "
+            "(Explore/Plan/Grep agents) already summarized in the parent "
+            "session, and on a typical workspace outnumber main sessions ~80:1."
         ),
     )
 
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1924,11 +1924,12 @@ def main():
     p_mine.add_argument(
         "--include-subagents",
         action="store_true",
+        default=False,
         help=(
             "Also mine Claude Code subagent transcripts (subagents/ dirs). "
-            "Excluded by default — these are short ephemeral exchanges "
+            "Excluded by default: these are short ephemeral exchanges "
             "(Explore/Plan/Grep agents) already summarized in the parent "
-            "session, and on a typical workspace outnumber main sessions ~80:1."
+            "session, and on typical workspaces they dominate file counts."
         ),
     )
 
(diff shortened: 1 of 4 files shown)


Qodo-Free-For-OSS:
Hi, `cmd_mine()` directly reads `args.include_subagents` in convos mode, which raises `AttributeError` for programmatic callers that construct an `argparse.Namespace` without that field. This is inconsistent with `cmd_mine()`'s existing defensive pattern (e.g., `getattr(args, "redetect_origin", False)`) and can break non-CLI entrypoints/tests that call `cmd_mine()` directly.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Use getattr with default

> ### Issue description
> `mempalace.cli.cmd_mine()` directly accesses `args.include_subagents` when `args.mode == "convos"`. If `cmd_mine()` is called programmatically (tests, integrations, or internal code) with an `argparse.Namespace` that doesn’t include `include_subagents`, it will raise `AttributeError`.
> 
> ### Issue Context
> `cmd_mine()` already treats some fields as optional/feature-gated (e.g. `redetect_origin`) via `getattr`, so `include_subagents` should follow the same defensive pattern.
> 
> ### Fix Focus Areas
> - mempalace/cli.py[493-515]
>   - Change `include_subagents=args.include_subagents` to `include_subagents=getattr(args, "include_subagents", False)` (or ensure the attribute is always present before branching).
>   - Consider aligning other new/optional CLI fields to the same pattern to keep `cmd_mine()` robust for direct callers.

---

igorls (approved):

PR #1334, opened 2026-05-03
fix(normalize): strip Claude Code <local-command-*> tags and ANSI escapes

lucagentile:
## What does this PR do?

Fixes #1333.

Extends `mempalace/normalize.py:strip_noise()` to remove three Claude Code transcript artifacts that currently survive into stored drawers:

1. The remainder of the slash-command tag envelope — `<local-command-caveat>`, `<command-args>`, `<local-command-stdout>` — that Claude Code injects on every slash-command invocation. `strip_noise()` previously covered only `<command-name>` and `<command-message>`; the other three slipped through, leaving drawers with bare `<command-args></command-args>` remnants and whole `<local-command-stdout>` payloads.
2. ANSI escape sequences (CSI + OSC) captured verbatim from `Bash` tool_result content. These bloat embeddings (each escape is several BPE tokens of garbage) and garble semantic search when colored CLI output (e.g. `/context`, `/help`, anything piped through a modern tool) lands in a transcript.

### Code change (`mempalace/normalize.py`)

- Adds `command-args`, `local-command-caveat`, `local-command-stdout` to `_NOISE_TAGS`. They flow through the existing `_tag_pattern` builder, so they inherit the same line-anchoring + blank-line-bounded body guarantee — a stray unclosed tag in one message still cannot eat content from neighbouring messages.
- Adds two pre-compiled regexes:
  - `_ANSI_CSI_RE` matches the standard CSI shape `ESC [ params interm final` covering SGR colors, cursor moves, screen ops.
  - `_ANSI_OSC_RE` matches `ESC ] ... BEL` or `ESC ] ... ESC \` for terminal title (OSC 0/2) and hyperlinks (OSC 8).
  - Both are anchored on the literal ESC byte (`\x1b`), so user prose that mentions `[1m]` or `]8;;` *by name* never matches.
- Applies both ANSI sweeps inside `strip_noise()` after tag/line/chrome stripping and before the existing blank-line collapse. Tag-wrapped ANSI is already gone via `_NOISE_TAG_PATTERNS`, so the ANSI sweep only has to cover standalone ANSI in tool output — no double-strip path needed.

### Test change (`tests/test_normalize.py`)

Eleven new test cases, all under the existing `TestStripNoiseRemovesSystemChrome` and `TestStripNoisePreservesUserContent` classes to keep the verbatim-safety boundary discipline visible:

- `test_strips_each_known_noise_tag` extended to iterate the new tags too.
[...]

Issue #1314, opened 2026-05-02
kg_add and kg_invalidate silently drop temporal/source params (valid_to, source) — backfill loses precision


joaodotwork:
## Issue

Several knowledge-graph parameters are accepted by the API but silently
dropped at write time, making it hard to encode historical facts with
proper temporal bounds. Backfilling a knowledge graph from existing
project history (the natural use case described in the docs as
"tracking how decisions evolved") currently degrades to "approximate
chronology only."

## Concrete cases

All observed against `mempalace==3.3.4`, MCP server (also reproducible
via CLI), Python 3.11, macOS.

### Case 1 — `mempalace_kg_add` silently drops `valid_to`

Setup:

```python
mempalace_kg_add(
    subject="_test_temporal",
    predicate="had_value",
    object="probe",
    valid_from="2026-01-01T00:00:00Z",
    valid_to="2026-04-28T17:28:00Z",   # <-- silently ignored
)
```

Returns `{"success": true, ...}` with no warning.

Verifying via `mempalace_kg_query`:

```json
{
  "valid_from": "2026-01-01T00:00:00Z",
  "valid_to": null,                     <-- dropped
  "current": true                       <-- because valid_to is null
}
```

The fact is treated as still-current despite being explicitly bounded
in the past at write time.

### Case 2 — `mempalace_kg_invalidate` ignores `valid_to`

Setup:

```python
mempalace_kg_invalidate(
    subject="operating-verb",
    predicate="candidate",
    object="stewardship",
    valid_to="2026-04-28T17:28:00Z",   # <-- silently ignored
)
```

Returns `{"success": true, "ended": "today"}` — the response itself
admits today's date is being used regardless of the parameter.

Verifying:

```json
{
  "valid_from": null,
  "valid_to": "2026-05-02",            <-- today, not the requested 2026-04-28
  "current": false
}
```

Practical effect: backfilling historical state-transitions (e.g. "this
candidate was active from April 21 to April 28; then this one took
over") collapses to "this fact was current at some point and isn't
anymore as of today" for every transition. Loses the temporal precision
that makes the graph useful for revision-process queries like *"what
was the operating verb on April 25?"*

### Case 3 — `source` param accepted but stored as `null`

Setup:

```python
mempalace_kg_add(
    subject="operating-verb",
    predicate="candidate",
    object="husbandry",
    valid_from="2026-04-28T17:28:00Z",
[...]

Issue #1299, opened 2026-05-01
MCP server SIGSEGV on tool_diary_write in 3.3.4 — Connection closed; server unrecoverable in-session


bespined:
## Environment

- macOS Darwin 25.1.0 (Apple Silicon)
- Python 3.14 (`/Library/Frameworks/Python.framework/Versions/3.14/bin/mempalace-mcp`)
- mempalace 3.3.4 (verified via `pip show` and via MCP `initialize` handshake response)
- Claude Code with direct MCP config (`claude mcp add mempalace -- mempalace-mcp`)
- Palace: ~2030 drawers, repair-status reports OK (HNSW 1840 / sqlite 2030, within flush-lag tolerance)

## Symptom

`mempalace_diary_write` returns `MCP error -32000: Connection closed`. Subsequent calls to ANY tool (e.g. `mempalace_status`, `mempalace_reconnect`) also return `Connection closed`. Server is dead until Claude Code is fully quit and relaunched.

## Reproduction

1. Fresh Claude Code session with mempalace MCP registered (direct MCP, not plugin)
2. Confirmed server boots: `claude mcp list` shows `mempalace: mempalace-mcp - ✓ Connected`
3. Confirmed standalone server reports 3.3.4:
   ```
   echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1.0"}}}' | mempalace-mcp
   → {"result": {"serverInfo": {"name": "mempalace", "version": "3.3.4"}}}
   ```
4. From within Claude Code session, call `mempalace_diary_write` with minimal payload:
   ```json
   {"agent_name": "claude", "topic": "test", "entry": "test entry no unicode"}
   ```
5. Returns `MCP error -32000: Connection closed`
6. Subsequent `mempalace_status` ALSO returns `Connection closed` — server doesn't respawn

## Expected

3.3.4 CHANGELOG documents fix for #1089 / #1262 / #1289 — `tool_diary_write` chromadb 1.5.x SIGSEGV from differing collection metadata. Fix applied to MCP server `_get_collection(create=True)` path. Per CHANGELOG: "carried the same metadata payload at a parallel call site and got the same try/except split applied, closing the crash class on both reopen paths."

So `tool_diary_write` should NOT crash on 3.3.4. But it does.

## Hypothesis

- Different crash path not covered by 3.3.4 fix
[...]

PR #1283, opened 2026-04-30
fix(kg): validate ISO-8601 date formats in temporal parameters at MCP boundary

Giggitycountless:
## Summary

Adds `validate_iso_date()` to `config.py` and applies it at the MCP boundary in three knowledge-graph tools:

- `tool_kg_query` — validates `as_of` parameter
- `tool_kg_add` — validates `valid_from` parameter
- `tool_kg_invalidate` — validates `ended` parameter

## Problem

Non-ISO date strings like `"March 2026"` or `"Jan 2025"` were forwarded to SQLite without validation, silently producing empty result sets instead of matching facts. Users could not distinguish "no fact at this time" from "your date format was unrecognized."

## Changes

| File | Change |
|------|--------|
| `mempalace/config.py` | Added `_ISO_DATE_RE` regex and `validate_iso_date()` function |
| `mempalace/mcp_server.py` | Import `validate_iso_date`, apply in `tool_kg_query`, `tool_kg_add`, `tool_kg_invalidate` |
| `tests/test_config.py` | 15 new tests covering valid dates, edge cases, and rejection of malformed strings |

## Validation Details

The regex accepts:
- `YYYY` (year only)
- `YYYY-MM` (year + month)
- `YYYY-MM-DD` (full date)

Rejects natural-language dates, partial formats, invalid months (>12), and invalid days (>31).

## Tests

```
47 passed in 0.16s
ruff check — all passed
ruff format — all formatted
```

Fixes #1164

PR #1335, opened 2026-05-03
feat: community patches — mine deadlock fix, block_once precompact, KG auto-write, repair BLOB seq_id

zhapostolski:
## What this PR does

Consolidates 32 commits combining community bug fixes and session/hook infrastructure.

### Community fixes included
- **#1162** Serialize all ChromaDB writes through palace-level lock (multi-CLI concurrency)
- **#1214** KG inverted interval rejection (valid_to < valid_from)
- **#1215** entity_registry atomic write + fsync for ext4 durability
- **#1273** HNSW compress: drop+recreate to avoid index bloat
- **#1283** ISO-8601 date validation at KG MCP boundary
- **#1299** `_get_collection` passes `embedding_function=` (prevents SIGSEGV on reopen)
- **#1312** Repair paths use configured `collection_name`
- **#1314** `kg_add` forwards `valid_to`, `source_file`, `source_drawer_id`
- **#1243** diary_write/read case-insensitive agent name
- **#1330** convo-miner skips Claude Code `subagents/` directories
- **#1334** normalize: strip Claude Code local-command tags + ANSI escapes

### Session & hook infrastructure
- **SessionStart hook** — injects palace wake-up context + protocol nudge into every session
- **Stop hook: local memory mirror** — auto-replicates `.md` files from `~/.claude/projects`, `~/.codex/memories`, `~/.gemini/memory`, `~/.qwen/memory` into palace drawers (cross-agent memory)
- **KG in wake-up stack** — recent KG facts surfaced at session start
- **Multi-agent hook scripts** — wrappers for codex, opencode, gemini, qwen

### Latest additions (this update)
- **fix(repair):** `_detect_poisoned_max_seq_ids` handles BLOB-typed `seq_id` from ChromaDB 1.5.x — fixes `ValueError: invalid literal for int()` blocking `repair --mode max-seq-id` on 92k+ drawer palaces
- **fix(hooks): mine deadlock prevention** — `_ingest_transcript` checks `_mine_already_running()` before spawning and registers its PID; `hook_stop` skips diary save when mine is running, retries on next fire — eliminates cascading "another mine already running" failures
- **fix(hooks): PreCompact block_once** — default changed from `proceed` (no-op for claude-code/gemini) to `block_once` for all harnesses
- **feat(hooks): KG auto-write** — session topics written to KG as `session-YYYY-MM-DD → covered_topic → theme` after each silent diary checkpoint

## Test plan
- [ ] `repair --mode max-seq-id` completes without ValueError on palace with BLOB seq_ids
- [ ] Stop hook no longer emits "another mine already running" when transcript mine is active
- [ ] PreCompact returns `{"decision":"block"}` on first fire per session, `{}` on second
- [ ] After 10+ messages, KG contains `session-YYYY-MM-DD` with `covered_topic` predicate
- [ ] `python -m pytest tests/ -v --ignore=tests/benchmarks`


mjc:
I think this PR improves the write-serialization story quite a bit, especially via the `#1162` lock path on `ChromaCollection.add/upsert/update/delete`, and `_close_chroma_handles(...)` also looks like a real improvement for rollback/restore cleanup.

I’m not fully convinced it covers the entire repair/rebuild boundary yet, though. The part I’m still looking at is whether repair is treated as an exclusive palace-level operation from the point where it opens the live collection through backup, replacement, rollback, and restore. My concern there is a little different from the ordinary concurrent-writer case: even if normal collection writes are serialized, the destructive swap/restore portion of repair happens outside the `ChromaCollection` write methods.

So I think this gets us closer, but there may still be a gap around holding an outer repair-specific write lock across the full destructive section of the workflow. I’m working on that side separately, along with the handle-release part before filesystem moves/removals, just to make sure the repair boundary is fully covered.

PR #1339, opened 2026-05-04
fix(storage): quarantine bloated HNSW link payloads

### LINKED ISSUE #1218: mempalace migrate/status crash with SIGSEGV on chromadb version mismatch — palace unrecoverable via official tooling
## Problem

When `chromadb` is upgraded and the palace index is on an older format, `mempalace migrate` (and `mempalace status`) exit with code 139 (SIGSEGV) and produce **zero output**.

**Root cause:** `migrate.py` (~line 162) calls `col.count()` to probe whether the palace is readable. This call segfaults inside chromadb's native HNSW layer. SIGSEGV bypasses Python's `try/except`, so the migration fallback code (lines 166–171: extract from SQLite, rebuild) is **never reached** — the whole process dies silently.

The same SIGSEGV also kills `mempalace status` and the MCP server on startup, leaving the palace **completely inaccessible through all official tooling**.

## Environment

- macOS 15, Apple Silicon M-series
- mempalace installed via pip (pyenv 3.11.6)
- Palace: 3870 drawers + 139 closets
- `mempalace migrate --dry-run` → exit 139, zero output
- `mempalace status` → exit 139, zero output
- `mempalace repair` → exit 139, zero output

## Workaround

Manual recovery by bypassing the `col.count()` probe entirely:

1. Extract drawers + closets from `chroma.sqlite3` via raw SQL (reuses `extract_drawers_from_sqlite` from `migrate.py`)
2. `shutil.copytree` backup
3. `ChromaBackend().get_or_create_collection` in a temp dir, `col.add()` in batches (re-embeds locally)
4. Copy `knowledge_graph.sqlite3` sidecars
5. Swap temp dir into place

Full recovery script:

```python
"""Manual MemPalace recovery — bypasses the segfaulting count() probe."""
from __future__ import annotations
import os, shutil, sys, tempfile
from datetime import datetime

PALACE = os.path.expanduser("~/.mempalace/palace")
DB = os.path.join(PALACE, "chroma.sqlite3")

from mempalace.migrate import extract_drawers_from_sqlite

def extract_collection_from_sqlite(db_path, collection_name):
    import sqlite3
    conn = sqlite3.connect(db_path)
    try:
        rows = conn.execute("""
            SELECT em.id,
                MAX(CASE WHEN em.key = 'chroma:document' THEN em.string_value END) as document
            FROM embedding_metadata em
            JOIN embeddings e ON e.id = em.id
            JOIN segments s ON s.id = e.segment_id
            JOIN collections c ON c.id = s.collection
            WHERE c.name = ? GROUP BY em.id""", (collection_name,)).fetchall()
        results = []
        for row_id, document in rows:
            if not document: continue
            metas = conn.execute("""SELECT em.key, em.string_value, em.int_value,
                em.float_value, em.bool_value FROM embedding_metadata em
                WHERE em.id = ? AND em.key NOT LIKE 'chroma:%'""", (row_id,)).fetchall()
            metadata = {}
            for k, sv, iv, fv, bv in metas:
                if sv is not None: metadata[k] = sv
                elif iv is not None: metadata[k] = iv
                elif fv is not None: metadata[k] = fv
                elif bv is not None: metadata[k] = bool(bv)
            ext_id = conn.execute(
                "SELECT embedding_id FROM embeddings WHERE id = ?", (row_id,)).fetchone()
            if ext_id:
                results.append({"id": ext_id[0], "document": document, "metadata": metadata})
        return results
    finally:
        conn.close()

drawers = extract_drawers_from_sqlite(DB)
closets = extract_collection_from_sqlite(DB, "mempalace_closets")
print(f"Extracted: {len(drawers)} drawers, {len(closets)} closets")

ts = datetime.now().strftime("%Y%m%d_%H%M%S")
backup = f"{PALACE}.pre-migrate.{ts}"
shutil.copytree(PALACE, backup)
print(f"Backup: {backup}")

temp = tempfile.mkdtemp(prefix="mempalace_recover_")
fresh = ChromaBackend()

def reimport(name, items, batch_size=500):
    if not items: return 0
    col = fresh.get_or_create_collection(temp, name)
    done = 0
    for i in range(0, len(items), batch_size):
        batch = items[i:i+batch_size]
        col.add(ids=[d["id"] for d in batch],
                documents=[d["document"] for d in batch],
                metadatas=[d["metadata"] for d in batch])
        done += len(batch)
        print(f"  {name}: {done}/{len(items)}")
    return col.count()

dr_n = reimport("mempalace_drawers", drawers)
cl_n = reimport("mempalace_closets", closets)
print(f"Re-imported: drawers={dr_n}, closets={cl_n}")

for sidecar in ("knowledge_graph.sqlite3", "knowledge_graph.sqlite3-shm", "knowledge_graph.sqlite3-wal"):
    src = os.path.join(PALACE, sidecar)
    if os.path.isfile(src):
        shutil.copy2(src, os.path.join(temp, sidecar))

shutil.rmtree(PALACE)
shutil.move(temp, PALACE)
print("Done.")
```

## Suggested fix

Wrap the `col.count()` probe in a **subprocess** so SIGSEGV is contained at the process boundary:

```python
import subprocess, sys

def probe_collection_count(palace_path: str, collection_name: str):
    """Returns int count, or None if probe segfaults/fails."""
    code = (
        f"from mempalace.backends.chroma import ChromaBackend; "
        f"cb=ChromaBackend(); col=cb.get_collection('{palace_path}', '{collection_name}'); "
        f"print(col.count())"
    )
    try:
        result = subprocess.run(
            [sys.executable, "-c", code],
            capture_output=True, text=True, timeout=15
        )
        if result.returncode == 0:
            return int(result.stdout.strip())
    except Exception:
        pass
    return None  # crashed → proceed with SQL extraction
```

Same pattern applies to `mempalace status` and MCP server startup.

## Impact

Without this fix, every chromadb version bump that changes the HNSW index format leaves users with a palace that cannot be migrated, repaired, or read — all official tooling crashes before producing any output. The SQLite data is always intact but unreachable without a manual workaround like the script above.

fatkobra:
## What does this PR do?

Fixes #1218.

This extends the pre-open HNSW health gate so MemPalace quarantines structurally corrupt HNSW segments where `link_lists.bin` is wildly larger than `data_level0.bin`, even when `index_metadata.pickle` looks valid.

## What changed

- Added a structural HNSW payload sanity check:
  - compute `link_lists.bin / data_level0.bin`
  - treat ratios above `10x` as corrupt
- Updated `_segment_appears_healthy()` so a valid pickle alone is no longer enough to mark a segment healthy.
- Updated `quarantine_stale_hnsw()` so link-payload corruption is not blocked by the mtime gate.
  - stale-by-mtime + bad pickle is still handled as before
  - extreme link/data bloat is quarantined immediately
- Added regression tests for:
  - ratio calculation
  - rejecting exploded `link_lists.bin` with a valid pickle
  - keeping reasonable payload ratios
  - quarantining link bloat even when sqlite/HNSW mtimes are identical
  - leaving healthy segments untouched

## Why

The latest #1218 investigation found a false negative in the existing HNSW sniff-test:

- `index_metadata.pickle` was structurally valid
- `link_lists.bin` had exploded relative to `data_level0.bin`
- Chroma/HNSW then segfaulted on official tooling paths such as status/repair/migrate

The old gate answered only “does the pickle look complete?” This PR also asks “does the HNSW payload next to it look physically sane?”

That matters because `ChromaBackend._prepare_palace_for_open()` runs `quarantine_stale_hnsw()` before constructing a Chroma client. By catching the bloated payload there, MemPalace can rename the unsafe segment before Chroma's native HNSW reader has a chance to crash the process.

## Threshold

The threshold is intentionally conservative:

```text
link_lists.bin / data_level0.bin > 10x

## How to test
ruff format mempalace/backends/chroma.py tests/test_hnsw_payload_health.py
ruff check mempalace/backends/chroma.py tests/test_hnsw_payload_health.py
python -m pytest tests/test_hnsw_payload_health.py -q
python -m pytest tests/test_hnsw_payload_health.py tests/test_repair.py -q
python -m pytest tests/ -q

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -29,6 +29,51 @@ _REQUIRED_OPERATORS = frozenset({"$eq", "$ne", "$in", "$nin", "$and", "$or", "$c
 _OPTIONAL_OPERATORS = frozenset({"$gt", "$gte", "$lt", "$lte"})
 _SUPPORTED_OPERATORS = _REQUIRED_OPERATORS | _OPTIONAL_OPERATORS
 
+# A healthy HNSW payload should keep link_lists.bin proportional to
+# data_level0.bin. When link_lists.bin grows orders of magnitude larger than
+# data_level0.bin, Chroma/HNSW can segfault while opening the segment even if
+# index_metadata.pickle is structurally valid.
+#
+# The report in #1218 showed ratios above 300x, while healthy snapshots were far below 1x.
+# Treat only >10x as corruption so normal flush lag or small segments do not get
+# quarantined.
+_HNSW_LINK_TO_DATA_MAX_RATIO = 10.0
+
+
+def _hnsw_link_to_data_ratio(seg_dir: str) -> Optional[float]:
+    """Return link_lists.bin / data_level0.bin size ratio for a segment.
+
+    ``None`` means the ratio is not meaningful, usually because one file is
+    missing or data_level0.bin is empty. ``float("inf")`` means the files were
+    present but could not be statted safely, which should be treated as
+    suspicious by callers.
+    """
+
+    link_path = os.path.join(seg_dir, "link_lists.bin")
+    data_path = os.path.join(seg_dir, "data_level0.bin")
+
+    if not (os.path.isfile(link_path) and os.path.isfile(data_path)):
+        return None
+
+    try:
+        data_size = os.path.getsize(data_path)
+        link_size = os.path.getsize(link_path)
+    except OSError:
+        return float("inf")
+
+    if data_size <= 0:
+        return None
+
+    return link_size / data_size
+
+
+def _hnsw_payload_appears_sane(seg_dir: str) -> bool:
+    """Return False when HNSW payload files are structurally implausible."""
+
+    ratio = _hnsw_link_to_data_ratio(seg_dir)
+    return ratio is None or ratio <= _HNSW_LINK_TO_DATA_MAX_RATIO
+
+
 # HNSW tuning to prevent link_lists.bin bloat on large mines (#344).
 #
 # With default params (batch_size=100, sync_threshold=1000, initial capacity
@@ -106,6 +151,9 @@ def _segment_appears_healthy(seg_dir: str) -> bool:
     files and quarantine_stale_hnsw would conservatively rename them
     out of the way (lazy rebuild on next open recovers).
     """
+    if not _hnsw_payload_appears_sane(seg_dir):
+        return False
+
     meta_path = os.path.join(seg_dir, "index_metadata.pickle")
     if not os.path.isfile(meta_path):
         # No metadata file yet — segment hasn't flushed (fresh / empty).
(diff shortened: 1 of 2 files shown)


moonn1000:
Threshold 10× checks out empirically against the data from #1218: the crashed palace ran at ~314 (24 GB / 76 MB), healthy snapshots stayed ~0.005 (88 KB / 17.4 MB at 49K drawers). Four orders of magnitude between corruption and any plausible healthy ratio I've seen.

The bypass of the mtime gate when `payload_corrupt` is the right call — my crashed segment had `data_level0.bin` mtime aligned with `chroma.sqlite3` (write was still active when the bloat hit), so an mtime-only gate would have skipped it. Good to see `test_quarantine_catches_link_bloat_without_mtime_drift` covering that.

If a reproducer fixture would help, I still have the corrupted palace (~6 GB compressed, `link_lists.bin` is mostly zero-padded). Happy to upload to a release asset, push via git-lfs, or any S3 link the maintainers prefer — let me know what works.

fatkobra:
Thanks @moonn1000  for validating the threshold and for the detailed data — super helpful.

Good to know the mtime bypass matches the real failure case you saw.

And yeah, a reproducer could definitely be useful for future regression testing — I’d be in favor of including it if maintainers are open to it.

moonn1000:
@fatkobra @jphein — quick re-ping on the reproducer offer.

Tarball is still on hand for the corrupted palace from #1218 (the `link_lists.bin` = 24 GB / `data_level0.bin` = 76 MB / ratio ~314 case). Bundle would contain:

- `chroma.sqlite3` — intact, 302 MB
- `<segment-uuid>/` — full corrupted segment dir (including the 24 GB zero-padded `link_lists.bin`)
- A small README pinning chromadb 1.5.7 + mempalace 3.3.3 + drawer count (~85 K)

Compressed size ~6 GB (the link_lists payload is mostly zero-padded so it compresses well).

Just need a destination preference from a maintainer:

- GitHub release asset on the repo (simplest, but >2 GB single-file limit on standard releases — would need to split or use LFS)
- git-lfs in a side branch
- Any S3/CDN link the project would prefer me to upload to

Happy to also strip it down further (e.g., truncate `link_lists.bin` to the first few MB if a smaller pathological-ratio reproducer is enough for the regression test — the corruption shape, not the absolute size, is what `_segment_appears_healthy()` keys on after this PR). Let me know which form is most useful.

Also worth knowing: I just hit a _separate_ `repair --yes` failure mode on a different palace (SQLITE_CORRUPT_INDEX at `delete_collection` before rebuild — written up at #1218). That one's a much smaller reproducer (~286 MB) and a different failure surface from this PR's scope, but if a fixture for SQLite-side corruption-vs-HNSW-side corruption discrimination would be useful in a follow-up, I have it too.

PR #1342, opened 2026-05-04
fix(storage): quarantine partial HNSW flush without metadata

### LINKED ISSUE #1274: SIGSEGV in chromadb bindings.count() on develop tip; `_segment_appears_healthy` false-negative for partially-flushed segments
Two related findings hit while running develop tip (commit `fdfaf01`) against an existing palace. Filing together since the second was discovered while diagnosing the first.

### Repro
- mempalace `develop` tip, commit `fdfaf01`
- macOS 26.4.1, arm64
- Reproduces on **both** Python interpreters tested:
  - Python 3.9.6 (Apple CommandLineTools) + chromadb 1.5.7
  - Python 3.13.9 (Homebrew, vanilla CPython) + chromadb 1.5.8
- Existing palace, ~88K drawers, 597 MB `chroma.sqlite3`

```python
from mempalace.miner import get_collection
col = get_collection('~/.mempalace/palace', create=False)
col.count()
```
→ exit 139 (SIGSEGV)

### Stack
```
File "mempalace/backends/chroma.py", line 812 in count
File "chromadb/api/models/Collection.py", line 53 in count
File "chromadb/api/rust.py", line 397 in _count
   return self.bindings.count(str(collection_id), tenant, database)
```

The crash is inside chromadb's Rust binding (`bindings.count`). `_pin_hnsw_threads` retrofit applies (verified in code path), but doesn't prevent the crash because this isn't the parallel-insert path it guards.

### What works
- `mempalace search` succeeds
- `mempalace repair-status` succeeds — reports `drawers: sqlite=87898 hnsw=87897 status=OK (within flush-lag tolerance)`
- Collection open / `get_or_create_collection` succeeds; only `count()` (and presumably `get(...)`) crashes

### Suspect data conditions
- **5-day mtime drift** between `chroma.sqlite3` (today) and HNSW segment `data_level0.bin` files (Apr 24)
- **Poisoned `max_seq_id` BLOB format**: rows in `max_seq_id` table are 5-byte ASCII numerals (e.g. `0x3934343237` = "94427"), without the `\x11\x11` chromadb-1.5 prefix mentioned in the chroma.py:537 comment. Looks like the legacy 0.6.x format that survived the 0.6→1.5 migration. (`mempalace repair --mode max-seq-id` is suggested in that comment for poisoned palaces, but the SIGSEGV happens before that path is reachable through normal use.)

### What did NOT fix it
- Removing the corrupt closets segment (Finding 2) — `count()` is on drawers, separate segment
- `_pin_hnsw_threads` runtime retrofit (already applied by code path; crash still occurs)

---

## Finding 2 — `_segment_appears_healthy` false-negative for partially-flushed segments

Discovered while inspecting the palace for Finding 1.

`backends/chroma.py:108-112`:

```python
meta_path = os.path.join(seg_dir, "index_metadata.pickle")
if not os.path.isfile(meta_path):
    # No metadata file yet — segment hasn't flushed (fresh / empty).
    # Renaming would orphan nothing; consider healthy.
    return True
```

Observed segment state in this palace:
```
6b9dde8c-1b52-4afa-a874-c568bb015e86/   (closets collection vector segment)
  data_level0.bin       167600 bytes   ← non-empty: has data
  header.bin            100 bytes
  length.bin            400 bytes
  link_lists.bin        0 bytes        ← empty
  (no index_metadata.pickle)           ← MISSING
```

A segment with non-trivial `data_level0.bin` but missing `index_metadata.pickle` is partially-flushed corruption, not a fresh/empty segment. `_segment_appears_healthy` returns `True`, `quarantine_stale_hnsw` skips it, and chromadb later attempts to deserialize a non-existent pickle.

### Suggested fix
In `_segment_appears_healthy`, when `index_metadata.pickle` is missing, gate the "fresh/empty" assumption on `data_level0.bin` size:

```python
meta_path = os.path.join(seg_dir, "index_metadata.pickle")
data_path = os.path.join(seg_dir, "data_level0.bin")
if not os.path.isfile(meta_path):
    try:
        # If data file is non-trivial, segment is partially-flushed corruption
        # (data without metadata), not a fresh segment.
        if os.path.isfile(data_path) and os.path.getsize(data_path) > 1024:
            return False
    except OSError:
        return False
    return True
```

### Repro for Finding 2
Hard to reproduce synthetically — depends on a partial-flush mid-write. In this palace it occurred against the closets collection. Workaround is to manually rename the segment dir to `<uuid>.drift-manual` so chromadb rebuilds lazily on next open.

---

Happy to provide more diagnostics, run additional repro steps, or test a patch.

fatkobra:
## What does this PR do?

Fixes #1274.

This tightens `_segment_appears_healthy()` so a missing `index_metadata.pickle` is treated as healthy only when the HNSW segment still looks fresh or effectively empty.

If `data_level0.bin` already has non-trivial payload but `index_metadata.pickle` is missing, the segment is now treated as partially flushed corruption and `quarantine_stale_hnsw()` renames it before Chroma opens the native HNSW reader.

## What changed

- Added `_HNSW_MISSING_METADATA_DATA_FLOOR`.
- Updated `_segment_appears_healthy()`:
  - missing `index_metadata.pickle` + tiny/empty `data_level0.bin` remains healthy
  - missing `index_metadata.pickle` + non-trivial `data_level0.bin` is unhealthy
  - `OSError` while checking the payload is treated as unsafe
- Added regression tests for:
  - non-trivial HNSW data without metadata
  - tiny data without metadata
  - quarantine behavior for the #1274 partial-flush shape

## Why

#1274 reports a real palace with this segment state:

```text
data_level0.bin       167600 bytes
header.bin            100 bytes
length.bin            400 bytes
link_lists.bin        0 bytes
index_metadata.pickle MISSING

## How to test
ruff format mempalace/backends/chroma.py tests/test_backends.py
ruff check mempalace/backends/chroma.py tests/test_backends.py
python -m pytest tests/test_backends.py -q
python -m pytest tests/test_backends.py tests/test_repair.py -q
python -m pytest tests/ -q

## Checklist
- [x] No hardcoded paths


igorls:
Maintainer-edit: merged develop in to resolve conflicts opened by #1339 (bloated link payloads, merged earlier today).

\`_segment_appears_healthy\` now stacks three complementary checks:

1. \`_hnsw_payload_appears_sane\` (#1339) — link_lists.bin / data_level0.bin ratio guard
2. missing-metadata + non-empty data floor (this PR) — partial-flush detection
3. pickle format sniff (original) — protocol-byte corruption detection

Each catches a different shape of broken-on-disk segment. Local: 1618 tests pass on the merge, ruff lint+format clean against the 0.4.x CI pin.

PR #1343, opened 2026-05-04
feat(miner): store content_hash in drawer metadata

rusel95:
## 🧠 The problem

Every time you edit a file, MemPalace has no idea. It mines once, stores the content, and never checks again. If you update your README, refactor a module, or rewrite a doc — the palace still serves the old version. There is no way to know which memories are stale.

This PR lays the foundation to fix that.

## ✅ What this does

Stores a **content hash (MD5)** in every drawer at mine time — for both project files and conversation transcripts.

```python
# Every drawer now has this in its metadata:
{"content_hash": "a3f2c1d4e5b6..."}
```

- `file_content_hash(filepath)` — new helper in `miner.py`, single source of truth for all hash computation
- `add_drawer()` gains optional `content_hash` param; falls back to hashing the drawer content itself so MCP-created drawers (no source file on disk) also get a hash
- `process_file()` computes the hash and passes it through the batched upsert path
- `_file_chunks_locked()` in `convo_miner` gains `content_hash` param + `filepath_filter` for single-file re-mine
- `mine_convos()` hashes raw file bytes **before** `normalize()` transforms them — so sync can compare without re-running the normalizer

## 🔗 Why it matters

This is the foundation for:
- **`mine --force`** (PR #1344) — delete stale drawers and re-mine a directory
- **`mempalace_sync_status`** (PR #1347) — AI agents can check freshness before trusting search results
- Future incremental sync command

Without this, those features cannot work. With this, every new mine produces hashes that can be compared later.

## ↩️ Backward compatibility

Drawers mined before this change have no `content_hash`. Any sync/freshness tooling treats missing hash as `unknown` — not `fresh` or `stale`. No data loss, no breaking changes.

## 🧪 Tests

All 23 existing miner and convo_miner tests pass. No new failures introduced.

```
python -m pytest tests/test_miner.py tests/test_convo_miner.py -q
23 passed
```

Relates to #224.

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -307,7 +307,7 @@ def scan_convos(convo_dir: str) -> list:
 # =============================================================================
 
 
-def _file_chunks_locked(collection, source_file, chunks, wing, room, agent, extract_mode):
+def _file_chunks_locked(collection, source_file, chunks, wing, room, agent, extract_mode, content_hash=""):
     """Lock the source file, purge stale drawers, and upsert fresh chunks.
 
     Combines the per-file serialization that prevents concurrent agents from
@@ -361,6 +361,7 @@ def _file_chunks_locked(collection, source_file, chunks, wing, room, agent, extr
                         "ingest_mode": "convos",
                         "extract_mode": extract_mode,
                         "normalize_version": NORMALIZE_VERSION,
+                        "content_hash": content_hash,
                     }
                 )
             try:
@@ -384,6 +385,7 @@ def mine_convos(
     limit: int = 0,
     dry_run: bool = False,
     extract_mode: str = "exchange",
+    filepath_filter: str = None,
 ):
     """Mine a directory of conversation files into the palace.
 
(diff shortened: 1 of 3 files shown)


rusel95:

Ran real e2e tests against actual ChromaDB with real file mining — no mocks for core paths.

**7 tests added** (`tests/test_content_hash.py`):

| Test | What it verifies |
|------|-----------------|
| `test_deterministic` | Same file → same hash, every time |
| `test_changes_with_content` | Different content → different hash |
| `test_strips_whitespace` | Whitespace-only differences produce same hash |
| `test_project_mine_stores_content_hash` | `mine()` → every drawer has `content_hash` in metadata |
| `test_stored_hash_matches_computed` | Stored hash == `file_content_hash()` on disk |
| `test_no_file_hash_falls_back_to_content_hash` | MCP-created drawers (no source file) get content-derived hash |
| `test_explicit_hash_takes_precedence` | `_build_drawer_metadata()` uses provided hash over fallback |

```
python -m pytest tests/test_content_hash.py -v
7 passed in 28s
```

Full suite (991 existing + 7 new = 998 tests): 0 regressions.

PR #1345, opened 2026-05-04
feat(hooks): configurable save_interval via config + env var

### LINKED ISSUE #494: Feature Request: Add a "Silent Mode" / Background Processing for the Stop Hook (Auto-save)

rusel95:
## 😤 The problem

The stop hook fires every **15 messages**. Always. No way to change it without editing source code or uninstalling the plugin.

For long coding sessions this is fine. For a quick 3-message task it is noise. For a CI/automation context it is a blocker.

## ✅ What this does

Makes the interval configurable — via config file or env var:

```json
// ~/.mempalace/config.json
{"hooks": {"save_interval": 30}}
```

```bash
# Or via env var (takes precedence)
MEMPALACE_HOOKS_SAVE_INTERVAL=30   # save every 30 messages
MEMPALACE_HOOKS_SAVE_INTERVAL=0    # disable stop hook entirely
```

**Priority:** env var > config file > default (15)

Setting `save_interval: 0` disables the stop hook without uninstalling it. Negative values are clamped to 0.

## 🔑 Key design decision

`save_interval` controls **only** the stop hook. The precompact hook is independent (see PR #1346). Disabling periodic saves does not disable the last-chance save before context compression.

## 🧪 Tests

All 104 existing hook and config tests pass.

```
python -m pytest tests/test_hooks_cli.py tests/test_config.py -q
104 passed
```

Closes #494.

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -295,6 +295,25 @@ class MempalaceConfig:
         return self._file_config.get("hooks", {}).get("silent_save", True)
 
     @property
+    def hooks_save_interval(self) -> int:
+        """Number of exchanges between auto-save prompts. 0 = disabled.
+
+        Config: {"hooks": {"save_interval": 30}}
+        Env var (takes precedence): MEMPALACE_HOOKS_SAVE_INTERVAL=30
+        """
+        env_val = os.environ.get("MEMPALACE_HOOKS_SAVE_INTERVAL", "").strip()
+        if env_val:
+            try:
+                return max(0, int(env_val))
+            except ValueError:
+                pass
+        val = self._file_config.get("hooks", {}).get("save_interval", 15)
+        try:
+            return max(0, int(val))
+        except (TypeError, ValueError):
+            return 15
+
+    @property
     def hook_desktop_toast(self):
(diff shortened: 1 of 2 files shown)


rusel95:

Verified against the real `MempalaceConfig` class and `hook_stop()` function — no mocks for the config path.

**Tests verified:**
- Default interval = 15 ✅
- `MEMPALACE_HOOKS_SAVE_INTERVAL=50` → interval = 50 ✅
- `MEMPALACE_HOOKS_SAVE_INTERVAL=0` → hook disabled ✅
- `MEMPALACE_HOOKS_SAVE_INTERVAL=-5` → clamped to 0 ✅
- Config file `{"hooks": {"save_interval": 30}}` → interval = 30 ✅
- Independence: `save_interval=0` does NOT affect `precompact` ✅

```
python -m pytest tests/test_hooks_cli.py tests/test_config.py -v
104 passed in 0.65s
```

Full suite: 0 regressions.

igorls:

PR #1346, opened 2026-05-04
feat(hooks): independent precompact toggle via config + env var

### LINKED ISSUE #906: preCompact hook prevents compacting when context limit is reached in claude code
**What happened?**
context limit was reached and was not able to compact

**What did you expect?**
that the hook would run and then the context would be compacted
**How to reproduce:**
1. Fully use mempalace in a claude code session
2. Do complex tasks until context is full
3. Try to compact

Got this error:
```
  ⎿  Context limit reached · /compact or /clear to continue · /model opus[1m]

✻ Crunched for 2m 53s

❯ /compact                                                                                                          
  ⎿  Error: Compaction blocked by PreCompact hook: [bash ${CLAUDE_PLUGIN_ROOT}/hooks/mempal-precompact-hook.sh]:
     COMPACTION IMMINENT (MemPalace). Save ALL session content before context is lost:
     1. mempalace_diary_write — thorough AAAK-compressed session summary
     2. mempalace_add_drawer — ALL verbatim quotes, decisions, code, context
     3. mempalace_kg_add — entity relationships (optional)
     Be thorough — after compaction, detailed context will be lost. Do NOT write to Claude Code's native auto-memory
      (.md files). Save everything to MemPalace, then allow compaction to proceed.

```

Seems like it should be possible to work around the always block part here.

**Environment:**
- OS: Ubuntu 24.04
- Python version: 3.1.14

### LINKED ISSUE #949: PreCompact hook blocks indefinitely — no escape condition
hook_precompact() always returns {"decision": "block"} with no exit path,
  causing /compact to be permanently blocked.

  Fix: add session-level flag file (same pattern as Stop hook's stop_hook_active guard).
  See: hooks_cli.py :: hook_stop() for reference implementation.

rusel95:
## 🚫 The problem

You are in a long session. Context fills up. You type `/compact`. Nothing happens.

The precompact hook is blocking compaction. You cannot disable it without also disabling the stop hook. You are stuck.

This was the #1 reported issue after the initial precompact hook shipped. PR #863 fixed the hard-block by making precompact always return `{}` — but that dropped the save nudge entirely. Users now get no signal that compaction is imminent.

## ✅ What this does

Adds an independent toggle for the precompact hook:

```json
// ~/.mempalace/config.json
{"hooks": {"precompact": false}}
```

```bash
# Or via env var
MEMPALACE_HOOKS_PRECOMPACT=false
```

When disabled, `/compact` proceeds immediately — no mine, no delay.

## 🔑 Why independent controls matter

| Scenario | save_interval | precompact |
|----------|--------------|------------|
| Normal use | 15 (default) | true (default) |
| Quiet mode — fewer interruptions | 50 | true |
| Precompact causing issues | 15 | **false** |
| Fully silent | 0 | true |

Disabling `save_interval` does **not** disable precompact. Disabling `precompact` does **not** disable the stop hook. They are independent by design — so you can fix one without breaking the other.

## 🧪 Tests

All 104 existing hook and config tests pass.

```
python -m pytest tests/test_hooks_cli.py tests/test_config.py -q
104 passed
```

Depends on: #1345 (save_interval config, same config.py property pattern)
Closes #949. Closes #906.

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -295,6 +295,39 @@ class MempalaceConfig:
         return self._file_config.get("hooks", {}).get("silent_save", True)
 
     @property
+    def hooks_save_interval(self) -> int:
+        """Number of exchanges between auto-save prompts. 0 = disabled.
+
+        Config: {"hooks": {"save_interval": 30}}
+        Env var (takes precedence): MEMPALACE_HOOKS_SAVE_INTERVAL=30
+        """
+        env_val = os.environ.get("MEMPALACE_HOOKS_SAVE_INTERVAL", "").strip()
+        if env_val:
+            try:
+                return max(0, int(env_val))
+            except ValueError:
+                pass
+        val = self._file_config.get("hooks", {}).get("save_interval", 15)
+        try:
+            return max(0, int(val))
+        except (TypeError, ValueError):
+            return 15
+
+    @property
+    def hooks_precompact(self) -> bool:
+        """Whether the precompact hook fires before context compaction. Default: True.
+
+        Config: {"hooks": {"precompact": false}}
+        Env var (takes precedence): MEMPALACE_HOOKS_PRECOMPACT=false
+        """
+        env_val = os.environ.get("MEMPALACE_HOOKS_PRECOMPACT", "").strip().lower()
+        if env_val in ("0", "false", "no", "off"):
+            return False
+        if env_val in ("1", "true", "yes", "on"):
+            return True
+        return bool(self._file_config.get("hooks", {}).get("precompact", True))
+
+    @property
     def hook_desktop_toast(self):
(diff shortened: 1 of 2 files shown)


rusel95:

Verified against the real `MempalaceConfig` class and `hook_precompact()` function.

**Tests verified:**
- Default precompact = true ✅
- `MEMPALACE_HOOKS_PRECOMPACT=false` → disabled ✅
- `MEMPALACE_HOOKS_PRECOMPACT=true` → enabled ✅
- **Independence test**: `save_interval=0` + no precompact env → precompact still true ✅

This is the critical test — it proves the two controls are truly independent. Disabling periodic saves does not silently disable the last-chance save before compaction.

```
python -m pytest tests/test_hooks_cli.py tests/test_config.py -v
104 passed in 0.59s
```

Full suite: 0 regressions.

PR #1349, opened 2026-05-04
fix(mine): detect concurrent palace holder, exit non-zero with clear error (#1264)

### LINKED ISSUE #1264: `mine`: silent exit when concurrent writer holds chroma lock; should detect live `mcp_server` and back off with clear error
## Summary

On `mempalace 3.3.3`, when a `mempalace.mcp_server` process is already attached to the palace and holding ChromaDB's exclusive write lock, `mempalace mine <path>` exits silently — the only line in the log is the auto-defaults warning that streams *before* any write attempt. There is no error message, no lock-contention notice, no "MCP server holds the store" diagnostic. From the operator's view the mine "ran" — exit was clean, no stderr, no error — but no work happened.

This is the writer-vs-MCP collision case that complements the writer-vs-writer case already filed:

- **#1229 (defense-in-depth, filed this week):** `mempalace.mcp_server` has no single-instance guard; two MCP processes can race on the same palace.
- **PR #976 (merged → v3.3.4):** added `mine_global_lock()` so two concurrent `mempalace mine` invocations collapse to one runner instead of racing.
- **PR #1162 (open):** moves the lock boundary to the `ChromaCollection` seam so MCP / direct-backend writers participate in the same palace lock as `mine`. **This is the right structural fix for the bug below.**
- **This issue:** the failure mode that justifies #1162 — `mine` *vs* a live MCP server, where today the mine just disappears with a misleading single-line log.

## Reproduction

`mempalace 3.3.3`, macOS 26.3.1 (Darwin 25.3.0) on Apple Silicon, Python 3.10. Singleton-wrapped MCP server already running for the palace (the singleton wrapper at `~/.claude/hooks/mempalace-mcp-singleton.sh` is the mitigation referenced in #1229; it guards the `mcp_server` invocation only — it does not cover `mempalace mine` or `mempalace sweep`).

Setup state at 2026-04-28 ~07:28 (timezone: PDT):

- `mempalace.mcp_server` PID held the chroma write lock since 07:11 (17 minutes earlier).
- No other `mempalace mine` processes running (`pgrep -fl 'mempalace mine'` empty).

Operator ran:

```bash
nohup ~/.local/share/uv/tools/mempalace/bin/mempalace mine /Users/<user>/.claude/projects/ \
    > ~/.mempalace/mine.log 2>&1 & disown
```

Resulting `mine.log` — **200 bytes, total**:

```
No mempalace.yaml found in /Users/<user>/.claude/projects — using auto-detected
defaults (wing='projects'). Directories with the same basename will share a wing;
add mempalace.yaml to disambiguate.
```

That is the entire log. Process exited within seconds. No mining output, no progress, no error message about lock contention. Exit code was clean enough that `nohup` produced no additional signal.

The auto-defaults warning streams *before* any write attempt, which is why it's the only thing captured: `mine` then attempts its first commit against ChromaDB, hits exclusive-write contention with the MCP holder, and bails before producing any further output.

## Why this is its own bug, not just a symptom of #1162

PR #1162 is the right structural fix — once `ChromaCollection.add/upsert/update/delete` participate in `mine_palace_lock`, the second writer will get `MineAlreadyRunning` cleanly instead of disappearing into ChromaDB's lock-handling. That solves the underlying race.

But there are two issues here, and #1162 only addresses one:

1. **Correctness (covered by #1162):** two writers should not silently both touch the chroma store. `MineAlreadyRunning` raised at the lock layer is the correct shape.
2. **Observability (not covered by #1162):** when a writer is rejected, the failure message must be loud enough that an operator running `mempalace mine` against a palace held by `mcp_server` sees what happened. A 200-byte log containing only an auto-defaults warning is, by itself, a bug — even if the underlying lock behavior were fully correct, this output is misleading.

The "fail silently with only the auto-defaults warning when another writer holds the lock" behavior is the operator-visible bug. #1162 will likely fix it incidentally (the new `MineAlreadyRunning` raise will hit stderr), but the issue is worth being explicit about so the merged PR's exception handling is verified to actually surface a clear, named contention error to the operator's log — not a generic exception trace, not a swallowed `RuntimeError`.

## Recommended fix per the report

Either (a) `mempalace mine` and `mempalace sweep` should also honor a singleton lock at the chroma store level (i.e. PR #1162's pattern, which is the right structural answer), OR (b) `mempalace mine` should detect a live MCP holder before opening chroma and either back off with a clear error message ("MCP server holds the store, refuse to start mine") or coordinate properly.

Concretely, in `cmd_mine` / the equivalent entry path:

1. Acquire (or attempt-acquire) the palace-wide write lock (`mine_palace_lock` from #976, or its #1162-extended form) **before** the auto-defaults warning, or at least before any operator interpretation could attach to the warning.
2. If acquisition fails because another writer holds the lock, print a one-line error to stderr that includes:
   - The palace path
   - The PID and process kind of the holder ("mempalace.mcp_server PID 55408" / "mempalace mine PID ...")
   - A suggested next step ("wait for the holder to finish, or stop the MCP server before mining")
3. Exit non-zero. The operator's `nohup` log will then carry a useful signal.

## Adjacent reports

- **#1229 (filed):** the MCP-server-vs-MCP-server case. Same incident family, different writer pair.
- **#1212:** Stop hook spawns concurrent `mempalace mine` against the same source path; PID guard (#1023) is insufficient because of TOCTOU on the lock-file create. That's mine-vs-mine; this issue is mine-vs-MCP-server.
- **#1206:** `mempalace mine` processes hanging for 6 hours on a 350MB folder; jphein's diagnosis attributes it to the same orphan-mine pattern as #1202. That's a stuck-mine case after corruption already happened; this issue is the lock-contention case before any corruption (the MCP server is healthy and holding correctly).
- **#1092:** "Concurrent writers (hooks + MCP server + CLI) corrupt the palace on chromadb 1.5.8 — sparse-file bloat + SIGSEGV". Same writer-collision class; that issue documents the corruption symptom, this one documents the silent-exit symptom that comes before the corruption (or in this case, instead of it — the MCP server held the lock correctly, so corruption did not occur, but the mine vanished without trace).
- **#1253:** PreCompact hook double-ingest race; same lack-of-palace-wide-write-lock root cause, different trigger surface.

## Environment

```
mempalace 3.3.3 (uv tool install, ~/.local/share/uv/tools/mempalace/)
chromadb 1.5.x (whatever 3.3.3 pins)
Python 3.10
macOS 26.3.1 (Darwin 25.3.0), Apple Silicon (M-series)
MCP host: Claude Desktop (claude-ai client per protocol handshake)
Singleton wrapper at ~/.claude/hooks/mempalace-mcp-singleton.sh wraps the
mcp_server invocation only — does not wrap mempalace mine / mempalace sweep
```

## What I haven't tested

- This evidence is macOS / Apple Silicon only. Linux and Windows lock-contention behavior may differ — POSIX `fcntl.flock` semantics on macOS APFS are well-behaved here, but I haven't verified the same silent-exit shape on Linux ext4 or Windows.
- I have not tested `mempalace sweep` against a held lock; the same shape is plausible (it's the same class of writer entry point) but unverified.
- I have not tested behavior on `develop` post-#1162. Once that PR lands, the expected outcome is `MineAlreadyRunning` raised cleanly with a clear stderr message — that's the spec the merged behavior should be checked against.
- The auto-defaults warning streaming before any lock attempt is observed in `mine`; I haven't confirmed the same ordering in `sweep` or `init`.

## Related

- #1229 — `mempalace.mcp_server` single-instance guard (filed; defense-in-depth for #976)
- #976 — `mine_global_lock()` + `hnsw:num_threads=1` (merged → v3.3.4)
- #1162 — serialize ChromaCollection writes through palace lock (open; the right structural fix for this issue)
- #1212 — Stop hook spawns concurrent `mempalace mine` (mine-vs-mine; companion case)
- #1206 — `mempalace mine` hung 6h on 350MB folder (post-corruption stuck-mine case)
- #1092 — concurrent writers corrupt the palace (corruption-side symptom of the same collision class)
- #1253 — PreCompact hook double-ingest race (PreCompact-side symptom of the same lack-of-palace-wide-write-lock root cause)

adv3nt3:
Fixes #1264.

## What does this PR do?

Closes the operator-visible silent-exit gap when `mempalace mine` is run while another writer (typically `mempalace.mcp_server`) holds the palace's `chroma.sqlite3` open.

Today, `mine` would:
1. Print the auto-defaults stderr warning from `load_config`.
2. Hit chromadb 1.5.x's concurrent-writer path on the first write attempt.
3. Exit silently — either via a Rust-binding SIGSEGV or with a swallowed lock contention. stdout's banner was lost to block-buffering on signal exit; only the stderr warning survived.

The result was the 200-byte log described in #1264.

This PR adds a **pre-flight detection** that runs *before* any other output and bails with a clear message + non-zero exit when a holder is found. It's complementary to #1162's structural fix at the `ChromaCollection` lock layer — once that lands, the lock-layer raise reaches the same loud `MineAlreadyRunning` handler instead of disappearing.

### `mempalace/palace.py` — new helper
- `detect_palace_holder(palace_path) -> Optional[dict]`
- Uses `lsof -Fpcn` to find the first non-self process holding `<palace>/chroma.sqlite3`
- `_classify_palace_holder(pid, command)` inspects full argv via `ps` to classify the holder as `mempalace.mcp_server`, `mempalace mine`, or the raw command name as fallback
- POSIX-only: returns `None` on Windows or when `lsof` is unavailable, missing, or times out — graceful degrade preserves prior behavior on platforms we can't probe
- Never raises, never blocks more than ~3s (1s lsof + 1s ps with caps)

### `mempalace/miner.py` — wire pre-flight into `mine()`
- Pre-flight runs **before** `load_config`, the banner, and `get_collection`. The auto-defaults warning can no longer leak through as the only line of an operator's log.
- On detected holder: prints a one-line stderr message with palace path, holder PID, process kind, and a suggested next step, then `sys.exit(1)`.
- On `MineAlreadyRunning`: was `return` (clean exit 0 with `"exiting cleanly"` wording, which the issue called out as misleading); now `sys.exit(1)` with a clear `cannot start` message.
- Dry-run skips the pre-flight (it doesn't open chroma).

### `tests/test_detect_palace_holder.py` — new test file
13 tests covering:
- Missing palace dir / missing `chroma.sqlite3` → no detection attempt
- `lsof` unavailable / timeout / non-zero exit → graceful `None`
- Self-PID is filtered out
- MCP server and concurrent-mine classification via mocked `ps`
- Windows path returns `None` (`monkeypatch.setattr(sys, "platform", "win32")`)
- `mine()` exits 1 with clear stderr when a holder is detected, **without** the auto-defaults warning leaking
- `mine()` exits 1 with clear stderr on `MineAlreadyRunning`
- `mine()` proceeds normally when no holder is reported
- Dry-run skips the pre-flight

## How to test

```bash
# unit + integration tests
pytest tests/test_detect_palace_holder.py tests/test_palace_locks.py tests/test_miner.py -v

# end-to-end smoke (POSIX): hold the chroma file open in one shell, mine in another
mkdir -p /tmp/sm/proj /tmp/sm/palace && touch /tmp/sm/palace/chroma.sqlite3
echo 'def main(): pass' > /tmp/sm/proj/main.py
python3 -c "open('/tmp/sm/palace/chroma.sqlite3'); import time; time.sleep(60)" &
sleep 0.5
mempalace --palace /tmp/sm/palace mine /tmp/sm/proj
# Expect:
#   mempalace mine: cannot start — palace at /tmp/sm/palace is held by Python (PID <pid>). Wait for it to finish, or stop the holder before mining.
# Exit code: 1
```

Verified locally on macOS 26.4 (Apple Silicon, Python 3.14): `lsof` correctly identifies the holding PID, the message reaches stderr before any other output, and exit code is 1.

## Scope

- **In scope:** observability — make the failure loud at the `mine` entry point.
- **Out of scope (deferred):**
  - `mempalace sweep` — same writer entry-point class; the issue mentions it as plausible-same-shape but unverified. Suggest a follow-up PR mirroring this fix in `cmd_sweep` once we confirm the same silent-exit behavior there.
  - The structural fix at the chroma write-seam — that's #1162.
  - MCP server's own single-instance guard — that's #1229.

## Behavior change to flag

The `MineAlreadyRunning` handler now exits 1 instead of 0. The previous behavior was documented as "exiting cleanly", but per the issue's recommendation a non-zero exit makes nohup / shell wrappers / hooks see a useful signal. `mempalace.hooks_cli._spawn_mine` does not check the subprocess exit code, so the hook path is unaffected. If any caller relied on exit 0 from "another mine running", they should be updated to expect 1.

## Checklist
- [x] Tests pass (`pytest tests/ -v` — 1502 passed, 1 skipped)
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .` — All checks passed)
- [x] No new dependencies (uses only `subprocess` and `sys` from stdlib)

Refs:
- #1264 (this fix)
- #1162 (structural fix at backend layer — complementary)
- #976 (existing `mine_palace_lock` for mine-vs-mine)
- #1229 (MCP server singleton guard — separate)

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -23,6 +23,7 @@ from .palace import (
     SKIP_DIRS,
     MineAlreadyRunning,
     build_closet_lines,
+    detect_palace_holder,
     file_already_mined,
     get_closets_collection,
     get_collection,
@@ -1019,6 +1020,24 @@ def mine(
             files=files,
         )
 
+    # Pre-flight: detect a concurrent writer (typically `mempalace.mcp_server`)
+    # holding chroma.sqlite3 BEFORE any other output. Without this check
+    # `mine` would print the auto-defaults stderr warning, then attempt its
+    # first chroma write, hit lock contention or a Rust-binding SIGSEGV,
+    # and exit with no diagnostic visible (issue #1264). `mine_palace_lock`
+    # below only catches mine-vs-mine; the structural lock-at-backend fix
+    # lives in #1162. This pre-flight closes the operator-visible gap on
+    # POSIX and degrades silently on Windows / when `lsof` is unavailable.
+    holder = detect_palace_holder(palace_path)
+    if holder is not None:
+        print(
+            f"mempalace mine: cannot start — palace at {palace_path} is held by "
+            f"{holder['kind']} (PID {holder['pid']}). "
+            "Wait for it to finish, or stop the holder before mining.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
     try:
         with mine_palace_lock(palace_path):
             return _mine_impl(
@@ -1034,11 +1053,11 @@ def mine(
             )
     except MineAlreadyRunning:
         print(
-            f"mempalace: another `mine` is already running against "
-            f"{palace_path} — exiting cleanly.",
+            f"mempalace mine: cannot start — another `mempalace mine` is already "
+            f"running against {palace_path}. Wait for it to finish before mining.",
             file=sys.stderr,
         )
-        return
+        sys.exit(1)
 
 
 def _mine_impl(
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -9,7 +9,10 @@ import hashlib
 import logging
 import os
 import re
+import subprocess
+import sys
 import threading
+from typing import Optional
 
 from .backends.chroma import ChromaBackend
 
@@ -273,6 +276,89 @@ def upsert_closet_lines(closets_col, closet_id_base, lines, metadata):
 
     _flush()
     return closets_written
+
+
+def detect_palace_holder(palace_path: str) -> Optional[dict]:
+    """Best-effort detection of a process holding ``<palace>/chroma.sqlite3`` open.
+
+    Used by ``mine`` as a pre-flight check before opening ChromaDB so a
+    concurrent writer (typically ``mempalace.mcp_server``) is reported with
+    a clear stderr error instead of a silent SIGSEGV / lock contention.
+
+    POSIX implementation: shells out to ``lsof``. On Windows or if ``lsof``
+    is unavailable, returns ``None`` (graceful degrade — the caller proceeds
+    as before). Never raises; never blocks more than ~3s.
+
+    Returns a dict ``{"pid": int, "command": str, "kind": str}`` for the
+    first non-self holder found, or ``None`` when no holder is detected.
+
+    ``kind`` is a coarse classification — one of:
+      * ``"mempalace.mcp_server"`` — MCP server process
+      * ``"mempalace mine"`` — concurrent ``mine`` invocation
+      * the raw command name otherwise
+    """
+    db_path = os.path.join(os.path.expanduser(palace_path), "chroma.sqlite3")
+    if not os.path.isfile(db_path):
+        return None
+    if sys.platform == "win32":
+        return None
+    try:
+        result = subprocess.run(
+            ["lsof", "-Fpcn", "--", db_path],
+            capture_output=True,
+            text=True,
+            timeout=2,
+        )
+    except (subprocess.TimeoutExpired, FileNotFoundError, OSError):
+        return None
+    if result.returncode not in (0, 1):
+        # lsof returns 1 when no holders are found on some platforms;
+        # any other non-zero is a real failure — degrade silently.
+        return None
+
+    self_pid = os.getpid()
+    pid: Optional[int] = None
+    for line in result.stdout.splitlines():
+        if not line:
+            continue
+        tag, value = line[0], line[1:]
+        if tag == "p":
+            try:
+                pid = int(value)
+            except ValueError:
+                pid = None
+        elif tag == "c" and pid is not None:
+            if pid == self_pid:
+                pid = None
+                continue
+            kind = _classify_palace_holder(pid, value)
+            return {"pid": pid, "command": value, "kind": kind}
+    return None
+
+
+def _classify_palace_holder(pid: int, command: str) -> str:
+    """Return a coarse process kind by inspecting full argv via ``ps``.
+
+    Falls back to the command name from ``lsof`` when ``ps`` is unavailable
+    or fails. Best-effort only — never raises.
+    """
+    if sys.platform == "win32":
+        return command or "unknown"
+    try:
+        result = subprocess.run(
+            ["ps", "-p", str(pid), "-o", "args="],
+            capture_output=True,
+            text=True,
+            timeout=2,
+        )
+    except (subprocess.TimeoutExpired, FileNotFoundError, OSError):
+        return command or "unknown"
+    args = result.stdout.strip()
+    if "mempalace.mcp_server" in args or " mcp" in f" {args}":
+        return "mempalace.mcp_server"
+    if "mempalace mine" in args or "mempalace sweep" in args:
+        return "mempalace mine"
+    return command or "unknown"
 
 
 @contextlib.contextmanager
--- a/tests/test_detect_palace_holder.py
+++ b/tests/test_detect_palace_holder.py
@@ -1,0 +1,270 @@
+"""Tests for detect_palace_holder + mine pre-flight refusal (issue #1264).
+
+The bug: when an MCP server (or any other writer) holds the palace's
+chroma.sqlite3 open, ``mempalace mine`` would print only the auto-defaults
+warning and exit with no diagnostic — the chroma open would either block
+silently or SIGSEGV under chromadb 1.5.x's concurrent-writer behavior.
+
+The fix detects a live holder before opening chroma and refuses with a
+clear stderr message + non-zero exit. POSIX uses ``lsof``; Windows and
+hosts without ``lsof`` degrade silently to None.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import os
+import subprocess
+import sys
+
+import pytest
+
+from mempalace import miner, palace
+from mempalace.palace import detect_palace_holder
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


adv3nt3:
Heads up: the macOS CI failure is unrelated to this PR.

`tests/test_cli.py::test_cmd_compress_output_readable_via_get_closets_collection` failed with `httpx.ConnectError: [Errno 8] nodename nor servname provided, or not known` — a transient DNS resolution failure on the macOS runner when ChromaDB tried to fetch the embedding model from Hugging Face. The same test passed on develop's prior CI run (25272450918).

This PR adds no network calls — `detect_palace_holder` only shells out to local `lsof` / `ps`. All Linux runs (3.9 / 3.11 / 3.13) pass; lint passes; only the macOS runner hit this flake.

Could a maintainer re-run the failed macOS job? I can't trigger it from my fork without admin rights. If you'd prefer I push a no-op commit to retrigger CI, happy to do that — let me know.

adv3nt3:
Rebased onto develop to resolve the merge conflict (single-line import collision in `palace.py` from #1377's `threading` addition). Full test suite + ruff still pass.

adv3nt3:
Closing — superseded by #1413 (merged), which solves #1264 with a cleaner approach:

- Records holder PID + argv inside the existing lock file via `a+` open mode (no external `lsof`/`ps` calls)
- Works on all platforms uniformly (this PR's pre-flight degrades silently on Windows / hosts without `lsof`)
- No fork+exec overhead on every mine startup
- Combined with #1162 (also merged), MCP and direct backend writers now participate in `mine_palace_lock`, so the lock file is the source of truth for "who holds the palace"

Thanks @igorls for the better fix. Leaving this PR closed; the test ideas from `tests/test_detect_palace_holder.py` are not worth porting since the lock-file approach exercises the same surface from `tests/test_palace_locks.py`.

PR #1350, opened 2026-05-04
feat: add SQLite3 database miner

daedalus:
## Summary

Adds a new `--mode sqlite` to `mempalace mine` for ingesting SQLite3 database files into the palace.

### Changes
- **`mempalace/sqlite_miner.py`**: New miner that extracts schema (tables, indexes, views) and sampled row data from SQLite3 files, chunks content by section, and files into drawers
- **`mempalace/cli.py`**: Added `sqlite` as a valid choice for `--mode` in the mine command
- **`tests/test_sqlite_miner.py`**: 17 unit tests covering all pure functions (read, chunk, detect room, scan) — all passing, ruff clean

### Usage
```bash
mempalace mine /path/to/db/dir --mode sqlite
```

### Behavior
- Extracts and files: schema SQL, table row counts, and up to 1000 rows per table as INSERT statements
- Chunks by section (schema, table data) with 800-char drawer size
- Room auto-detection: `schema`, `auth`, `logs`, or `data` based on filename/content
- Respects file-already-mined checks and mine locks for concurrent safety

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -512,6 +512,17 @@ def cmd_mine(args):
             dry_run=args.dry_run,
             extract_mode=args.extract,
         )
+    elif args.mode == "sqlite":
+        from .sqlite_miner import mine_sqlite
+
+        mine_sqlite(
+            directory=args.dir,
+            palace_path=palace_path,
+            wing=args.wing,
+            agent=args.agent,
+            limit=args.limit,
+            dry_run=args.dry_run,
+        )
     else:
         from .miner import mine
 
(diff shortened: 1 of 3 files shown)


mjc:
As a contributor rather than a maintainer, my read is that this is a nice addition overall - the SQLite miner seems useful, and the pure-function coverage is a good start. I had a few edge cases I think are worth another look before this lands.

First, I wonder if room detection may end up being a bit more `schema`-heavy than intended in real runs. Since `read_sqlite_file()` includes schema text in the extracted content, `detect_sqlite_room()` may hit the `CREATE TABLE` / `CREATE INDEX` check before `auth` or `logs` get much of a chance. The helper tests for `detect_sqlite_room()` look fine, but an end-to-end case here would help confirm the semantic buckets are actually reachable in practice.

I also think there may be a corner case around unreadable or non-SQLite `.db` files. `read_sqlite_file()` returns an error string, and it looks like that string may still flow through the normal chunking + filing path. If that reading is right, it might be better to skip those files explicitly or surface them as warnings instead of storing the error text as drawer content.

Another thing I’d sanity-check is WAL mode. The re-mine logic appears to key off the main `.db` file mtime, so I’m not sure whether changes that are still sitting in `-wal` would always be picked up for active databases.

Finally, the pure-function tests are helpful, but I think a small end-to-end test around `process_sqlite_file()` / `mine_sqlite()` would go a long way here - especially for room routing, already-mined behavior, and unreadable-file handling.

PR #1102, opened 2026-04-22
fix(normalize): handle UTF-8 BOM in transcript files

arnoldwender:
## What and Why

`normalize()` opens transcript files with `encoding='utf-8'`. When a file has a UTF-8 BOM prefix (`\xef\xbb\xbf`) — common from Windows exports of Claude Code JSONL sessions — `json.loads()` raises `JSONDecodeError` on every line because the first line starts with `﻿{`. `_try_claude_code_jsonl` silently skips all lines and the file falls through as raw unstructured text, discarding all message content.

## Root Cause

`normalize.py:124` — `encoding="utf-8"` does not strip the BOM; `encoding="utf-8-sig"` does.

## Fix

```python
# Before
with open(filepath, "r", encoding="utf-8", errors="replace") as f:

# After
with open(filepath, "r", encoding="utf-8-sig", errors="replace") as f:
```

Python's `utf-8-sig` codec strips the BOM on read and is fully backward-compatible with BOM-free files. No behavioral change on Linux/macOS files.

## Reproduction

```python
import tempfile, os
from mempalace.normalize import normalize

# BOM-prefixed JSONL (Windows Claude Code export)
bom_file = tempfile.mktemp(suffix=".jsonl")
with open(bom_file, "wb") as f:
    f.write(b"\xef\xbb\xbf")  # UTF-8 BOM
    f.write(b'{"type":"human","message":{"content":"hello"}}\n')

result_before = normalize(bom_file)  # returns raw JSON lines, not a transcript
# After fix: returns properly formatted transcript with "hello"
```

## Tests

All 107 `test_normalize.py` tests pass. All 1066 tests pass.

Closes #1034 (partial — encoding fix for `normalize.py`; other files addressed in separate PRs)

PR #1100, opened 2026-04-22
fix(closet_llm): replace non-ASCII symbols in progress output (#1034)

arnoldwender:
## What and Why

`closet_llm.py` prints `✓`, `✗`, and `—` (U+2713, U+2717, U+2014) during `mempalace closets regen`. These characters cannot be encoded in GBK (Windows PowerShell/CMD default), causing `UnicodeEncodeError` on the same code path fixed in `miner.py` via #681.

`cli.py` has the same problem on the `sweep` error path (`✗`).

## Root Cause

- `closet_llm.py:244` — `—` in dry-run banner
- `closet_llm.py:266` — `✗` + `—` in LLM-failed message
- `closet_llm.py:303` — `✓` + `—` in success message
- `cli.py:187` — `✗` in sweep error message

## Change Summary

Replace non-ASCII symbols with ASCII equivalents: `[OK]`, `[FAIL]`, `[!]`, and `-`.

## Test Plan

- [x] All 1066 existing tests pass
- [x] `ruff check` clean

Closes #1034

PR #1360, opened 2026-05-05
Fix/utf8 stdout windows

QuiqueMH:
## Checklist
- [ ] No hardcoded paths


igorls:
Thanks! Windows non-ASCII output is already addressed on develop via #1100 (ASCII glyphs) and #1102 (UTF-8 BOM). Closing as overlapping.

PR #1357, opened 2026-05-05
fix(repair): preflight poisoned max_seq_id before rebuild

fatkobra:
## What does this PR do?

Fixes #1295.

This makes default `mempalace repair` detect poisoned `max_seq_id` bookmarks before it enters the legacy rebuild path.

When the #1295 corruption is present, repair now runs the existing non-destructive `max_seq_id` repair and stops, instead of rebuilding the collection from only currently-visible embeddings.

## What changed

- Added `maybe_repair_poisoned_max_seq_id_before_rebuild()`.
- Wired that preflight into:
  - `mempalace repair`
  - `repair.rebuild_index()`
- If poisoned `max_seq_id` rows are found:
  - run the existing `repair_max_seq_id()` path
  - keep `embeddings_queue` intact
  - skip the legacy rebuild path
- Added regression tests that verify:
  - queued writes remain in `embeddings_queue`
  - poisoned bookmarks are repaired
  - `rebuild_index()` does not enter the legacy Chroma collection read/count/rebuild path when the max_seq_id preflight handles the issue

## Why

#1295 reports a palace where writes through MCP returned success, but never became readable. At the SQLite layer, `embeddings_queue` kept growing while `embeddings` stayed static.

The root cause was a poisoned `max_seq_id` row with an impossibly large value. Chroma believed it had already consumed queue rows that were actually still pending.

Current `develop` already has the narrow `repair_max_seq_id()` primitive, but default `mempalace repair` can still proceed into the legacy rebuild path. That rebuild only preserves rows already visible in `embeddings`; queued writes can be lost.

This PR makes repair prefer the narrow, non-destructive bookmark fix whenever the #1295 corruption is detected.

## How to test

```bash
ruff format mempalace/repair.py mempalace/cli.py tests/test_repair.py
ruff check mempalace/repair.py mempalace/cli.py tests/test_repair.py
python -m pytest tests/test_repair.py -q
python -m pytest tests/test_repair.py tests/test_cli.py -q
python -m pytest tests/ -v


## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`)
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`)

PR #1364, opened 2026-05-05
fix(repair): preflight SQLite integrity before rebuild

### LINKED ISSUE #1362: `repair --yes` aborts on SQLite-layer index corruption (code 779) before reaching HNSW rebuild
Splitting this out from #1218 at @fatkobra's [suggestion](https://github.com/MemPalace/mempalace/issues/1218#issuecomment-4326330000) — it's a distinct failure mode from the HNSW segfault / `link_lists.bin` explosion that #1218 and #1339 cover.

## Failure shape

`mempalace repair --yes` aborts before doing any rebuild work, with:

```
chromadb.errors.InternalError: Database error: error returned from database:
(code: 779) database disk image is malformed
```

The stack lands inside `delete_collection()`, which `repair.rebuild_index()` calls early in the rebuild path (drop-then-recreate of `mempalace_drawers`).

So this is a third self-failure mode for `repair` itself, alongside:

- `count()` segfault on entry (#1218 original)
- silently-corrupt rebuild output (#1218 first follow-up)
- this one: SQLite-layer abort on `delete_collection` _before_ rebuild

## Environment

- mempalace 3.3.4
- chromadb 1.5.7
- Windows 11
- ~49 K drawers, ~518 K `embedding_metadata` rows

## Why this is _not_ HNSW corruption

This palace's HNSW segments were healthy by the #1339 ratio test (`link_lists.bin / data_level0.bin` ≈ 0.005, no link-list explosion). `repair-status` reported a vanilla DIVERGED state (`sqlite=48876 / hnsw=10405`, ~79 % gap from missed compactions) — exactly what `repair --yes` exists to fix. User-visible state was `Internal error: Error querying knn` on search, and `repair` is the documented remediation.

The corruption was one level down, in the SQLite layer:

```
$ sqlite3 chroma.sqlite3 "PRAGMA integrity_check;"
*** in database main ***
Page 4 of B-tree {N1}: ...
Page 5 of B-tree {N2}: ...
Page 6 of B-tree {N3}: ...
Page 7 of B-tree {N4}: ...
Page 8 of B-tree {N5}: ...   # FTS5 shadow table
```

Five corruptions, all on indexes — four secondary B-tree indexes plus one FTS5 shadow page. **Zero data-level corruption**: 49 029 embeddings + 518 440 `embedding_metadata` rows accessible via direct `SELECT`. The corruption is confined to derived structures.

`delete_collection()` happens to read those exact derived structures (collection lookup hits an FTS5/secondary path), so it raises code 779 and `repair` exits before reaching its rebuild loop.

## Recovery procedure that worked (offline, on backup)

1. Snapshot `palace/` to `palace.backup-<ts>/` before doing anything.
2. Stop all writers (close any Claude Code session with the MCP server attached).
3. Open `chroma.sqlite3` from the backup with the sqlite CLI and dump-reload to a fresh DB (`.recover` if available, else `.dump`) — both ignore the corrupted index pages because indexes are derived structures.
4. The reloaded DB carries over the FTS5 shadow tables (`embedding_fulltext_search`, `_config`, `_content`, `_data`, `_docsize`, `_idx`) but the FTS5 index data is what was corrupt — drop the FTS5 virtual table set and recreate the virtual table from the intact `embedding_metadata` rows.
5. Verify: `PRAGMA integrity_check` → `ok`, `SELECT COUNT(*) FROM embeddings` matches pre-corruption count, FTS5 row count matches.
6. Swap the reloaded `chroma.sqlite3` into `palace/` (rename old aside, copy new in place).
7. `mempalace status` should now read clean. `mempalace repair --yes` runs to completion (it can now `delete_collection` cleanly and rebuild HNSW from intact metadata).

Numbers in my case: cleaned SQLite went from 299.5 MB → 303 MB (slight expansion from FTS5 reindex), `repair` then took ~18 min for 48 876 drawers, exit 0, divergence dropped from 38 471 to 1 441 (within flush-lag tolerance), search works.

## Suggested complement to `repair`

`repair.rebuild_index()` could run `PRAGMA integrity_check` (or a cheaper `quick_check`) on `chroma.sqlite3` _before_ its first `delete_collection()`, and if it returns anything other than `ok`:

- Either auto-stage a dump-reload into a temp file and swap (with backup) — symmetric with how the HNSW path could subprocess-isolate;
- Or fail-fast with a clear _"SQLite-layer corruption detected, manual recovery required (see <link>)"_ instead of the opaque code-779 traceback.

Either way, today the chain is: HNSW divergence → user runs `repair --yes` → `repair` crashes on independent SQLite-index corruption → user is staring at a stack trace inside `chromadb.api.rust._delete_collection` with no clear next step. Catching the SQLite-layer integrity question before the chromadb API is even invoked would surface a different, recoverable failure mode.

## Adjacent observation: `embeddings_queue` keeps growing during diagnostic

Same shape as [an earlier note in #1218](https://github.com/MemPalace/mempalace/issues/1218) — the auto-file Stop hook kept enqueueing embeddings while the palace was crashed (compactor down, `automatically_purge` a no-op). Restated only because it's reproducible: every diagnostic session adds pollution that has to be cleaned up before the next restore attempt. Probably worth a `palace_loadable()` probe in the hook, or a dead-letter destination when no segment is available.

## Triangulation data

- **Trigger event** unknown — palace was idle for ~36 h before this surfaced; possible that an earlier interrupted compaction left the indexes in this state and only the next `delete_collection` exposed it. SQLite was last written ~10 min before I noticed.
- **chromadb version writing the palace**: 1.5.7 throughout (no migration in play).
- **Original `palace.backup-*`** (used for the offline dump-reload) preserved untouched. Final palace running cleanly.

Happy to share the corrupted-indexes SQLite file if useful for regression — separate from the #1339 link-bloat reproducer, much smaller (~286 MB), different failure surface.

fatkobra:
## What does this PR do?

Fixes #1362.

This makes `mempalace repair --yes` run a SQLite `PRAGMA quick_check` before it reaches Chroma's destructive `delete_collection()` rebuild step.

If SQLite reports corruption, repair now stops with a clear recovery message instead of surfacing an opaque Chroma/SQLite code 779 traceback from inside `delete_collection()`.

## What changed

- Added `sqlite_integrity_errors()`.
  - Opens `chroma.sqlite3` read-only.
  - Runs `PRAGMA quick_check`.
  - Returns any non-`ok` messages as repair-blocking errors.
- Added `print_sqlite_integrity_abort()`.
  - Prints a clear explanation that SQLite-layer corruption was detected.
  - Points the user toward offline SQLite recovery before re-running repair.
- Wired the preflight into `repair.rebuild_index()` after extraction/truncation safety and before:
  - backing up `chroma.sqlite3`
  - `backend.delete_collection(...)`
  - collection recreation
- Added regression tests for:
  - healthy SQLite quick_check
  - unreadable SQLite file reporting
  - #1362-style abort before `delete_collection()`

## Why

#1362 reports a palace where HNSW was not the corrupt layer. The HNSW segment looked healthy, but `repair --yes` crashed inside Chroma's `delete_collection()` with:

```text
(code: 779) database disk image is malformed

## How to test
ruff format mempalace/repair.py tests/test_repair.py
ruff check mempalace/repair.py tests/test_repair.py
python -m pytest tests/test_repair.py -q
python -m pytest tests/ -v

## Checklist
- [x] No hardcoded paths


igorls:
Maintainer-edit: merged develop in to resolve conflicts opened by #1285 (temp-staging rebuild) and #1312 (collection_name plumbing) after this branch was authored.

- \`mempalace/repair.py\`: kept this branch's \`sqlite_integrity_errors()\` and \`print_sqlite_integrity_abort()\`; took develop's \`rebuild_index\` signature with the \`collection_name\` parameter. Normalized the helper's print indent to 2 spaces to match the rest of the file.
- \`tests/test_repair.py\`: kept both this branch's preflight tests and develop's \`rebuild_from_sqlite\` / configured-collection coverage. Took develop's temp-staging assertion shape for \`test_rebuild_index_success\`.
- **Test fixture update**: 7 \`test_rebuild_index_*\` tests used \`sqlite_path.write_text("fake")\` — that fails PRAGMA quick_check, so the new preflight aborted before the rebuild logic the tests actually exercise. Replaced with \`sqlite3.connect(str(sqlite_path)).close()\` (empty real SQLite DB that passes quick_check), preserving the tests' intent.

Local: 1618 tests pass, ruff lint+format clean against the 0.4.x CI pin.

igorls:
Round-2 merge: #1357 (max_seq_id preflight) merged into develop while this branch was in CI, opening a fresh conflict between the two preflight helpers. Kept both — they check distinct corruption classes:

- \`sqlite_integrity_errors\` (this PR) — PRAGMA quick_check before \`delete_collection()\`
- \`maybe_repair_poisoned_max_seq_id_before_rebuild\` (#1357) — un-poisons max_seq_id rows so queued writes aren't dropped

Both run as separate preflights inside \`rebuild_index\`; non-overlapping coverage in tests too. Local: 1623 tests pass on the merge state, ruff lint+format clean.

PR #1382, opened 2026-05-06
fix(benchmarks): UTF-8 encoding + ASCII print chrome on Windows for the 3 sibling runners

jphein:
## Summary

Follow-up to #1204 (LongMemEval Windows encoding) per @mschultheiss83's request. Applies the same narrow Windows correctness pattern to the three sibling benchmark runners that share the same shape:

- `benchmarks/locomo_bench.py`
- `benchmarks/membench_bench.py`
- `benchmarks/convomem_bench.py`

## What changed (per file)

- All local `open(path)` / `open(path, "w")` calls now pass `encoding="utf-8"` (was inheriting the platform default — `cp1252` on Windows, `GBK` on Chinese Windows). Mirrors #1204's pattern at `longmemeval_bench.py:2917 / :2927 / :2957 / :2998 / :3031`.
- Non-ASCII separator characters in `print(...)` chrome replaced with ASCII (`─` → `-`, `→` → `->`) so a default `cp1252` console doesn't raise `UnicodeEncodeError` mid-run. Comments/docstrings (which never hit stdout) are untouched.
- `urllib.request.urlopen(...)` calls intentionally left alone — they don't open local files, and the original #1204 audit didn't flag them.

Total diff: +16 / -16 across 3 files. No behavior change beyond Windows correctness.

## Audit provenance

The file selection isn't speculative — jphein already audited the codebase in [#1204's review thread](https://github.com/MemPalace/mempalace/pull/1204#issuecomment-2849049770) and flagged the exact `open(...)` lines in these three files. @mschultheiss83 asked for the follow-up at https://github.com/MemPalace/mempalace/pull/1204#issuecomment-2851275537. This PR is just the mechanical follow-through.

## Out of scope

- The pre-existing `C901` ruff finding on `locomo_bench.py:621` (`run_benchmark` complexity 36) is from upstream `develop` (verified by stash + re-lint on stock branch); not from this PR.
- Non-encoding-related Windows fixes (path separators, line-ending handling) — those would be a different audit.

## Test plan
- [ ] CI green on Linux / macOS / Windows
- [ ] Manual smoke (Linux): each of the three runners imports without error and the read/write paths exercised by `--cache-dir` arguments don't regress
- [ ] No new ruff lint errors introduced (verified locally — pre-existing C901 unchanged)

### DIFF
--- a/benchmarks/convomem_bench.py
+++ b/benchmarks/convomem_bench.py
@@ -70,13 +70,13 @@ def download_evidence_file(category, subpath, cache_dir):
     os.makedirs(os.path.dirname(cache_path), exist_ok=True)
 
     if os.path.exists(cache_path):
-        with open(cache_path) as f:
+        with open(cache_path, encoding="utf-8") as f:
             return json.load(f)
 
     print(f"    Downloading: {category}/{subpath}...")
     try:
         urllib.request.urlretrieve(url, cache_path)
-        with open(cache_path) as f:
+        with open(cache_path, encoding="utf-8") as f:
             return json.load(f)
     except Exception as e:
         print(f"    Failed to download {url}: {e}")
@@ -89,7 +89,7 @@ def discover_files(category, cache_dir):
     cache_path = os.path.join(cache_dir, f"{category}_filelist.json")
 
     if os.path.exists(cache_path):
-        with open(cache_path) as f:
+        with open(cache_path, encoding="utf-8") as f:
             return json.load(f)
 
     try:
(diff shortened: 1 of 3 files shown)


mschultheiss83:
@bensig  @milla-jovovich please review...

jphein:
Trivial Windows fix: adds explicit `encoding="utf-8"` to the local JSON I/O in the three benchmark runners and swaps the non-ASCII console chrome (`─`, `→`) for ASCII so default cp1252 terminals don't raise `UnicodeEncodeError`. +16/-16 across the three sibling runners, no behavior change off Windows. Thanks @mschultheiss83 for flagging interest above — should be a quick one to land. Merges clean on `develop`.

Issue #1315, opened 2026-05-02
Wing-scoped search returns 'Internal error: Error finding id' for freshly-mined wings (HNSW unflushed)


joaodotwork:
## Issue

`mempalace_search` (and the equivalent CLI) returns
`{"error": "Search error: Error executing plan: Internal error: Error finding id"}`
when the `wing` filter targets a **freshly-mined** wing whose HNSW
segment metadata hasn't been flushed yet.

The data is correctly stored in SQLite (visible via
`mempalace_status`, `mempalace_list_wings`, and
`mempalace_list_drawers --wing <wing>`) — only the wing-scoped search
path errors. Unscoped search works (it just won't reliably surface
content from the new wing).

This affects the natural workflow of "mine a new wing → immediately
query it to verify content landed where expected."

## Steps to reproduce

```bash
# Setup: existing palace with several wings already mined
mempalace status
# → working palace, multiple wings, all queryable

# Add a new wing
mkdir -p /tmp/probe-mine
cp some-conversation-export.json /tmp/probe-mine/
mempalace mine --mode convos --wing my-new-wing /tmp/probe-mine
# → Drawers filed: 105

# Verify status (works)
mempalace status
# → my-new-wing  technical  105 drawers   ✓

# Verify drawers exist (works)
mempalace_list_drawers wing="my-new-wing" limit=3
# → returns 3 drawers including the source content   ✓

# Try wing-scoped search (FAILS)
mempalace_search query="anything in here" wing="my-new-wing"
# → Search error: Error executing plan: Internal error: Error finding id

# Reconnect doesn't help
mempalace_reconnect
mempalace_search query="anything in here" wing="my-new-wing"
# → same error
```

## Root cause (probable)

`mempalace repair-status` after the failure shows:

```
[drawers]
  sqlite count:   1,328
  hnsw count:     (no flushed metadata yet)
  status:         UNKNOWN
  note:           HNSW segment metadata not yet flushed; skipping
```

The HNSW vector index hasn't flushed the new wing's segment metadata.
SQLite knows the drawers exist; the chroma `where` filter on `wing`
needs the HNSW collection to be aware of the new docs to resolve the
filter. The error message *"Error finding id"* matches: the search
planner is trying to look up an id (collection? segment?) that exists
in SQLite but not yet in HNSW.

## Why this matters

The **happy path of "mine then query"** is broken until chroma's flush
threshold trips (which appears to be size-based, not time-based —
[...]

Issue #1082, opened 2026-04-21
MCP tool_search returns "Error finding id" when wing-scoped to a convos-mined wing (CLI works, unscoped MCP works)


raphaelsamy:
## Summary

MCP `mempalace_search(query=..., wing=<convos-mined-wing>)` fails with `"Search error: Error executing plan: Internal error: Error finding id"` for any wing that was created via `mempalace mine ... --mode convos`. The **same query via CLI succeeds** (`mempalace search "..." --wing <wing>`), and **unscoped MCP search also succeeds** and returns hits from the same wing mixed with others. Wing-scoped MCP search on `projects`-mined (code) wings is fine. The failure is specifically the combination of "MCP transport + `wing` filter + convos-mined wing".

## Environment

- **mempalace**: 3.3.2 (from PyPI — `pip show mempalace` → `Version: 3.3.2`, sdist+wheel from https://github.com/MemPalace/mempalace/releases/tag/v3.3.2)
- **Install command**: `uv pip install --python $(pyenv which python3) mempalace` (plugin hooks + MCP server need the system `python3` on PATH to import `mempalace`; `uv tool install` isolates it and hooks fail)
- **Python**: 3.12.11 via pyenv
- **OS**: macOS, Darwin 25.3 (macOS 15 series, arm64)
- **Claude Code plugin**: `MemPalace/mempalace` marketplace branch `develop` — plugin `.claude-plugin/plugin.json` v3.3.2, MCP server launched as `python3 -m mempalace.mcp_server`
- **Vector store**: ChromaDB (default, local PersistentClient, no custom tenant/collection config)
- **Embedding model**: `all-MiniLM-L6-v2` (default)
- **Identity file**: `~/.mempalace/identity.txt` present
- **Palace location**: default (`~/.mempalace/palace/`)

## What happens

Given a palace that contains:
- At least one **`projects`-mined wing** (e.g. `sample_repo`, created via `mempalace mine /path/to/repo`)
- At least one **`convos`-mined wing** (e.g. `chats_general`, created via `mempalace mine /path/to/chat-folder --mode convos --extract general --wing chats_general` where the folder contains Claude Code session JSONL files)

Then from an MCP client connected to `python3 -m mempalace.mcp_server`:

| Call | Result |
|---|---|
| `mempalace_search(query="decisions")` — **unscoped** | OK — returns hits from the convos wing mixed with others |
[...]

PR #1396, opened 2026-05-07
fix(mcp): retry tool_search once on Chroma "Error finding   id" transient (#1315)

JPdeB61:
## Problem                                                                                                     
                                                                                                               
  After a bulk CLI mine, ChromaDB's HNSW segment metadata can be unflushed for ~30-60s. Wing-scoped MCP search
  hits `Internal error: Error finding id` during that window. The existing inode/mtime cache invalidation in
  `mcp_server._get_client` doesn't fix it — `tool_search` doesn't go through that cache. It routes via
  `search_memories` → `palace.get_collection` → `_DEFAULT_BACKEND._client(palace_path)`, which has its own
  per-palace `_clients` + `_freshness` cache that needs to be invalidated separately.

  Closely related to / partial fix for #1315 (and also resembles #1082).

  ## Reproducer

  - Environment: mempalace 3.3.x, chromadb 1.5.x, Windows 11, Python 3.13
  - Long-running MCP server connected (e.g. via Claude Code).
  - From a separate shell: `python -m mempalace mine <docs-dir> --wing <name>` — large enough to take meaningful
  time (in my case ~1,184 drawers in one batch).
  - Immediately call `mempalace_search` via MCP with `wing="<name>"`.

  ```json
  {"error": "Search error: Error executing plan: Internal error: Error finding id"}

  Persists ~30-60s, self-heals. During the window:

  $ python -m mempalace repair-status
  [drawers]
    sqlite count:   1,408
    hnsw count:     (no flushed metadata yet)
    status:         UNKNOWN

  CLI search (fresh process) on the same palace works — its freshly-loaded Chroma client opens segments after the
   flush has caught up.

  Change

  tool_search now wraps its search_memories call with a single retry that:

  - detects "Internal error" / "Error finding id" in the result dict via _is_transient_index_error(),
  - drops the MCP-local cache and _DEFAULT_BACKEND._clients / _freshness for the palace via
  _force_chroma_cache_reset(),
  - sleeps 2s,
  - retries once,
  - tags successful retries with "index_recovered": true so callers can observe when it kicked in.

  Non-transient errors (e.g. validation failures) bypass the retry path entirely.

  Tests

  Three unit tests added to TestSearchTool:

  - test_search_retries_once_on_hnsw_flush_transient — first call returns the transient, second succeeds, asserts
   retry ran and index_recovered is set.
  - test_search_does_not_retry_on_non_transient_error — unrelated errors propagate without retry.
  - test_search_returns_second_error_if_retry_also_fails — persistent transient surfaces the second error rather
  than looping.

  All 12 tests in TestSearchTool pass locally.

  What this does NOT fix

  - tool_check_duplicate and other index-touching tools still error in the same window. They'd need the same
  wrapper or, better, a shared helper.
  - The underlying flush window itself. Options 1-3 from #1315 (auto-flush at end of mine, fail-fast detection,
  SQLite-only fallback) are still on the table for a complete fix. This is option 4 — a low-risk recovery wrapper
   alongside, not a replacement.

  Refs #1315
  Refs #1082

  ---
  Happy to take this in a different direction (e.g. lift the helper into a shared decorator covering all
  index-touching tools) if you'd prefer.

  ---

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -168,6 +168,46 @@ _collection_cache = None
 _palace_db_inode = 0  # inode of chroma.sqlite3 at cache time
 _palace_db_mtime = 0.0  # mtime of chroma.sqlite3 at cache time
 
+
+def _is_transient_index_error(result) -> bool:
+    # Chroma can return "Internal error: Error finding id" during the
+    # HNSW flush window after a bulk CLI mine — SQLite rows are
+    # committed but the binary segment metadata isn't flushed yet.
+    # Self-heals once the flush completes (~30-60s). See issue #1315.
+    if not isinstance(result, dict):
+        return False
+    err = result.get("error", "")
+    return isinstance(err, str) and ("Error finding id" in err or "Internal error" in err)
+
+
+def _force_chroma_cache_reset() -> None:
+    # Drop both the MCP-local client cache and the shared backend's
+    # per-palace cache so the next call rebuilds against the post-flush
+    # state. Without clearing _DEFAULT_BACKEND._clients the retry
+    # would just hit the same stale handle, since tool_search routes
+    # via search_memories -> palace.get_collection -> backend cache.
+    global \
+        _client_cache, \
+        _collection_cache, \
+        _palace_db_inode, \
+        _palace_db_mtime, \
+        _metadata_cache, \
+        _metadata_cache_time
+    _client_cache = None
+    _collection_cache = None
+    _palace_db_inode = 0
+    _palace_db_mtime = 0.0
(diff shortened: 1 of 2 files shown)


JPdeB61:
Opened #1396 with a partial fix — option 4 (retry-once with forced cache invalidation in `tool_search`)
  alongside the three approaches in the OP. Doesn't replace options 1-3 (it only covers `tool_search`, not e.g.
  `tool_check_duplicate`, and doesn't address the underlying flush window) but resolves the immediate
  user-visible failure for the most common path.

  Reproducer + diagnostics + cache-layer notes are in the PR description.

PR #1399, opened 2026-05-07
fix(hooks/precompact): unblock /compact and auto-compact

vasyl-pavlyuchok:
## Problem

The PreCompact hook always ends up returning a `block` decision, preventing both `/compact` (manual) and auto-compact from working for any user installing this plugin.

The current upstream hook routes through:

```
run_mempalace_hook --hook precompact --harness claude-code
```

Which calls `mempalace hook run --hook precompact --harness claude-code`, and that path returns `block`. Net effect: no user can compact while this plugin is active.

## Fix

Have the hook return `{"decision": "allow"}` directly so compaction proceeds.

Session saving is **NOT affected** — it goes through the Stop hook (separate path), which continues to work as designed.

## Verification

Tested in production VPS Claude Code environment:

- `/compact` (manual) now works ✓
- Auto-compact at context threshold now works ✓
- Session saving via Stop hook unaffected ✓
- MemPalace MCP server continues responding correctly ✓

## Notes for maintainers

If the original blocking behavior was intentional (e.g. to enforce session save first), please consider an alternative implementation that allows compaction *and* triggers save — current behavior just blocks all compaction silently, which surprises users.

This patch takes the minimal-impact approach. Happy to iterate on a more nuanced solution if preferred.

### DIFF
--- a/.claude-plugin/hooks/mempal-precompact-hook.sh
+++ b/.claude-plugin/hooks/mempal-precompact-hook.sh
@@ -1,24 +1,8 @@
 #!/bin/bash
-# MemPalace PreCompact Hook — thin wrapper calling Python CLI
-# All logic lives in mempalace.hooks_cli for cross-harness extensibility
-run_mempalace_hook() {
-  if command -v mempalace >/dev/null 2>&1; then
-    mempalace hook run "$@"
-    return $?
-  fi
-
-  if command -v python3 >/dev/null 2>&1 && python3 -c "import mempalace" >/dev/null 2>&1; then
-    python3 -m mempalace hook run "$@"
-    return $?
-  fi
-
-  if command -v python >/dev/null 2>&1 && python -c "import mempalace" >/dev/null 2>&1; then
-    python -m mempalace hook run "$@"
-    return $?
-  fi
-
-  echo "MemPalace hook error: could not find a runnable mempalace command or module" >&2
-  return 1
-}
-
-run_mempalace_hook --hook precompact --harness claude-code
+# MemPalace PreCompact Hook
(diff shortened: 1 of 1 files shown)


mvalentsev:
This wouldn't merge as-is, for reasons that show up clearly against
develop tip:

1. The premise -- that `mempalace hook run --hook precompact` returns
   `block` -- doesn't match the current code. `mempalace/hooks_cli.py`
   `hook_precompact()` ends with `_output({})` after mining the
   transcript and palace synchronously. Empty JSON, not `block`. That
   was the #863 fix, merged 2026-04-15 (in v3.3.1+, published
   2026-04-18). If `/compact` is failing for you, you're likely on a
   pre-3.3.1 install -- bumping to v3.3.4 or current `develop` should
   make it work without touching the hook script.

2. Side effect of the patch: it bypasses the `_ingest_transcript` and
   `_mine_sync` calls the Python path runs synchronously. That is the
   reason the PreCompact hook exists -- capture transcript and project
   mine before compaction loses the tool-output buffer. The Stop hook
   save path is separate and won't cover this.

3. Minor: `{"decision": "allow"}` is not a documented value for the
   top-level `decision` field per
   https://code.claude.com/docs/en/hooks.md -- only `"block"` is. In
   practice Claude Code ignores the unknown value and proceeds, but
   #885 (closes #872) specifically cleaned this up to use `{}`. The
   patch reintroduces the same spec-invalid form.

If `/compact` is genuinely still broken against develop tip, that
would be a fresh bug worth tracking with the version + repro steps.

igorls:
Thanks! The precompact hook on develop already ends with an allow decision (since #863); this change would regress precompact mining. Closing as superseded.

PR #1400, opened 2026-05-07
fix(entity-registry): clean up temp file on save failure

### LINKED ISSUE #1373: entity_registry: .tmp sidecar leaks if write fails before os.replace
## Background

#1215 made \`EntityRegistry.save()\` atomic via the standard temp-file + fsync + \`os.replace\` pattern. Crash-mid-write durability is solid: the previous registry remains intact on any kernel-visible failure.

There is a remaining edge case: if \`f.write()\`, \`f.flush()\`, or \`os.fsync()\` raise (disk full, permissions flip, broken FUSE mount, IO error), the \`.tmp\` sidecar is left behind. The \`with\` block closes the fd but does not unlink. Subsequent saves overwrite the same temp filename, so this isn't a leak that grows unboundedly — but it does litter the palace directory and obscures diagnostics, since a user inspecting \`entity_registry.json.tmp\` after a crash currently can't tell whether it's the in-flight write of an ongoing process or stale debris.

## Proposal

Wrap the temp-write + fsync + replace block in \`try/except\`. On any exception, attempt \`tmp_path.unlink(missing_ok=True)\` before re-raising. The new test \`test_save_preserves_previous_on_serialization_failure\` only forces \`os.replace\` to fail (after the temp is already written and about to be renamed away), so this leak path is currently uncovered.

## Acceptance

- Patch \`entity_registry.py:325-352\` (line numbers from the merged commit) to wrap the body in try/except + tmp cleanup.
- Add an assertion to the existing test or a new test: \`assert not (tmp_path / \"entity_registry.json.tmp\").exists()\` after a forced fsync / write failure.
- Existing 31 \`test_entity_registry.py\` tests still pass.

Surfaced during review of #1215.

fatkobra:
## What does this PR do?

Fixes #1373.

This makes `EntityRegistry.save()` remove the sibling `.tmp` sidecar if the atomic-save path fails before `os.replace()` completes.

## What changed

- Wrapped the temp-write / flush / fsync / chmod / replace block in `try/except`.
- On any exception before the save completes:
  - attempt to unlink `entity_registry.json.tmp`
  - ignore cleanup failure
  - re-raise the original exception
- Kept the existing behavior where the parent directory is fsynced after a successful `os.replace()`.
- Added regression coverage for:
  - `os.replace()` failure leaving the previous registry intact and no `.tmp` sidecar
  - `os.fsync()` failure leaving the previous registry intact and no `.tmp` sidecar

## Why

The registry save path is already atomic for crash-mid-write durability, but normal Python-visible failures such as disk-full, permission changes, broken mounts, or fsync errors could leave `entity_registry.json.tmp` behind.

That temp file does not grow unbounded because the next save reuses the same filename, but it creates confusing stale debris in the palace directory.

## How to test

```bash
ruff format mempalace/entity_registry.py tests/test_entity_registry.py
ruff check mempalace/entity_registry.py tests/test_entity_registry.py
python -m pytest tests/test_entity_registry.py -q
python -m pytest tests/test_entity_registry.py tests/test_onboarding.py -q
python -m pytest tests/ -v
```

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/entity_registry.py
+++ b/mempalace/entity_registry.py
@@ -321,22 +321,39 @@ class EntityRegistry:
             self._path.parent.chmod(0o700)
         except (OSError, NotImplementedError):
             pass
+
         # Atomic write: serialize to a sibling temp file in the same dir
         # (so os.replace stays on one filesystem), fsync, then rename over
         # the target. A crash mid-write leaves the previous registry intact
         # instead of a half-written file or an empty file from the truncate.
+        #
+        # If a normal Python-visible write/flush/fsync/replace error occurs
+        # before the rename completes, remove the temp sidecar before
+        # re-raising so entity_registry.json.tmp does not linger as stale
+        # debris in the palace directory.
         payload = json.dumps(self._data, indent=2)
         tmp_path = self._path.with_name(self._path.name + ".tmp")
-        with open(tmp_path, "w", encoding="utf-8") as f:
-            f.write(payload)
-            f.flush()
-            os.fsync(f.fileno())
+
         try:
-            tmp_path.chmod(0o600)
-        except (OSError, NotImplementedError):
-            pass
-        os.replace(tmp_path, self._path)
-        # On ext4 (and similar) the rename's durability across power loss
+            with open(tmp_path, "w", encoding="utf-8") as f:
+                f.write(payload)
+                f.flush()
+                os.fsync(f.fileno())
+
+            try:
+                tmp_path.chmod(0o600)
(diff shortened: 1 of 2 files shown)


fatkobra:
Closing this as superseded by #1408.

This PR was intended to fix #1373 by cleaning up the `entity_registry.json.tmp` sidecar on failed save paths. Since #1408 has now been merged into `develop` and closes #1373 with the same cleanup behavior and regression coverage, keeping this PR open would duplicate already-merged work.

Thanks!

PR #1407, opened 2026-05-07
fix(repair): harden recovery flows and serialize writers

mjc:
## What does this PR do?

This PR hardens MemPalace repair and recovery for damaged Chroma palaces and reduces rebuild time on the common recovery path.

Rebased on 2026-05-15 onto `upstream/develop` at `bb31396`. Since the original PR, upstream has already landed overlapping repair hardening for rebuild progress/ETA and empty-metadata sanitization. This PR now focuses on the remaining repair-performance and writer-safety pieces that are still unique to this branch.

## Current delta after rebase

- Harden Chroma startup preflight/reconnect behavior so invalid HNSW metadata and stale HNSW state are handled consistently after palace replacement or reconnect.
- Make repair and recovery honor the configured primary collection name instead of assuming `mempalace_drawers`.
- Speed up legacy and from-SQLite rebuilds by reusing stored embeddings when complete, staging into a temporary collection, swapping by rename, and validating source/destination inventory before reporting success.
- Add a shared per-palace write boundary so `repair`, `mine`, repair cleanup, MCP write tools, and direct backend mutators do not race rebuild/swap paths. MemPalace-managed writes against the same palace no longer write the HNSW index concurrently.
- Fold in review followups around repair-lock release, rebuild/swap failure boundaries, migration no-op behavior, and tests for the above flows.

## User-visible outcomes

- repair/recovery succeeds against more corrupted-palace states without silently dropping rows
- custom collection names survive rebuilds correctly
- healthy palaces do not prompt for destructive migration confirmation when no rebuild is needed
- reconnect paths are more resilient to newly-invalid HNSW metadata
- live writers no longer race repair/rebuild work
- rebuilds can avoid expensive re-embedding when stored vectors are complete

## Verification

After the rebase:

```bash
.venv/bin/python -m ruff check mempalace/config.py mempalace/palace.py mempalace/repair.py mempalace/mcp_server.py tests/test_repair.py
.venv/bin/python -m pytest tests/test_repair.py tests/test_palace_locks.py tests/test_mcp_server.py -q
```

Result: `272 passed`; targeted ruff check clean.

## Broader test plan

- `uv run pytest tests/test_backends.py tests/test_searcher.py tests/test_repair.py tests/test_migrate.py tests/test_cli.py tests/test_mcp_server.py tests/test_miner.py -q`
- Optional manual spot checks:
  - run `mempalace repair` against a palace with a non-default configured collection name
  - run `mempalace migrate` against a healthy readable palace and confirm it exits successfully without prompting for destructive confirmation
  - verify MCP writes queue/refuse appropriately while repair is active

## Checklist

- [x] Targeted tests pass after rebase
- [x] Targeted linter check passes after rebase
- [x] No hardcoded paths

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -1231,14 +1231,9 @@ class ChromaBackend(BaseBackend):
         )
 
         if cached is None or inode_changed or mtime_changed or mtime_appeared:
-            # An inode swap means we are reopening a different physical DB
-            # (post-restore, fresh palace at the same path, etc.); drop the
-            # per-process gate so the quarantine pre-checks run again
-            # against the new disk state instead of trusting cached "we
-            # already cleaned this path" credit from the prior inode.
-            if inode_changed:
-                ChromaBackend._quarantined_paths.discard(palace_path)
-            ChromaBackend._prepare_palace_for_open(palace_path)
+            ChromaBackend._prepare_palace_for_open(
+                palace_path, reset_quarantine=inode_changed
+            )
             cached = chromadb.PersistentClient(path=palace_path)
             self._clients[palace_path] = cached
             # Re-stat after the client constructor runs: chromadb creates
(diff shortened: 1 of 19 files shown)


mjc:
I'm still experiencing occasional corruption but this cut it down a lot. unsure if its worth taking yet though.

PR #1415, opened 2026-05-08
fix(hooks): per-target PID guard with atomic claim (#1212, #1206)

### LINKED ISSUE #1206: Mempalace mine multiple processes hanging on claude folder
**What happened?**

```USER               PID  %CPU %MEM      VSZ    RSS   TT  STAT STARTED      TIME COMMAND
hunter2         22658  98.3  0.5 442773008 303856 s000  R     1:11PM 358:31.52 /Users/hunter2/.pyenv/versions/3.12.7/bin/python3 -m mempalace mine /Users/hunter2/.claude/projects/-Users-hunter2-repos-hunter2
hunter2         46719  58.5  0.3 442568096 178688 s000  R+    7:08PM   0:03.05 /Users/hunter2/.pyenv/versions/3.12.7/bin/python3 -m mempalace mine /Users/hunter2/.claude/projects/-Users-hunter2-repos-hunter2 --mode convos --wing sessions```

**What did you expect?**

I do not expect mempalace to be running for 6 hours on my machine. What is this thing doing? The folder size it's trying to mine is 350MB. 

**How to reproduce:**

1. Just using claude code; it seems the MCP is spawning these processes.
2.
3.

**Environment:**
- OS: MacOS Tahoe 26.4.1
- Python version: 3.12.7
- MemPal version: (check `python mempal.py --version` or git SHA) 3.3.3

### LINKED ISSUE #1212: Stop hook spawns concurrent `mempalace mine` processes that survive across sessions and bypass PID guard

igorls:
## Summary

The hook PID guard used a single global `~/.mempalace/hook_state/mine.pid` file, which failed two distinct ways:

1. **TOCTOU race**: `_mine_already_running()` read-then-spawn meant two near-simultaneous Stop hook fires both passed the existence/liveness check before either wrote — so both ended up spawning a mine.
2. **Unconditional overwrite**: `_spawn_mine()` overwrote the global PID file with the new child's PID. The first PID was lost, orphaning the first child. The user-visible result in #1212 was two concurrent `mempalace mine` processes running against the same source, driving HNSW inserts in parallel — exactly the corruption pattern the guard was meant to prevent. #1206 reported the same shape from the user perspective (two mines hung on a 350MB folder).

## Fix

Replace the global file with per-target slots under `~/.mempalace/hook_state/mine_pids/`, keyed by `sha256` of the mine sub-arguments (everything after `mine`). Two key invariants:

- **Atomic claim**: the slot is claimed via `O_CREAT | O_EXCL` so the claim is atomic — two simultaneous fires can never both pass.
- **Per-target keying**: different targets (e.g. project mine vs. transcript ingest, or two different `MEMPAL_DIR`s) get independent slots and run in parallel.

Stale slots (PID exists but is dead) are reclaimed transparently, so orphan miners that crashed without cleanup don't permanently block future hook fires.

The mine subprocess receives its slot path via `MEMPALACE_MINE_PID_FILE` env var; `miner._cleanup_mine_pid_file` reads that var on exit and removes the slot if it points at our PID, so orphaned slots from crashed mines don't accumulate.

Also: `_ingest_transcript` now routes through `_spawn_mine`, so the transcript ingest path participates in the same dedup. Repeated Stop fires for the same transcript no longer stack parallel mines (which was a separate gap — `_ingest_transcript` previously called `subprocess.Popen` directly with no PID guard at all).

## Changes

- `mempalace/hooks_cli.py`:
  - `_MINE_PID_FILE` (single file) → `_MINE_PID_DIR` (directory of per-target slots).
  - New `_pid_file_for_cmd(cmd)` derives the slot path from the command arguments.
  - `_mine_already_running()` (no-arg) → `_mine_already_running(cmd)` (per-target).
  - New `_claim_mine_slot(cmd)` does the atomic `O_CREAT | O_EXCL` claim with stale-PID reclaim.
  - `_spawn_mine` claims the slot before spawning, passes the slot path via `MEMPALACE_MINE_PID_FILE` env var to the child, and releases the slot on `OSError` so a failed spawn doesn't permanently block.
  - `_ingest_transcript` now routes through `_spawn_mine`.

- `mempalace/miner.py`:
  - `_cleanup_mine_pid_file` reads the path from `MEMPALACE_MINE_PID_FILE` instead of the now-removed global file.

- Tests:
  - `tests/test_hooks_cli.py`: existing PID-guard tests migrated to per-target slots; new tests cover the atomic claim, per-target independence, stale reclaim, OSError slot release, env var propagation, and transcript dedup.
  - `tests/test_miner.py`: PID cleanup tests now use the env var instead of the removed global.

## Test plan

- [x] `uv run pytest tests/test_hooks_cli.py tests/test_miner.py` — 121 passed (2 unrelated onnx flakes only when run alongside in a full pytest run; all PID-related tests green)
- [x] `uvx --from 'ruff>=0.4.0,<0.5' ruff check` + format check — clean

Closes #1212
Closes #1206

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -14,6 +14,7 @@ import subprocess
 import sys
 from datetime import datetime
 from pathlib import Path
+from typing import Optional
 
 SAVE_INTERVAL = 15
 STATE_DIR = Path.home() / ".mempalace" / "hook_state"
@@ -343,7 +344,7 @@ def _mine_already_running(cmd: list[str]) -> bool:
     return _pid_alive(int(recorded))
 
 
-def _claim_mine_slot(cmd: list[str]) -> Path | None:
+def _claim_mine_slot(cmd: list[str]) -> Optional[Path]:
     """Atomically reserve the per-target PID slot for ``cmd``.
 
     Returns the slot path on success, or ``None`` if the target is
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -6,6 +6,7 @@ Supported hooks: session-start, stop, precompact
 Supported harnesses: claude-code, codex (extensible to cursor, gemini, etc.)
 """
 
+import hashlib
 import json
 import os
 import re
@@ -256,7 +257,45 @@ def _get_mine_targets() -> list[tuple[str, str]]:
     return targets
 
 
-_MINE_PID_FILE = STATE_DIR / "mine.pid"
+# Per-target PID guard.
+#
+# Hook fires ingest mines in the background. If a previous fire's child is
+# still running for the *same* target (same source dir, mode, wing), the new
+# fire should skip rather than pile up — multiple concurrent mines against the
+# same source corrupt the HNSW index and exhaust disk via duplicate upserts
+# (#1212, #1206). But mines targeting *different* sources / modes must remain
+# independent so the user can have e.g. project-mining and transcript-ingest
+# running in parallel.
+#
+# The single ``mine.pid`` global file used previously failed both ways: the
+# guard was rebuilt every spawn (so two near-simultaneous fires both passed
+# the check before either wrote), and the file was unconditionally overwritten
+# (so the second spawn lost the first PID, orphaning it). The replacement is
+# a directory of per-target slots, claimed via ``O_CREAT | O_EXCL`` so the
+# claim is atomic and per-target.
+_MINE_PID_DIR = STATE_DIR / "mine_pids"
+
+# The per-process PID file path is communicated to the mine subprocess via
+# this env var so the child's cleanup hook (in miner.py) can remove its
+# own slot on exit without scanning the whole directory.
+_MINE_PID_FILE_ENV = "MEMPALACE_MINE_PID_FILE"
+
+
+def _pid_file_for_cmd(cmd: list[str]) -> Path:
+    """Return the per-target PID file path for a mine subcommand.
+
+    The key is derived from the mine arguments (everything after ``mine``)
+    so different (dir, mode, wing) combinations get independent slots.
+    Two fires with the same arguments collapse to the same slot — which is
+    exactly the dedup we want.
+    """
+    try:
+        idx = cmd.index("mine")
+        key = " ".join(cmd[idx:])
+    except ValueError:
+        key = " ".join(cmd)
+    digest = hashlib.sha256(key.encode("utf-8")).hexdigest()[:16]
+    return _MINE_PID_DIR / f"mine_{digest}.pid"
 
 
 def _pid_alive(pid: int) -> bool:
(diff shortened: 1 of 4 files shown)


lealbrunocalhau (commented):
lgtm

PR #1080, opened 2026-04-21
feat(mine): add --exclude flag + fix status metadata pagination

vnguyen-lexipol:
## Problem

The `mine` command currently has no way to skip files or directories without adding patterns to `.gitignore`. This is a blocker for monorepos and projects where assets like JSON config files, resource bundles, or generated directories need to stay in version control but should never be indexed into the palace.

Separately, `status` can fail on large palaces with SQLite variable limits when metadata is fetched in one large call.

## Solution

This PR now includes two scoped changes:

### 1) Exclude patterns for `mine`

Adds glob-based exclusion via two complementary mechanisms:

**a. `exclude:` key in `mempalace.yaml`** (per-project, committed with the repo):
```yaml
wing: my-app
exclude:
  - "**/*.json"
  - "resources/**"
  - "src/generated/**"
rooms:
  - name: src
    description: Source files
```

**b. `--exclude` CLI flag** (one-off or scripted):
```bash
mempalace mine ~/dev/projects/myapp --exclude "**/*.json" --exclude "resources/**"
# or comma-separated:
mempalace mine ~/dev/projects/myapp --exclude "**/*.json,resources/**"
```

### 2) Status pagination fix

`status()` now reads metadata in batches (`limit/offset`) instead of a single `limit=total` fetch. This avoids SQLite `too many SQL variables` errors on large palaces.

## Behaviour

- Patterns use `fnmatch` glob syntax, matched against paths **relative to the project root**
- CLI `--exclude` flags **append** to config-level `exclude:` patterns (both apply)
- **`include_ignored` still wins** — force-included files bypass exclude patterns, consistent with existing `.gitignore` override behaviour
- Excluded patterns are printed in the mine header under `Exclude:` for visibility
- Images and binaries are still skipped by the existing `READABLE_EXTENSIONS` whitelist
- `status` output counts are aggregated from paginated metadata batches and the header total uses `col.count()`

## Changes

- `mempalace/miner.py`
  - `scan_project()` + `mine()`: new `exclude_patterns` param and matching logic
  - `status()`: paginated metadata fetch to prevent SQLite variable-limit failures
- `mempalace/cli.py` — `--exclude` argparse option (append, comma-split, mirrors `--include-ignored`)
- `tests/test_miner.py` — exclude-pattern coverage and precedence tests

## Test plan

- [x] `python3 -m pytest tests/test_miner.py tests/test_cli.py`
- [x] Local `mempalace status` run completes successfully on a large local palace (no SQLite variable-limit crash)

PR #1421, opened 2026-05-08
feat(sync): add gitignore-aware drawer prune (#1252)

### LINKED ISSUE #1252: feat(mine): add --sync flag to prune drawers for files removed from git tracking
## Problem

When a file is added to `.gitignore` (or otherwise untracked from git), its drawers remain in the palace indefinitely. There is no way to detect and remove these orphaned drawers without manually running `mempalace purge --wing <x> --room <y>` or writing a custom ChromaDB script.

**Concrete example:** A project folder with ~10k files was mined into a room. The folder was later added to `.gitignore`. The room still holds 10,547 drawers with no supported way to bulk-remove them short of a custom Python script querying ChromaDB directly.

## Proposed solution

Add a `--sync` flag to `mempalace mine` (or a standalone `mempalace sync` command) that:

1. Reads `source_path` metadata from all drawers in the target wing/room
2. Checks each path against the current git-tracked file list (`git ls-files`) and the `.gitignore` rules
3. Deletes drawers whose source file is no longer tracked (gitignored, deleted, or moved)

```bash
mempalace mine --sync                     # prune orphans in current repo's wing
mempalace mine --sync --wing iis          # scope to one wing
mempalace sync --wing iis --room old-dir  # explicit sync command
```

## Why this should be default (or at least easy)

The mine workflow already tracks `filed_away` state per file to avoid re-mining. The inverse — removing stale entries — is the natural complement. Without it, palaces grow unboundedly and searches return results for files that no longer exist in the project.

A reasonable default would be: when `mempalace mine` is run in a git repo, automatically prune drawers whose `source_path` is now gitignored. This mirrors how tools like ripgrep and `git ls-files` naturally respect `.gitignore`.

## Related

- #1087 adds manual `mempalace purge --wing/--room` (workaround, but manual)
- #1080 adds `--exclude` flag for mine (prevents future mining, doesn't clean existing)

Neither addresses retrospective cleanup of gitignored files.

mvalentsev:
## Summary

Adds `mempalace sync` CLI command and `mempalace_sync` MCP tool that
prune drawers whose source files are now gitignored, deleted, or moved
out of the project. Reuses the existing `GitignoreMatcher` infrastructure
in `mempalace/miner.py` so the same gitignore rules that block ingest
also drive the corresponding cleanup.

Closes #1252.

## Why

Reporter sokol-matija put a 10k-file folder into `.gitignore` after
mining and ended up with 10547 stranded drawers. The miner respects
gitignore on the way in but never re-checks on subsequent runs, and
there is no way to remove the orphans short of
`mempalace purge --wing X --room Y` (manual, not detection-based).

## What this does

- New module `mempalace/sync.py` with `sync_palace(palace_path,
  project_dirs, wing, dry_run, batch_size, wal_log)` returning a typed
  `SyncReport`.
- New CLI: `mempalace sync [dir] [--wing W] [--root R] [--apply]`.
  Default is dry-run preview; `--apply` opts into deletion. Output
  shows DRY RUN or APPLY mode header so callers always see which
  branch ran.
- New MCP tool `mempalace_sync(project_dir, wing, apply)`. Returns
  `{success: true, ...}` on success and `{success: false, error: ...}`
  on every error path (no palace, lock contention, scope guard,
  unexpected backend failure).
- Five classification buckets: `kept`, `gitignored`, `missing`,
  `no_source`, `out_of_scope`. Drawers without `source_file` (convo /
  explicit-add memories), convo-miner registry sentinels (`_reg_*` IDs
  or `room=_registry` or `ingest_mode=registry`), and drawers outside
  the requested project root are always preserved.
- `mine_palace_lock` is held for the entire call (classify + apply) so
  a concurrent miner cannot shift offsets under the scan; sync raises
  `MineAlreadyRunning` (CLI exit 1, MCP structured error) if another
  mine is in progress.
- Apply refuses to run when neither `wing` nor `project_dirs` is set,
  so a caller cannot accidentally prune every wing in a multi-project
  palace via auto-detected roots. Empty `project_dirs=[]` also rejected.
- Auto-detection picks the deepest `.git` / `.gitignore` ancestor
  (not the shallowest), with deterministic ordering when paths share
  string length.
- Closet purge runs alongside drawer deletion via a single
  `where={"source_file": {"$in": [...]}}` batch get + one bulk
  `delete(ids=...)` call, with a logged warning if the closets
  collection is unavailable.
- WAL logging via the same `_wal_log` plumbing as `delete_drawer` from
  both CLI and MCP paths; result payload carries `removed_count`,
  params allow-listed to `first_id` only (no source/content leaks).

## What this does NOT do

- Not a generic prune by ID or wing/room. `mempalace purge` (#1087)
  covers the manual destructive case; `mempalace prune` (#522) covers
  existence/mtime/orphans cleanup orthogonal to gitignore.
- Does not auto-run during `mine`. Sync is destructive; conflating it
  with mine surprises users who run `mine` from cron.
- Does not rematch moved files by content hash. A moved file's old
  drawer disappears under `missing`; the new location is picked up on
  the next `mine`.
- Does not modify the knowledge graph. Entities can outlive any single
  source.

## Surface

```
mempalace sync /path/to/repo                  # dry-run, auto-detect roots
mempalace sync /path/to/repo --apply          # commit deletions
mempalace sync --wing project_x --apply       # scope to one wing
mempalace sync --root /a --root /b            # multiple project roots
```

```
mempalace_sync(project_dir="/repo")              # dry-run via MCP
mempalace_sync(project_dir="/repo", apply=true)  # commit via MCP
```

## Test plan

- 38 tests in `tests/test_sync.py` across `TestSyncPalace` (28),
  `TestSyncMcpTool` (6), and `TestSyncCli` (4) covering: all 5
  classification buckets; dry-run vs apply mutation safety; wing
  scope; negated and nested gitignore rules; closet purge with
  collection-unavailable warning; empty palace; WAL emission with
  allow-listed audit shape; registry-sentinel preservation across all
  3 detection paths; auto-detect deepest-root selection (white-box);
  apply-without-scope guard (both runtime ValueError and CLI exit 2);
  empty `project_dirs=[]` guard; `MineAlreadyRunning` propagation
  under fcntl contention from CLI and MCP; idempotent repeated apply;
  relative `source_file` defensive routing; overlapping `project_dirs`
  longest-prefix selection; `batch_size` boundary at 2 / 5 chunks;
  SyncReport key-shape stability; cache invalidation on exception
  with explicit explode-mock counter; structured success/error
  contract on all MCP entry points; CLI `_wal_log` round-trip; CLI
  argparse round-trip for both dry-run and apply; symmetric
  `.resolve()` of `source_file` under tmpfs symlinks (closes
  macOS/Windows CI gap); per-file classification cache verified via
  `_classify_drawer` counter monkeypatch; closet purge collapsed to
  one `$in`-batched `get`/`delete` verified via call-counting
  wrapper; registry-row check runs before cache lookup so a sentinel
  cannot inherit a non-registry drawer's cached verdict;
  `_normalize_project_dirs` deterministic tie-break on equal-length
  paths.
- Manual end-to-end on a synthetic repo (30 files in `build/`
  gitignored + 1 `*.log` gitignored + 1 deleted file + 1 kept file +
  1 `_reg_` sentinel for a gitignored transcript): dry-run reports 31
  gitignored + 1 missing, `--apply` removes exactly 32 drawers leaving
  the keep file and the registry sentinel intact.
- Full pytest suite passes locally; all 6 CI jobs (lint,
  test-linux 3.9/3.11/3.13, test-macos, test-windows) green at
  HEAD `1822756`.

## Refs

- Closes #1252
- Adjacent: #522 (orthogonal `pruner` for existence/mtime/orphans),
  #1087 (`mempalace purge` manual), #1080 (`--exclude` mining-time)

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -579,6 +579,84 @@ def cmd_sweep(args):
         sys.exit(1)
 
 
+def cmd_sync(args):
+    """Prune drawers whose source files are gitignored, deleted, or moved (#1252)."""
+    from .mcp_server import _wal_log
+    from .palace import MineAlreadyRunning
+    from .sync import sync_palace
+
+    palace_path = os.path.expanduser(args.palace) if args.palace else MempalaceConfig().palace_path
+
+    if not os.path.isdir(palace_path):
+        print(f"\n  No palace found at {palace_path}")
+        return
+
+    project_dirs = []
+    if args.dir:
+        project_dirs.append(os.path.expanduser(args.dir))
+    project_dirs.extend(os.path.expanduser(r) for r in args.root)
+    project_dirs = project_dirs or None
+
+    print(f"\n{'=' * 55}")
+    print("  MemPalace Sync — Gitignore-aware drawer prune")
+    print(f"{'=' * 55}")
+    print(f"  Palace:   {palace_path}")
+    if args.wing:
+        print(f"  Wing:     {args.wing}")
+    if project_dirs:
+        for p in project_dirs:
+            print(f"  Project:  {p}")
+    if args.dry_run:
+        print("  Mode:     DRY RUN (no deletions)")
+    else:
+        print("  Mode:     APPLY (deleting drawers)")
+    print(f"{'-' * 55}\n")
+
+    try:
+        report = sync_palace(
+            palace_path=palace_path,
+            project_dirs=project_dirs,
+            wing=args.wing,
+            dry_run=args.dry_run,
+            wal_log=_wal_log,
+        )
+    except MineAlreadyRunning as exc:
+        print(f"mempalace: {exc}", file=sys.stderr)
+        sys.exit(1)
+    except ValueError as exc:
+        print(f"mempalace: {exc}", file=sys.stderr)
+        sys.exit(2)
+    except Exception as exc:
+        print(f"mempalace: sync failed: {exc}", file=sys.stderr)
+        sys.exit(1)
+
+    removed_suffix = "(would remove)" if args.dry_run else "(removed)"
+    print(f"  Scanned:        {report['scanned']}")
+    print(f"  Kept:           {report['kept']}")
+    print(f"  Gitignored:     {report['gitignored']}  {removed_suffix}")
+    print(f"  Missing:        {report['missing']}  {removed_suffix}")
+    print(f"  No source:      {report['no_source']}  (kept)")
+    print(f"  Out of scope:   {report['out_of_scope']}  (kept)")
+
+    by_source = report.get("by_source") or {}
+    if by_source:
+        top = sorted(by_source.items(), key=lambda kv: -kv[1])[:5]
+        label = "Top sources to remove" if args.dry_run else "Top sources removed"
+        print(f"\n  {label}:")
+        for src, n in top:
+            print(f"    {src}  ({n})")
+
+    if args.dry_run:
+        if report["gitignored"] + report["missing"] > 0:
+            print("\n  Re-run with --apply to commit these deletions.")
+    else:
+        print(
+            f"\n  Removed {report['removed_drawers']} drawers, {report['removed_closets']} closets."
+        )
+
+    print(f"\n{'=' * 55}\n")
+
+
 def cmd_search(args):
     from .searcher import search, SearchError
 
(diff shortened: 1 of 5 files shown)


mvalentsev:
All 5 picked up. One tweak on the cache: registry sentinel rows (`_reg_*`, `room=_registry`, `ingest_mode=registry`) route to "kept" before the cache lookup, since a sentinel that carries source_file metadata could otherwise read a stale "gitignored" verdict cached by an earlier non-registry chunk with the same source_file.

Also a separate fix for the macOS/Windows CI failures: source_file was being compared raw against `.resolve()`'d project_roots, which under tmpfs symlinks (`/var/folders` -> `/private/var`) buckets every drawer as `out_of_scope`. `_classify_drawer` now resolves symmetrically.

Regression tests added: symlinked-tmp resolution, classification cache (counter on `_classify_drawer`), closet batch (single delete via `$in`), plus registry+cache interaction and `_normalize_project_dirs` tie-break unit.

PR #1235, opened 2026-04-27
fix: handle lone surrogates in MCP write tools

YuanYiZheXue:
## Problem

MCP write tools (mempalace_add_drawer, mempalace_diary_write) fail with UnicodeEncodeError when processing strings containing lone surrogates injected by MCP clients like WorkBuddy.

## Solution

Add _clean() function that removes lone surrogates using surrogatepass/surrogateignore encoding before writing to ChromaDB.

## Changes

- tool_add_drawer: Clean content, wing, room, added_by before ChromaDB upsert
- tool_diary_write: Clean entry, topic, agent_name, wing before ChromaDB add
- SHA256 hash: Use surrogatepass encoding to avoid errors

## Testing

```bash
mcp__mempalace__mempalace_add_drawer --wing test --room unicode --content "????"
mcp__mempalace__mempalace_diary_write --agent_name test --entry "SESSION:2026-04-27"
```

See docs/fix-lone-surrogate.md for full details.

PR #1422, opened 2026-05-09
fix(mcp): sanitize lone surrogates before ChromaDB ops (#1235)

YuanYiZheXue:


igorls (changes requested):
Thanks for the fix — the implementation is correct and addresses a real crash.

**Requested changes before merge:**

1. **Add test coverage.** A helper touching 5 production tool paths needs at least a unit test for `_clean()` (strings with/without surrogates, valid surrogate pairs) and one integration test exercising a tool path (e.g. `tool_add_drawer`) with a surrogate-laden payload to confirm ChromaDB no longer crashes.

2. **Trim the docstring.** The 14-line explanation belongs in the commit message / issue #1235. Per project conventions, keep the docstring to 2–3 lines max.

3. **PR title** — already updated to a conventional-commit style.

YuanYiZheXue:
## Tests Added

Tests have been added in tests/test_clean_lone_surrogates.py:

**Unit tests (11)** for _clean(): lone surrogates, real emoji preservation, empty strings, SHA-256 compatibility

**Integration tests (6)** covering all 5 tool paths:
- tool_add_drawer (content + metadata fields)
- tool_check_duplicate
- tool_search
- tool_update_drawer
- tool_diary_write

Commit: 726495d — all 17 tests pass.

Note: A test-environment fix is included. ONNXMiniLM_L6_V2.DOWNLOAD_PATH is patched at module import time using conftest._original_env to avoid 79 MB ONNX model re-downloads on every test run.

igorls:
Pushed a follow-up commit (c6bdd8b) that addresses the architectural feedback from the review:

- `_clean()` is now `strip_lone_surrogates()` in `mempalace/config.py`, called from inside `sanitize_content`, `sanitize_kg_value`, and `sanitize_query` so every MCP tool inherits the fix automatically and future tools don't need to remember a separate helper.
- Implementation is a single regex pass (`re.sub(r'[\ud800-\udfff]', '�', text)`), so one bad code point becomes one U+FFFD instead of three.
- Moved the ONNX model-cache patch out of the test module and into `conftest.py` as session-scoped setup.
- Updated test expectations to match the 1×U+FFFD substitution.

Full suite passes locally (1702/1702). Thanks for the fix — planning to squash-merge once CI is green.

igorls (approved):

PR #1424, opened 2026-05-09
fix(hooks): correct wing extraction for hyphenated project names (#1410)

### LINKED ISSUE #1410: _wing_from_transcript_path truncates hyphenated project names — wing collisions for *-code, *-native, etc.
**Version: 3.3.4 (Python 3.14, macOS)**         

### Summary                                                                                                                                                                                                  
   
 `mempalace.hooks_cli._wing_from_transcript_path` derives a project wing by taking only the last dash-separated token of Claude Code's encoded project folder. Because Claude Code encodes the source directory by replacing / with -, any project whose folder name itself contains a - is silently truncated, producing the wrong wing.
                                                                                                                                                                                                           
### Reproduction                                              

`from mempalace.hooks_cli import _wing_from_transcript_path
`
```
_wing_from_transcript_path("/Users/me/.claude/projects/-Users-me-git-code-project/abc.jsonl")                                                                                                                                                                                                        
  # returns: 'wing_project'                                 
  # expected: 'wing_code_project' (or similar — anything that captures the project name)
```
                                                                                                                                                                                                           
### Source 
`mempalace/hooks_cli.py:518`

```
  match = re.search(r"/\.claude/projects/-([^/]+)", normalized)
  if match:                                                                                                                                                                                                
      encoded = match.group(1)
      project = encoded.rsplit("-", 1)[-1]   # <-- only keeps last token                                                                                                                                   
      if project:                                                                                                                                                                                          
          return f"wing_{project.lower().replace(' ', '_')}"
```
                                                                                                                                                                                                           
### Impact                                                    

Any project with a - in its folder name gets a silently corrupted wing:                                                                                                                                  
   
  ```
│ Project folder  │ Current result │   Likely intended    │
  │ claude-code     │ wing_code      │ wing_claude_code     │
  │ react-native    │ wing_native    │ wing_react_native    │
  │ customer-portal │ wing_portal    │ wing_customer_portal │
  │ next-app        │ wing_app       │ wing_next_app        │
```
                                                            
Two real consequences:                                                                                                                                                                                   
   
1. Project-scoped queries miss diary entries. mempalace wake-up --wing <project> won't find checkpoints written by the Stop hook, because they're in the truncated wing.                                 
2. Multi-project collision. Any two projects ending in the same final token (customer-portal and admin-portal → both wing_portal) get their diary entries merged into one wing, defeating the wing isolation model.                                                                                                                                                                                         
                                                            
tool_diary_read is unaffected because it deliberately doesn't filter by wing when none is supplied (per its own docstring), but everything else that scopes by wing does break.                          
                                                            
### Suggested fix                                                                                                                                                                                            
                                                            
Strip OS user-home and common parent-dir prefixes from the encoded path, then keep the rest as the project. Tested locally:                                                                              
   
```
  match = re.search(r"/\.claude/projects/-([^/]+)", normalized)                                                                                                                                            
  if match:                                                 
      encoded = match.group(1)
      m = re.match(r"(?:Users|home)-[^-]+-(.+)", encoded)
      if m:                                                                                                                                                                                                
          encoded = m.group(1)
      for pfx in ("git-", "dev-", "projects-", "Projects-",
                  "src-", "code-", "work-", "Documents-"):
          if encoded.startswith(pfx):
              encoded = encoded[len(pfx):]
              break
      project = encoded.lower().replace(" ", "_").replace("-", "_")
      if project:
          return f"wing_{project}"
```

**Verified across:**

  ```
│               Input               │               Output               │
  │ -home-alice-projects-react-native │ wing_react_native                  │
  │ -Users-bob-myproject              │ wing_myproject                     │
  │ -Users-c-dev-claude-code          │ wing_claude_code                   │
  │ /some/other/path.jsonl            │ wing_sessions (fallback unchanged) │
```

A more robust long-term fix would be to read the actual cwd from the JSONL transcript (Claude Code records it per-message) instead of trying to round-trip the encoded folder name — but the prefix-strip heuristic above covers the common cases without that extra parsing.

Happy to send a PR if it's helpful.

arnoldwender:
## What and Why

`_wing_from_transcript_path` derived the wing name by taking only the last dash-separated token of Claude Code's encoded project folder. Since Claude Code encodes the source directory by replacing `/` with `-`, any project whose folder name itself contained a dash was silently truncated:

| Encoded folder | Old result | Correct result |
|---|---|---|
| `-Users-me-claude-code` | `wing_code` | `wing_claude_code` |
| `-Users-me-react-native` | `wing_native` | `wing_react_native` |
| `-Users-me-customer-portal` | `wing_portal` | `wing_customer_portal` |
| `-Users-me-admin-portal` | `wing_portal` | `wing_admin_portal` |

Two real consequences flagged in #1410:
1. Project-scoped queries (`wake-up --wing <project>`) miss diary entries that the Stop hook filed under the truncated wing.
2. Multi-project collision — any two projects ending in the same final token (e.g. `customer-portal` and `admin-portal` → both `wing_portal`) get merged into one wing, defeating wing isolation.

## Reproduction

```python
from mempalace.hooks_cli import _wing_from_transcript_path
_wing_from_transcript_path("/Users/me/.claude/projects/-Users-me-claude-code/abc.jsonl")
# old:      'wing_code'
# expected: 'wing_claude_code'
```

## Root Cause

`hooks_cli.py:694` (pre-fix):
```python
project = encoded.rsplit("-", 1)[-1]  # only keeps the last dash-separated token
```

## Fix

Two-tier strategy:

1. **Primary — `cwd` from JSONL.** Claude Code records the absolute working directory on most message types (tool_use, tool_result, user/assistant turns). Reading the first JSONL line that has a non-empty `cwd` gives the canonical project path; the leaf segment becomes the wing. Bounded scan (200 lines) keeps the lookup well inside the hook's 500ms budget. This is the long-term-robust fix the reporter described.

2. **Fallback — improved encoded-folder heuristic.** When cwd isn't recorded (older Claude Code, queue-only transcripts), decode the `.claude/projects/-...` folder: strip the platform user-home prefix (`Users-<user>-` / `home-<user>-`) and one common parent-dir token (`git-`, `dev-`, `projects-`, `Projects-`, `src-`, `code-`, `work-`, `Documents-`), then convert remaining dashes to underscores. May include extra parent-dir noise in the wing name when the project lives several levels deep, but **never silently truncates** a hyphenated project name. Approach borrowed from the heuristic @ReporterUnknown sketched in #1410 — thanks for the well-specified report.

## Change Summary

- `mempalace/hooks_cli.py` — replaced the `rsplit("-", 1)[-1]` truncation with cwd-primary + improved fallback. Added `_wing_from_jsonl_cwd()` helper. No new imports (`json`, `re`, `Path`, `Optional` were already present).
- `tests/test_hooks_cli.py` — updated 2 existing tests whose assertions encoded the old truncation (they happened to give the right answer when project = leaf dir of a deep path), added 11 new tests:
  - 4 regression tests for hyphenated project names + collision avoidance
  - 7 tests for the cwd-from-JSONL primary path (happy path, missing file, malformed JSON, non-string cwd, no-cwd-in-file fallback, hyphenated cwd, queue-operation lines without cwd)

## Test Plan

- [x] `uv run ruff check .` — clean
- [x] `uv run ruff format --check .` — clean
- [x] `uv run pytest tests/test_hooks_cli.py -k wing_from_transcript_path -v` — 19 passed (6 original + 13 new)
- [x] `uv run pytest tests/test_hooks_cli.py` — 100 passed, 1 skipped
- [x] `uv run pytest tests/ --ignore=tests/benchmarks` — **1698 passed, 1 skipped**

## Notes for review

- The two existing tests I updated (`non_projects_layout` → `wing_mempalace_mempalace`, `nested_deep` → `wing_clients_acme_frontend`) used to pass only because the old "last token" heuristic happened to coincide with the actual project name on those specific deep paths. Their inputs were ambiguous on encoding alone (no way to tell `dev-MemPalace-mempalace` from `dev-MemPalace-mempalace` as one project name). The new fallback preserves all post-`Users-<user>-`/`home-<user>-` segments, which is collision-safe; the primary cwd-from-JSONL path produces clean `wing_mempalace` / `wing_frontend` for these cases when the JSONL records cwd.
- The 200-line bounded scan in `_wing_from_jsonl_cwd` matches Claude Code's pattern of recording cwd on most-but-not-all message types (queue-operation lines lack it). Hook latency overhead measured at <1ms on a 1MB transcript locally; falls within the documented 500ms budget.

Closes #1410.

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -670,36 +670,124 @@ def _parse_harness_input(data: dict, harness: str) -> dict:
     }
 
 
+# Common parent-dir tokens stripped from the encoded folder when no
+# explicit ``-Projects-`` segment is present. Order matters: only the
+# first match strips. These cover the bulk of Unix layouts; cwd-from-JSONL
+# (the primary path) handles the long tail correctly without heuristics.
+_ENCODED_PARENT_PREFIXES = (
+    "git-",
+    "dev-",
+    "projects-",
+    "Projects-",
+    "src-",
+    "code-",
+    "work-",
+    "Documents-",
+)
+
+
+def _wing_from_jsonl_cwd(transcript_path: str) -> Optional[str]:
+    """Read ``cwd`` from the first JSONL line that records it.
+
+    Claude Code stores the absolute working directory on most message
+    types (tool_use, tool_result, user/assistant turns), but not all
+    (e.g. queue-operation lines lack it). Scan up to 200 lines to find
+    the first record that includes a non-empty cwd, then derive the
+    wing from its leaf path segment. Returns ``None`` if the file is
+    unreadable, empty, or contains no cwd.
+    """
+    try:
+        path = Path(transcript_path).expanduser()
+        if not path.is_file():
+            return None
+        with path.open("r", encoding="utf-8", errors="replace") as f:
+            for i, line in enumerate(f):
+                if i >= 200:
+                    break
+                line = line.strip()
+                if not line or '"cwd"' not in line:
+                    continue
+                try:
+                    data = json.loads(line)
+                except (json.JSONDecodeError, ValueError):
+                    continue
+                cwd = data.get("cwd")
+                if not cwd or not isinstance(cwd, str):
+                    continue
+                cwd_norm = cwd.replace("\\", "/").rstrip("/")
+                if not cwd_norm:
+                    continue
+                project = cwd_norm.rsplit("/", 1)[-1]
+                if project:
+                    slug = project.lower().replace(" ", "_").replace("-", "_")
+                    return f"wing_{slug}"
+    except OSError:
+        pass
+    return None
+
+
 def _wing_from_transcript_path(transcript_path: str) -> str:
     """Derive a project wing name from a Claude Code transcript path.
 
-    Claude Code encodes the project's source directory by replacing path
-    separators with dashes, producing folders like:
-        ~/.claude/projects/-home-<user>-Projects-<project>/session.jsonl
-        ~/.claude/projects/-home-<user>-dev-<parent>-<project>/session.jsonl
-        ~/.claude/projects/-Users-<user>-<folder>-<project>/session.jsonl
+    Strategy (in priority order):
 
-    The project directory name is the final dash-separated token of the
-    encoded folder. Returns ``wing_<project>`` (lowercased, spaces → ``_``).
-    Falls back to ``wing_sessions`` if the path does not match a Claude Code
-    project-folder layout.
+    1. PRIMARY — Read ``cwd`` from the JSONL transcript. Claude Code records
+       the absolute working directory on most message types, so the project
+       name is whatever the leaf path segment of cwd is. This is the
+       canonical answer when present.
+
+    2. FALLBACK — Decode the encoded folder under ``.claude/projects/``.
+       Claude Code flattens path separators to dashes (``/Users/me/code/foo``
+       → ``-Users-me-code-foo``), so the original directory boundaries are
+       lost. We strip the platform user-home prefix (``Users-<user>-`` or
+       ``home-<user>-``) and one common parent-dir token (``git-``, ``dev-``,
+       ``projects-``, etc.), then convert the remaining dashes to
+       underscores. Unlike the previous "last token only" heuristic, this
+       never silently truncates a hyphenated project folder name like
+       ``claude-code``, ``react-native``, or ``customer-portal``.
+
+    3. LEGACY — Match an explicit ``-Projects-<name>`` segment for
+       transcripts not under the standard Claude Code projects dir.
+
+    4. DEFAULT — ``wing_sessions``.
+
+    Closes #1410.
     """
+    # 1. Primary — cwd from JSONL is the canonical source of truth
+    cwd_wing = _wing_from_jsonl_cwd(transcript_path)
+    if cwd_wing:
+        return cwd_wing
+
     # Normalize path separators for cross-platform (Windows backslashes)
     normalized = transcript_path.replace("\\", "/")
-    # Primary: pull the encoded project folder out of ``.claude/projects/``
-    # and take its last dash-separated token.
+
+    # 2. Fallback — encoded project folder under .claude/projects/
     match = re.search(r"/\.claude/projects/-([^/]+)", normalized)
     if match:
         encoded = match.group(1)
-        project = encoded.rsplit("-", 1)[-1]
+        # Strip platform user-home prefix so the wing isn't dominated by
+        # /Users/<user>/ or /home/<user>/.
+        m = re.match(r"(?:Users|home)-[^-]+-(.+)", encoded)
+        if m:
+            encoded = m.group(1)
+        # Strip one common parent-dir token if present, keeping the rest as
+        # the project path. Hyphens become underscores to preserve
+        # uniqueness for hyphenated project folder names.
+        for prefix in _ENCODED_PARENT_PREFIXES:
+            if encoded.startswith(prefix):
+                encoded = encoded[len(prefix) :]
+                break
+        project = encoded.lower().replace(" ", "_").replace("-", "_")
         if project:
-            return f"wing_{project.lower().replace(' ', '_')}"
-    # Legacy fallback: explicit ``-Projects-<name>`` segment, useful for
-    # transcripts not under the standard Claude Code projects dir.
+            return f"wing_{project}"
+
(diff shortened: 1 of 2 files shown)


arnoldwender:

jphein:
+1 from operator-experience. We hit the same problem on the jphein/mempalace fork — Claude Code transcript paths for projects with dashes in the directory name (e.g. `-home-jp-Projects-realm-watch` → wing `watch` when it should be `realm_watch`) were silently breaking project wing assignment.

Our fork landed a narrower fix in commit `d76134d` that's encoded-folder-decode only — preserves dashes when an explicit `-Projects-<name>` segment is present, falls back to last-token-after-dash otherwise. It works but it's a heuristic on top of an encoded representation.

The JSONL-cwd-primary approach in this PR is structurally better: `cwd` is the canonical signal Claude Code records on most message types, so it's what the project actually IS rather than what its encoded folder happens to look like. Once this merges and we sync, we'll retire our fork-side version.

Thanks for the more correct fix.

jphein:
Operator follow-up — we hit this exact bug class in our [techempower-org/palace-daemon](https://github.com/techempower-org/palace-daemon) fork's hook (`_project_wing`) and landed a more general fix yesterday. Sharing the shape in case it's useful for the upstream direction.

## The general fix

Rather than fixing the last-token-after-dash heuristic for hyphenated project names, replace path-encoding decoding with a deliberate inverse:

```python
def _wing_from_claude_transcript_path(transcript_path: str) -> Optional[str]:
    """Claude Code encodes /Users/me/project-name → -Users-me-project-name
    (path-separators → dashes, leading slash → leading dash). The inverse
    is: strip leading dash, replace remaining dashes with slashes, take
    final path segment.
    """
    parts = Path(transcript_path).parts
    for i, part in enumerate(parts):
        if part == "projects" and i + 1 < len(parts):
            encoded = parts[i + 1]  # "-Users-me-customer-portal"
            decoded = encoded.lstrip("-").replace("-", "/")  # "Users/me/customer/portal"
            return decoded.rsplit("/", 1)[-1]  # "portal"  ← still wrong!
```

Wait — that still has the original bug. The actual fix needs to know that Claude Code's encoding is **lossy**: the dash that originally separated the project basename from its parent directory looks identical to a dash inside the project name. So `customer-portal` and `customer/portal` encode to the same form.

Our solution: assume the **basename** is the wing (rather than trying to recover the original directory structure). Apply slug normalization to whatever's left:

```python
def normalize_wing_slug(s: str) -> str:
    if not s:
        return "unknown"
    s = s.lower()
    if s.startswith("wing_"):
        s = s[5:]
    return re.sub(r"[^a-z0-9_]", "_", s) or "unknown"

# transcript path "-Users-me-customer-portal" → decode → "Users/me/customer/portal"
# → basename "portal" → BUT we resolved the multi-project collision differently:
# fall back to cwd basename when the transcript path doesn't uniquely identify a project
```

The collision JP raised (`customer-portal` vs `admin-portal` both → `wing_portal`) hits anyone running multiple projects per machine. Our fix sidesteps this by **preferring `mempalace.yaml` over transcript-derived wing** when present. The `mempalace.yaml` lives in the project root and carries the canonical wing slug — see config.py's existing wing lookup. The transcript-path derivation is only the fallback for projects without `mempalace.yaml`.

Operator-level: in our deployment we run a `~/Projects/lexicon.realm.watch/catalog/projects.yaml` (with 89 distinct project entries) and resolve the encoded-path wing against that catalog as a secondary fallback. Hairy edge case fully resolved.

## Suggestion

If you want a fully general fix here, it's worth either:
1. Documenting Claude Code's encoding as lossy and requiring `mempalace.yaml` for unambiguous wing assignment (current PR's `conversations` fallback is fine but elides that there's no fully-correct path-only solution)
2. Reading project metadata from elsewhere — Claude Code's session file has the cwd recorded as `data.cwd` per the [hook spec](https://docs.anthropic.com/claude-code/hooks#stop). Using cwd directly avoids the encoding round-trip entirely.

We chose (2) as our primary path with `mempalace.yaml` as override. Happy to share more if useful.

arnoldwender:
Thanks for the operator-level follow-up — the lossy-encoding framing is more precise than how I scoped the original bug. Claude Code's path encoding can't be uniquely inverted once a project basename contains dashes; that's exactly why the merged fix reads `data.cwd` from the JSONL as the primary path, with the prefix-decoded encoded folder only as fallback when transcripts predate that field.

The mempalace.yaml-first idea is interesting — that'd give an authoritative wing regardless of how the transcript was recorded. Worth a follow-up issue if @igorls wants to formalize that as the canonical resolution order. Happy to file it if useful.

PR #1425, opened 2026-05-09
feat(search): opt-in recency-aware ranking via exponential decay

zhapostolski:
## Summary

Add an optional third axis to the existing hybrid (vector + BM25) ranker in ``_hybrid_rank``: an exponential-decay recency factor on each drawer's ``filed_at`` timestamp.

This addresses the common *"what was the latest decision on X?"* failure mode where a palace accumulates years of session transcripts and an old high-similarity drawer outranks a recent, more relevant one. Pure cosine + BM25 has no notion of time, so a 12-month-old infrastructure note can mask the answer recorded in last week's session.

## Behavior

**Default (``MEMPALACE_RECENCY_WEIGHT=0.0``) is a strict no-op** — the score formula collapses to the existing convex combination:

```
score = 0.6 * vec_sim + 0.4 * bm25_norm
```

Setting ``MEMPALACE_RECENCY_WEIGHT`` to a value in ``(0, 1]`` blends in ``recency_factor``:

```
score = (1 - w) * (0.6 * vec_sim + 0.4 * bm25_norm)
      + w       *  recency_factor

recency_factor = 2 ** (-age_days / half_life_days)
age_days       = max(0, now - filed_at)
```

## Configuration

| Env var | Default | Meaning |
|---|---|---|
| ``MEMPALACE_RECENCY_WEIGHT`` | ``0.0`` | Blend factor; 0 = old behavior, 1 = pure recency |
| ``MEMPALACE_RECENCY_HALF_LIFE_DAYS`` | ``30`` | Time for ``recency_factor`` to halve |

Programmatic callers can also override per-call:

```python
_hybrid_rank(results, query, recency_weight=0.3, half_life_days=14)
```

Negative weights clamp to 0; values >1 clamp to 1 — a callsite that plumbs through user input can never invert the score direction.

## Edge cases handled

* Missing or unparseable ``filed_at`` → ``recency_factor = 0.0`` (a drawer with no timestamp can never beat a recent dated drawer on the recency axis alone)
* Future-dated entries (clock skew) → ``recency_factor = 1.0`` (treated as "just now", not negative)
* Trailing ``Z`` (Zulu/UTC) timestamps → parsed correctly via the standard ``datetime.fromisoformat`` ``+00:00`` substitution
* Timezone-naive timestamps → assumed UTC
* Empty results list → unchanged no-op

When recency blending is active, each result dict gains a ``"recency_factor"`` key alongside the existing ``"bm25_score"``, so introspection / debugging still works (CLI ``mempalace search`` can display it; MCP responses pass through arbitrary keys to the agent).

## Why opt-in by default

* Existing palaces see literally no behavior change without explicit env var
* Long-tail-stable palaces (reference docs, code patterns, identity files) usually want ``recency=0``; aging them out would be wrong
* Users with mostly session-transcript content benefit most
* Half-life of 30 days matches the "what's current?" intent of typical agent queries without aging out reference material too aggressively

## Test plan

Smoke-tested in a sanity script:

```
=== _recency_factor ===
  now           -> 1.000
  30d ago       -> 0.500   ← exact half-life
  60d ago       -> 0.250
  365d ago      -> 0.000
  unknown       -> 0.000
  None          -> 0.000
  malformed     -> 0.000
  trailing-Z    -> 1.000
  future        -> 1.000
  90d half=10   -> 0.002

=== _hybrid_rank ===
  recency_weight=0   -> top: 'older but exact match' (backward-compat ✓)
  recency_weight=.5  -> top: 'newer but weaker semantic match' (recency wins ✓)
  recency_weight=99  -> clamped to 1.0 (pure recency)
  recency_weight=-5  -> clamped to 0.0 (no effect)
  empty list         -> no-op ✓
```

## Followups (not in this PR)

These were considered but kept out for focused review; happy to open separate PRs if the maintainer is interested:

* **Embedder upgrade as opt-in** — current ``all-MiniLM-L6-v2`` is the 2019 baseline; ``BGE-base-en-v1.5`` typically gets +30-40% recall@10. Needs a re-embed migration tool because vectors are dim-incompatible (384 → 768).
* **Cross-encoder reranker on the live MCP path** — benchmark code uses LLM-as-reranker (``llm_rerank_locomo``) which is too slow / costly for production. A real cross-encoder (``bge-reranker-base``) would add ~50-200 ms per query but materially improve top-K precision; would introduce a new optional dep.

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -60,6 +60,7 @@ def _recency_factor(filed_at_str, half_life_days: float = _RECENCY_HALF_LIFE_DAY
         return 1.0
     return 2.0 ** (-age_days / half_life_days)
 
+
 # Closet pointer line format: "topic|entities|→drawer_id_a,drawer_id_b"
 # Multiple lines may join with newlines inside one closet document.
 _CLOSET_DRAWER_REF_RE = re.compile(r"→([\w,]+)")
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -14,9 +14,51 @@ import math
 import os
 import re
 import sqlite3
+from datetime import datetime, timezone
 from pathlib import Path
 
 from .palace import get_closets_collection, get_collection
+
+
+# --- Recency-aware ranking config -------------------------------------------
+# Backwards compatible: with weight=0.0 (default) ``_hybrid_rank`` behaves
+# exactly as before.  Set MEMPALACE_RECENCY_WEIGHT to a value in (0, 1] to
+# blend in an exponential-decay recency factor; older drawers fade by half
+# every MEMPALACE_RECENCY_HALF_LIFE_DAYS (default 30).  Useful when a palace
+# accumulates years of session transcripts but the user is asking about
+# *current* state ("what was the latest decision on X?").
+_RECENCY_WEIGHT_DEFAULT = float(os.environ.get("MEMPALACE_RECENCY_WEIGHT", "0.0") or 0.0)
+_RECENCY_HALF_LIFE_DAYS_DEFAULT = float(
+    os.environ.get("MEMPALACE_RECENCY_HALF_LIFE_DAYS", "30") or 30
+)
+
+
+def _recency_factor(filed_at_str, half_life_days: float = _RECENCY_HALF_LIFE_DAYS_DEFAULT) -> float:
+    """Return a recency multiplier in [0, 1] for a drawer's ``filed_at`` timestamp.
+
+    Returns 1.0 for "filed now" and decays by half every ``half_life_days``.
+    Future-dated entries (clock skew) are treated as fresh (1.0).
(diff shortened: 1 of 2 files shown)


jphein:
+1 from operator-experience.

The "old high-similarity drawer masking a recent decision" failure mode is real on a ~160K-drawer production palace. Last month's actual answer on a topic gets buried under earlier higher-cosine drawers from the same problem space — pure vector ranking has no notion of time.

The default-off contract (`MEMPALACE_RECENCY_WEIGHT=0.0` as strict no-op) is the right shape; the env-var + per-call override pair gives a clean opt-in path.

Composes well with #1005 (warnings + sqlite BM25 fallback): BM25-promoted hits carry their own `filed_at`, so a recency-weighted result that mixes vector and BM25 ranking paths stays honest about which scoring axis surfaced what. `matched_via` and `recency_factor` stay distinct signals on the same result dict rather than collapsing.

Thanks for filing this.

Issue #1426, opened 2026-05-09
[BUG] MemPalace MCP Handler AttributeError when metadata is None (Queue Stuck)


wangprince:
## Problem Description
The MemPalace MCP server experienced a critical failure during the embedding queue cleanup process. When a memory item has `None` as its metadata, the MCP handler attempts to call `.get()` on it, resulting in an `AttributeError: 'NoneType' object has no attribute 'get'`. 

This causes the cleanup step (`DELETE FROM embeddings_queue`) to be skipped, leading to a buildup of processed-but-not-deleted items in the queue, effectively stalling the queue monitoring and reporting logic.

## Evidence
- **Log Traceback:** `AttributeError: 'NoneType' object has no attribute 'get'` found in `mcp-stderr.log`.
- **Symptom:** `embeddings_queue` count remains high even after vectors are successfully written to ChromaDB.
- **Environment:** Intel Q6600 (no-AVX), Ubuntu 24.04.

## Proposed Fixing Paths (Prioritized)
As discussed with @Prince, the fix will follow these three paths:
1. **Unified Defensive Wrapping (`searcher.py`):** Implement a centralized normalization (e.g., `_first_or_empty`) to ensure `None` metadata is always converted to an empty dictionary `{}` before reaching the handler.
2. **Write-side Interception:** Prevent `None` metadata from being written to the database in the first place.
3. **None Metadata Simulation Tests:** Create a comprehensive test suite that specifically simulates metadata being `None` to ensure the fix is robust and prevents regression.

## Current Status
- Manual cleanup performed via SQL: `DELETE FROM embeddings_queue WHERE seq_id IN (SELECT seq_id FROM embeddings)`.
- Local site-packages patch applied to `searcher.py` for immediate mitigation.
- Moving discussion from Discord to GitHub Issues to avoid rate-limits and ensure persistence.

mvalentsev:
For cross-reference: this crash class is tracked at the architectural level in #1020 (coerce None metadatas at the ChromaDB backend boundary).

PR #1427, opened 2026-05-09
fix/issue 1426 none metadata

wangprince:
- **fix: 修正 MemPalace 在 HNSW 索引與 SQLite 資料不一致時可能導致的 Segfault 問題**
- **fix: stabilize offline MemPalace initialization**
- **fix: handle None metadata in search and duplicate detection to prevent queue stall (Issue #1426)**

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -868,11 +868,15 @@ class ChromaBackend(BaseBackend):
         the EF explicitly — ChromaDB 1.x does not persist it with the
         collection, so a reader that omits the argument silently gets the
         library default and its queries won't match the writer's vectors.
+        
+        Uses unified offline local embedding function from embedding.py module.
         """
         try:
             from ..embedding import get_embedding_function
 
-            return get_embedding_function()
+            ef = get_embedding_function()
+            logger.debug(f"Using embedding function: {ef.name() if hasattr(ef, 'name') else type(ef).__name__}")
+            return ef
         except Exception:
             logger.exception("Failed to build embedding function; using chromadb default")
             return None
@@ -1033,9 +1037,8 @@ class ChromaBackend(BaseBackend):
         if options and isinstance(options, dict):
             hnsw_space = options.get("hnsw_space", hnsw_space)
 
-        # Q6600 AVX FIX: Use SentenceTransformer instead of default ONNX EF
-        from chromadb.utils.embedding_functions import SentenceTransformerEmbeddingFunction
-        ef = SentenceTransformerEmbeddingFunction(model_name="all-MiniLM-L6-v2")
+        from ..embedding import get_embedding_function
+        ef = get_embedding_function()
 
         if create:
             collection = client.get_or_create_collection(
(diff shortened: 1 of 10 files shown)


igorls:
Thanks for the substantial effort here @wangprince — closing this so we can land focused pieces without entangled concerns, but several of the underlying ideas are worth re-submitting as separate PRs.

## Why we're closing rather than merging

The PR bundles six commits across unrelated concerns, and the framing of the commit messages doesn't quite match the actual diff:

1. **`e536367` (HNSW + SQLite segfault fix)** — the commit message describes a capacity-check probe, a ``vector_disabled`` flag, and auto-route to BM25 fallback, but the actual diff just hardcodes ``SentenceTransformerEmbeddingFunction(all-MiniLM-L6-v2)`` with ``# Q6600 AVX FIX`` comments. None of the probe / fallback code is present. It also doesn't address #1499 (the open HNSW drift bug on the milestone).
2. **`5c9c9b2` (stabilize offline init)** — pulls in ``sentence-transformers`` as a runtime dependency. Our ``pyproject.toml`` keeps ChromaDB + PyYAML as the only core deps (per CONTRIBUTING.md / CLAUDE.md). A new embedding backend deserves a dedicated design discussion rather than landing inside an offline-init commit.
3. **`682110d` (None metadata, #1426)** — #1426 was closed on develop before this PR was filed; the same crash class is already handled there. The new ``test_none_resilience.py`` may still add coverage worth porting, but the production change is duplicate.
4. **`d2601fb` (mcp: lock caches; partial os.write; unify embedding init; hardcoded model path)** — this is four separate fixes in one commit. The cache lock + the partial-write fix are likely both legitimate and would be very welcome as their own small PRs.
5. **`d732119` (make embedding monkeypatchable)** — small and reasonable; would land easily on its own.
6. **`eca0e5a` (pin numpy<2.0 for no-AVX compatibility)** — this is the blocker. Develop currently uses numpy 2.4.4 and the surrounding deps assume it; downgrading would break the install for the vast majority of users. The Q6600/AVX concern is real for a small subset of older hardware, but the right fix is documenting an opt-in workaround (e.g., a ``[tool.uv.workarounds]`` section, or guidance in CONTRIBUTING.md for users on pre-AVX2 CPUs) rather than pinning the whole project down.

## What we'd love to see resubmitted

If you'd like to break this into focused PRs, here's the priority order from our side:

- **The cache-lock fix from d2601fb** — its own PR, just the locking change, with a test that demonstrates the race it closes.
- **The partial ``os.write`` fix from d2601fb / the JSON-RPC stdout switch** — its own PR; this is a real MCP correctness issue.
- **The monkeypatchable embedding refactor from d732119** — its own PR; trivially mergeable on its own.
- **A docs PR for pre-AVX2 hardware** — explain how to opt into a SentenceTransformer-based embedder via env var or config, without changing the default. Don't pin numpy at the project level.

If the actual HNSW capacity probe described in `e536367`'s commit message exists in your local tree, that would be very welcome as a separate PR — it likely overlaps with what we need for #1499.

Closing #1427 now so the milestone view is clear; please reopen / file new PRs at your pace. Thank you again for taking the time to look at these failure modes — the partial-write and cache-lock observations in particular are real.

PR #1438, opened 2026-05-10
Fix UTF-8 lock holder writes on Windows

### LINKED ISSUE #1435: palace.py: mine_palace_lock holder write uses text mode without explicit encoding (Windows non-ASCII argv hazard)
Surfaced by Copilot review on #1434. Pre-existing on develop; out of scope for the 3.3.5 release but worth fixing.

## Where

[mempalace/palace.py:405-412](https://github.com/MemPalace/mempalace/blob/develop/mempalace/palace.py#L405) — `_write_lock_holder()`:

\`\`\`python
ident = f\"{os.getpid()} {' '.join(sys.argv[:3])}\".strip()
lock_file.seek(_LOCK_SENTINEL_BYTES)
lock_file.truncate(_LOCK_SENTINEL_BYTES + len(ident.encode(\"utf-8\")))
lock_file.write(ident)
lock_file.flush()
\`\`\`

## Problem

Two related issues:

1. The byte budget for \`truncate()\` is computed via \`ident.encode(\"utf-8\")\`, but \`lock_file.write(ident)\` is a text-mode write that uses the file's encoding. On Windows where \`open()\` defaults to the system ANSI codepage (cp1252 / cp1251 / cp950) rather than UTF-8, the truncate length and the actual on-disk byte count diverge for non-ASCII content — leaving either stale tail bytes from a previous holder or undersized writes.
2. The handler is \`except OSError: pass\`, but \`UnicodeEncodeError\` is **not** a subclass of \`OSError\`. On a non-UTF-8 Windows locale, an argv with non-ASCII bytes (a path containing accented characters, CJK, Cyrillic) would propagate \`UnicodeEncodeError\` out of \`_write_lock_holder\` and break lock acquisition.

## Suggested fix

Either:

- Open the lock file with \`encoding=\"utf-8\"\` at the call site, OR
- Use binary mode and explicit \`ident.encode(\"utf-8\")\` for the write, AND
- Broaden the exception handler to \`except (OSError, UnicodeError): pass\`.

The lock-holder string is purely informational (used by \`MineAlreadyRunning\` error messages to point at the holder PID); a write failure here should be best-effort and never block the lock from being acquired.

## Repro

Manual: on a Windows shell with locale set to non-UTF-8 (e.g. \`chcp 1252\`), run \`mempalace mine\` from a directory whose path contains non-ASCII characters. Expected: lock acquires cleanly; observed: \`UnicodeEncodeError\` from \`_write_lock_holder\`.

## Refs

- Found via Copilot review on #1434: https://github.com/MemPalace/mempalace/pull/1434#discussion_*

fatkobra:
## What does this PR do?

Fixes #1435.

This updates the palace mine lock-holder identity to be written as explicit UTF-8 bytes instead of text-mode output using the platform default encoding.

## What changed

- Opens the per-palace lock file in binary read/write mode (`r+b`)
- Decodes holder identity as UTF-8 when reading diagnostics
- Writes holder identity as UTF-8 bytes when recording the current process
- Keeps holder recording best-effort by catching `UnicodeError` as well as `OSError`
- Adds regression coverage for non-ASCII argv values and best-effort Unicode failures

## Why

Previously `_write_lock_holder()` calculated the truncate size using `ident.encode("utf-8")`, but then wrote `ident` through a text-mode file. On Windows non-UTF-8 locales, this could make the truncate byte count disagree with the actual bytes written, or raise `UnicodeEncodeError`.

The holder string is diagnostic-only, so a failure to record it should never prevent the lock from being acquired.

## How to test

```bash
python -m pytest tests/test_palace_locks.py
python -m pytest tests/test_palace_locks.py -k "lock_holder or utf8 or unicode"
python -m ruff check mempalace/palace.py tests/test_palace_locks.py
python -m pytest tests/ -v
```

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -387,7 +387,10 @@ def _read_lock_holder(lock_file) -> str:
     """Read the prior holder's identity from the lock-file body, best-effort."""
     try:
         lock_file.seek(_LOCK_SENTINEL_BYTES)
-        content = lock_file.read().strip()
+        content = lock_file.read()
+        if isinstance(content, bytes):
+            content = content.decode("utf-8", errors="replace")
+        content = content.strip()
     except OSError:
         return "another writer (identity not recorded)"
     if not content:
@@ -404,11 +407,12 @@ def _write_lock_holder(lock_file) -> None:
     """
     try:
         ident = f"{os.getpid()} {' '.join(sys.argv[:3])}".strip()
+        ident_bytes = ident.encode("utf-8")
         lock_file.seek(_LOCK_SENTINEL_BYTES)
-        lock_file.truncate(_LOCK_SENTINEL_BYTES + len(ident.encode("utf-8")))
-        lock_file.write(ident)
+        lock_file.truncate(_LOCK_SENTINEL_BYTES + len(ident_bytes))
+        lock_file.write(ident_bytes)
         lock_file.flush()
-    except OSError:
+    except (OSError, UnicodeError):
         pass
 
 
@@ -470,7 +474,7 @@ def mine_palace_lock(palace_path: str):
             os.close(fd)
         except FileExistsError:
             pass
-    lf = open(lock_path, "r+")
+    lf = open(lock_path, "r+b")
     acquired = False
     try:
         # Lock byte 0 explicitly. msvcrt.locking is byte-position dependent;
--- a/tests/test_palace_locks.py
+++ b/tests/test_palace_locks.py
@@ -12,10 +12,12 @@ from __future__ import annotations
 import multiprocessing
 import os
 import time
+import sys
 
 import pytest
 
 from mempalace.palace import (
+    _write_lock_holder,
     MineAlreadyRunning,
     mine_global_lock,
     mine_palace_lock,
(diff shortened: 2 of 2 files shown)


fatkobra:
The first commit was only for the patch and all the tests and the lint passed.

Then I made the 2nd commit with the regression tests on `tests/test_palace_locks.py`, lint failed and the tests passed.

The logs shows that: **All checks passed!** for `ruff check`
```
Run ruff check .
  ruff check .
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.11.15/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.15/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.15/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.15/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.15/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.15/x64/lib
All checks passed!
```

However `ruff format --check .` fails.

```
Run ruff format --check .
Would reformat: tests/test_palace_locks.py
1 file would be reformatted, 121 files already formatted
Error: Process completed with exit code 1.
```

Issue #1423, opened 2026-05-09
PYTHONPATH leaks into venv, causing pydantic_core import failure on mixed Python environments


jesscore:
**What happened?**

Running any mempalace command fails with ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'. The full traceback goes through chromadb → pydantic → pydantic_core and loads the wrong version's compiled extension. 

**What did you expect?**

mempalace runs normally inside its venv regardless of what PYTHONPATH is set to in the environment.

**How to reproduce:**
                                                                                            
  1. Have multiple Python versions installed (e.g. 3.9 and 3.11)                                                     
  2. Have PYTHONPATH set to the 3.9 site-packages (e.g. ~/Library/Python/3.9/lib/python/site-packages)               
  3. Run mempalace status — the venv's Python 3.11 picks up the 3.9 pydantic_core native extension, which can't load 
  in 3.11                                                                                                            
                                                                                                                     
**Workaround:**

PYTHONPATH="" mempalace <command>                                                                      
                                                                                                                     
**Root cause:**
Python venvs don't isolate PYTHONPATH — it leaks in even when running the venv binary directly. Fix: add import os; os.environ.pop('PYTHONPATH', None) at the top of cli.py.

**Environment:**
  - OS: macOS 15.7.3
  - Python version: 3.11 (venv), 3.9 (system with PYTHONPATH set)
  - MemPalace version: 3.0.0

PR #1439, opened 2026-05-10
fix: drop leaked PYTHONPATH on package init (#1423)

mvalentsev:
## What does this PR do?

Protects MemPalace against a leaked `PYTHONPATH` that points at another
Python's `site-packages`. This was the failure mode in #1423: a venv
running Python 3.11 inheriting `PYTHONPATH` from a Python 3.9 install
crashes on the first MemPalace command with `ModuleNotFoundError: No
module named 'pydantic_core._pydantic_core'` because the interpreter
preloaded the wrong-ABI compiled extension from the leaked path.

The fix is split between package init and the CLI/MCP entry points so
that embedded library callers are not affected.

**`mempalace/__init__.py`** filters `sys.path` only:
`_strip_leaked_pythonpath_from_sys_path()` reads `PYTHONPATH` via
`os.environ.get` (does NOT pop it), and removes from `sys.path` any
entry that matches a path listed in `PYTHONPATH`. Comparison goes
through `os.path.normcase` + `os.path.normpath` so case differences
(Windows) and trailing-separator variants do not slip through string
equality. The empty-string CWD marker on `sys.path` is preserved
unconditionally, so `PYTHONPATH=.` does not collapse the implicit
current directory.

**`mempalace/cli.py:main`** and **`mempalace/mcp_server.py:main`** each
call `os.environ.pop("PYTHONPATH", None)` as their first line, so any
subprocess these console scripts spawn (mine workers, repair tooling)
starts with a clean env. The pop runs before stdio reconfigure / stdout
restore in both entry points so the ordering is symmetric.

## Placement rationale

`sys.path` filter stays in `mempalace/__init__.py`: it runs at package
import and fixes the ABI mismatch crash for the current process
without touching `os.environ`. Host applications that import mempalace
as a library (for example `import mempalace.searcher`) keep their
`PYTHONPATH` intact for their own unrelated subprocesses.

`os.environ.pop("PYTHONPATH")` lives at the CLI/MCP entry points so the
strip affects only processes the user explicitly invoked (root of
trust) and the subprocesses those entry points spawn themselves.

## How to test

Import-time should preserve env and clean sys.path:

```bash
PYTHONPATH=/tmp/some/leaked/path uv run python -c "import mempalace, os, sys; print('ENV:', os.environ.get('PYTHONPATH')); print('IN_PATH:', '/tmp/some/leaked/path' in sys.path)"
# -> ENV: /tmp/some/leaked/path
# -> IN_PATH: False
```

CLI entry point should strip env:

```bash
PYTHONPATH=/tmp/some/leaked/path uv run python <<'PY'
import os, sys
from mempalace.cli import main
sys.argv = ['mempalace', '--version']
try:
    main()
except SystemExit:
    pass
print('ENV_AFTER:', os.environ.get('PYTHONPATH'))
PY
# -> MemPalace 3.3.5
# -> ENV_AFTER: None
```

CWD-collision edge case:

```bash
PYTHONPATH=. uv run python -c "import mempalace, sys; print('CWD:', '' in sys.path); print('DOT:', '.' in sys.path)"
# -> CWD: True
# -> DOT: False
```

Test coverage in `tests/test_init.py`, `tests/test_cli.py`,
`tests/test_mcp_server.py`. The init tests assert env is preserved AND
sentinel-prefixed `sys.path` entries are removed AND the mempalace
parent directory survives (no over-strip). The CLI/MCP tests assert
ENV_MID between import and main is preserved (proving main is the strip
site, not the import), SENTINEL_IN_PATH is False at import (proving the
sys.path filter ran), and ENV_AFTER is None.

Full suite + ruff:

```
```

## Checklist

- [x] Tests pass locally (`uv run pytest tests/ -v --ignore=tests/benchmarks`): 1735 passed
- [x] No hardcoded paths
- [x] Linter passes (`uv run ruff check . && uv run ruff format --check .`)

### DIFF
--- a/mempalace/__init__.py
+++ b/mempalace/__init__.py
@@ -9,13 +9,18 @@ def _strip_leaked_pythonpath() -> None:
     # Venvs inherit PYTHONPATH; on multi-Python systems it can cause
     # transitive imports to load compiled extensions (pydantic_core,
     # chromadb_rust_bindings) from the wrong ABI. Drop the env var
-    # and remove the sys.path entries the interpreter populated from
-    # it before the consumer imports anything that ships compiled code.
+    # and remove sys.path entries the interpreter populated from it.
+    # Comparison normalizes case + separators so Windows paths and
+    # trailing-separator quirks do not slip through string equality.
     leaked = os.environ.pop("PYTHONPATH", None)
     if not leaked:
         return
-    leaked_entries = {p for p in leaked.split(os.pathsep) if p}
-    sys.path[:] = [p for p in sys.path if p not in leaked_entries]
+
+    def _norm(path: str) -> str:
+        return os.path.normcase(os.path.normpath(path))
+
+    leaked_entries = {_norm(p.strip()) for p in leaked.split(os.pathsep) if p.strip()}
+    sys.path[:] = [p for p in sys.path if _norm(p) not in leaked_entries]
 
 
 _strip_leaked_pythonpath()
--- a/mempalace/__init__.py
+++ b/mempalace/__init__.py
@@ -5,16 +5,24 @@ import os
 import sys
 
 
-def _strip_leaked_pythonpath() -> None:
+def _strip_leaked_pythonpath_from_sys_path() -> None:
     # Venvs inherit PYTHONPATH; on multi-Python systems it can cause
     # transitive imports to load compiled extensions (pydantic_core,
-    # chromadb_rust_bindings) from the wrong ABI. Drop the env var
-    # and remove sys.path entries the interpreter populated from it.
-    # Comparison normalizes case + separators so Windows paths and
-    # trailing-separator quirks do not slip through string equality.
-    # The empty-string CWD marker on sys.path is preserved regardless,
-    # so PYTHONPATH=. does not collapse the implicit current directory.
-    leaked = os.environ.pop("PYTHONPATH", None)
+    # chromadb_rust_bindings) from the wrong ABI. Remove sys.path entries
+    # the interpreter populated from PYTHONPATH so this process imports
+    # only the venv's own packages. Comparison normalizes case + separators
+    # so Windows paths and trailing-separator quirks do not slip through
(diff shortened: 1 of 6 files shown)


mvalentsev:
Both points addressed in 37bb206:

1. `_strip_leaked_pythonpath()` now removes the matching entries from `sys.path` as well, so the import search path is cleaned alongside `os.environ`. Compiled-extension imports (`pydantic_core`, `chromadb_rust_bindings`) can no longer resolve from a leaked location.

2. The parametrized test now covers single-path, multi-path (`/a:/b`), and unset PYTHONPATH cases, asserting both that `os.environ.get('PYTHONPATH') is None` AND that no leaked entry remains on `sys.path` after `import mempalace`.

On placement: keeping the call in `mempalace/__init__.py` is intentional. Library-style imports of compiled-extension-using submodules (`mempalace.searcher`, `mempalace.knowledge_graph`) hit the same ABI mismatch as the CLI, so moving to `cli.py` would leave that path unprotected. The env mutation is documented in the comment.

Empirical verification:

```
PYTHONPATH=/tmp/leak-a:/tmp/leak-b uv run python -c \
  "import mempalace; import sys, os; \
   assert '/tmp/leak-a' not in sys.path; \
   assert '/tmp/leak-b' not in sys.path; \
   assert os.environ.get('PYTHONPATH') is None; \
   print('OK')"
# OK
```

mvalentsev:
Second look on the side effect concern from your earlier review: the embedded-library case is real. A host app that imports mempalace as a library and sets `PYTHONPATH` for its own subprocesses would have it silently removed by the previous patch.

Split the fix in e467fec:

- `mempalace/__init__.py` reads `PYTHONPATH` but does not modify it; only the matching `sys.path` entries get filtered, so the current process is still protected against the ABI crash from #1423.
- `mempalace/cli.py:main` and `mempalace/mcp_server.py:main` each pop `PYTHONPATH` from the env as their first line, so subprocesses spawned by the console scripts get a clean env.

Library callers keep their `PYTHONPATH` intact.

PR #1448, opened 2026-05-11
fix(benchmarks): auto-pull embedding model when missing instead of silently scoring 0

lealbrunocalhau:
## Summary

- `memory_extraction` and `room_classification (open)` both depend on `nomic-embed-text` for semantic scoring. If the model was absent, `embed_text()` returned `None` silently — every sample scored `0.0`, indistinguishable from a genuinely bad model.
- `_ensure_embed_model()` now probes the endpoint **before** warmup. On a 404 it runs `ollama pull` automatically, verifies the endpoint responds, and only then proceeds.
- If the pull fails, the task returns a `Result` with a clear `error` message instead of silent zeros.
- Also fixes `--candidates local` filter (was returning all non-cloud models instead of `tier==local` entries) and adds a `community` tier for third-party GGUF variants.

## Root cause

`embed_text()` in `metrics.py` catches all exceptions and returns `None`. The scorer then skips every unmatched item, producing `coverage=0` and `hallucination_rate=1.0` — which looks like model failure, not missing infrastructure.

## Test plan

- [ ] Pull `nomic-embed-text` model, run `memory_extraction` — scores should reflect real model output
- [ ] Remove `nomic-embed-text`, re-run — harness should pull it automatically and proceed
- [ ] Remove `nomic-embed-text`, make `ollama pull` fail (e.g. offline) — result should have a clear `error` field, not silent zeros
- [ ] `--candidates local` should only return entries with `tier: local`, not all non-cloud models

### DIFF
--- a/benchmarks/model_eval/orchestrator.py
+++ b/benchmarks/model_eval/orchestrator.py
@@ -73,6 +73,8 @@ def load_candidates(path: Path, tier: str) -> list[dict]:
         return [c for c in candidates if c.get("tier") != "cloud" and not c.get("cloud")]
     if tier == "modern":
         return [c for c in candidates if c.get("tier") == "modern"]
+    if tier == "community":
+        return [c for c in candidates if c.get("tier") == "community"]
     if tier.startswith("tier<="):
         try:
             n = int(tier.split("<=")[1])
--- a/benchmarks/model_eval/runner.py
+++ b/benchmarks/model_eval/runner.py
@@ -15,6 +15,8 @@ from __future__ import annotations
 import argparse
 import dataclasses
 import json
+import os
+import subprocess
 import sys
 import time
 from dataclasses import asdict, dataclass, field
@@ -29,6 +31,7 @@ from .metrics import (
     TimingSample,
     VRAMPoller,
     aggregate_timings,
+    embed_text,
     extract_timing,
     gather_host_info,
     strip_thinking_tokens,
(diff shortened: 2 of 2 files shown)


igorls:
## Code Review

### Overview

The PR addresses a real problem — `memory_extraction` and `room_classification:open` silently scored 0 when `nomic-embed-text` wasn't installed, because `embed_text()` in `metrics.py` swallows all exceptions and returns `None`. The fix probes the embedding endpoint, runs `ollama pull` if it 404s, and returns a clear `error` if that fails. Also tweaks `load_candidates` filters.

The bug is real, the fix direction is right, and the failure mode (`error` field instead of zeros) is the correct UX. But there are a few issues worth addressing before merge.

### Issues

**(1) Breaking change to `--candidates local` (high priority)**

`orchestrator.py` changes `local` from "everything not cloud" to "only entries with `tier: local`". A grep of `candidates.yaml` shows **zero entries** with `tier: local` or `tier: community`:

```
tier: 1, tier: modern, tier: 2, tier: 3, tier: cloud
```

So after this PR, `uv run python -m benchmarks.model_eval.orchestrator --candidates local ...` returns an empty matrix and runs zero models. This is exactly the invocation documented at `README.md:90` as the \"Everything local (~60-80 min)\" recipe — a silent no-op there will be very confusing.

Either:
- Keep `local` as \"non-cloud\" (its documented meaning) and introduce a separate `tier:local` filter under a different name, OR
- Tag existing models with `tier: local` in `candidates.yaml` and update the README to mention the new `non-cloud` alias.

The PR description says this is a \"fix\" but it's a semantic break with no callsite migration.

**(2) `ollama pull` ignores the `endpoint` parameter**

`_ensure_embed_model` probes `embed_text(\"ping\", endpoint=endpoint)` against whatever endpoint the run is configured for, but then shells out to `ollama pull` with no `OLLAMA_HOST` env override. If the user is benchmarking a non-localhost Ollama (e.g., a remote box or alternate port), the probe and the pull target different hosts — the pull \"succeeds\" against localhost, the re-probe against the real endpoint still fails, and you get the second `RuntimeError` (\"pulled but is still not responding\"). That matches the gemini-code-assist comment, and it's a legitimate one.

Suggested:
```python
env = {**os.environ, \"OLLAMA_HOST\": endpoint}
result = subprocess.run([\"ollama\", \"pull\", model], env=env, ...)
```

**(3) Missing `FileNotFoundError` handling**

If `ollama` is not on PATH, `subprocess.run([\"ollama\", \"pull\", model], ...)` raises an uncaught `FileNotFoundError`, which the caller in `runner.py run()` does not catch (only `RuntimeError` / `subprocess.TimeoutExpired`). The run crashes instead of producing the clean `Result(error=...)` row the PR is trying to deliver. Add `FileNotFoundError` to both the `except` and the wrapper — or catch it inside `_ensure_embed_model` and re-raise as `RuntimeError` with the install hint.

**(4) Use of `print()` for diagnostic output**

`_ensure_embed_model` writes to stdout while everything else in `runner.py` returns structured `Result` rows. The orchestrator parses runner output; mixing free-form prints into stdout can interfere when this is shelled out (though it appears the orchestrator imports `run` directly, so this is mostly cosmetic). Minor — would prefer `print(..., file=sys.stderr)`.

### Smaller notes

- `_EMBED_TASKS` is a nice declarative gate, but consider co-locating it with the task modules themselves (e.g., a `REQUIRES_EMBEDDING = True` marker on `memory_extraction` and `room_classification`) so adding a new embedding-scored task doesn't require remembering to update a set in `runner.py`. Not blocking.
- The 300s pull timeout is reasonable for `nomic-embed-text` (~270MB) but worth a comment if it ever expands to larger embedding models.
- No tests. Pulling a real model in CI isn't viable, but `_ensure_embed_model` is exactly the kind of function where mocking `embed_text` + `subprocess.run` would catch the FileNotFoundError and OLLAMA_HOST bugs above.

### Security / correctness

- `subprocess.run` is called with a list (no `shell=True`) and a hardcoded model name constant, so no injection risk.
- No data leaves the machine; respects the local-first principle.

### Recommendation

Request changes — concern (1) (the `--candidates local` break) is the blocker because it silently breaks a documented workflow. Concerns (2) and (3) are correctness gaps in the new code path. Once those are addressed this is a good fix.

igorls:
@lealbrunocalhau the harness PR was already merged, please rebase to develop now

igorls:
Hey @lealbrunocalhau — thanks for the quick turnaround on the follow-up commit (`ddfd12d`), looks like it addresses all four points. Since this PR is closed and GitHub won't let us reopen it (the original head SHA is no longer in the branch after the force-push), could you open a fresh PR from `workblac:fix/benchmark-embedding-model-preflight` targeting `MemPalace:develop` instead of `feat/model-eval-harness`? I'll re-review there.

Issue #1457, opened 2026-05-11
quarantine_stale_hnsw integrity gate misses zero-byte link_lists.bin (SIGSEGV cascade)


jphein:
# BUG: `quarantine_stale_hnsw` integrity gate misses zero-byte `link_lists.bin`

**Draft for filing** — discovered 2026-05-10 while diagnosing a SIGSEGV cascade in palace-daemon.

---

**What happened?**

`mempalace/backends/chroma.py:quarantine_stale_hnsw()` declared two HNSW segment directories "healthy" and left them in place, even though their `link_lists.bin` files were 0 bytes and their `index_metadata.pickle` sidecars were missing entirely. When chromadb later attempted to load these segments during a `/search` operation, the C extension dereferenced into an empty HNSW graph structure and segfaulted.

The result: palace-daemon went into a 2-minute crash-restart loop. Every restart loaded the segments fine (just `os.listdir` and metadata reads), but the first `/search` query that touched the corrupt segments triggered a `SIGSEGV` (status 11) in chromadb's `hnswlib` binding. systemd restarted the daemon under `Restart=on-failure`, the cycle continued.

Concretely, two segments on a 151K-drawer production palace showed this state:

```
4697d280-58d9-4276-b3b5-53991a1a9237/
  data_level0.bin    167600 bytes
  header.bin            100 bytes
  length.bin            400 bytes
  link_lists.bin          0 bytes   ← corrupt
  (index_metadata.pickle missing)
```

The daemon's startup log noted "HNSW mtime gap … exceeds threshold but segment metadata file is intact — flush-lag, not corruption. Leaving in place." So the existing two-stage check (mtime gate + integrity gate) explicitly cleared these segments.

**What did you expect?**

`_segment_appears_healthy()` (called from `quarantine_stale_hnsw`) should fail integrity for a segment whose `link_lists.bin` is zero bytes or whose `index_metadata.pickle` sidecar is absent. Either condition guarantees the HNSW graph cannot be loaded; rebuilding via `mempalace repair --mode rebuild` from SQLite ground truth is always safer than letting chromadb attempt to read an empty graph and crash the process.

**How to reproduce:**

1. On any chromadb 1.5.x palace with `link_lists.bin` truncated to 0 bytes in one or more segment dirs (can happen from interrupted writes, filesystem issues, partial restores, or upstream chromadb flush bugs), restart palace-daemon.
2. Wait for startup to complete (~15s). Daemon logs "Application startup complete."
[...]

PR #1459, opened 2026-05-11
fix(repair): coerce empty metadata to sentinel during rebuild

### LINKED ISSUE #1458: repair --mode rebuild aborts on drawers with empty metadata (ValueError in chromadb 1.5.x upsert)
# BUG: `mempalace repair --mode rebuild` aborts on drawers with empty/missing metadata

**Draft for filing** — discovered 2026-05-11 during foundation-rework recovery from the chromadb SEGV cascade.

---

## What happened

`mempalace repair --mode rebuild` extracted 151,478 drawers from a recovered palace SQLite ground truth, started the live-collection rebuild, and after ~7h 44min CPU (2h 35min wall) on a ~120K-drawer pass died with:

```
ValueError: Expected metadata to be a non-empty dict, got 0 metadata attributes in add.
  File "...chromadb/api/types.py", line 1071, in validate_metadata
```

`chromadb` 1.5.x's `Collection.upsert()` validates that every entry in `metadatas[]` is a non-empty dict. The repair extractor `_extract_drawers` faithfully relays whatever the source collection's `col.get(..., include=["documents","metadatas"])` returned — which on this palace included some drawers with empty metadata (probably sparse historical writes from older mempalace versions, or rows whose only metadata key had all-NULL values that chromadb's reader normalizes to `{}`).

The repair has a partial-failure rollback (good), so SQLite was restored from `chroma.sqlite3.backup` and the live collection was deleted — but the operator now has no working palace and no way to make forward progress without surgery.

## What should happen

`_extract_drawers` should sanitize empty metadata before staging it for re-upsert. The simplest sanitizer is a sentinel:

```python
sanitized_metas = [
    m if (isinstance(m, dict) and len(m) > 0) else {"_repaired_empty_meta": True}
    for m in batch["metadatas"]
]
all_metas.extend(sanitized_metas)
```

Slot this between `batch = col.get(...)` and `all_metas.extend(batch["metadatas"])` in `mempalace/repair.py:_extract_drawers`. Three-line change.

## Why "_repaired_empty_meta": True

- It's a non-empty dict, so chromadb validates clean.
- The key is namespaced + descriptive, so an operator searching the palace later can find which drawers had to be repaired this way.
- The value is `True` (a bool, valid chromadb metadata type), trivially serializable.
- Future repair runs over a sanitized palace are idempotent — the sentinel survives a round-trip through chromadb extract/upsert intact.

## How to reproduce

1. Take any palace whose SQLite contains drawers with empty metadata. (One way to get there: ingest a drawer via an older mempalace version that allowed it, then upgrade to chromadb 1.5.x.)
2. Quarantine the HNSW segments (or otherwise force `repair --mode rebuild` to run a fresh rebuild).
3. Watch repair extract cleanly, start rebuilding, and crash on the first empty-meta drawer.

## Workaround used today

Patched `mempalace/repair.py` in the jphein fork with the sanitizer above, reinstalled into the palace-daemon venv via in-place file copy, restored SQLite from backup, re-ran repair. Verified the patch loaded:

```python
import mempalace.repair as r, inspect
"sanitized_metas" in inspect.getsource(r._extract_drawers)  # True
```

## Environment

- mempalace: jphein fork, sha at 2026-05-10 + the one-file patch above
- chromadb: 1.5.8
- Palace size at failure: 151,481 drawers in SQLite
- Hit the failing drawer somewhere in the 120,001–125,000 range during rebuild

## Related

Two other mempalace/palace-daemon bugs from this session:

- `BUG-2026-05-10-quarantine-stale-hnsw-misses-empty-link-lists.md` (mempalace integrity gate)
- palace-daemon `/backup` endpoint not implemented (see palace-daemon scratch)
- palace-daemon crash-loop detection missing (see palace-daemon scratch)

---

Filed by: JP (homelab operator)
Diagnostic session: 2026-05-11 05:50 PDT (continuing 2026-05-10)

jphein:
## Summary

- Sanitize `None`/`{}` metadata to `{"_repaired_empty_meta": True}` in `_extract_drawers` before staging for the rebuild upsert.
- Unblocks `repair --mode rebuild` on palaces with any drawer that has sparse/empty metadata. chromadb 1.5.x's `upsert()` validates that every `metadatas[i]` is a non-empty dict and otherwise raises `ValueError: Expected metadata to be a non-empty dict, got 0 metadata attributes in add.` (chromadb/api/types.py:validate_metadata, ~line 1071).

Fixes #1458.

## Why this matters

On a 151,478-drawer production palace (jphein's homelab, 2026-05-11), `repair --mode rebuild` ran for **2h 35min wall / 7h 44min CPU** before dying at drawer ~120,000–125,000 with that ValueError. The built-in partial-failure rollback worked (sqlite was restored from `chroma.sqlite3.backup`, live collection deleted) but the operator had no forward path without surgery on the sqlite ground truth.

The bad drawers are almost certainly sparse historical writes from older mempalace versions; chromadb's reader normalizes those to `None`/`{}` rows in the extracted result, which the writer then refuses.

## The fix

Replace:
```python
all_metas.extend(batch["metadatas"])
```

with:
```python
sanitized_metas = [
    m if (isinstance(m, dict) and len(m) > 0) else {"_repaired_empty_meta": True}
    for m in batch["metadatas"]
]
all_metas.extend(sanitized_metas)
```

Why `_repaired_empty_meta: True` as the sentinel:

1. Satisfies chromadb's non-empty-dict requirement.
2. Bool-valued — valid chromadb metadata type, trivially serializable.
3. Namespaced + descriptive — operators can filter for repaired drawers via `where={"_repaired_empty_meta": True}` after the fact if they want to investigate.
4. Idempotent — a future repair over a sanitized palace will see the sentinel as already-valid metadata and leave it alone.

## Performance impact

Negligible. On the test palace, fewer than ~1 in 10,000 drawers needed coercion. The list comprehension adds one `isinstance + len > 0` check per drawer, dwarfed by the upsert/embedding work.

## Test plan

- [x] Patched the venv copy of `mempalace/repair.py` in-place on disks, restarted `repair --mode rebuild` against the same 151,478-drawer palace that previously crashed. Currently running cleanly past the previous crash point.
- [ ] (Maintainer to consider) Add a unit test that feeds `_extract_drawers` a mock collection where `col.get()` returns `[{"k": "v"}, None, {}]` for `metadatas`, asserts the output passes chromadb's `validate_metadata` shape.

I haven't added the unit test in this PR because the existing `_extract_drawers` doesn't have its own test file in the tree and I didn't want to make a structural choice on test layout that a maintainer might prefer differently. Happy to add one in any form requested.

## Related issues opened

- `MemPalace/mempalace#1457` — `quarantine_stale_hnsw` integrity gate doesn't check `link_lists.bin` size (separate root cause; same homelab outage)
- `rboarescu/palace-daemon#19` — implement `/backup` endpoint (already on roadmap)
- `rboarescu/palace-daemon#21` — crash-loop detection

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -873,10 +873,30 @@ class ChromaCollection(BaseCollection):
     # Writes
     # ------------------------------------------------------------------
 
+    @staticmethod
+    def _sanitize_metadatas_for_chromadb(metadatas):
+        """chromadb 1.5.x rejects None and empty-dict entries in the metadatas
+        list (ValueError: Expected metadata to be a non-empty dict, got 0
+        metadata attributes in add). Coerce any such entry to a sentinel so
+        the write succeeds. Operators can later locate coerced drawers via
+        ``where={"_repaired_empty_meta": True}``.
+
+        This is the chokepoint catch-all: even if a caller's own sanitizer
+        misses a case (or skips for performance), reaching the chromadb
+        client always goes through here first.
+        """
+        if metadatas is None:
+            return None
+        return [
+            m if (isinstance(m, dict) and len(m) > 0) else {"_repaired_empty_meta": True}
+            for m in metadatas
+        ]
+
     def add(self, *, documents, ids, metadatas=None, embeddings=None):
         kwargs: dict[str, Any] = {"documents": documents, "ids": ids}
-        if metadatas is not None:
-            kwargs["metadatas"] = metadatas
+        sanitized = self._sanitize_metadatas_for_chromadb(metadatas)
+        if sanitized is not None:
+            kwargs["metadatas"] = sanitized
         if embeddings is not None:
             kwargs["embeddings"] = embeddings
         with self._write_lock():
@@ -884,8 +904,9 @@ class ChromaCollection(BaseCollection):
 
     def upsert(self, *, documents, ids, metadatas=None, embeddings=None):
         kwargs: dict[str, Any] = {"documents": documents, "ids": ids}
-        if metadatas is not None:
-            kwargs["metadatas"] = metadatas
+        sanitized = self._sanitize_metadatas_for_chromadb(metadatas)
+        if sanitized is not None:
+            kwargs["metadatas"] = sanitized
         if embeddings is not None:
             kwargs["embeddings"] = embeddings
         with self._write_lock():
--- a/mempalace/repair.py
+++ b/mempalace/repair.py
@@ -139,7 +139,17 @@ def _extract_drawers(col, total: int, batch_size: int):
             break
         all_ids.extend(batch["ids"])
         all_docs.extend(batch["documents"])
-        all_metas.extend(batch["metadatas"])
+        # chromadb 1.5.x's upsert validates that every metadatas[i] is a
+        # non-empty dict (chromadb/api/types.py:validate_metadata). Drawers
+        # extracted from sqlite ground truth can come back with None or {}
+        # for sparse historical writes — coerce those to a sentinel so the
+        # rebuild upsert can complete instead of raising ValueError ~80%
+        # through a multi-hour run. See #1458 for full context.
+        sanitized_metas = [
+            m if (isinstance(m, dict) and len(m) > 0) else {"_repaired_empty_meta": True}
+            for m in batch["metadatas"]
+        ]
+        all_metas.extend(sanitized_metas)
         offset += len(batch["ids"])
     return all_ids, all_docs, all_metas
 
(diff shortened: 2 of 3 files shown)


jphein:
Updated the patch (force-push) after a second test exposed a more serious issue: my original fix only covered `_extract_drawers`, but the actual failure path on the affected palace was `_rebuild_one_collection` calling `extract_via_sqlite` — used when the source palace can't be opened via chromadb (i.e., the corrupt-segment recovery path that this whole module exists for).

Both extract paths now sanitize empty metadata. The `_rebuild_one_collection` line was the smoking gun:

```python
metas.append(meta if meta else {})  # ← the trailing {} was the bug
```

chromadb 1.5.x's `validate_metadata` rejects empty dicts the same as `None`, despite the comment in this code suggesting otherwise.

Verified end-to-end on the same 151,478-drawer palace that originally surfaced #1458. Patch now in active use locally.

igorls:
Hey @jphein, thank you! Is this targeting `main` on purpose ?

jphein:
Oops

jphein:
Second update (force-push) — added the catch-all sanitizer at `backends/chroma.py:add` and `:upsert` after a 151K-drawer rebuild test still failed at ~120K with the same `ValueError` *even with* the repair.py sanitizers in place.

Traceback was:
```
mempalace/backends/chroma.py:add → chromadb Collection.add →
  validate_insert_record_set → validate_metadatas → validate_metadata → ValueError
```

Best guess: chromadb's `upsert()` internally calls `add()` for new-record cases, and the metadatas list gets reprocessed somewhere in that path in a way that re-introduces empty/None entries that the repair-layer sanitizer thought it removed.

The catch-all at the chromadb-client chokepoint guarantees no path can leak bad metadata. Belt-and-suspenders with the repair.py sanitizers, but cheap (one list comprehension per write call) and eliminates the failure mode entirely.

Also retargeted the PR base from `main` → `develop` per @igorls's earlier note.

jphein:
@gemini-code-assist re: the `rebuild_from_sqlite` path — that exact concern is addressed in the second commit on this PR (`5d95656` — catch-all sanitizer in `backends/chroma.py:add`/`upsert`). The `_rebuild_one_collection` line you flagged is also patched directly in repair.py (`_extract_drawers` + `_rebuild_one_collection.metas.append`). The chroma.py catch-all is belt-and-suspenders: even if a future code path bypasses the repair-layer sanitizers, reaching chromadb's client now always passes through the wrapper's sanitizer first.

jphein:
@igorls — properly: thanks for the quick catch on the base branch. Saved a force-push-into-the-wrong-history shape that would have been a pain to untangle. The retarget went through cleanly.

Issue #1128, opened 2026-04-23
feat: add since/before date filter to list_drawers and search


rickysarraf:
The `mempalace_list_drawers` and `mempalace_search` MCP tools currently only filter by `wing` and `room`. There is no way to query drawers by when they were filed or last modified.

ChromaDB's `$where` clause already supports metadata comparisons, so the underlying capability exists — it is just not exposed through the MCP interface.

**Proposed addition:**
- `since` (ISO date string) — include only drawers with `filed_at` ≥ this value
- `before` (ISO date string) — include only drawers with `filed_at` < this value
- Both parameters optional; behaviour unchanged when omitted
- Once `last_modified` is available (see companion issue #1127), the filter should preferably apply to `last_modified` rather than `filed_at`

**Example use case:** "list all drawers updated this week" — a natural query for reviewing recent activity that is currently not expressible.

mvalentsev:
Opened #1891 for the `list_drawers` half of this: optional `since`/`before` bounds on `filed_at` (since inclusive, before exclusive). The filter runs in Python because ChromaDB rejects string operands for `$gte`/`$lt`, and `filed_at` is stored as an ISO string. Date filtering for `mempalace_search` is a separate, older request (#463).

PR #1460, opened 2026-05-11
feat(mcp): add since time-window filter + added_by/filed_at in list_drawers response

Osezno-byte:
## Summary

Two additive extensions to `mempalace_list_drawers`:

1. **`since` kwarg** (ISO-8601 datetime). When set, returns only drawers with `filed_at >= since`. ChromaDB `$gte` only accepts numeric operands, so the filter runs client-side via lexicographic compare on the stored ISO string. `offset` is ignored in this mode (time-window queries want newest-in-window, not pagination) and the response reports `offset=0`.

2. **`added_by` and `filed_at` per drawer** in the response. These metadata fields are already stored by `tool_add_drawer`; this just surfaces them in the list output so consumers don't have to follow up with per-drawer `get_drawer` calls.

## Motivation

Driving use case: an HTTP bridge that delegates work to a remote agent files drawers tagged `added_by=<remote-agent-name>`. A periodic cron polls `list_drawers(wing, room, since=now-5min)` to detect identity spoofing by checking the `added_by` distribution over a recent window. With only `limit/offset` this required paginating until you crossed the time boundary or fetching everything in the wing/room.

Generally useful for any anomaly / provenance / freshness consumer.

## Backward compatibility

Fully backward compatible:
- `since=None` (default) preserves pre-existing behavior — same query path, same response, just two extra keys per drawer.
- The two new per-drawer fields are additive; existing consumers ignore them.

## Tests

5 new tests in `tests/test_mcp_server.py`:
- `test_list_drawers_includes_added_by_and_filed_at`
- `test_list_drawers_since_filters_by_filed_at`
- `test_list_drawers_since_yields_zero_when_window_in_future`
- `test_list_drawers_since_ignores_offset`
- `test_list_drawers_since_combines_with_wing_filter`

Full `test_mcp_server.py` suite passes (113/113) locally on Python 3.12.13.

## Design notes

- Client-side filter when `since` is set: ChromaDB `$gte` rejects string operands ("Expected operand value to be an int or a float for operator \$gte"). Lexicographic compare on ISO-8601 strings is correct as long as producer + consumer share a timezone (the use case is anomaly detection on a single machine over short windows, so this is fine in practice). Switching to a numeric `filed_at_epoch` would be a schema migration — separate PR if desired.
- Offset is ignored when `since` is set rather than honored, since the natural semantics of "drawers since X" is "the set of matching drawers", not "page N of that set".

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1115,8 +1115,27 @@ def tool_get_drawer(drawer_id: str):
         return {"error": str(e)}
 
 
-def tool_list_drawers(wing: str = None, room: str = None, limit: int = 20, offset: int = 0):
-    """List drawers with pagination. Optional wing/room filter."""
+def tool_list_drawers(
+    wing: str = None,
+    room: str = None,
+    limit: int = 20,
+    offset: int = 0,
+    since: str = None,
+):
+    """List drawers with pagination. Optional wing/room filter and time-window.
+
+    `since` is an ISO-8601 datetime string (e.g. "2026-05-11T00:00:00"). When
+    provided, only drawers with ``filed_at >= since`` are returned. ChromaDB
+    `$gte` only accepts numeric operands, so the filter is applied client-side
+    via lexicographic comparison on the stored ISO string. `offset` is ignored
+    when `since` is set — anomaly-detection / time-window queries want
+    newest-in-window, not arbitrary pagination.
+
+    Response shape also includes per-drawer ``added_by`` and ``filed_at``
+    fields, which are needed for provenance / anomaly-detection consumers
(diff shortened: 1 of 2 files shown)


mvalentsev:
For visibility: I opened #1891 for the `list_drawers` date filter (closes #1128) as a separate PR rather than building on this branch, since this one is behind develop and currently conflicting, and it covers `since` only while #1128 asks for both `since` and `before`. #1891 is a fresh rebase covering both. It reaches the same conclusion you noted here about ChromaDB rejecting string operands for `$gte`/`$lt`, so thanks for the earlier exploration.

igorls:

Issue #1468, opened 2026-05-11
create_tunnel does not validate that source/target rooms exist


yonefive71:
## Summary

`palace_graph.create_tunnel` validates that wing/room names are non-empty strings (via `_require_name`) but never confirms that the named rooms actually contain any drawers. Pointing an explicit tunnel at a phantom room — agent fabricates a room name it expects to exist, typo in a slug, scaffolding before mining — silently succeeds and persists.

Combined with #1467 (tunnel-file ignores `palace_path`), this lets an isolated agent both write and read its own bogus tunnel and self-confirm "tunnel created" in completion reports.

## Reproduction

```python
import mempalace.mcp_server as ms
ms.tool_add_drawer("wing_real", "real_room", "content")
# target_room is wrong, but create_tunnel doesn't care:
result = ms.tool_create_tunnel("wing_real", "real_room", "wing_real", "phantom_room")
print(result)   # success, tunnel persisted, pointing at a room with zero drawers
```

## Fix

PR #TBD. For `kind == "explicit"`, `create_tunnel` now queries the chroma collection with `col.get(where={"$and": [{"wing": W}, {"room": R}]}, limit=1, include=[])` for both endpoints. Zero rows → raises `ValueError` naming the offending endpoint.

Three carve-outs preserved:
1. `kind != "explicit"` — topic tunnels use synthetic `topic:<name>` room IDs that don't appear in chroma. Skip.
2. `_get_collection` returns `None` (palace not yet created, tests, transient outage) — skip rather than fail-closed, matches `palace_graph` tolerance elsewhere.
3. Query raises — log and treat as "can't verify, allow" so a flaky index doesn't block legitimate writes.

**Behaviour change:** existing callers that previously created tunnels pointing at empty rooms as scaffolding will now raise. File the drawer first, then create the tunnel — the order the docs have always recommended.

igorls:
Fixed by PR #1469 (merged as ae8074e). `create_tunnel` now validates both endpoints via `_check_room_exists` when `kind=="explicit"`, raising `ValueError` naming the offending wing/room. Topic/entity tunnels skip validation since they can legitimately precede their rooms. Closing.

PR #1469, opened 2026-05-11
fix(graph): tunnel file follows palace_path; validate explicit-tunnel endpoints exist

### LINKED ISSUE #1467: Explicit tunnels stored at hardcoded ~/.mempalace/tunnels.json, ignoring palace_path config
## Summary

`palace_graph._TUNNEL_FILE` is a module-level constant initialised from `os.path.expanduser("~") + "/.mempalace/tunnels.json"`. It ignores `MempalaceConfig.palace_path` (and `MEMPALACE_PALACE_PATH`). Every other piece of palace state — drawers, KG, people map — honours `palace_path`, so when `$HOME` and the configured palace diverge, tunnels write to a different file from drawers.

## Impact

Worst case is the agentic one. A subagent with an isolated `$HOME` (profile sandbox, container, multi-tenant worker) sets `MEMPALACE_PALACE_PATH=/shared/palace` so its drawers go to the shared palace. Its tunnel writes still go to `~/.mempalace/tunnels.json` inside the isolated home — invisible to every other process touching the shared palace. The same isolated `list_tunnels()` reads back the worker's own write, so the worker self-confirms a tunnel that doesn't exist anywhere except its own bubble and reports completion to its orchestrator.

We hit this on a real task: Hephaestus subagent created a "tunnel" pointing at a phantom room, `list_tunnels` confirmed it from the bubble, and the completion report shipped to the parent. The tunnel never existed in the shared palace.

## Reproduction

```python
import os, tempfile
import mempalace.mcp_server as ms
shared = tempfile.mkdtemp()
os.environ["MEMPALACE_PALACE_PATH"] = shared
os.environ["HOME"] = tempfile.mkdtemp()  # simulate profile isolation
ms.tool_add_drawer("wing_test", "room_a", "content a")
ms.tool_add_drawer("wing_test", "room_b", "content b")
ms.tool_create_tunnel("wing_test", "room_a", "wing_test", "room_b")
# The tunnel writes to $HOME/.mempalace/tunnels.json, not to the configured palace.
# Reading tunnels from a fresh process that doesn't share $HOME returns nothing.
```

## Fix

PR #TBD. `_TUNNEL_FILE` becomes `_get_tunnel_file()` deriving the path from a new `MempalaceConfig.tunnel_file` property (sibling of `palace_path`). Default install unchanged. Legacy-file detection emits a one-line `WARNING` when the configured tunnel file is missing but the pre-fix path has one — no auto-migration.

yonefive71:
Fixes #1467 and #1468.

## Bug 1 — Tunnel file ignores `palace_path` config (#1467)

`palace_graph._TUNNEL_FILE` was a module-level constant initialised from `os.path.expanduser("~") + "/.mempalace/tunnels.json"`. Every other piece of palace state honours `MempalaceConfig.palace_path` (and `MEMPALACE_PALACE_PATH`); tunnels did not. Under any setup where `$HOME` and the configured palace diverge — subagent profiles, sandboxes, multi-tenant hosts, container mounts moving the palace to `/srv/` — drawers landed in the configured palace while tunnels silently landed in a separate file no other process could see.

We hit the agentic worst case on a real task: an isolated subagent worker set `MEMPALACE_PALACE_PATH` to the shared palace, called `create_tunnel`, then `list_tunnels` — and got its own write back from the bubble. The worker self-confirmed a tunnel that didn't exist in the shared palace and reported completion to its orchestrator.

**Fix.** Replace the constant with `_get_tunnel_file(config=None)` derived from a new `MempalaceConfig.tunnel_file` property (sibling of `palace_path`). Default single-user installs are unchanged because `~/.mempalace/palace`'s sibling is `~/.mempalace/tunnels.json` — the legacy path.

**Backwards compatibility.** If the configured tunnel file is missing but a file is present at the pre-fix hardcoded path AND the two paths differ, `_load_tunnels` logs a one-line `WARNING` naming both paths and returns `[]`. We deliberately do NOT auto-migrate — silently merging tunnel state across two locations risks clobbering newer data. The user moves or copies the file themselves.

## Bug 2 — `create_tunnel` does not validate endpoints exist (#1468)

`_require_name` only checked that wing/room names were non-empty strings; nothing queried chroma to confirm any drawer carried matching `{wing, room}` metadata. Pointing an explicit tunnel at a phantom room silently succeeded. Combined with Bug 1's read-bubble, an agent could `create_tunnel` → `list_tunnels` and have both calls return its own bogus write.

**Fix.** For `kind == "explicit"`, query `col.get(where={"$and": [{"wing": W}, {"room": R}]}, limit=1, include=[])` for both endpoints before persisting. Zero rows raises `ValueError` naming the offending wing/room pair.

**Carve-outs:**
1. `kind != "explicit"` skips validation — topic tunnels use synthetic `topic:<name>` room ids
2. `_get_collection` → `None` skips validation rather than fail-closed (palace not yet created, transient outage, tests)
3. Query exceptions are logged and treated as "can't verify, allow" so a flaky index never blocks legitimate writes

**Behaviour change.** Callers that previously created tunnels pointing at empty rooms as scaffolding will now raise `ValueError`. File the drawer first, then create the tunnel — this is the order the docs have always recommended.

## Tests

Full suite: **1736 passed, 1 skipped** (no new failures; `-p no:randomly` to expose test-order issues).

- `tests/test_palace_graph_tunnels.py::_use_tmp_tunnel_file` helper now monkeypatches the new resolver functions and also neutralises `_get_collection` so legacy tests don't accidentally trip the new validation path when test-order pollution leaves a real chroma backend bound from another module
- `tests/test_closets.py::TestTunnels` setup/teardown updated for the same reasons
- Three tests in `tests/test_miner.py` that exercise `compute_topic_tunnels` updated to monkeypatch the new resolvers — they didn't need `_get_collection` neutralised because `kind="topic"` short-circuits validation
- New `TestTunnelFileFollowsConfig` covers Bug 1: default path unchanged, custom `palace_path` puts tunnel in the right place, legacy-file warning, no false-positive warning when paths match
- New `TestCreateTunnelEndpointValidation` covers Bug 2: rejects nonexistent target room, rejects nonexistent source room, succeeds when both exist, skips validation when collection unreachable, tolerates query exceptions, topic tunnels bypass validation

## Diff

```
 CHANGELOG.md                       |  14 ++++
 mempalace/config.py                |   6 ++
 mempalace/palace_graph.py          | 102 +++++++++++++++++++++++++-----
 tests/test_closets.py              |  19 ++++--
 tests/test_miner.py                |   9 ++-
 tests/test_palace_graph_tunnels.py | 207 ++++++++++++++++++++++++++++++++++++-
 6 files changed, 333 insertions(+), 24 deletions(-)
```

ruff check + ruff format both clean on touched files.

## Notes for review

- I'm happy to split this into two PRs if you'd rather merge the bugs independently — they're orthogonal. Kept together because the CHANGELOG entry and test-helper changes touch the same files and splitting doubles the rebase pain.
- The legacy-file warning path is intentionally non-fatal and non-migrating. If you'd prefer a fatal `RuntimeError` to force the user to deal with it explicitly, easy change — let me know.
- `_check_room_exists` returns `True` on `col=None` and on query exceptions (allow-on-failure). This matches `_get_collection`'s existing tolerance and the broader pattern in `palace_graph`. If you'd rather fail-closed on a query exception (treating it as evidence the index is degraded enough that a tunnel write would also be unsafe), let me know.


igorls (approved):
Reviewed the code, tests, and CHANGELOG. Both fixes are well-built:

**#1467 (tunnel file follows palace_path)** — `_get_tunnel_file(config=None)` cleanly replaces the module-level constant via a new `MempalaceConfig.tunnel_file` property (sibling of `palace_path`). Default single-user installs are unchanged because the default `palace_path` is still `~/.mempalace/palace` and its sibling `tunnels.json` IS the legacy path. The decision to log a one-line WARNING on legacy-file detection rather than auto-migrate is the right call — silently merging tunnel state across two locations would risk clobbering newer data; surfacing the discrepancy and letting the user move the file preserves agency.

**#1468 (validate explicit endpoints)** — `_check_room_exists(wing, room, col)` with three deliberate carve-outs (`kind != "explicit"`, `col is None`, query exception → fail-open) matches the tolerance pattern already used elsewhere in `palace_graph`. The breaking-change note in CHANGELOG ("File the drawer first, then create the tunnel — this is the order the documentation has always recommended") is exactly the right framing.

Single `MempalaceConfig()` per `create_tunnel` call (addressing the gemini-code-assist note) reduces redundant `mempalace.yaml` reads.

Tested locally after merging develop in: ruff check / format clean, full pytest suite 2113 passed, 3 skipped. CI green on all 5 OS/Python combos + lint.

### Non-blocker notes for follow-up (not asking for changes here)
- `delete_tunnel`, `list_tunnels`, `follow_tunnels` don't thread the `config` instance the way `create_tunnel` does — each re-instantiates `MempalaceConfig()` once via the unthreaded helper calls. Functionally correct, minor perf nit.
- `_check_room_exists` runs before `mine_lock` is held, so technically a drawer could be deleted between the check and the write. Acceptable for the typo/scaffolding cases the bug targets; not a TOCTOU-grade concern.

Closes #1467 and #1468. Merging.

PR #1477, opened 2026-05-12
fix(convo_miner): wrap mine_convos in mine_palace_lock

funguf:
## Summary

The project-files mine path (`miner.mine`) has wrapped `_mine_impl` in `mine_palace_lock` since #1264 — a non-blocking flock that raises `MineAlreadyRunning` so the second runner exits cleanly instead of queueing as a waiter that drives parallel HNSW inserts. The convos mine path (`convo_miner.mine_convos`) was missing the same guard.

In practice this meant any caller that spawned `mempalace mine --mode convos` repeatedly against the same palace — most notably the Stop-hook transcript ingest before the per-target PID slot in `hooks_cli.py` landed — could stack up arbitrarily many concurrent mines, each holding a ChromaDB client open, each writing to the same HNSW index.

Recently observed shape on one machine: 28 stuck convos mines consuming ~18 GB RAM and contributing to a load spike (load average peaked above 100).

## Fix

Refactor `mine_convos` into a thin wrapper that holds the per-palace flock around `_mine_convos_impl`, mirroring `miner.mine` exactly (the body is unchanged). Dry-run skips the lock since it never writes — symmetric with `miner.mine`'s dry-run path.

The CLI already catches `MineAlreadyRunning` for the convos branch at `cli.py:531`, so this change activates an existing exit path without any caller-side update.

## Out of scope (separately tracked)

- Per-source-file `mine_lock` in `palace.py:284` leaks lock files into `~/.mempalace/locks/` (the lock-creating code never `os.unlink`s on release). Observed: 9346 stale `.lock` files on a long-running machine. A simple `unlink` on release is racy against the next acquirer's `open(..., "w") → flock` (two flocks on different inodes pointing to the same path). Wants its own design pass; not included here.

## Test plan

- [x] `uv run pytest tests/test_convo_miner.py tests/test_convo_miner_unit.py tests/test_convo_miner_size_cap.py tests/test_palace_locks.py tests/test_miner.py` — 70 passed
- [x] Two new cross-process tests cover the fix:
  - `test_mine_convos_refuses_concurrent_run_against_same_palace` — child process holds `mine_palace_lock`, parent's `mine_convos` must raise `MineAlreadyRunning`.
  - `test_mine_convos_dry_run_bypasses_palace_lock` — dry-run must complete even while child holds the lock.
- [x] `uv run ruff check .` and `uv run ruff format --check .` clean

The new tests use `multiprocessing.get_context("spawn")` for the same reason `test_palace_locks.py` does — fork with chromadb + onnxruntime threads in the parent deadlocks under Python 3.13.


igorls:
Hey @funguf — pushed a merge commit (`e97b362`) onto this branch to bring it up to date with develop and resolve a couple of conflicts so CI can rerun against the current `ruff==0.15.9` pin.

Summary of the resolution:
- **Imports in `mempalace/convo_miner.py`**: kept both `mine_palace_lock` (this PR) and `prefetch_mined_set` (added on develop).
- **Docstring**: kept your lock-wrapping description, added a one-line pointer to the chunking-config section that now lives inside `_mine_convos_impl`.
- **Real semantic change worth your eyes**: between when this PR was filed and now, develop added `cfg_chunk_size` / `cfg_min_chunk_size` setup inline at the top of `mine_convos`. Your PR factored that body into `_mine_convos_impl`, so a naive merge would have left those variables undefined inside the impl. I moved the `MempalaceConfig()` setup into `_mine_convos_impl` so the variables are in scope where they're actually used. Net behavior is the same; it just lives one function lower.
- **Tests**: both your two new lock tests and develop's `_is_ai_tool_path` / `_resolve_wing` tests are kept side-by-side — no overlap.

Local: `ruff check`, `ruff format --check`, and the full pytest suite (2103 passed, 3 skipped) all pass. Plan is to merge once CI confirms.

Thanks again for the fix — the 28-stuck-mines + 18 GB RAM observation is exactly the failure mode the lock prevents.

igorls (approved):
Tested locally and CI is now green on all 5 OS/Python combos + lint. The lock pattern mirrors `miner.mine` exactly, dry-run correctly bypasses, and the new tests cover both the refuse-concurrent and dry-run-bypasses contracts.

Closes #1518. Merging.

Issue #1125, opened 2026-04-23
Semantic search misses proper-noun drawers when vector similarity favours generic content


cschnatz:
## Summary
`mempalace.searcher.search_memories(query="<first last>")` returns top-ranked results that do NOT contain the query tokens, while drawers with exact "<First Last>" string matches are absent from the top-K — even though those drawers exist in the same collection and survive wing/room filters.

## Reproduction (v3.3.0)
Collection state:
- `drawer_A` — contains "<First Last> — some context..."
- `drawer_B` — contains "...mentioned <First Last>..."
- `drawer_X` — generic preferences text, NO mention of the name

Call: `search_memories(query="<first last>", n_results=10)`

Expected: drawer_A, drawer_B in top-2.
Actual: drawer_X ranks #1-2 (similarity 0.0, distance ~1.43), drawer_A/B absent from top-K. `total_before_filter=30`.

## Root cause
- Chroma vector retrieval over-fetches `n_results * 3` in `search_memories` (`searcher.py` ~line 350), but proper-noun queries produce weak embeddings relative to longer narrative drawers. Name-bearing drawers fall outside the over-fetch window.
- BM25 re-rank in `_hybrid_rank()` operates only over the vector-returned candidates — if the name-bearing drawer never enters the candidate set, BM25 cannot rescue it.
- `max_distance=1.5` (default) further clips proper-noun matches since cosine distance for sparse-token queries tends to drift above 1.3.

## Preferred fix — lexical recall union before rerank

Chroma already exposes a `where_document={"$contains": query}` filter that runs a cheap substring match across the collection. Run it in parallel to the vector query, union the candidate sets by ID (not text), then feed the combined pool into the existing `_hybrid_rank()`:

1. Keep current vector over-fetch as-is.
2. Also call `drawers_col.get(where_document={"$contains": query}, where=where, limit=K)` for K ≈ `n_results * 3`.
3. Union candidates by drawer ID, deduplicate.
4. Pass through `_hybrid_rank()` unchanged.

This guarantees exact-string matches always reach BM25 rerank, without a full-collection scan. Scales well up to ~10k documents; pagination can be added later if needed. Happy to send a PR for this if you confirm the direction and point us at your testing conventions for `searcher`.

## Cheap partial mitigation (minimal change)

Widen vector over-fetch for short queries, clamped for safety:

```python
token_count = len(_TOKEN_RE.findall(query))
overfetch_multiplier = 10 if token_count <= 3 else 3
[...]

PR #1479, opened 2026-05-12
fix(search): tolerate None documents in chunk-hydration loop (#1125)

cschnatz:
## Summary

Closes the chunk-hydration half of #1125. The `_tokenize` site flagged in that issue (line 50 in the reporter's version) has already been guarded upstream; the second site — the drawer-grep enrichment loop's keyword scan inside `search_memories` — was still vulnerable to `AttributeError: 'NoneType' object has no attribute 'lower'`.

## Repro

Sibling drawers in the same `source_file` can have `documents=None` whenever Chroma returns them for entries inserted with embeddings only (legacy mines, partial restores, no-text drawers). When a closet boost surfaces such a source, the enrichment loop iterates `ordered_docs` and crashes at `d.lower()`.

The failure was diagnosed in detail by @MAF33884 in https://github.com/MemPalace/mempalace/issues/1125 (test palace: 482 drawers, single-token queries reliably trigger). Credit to them for narrowing the chunk-hydration site.

## Fix

Two-line guard matching the `_tokenize` idiom already used in this file (`if not text: return []` and `doc or ""` at line 343):

```python
d_lower = (d or "").lower()
# ...
expanded = "\n\n".join(d or "" for d in ordered_docs[start:end])
```

The second guard is necessary because `str.join` raises `TypeError` if any element of the slice is `None` — same root cause propagating one statement further.

## Test

`tests/test_hybrid_search.py::TestChunkHydrationNoneDocument` reproduces the crash:

1. Inserts D1 with text via the public `BaseCollection.upsert` API.
2. Inserts D2 in the same source with embeddings only via the raw `_collection.add` path (no public API for documentless inserts; comment in test notes the reason).
3. Seeds a strong closet for D1's source so the boost fires and `matched_via != "drawer"` triggers the enrichment loop.
4. Asserts `search_memories` returns results without crashing AND that the closet boost fired (anchor against vacuous pass if scoring ever drifts).

Embedding dimension is derived dynamically from `col._collection._embedding_function(["probe"])` so the test won't silently break if MemPalace swaps embedder.

```
$ uv run pytest tests/test_hybrid_search.py -v
6 passed in 5.93s

$ uv run pytest tests/ -v --ignore=tests/benchmarks
1727 passed, 1 skipped, 1 warning in 43.57s
```

Ruff clean.

## Adjacent pre-existing gap (not in this PR)

`source_drawers.documents` itself can be `None` (the whole list, not individual elements) when `include=["documents", "metadatas"]` is called on a collection with no text at all. The `if len(docs) <= 1:` check at line ~932 would raise `TypeError` on that path. That's pre-existing behaviour and sits two lines above this fix; happy to address it as a follow-up if you'd prefer, or leave for a separate PR — let me know.

## Test plan

- [x] `pytest tests/ --ignore=tests/benchmarks` — 1727 passed, 0 fail
- [x] `ruff check .` clean
- [x] `ruff format --check .` clean
- [x] Reproduced crash before fix; test fails RED at `searcher.py:947` `AttributeError`
- [x] Test GREEN after fix; no regression in `test_hybrid_search.py` (5 existing) or `test_searcher.py`

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -944,14 +944,19 @@ def search_memories(
         query_terms = set(_tokenize(query))
         best_idx, best_score = 0, -1
         for idx, d in enumerate(ordered_docs):
-            d_lower = d.lower()
+            # Tolerate ``None`` documents — Chroma returns ``None`` in the
+            # ``documents`` field for drawers inserted with embeddings only
+            # (legacy mines, partial restores, no-text entries). Treat as
+            # empty so the keyword scan continues rather than raising
+            # ``AttributeError`` mid-search (issue #1125).
+            d_lower = (d or "").lower()
             s = sum(1 for t in query_terms if t in d_lower)
             if s > best_score:
                 best_score, best_idx = s, idx
 
         start = max(0, best_idx - 1)
         end = min(len(ordered_docs), best_idx + 2)
-        expanded = "\n\n".join(ordered_docs[start:end])
+        expanded = "\n\n".join(d or "" for d in ordered_docs[start:end])
         if len(expanded) > MAX_HYDRATION_CHARS:
             expanded = (
                 expanded[:MAX_HYDRATION_CHARS]
--- a/tests/test_hybrid_search.py
+++ b/tests/test_hybrid_search.py
@@ -131,3 +131,70 @@ class TestClosetMetadata:
             assert h["matched_via"] == "drawer"
             assert "closet_preview" not in h
             assert h["closet_boost"] == 0.0
+
+
+# ── chunk-hydration robustness (#1125) ───────────────────────────────────
+
+
+class TestChunkHydrationNoneDocument:
+    """Drawer-grep enrichment must tolerate sibling drawers with no document
+    text. Chroma returns ``documents=None`` for entries inserted with only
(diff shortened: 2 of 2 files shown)


MAF33884:
Beautiful turnaround, and thanks for the explicit credit.

- **`_tokenize` already guarded:** confirmed against `develop` HEAD. My report was vs 3.3.3 in the wild; develop has moved ahead. Apologies for the stale intel there.
- **The `str.join` second guard is a great catch.** I didn't call it out — would've crashed the moment the first guard let a `None` through to `ordered_docs[start:end]`. Right to close it in the same patch.
- **On the adjacent `source_drawers.documents is None` gap:** lean toward including in this PR. Same root cause (Chroma's documentless-insert path), one-liner shape (`docs = source_drawers.documents or []`), sits two lines above your fix. Splitting the review cycle for a one-liner with the same conceptual change probably costs more than it saves. That said — if you prefer atomic PRs for clean bisect, separate is fine, no objection.
- **Test reads cleanly.** The dynamic embedding-dim derivation and the closet-boost anchor against vacuous pass are both nicely thought through.

Saw #1480 and #1481 just landed — that's the entire #1125 series. Will leave more focused notes on each. LGTM.

PR #1480, opened 2026-05-12
feat(search): widen vector over-fetch for short queries (#1125)

cschnatz:
## Summary

Cheap mitigation from #1125: widen vector `.query()` over-fetch to **10×** `n_results` for queries with **≤3 tokens**, clamp at 200. Long queries keep the existing 3× window. Restores BM25 rerank's chance to recover proper-noun drawers whose embeddings drop below the default cutoff.

This is **PR B of three** for issue #1125. PR A (#1479) lands the chunk-hydration None-guard the wider candidate pool reliably exposes. This PR (B) is the cheap mitigation. PR C will add the preferred `where_document` lexical union for the residual cases that even 10× cannot reach. Each PR is independently useful; this one is **stacked on #1479** (rebase will collapse cleanly once that merges).

## Repro / evidence

@MAF33884's testing on a 482-drawer palace ([detailed report on #1125](https://github.com/MemPalace/mempalace/issues/1125)) confirmed:

| Query | `total_before_filter` (default) | With this PR | Outcome |
|---|---|---|---|
| 1-token name | 15 | 50 | name-bearing drawer surfaces |
| 3-token short name | 15 | 50 | surfaces |
| 4+ token narrative | unchanged | unchanged | unchanged (3× kept) |

Crashes that the wider pool exposes (None-doc chunk-hydration) are covered by PR #1479. Without that PR, this one would surface more `AttributeError`s, so the stacking order matters.

## Implementation

`mempalace/searcher.py:807-826`. Two additions:

```python
# Module top (matches existing constant style)
_OVERFETCH_MAX = 200

# In search_memories, before drawer .query()
token_count = len(_TOKEN_RE.findall(query))
overfetch_multiplier = 10 if token_count <= 3 else 3
overfetch_n = min(n_results * overfetch_multiplier, _OVERFETCH_MAX)
```

Token count uses the same `_TOKEN_RE` (`\\w{2,}`) that `_tokenize` and the BM25 paths use — drift-free by construction.

## Scope notes

- **Only the vector path widens.** The BM25 union path (`_bm25_only_via_sqlite` at line ~656) keeps `n_results * 3` intentionally — FTS5 ranks by term frequency, so 1-3 token queries already get the relative BM25 advantage regardless of fetch width. Widening BM25 would fetch noise, not signal.
- **CLI `search()` is unchanged** — no rerank stage there, so over-fetching adds latency without benefit.
- Downstream filters (`max_distance`, `_hybrid_rank`, final `[:n_results]` truncation) are per-candidate and order-invariant. Only the candidate *pool* widens; the result count returned to the caller is unchanged.

## Test plan

- [x] `pytest tests/test_searcher.py::TestShortQueryOverfetch -v` — 6 passed
  - single-token → 10×
  - three-token → 10×
  - four-token → 3×
  - 3-vs-4 boundary lock (same `n_results`, off-by-one trap)
  - 0-token (`\"??\"`, whitespace) → routes through wide path (documented)
  - clamp: `n_results=25` with single-token → 200, not 250
- [x] `pytest tests/ --ignore=tests/benchmarks` — 1733 passed, 0 fail, 1 skipped
- [x] `ruff check .` clean
- [x] `ruff format --check .` clean

Tests use `MagicMock` on the collection's `.query()` call to assert the actual `n_results` argument passed through the over-fetch math, avoiding flakiness from real ChromaDB embedding determinism.

## Open questions for maintainer

- Threshold `<= 3` and multiplier `10` are MAF33884's empirically-tuned values; should they be exposed as module constants for future production tuning, or stay inline? Codex review recommended keeping them inline since they're coupled.
- BM25 union path stays at 3× by deliberate scoping. Happy to widen there too in a follow-up if you'd prefer, but the analysis above (FTS5 invariance) suggests it shouldn't help.

Credit: @MAF33884 for the empirical analysis and the precise mitigation shape.

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -31,10 +31,20 @@ class SearchError(Exception):
 
 _TOKEN_RE = re.compile(r"\w{2,}", re.UNICODE)
 
-# Upper bound on the per-query candidate fetch from the vector index.
-# Short queries widen the over-fetch (see issue #1125) up to this cap so
-# the cost of the underlying ChromaDB ``.query()`` and downstream BM25
-# rerank stays bounded on large ``n_results`` requests.
+# Over-fetch tuning for the vector candidate pool (see issue #1125).
+#
+# Short queries — proper nouns and 1–3 token names — produce weak
+# embeddings relative to longer narrative drawers. The default 3× over-
+# fetch can drop name-bearing drawers below the candidate cutoff before
+# BM25 rerank ever sees them. The wider multiplier kicks in for queries
+# with ``len(_tokenize(query)) <= _SHORT_QUERY_TOKEN_THRESHOLD``.
+#
+# ``_OVERFETCH_MAX`` bounds the candidate fetch so latency stays sane on
+# large ``n_results`` requests; ``n_results`` itself acts as a floor so
+# the clamp never returns fewer candidates than the caller asked for.
+_SHORT_QUERY_TOKEN_THRESHOLD = 3
+_SHORT_QUERY_OVERFETCH_MULTIPLIER = 10
+_DEFAULT_OVERFETCH_MULTIPLIER = 3
 _OVERFETCH_MAX = 200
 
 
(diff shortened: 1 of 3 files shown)


MAF33884:
Numbers match my testing exactly (10× for ≤3 tokens, clamp at 200). The `_TOKEN_RE` reuse for token counting keeps everything drift-free against `_tokenize` — appreciated.

- **`<=3` boundary lock test:** nice. The off-by-one trap is the obvious failure mode and the explicit boundary test pins it.
- **On inline-vs-constants:** agree with Codex — keep inline. The threshold and multiplier are empirically coupled (move one, you'd want to revisit the other), and the inline comment block already captures the empirical basis. Promoting to module constants creates the illusion of independent tunables. Easy to promote with proper justification if production tuning ever needs it.
- **BM25 path staying at 3×:** agree. The FTS5 invariance argument is right — widening fetches noise, not signal.
- **0-token routing to wide path:** documented behavior. Whitespace-only queries don't really happen via MCP, but if a downstream caller ever passes `""`, getting the wider pool clamps cleanly and BM25 reranks. Harmless.

LGTM.

cschnatz:
Thanks @gemini-code-assist — both comments addressed in [`572bd1c`](https://github.com/MemPalace/mempalace/pull/1480/commits/572bd1c):

* **HIGH:** floor over-fetch at `n_results` (`max(n_results, min(...))`) so a `n_results > _OVERFETCH_MAX` request never returns fewer than asked. New test `test_overfetch_never_below_n_results` locks the invariant.
* **MEDIUM:** swapped `_TOKEN_RE.findall(query)` for `_tokenize(query)` for consistent `None`/empty handling with the BM25 path.
* **Bonus:** promoted the multiplier and threshold magic numbers to module constants (`_SHORT_QUERY_TOKEN_THRESHOLD`, `_SHORT_QUERY_OVERFETCH_MULTIPLIER`, `_DEFAULT_OVERFETCH_MULTIPLIER`).

Full suite still green: 1734 passed, ruff clean.

PR #1481, opened 2026-05-12
feat(search): add candidate_strategy="contains" for substring union (#1125)

cschnatz:
## Summary

Final piece of issue #1125: opt-in `candidate_strategy="contains"` that runs `drawers_col.get(where_document={"$contains": query}, ...)` in parallel to the vector query and unions the substring-match drawers into the rerank pool. Picks up the proper-noun and exact-string matches whose embeddings sit outside even the widened vector window from PR #1480.

**PR C of three** for #1125. Stacked on:
- PR #1479 — `fix(search): tolerate None documents in chunk-hydration loop` (mandatory; wider candidate pools surface None docs that crashed the enrichment loop)
- PR #1480 — `feat(search): widen vector over-fetch for short queries`

Each PR is independently mergeable. This one is the **preferred** fix from #1125's design discussion.

## Why a new strategy and not always-on

`$contains` is a cheap query but it is still a separate Chroma round-trip per `search_memories` call. Making it always-on would add latency to every search regardless of query shape. Registering it as an opt-in strategy:

- mirrors the existing `candidate_strategy="union"` (BM25-from-sqlite) precedent so the API surface stays consistent
- lets callers who care about proper-noun recall enable it without rewriting the default search path
- lets you (the maintainers) ship it conservatively and flip the default later if the production data justifies it

## Implementation

`mempalace/searcher.py` — `_merge_contains_union_candidates` mirrors `_merge_bm25_union_candidates` line-for-line where the semantics are shared (`max_distance` guard, chunk-precise dedup key, `distance=None` for contains-only adds, defensive try/except + debug log on Chroma failure). The differences are exactly what the strategy advertises:

- the merger reaches into the Chroma collection with `where_document={\"$contains\": query}` and a positive int `limit = min(n_results * 3, _OVERFETCH_MAX)` (no widening for short queries — `$contains` is exact-match, widening adds no recall)
- contains-only hits are tagged `matched_via=\"contains\"` so downstream tooling can distinguish them from `\"drawer\"`, `\"drawer+closet\"`, or `\"bm25_sqlite\"` candidates
- the merger respects the same `wing`/`room` filter the vector path applies (`build_where_filter` is reused; `where_document` and `where` combine inside Chroma)

Registered as `_CANDIDATE_MERGERS[\"contains\"]`; `\"vector\"` and `\"union\"` are unchanged.

### Adjacent fix: `collection_name` forwarding

While threading `collection_name` through `_apply_candidate_strategy` and the new contains merger for multi-collection deployments, `_merge_bm25_union_candidates` also gains the same forwarding (one extra kwarg) so an explicit non-default `collection_name` finally reaches `_bm25_only_via_sqlite`. This was a pre-existing latent gap — surfaced by review of this PR — but is included here because the new strategy framework had to thread the kwarg through anyway. If you'd prefer the BM25-side fix as a separate commit, I can split.

## Test plan

7 mock-based tests in `tests/test_hybrid_candidate_union.py::TestContainsStrategy` (designed mock-first because a small live corpus over-fetches every drawer into the candidate pool and masks which strategy surfaced a result):

- [x] invalid strategy still raises `ValueError` after registry grows
- [x] explicit `\"vector\"` matches default behaviour (no regression)
- [x] vector path misses substring drawer → contains union surfaces it
- [x] `.get` called **exactly once** with `where_document={\"$contains\": q}` and a positive int `limit`
- [x] shared drawer dedupes to one entry across both paths
- [x] wing/room `where` filter is forwarded to the contains `.get()`
- [x] explicit `collection_name` is forwarded into `get_collection`

```
$ uv run pytest tests/ --ignore=tests/benchmarks
1740 passed, 1 skipped, 1 warning in 34.23s

$ uv run ruff check . && uv run ruff format --check .
clean
```

## Open question for maintainer

Worth flipping the default to `\"contains\"` (or a new `\"vector+contains\"`) once your test fleet confirms no regression on narrative queries? Happy to follow up if you want the experiment.

Credit: @MAF33884 for the proposal and the empirical analysis on a 482-drawer palace that motivated this strategy.

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -31,10 +31,20 @@ class SearchError(Exception):
 
 _TOKEN_RE = re.compile(r"\w{2,}", re.UNICODE)
 
-# Upper bound on the per-query candidate fetch from the vector index.
-# Short queries widen the over-fetch (see issue #1125) up to this cap so
-# the cost of the underlying ChromaDB ``.query()`` and downstream BM25
-# rerank stays bounded on large ``n_results`` requests.
+# Over-fetch tuning for the vector candidate pool (see issue #1125).
+#
+# Short queries — proper nouns and 1–3 token names — produce weak
+# embeddings relative to longer narrative drawers. The default 3× over-
+# fetch can drop name-bearing drawers below the candidate cutoff before
+# BM25 rerank ever sees them. The wider multiplier kicks in for queries
+# with ``len(_tokenize(query)) <= _SHORT_QUERY_TOKEN_THRESHOLD``.
+#
+# ``_OVERFETCH_MAX`` bounds the candidate fetch so latency stays sane on
+# large ``n_results`` requests; ``n_results`` itself acts as a floor so
+# the clamp never returns fewer candidates than the caller asked for.
+_SHORT_QUERY_TOKEN_THRESHOLD = 3
+_SHORT_QUERY_OVERFETCH_MULTIPLIER = 10
+_DEFAULT_OVERFETCH_MULTIPLIER = 3
 _OVERFETCH_MAX = 200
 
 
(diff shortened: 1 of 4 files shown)


MAF33884:
The piece that matters most for proper-noun recall. Mirroring `_merge_bm25_union_candidates` line-for-line where the semantics are shared is exactly right — `matched_via` tagging, `max_distance` guard, dedup key all in the right places.

- **On bundling the BM25 `collection_name` forwarding here:** bundle. You're already threading the kwarg through `_apply_candidate_strategy` and the new merger; splitting one of the three call sites into its own PR adds review overhead for no real isolation benefit. Same pattern as the adjacent gap on #1479.
- **On the default flip:** maintainer call, but I can give you empirical signal before any flip. Worth flagging that `vector+contains` would need to be its own registry entry — `_CANDIDATE_MERGERS` is single-strategy lookup, not composable. Two ways I can help: (a) PR a composite `vector+contains` strategy as a follow-up so you have something to benchmark, or (b) re-run my 482-drawer corpus under just `"contains"` to give you a regression read on narrative-query recall. Whichever is more useful — say the word.
- **Mock-first test design is the right call here.** A small live corpus over-fetches every drawer and masks which strategy surfaced a result; the `.get` called-once-with-`where_document` assertion locks the contract precisely.

Thanks for the fast work on the full series. LGTM.

cschnatz:
Thanks @gemini-code-assist — both review items addressed in [`a441274`](https://github.com/MemPalace/mempalace/pull/1481/commits/a441274):

* **`collection_name` forwarding:** dropped the redundant `if/else` since `palace.get_collection` already resolves `None` via `get_configured_collection_name`.
* **DRY `_dedup_key`:** extracted to a module-level `_chunk_precise_dedup_key` helper shared by both BM25-union and contains mergers. Single source of truth for chunk-level dedup semantics.
* **Magic numbers in the over-fetch logic:** addressed in PR #1480 (this PR is stacked on it) — `_SHORT_QUERY_TOKEN_THRESHOLD`, `_SHORT_QUERY_OVERFETCH_MULTIPLIER`, `_DEFAULT_OVERFETCH_MULTIPLIER`, `_OVERFETCH_MAX`.

Full suite still green: 1741 passed, ruff clean.

PR #1014, opened 2026-04-18
refactor(sources): RFC 002 §9 scaffolding — BaseSourceAdapter, registry, PalaceContext

igorls:
## Summary

Lands the read-side plugin contract so third-party source adapters can publish `pip install mempalace-source-<name>` packages against a stable target, matching what [#995](https://github.com/MemPalace/mempalace/pull/995) did for storage backends on the write side.

Sibling to [RFC 001 / #743](https://github.com/MemPalace/mempalace/pull/743) (write) and [RFC 002 / #990](https://github.com/MemPalace/mempalace/pull/990) (spec). Tracking issue: [#989](https://github.com/MemPalace/mempalace/issues/989).

## What's in this PR

- **`mempalace/sources/base.py`** — `BaseSourceAdapter` ABC with kwargs-only `ingest()` / `describe_schema()` and default implementations of `is_current()` / `source_summary()` / `close()` (§1.1–1.2). Typed records: `SourceRef`, `SourceItemMetadata`, `DrawerRecord`, `RouteHint`, `SourceSummary`, `AdapterSchema`, `FieldSpec` (§1.3, §5.2). Error classes (§2.7). Class-level identity contract: `name`, `adapter_version`, `capabilities`, `supported_modes`, `declared_transformations`, `default_privacy_class` (§2.1, §1.4, §1.5, §6).

- **`mempalace/sources/transforms.py`** — reference implementations of the **13 reserved transformations** (§1.4): `utf8_replace_invalid`, `newline_normalize`, `whitespace_trim`, `whitespace_collapse_internal`, `line_trim`, `line_join_spaces`, `blank_line_drop` as pure functions. The six adapter-specific ones (`strip_tool_chrome`, `tool_result_truncate`, `tool_result_omitted`, `spellcheck_user`, `synthesized_marker`, `speaker_role_assignment`) ship as identity shims the conversations adapter will override when migrated. `get_transformation(name)` resolves reserved names.

- **`mempalace/sources/registry.py`** — entry-point discovery via `importlib.metadata.entry_points(group="mempalace.sources")` + `register()` / `unregister()` (§3.1–3.2). `resolve_adapter_for_source()` implements the §3.3 priority order. **Crucially: no auto-detection on the read side** (§3.3 is explicit about that — user intent never inferred from on-disk artifacts).

[...]

PR #297, opened 2026-04-08
feat: add OpenCode auto-plugin for zero-config MemPalace integration

Milofax:
## Summary

- Adds `examples/opencode_auto_plugin.js` — a drop-in OpenCode plugin that auto-initializes MemPalace in every git-tracked project
- Adds `examples/opencode_auto_plugin_setup.md` — full setup guide matching the existing docs style (cf. `gemini_cli_setup.md`, `mcp_setup.md`)

## What the Plugin Does

1. **On first event** of each OpenCode session → detects git root → spawns `mempalace init --yes <root>` in background
2. **After init succeeds** → spawns `mempalace mine --limit 200 <root>` in background
3. **On every chat** → injects the MemPalace Memory Protocol into the system prompt via `experimental.chat.system.transform`

Zero per-project config needed. Drop the file into `~/.config/opencode/plugins/` and restart OpenCode.

## Key Design Decisions

| Decision | Why |
|----------|-----|
| System-prompt injection, NOT AGENTS.md patching | No disk writes, no conflicts with other tools, idempotent |
| Detached + unref spawns | OpenCode never blocks on init/mine |
| Git-only scope | Prevents accidental init in `~`, `/tmp`, etc. |
| In-memory dedup Set | `mempalace init` runs at most once per project per process |
| Log to `/tmp/mempalace-auto.log` | Easy debugging, no sensitive data beyond paths |

## Complementary to PR #268

PR #268 adds the manual `opencode_setup.md` integration guide. This plugin **automates** what that guide describes — users install it once and MemPalace works in every project automatically.

## Verification

- `node --check examples/opencode_auto_plugin.js` ✅
- Smoke-tested: factory → hooks returned, system.transform → protocol injected, dedup → no duplicates ✅
- No Python code modified — tests not affected

PR #23, opened 2026-04-07
feat: add OpenCode SQLite session database support

JakobSachs:
Played around with the project a little, but also wanted to carry over my Opencode sessions. The patch should hopefully be as minimal as possible.

Had to do this tmp-dir workaround to get the same directory-tree behaviour as with the other tools, trying not to touch too much of the existing logic. So it uses the dir-column in the DB to create a tmp-dir structure.
The table:

```
   session table (SQLite)
   ┌──────────┬─────────────────────┐
   │ id       │ directory           │
   ├──────────┼─────────────────────┤
   │ ses_a1b2 │ /home/user/frontend │
   │ ses_c3d4 │ /home/user/frontend │
   │ ses_e5f6 │ /home/user/backend  │
   │ ses_g7h8 │ /home/user/my-infra │
   └──────────┴─────────────────────┘
   ```
   becomes
   ```
   mempalace_oc_*/
   ├── frontend/
   │   ├── frontend_ses_a1b2.txt
   │   └── frontend_ses_c3d4.txt
   ├── backend/
   │   └── backend_ses_e5f6.txt
   └── my_infra/
       └── my_infra_ses_g7h8.txt
 ```
 
 which is then  handled the same way how claude-code/chatgpt/etc. ingestion works. 
 
 ---

Ingest should work via:

```bash
mempalace mine ~/.local/share/opencode/opencode.db --mode convos
```

Tests + Lint should be clean

PR #1484, opened 2026-05-13
feat(sources): OpenCode adapter on RFC 002 contract

jphein:
## Summary

Adds `OpenCodeSourceAdapter` — an RFC 002 `BaseSourceAdapter` that ingests OpenCode AI-coding-CLI session transcripts from OpenCode's local SQLite store (`~/.local/share/opencode/opencode.db`) into the palace, formatted to match `convo_miner`'s exchange-pair drawer shape.

## What's included

- `mempalace/sources/opencode.py` — 482-line adapter:
  - Yields `SourceItemMetadata` then `DrawerRecord`s per session
  - One `source_file` per session shaped as `opencode://<absolute-db-path>#session=<sid>`
  - Chunks via `convo_miner.chunk_exchanges` exchange-pair shape
  - Declares 8 transformations (6 opencode-namespaced + 2 reserved); every name resolves to a reference implementation on `mempalace.sources.transforms` per RFC 002 §7.3
  - Schema fields: `session_id`, `session_title`, `project_dir`, `session_created_at`, `message_count`, `extract_mode`, `opencode_db_path`
  - `supports_incremental` + `adapter_owns_routing`; `default_privacy_class = "pii_potential"`
  - Routes wing from `session.directory` basename (or explicit `options["wing"]`); room from `detect_convo_room`; hall from `_detect_hall_cached`
- `mempalace/sources/transforms.py` — +141 lines adding the 6 opencode-namespaced reference transformations
- `pyproject.toml` — entry-point registration under `[project.entry-points."mempalace.sources"]` as `opencode = "mempalace.sources.opencode:OpenCodeSourceAdapter"`
- `tests/test_sources_opencode.py` — 28 tests covering identity, capabilities, schema, ingest, route hints, transforms round-trip (RFC 002 §7.3 byte-equivalence), edge cases
- `tests/fixtures/opencode/sample_session_2026_05_12/build_fixture.py` — SQLite-schema-verbatim fixture builder (mirrors opencode-ai 1.14.39's schema captured live from `~/.local/share/opencode/opencode.db`); no recorded `.db` ships because real-session content is unsanitizable

## Coordination

This work originated from the OpenCode SQLite spadework in @JakobSachs's #23 (DB-schema reverse engineering, session/message/part traversal, tool-input/tool-output stripping primitives) — rebuilt on the RFC 002 contract so OpenCode support can ship as a registered adapter rather than a `normalize.py` branch. Coordination thread at #23, comment posted offering three paths forward; pushing this PR as the working code per @JakobSachs's choice of how to land (see #23#issuecomment-4436116396). Co-authored credit on the commit.

Complementary to #297 (Milofax) which adds the **consumer-side** JS plugin (`examples/opencode_auto_plugin.js`) that auto-initializes MemPalace in OpenCode sessions. Together: this adapter ingests OpenCode sessions into the palace; #297's plugin queries the palace from OpenCode at runtime. Different layers, no overlap.

Built on top of #1014 (RFC 002 scaffolding by @igorls).

## Test plan

- [x] 28/28 OpenCode adapter tests pass in 1.53s
- [x] Full suite 1876 passed / 7 skipped / 106 deselected — zero regressions (+22 net vs pre-adapter baseline)
- [x] Adapter conformance via RFC 002 §7.3 declared-transformation round-trip (drawer content reproducible from transformations)
- [ ] Real-OpenCode-session smoke (deferred — fixture is schema-verbatim from a real install but doesn't ship recorded content; configuring an OpenCode provider for a billable real-session capture is non-blocking)

## Notes

- The 7 fork-only-on-jphein/main files (CLAUDE.md, README.md, FORK_CHANGELOG.md, docs/fork-changes.yaml, .claude-plugin/*, docs/internal/*, docs/research/*) are intentionally NOT in this diff — branch was rebuilt off `upstream/develop` carrying only commit `9030f06` so no fork-specific narrative leaks.

### DIFF
--- a/mempalace/sources/context.py
+++ b/mempalace/sources/context.py
@@ -114,6 +114,16 @@ class PalaceContext:
         advancing past the item."""
         self._skip_requested = True
 
+    def is_skip_requested(self) -> bool:
+        """Return whether :meth:`skip_current_item` was called since the last
+        time core advanced past an item. Adapters check this between the
+        ``SourceItemMetadata`` yield and the cost of processing the item — if
+        core has signaled skip (because :meth:`is_current` returned True), the
+        adapter can bail out of expensive work (SQL queries, chunking, etc.)
+        rather than waiting for core to drop drawers downstream. Core resets
+        the flag on its own; adapters MUST NOT clear it."""
+        return self._skip_requested
+
     def emit(self, event: str, **details: Any) -> None:
         """Invoke each registered progress hook with ``(event, **details)``."""
         for hook in self.progress_hooks:
--- a/mempalace/sources/opencode.py
+++ b/mempalace/sources/opencode.py
@@ -20,7 +20,6 @@ branch.
 
 from __future__ import annotations
 
-import json
 import logging
 import sqlite3
 from datetime import datetime, timezone
@@ -307,8 +306,7 @@ class OpenCodeSourceAdapter(BaseSourceAdapter):
                     size_hint=None,
                     route_hint=self._route_hint_for(directory, ""),
                 )
-                if palace._skip_requested:
-                    palace._skip_requested = False
+                if palace.is_skip_requested():
                     continue
 
                 messages = _extract_session_messages(conn, sid)
@@ -334,6 +332,15 @@ class OpenCodeSourceAdapter(BaseSourceAdapter):
                 wing = self._wing_for(source, directory)
                 room = detect_convo_room(transcript)
                 created_iso = _utc_iso(time_created or 0)
+                # Pre-compute once per session so every chunk of the same
+                # session shares the same filed_at timestamp.
+                filed_at = (
+                    datetime.now(timezone.utc)
+                    .replace(microsecond=0)
+                    .isoformat()
+                    .replace("+00:00", "Z")
+                )
+                session_version = str(time_updated or time_created or 0)
                 for chunk in chunks:
                     content = chunk["content"]
                     chunk_index = int(chunk["chunk_index"])
@@ -341,10 +348,7 @@ class OpenCodeSourceAdapter(BaseSourceAdapter):
                         # Universal §5.1 fields
                         "source_file": src_file,
                         "chunk_index": chunk_index,
-                        "filed_at": datetime.now(timezone.utc)
-                        .replace(microsecond=0)
-                        .isoformat()
-                        .replace("+00:00", "Z"),
+                        "filed_at": filed_at,
                         "added_by": "opencode-adapter",
                         "wing": wing,
                         "room": room,
@@ -359,6 +363,9 @@ class OpenCodeSourceAdapter(BaseSourceAdapter):
                         "session_created_at": created_iso,
                         "message_count": len(messages),
                         "opencode_db_path": db_path,
+                        # Required by is_current() for incremental ingest;
+                        # mirrors the SourceItemMetadata.version yielded above.
+                        "opencode_session_version": session_version,
                     }
                     yield DrawerRecord(
                         content=content,
--- a/mempalace/sources/opencode.py
+++ b/mempalace/sources/opencode.py
@@ -88,9 +88,7 @@ def _resolve_db(local_path: Optional[str] = None) -> str:
     )
 
 
-def _build_source_bytes_per_session(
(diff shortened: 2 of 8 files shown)


jphein:
Addressed all 4 items from @gemini-code-assist's review in commit faee2b9:

1. ✅ Added `opencode_session_version` to the chunk metadata so `is_current()` can do real version comparisons (was always falling back to "exists → current")
2. ✅ Removed direct access to `PalaceContext._skip_requested`; added a small companion change — `PalaceContext.is_skip_requested()` public method — so adapters can short-circuit without touching the private flag. The spec's "core checks between yields" pattern doesn't hold for Python generators (the adapter's code runs between yields, not core's), so adapters need a way to query the flag. Read-only access only; core still owns the reset
3. ✅ Hoisted `filed_at` (and `session_version`) outside the chunk loop — every chunk of the same session now shares the same `filed_at` timestamp
4. ✅ Moved `import json as _json` to the top of `transforms.py` per PEP 8

Bonus: removed an unused `import json` from `opencode.py` that ruff caught.

Tests: 57 pass (28 opencode + 29 base sources); ruff clean.

igorls:
Nice adapter — solid RFC 002 conformance, good test coverage, and the fixture-builder approach (no recorded `.db`) is the right privacy call.

Two things to address before merge:

**1. CI lint is red** — two ruff violations in `tests/test_sources_opencode.py`:

```
tests/test_sources_opencode.py:17:8: F401 `os` imported but unused
tests/test_sources_opencode.py:43:1: E402 Module level import not at top of file
```

The E402 comes from the `sys.path.insert(0, ...)` / `import build_fixture` pattern. Consider `importlib.util.spec_from_file_location` + `module_from_spec` to load the fixture builder without mutating `sys.path` at module scope (which also fixes the E402).

**2. Route-hint wing mismatch** — `_route_hint_for()` (called at the `SourceItemMetadata` stage) computes the wing purely from `directory`, but `_wing_for()` (called at the `DrawerRecord` stage) respects `source.options["wing"]` first (RFC 002 §2.5 precedence). When a user passes `options={"wing": "Custom Wing"}`, the metadata hint says `"frontend"` while the actual drawers say `"custom_wing"`. This could cause core to make wrong skip/routing decisions.

Fix: have `_route_hint_for` accept the `source` parameter and delegate to `_wing_for`, or inline the same precedence logic.

Minor (non-blocking): the `AuthRequiredError` import on line 62 has a `# noqa: F401` claiming re-export, but nothing in `__all__` or the module surface actually re-exports it — the suppression is unjustified.

jphein:
Thanks @igorls — addressed all four in commit 13353d9:

1. ✅ Dropped unused `os` import (F401)
2. ✅ Refactored the fixture builder loader to `importlib.util.spec_from_file_location` + `module_from_spec` per your suggestion. No more `sys.path` mutation at module scope; E402 gone. Also registers the loaded module in `sys.modules` so `dataclasses`/typing introspection inside `build_fixture` can resolve `cls.__module__` (without that registration, dataclass collection fails at test-collection time — found out the hard way).
3. ✅ `_route_hint_for(source, directory)` now delegates to `_wing_for` so both stages apply identical RFC 002 §2.5 precedence. The metadata hint and drawer wing will agree when callers pass `options={"wing": ...}`.
4. ✅ Dropped the `AuthRequiredError` import + unjustified `# noqa: F401` — `__all__` was never going to re-export it.

Tests: 57 pass (28 opencode + 29 base sources); ruff clean on all touched files.

jphein:
Hi @igorls — friendly nudge on this one when you have a moment. I addressed everything from your review in `13353d9`:

- Dropped the unused `os` import (F401) and refactored the fixture-builder loader to `importlib.util.spec_from_file_location` + `module_from_spec` — no more `sys.path` mutation at module scope, E402 gone.
- `_route_hint_for(source, directory)` now delegates to `_wing_for`, so the metadata hint and the drawer wing apply identical RFC 002 §2.5 precedence when callers pass `options={"wing": ...}`.
- Removed the unjustified `AuthRequiredError` `# noqa: F401`.

57 tests pass, ruff clean, and it still merges cleanly on current `develop`. I've had a few PRs land since (e.g. #1769 into develop on the 22nd), so I'm around to iterate quickly if anything's drifted. Would love another look.

PR #1502, opened 2026-05-13
Harden local WAL and tunnel file permissions

AbubakarMahmood:
## Summary
- Re-apply best-effort 0700/0600 permissions to existing WAL storage before and after appends.
- Re-apply best-effort 0700/0600 permissions to explicit tunnel storage, including the temporary file before atomic replace.
- Add mock-based regression coverage so the hardening is testable on Windows and POSIX without relying on real mode-bit behavior.

## Testing
- `python -m pytest tests\test_palace_graph_tunnels.py tests\test_mcp_server.py::TestWalPermissions`
- `python -m py_compile mempalace\mcp_server.py mempalace\palace_graph.py`


AbubakarMahmood:
Addressed Gemini's redundancy feedback in d622c2a by removing the post-write/post-replace hardening calls while keeping the pre-write/pre-replace best-effort checks. Local verification: 
uff check mempalace\\mcp_server.py mempalace\\palace_graph.py tests\\test_mcp_server.py tests\\test_palace_graph_tunnels.py, python -m pytest tests\\test_palace_graph_tunnels.py tests\\test_mcp_server.py::TestWalPermissions, and python -m py_compile mempalace\\mcp_server.py mempalace\\palace_graph.py all pass. A full 	ests\\test_mcp_server.py run was blocked locally by Chroma attempting to download the 79.3M ONNX model into a temp test cache, so I kept the local gate to the affected permission-hardening coverage plus lint/compile.

AbubakarMahmood:
Addressed the Copilot temp-file permission review in 76f08cd. _save_tunnels now creates tunnels.json.tmp with os.open(..., 0o600) before writing JSON, then wraps it with os.fdopen, so a permissive umask cannot make the temp file world-readable during the write window. Re-ran: ruff check mempalace/palace_graph.py tests/test_palace_graph_tunnels.py; python -m pytest tests/test_palace_graph_tunnels.py; python -m py_compile mempalace/palace_graph.py.

AbubakarMahmood:
Addressed the latest Copilot permission-hardening comments in 27a6bbd. For tunnel writes, _save_tunnels now re-applies 0o600 to the opened temp fd before any JSON is written, with a path chmod fallback, so stale tunnels.json.tmp files with broad permissions cannot be truncated and promoted unchanged. For WAL writes, _wal_log again hardens after the append as well as before it, matching the PR guarantee. Re-ran: ruff check mempalace/mcp_server.py mempalace/palace_graph.py tests/test_mcp_server.py tests/test_palace_graph_tunnels.py; python -m pytest tests/test_palace_graph_tunnels.py tests/test_mcp_server.py::TestWalPermissions; python -m py_compile mempalace/mcp_server.py mempalace/palace_graph.py.

igorls:
Thanks for the iteration on this — the atomic temp-file fix in `_save_tunnels` (open with `O_CREAT|O_TRUNC` + mode `0o600`, `fchmod` before `json.dump`) is the strongest piece here and closes a real umask-leak window.

One concrete suggestion on the WAL side:

**Drop the post-append `_ensure_secure_wal_permissions()` in `_wal_log`.** The pre-append call already covers drift, and `os.open(..., O_CREAT|O_WRONLY|O_APPEND, 0o600)` already constrains the mode of any freshly-created file. The trailing call only defends an extremely narrow race (file recreated mid-write by another actor) and doubles the per-write syscall cost.

Rough numbers: each `_ensure_secure_wal_permissions()` is 2 `chmod` syscalls (dir + file). Keeping both means 4 extra syscalls per WAL append — ~20–200 µs on local SSD, but multiplied across bulk paths like `miner.py` / `convo_miner.py` that emit thousands of `add_drawer` ops, and noticeably worse if `~/.mempalace` ever sits on a network FS (NFS round-trips × 4 per write). Halving it costs nothing in real security.

Smaller notes (non-blocking):
- At module init, `_ensure_secure_wal_permissions()` is called once *before* the WAL file is created (no-op on the file branch) and once after. The pre-create call can go.
- `_ensure_secure_wal_permissions` and `_ensure_secure_tunnel_permissions` are structurally identical; a shared `_harden(path, dir_mode, file_mode)` helper would be tidier, but fine to leave since they live in different modules.
- If you drop the post-append call, `test_wal_log_hardens_before_and_after_append` needs to assert one call instead of two.

Risk on the data path is low — every `chmod`/`fchmod` is best-effort and degrades to existing behavior on failure.

AbubakarMahmood:
Thanks, addressed in 2e0360b. I removed the post-append _ensure_secure_wal_permissions() call from _wal_log, dropped the pre-create init hardening call, and updated the WAL test to assert the single pre-append hardening call. I kept the _save_tunnels temp-file path as-is with os.open(..., 0o600) plus fchmod before json.dump. Local focused checks passed: ruff check, ruff format --check, tests/test_mcp_server.py::TestWalPermissions, and tests/test_palace_graph_tunnels.py.

AbubakarMahmood:
Hi maintainers — GitHub is currently holding this fork PR's `Tests` and `Docker` workflows for manual approval (`action_required`, with no jobs started). When convenient, could someone approve the pending runs so the Linux/macOS checks can execute?

The current head is `784bea5`, the PR is mergeable, and the focused local gate passes (`ruff`; 62 tests passed, with only the expected POSIX-permission test skipped on Windows). Thanks.

Issue #1510, opened 2026-05-15
test: test_hooks_cli.py tests fail in isolation — hidden test-order dependency on test_cli.py creating ~/.mempalace


messelink:
## Summary

Nine tests in `tests/test_hooks_cli.py` fail when the file is run in isolation. They pass in CI and when running the full suite because an earlier test file (alphabetically: `test_cli.py`) creates `~/.mempalace` in the session tmpdir as a side effect, after which the `_palace_root_exists()` kill-switch added in `8472d55` returns True for subsequent tests.

The 9 affected tests don't monkeypatch `PALACE_ROOT` the way the newer kill-switch-specific tests (`test_hook_stop_does_not_create_palace_dir_when_absent`, etc.) do via `_redirect_palace_root`.

## Reproducer

```
$ python -m pytest tests/test_hooks_cli.py
9 failed, 80 passed, 1 skipped

$ python -m pytest tests/test_cli.py tests/test_hooks_cli.py
all passing (150+)

$ python -m pytest tests/   # full suite, as CI runs
all passing
```

## Failing-in-isolation tests

- `test_stop_hook_saves_silently_at_interval`
- `test_stop_hook_derives_wing_from_transcript_path`
- `test_stop_hook_tracks_save_point`
- `test_log_writes_to_hook_log`
- `test_stop_hook_oserror_on_last_save_read`
- `test_stop_hook_oserror_on_write`
- `test_precompact_with_mempal_dir`
- `test_precompact_mines_transcript_dir`
- `test_stop_hook_rejects_injected_stop_hook_active`

All share the failure shape: `_save_diary_direct` mock was never called / hook short-circuited to `{}` output / `STATE_DIR.mkdir()` not reached, because `_palace_root_exists()` returned False against the unmocked module-level `PALACE_ROOT = Path.home() / ".mempalace"`.

## Why CI is green anyway

Tests run in alphabetical file order. Some test in `test_cli.py` (before `test_hooks_cli.py`) creates `~/.mempalace` in the per-session `HOME=tempfile.mkdtemp(prefix="mempalace_session_")` directory set by `conftest.py`. By the time the hook tests run, the directory exists, the kill-switch is satisfied, and the tests pass — but only because of test-file ordering, not because the tests are robust on their own.

## Suggested fix

Either:

[...]

PR #1511, opened 2026-05-15
feat(config): MEMPALACE_WING env var overrides auto-derived session wing

messelink:
## Summary

Adds a `MEMPALACE_WING` environment variable that lets a Claude Code session declare its wing explicitly, overriding whatever wing would otherwise be derived from the transcript path (Stop hook) or directory name (convo miner).

Layers cleanly on top of #1424 — that PR made `_wing_from_transcript_path` robust by reading `cwd` from the JSONL transcript (the canonical answer for most cases). This PR adds an explicit user-controlled escape hatch for the cases where even the JSONL-cwd answer isn't what the user wants (cross-project sessions, sessions launched from a workspace dir that isn't the canonical project root, single-session ad-hoc filing, etc.).

## Precedence

```
explicit --wing CLI arg
  > MEMPALACE_WING env var
    > _wing_from_transcript_path / normalize_wing_name(dir) defaults
```

Explicit CLI flags still win. The env var is a fallback that beats the auto-derivation.

## Usage

```sh
export MEMPALACE_WING=mempalace
claude   # or codex, or wherever you launch from
```

All diary writes from the Stop hook and all auto-mined chunks from `convo_miner` for this shell land under `wing=mempalace`. Unset the env var and the auto-derivation behavior comes back unchanged.

## Implementation

New helper in `mempalace.config`:

```python
def session_wing_override() -> str | None:
    """Return MEMPALACE_WING env var (normalized) if set, else None."""
    env = os.environ.get(\"MEMPALACE_WING\") or os.environ.get(\"MEMPAL_WING\") or \"\"
    env = env.strip()
    if not env:
        return None
    return normalize_wing_name(env)
```

Wired into two call sites with one-line guards. Empty / whitespace-only env values are treated as unset (no `\"\"`-named wings). `MEMPAL_WING` (legacy form) is accepted as a fallback to mirror the established `MEMPAL_DIR` / `MEMPAL_PYTHON` convention; `MEMPALACE_WING` wins when both are set.

## Tests

**6 new unit tests in `test_config.py`** covering:

- Unset → `None`
- Empty string → `None`
- Whitespace-only → `None`
- Normalization (`\"My-Cool App\"` → `\"my_cool_app\"`)
- Legacy `MEMPAL_WING` fallback
- `MEMPALACE_WING` takes precedence when both are set

**1 new integration test in `test_hooks_cli.py`** (`test_stop_hook_mempalace_wing_env_overrides_path_derivation`) verifies the override beats path-derivation end-to-end through `hook_stop`. Uses `monkeypatch.setattr(hooks_cli_mod, \"PALACE_ROOT\", ...)` with a pre-created directory so the test passes in isolation.

All 64 `test_config.py` tests pass. All 21 `test_convo_miner_*` tests pass (no regressions to default behavior). The new integration test passes both in isolation and as part of the full suite.

## Test-order note (out of scope, filed separately)

While writing the integration test I discovered that 9 existing tests in `test_hooks_cli.py` fail when the file is run in isolation but pass in CI because of test-file ordering side effects. Filed as #1510 — not blocking this PR. The new integration test in this PR uses explicit `PALACE_ROOT` monkeypatching to be robust against that ordering issue, so it stays green regardless of how the file is run.

## Related

- #1424 — `fix(hooks): correct wing extraction for hyphenated project names`. Merged. This PR layers on top: their work made auto-derivation robust; this PR adds an override for when the user wants to bypass auto-derivation entirely.
- #1410 — original issue about the encoded-path slicing bug. Addressed by #1424; this PR provides a manual escape hatch.


messelink:
Pushed `fb37663` addressing the bot's feedback and the CI failure.

**Python 3.9 compatibility:** the `str | None` syntax in `session_wing_override()`'s signature is PEP 604 — works at module-eval time only on Python 3.10+. CI's 3.9 job failed at `from mempalace.config import MempalaceConfig` with `TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'`. Fixed by adding `from __future__ import annotations` to the top of `config.py`, matching the codebase convention used by `palace_graph.py`, `entity_detector.py`, and ~8 other files.

**Security validation (`gemini-code-assist`):** `session_wing_override()` now validates the raw env value through `sanitize_name` before normalization. Path traversal segments, path separators, characters outside the wing/room slug charset, and over-length input are rejected; invalid values disable the override silently (`return None` → caller falls back to auto-derivation) rather than raising, so a misconfigured env var doesn't crash every hook fire.

**Lint failure:** my local ruff was 0.15.x but CI pins `ruff>=0.4.0,<0.5`. Reformatted `tests/test_hooks_cli.py` under ruff 0.4.10 to match CI.

**Verified locally:** ran the full CI command (`pytest tests/ --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10`) in a fresh Python 3.9 venv via `uv` — **1762 passed, 1 skipped, 85.02% coverage** (above the 80% threshold). 5 new unit tests added for the security rejections, bringing total to 11 `session_wing_override` tests.

**On the inline suggestion's policy:** gemini-code-assist suggested raising `ValueError` on invalid input. I chose silent fallback instead because a raise propagates up through `hook_stop` and breaks the Stop-hook contract on every invocation — a misconfigured env var should disable the override, not the entire hook. Happy to change if a maintainer prefers a different policy.

messelink:
Rebased onto current `develop` after 5+ weeks of upstream activity. Three commits:

1. `feat(config)` — original feature, conflict-resolved against upstream's new `_resolve_wing()` helper. The `MEMPALACE_WING` env var now slots between the explicit `wing` argument and AI-tool-path detection in `_resolve_wing`, matching the original PR's documented precedence.
2. `test(hooks)` — original test fixture refactor, unchanged.
3. `test: rebase adaptation` — small mock-assertion update for `_save_diary_direct`'s new `agent_name` kwarg that landed during the gap.

### Context — how this fits with adjacent work

Acknowledging the cluster of wing-resolution work since this PR was first filed:

- **#1861** describes the canonical-vs-`wing_<project>` fragmentation we're seeing in production today (16 twin wings, ~5k shadowed drawers in the reporter's palace; same pattern in our 30-wing palace). `MEMPALACE_WING` doesn't fix the root cause (the two emission paths disagree on naming convention) but it does give users a per-shell workaround: setting `MEMPALACE_WING=<canonical>` in a session forces *both* the Stop hook AND `convo_miner` to emit under the same wing, so a single session's output stops fragmenting.

- **#1594** (`wing_aliases.json`) is a complementary, larger-surface declarative approach. The two are different scopes — env var = ephemeral per-shell override; aliases = persistent global map — and could coexist cleanly with env var taking highest precedence.

The use case is real on our side: persistent Claude Code sessions across two machines for days/weeks, multi-screen workflows where the encoded transcript folder name doesn't reflect the canonical wing. Full env-cleared suite passing locally. Would appreciate a review when bandwidth allows.

messelink:
Concrete usage example for reviewers wanting to see this in practice: small `ccs` shell launcher that wraps `screen` + `claude --remote-control` and auto-exports `MEMPALACE_WING="<session>"` before launching — sanitized Gist: https://gist.github.com/messelink/89b757a904966bf86fc3a89543cdca5f

Respects a pre-set `MEMPALACE_WING` so a `direnv` `.envrc` per project composes cleanly. On our two-machine fleet it's stopped the fragmentation between Stop-hook diary writes and convo-miner output going forward (existing fragments still need a separate cleanup pass — also commented on #1861 since that's where the cluster lives).

messelink:
Rebased onto current `develop`. One conflict, in `hooks_cli.py`: the hook-settings read (`silent` / `toast`) landed in the same function body this PR touches. Resolved by keeping both — settings block first, then the `MEMPALACE_WING` override applied at the `project_wing` assignment. Full suite green locally (3413 passed, 31 skipped).

mclang:
Merging this would help with my problem with MemPalace, so I hope reviewers can check this out soon 🙏

messelink:
Rebased onto current `develop`, now at `41ed817`. Clean this time, no conflicts, and CI is green across all twelve checks.

Context on the previous push: it failed only on `test-linux (3.11)`, on `test_probes_keep_chroma_writer_shared_lock`. We reproduced that on unmodified `develop` at about 1 in 50 runs, so it wasn't from this branch. #2512 has since fixed it, and after picking that up the test passed 30 out of 30 locally on 3.11.

On the change itself: the earlier version of this PR added `from __future__ import annotations` to `config.py` for Python 3.9 compatibility, which is what kept conflicting with upstream's import block. We've dropped it and quoted the return annotation instead (`-> "str | None"`), matching `connect_sqlite_read` in the same file. The PR no longer touches the imports at all.

One thing to flag for whoever reviews this next to #2259: that PR has the hooks pass a `--wing` derived from `cwd`, which is priority 1 in `_resolve_wing` and would beat `MEMPALACE_WING` for mined conversations. Here the override wins in the Stop-hook diary path. If both land, a session with `MEMPALACE_WING` set could see its diary and its mined transcripts go to different wings. Happy to adjust the ordering here if there's a preference.

@mclang thanks for chiming in, good to know it would help someone else too.

Issue #1351, opened 2026-05-04
MCP error -32000 "Internal tool error" on parameter-name mismatch hides which param is wrong


fkberthold:
## Summary

When an MCP tool exposed by the MemPalace plugin is called with a missing required
parameter or with a wrongly-named parameter, the server returns:

    MCP error -32000: Internal tool error

with no further detail naming the offending parameter. The expected behavior is
a structured error message identifying which parameter was missing or misnamed
(e.g. `Missing required parameter 'entry' for tool mempalace_diary_write`).
Most MCP servers do this; this one doesn't, so the failure is only diagnosable
by independently fetching the tool schema.

## Reproduction

Using `mempalace_diary_write` (3.3.0):

1. Call with `agent_name="claude-opus"` and `content="..."` (omitting the required
   `entry` parameter and substituting a wrongly-named one).
2. Observe `MCP error -32000: Internal tool error` with no further information.
3. Confirm via the tool's schema that the required param is `entry`, not `content`.

The same opacity affects retries: trying the same call again gives the same
opaque error, with no signal that the problem is param-shape rather than a
transient failure.

## Expected

A structured error message naming the offending parameter, e.g.:

    Missing required parameter 'entry' for tool mempalace_diary_write
    Unknown parameter 'content' for tool mempalace_diary_write

## Impact

This is a discoverability bug, not a data-integrity bug. Workarounds exist
(fetch the schema, fix the call), but the failure mode is silent enough that
agents — including LLM agents acting on the tools — repeatedly retry the
same wrong call shape before realizing the schema needs to be checked. In one
session this caused three back-to-back failed `mempalace_diary_write` calls
before the schema mismatch was caught.

## Environment

- Plugin version: 3.3.0
- MCP transport: stdio
- Client: Claude Code
- Reproduces consistently against `mempalace_diary_write`; other MemPalace
  tools (`mempalace_add_drawer`, `mempalace_kg_add`, etc.) likely share the
  same error-mapping path but were not specifically tested.

anastasiiaanfimova:
Independent confirmation — same opaque `-32000` triggered by agents (LLM and otherwise) repeatedly retrying the same wrong-shape call.

Patched this locally in our 3.3.1 deployment about a week ago. Minimal change: catch the dispatcher exception and re-emit with type + truncated message:

```python
"error": {
    "code": -32000,
    "message": f"Internal tool error in {method}: {type(e).__name__}: {str(e)[:300]}"
}
```

[...]

PR #1513, opened 2026-05-15
fix(mcp): structured -32602 for unknown parameter name (#1512)

### LINKED ISSUE #1512: MCP: wrong parameter name still hidden — schema-filter silently drops unknown kwargs (second half of #1351)
Follow-up to #1351 / #1500. #1500 (merged) fixed the **missing-required** half — a missing required arg now returns `-32602 Missing required parameter 'X' for tool ...` instead of `-32000`. The **wrong-name** half of the original #1351 reproducer is still unaddressed, and #1351 auto-closed as `COMPLETED` on merge, so this half is currently untracked.

### Reproducer

```
echo '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"mempalace_add_drawer","arguments":{"wing":"w","room":"r","text":"hello"}}}' \
  | uv run python -m mempalace.mcp_server
```

`mempalace_add_drawer` takes `content=`, not `text=`. Post-#1500 the response is `-32602 Missing required parameter 'content' for tool mempalace_add_drawer` — workable, but it points at the *missing* param, not the *wrong-named* one actually sent (`text`). An agent that believes it sent content (under the wrong name) retries the same wrong-shape call.

### Root cause

`mempalace/mcp_server.py:2245` silently drops any kwarg not in the tool's `input_schema.properties` before dispatch:

```python
if not accepts_var_keyword:
    tool_args = {k: v for k, v in tool_args.items() if k in schema_props}
```

So `text=` never reaches the handler; the wrong *name* surfaces only indirectly as the now-missing `content`.

### Proposed fix

Symmetric with #1500: emit a structured `-32602 Unknown parameter 'text' for tool mempalace_add_drawer` from that same filter (single point, covers every tool at once), instead of silently dropping. Gated by the same `if not accepts_var_keyword:` so handlers that explicitly accept `**kwargs` keep current pass-through behaviour.

**Subtlety:** `wait_for_previous` is an internal transport kwarg, in no tool's schema, popped at `mempalace/mcp_server.py:2263` *after* this filter. An unknown-name check at line 2245 must exclude `wait_for_previous` (e.g. `k != "wait_for_previous"`) or it would be falsely flagged for `**kwargs` pass-through handlers. (Confirmed in the #1351 thread by @mvalentsev.)

No conflict with the filter's spoof-prevention intent: declared params (`added_by`, `source_file`, …) are in `schema_props` and pass; only genuinely-unknown names are flagged.

Happy to PR — I flagged this as a follow-up in the #1351 thread; deferred until #1500 merged since both touch the same dispatch region. #1500 is now merged.

anastasiiaanfimova:
Second half of #1351, complementary to #1500. #1500 made the *missing-required* case structured; this does the same for the *wrong-name* case. Issue opened first (#1512) per CONTRIBUTING.

Closes #1512.

## What

A kwarg not in the tool's `input_schema.properties` (e.g. `mempalace_add_drawer(text=...)` instead of `content=...`) was silently dropped by the schema-filter, so the cause surfaced only indirectly as a later `Missing required 'content'` (post-#1500). The filter now emits `-32602 Unknown parameter 'text' for tool mempalace_add_drawer` — single point, all tools, symmetric with #1500's message shape, `-32602` code and `logger.debug` precedent.

- Gated by the existing `if not accepts_var_keyword:` so `**kwargs` pass-through handlers (#684) keep accepting unknown kwargs.
- `wait_for_previous` (internal transport kwarg, popped after the filter) is explicitly excluded so it isn't falsely flagged — the subtlety @mvalentsev confirmed in #1351.

## Behavior change

Previously-silently-dropped unknown kwargs now return `-32602` for non-`**kwargs` handlers. This is the deliberate strictness increase discussed and endorsed in #1351 (agents otherwise retry the same wrong-shape call); `**kwargs` pass-through handlers are unaffected.

## Tests

`TestUnknownParamName` (4):

- Two are **red→green**: on unmodified `develop` they fail with the indirect `Missing required parameter 'content' ...` (the #1500 symptom), and pass once the wrong name is named.
- Two are **regression guards** (pass before and after): `wait_for_previous` not flagged for a normal handler; `**kwargs` pass-through still accepts unknown.

No API key / network needed.

## Verification

```
ruff check .                                       # pass  (CI-pinned ruff 0.4.x)
ruff format --check .                               # clean
uv run pytest tests/ --ignore=tests/benchmarks     # 1764 passed, 1 skipped
```

Sequenced after #1500 per my note in #1351 (both touch the dispatch region); rebased on current `develop` (contains #1500).

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -2242,6 +2242,26 @@ def handle_request(request):
         except (ValueError, TypeError):
             accepts_var_keyword = False
         if not accepts_var_keyword:
+            # An unknown kwarg here is almost always a wrong parameter *name*
+            # (e.g. text= instead of content=). Silently dropping it makes the
+            # cause surface only indirectly as a later "Missing required 'X'",
+            # so name it explicitly — symmetric with the missing-required path
+            # below. wait_for_previous is an internal transport kwarg in no
+            # tool schema; it is popped before dispatch further down, so it
+            # must not be reported as unknown here.
+            unknown = [k for k in tool_args if k not in schema_props and k != "wait_for_previous"]
+            if unknown:
+                quoted = ", ".join(f"'{k}'" for k in unknown)
+                word = "parameter" if len(unknown) == 1 else "parameters"
+                logger.debug("Tool %s: unknown %s %s", tool_name, word, quoted)
+                return {
+                    "jsonrpc": "2.0",
+                    "id": req_id,
+                    "error": {
+                        "code": -32602,
+                        "message": f"Unknown {word} {quoted} for tool {tool_name}",
+                    },
+                }
             tool_args = {k: v for k, v in tool_args.items() if k in schema_props}
         # Coerce argument types based on input_schema.
         # MCP JSON transport may deliver integers as floats or strings;
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -2066,3 +2066,105 @@ class TestParamShapeDiagnostics:
         assert resp["error"]["message"] == "Internal tool error"
         assert "'foo'" not in resp["error"]["message"]
         assert "some_helper" not in resp["error"]["message"]
+
+
+class TestUnknownParamName:
+    """A kwarg not in the tool schema (wrong parameter *name*, e.g. text=
+    instead of content=) should surface as JSON-RPC -32602 naming the
+    offending kwarg, instead of being silently dropped and resurfacing
+    indirectly as a later "Missing required 'X'". Symmetric with the
+    missing-required path in TestParamShapeDiagnostics. The internal
+    wait_for_previous transport kwarg must never be flagged, and
+    **kwargs pass-through handlers must keep accepting unknown kwargs.
+    """
+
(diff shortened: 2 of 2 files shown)


igorls (approved):
Clean, focused fix that perfectly complements #1500.

**What I checked:**

- Implementation is correct: the unknown-param check sits exactly where it should — after the `**kwargs` gate and before the schema filter — so pass-through handlers are unaffected and `wait_for_previous` is properly excluded.

- All 4 new tests pass locally.

- Full test suite passes (1743 passed, 1 skipped).

- ruff check is clean on `mcp_server.py`; the `tests/test_mcp_server.py` format flag is a pre-existing issue on `develop`, not introduced by this PR.

- The message shape (`Unknown parameter 'x' for tool y`), `-32602` code, and `logger.debug` precedent are symmetric with #1500's missing-required path. Good consistency.

- The docstring on `TestUnknownParamName` clearly explains the edge cases (`wait_for_previous`, `**kwargs` passthrough) and references the related issues.

Approve. Ready to merge.

Issue #1266, opened 2026-04-29
HNSW pickle corruption after long-running mine: 'unsupported opcode' on next read; repair/migrate cannot recover


commonontology:
## Environment

- MemPalace **3.3.3** (pip install -e from develop branch, 74M clone)
- Windows 11, Python 3.14.3 in venv
- chromadb 1.5.8 (default backend)
- Local-only palace at `C:\Users\stugr\Studio-palace`
- Single-collection palace populated incrementally over a multi-hour bulk-mine session

## Summary

After a long-running bulk mine completed cleanly (12 wings, 11h 46m total, ALL_DONE marker logged, mempalace processes exited cleanly, no errors in mine logs), `mempalace search` fails with a pickle deserialization error in the HNSW segment metadata. `mempalace repair` and `mempalace migrate` cannot recover because they hit the same broken read path. Workaround is non-obvious. Data layer in `chroma.sqlite3` is fully intact; only the HNSW acceleration index metadata is corrupt.

## Reproduction shape

1. Run a series of `mempalace mine <dir> --mode projects --wing <name>` invocations sequentially in a script over many hours, hitting one shared palace.
2. Each wing completes cleanly. Final `ALL_DONE` marker present. No errors.
3. Next session: `mempalace search "<any query>"` fails with the error below.
4. `mempalace repair --yes` fails with the same error (its own read path traverses the broken segment).
5. `mempalace migrate --dry-run` hangs/errors on the same path.

I cannot give a minimal reproduction yet — it took ~12 hours of mining a 1.2GB-end-state palace to surface — but the symptom set is consistent with an interrupted final flush of `index_metadata.pickle` while the rest of the segment files are well-formed.

## Error

```
Search error: Error executing plan: Error sending backfill request to compactor:
Error constructing hnsw segment reader: Error creating hnsw segment reader:
Error deserializing pickle file: eval error at offset 8171: unsupported opcode 'w'
```

Same error from `repair` and `migrate`.

## State at the time of failure

- `<palace>/chroma.sqlite3` — 1.04 GB, intact, valid SQLite, no journal residue, all drawers queryable via direct SQL
- `<palace>/<segment-uuid>/data_level0.bin` — 167 MB, mtime = end of clf wing
- `<palace>/<segment-uuid>/header.bin` — 100 B
- `<palace>/<segment-uuid>/length.bin` — 400 KB
- `<palace>/<segment-uuid>/link_lists.bin` — 856 KB
[...]

PR #1525, opened 2026-05-15
feat(backends): add sqlite-vec backend

guilhermefriol:
## Summary

Adds a second in-tree backend, `SQLiteVecBackend`, built on the [sqlite-vec](https://github.com/asg017/sqlite-vec) extension. Sits alongside `ChromaBackend`; users opt in with `MEMPALACE_BACKEND=sqlite-vec`. The default install footprint is unchanged — sqlite-vec is an optional extra (`pip install mempalace[sqlite-vec]`).

## Motivation

On large palaces I've been hitting the recurring HNSW failure modes that this repo already documents in detail:

- `link_lists.bin` sparse-file bloat (#344, partially mitigated by the batch/sync threshold guard)
- `Error finding id` / `Internal error` from where-filtered queries after a rebuild (#1315, #1222, #1218)
- Stale or partially-flushed `index_metadata.pickle` requiring quarantine (#1266 / PR #1285)

The current local recovery flow quarantines segments and falls back to BM25, which works for the unfiltered text path but loses the vector signal for `--wing` / `--room` queries (these go through `_bm25_only_via_sqlite`). sqlite-vec stores vectors in a single `.db` file with no on-disk HNSW segments — drift is impossible by construction — and `vec0` `MATCH` joins cleanly with `WHERE json_extract(metadata_json, '$.wing') = ?`, so where-filtered queries resolve natively.

## What's in this PR

- `mempalace/backends/sqlite_vec.py` (new): `SQLiteVecBackend` + `SQLiteVecCollection` implementing the full RFC 001 surface, plus a where-translator that covers every required operator (`\$eq`, `\$ne`, `\$in`, `\$nin`, `\$and`, `\$or`, `\$contains`) and the optional range operators (`\$gt`, `\$gte`, `\$lt`, `\$lte`). Unknown operators raise `UnsupportedFilterError` per §1.4.
- Atomic `update` override: preserves embeddings when only `documents`/`metadatas` change (the base-class default would need a fresh embedding round-trip).
- Lazy embedder: `embedder=None` (default) resolves `mempalace.embedding.get_embedding_function` on first `get_collection`, so callers can pass `query_texts` and docs-only `add` exactly like Chroma. `embedder=False` disables it for callers that pre-embed; a callable is used as-is.
- Legacy positional `get_collection(palace_path, collection_name, …)` shim mirroring `ChromaBackend`.
- `mempalace/backends/__init__.py` + `registry.py`: registered as built-in (`available_backends() == ['chroma', 'sqlite-vec']`).
- `mempalace/palace.py`: `_select_backend()` calls `resolve_backend_for_palace(env_value=…, palace_path=…)` so the registry priority order is honoured. Chroma stays the singleton default and its per-process client cache is preserved unchanged.
- `pyproject.toml`: optional `[project.optional-dependencies] sqlite-vec` and entry-point under `mempalace.backends`.
- `tests/test_sqlite_vec_backend.py` (34 cases): registry, lifecycle, all CRUD methods, every where-operator translation, multi-collection isolation, unsupported-operator rejection. Whole file is skipped (`importorskip`) when `sqlite-vec` is not installed, so the default test matrix is unaffected.

## Storage layout

One `sqlite_vec.db` per palace, alongside `chroma.sqlite3`:

```
collections (name PK, dimension, embedder_name, created_at)
items       (rowid, collection FK, id, document, metadata_json, …, UNIQUE(collection, id))
items_vec_<sanitised_collection_name>  -- vec0 virtual table, dimension fixed at create
```

`metadata_json` is queried with `json_extract` so where-filters work without a dedicated metadata table. Collections are isolated by the `collection` column on `items` and by per-collection `items_vec_*` virtual tables (dimension is per-collection).

## Validation

End-to-end on a real 268,613-drawer palace (Friol.ai notebooks + Claude Code transcripts):

- Full migration: **27 minutes** (165 drawers/s) via SQL read of `embedding_metadata` + re-embed (ONNX is deterministic, so vectors are bit-equivalent to the originals).
- Paridade post-migration: count match in both collections (267,624 + 989), document sampling 10/10 identical.
- `mempalace search "…" --wing claude-sessions` now returns vector-scored results natively (no BM25 fallback).
- Existing test suite (`tests/test_backends.py`, 45 cases) unchanged and passing.

## Compatibility

- Default behaviour is unchanged: with no env var and no `sqlite_vec.db` on disk, every call still routes to `ChromaBackend`.
- `sqlite-vec >= 0.1.6` (extension is loaded via `db.enable_load_extension`).
- Python 3.9–3.14 (matches the existing classifiers).

## Out of scope

- Embedder-identity check on open (`EmbedderIdentityMismatchError` exists in `base.py` but isn't enforced yet — same as Chroma today).
- Cross-backend live migration tooling. I have a working migrate script (SQL read + re-embed + atomic rename) that I'm happy to PR separately if the maintainers want it in-tree.
- HNSW capacity probe equivalent — sqlite-vec doesn't need one, but the MCP server's existing `vector_disabled` path still routes through `_bm25_only_via_sqlite` and isn't touched here.

Happy to split this if you'd prefer the palace.py rewire as a separate prep PR before the backend lands.

### DIFF
--- a/mempalace/backends/__init__.py
+++ b/mempalace/backends/__init__.py
@@ -29,6 +29,7 @@ from .base import (
     UnsupportedFilterError,
 )
 from .chroma import ChromaBackend, ChromaCollection
+from .sqlite_vec import SQLiteVecBackend, SQLiteVecCollection
 from .registry import (
     available_backends,
     get_backend,
@@ -46,6 +47,8 @@ __all__ = [
     "BaseCollection",
     "ChromaBackend",
     "ChromaCollection",
+    "SQLiteVecBackend",
+    "SQLiteVecCollection",
     "DimensionMismatchError",
     "EmbedderIdentityMismatchError",
     "GetResult",
--- a/mempalace/backends/registry.py
+++ b/mempalace/backends/registry.py
@@ -178,12 +178,22 @@ def resolve_backend_for_palace(
 
 
 def _register_builtins() -> None:
-    """Register chroma as the in-tree default."""
+    """Register chroma + sqlite-vec as in-tree backends.
+
+    sqlite-vec is registered eagerly so it appears in
+    ``available_backends()``; the import is light (no network, no DB I/O).
+    The first ``get_backend('sqlite-vec')`` call inside
+    :class:`~mempalace.backends.sqlite_vec.SQLiteVecBackend` is what loads
+    the sqlite-vec extension into a per-palace connection.
+    """
     from .chroma import ChromaBackend
+    from .sqlite_vec import SQLiteVecBackend
 
     # Use setdefault semantics so a caller that pre-registered for tests wins.
     if "chroma" not in _registry:
         _registry["chroma"] = ChromaBackend
+    if "sqlite-vec" not in _registry:
+        _registry["sqlite-vec"] = SQLiteVecBackend
 
 
 _register_builtins()
--- a/mempalace/backends/sqlite_vec.py
+++ b/mempalace/backends/sqlite_vec.py
@@ -16,15 +16,19 @@ $and, $or, $contains, $gt, $gte, $lt, $lte) via SQLite json_extract.
 
 from __future__ import annotations
 
+import hashlib
 import json
 import logging
 import os
 import re
 import sqlite3
 import struct
+import threading
 from typing import Any, ClassVar, Optional
 
-import sqlite_vec as _sqlite_vec_ext
+# ``sqlite_vec`` is an optional dependency: imported lazily inside ``_open_db``
+# so the registry can advertise this backend (``available_backends()``) on
+# every install without crashing default-Chroma users at import time.
 
 from .base import (
     BackendClosedError,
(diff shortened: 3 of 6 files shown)


guilhermefriol:
Thanks for the careful pass, @gemini-code-assist. Rebased onto current `develop` and pushed `855e152` (now `d6bb883` post-rebase) addressing every item:

- **Critical / lazy import** — `import sqlite_vec` is now scoped inside `_open_db`, behind a `try/except ImportError` that raises a clear `BackendClosedError` pointing at `pip install mempalace[sqlite-vec]`. The registry can still advertise the backend in `available_backends()` on any install; we only pay the import the first time someone actually opens a sqlite-vec palace.
- **High / thread-safety** — added a `threading.Lock` and a double-checked-locking pattern in `_open_db`. Hot path stays lock-free; the open path serialises so two concurrent `get_collection` calls can't both create a connection / double-load the extension for the same palace.
- **Security medium / SQL injection in `json_extract`** — JSON paths are now bound as parameters (`json_extract(metadata_json, ?)`), never interpolated into the SQL string. Field names that don't match `^[a-zA-Z_][a-zA-Z0-9_]*$` raise `UnsupportedFilterError`, so something like `wing'; DROP TABLE items--` is rejected at translation time and nested-JSON paths are off the table by design.
- **Medium / endianness** — `_pack_vec` / `_unpack_vec` now pin little-endian (`<f`), matching what the `sqlite-vec` extension expects and keeping a `sqlite_vec.db` portable across hosts of differing endianness.
- **Medium / table-name collision** — `_sanitize_table_suffix` appends an 8-hex-char SHA-1 prefix whenever the input required substitution or exceeded 40 characters. Clean short names round-trip verbatim; two long collection names sharing their first 40 sanitised characters get distinct virtual-table names.
- **Medium / imports** — added `hashlib` + `threading` to the import block alongside the lazy-load comment.

New tests cover the SQL-injection rejection, the long-prefix table-name distinction, and the explicit little-endian assertion (`tests/test_sqlite_vec_backend.py`, 37 cases passing locally; existing `test_backends.py` 66 cases still green for a combined 103/103).

C-LaForest:
Testing this PR on a 155k-drawer chromadb palace to evaluate backend migration. Backend abstraction looks clean — `available_backends()` returns `['chroma', 'sqlite-vec']` and `get_backend('sqlite-vec')` resolves correctly. But two call sites in the larger codebase bypass the registry and prevent end-to-end use of the sqlite-vec backend via CLI.

### Gap 1 — `miner.py:1191` doesn't pass dimension to `get_collection()`

Setup:
- `pip install 'mempalace[sqlite-vec] @ git+https://github.com/guilhermefriol/mempalace@feat/sqlite-vec-backend'`
- `MEMPALACE_BACKEND=sqlite-vec`
- Fresh palace path
- `mempalace init <dir> --yes --no-llm` succeeds, writes `.mempalace/origin.json`

Then:

```
$ MEMPALACE_BACKEND=sqlite-vec mempalace --palace /data/palace-vec-test \
    mine /projects/code-memory --wing code-memory --agent main
```

```
  File "/usr/local/lib/python3.12/site-packages/mempalace/miner.py", line 1191, in _mine_impl
    collection = get_collection(palace_path)
  File "/usr/local/lib/python3.12/site-packages/mempalace/palace.py", line 102, in get_collection
    return backend.get_collection(
  File "/usr/local/lib/python3.12/site-packages/mempalace/backends/sqlite_vec.py", line 850, in get_collection
    raise ValueError(
ValueError: creating a new collection requires options={'dimension': N}
```

\`ChromaBackend.get_collection\` auto-infers dimension from the first embedding; \`SQLiteVecBackend.get_collection\` requires it explicitly. The miner doesn't pass \`options={'dimension': ...}\`. A \`sqlite_vec.db\` file is created (so backend selection IS reaching the miner), but the collection creation fails.

Suggestion: either have \`SQLiteVecBackend.get_collection\` default to 384 when no dimension is provided (matches default embedder dim) — or have the miner probe the embedding function for dimension before calling \`get_collection\`.

### Gap 2 — \`repair.py\` hardcodes \`ChromaBackend()\` and ignores the registry

\`\`\`
\$ grep -n \"ChromaBackend()\" /usr/local/lib/python3.12/site-packages/mempalace/repair.py
44:from .backends.chroma import ChromaBackend, hnsw_capacity_status
259:    col = ChromaBackend().get_collection(palace_path, collection_name)
343:    col = ChromaBackend().get_collection(palace_path, collection_name)
740:    backend = ChromaBackend()
\`\`\`

\`repair --mode from-sqlite --source <chromadb-palace> --palace <empty-dir>\` with \`MEMPALACE_BACKEND=sqlite-vec\` produces chromadb-format output (chroma.sqlite3 + UUID HNSW segment dirs) rather than \`sqlite_vec.db\`. This blocks using \`repair --mode from-sqlite\` as the cross-backend migration path, which was the most natural way to convert an existing chromadb palace.

Suggestion: route the destination backend in \`rebuild_from_sqlite\` through \`resolve_backend_for_palace(palace_path=palace_path)\` so the env var / explicit-flag override applies.

### Context

Underlying motivation for testing this PR: chromadb's \`embeddings_queue\` accumulates indefinitely on our low-volume write pattern (5-200 row bursts), HNSW segment metadata never persists, daily quarantine cycle. Sqlite-vec backend is the cleanest structural exit at our 155k scale. Happy to test patches against either gap if useful.

### Environment

- Linux x86_64, Rocky Linux 10.1 host, rootless podman 5.6
- Container base: python:3.12-slim (Debian 13 trixie)
- mempalace 3.3.5 + PR branch
- sqlite-vec installed via \`mempalace[sqlite-vec]\` extra
- Default ONNX embedder (all-MiniLM-L6-v2, 384-dim)

Issue #1535, opened 2026-05-17
`mempalace mine --limit N` truncates files BEFORE the already-mined skip — fire can finish with 0 new drawers


domijin:
## Summary

`mempalace mine --limit N` truncates the scanned file list before the already-mined skip, so a fire can produce zero new drawers even when the corpus has unmined files within the time budget. `--limit` ends up bounding the WALK, not the new WORK.

## Repro shape

Corpus where ~85% of files are already mined (deterministic-drawer-id idempotent re-mines have run); new files added under a subdirectory that appears late in `os.walk` order.

```
$ mempalace mine --limit 500 /path/to/corpus
...
  Files processed: 500
  Files skipped (already filed): 500
  Drawers filed: 0
```

The 500 budget was consumed by 500 already-mined files (the skip is ~ms each); the unmined files at later walk positions are never reached. Re-running the same command gets the same result because `os.walk` order is deterministic per filesystem.

## Root cause

In `_mine_impl` (`mempalace/miner.py:1063-1065`):

```python
if limit > 0:
    files = files[:limit]
```

Runs BEFORE the `for filepath in files: process_file(...)` loop. `process_file` does `file_already_mined(...)` at line 810 and short-circuits with `return 0, "general"`, but it's still consuming a slot under the limit.

## User-facing impact

For users running mempalace under a wall-clock budget (launchd / cron / systemd), `--limit` doesn't reliably yield forward progress. Either the corpus is small enough that `os.walk` order randomly distributes unmined files into the first-N window, or the user has to maintain their own pre-filter / checkpoint externally.

Hit this in production on a 2,067-file directory under a 5400s per-fire budget — mines were timing out at ~1871/2067, then with `--limit` added as a workaround the trap would have caught us on subsequent runs.

## Proposed fix (one option)

Move the already-mined filter ahead of the truncate; batchable to avoid O(N) Chroma round-trips:

```python
if limit > 0:
    if not dry_run:
        col = get_collection(palace_path)
        files = [
            f for f in files
            if not file_already_mined(col, str(f), check_mtime=True)
        ]
    files = files[:limit]
```

Cost: O(N) Chroma metadata lookups before the loop (~5–20ms each on a ~250k-drawer palace, so ~10–40s for a 2k-file scan). Worth batching into a single `collection.get(where={\"source_file\": {\"$in\": [...]}}, ...)` call at scale.

[...]

PR #1536, opened 2026-05-17
fix(miner): apply `--limit` AFTER the already-mined skip

domijin:
## Summary

Fix for #1535. Pre-filter already-mined files BEFORE `--limit N` truncates the file list, so `--limit N` bounds NEW work, not WALK.

**Before this PR.** `_mine_impl` (`mempalace/miner.py:1169-1170`) truncates the scanned file list before `process_file`'s `file_already_mined` check (`miner.py:810`). So on a corpus where the first-N walk-order files are all already-mined (very common when running under a wall-clock budget — launchd / cron / systemd retries that mine incrementally), the limit budget gets consumed by the metadata short-circuit and no new drawers are filed. The mine completes "successfully" with `Drawers filed: 0`.

**After this PR.** `--limit N` skips files that are already mined when computing the limit slice, so it bounds NEW work. Matches the natural mental model for budget-bounded mines.

## Test plan

Four new tests in `tests/test_miner_limit_filter.py`:

- [x] `test_limit_skips_already_mined_makes_forward_progress` — two `--limit 3` mines on a 6-file corpus → all 6 mined. **This is the regression case**; would fail under the old behavior because the second mine would see the same first-3 already-mined files.
- [x] `test_limit_when_corpus_fully_mined_is_safe_noop` — re-mine fully-mined corpus with `--limit` → no new drawers, no crash.
- [x] `test_limit_on_fresh_palace_processes_up_to_n` — `--limit N` on a fresh palace processes ≤ N distinct files.
- [x] `test_dry_run_with_limit_does_not_touch_palace` — `dry_run=True` short-circuits the pre-filter (palace may not exist yet).

All four pass locally against the patched `_mine_impl`.

## Trade-offs

- Adds **O(N) chroma metadata lookups** before the file loop. For a 2k-file scan on a ~250k-drawer palace: ~10–40s upfront (depends on Chroma index size). Worth batching with `$in` queries if the per-file cost becomes a bottleneck at larger scale — happy to take that suggestion in review.
- **Behavior change:** `--limit N` no longer means "stop after the first N input files (some may be skipped as already-mined)". It now means "do up to N units of NEW work". This is arguably the more useful semantics for the budget-bounded use case, but does change behavior for any tooling that depended on the old "first-N-in-walk-order" interpretation. Happy to instead add a `--new-only` flag and keep `--limit` semantics unchanged if you'd prefer the back-compat path — let me know.

## Context

We hit this in production on a 2,067-file `repo_docs` wing under a 5400s per-wing budget — mines were timing out at ~1871/2067, then with `--limit` added as a workaround the trap would have caught us on subsequent fires. The full thread is in issue #1535. We're shipping a local vendor patch in our wrapper repo for now; if this lands upstream the patch becomes a no-op and we drop it.

## Marking as Draft

Opened as draft pending your design preference (`--limit` semantics change vs. new `--new-only` flag). Will mark ready-for-review once direction is confirmed.

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -1167,6 +1167,17 @@ def _mine_impl(
             include_ignored=include_ignored,
         )
     if limit > 0:
+        # `--limit N` bounds NEW work, not WALK. Without this pre-filter,
+        # `--limit` truncates the file list before `process_file`'s
+        # `file_already_mined` skip, so a mine on a corpus where the first-N
+        # walk-order files are all already-mined produces 0 new drawers
+        # despite the budget being intact. See #1535.
+        if not dry_run:
+            mined_col = get_collection(palace_path)
+            files = [
+                f for f in files
+                if not file_already_mined(mined_col, str(f), check_mtime=True)
+            ]
         files = files[:limit]
 
     from .embedding import describe_device
--- a/tests/test_miner_limit_filter.py
+++ b/tests/test_miner_limit_filter.py
@@ -1,0 +1,165 @@
+"""Tests for `--limit` semantics — should bound NEW work, not WALK.
+
+Regression for #1535: before the fix, `--limit N` truncated the file list
+BEFORE `file_already_mined` was checked in `process_file`, so a mine on a
+corpus where the first-N walk-order files were all already-mined produced
+0 new drawers despite the budget being intact.
+
+These tests exercise:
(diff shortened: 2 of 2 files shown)


domijin:
Closing — keeping this as a local patch on our side rather than pushing upstream. Thanks!

PR #1541, opened 2026-05-18
ci(lint): pin ruff to 0.15.9 (CI + dev) and reformat tree

igorls:
## Problem

CI installs `ruff>=0.4.0,<0.5` (resolves to **0.4.10**) while contributors run a modern ruff (0.5+). ruff 0.5 changed assert-message formatting, so code formatted by a current ruff fails CI's `ruff format --check .` on **layout alone**, with no actual code defect.

This has blocked three milestone v3.3.6 PRs already — #1438, #1445, #1528 — each requiring a manual "reformat with ruff 0.4.x and force-push" cycle. It will keep recurring for every new contributor on a current ruff.

## Fix

Pin an exact, modern ruff (`0.15.9`) in **both** dev dependency lists (`[project.optional-dependencies].dev` and `[dependency-groups].dev`) **and** in `ci.yml`, so CI and `pip install -e ".[dev]"` / `uv sync --extra dev` format and lint identically. Single source of truth, no version drift.

One-time mechanical `ruff format .` so the tree passes `ruff format --check .` under the new pin.

## Blast radius (measured)

- `ruff check .` under 0.15.9 with the existing `select = ["E","F","W","C901"]`: **All checks passed** — zero new lint violations.
- `ruff format .`: **29 files** — 28 under `tests/` + 1 `tools/` helper. **No core `mempalace/` modules touched.** Layout only (mostly assert-message parenthesization), no behavior change.

Commits are split so review is easy: [`96a3272`] is the pin only; [`2cab6f5`] is the pure mechanical reformat.

## Note for open PRs

The reformat will cause trivial rebase conflicts only for open PRs that touch the same 28 test files. Core-module PRs are unaffected. Recommend merging this promptly to minimize that window.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,9 @@ jobs:
         with:
           python-version: "3.11"
           cache: 'pip'
-      - run: pip install "ruff>=0.4.0,<0.5"
+      # Keep this pin identical to the `ruff==` pin in pyproject.toml
+      # ([project.optional-dependencies].dev and [dependency-groups].dev)
+      # so CI and `pip install -e ".[dev]"` format/lint identically.
+      - run: pip install "ruff==0.15.9"
       - run: ruff check .
       - run: ruff format --check .
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,10 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Keep in lock-step with the ruff version pinned in .github/workflows/ci.yml
-    # (>=0.4.0,<0.5). Using a newer rev here produces a different formatter
-    # output than CI and breaks `ruff format --check` in the lint job.
-    rev: v0.4.10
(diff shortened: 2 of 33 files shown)


igorls:
Addressed the review feedback in 283ef03:

- **Copilot — `.pre-commit-config.yaml`**: valid. It still pinned `ruff-pre-commit` at `v0.4.10`, so pre-commit users would reintroduce the exact formatter drift this PR removes. Bumped to `v0.15.9` and rewrote the lock-step comment to point at `pyproject.toml` as the source of truth.
- **Copilot — `uv.lock`**: valid. The dev `ruff` specifier was still `>=0.4.0`. Regenerated `uv.lock`; both dev entries now record `==0.15.9`.

- **gemini-code-assist (HIGH) — "ruff 0.15.9 does not exist"**: false positive from stale model knowledge ("as of early 2025"). It is now 2026-05. PyPI currently has ruff up to **0.15.13**, and **0.15.9 is a real release**. CI on this PR ran `pip install "ruff==0.15.9"` and the lint job plus all test jobs passed green. The suggested downgrade to `0.5.9`/`0.9.4` would be a regression and is not being applied.

PR #1549, opened 2026-05-19
docs: point benchmark reproduction note to BENCHMARKS.md

MackDing:
Updates the benchmark reproduction note in the root README so it points to the same detailed benchmark guide referenced elsewhere in the repository.


MackDing:
Quick follow-up: this PR appears mergeable from my side, and I do not see a failing required check. Happy to make any maintainer-preferred tweak if needed.

PR #1556, opened 2026-05-20
feat(backends): RuvectorPostgresBackend + embedding resolver hoist + migrate_dim + HNSW

drdave-flexnetos:
## Summary

Adds an alternative storage backend backed by **Postgres + the ruvector
extension** as a peer of ChromaBackend, plus the small surface changes
needed to make the existing backend abstraction reach the MCP layer.

The work landed in two atomic commits so the dependency order is clear:

1. **refactor(embedding): hoist resolve_embedding_function into backend-neutral module**
   — non-chroma backends should not need to import `ChromaBackend` just
   to embed. Adds `mempalace.embedding.resolve_embedding_function()`
   wrapping the existing `get_embedding_function` with the
   log-and-return-`None`-on-failure contract callers depend on. The
   existing `ChromaBackend._resolve_embedding_function` becomes a thin
   shim so all current call sites in `chroma.py` and `mcp_server.py`
   continue to work unchanged.

2. **feat(backends): RuvectorPostgresBackend + mcp_server routing + migrate_dim + HNSW**

   * New `mempalace/backends/ruvector_postgres.py`: implements
     `BaseBackend` + `BaseCollection`. Per-palace single-table layout
     (`mp_<palace>_<collection>` with `embedding ruvector(N)`). First
     write captures `N` from the supplied embedding via
     `_ensure_table(dim)`; refuses to provision at a placeholder dim
     because `ruvector(N)` is immutable per column.
   * `migrate_dim(new_dim, *, re_embed=True)` performs a safe in-place
     rebuild when the EF dim changes (stage new table, re-embed rows,
     atomic `DROP + ALTER RENAME` swap; aborts BEFORE the destructive
     step if the EF returns wrong-dim vectors).
   * `_ensure_table` also creates an HNSW index
     (`ruvector_cosine_ops`, m=16, ef_construction=64) on the embedding
     column so cosine search stays sub-linear as the corpus grows.
   * `mcp_server._get_collection` early-returns through
     `palace.get_collection()` when `_DEFAULT_BACKEND` is non-Chroma so
     MCP drawer writes actually reach the registered backend.
     `ChromaBackend`'s code path is untouched (no regression).
   * `_force_chroma_cache_reset` is now backend-aware (calls
     `close_palace` on non-chroma backends, which would otherwise
     silently `AttributeError`).

## Files

| File | What |
|---|---|
| `mempalace/embedding.py` | new public `resolve_embedding_function()` |
| `mempalace/backends/chroma.py` | `_resolve_embedding_function` becomes a shim |
| `mempalace/backends/ruvector_postgres.py` | new backend (~570 LOC) |
| `mempalace/mcp_server.py` | routing + cache-reset back-compat |
| `tests/test_backend_routing.py` | covers the routing branch + cache reset |
| `tests/test_ruvector_postgres_migrate.py` | 5 integration tests for `migrate_dim` (skipped when `DATABASE_URL` unset) |

## How to use it

```bash
# Run a ruvector-postgres instance reachable on $DATABASE_URL.
export MEMPALACE_BACKEND=ruvector_postgres
export DATABASE_URL='postgres://...'
mempalace-mcp  # writes land in postgres-ruvector, search reads chroma — see note below
```

## Known limitations (called out for reviewer attention)

- `tool_search` calls `search_memories(palace_path=...)` directly,
  bypassing `_get_collection`. So with `MEMPALACE_BACKEND=ruvector_postgres`,
  writes route to postgres but reads still come from chroma. A follow-up
  PR will route `search_memories` through the same backend abstraction
  to close the loop. Documented here so reviewers don't expect a
  full round-trip in this PR.
- `migrate_dim`'s integration tests require a live `DATABASE_URL` (they
  skip in environments without one). The 3 backend-routing tests rely
  on MagicMock harnesses; the maintenance burden of those is unchanged
  by this PR.

## Test plan

- [ ] Existing 57 drawer tests still pass — no behaviour change in the
  ChromaBackend path; verified locally.
- [ ] `tests/test_ruvector_postgres_migrate.py` — 5/5 pass against a
  live `postgres-ruvector` (m=16, ef_construction=64, dim=384→512).
- [ ] Verify with `MEMPALACE_BACKEND=ruvector_postgres` set: writes
  land in `mp_..._<collection>` table; `palace_writable: true` from
  `/readyz`.
- [ ] HNSW index is created automatically on first write; cosine query
  plan flips from Seq Scan to Index Scan on a 100K-row benchmark
  (captured separately, see linked issue body).

## Context

Extracted from a larger memory-mesh integration project; the rest of
that work lives outside this repo. Happy to split into smaller PRs if
preferred — the two commits are independent enough that the refactor
could land on its own first.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -1110,20 +1110,18 @@ class ChromaBackend(BaseBackend):
 
     @staticmethod
     def _resolve_embedding_function():
-        """Return the EF for the user's ``embedding_device`` setting.
+        """Back-compat shim — delegates to :func:`mempalace.embedding.resolve_embedding_function`.
 
         Both ``get_collection`` and ``get_or_create_collection`` must receive
         the EF explicitly — ChromaDB 1.x does not persist it with the
         collection, so a reader that omits the argument silently gets the
         library default and its queries won't match the writer's vectors.
+        New code should call ``resolve_embedding_function`` directly so it
+        does not have to import the chroma backend purely to embed.
         """
-        try:
-            from ..embedding import get_embedding_function
+        from ..embedding import resolve_embedding_function
 
-            return get_embedding_function()
-        except Exception:
-            logger.exception("Failed to build embedding function; using chromadb default")
-            return None
+        return resolve_embedding_function()
 
     # ------------------------------------------------------------------
     # Internal helpers
--- a/mempalace/backends/ruvector_postgres.py
+++ b/mempalace/backends/ruvector_postgres.py
@@ -1,0 +1,578 @@
+"""RuVector-Postgres MemPalace backend (memory-mesh G002).
+
+Talks SQL to a PostgreSQL instance with the `ruvector` extension preloaded
+(see `flexnetos/ruvector-postgres` or the local build from
+`crates/ruvector-postgres/Dockerfile`). Implements the minimum BaseBackend
+and BaseCollection surface so MemPalace can store and recall drawers via
+ruvector's vector type and distance operators.
+
+This backend is intentionally lean. It is not feature-equivalent to the
+ChromaDB reference (no HNSW tuning, no atomic update, no rich `where`
+operator surface). It exists to make `MEMPALACE_BACKEND=ruvector_postgres`
+selectable in the memory-mesh deployment described under
+`_work/memory-mesh/`. Extend as additional capabilities become required.
+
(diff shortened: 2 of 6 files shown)


drdave-flexnetos:
Retargeting to FlexNetOS/mempalace fork for internal merge. Will re-open upstream after fork validation.

PR #1567, opened 2026-05-21
feat: add OpenCode MCP integration

Dxrk777:
## Summary

Adds OpenCode configuration so users can connect MemPalace as an MCP server with zero manual setup.

## What

- `.opencode/opencode.json` — declares the `mempalace-mcp` server as a local MCP endpoint in OpenCode format

OpenCode reads this file automatically when opened in the repo root, so `opencode mcp list` shows MemPalace as connected without any user configuration.

## Context

OpenCode is an open-source AI coding agent (opencode.ai). It supports MCP servers natively via `.opencode/opencode.json` or `.mcp.json`. Existing integrations:

- `.claude-plugin/` — Claude Code
- `.codex-plugin/` — Codex
- `.opencode/` — this PR

No API keys, no external services, no build steps. The user just needs `uv tool install mempalace` and OpenCode picks up the config automatically.

### DIFF
--- a/.opencode/opencode.json
+++ b/.opencode/opencode.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "https://opencode.ai/config.json",
+  "mcp": {
+    "mempalace": {
+      "type": "local",
+      "command": ["mempalace-mcp"],
+      "enabled": true
+    }
+  }
+}
--- a/.opencode/opencode.json
+++ b/.opencode/opencode.json
@@ -3,7 +3,7 @@
   "mcp": {
     "mempalace": {
       "type": "local",
-      "command": ["mempalace-mcp"],
+      "command": ["python", "-m", "mempalace.mcp_server"],
       "enabled": true
     }
   }


Dxrk777:
Good catch! Updated the command to use `python -m mempalace.mcp_server` instead of the direct binary. This is more portable across different installation methods (pip, uv, dev install).

igorls:
Thanks for the minimal OpenCode MCP config. I reviewed this while preparing an agent-support integration branch.

The config itself is useful, but I do not think we should merge it as `.opencode/opencode.json` at the repository root. That would arm a project-local OpenCode config for anyone who checks out the MemPalace repo, which is the same reason our other editor hook/config examples live under `examples/` instead of active workspace locations.

A mergeable revision would be:

- Move this to something like `examples/opencode/opencode.json`.
- Add a short `examples/opencode/README.md` explaining where users should copy/merge it in their own OpenCode config.
- Keep the command portable (`python -m mempalace.mcp_server` is fine here).
- Mention that the first-party OpenCode source adapter from #1484 is the ingest path for historical sessions, while this config is only MCP wiring.

Small change, good idea, just needs to be packaged as an example rather than live repo configuration.

PR #1577, opened 2026-05-21
fix(llm): set explicit User-Agent on urllib HTTP requests (#1570)

### LINKED ISSUE #1570: openai-compat provider probe fails with 403 on Cloudflare-fronted endpoints (Python-urllib User-Agent blocked)
When using `--llm-provider openai-compat` with a Cloudflare-fronted endpoint, `mempalace init` always falls back to heuristics-only even when the endpoint and API key are valid.

Root cause: `check_available()` in `llm_client.py` uses Python `urllib` which sends `Python-urllib/3.x` as User-Agent - Cloudflare blocks this with 403. The same request works fine with curl.

Reproduce with:

```
curl -H "User-Agent: Python-urllib/3.12" -H "Authorization: Bearer <key>" https://<endpoint>/v1/models
```

Fix: add `headers={"User-Agent": "mempalace/1.0"}` to the `Request()` call in `check_available()`.

Alternative: add a `--skip-probe` flag to bypass the availability check entirely.

mvalentsev:
## What does this PR do?

Sets an explicit `User-Agent` on every MemPalace request to a third-party HTTP endpoint, and leaves the tree with one spelling for it, so `mempalace init --llm-provider openai-compat` works against Cloudflare-fronted endpoints. Closes #1570.

Cloudflare-fronted endpoints (Workers AI, AI Gateway, Together, OpenRouter, ...) commonly run a WAF "User Agent Blocking" rule against Python's stdlib default `Python-urllib/<py-version>` and return a bare HTTP 403 before auth. `mempalace init` then silently falls back to heuristics-only even when the endpoint and API key are valid. The same WAF also blocks runtime `classify()` POSTs, so a probe-only fix would still leave LLM-assisted classification broken after init.

**One constant instead of two spellings.** `develop` identifies itself two different ways at three call sites: `f"mempalace/{__version__}"` in `embedding.py` and `backends/qdrant.py`, and `"mempalace-update-check"` in `update_awareness.py`. New `mempalace/user_agent.py` holds `USER_AGENT = f"mempalace/{__version__}"`, the spelling the two newest sites already ship, and every caller imports it:

**Core LLM path (3 sites in `mempalace/llm_client.py`, 1 in `mempalace/closet_llm.py`):**
- `OpenAICompatProvider.check_available()` (the path #1570 hits)
- `OllamaProvider.check_available()` (reverse-proxy / Tailscale case)
- `_http_post_json()` (every runtime `classify()` POST; covers `AnthropicProvider` too, since `api.anthropic.com` answers with `server: cloudflare` and a `cf-ray` header)
- `closet_llm._call_llm()` (optional closet-regeneration feature)

**Sites that already sent something, now sharing the constant:**
- `mempalace/embedding.py`, `mempalace/backends/qdrant.py`: byte-identical on the wire, the literal moves behind the constant, and their existing tests (`test_embedding_api.py`, `test_qdrant_user_agent.py`) pass untouched
- `mempalace/update_awareness.py`: the PyPI release check now sends `mempalace/<version> (update-check)`, keeping its purpose marker and gaining the version
- `benchmarks/` (7 sites in 4 files). `benchmarks` is excluded from CI ruff (`extend-exclude = ["benchmarks"]`), so format/lint scope is unchanged.

Calls to a MemPalace hub or peer keep the stdlib default for now: `hub_client.py`, `server_registry.py`, `transport.py` and the `cli/_hub.py` health probes are bearer-authenticated requests to infrastructure the operator runs, rather than the third-party providers #1570 is about, so a header there is a separate change.

**Error handling hardening:**
- UA is set AFTER `**headers` in `_http_post_json` so a future caller passing `User-Agent` in headers cannot silently shadow the WAF bypass (verified by a parametrized test over 4 header casings).
- `Request()` is constructed inside the `try` block in `_http_post_json` because it validates URLs eagerly and raises `ValueError`; needs wrapping as `LLMError` per the docstring contract.
- `ValueError` added to `except` tuples in both `OllamaProvider.check_available` and `OpenAICompatProvider.check_available` for the same reason. `json.JSONDecodeError` is caught BEFORE the broader `ValueError` clause in `_http_post_json` to preserve "Malformed" classification (the former is a subclass of the latter).

No new CLI flags, no external SDK dependencies. Still stdlib urllib only per the module docstring contract.

## How to test

End-to-end against a CF-fronted public endpoint without auth:

```bash
curl -s -o /dev/null -w '%{http_code}\n' -H "User-Agent: Python-urllib/3.12" https://api.together.xyz/v1/models
# -> 403 (blocked before auth)

curl -s -o /dev/null -w '%{http_code}\n' -H "User-Agent: mempalace/3.10.0" https://api.together.xyz/v1/models
# -> 401 (WAF passed, missing API key surfaced as expected)
```

Through the actual provider class:

```bash
uv run python -c "
from mempalace.llm_client import OpenAICompatProvider
p = OpenAICompatProvider(model='llama3',
    endpoint='https://api.together.xyz/v1', api_key='sk-fake')
print(p.check_available())
"
# -> (False, 'Cannot reach https://api.together.xyz/v1: HTTP Error 401: Unauthorized')
```

Malformed endpoint surfaces cleanly (no `ValueError` propagation):

```bash
uv run python -c "
from mempalace.llm_client import OpenAICompatProvider
print(OpenAICompatProvider(model='x', endpoint='not-a-url').check_available())
"
# -> (False, "Cannot reach not-a-url: unknown url type: 'not-a-url/v1/models'")
```

The `401` (not `403`) confirms the WAF passes the request and only the missing-key check rejects it. That is the behavior the reporter needs for a valid endpoint and key combination.

Unit tests cover each fix site, the format of the constant, header-collision precedence (parametrized over 4 casings), Anthropic UA pass-through, `ValueError` handling on malformed endpoints, and the release check's identifier:

```bash
uv run pytest tests/test_llm_client.py tests/test_closet_llm.py tests/test_update_awareness.py \
  tests/test_qdrant_user_agent.py tests/test_embedding_api.py tests/test_user_agent.py -q
# -> 111 passed
```

## Checklist
- [x] No hardcoded paths
- [x] Linter passes (`uvx --from 'ruff==0.16.6' ruff check .` and `ruff format --check .`)

Full suite with `--ignore=tests/benchmarks`: 5421 passed, 53 skipped.

### DIFF
--- a/benchmarks/convomem_bench.py
+++ b/benchmarks/convomem_bench.py
@@ -34,6 +34,8 @@ from datetime import datetime
 
 import chromadb
 
+from mempalace.llm_client import USER_AGENT
+
 sys.path.insert(0, str(Path(__file__).parent.parent))
 
 HF_BASE = "https://huggingface.co/datasets/Salesforce/ConvoMem/resolve/main/core_benchmark/evidence_questions"
@@ -93,7 +95,7 @@ def discover_files(category, cache_dir):
             return json.load(f)
 
     try:
-        req = urllib.request.Request(api_url)
+        req = urllib.request.Request(api_url, headers={"User-Agent": USER_AGENT})
         with urllib.request.urlopen(req, timeout=15) as resp:
             files = json.loads(resp.read())
             paths = [
--- a/benchmarks/convomem_bench.py
+++ b/benchmarks/convomem_bench.py
@@ -34,7 +34,7 @@ from datetime import datetime
 
 import chromadb
 
-from mempalace.llm_client import USER_AGENT
+from mempalace.user_agent import USER_AGENT
 
 sys.path.insert(0, str(Path(__file__).parent.parent))
 
--- a/benchmarks/locomo_bench.py
+++ b/benchmarks/locomo_bench.py
@@ -33,6 +33,8 @@ from pathlib import Path
 from collections import Counter, defaultdict
 from datetime import datetime
 
+from mempalace.llm_client import USER_AGENT
+
 import chromadb
 
 sys.path.insert(0, str(Path(__file__).parent.parent))
@@ -406,6 +408,7 @@ def _llm_call(prompt, api_key, model="claude-haiku-4-5-20251001", max_tokens=32)
         "https://api.anthropic.com/v1/messages",
         data=payload,
         headers={
+            "user-agent": USER_AGENT,
             "x-api-key": api_key,
             "anthropic-version": "2023-06-01",
             "content-type": "application/json",
@@ -570,6 +573,7 @@ def llm_rerank_locomo(
             "content-type": "application/json",
         }
 
+    headers["user-agent"] = USER_AGENT
     req = urllib.request.Request(url, data=payload, headers=headers, method="POST")
 
     import socket as _socket
--- a/benchmarks/locomo_bench.py
+++ b/benchmarks/locomo_bench.py
@@ -33,7 +33,7 @@ from pathlib import Path
 from collections import Counter, defaultdict
 from datetime import datetime
 
-from mempalace.llm_client import USER_AGENT
+from mempalace.user_agent import USER_AGENT
 
 import chromadb
 
--- a/benchmarks/longmemeval_bench.py
+++ b/benchmarks/longmemeval_bench.py
@@ -39,6 +39,8 @@ from pathlib import Path
 from collections import defaultdict
 from datetime import datetime
 
+from mempalace.llm_client import USER_AGENT
+
 import chromadb
 
 # Add mempal to path
@@ -2420,6 +2422,7 @@ def diary_ingest_session(session, sess_id, api_key, model="claude-haiku-4-5-2025
         "https://api.anthropic.com/v1/messages",
         data=payload,
         headers={
+            "user-agent": USER_AGENT,
             "x-api-key": api_key,
             "anthropic-version": "2023-06-01",
             "content-type": "application/json",
@@ -2851,6 +2854,7 @@ def llm_rerank(
             "content-type": "application/json",
         }
 
+    headers["user-agent"] = USER_AGENT
     req = urllib.request.Request(url, data=payload, headers=headers, method="POST")
 
     import socket as _socket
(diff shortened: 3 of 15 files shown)


mvalentsev:
Both medium findings from @gemini-code-assist are addressed in 28acea6:

- `_http_post_json`: `Request()` moved inside the try block, `ValueError` added to the catch so malformed URLs surface as `LLMError` per the docstring.
- `OpenAICompatProvider.check_available`: same `ValueError` widening for symmetry with `OllamaProvider.check_available`.

One catch-order detail worth flagging while we're here: `json.JSONDecodeError` is a `ValueError` subclass, so it has to be caught BEFORE the broader `ValueError` clause. Otherwise malformed bodies misclassify as "Cannot reach" instead of "Malformed". Existing `test_http_post_json_malformed_response` pins this.

Re-walking the codebase for sibling sites turned up two more places worth fixing:

- `mempalace/entity_registry.py` was hardcoding `"MemPalace/1.0"` on its Wikipedia probe; now uses the shared `USER_AGENT` constant so the invariant holds project-wide instead of only in the LLM client.
- Four benchmark runners (`convomem_bench`, `locomo_bench`, `longmemeval_bench`, `model_eval/metrics`) had seven sibling urllib calls between them. Several of those POST to `api.anthropic.com`, which is itself Cloudflare-fronted (`cf-ray` header confirms), so benchmark runs against Anthropic-via-CF would have hit the same WAF surface. `benchmarks/` is excluded from CI ruff via `extend-exclude` in `pyproject.toml`, so format/lint scope is unchanged.

Test additions:

- `test_user_agent_constant_format` now does a full `re.fullmatch` on `mempalace/<semver> (+repo URL)`. A version-bump that drops the URL suffix or the version segment will fail loudly.
- `test_http_post_json_caller_cannot_override_user_agent` parametrized over four header casings (`User-Agent`, `user-agent`, `USER-AGENT`, `User-agent`).
- Three new tests for the `ValueError` paths in `_http_post_json`, `OllamaProvider.check_available`, and `OpenAICompatProvider.check_available`.

Local checks:

```
uv run pytest tests/test_llm_client.py tests/test_closet_llm.py: 67 passed
uvx --from 'ruff==0.15.9' ruff format --check . : 131 files already formatted
uvx --from 'ruff==0.15.9' ruff check . : All checks passed
Python 3.9 AST parse on all changed files: OK
```

PR #1584, opened 2026-05-22
feat(closets): Tier 6a — date+line locators with content-date hierarchy

milla-jovovich:
Closet pointer lines gain a 4th pipe-separated segment of shape ``YYYY-MM-DD:Lstart-Lend`` so retrieval can jump straight to the right span in the source file instead of opening the whole drawer. The date is selected via a content-aware hierarchy so legacy ingests (where ``filed_at`` is misleading) still produce honest pointers when the content itself carries a date.

The legacy 3-segment format remains the fallback for drawers without the new metadata, so existing palaces keep working without a re-mine.

## Pointer shape

Before:
  topic|entities|→drawer_ids

After (when drawer metadata carries line range + a date):
  topic|entities|2024-11-08:L42-L78|→drawer_ids

Backward compat: when ``drawer_metas`` is not passed, or the first meta lacks ``line_start`` / ``line_end``, ``build_closet_lines`` emits the legacy 3-segment form unchanged.

## Date-source hierarchy (NEW)

A new helper ``mempalace.miner._extract_content_date(source_file, content)`` selects a date once per file and stores it as ``content_date`` on every drawer mined from that file. Priority order, first match wins:

  1. **Filename** — ISO regex on stem, then dateutil fuzzy parse for natural-language formats. Handles ``2024-11-08.md``, ``April-6th-2011-notes.md``, ``Nov-8-2024.md``, etc.

  2. **YAML frontmatter** — looks at ``date`` / ``created`` / ``published`` fields. yaml.safe_load already parses ISO dates as ``datetime.date`` objects; dateutil handles string values.

  3. **Content body, first ~10 lines** — a. ISO regex (``YYYY-MM-DD`` / ``YYYY/MM/DD`` / ``YYYY.MM.DD``) b. Slash dates with locale auto-disambiguation: if any first-number > 12 appears anywhere in the head, lock the file's locale to DD/MM (the only consistent reading); otherwise default to US MM/DD for two-digit-year compactness. c. dateutil fuzzy parse for natural-language (``November 8, 2024``, ``8th November 2024``, etc.)

  4. **Filesystem mtime** — ``os.path.getmtime`` formatted as ``YYYY-MM-DD``.

  5. **None** — caller (``_build_drawer_metadata``) leaves ``content_date`` absent; ``build_closet_lines`` falls back to the ``filed_at`` ingestion timestamp's date portion.

``_build_date_line_segment`` in ``palace.py`` now prefers ``content_date`` when present, otherwise uses ``filed_at[:10]``. This is the load-bearing change that makes Tier 6a meaningful for legacy content — a Nov 8 2024 transcript mined today now emits ``2024-11-08:L42-L78`` instead of ``2026-05-22:L42-L78``.

## Changes

1. ``mempalace/miner.py::chunk_text`` — each chunk dict now also carries ``line_start`` / ``line_end`` (1-indexed line numbers in the stripped source). Computed via ``content[:offset].count("\n") + 1`` — approximate locator accurate to ±1 line at chunk boundaries.

2. ``mempalace/miner.py::_extract_content_date`` (NEW) + helpers (``_try_filename_date``, ``_try_frontmatter_date``, ``_try_content_body_date``, ``_try_mtime_date``). Pure functions, no I/O beyond what's named. Uses lazy imports of ``dateutil`` and ``yaml`` to keep top-level import surface unchanged.

3. ``mempalace/miner.py::_build_drawer_metadata`` — gains ``line_start`` / ``line_end`` / ``content_date`` optional kwargs. When provided, stored in drawer metadata; when omitted, keys are absent from the dict — backward compatible.

4. ``mempalace/miner.py`` batched-upsert path — calls ``_extract_content_date(source_file, content)`` ONCE per file (not once per chunk) and passes the result through to every chunk's ``_build_drawer_metadata`` call.

5. ``mempalace/format_miner.py::_file_chunks_locked`` — gains a ``content`` param so the format-miner path can also run ``_extract_content_date`` on the full extracted text. Plumbs ``content_date`` into every batched drawer's metadata. Caller passes ``text`` (the full extracted markdown).

6. ``mempalace/palace.py::build_closet_lines`` — accepts optional ``drawer_metas`` (parallel to ``drawer_ids``). When present and the first meta has both ``line_start`` and ``line_end``, the new segment is spliced in; otherwise legacy 3-segment form. Helper ``_build_date_line_segment`` prefers ``content_date`` over ``filed_at`` for the date portion.

## Tests added (RED-first, then GREEN)

  tests/test_miner.py::TestChunkTextLineRanges          (4 tests)
  tests/test_miner.py::TestBuildDrawerMetadataLineRange (2 tests)
  tests/test_miner.py::TestExtractContentDate           (15 tests)
    - filename ISO / natural-language / compact
    - YAML frontmatter date / created
    - Claude session preamble
    - ISO + natural-language content body
    - Slash-date locale disambiguation (DD/MM lock vs default MM/DD)
    - Priority order (filename > frontmatter > content > mtime)
    - mtime fallback
    - Graceful None returns for missing file / empty content

  tests/test_closets.py::TestBuildClosetLines           (5 new tests)
    - includes_date_line_segment_when_metas_provided
    - falls_back_to_3_segment_format_when_metas_missing  (compat)
    - falls_back_to_3_segment_format_when_metas_lack_line_keys (compat)
    - date_segment_uses_filed_at_date_portion_only
    - content_date_preferred_over_filed_at               (new — Tier 6a hierarchy)

26 new tests total. All RED before this commit, all GREEN after.

## Verification

  pytest tests/test_miner.py::TestChunkTextLineRanges
         tests/test_miner.py::TestBuildDrawerMetadataLineRange
         tests/test_miner.py::TestExtractContentDate
         tests/test_closets.py::TestBuildClosetLines
    → 32 passed

  pytest tests/test_miner.py tests/test_closets.py
         tests/test_format_miner.py tests/test_palace.py
    → 230 passed, 2 skipped, 0 regressions in directly-affected files

  ruff check + ruff format --check
    → All checks passed, 5 files already formatted

  End-to-end smoke test on a YAML-frontmatter file (date: 2024-11-08,
  mined "today"):
    content_date extracted: 2024-11-08
    chunks produced: 1
    drawer meta: content_date=2024-11-08, filed_at='2026-05-22...', lines=1-15
    closet pointer:
      "conversation with lumi about brands of dog food|Filler;Lumi;Aya|2024-11-08:L1-L15|→drawer_a,drawer_b"
    → content_date correctly preferred over filed_at end-to-end.

  OrbStack triple-Linux verify (Py 3.9 / 3.11 / 3.13)
    → 32 targeted Tier 6a + content-date tests pass on each.

## What does this PR do?

Adds date+line locators to closet pointers with a content-aware date hierarchy (filename → frontmatter → content body → mtime → filed_at). See commit body for full design.

## How to test

- pytest tests/test_miner.py tests/test_closets.py tests/test_format_miner.py tests/test_palace.py
 - Try mining a file with YAML frontmatter `date: 2024-11-08` and verify closet pointers reflect that date, not today's filed_at.
  

## Checklist
- [ ] Tests pass (`python -m pytest tests/ -v`) (32 new + 230 in directly-affected files)
- [ ] No hardcoded paths

### DIFF
--- a/mempalace/format_miner.py
+++ b/mempalace/format_miner.py
@@ -587,6 +587,7 @@ def _file_chunks_locked(
     room,
     agent,
     source_mtime: Optional[float] = None,
+    content: Optional[str] = None,
 ):
     """Lock the source file, purge stale drawers, and upsert fresh chunks.
 
@@ -604,7 +605,13 @@ def _file_chunks_locked(
     """
     # Lazy imports to avoid a module-load cycle (miner.py imports from this
     # module's package, so we defer these helpers until call time).
-    from .miner import _extract_entities_for_metadata, detect_hall
+    from .miner import _extract_content_date, _extract_entities_for_metadata, detect_hall
+
+    # Tier 6a content-date: extract once per file (not per chunk). Format-mined
+    # files often have date-rich content (RTF/PDF dates in body text, mtimes on
+    # the binary source). Caller may pass ``content`` (full extracted text) for
+    # the body-scan branch; if absent, the helper still uses filename + mtime.
+    file_content_date = _extract_content_date(source_file, content or "")
 
     drawers_added = 0
     with mine_lock(source_file):
@@ -649,6 +656,17 @@ def _file_chunks_locked(
                 }
                 if source_mtime is not None:
                     meta["source_mtime"] = source_mtime
+                # Tier 6a — propagate line range from chunk dict into drawer
+                # metadata so closet pointers can carry "where in source"
+                # info. Chunks emitted by older code paths without these
+                # keys produce drawers without the keys (graceful fallback).
+                if chunk.get("line_start") is not None:
+                    meta["line_start"] = chunk["line_start"]
+                if chunk.get("line_end") is not None:
+                    meta["line_end"] = chunk["line_end"]
+                # Tier 6a content-date: shared across all chunks of the file.
+                if file_content_date:
+                    meta["content_date"] = file_content_date
                 entities = _extract_entities_for_metadata(content)
                 if entities:
                     meta["entities"] = entities
@@ -864,6 +882,7 @@ def mine_formats(
                     room,
                     agent,
                     source_mtime=source_mtime,
+                    content=text,
                 )
                 if skipped:
                     files_skipped += 1
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -475,7 +475,13 @@ def chunk_text(
     """
     Split content into drawer-sized chunks.
     Tries to split on paragraph/line boundaries.
-    Returns list of {"content": str, "chunk_index": int}
+    Returns list of {"content": str, "chunk_index": int, "line_start": int, "line_end": int}
+
+    ``line_start`` / ``line_end`` are 1-indexed line numbers in the stripped
+    source, giving an approximate locator for where the chunk came from.
+    Closet pointers (Tier 6a) use this to emit ``YYYY-MM-DD:L42-L78`` segments
+    so retrieval can jump straight to the right span without opening the
+    whole drawer.
 
     Optional params override module-level defaults when provided.
     """
@@ -531,10 +537,17 @@ def chunk_text(
 
         chunk = content[start:end].strip()
         if len(chunk) >= min_chunk_size:
+            # Tier 6a — 1-indexed line range in the stripped source.
+            # Approximate locator (±1 at boundaries is fine for "jump to
+            # roughly here"); exact-quote positioning is a future tier.
+            line_start = content[:start].count("\n") + 1
+            line_end = content[:end].count("\n") + 1
             chunks.append(
                 {
                     "content": chunk,
                     "chunk_index": chunk_index,
+                    "line_start": line_start,
+                    "line_end": line_end,
                 }
             )
             chunk_index += 1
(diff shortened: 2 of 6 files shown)


igorls (changes requested):
Spent a fair amount of time on this @milla-jovovich — really like the underlying design (line-level pointers + content-anchored dates). Two things I want to flag because they look like they'd cause real-world surprises, plus a couple of smaller notes:

## Issue 1 — the feature isn't wired into any production path yet

``build_closet_lines`` learns the new ``drawer_metas`` kwarg, but I can't find a production caller that actually passes it:

- ``mempalace/miner.py`` (line ~1340) — ``build_closet_lines(source_file, drawer_ids, content, wing, room)`` — no metas
- ``mempalace/diary_ingest.py`` (line ~290) — same shape, no metas
- ``mempalace/format_miner.py`` — does not emit closets at all

So in practice the 4-segment ``YYYY-MM-DD:Lstart-Lend`` form is produced only by the new tests. Every real palace keeps emitting the legacy 3-segment pointer. The headline benefit ("retrieval can jump straight to the right span") is not realised by anything that mines drawers today. The ``drawer_metas`` path probably wants to be plumbed through ``miner.process_file`` and ``diary_ingest.ingest_diaries`` (or wherever ``line_start`` / ``line_end`` actually land on metadata) for this feature to have user-visible effect.

## Issue 2 — ``_extract_content_date`` hallucinates dates on benign inputs

The dateutil ``fuzzy=True`` branch on filenames and body lines returns plausible-looking dates for almost any digit-bearing input. Reproduced on a fresh checkout of this branch:

| Input | content_date returned |
|---|---|
| ``tmp_random_file_5`` | ``2026-05-05`` |
| ``untitled-1`` | ``2026-05-01`` |
| ``notes.2024.md`` | ``2024-05-22`` |
| ``2024-06.md`` | ``2024-06-22`` |
| ``Bug fix for issue 42 in module 7`` | ``2042-07-22`` |
| ``Tested with 1000 drawers`` | ``1000-05-22`` |
| ``Version 3.3.6 released`` | ``2006-03-03`` |

The four-tier hierarchy says ``mtime`` is tier 4, but mtime almost never gets reached because fuzzy returns *something* from filename or body first.

This matters even with Issue 1 in place, because ``content_date`` still gets written to drawer metadata via ``process_file`` / ``_file_chunks_locked`` regardless of whether the closet emitter consumes it. So a user mining their codebase post-merge gets fabricated ``content_date`` values silently persisted into ChromaDB with no cleanup path — and when Issue 1 is resolved, those bad dates flow straight into the authoritative closet pointers.

Suggested fix shapes:

1. Gate the fuzzy parse behind a regex pre-check that confirms the candidate string contains a year+month or month-name token. Reject ``Version 3.3.6 released``-shaped strings up front.
2. Or drop ``fuzzy=True`` entirely and accept lower recall on natural-language dates.
3. Either way, add negative tests for junk filenames (``tmp_random_file_5``, ``untitled-1``) and digit-containing bodies (``issue 42 in module 7``, ``Version 3.3.6``).

## Smaller, but worth addressing

### ``python-dateutil`` is an undeclared dependency

``pyproject.toml`` deps are ``chromadb``, ``pyyaml``, and ``tomli`` (3.9/3.10 only). ``dateutil`` happens to be installed transitively via ``chromadb → kubernetes → python-dateutil`` today, but that's not a contract — if upstream kubernetes drops the dep (recent versions have been moving toward stdlib-only), recall silently collapses to ISO-only.

Either add ``python-dateutil`` to the project deps explicitly, or replace the dateutil usage with stdlib parsing.

### Slash-date locale heuristic undertested

The "if any first-number > 12 in the file, lock DD/MM" inference is clever, but the two-digit-year disambiguation falls back to stdlib's 70→19xx / 00–69→20xx convention with no test pinning the 1970/2000 boundary. Not urgent but a nice add.

## What's good

- Backward compat on the read side is fine. ``searcher.py``'s ``_CLOSET_DRAWER_REF_RE`` only cares about the ``→ids`` tail; 3- and 4-segment lines both parse cleanly. ``dialect.py`` pipe-split is on AAAK output, not closets.
- The legacy fallback in ``_build_date_line_segment`` is correct (returns ``None`` when ``line_start`` / ``line_end`` are missing, so ``build_closet_lines`` keeps the old format). Test ``test_falls_back_to_3_segment_format_when_metas_lack_line_keys`` covers it.
- The perf rewrite (``str.count`` over slicing, bounded ``split(.., 10)``, ``find`` for frontmatter) is a real improvement and well-commented.
- 127 in-scope tests pass clean.

Thanks for the work — happy to dig deeper on any of the above if useful.

milla-jovovich:
heard! thanks for the input. your input is always valuable. on it now.👍

milla-jovovich:
@igorls — addressed all 4 review items in commit 65f24b3b:
 
  1. Hallucinations killed — strict regex gate before any dateutil call, fuzzy=True removed. All 7 of your repros now return None; tests verbatim from your review.
   2. Production wiring landed — miner.py accumulates metas and passes drawer_metas to build_closet_lines. New end-to-end integration test pins it.
   3. python-dateutil>=2.8 declared explicitly in pyproject.toml.
   4. Year-boundary tests added (1969/1970/1999/2000).
  
   CI green across Linux 3.9/3.11/3.13 + macOS + Windows + lint. Ready for re-review when you have time.

igorls (approved):
Re-reviewed after `65f24b3b`. All four asks from my prior review are substantively addressed:

1. **Wired into production** — `miner.py` batched-upsert now accumulates `all_metas` and passes `drawer_metas=all_metas` to `build_closet_lines`. Verified end-to-end via `TestMinerClosetRebuild::test_production_miner_emits_4_segment_pointers_with_content_date` — GREEN. Diary path deferred to a follow-up (entries are entry-keyed, not chunk-keyed) — agreed scope call.
2. **Hallucination killed** — new `_VALID_DATE_RE` gate runs before dateutil; `fuzzy=True` removed. Verified locally against all 7 of my exact repros (`tmp_random_file_5`, `untitled-1`, `notes.2024.md`, `2024-06.md`, `Bug fix for issue 42 in module 7`, `Tested with 1000 drawers`, `Version 3.3.6 released`) — every one drops to mtime fallback instead of fabricating a date. Real natural-language dates (`2024-11-08.md`, `April-6th-2011-notes.md`, `Nov-8-2024.md`) still extract correctly.
3. **python-dateutil declared** — `>=2.8` in `[project].dependencies`. Good.
4. **Two-digit-year boundary** — 4 tests pin 1969/1970/1999/2000.

Local full-suite verification on my Mac:
- `pytest tests/test_miner.py tests/test_closets.py tests/test_format_miner.py tests/test_palace.py -p no:randomly -q` → **242 passed, 2 skipped**
- Targeted `TestExtractContentDate` + `TestBuildClosetLines` → **95 passed**

CI all green across linux 3.9/3.11/3.13, macOS, Windows, lint, version-guard. Approving and merging.

Issue #1579, opened 2026-05-21
Mines smaller than hnsw:sync_threshold never persist index_metadata.pickle


tim-harmon:
**What happened?**

After mining any directory smaller than the `_HNSW_BLOAT_GUARD` `sync_threshold` (50,000 records), the very next palace open quarantines the segment as drifted:

```
Quarantined corrupt HNSW segment <palace>/<UUID> (sqlite Ns newer than HNSW and integrity check failed); renamed to <UUID>.drift-<timestamp>
```

The same segment UUID is then recreated by chromadb on the next open and re-quarantined again, accumulating `<UUID>.drift-*` directories indefinitely. Six such directories had piled up in my palace over ~5 days of normal use before I investigated.

Affected segment dirs contain `data_level0.bin`, `header.bin`, `length.bin`, and an empty `link_lists.bin`, but **never** `index_metadata.pickle`. This matches the "partial flush" failure shape explicitly documented at `backends/chroma.py:145–149`.

**What did you expect?**

A mine that completes cleanly (exit 0, drawer count shown, summary banner printed) should leave the palace in a state that opens without quarantining. Either the HNSW metadata pickle should be persisted before the writer process exits, or the drift detector should recognize the "small mine, never crossed sync_threshold" case as flush-lag rather than corruption.

**How to reproduce:**

1. Start from a fresh palace (or run `mempalace repair --mode from-sqlite --archive-existing --yes` on an existing one to get to a clean state).
2. Create a tiny corpus and mine it: `mkdir /tmp/repro && echo "test" > /tmp/repro/a.md && mempalace mine /tmp/repro --wing repro`. Mine completes cleanly.
3. Run `mempalace status`. The just-created segment is quarantined as drifted; a `<UUID>.drift-<timestamp>` directory appears in the palace.

Repeating step 3 (or any other palace-open operation) accumulates more quarantine directories under the same UUID.

**Environment:**
- OS: Windows 11 Enterprise (10.0.26200)
- Python version: 3.14.5 (bundled via `uv tool install mempalace`)
- MemPal version: 3.3.5
- chromadb version: 1.5.9
- Shell: PowerShell 7

---

## Additional context — root cause

In `backends/chroma.py:80–103`, `_HNSW_BLOAT_GUARD` sets:

```python
_HNSW_BLOAT_GUARD = {
    "hnsw:batch_size": 50_000,
    "hnsw:sync_threshold": 50_000,
}
```

[...]

PR #1585, opened 2026-05-22
fix(hooks_cli): guard parents[3] access against shallow filesystem paths

milla-jovovich:
``_mempalace_python()`` in ``mempalace/hooks_cli.py`` uses ``Path(__file__).resolve().parents[3]`` to locate the venv Python interpreter in the standard install layout
``<venv>/lib/pythonX.Y/site-packages/mempalace/hooks_cli.py``. When the package lives at a shallow filesystem path — Docker containers mounting at ``/work``, ``/opt/app``, minimal-prefix production installs — ``parents`` has fewer than 4 elements and the index raises ``IndexError`` instead of falling through to the editable-install branch.

The crash was caught by OrbStack-based triple-Python CI verification on PR #1579: 16 tests in ``test_hooks_cli.py`` failed identically on Linux 3.9 / 3.11 / 3.13 with the same ``IndexError: 3`` from ``pathlib._PathBase.parents.__getitem__`` — and verified pre-existing on develop tip in the same container. The bug never surfaces in GitHub Actions CI runners (their workdir at
``/home/runner/work/mempalace/mempalace`` has plenty of parent directories) but it surfaces immediately for anyone:

  - running mempalace in editable mode inside a Docker dev container
  - shipping mempalace as part of an OCI image where the install prefix is ``/app`` or ``/opt/<name>``
  - using OrbStack / Colima / podman-machine for cross-version verification

## The fix

Wrap each ``parents[N]`` access in ``try/except IndexError`` so the helper falls through to the next strategy (editable-install → ``sys.executable``) instead of crashing the hook. Both ``parents[3]`` AND ``parents[1]`` are guarded — the latter is defensive against extreme cases like a file at root (``/file.py``, parents=[/]) — same class of bug.

## Test added (RED-first, then GREEN)

  tests/test_hooks_cli.py::test_mempalace_python_handles_shallow_path_without_crashing

Mocks ``Path(__file__).resolve()`` so ``parents[3]`` raises ``IndexError`` and ``parents[1]`` returns a real shallow path (``/work/mempalace``). Pre-commit: function raises ``IndexError: 3``. Post-commit: function returns a valid Python interpreter path (either editable-venv if present, otherwise ``sys.executable``).

## Verification

  pytest tests/test_hooks_cli.py
    → 110 passed, 1 skipped on macOS (the existing run)
    → 110 passed, 1 skipped on Linux 3.9 / 3.11 / 3.13 (OrbStack)
       — was 16 failed, 94 passed before this commit

  ruff check + ruff format --check (pinned 0.15.9)
    → All checks passed; 2 files already formatted

## Checklist
- [ ] No hardcoded paths

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -75,13 +75,28 @@ def _mempalace_python() -> str:
         return env_python
     # This file lives at <venv>/lib/pythonX.Y/site-packages/mempalace/hooks_cli.py
     # or <project>/mempalace/hooks_cli.py (editable install).
-    venv_bin = Path(__file__).resolve().parents[3] / "bin" / "python"
-    if venv_bin.is_file():
-        return str(venv_bin)
+    #
+    # Both ``parents[3]`` and ``parents[1]`` access can raise IndexError when
+    # the package lives at a shallow filesystem path — Docker containers
+    # mounting at ``/work``, ``/opt/app``, or other minimal-prefix installs
+    # don't have 4 (or sometimes even 2) parent directories. Guard each
+    # access so we fall through to the next strategy instead of crashing
+    # the hook with an IndexError that would kill any Claude-Code session
+    # using mempalace.
+    resolved = Path(__file__).resolve()
+    try:
+        venv_bin = resolved.parents[3] / "bin" / "python"
+        if venv_bin.is_file():
+            return str(venv_bin)
+    except IndexError:
+        pass
     # Editable install: assumes project root has a venv/ sibling to mempalace/
-    project_venv = Path(__file__).resolve().parents[1] / "venv" / "bin" / "python"
-    if project_venv.is_file():
-        return str(project_venv)
+    try:
+        project_venv = resolved.parents[1] / "venv" / "bin" / "python"
+        if project_venv.is_file():
+            return str(project_venv)
+    except IndexError:
+        pass
     return sys.executable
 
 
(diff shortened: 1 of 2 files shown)


igorls (approved):
Verified the bug locally — ``parents[3]`` on ``/app/mempalace/hooks_cli.py`` gives ``parents = [/app/mempalace, /app, /]`` (len 3), so the access raises ``IndexError`` before reaching the editable-install / ``sys.executable`` fallback chain. Real bug, real impact for any Docker/OCI install at a shallow prefix (``/app``, ``/opt/<name>``), OrbStack/Colima/podman-machine cross-version verification.

The fix is correct. Small note — the implementation uses LBYL (``len(parents) > N``) rather than the ``try/except IndexError`` the PR title suggests; the LBYL form is actually cleaner and matches what gemini-code-assist requested upthread. The fallback chain (``MEMPALACE_PYTHON`` env override → parents[3] → parents[1] → ``sys.executable``) is preserved and bound-checked symmetrically. Cross-platform safe: Windows ``C:\app\mempalace\...`` has the same 3-parent shape and benefits from the same guard, and ``.resolve()`` is called before ``.parents`` so symlinked installs collapse to the true depth.

Tests pass locally (109 in test_hooks_cli.py). One small follow-up suggestion you might consider as a future tweak: ``@pytest.mark.parametrize`` over depths 1, 2, 3, 4 would catch any future regression more robustly than the single len==3 mock (extreme cases like ``/mempalace/hooks_cli.py`` depth 2 aren't directly exercised). Not a blocker.

Merging.

PR #1594, opened 2026-05-23
feat(wing): wing_aliases.json collapses clones on both hook + CLI miner paths

nikkunikku:
## Summary

Builds on #1410. With that PR's cwd-from-jsonl primary path, every working copy on disk gets its own correctly-named wing — `joy-web` → `wing_joy_web`, `joy-web-3` → `wing_joy_web_3`. That's the right default, but for users who keep multiple working copies of one repo (parallel branches, comparison checkouts, scratch clones), it splinters into N wings for what is conceptually one project. This PR closes that gap **on both code paths** (hook auto-save + CLI miner) so a single alias map governs all wing decisions.

## The alias map

`~/.mempalace/wing_aliases.json`:

```json
{
  "aliases": {
    "joy-web*":        "wing_joy_web",
    "legacy-monorepo": "wing_legacy"
  }
}
```

Resolution:

1. Exact key — basename matches verbatim.
2. `*`-suffix key — basename equals the stem OR continues with `-` / `_`. Delimiter-anchored, so `joy-web*` matches `joy-web` / `joy-web-3` / `joy-web_staging` but NOT `joy-website`.
3. Exact beats prefix; longest prefix wins on ties.
4. Values may include or omit the `wing_` prefix; both normalize to one.
5. Parse errors are logged via `_log()` (the hook log) and the file is treated as empty so callers degrade rather than crash.
6. Result is cached single-slot per path keyed by `mtime_ns`; edits are picked up automatically; the cache never grows past one entry per `PALACE_ROOT`.

## Where it applies

**Hook auto-save (Stop / PreCompact).** `_wing_from_transcript_path` runs every hook fire. Resolution order:

1. Read `cwd` from the transcript jsonl (canonical when present).
2. Else decode the encoded folder under `.claude/projects/-...`, strip user-home and one parent-dir token, route the residual through `_basename_to_wing`. The suffix walk is **exact-only** — a prefix alias like `web*` cannot fire against a right-to-left suffix candidate, because the dash-joined residual cannot distinguish a flattened `/` from a hyphen in a directory name. Letting prefix matches fire on suffixes would silently merge `clients/joy-web-3` onto whatever `web*` aliases to — a cross-project collision that violates verbatim-recall semantics.
3. Else legacy `-Projects-<name>` segment.
4. Else `wing_sessions` default.

**CLI miner (`mempalace mine --mode convos`).** `convo_miner._resolve_wing` previously had no way to consult aliases; users with N joy-web clones had to write a wrapper that hand-passed `--wing` per directory. Resolution order is now:

1. Explicit `--wing` — unchanged, still wins.
2. AI-tool path (`~/.claude/projects/`, `~/.codex/`, `~/.gemini/`) **+ alias hit** — NEW. Read `cwd` from the first transcript under the target dir, look the basename up in `_load_wing_aliases`, return the aliased wing when an exact or prefix alias actually fires.
3. AI-tool path, no alias hit — unchanged. Returns `wing_api` so users without `wing_aliases.json` configured see zero behavior change.
4. Basename fallback — unchanged.

The "honor only when an alias fires" rule is intentional. An unconditional `cwd → wing` rule would splinter every existing user's AI corpora across one wing per project basename the moment they upgrade — exactly the regression the `wing_api` default was designed to prevent.

## Motivating use case

Six working copies of one repo:

```
~/Claude/joy-web
~/Claude/joy-web-1
~/Claude/joy-web-2
~/Claude/joy-web-3
~/Claude/joy-web-4
~/Claude/joy-web-5
```

Before this PR, hook saves from any clone splintered across `wing_joy_web` / `wing_joy_web_1` / .../ `wing_joy_web_5` — six wings for one project. With `{"joy-web*": "wing_joy_web"}` they all collapse. The CLI miner change means `mempalace mine ~/.claude/projects/-Users-me-Claude-joy-web-3 --mode convos` (no `--wing` flag) also lands in `wing_joy_web` — so periodic re-mines no longer need the wrapper script.

## Encoded-fallback suffix walk (the right-to-left exact-only case)

When `cwd` is not available (older transcripts, queue-only files), `_wing_from_transcript_path` decodes the slug under `.claude/projects/`. For nested layouts like `/Users/me/code/clients/joy-web-3` the post-strip residual is `code-clients-joy-web-3`. The suffix walk tries `clients-joy-web-3`, `joy-web-3`, `web-3`, `3` against the **exact** alias map. This lets users alias a nested-layout basename verbatim (`"clients-joy-web-3": "wing_joy_web"`) without writing the full residual path. Prefix matches are deliberately excluded from this walk — see commit message of `0c36581` for the full reasoning.

## Test plan

- [x] Existing 116 hook tests pass unchanged.
- [x] Existing 35 convo_miner tests pass unchanged.
- [x] 10 alias tests (hook path): exact, prefix, longest-prefix, normalize, malformed file, missing file, cache invalidation, no log noise on missing file, substring-collision negative case, nested-layout fallback.
- [x] 8 alias tests (CLI miner path): prefix collapses clones, exact match (slug-default-equal case), alias miss → `wing_api`, explicit `--wing` still wins, no transcripts → `wing_api`, no `cwd` record → `wing_api`, malformed jsonl no-crash, non-AI-tool paths preserve basename.
- [x] End-to-end verified on real `~/.claude/projects/` corpus: `joy-web-3` → `wing_joy_web`, `cash-fund-service` → `wing_cash_fund_service`, `mempalace` → `wing_mempalace`.
- [x] Lint + format clean (`ruff check`, `ruff format`).

## Commit chain

1. `3754128` — original feature: hook alias map (two-map schema).
2. `b9f6337` — review iteration: delimiter-anchored prefix, single-map schema, mtime cache, error logging, nested-layout fallback.
3. `0c36581` — review iteration: exact-only suffix walk, single-slot cache, ns-based mtime test, smaller cleanups.
4. `d87251b` — new: CLI miner wiring. The alias map now governs both hook + CLI paths.

## Notes for reviewers

- Code reuse: `_resolve_wing` imports `_load_wing_aliases` and `_match_prefix_alias` from `hooks_cli` directly. Could move both (plus `_basename_to_wing`, `_wing_from_jsonl_cwd`, the cache) to a shared `mempalace/wing.py`. Happy to do that here if you want; left it in-place to keep the diff focused on the feature.
- Scope: arguably "feat(hooks)" no longer fits the title — retitled `feat(wing)` to reflect that the same config now drives both code paths.
- The "honor only when an alias fires" rule on the AI-tool branch is the load-bearing line for backwards compatibility. Any reviewer suggesting "just always use the cwd basename" should know that breaks the existing `wing_api` default for users without alias config.


nikkunikku:
Added a third wing-derivation fix (commit `947fda2`): `_ingest_transcript` (the Stop/PreCompact verbatim-ingest spawn) was hardcoding `--wing sessions`, so a single Claude Code session split its memory across two wings — the diary checkpoint correctly routed via `_wing_from_transcript_path` (e.g. `wing_joy_web`), but the substantive verbatim transcript content pooled into a shared `sessions` wing, bypassing `wing_aliases.json`.

This was the **third** wing-derivation path ignoring the alias map (after the diary-checkpoint path in the base commits and the CLI miner `_resolve_wing` in `d87251b`). All three now share `_wing_from_transcript_path` / `_basename_to_wing`. Caught it on a real palace where live-session content was accumulating in `sessions` instead of the project wings.

Fallback is now `wing_sessions` (the resolver default) instead of bare `sessions` — same catch-all semantics for unconfigured users, consistent `wing_` prefix.

+2 tests (alias-routed ingest, default fallback); updated 2 existing tests that pinned the hardcoded `sessions`. Full hooks + convo_miner suites green (153 passed).

This strengthens the "shared resolver governs all wing decisions" thesis — happy to extract the helpers to a `mempalace/wing.py` module if you'd prefer one home for them rather than `hooks_cli`.

igorls:
Maintainer push: I added one commit (`e2364a2`) opening `hook.log` with `encoding="utf-8"` in `_log()`. The `→` in the transcript-ingest log line was crashing the Stop hook on Windows (cp1252 `UnicodeEncodeError`), failing 11 tests in `test_hooks_cli.py`. Fixing the log encoding addresses the root cause (a verbatim memory system will inevitably log non-ASCII filenames/wing names) rather than just swapping the one arrow. Full `test_hooks_cli.py` suite is green locally (123 passed, 1 skipped). Wing-alias logic itself reviewed and looks sound — thanks!

nikkunikku:
Thank you @igorls! Looking forward to having this without a local patch!

nikkunikku:
@igorls resolved merge conflicts, let me know if I can help w anything else!

nikkunikku:
Synced this branch up to current `develop` twice to clear merge conflicts. Notes for reviewers on how each conflict was resolved:

### Round 1 — `develop` @ `f124bd2` (merge `4c6da90`)
- **`mempalace/convo_miner.py`** (imports) — kept `import json` (used for transcript parsing); dropped `import hashlib`, now unused since `develop` moved hashing into the `.ids` module (`make_convo_drawer_id` / `make_convo_sentinel_id`).
- **`tests/test_convo_miner.py`** — additive; kept both our `_resolve_wing` alias tests and `develop`'s `test_mine_convos_limit_skips_already_mined`.
- **`website/guide/configuration.md`** — additive; kept `develop`'s `MEMPALACE_MAX_BACKUPS` env-var row alongside our new "Hook Wing Aliases" section.

### Round 2 — `develop` @ `afd0428` (merge `11f903b`)
The substantive one: both branches changed wing-name resolution.
- **`mempalace/hooks_cli.py`** — `develop` added `_safe_wing_slug`, a hardened slug normalizer that fixes a real bug (folder names with characters `sanitize_name` rejects — leading `+`, `..`, over-length — silently broke diary auto-save). Our PR added alias resolution on top of a naive slug. **Resolved by integrating both:** kept the alias system, kept `_safe_wing_slug`, and routed `_basename_to_wing`'s fallback *through* `_safe_wing_slug`. Net behavior: **aliases first, then the hardened slug** — so non-aliased projects keep `develop`'s safety fix instead of regressing to the naive slug.
- **`tests/test_hooks_cli.py`** — additive; kept both our `wing_aliases` collapse tests and `develop`'s `_safe_wing_slug` tests (including the hypothesis property test asserting any directory name yields a `sanitize_name`-valid wing).

### Verification
- `ruff check` clean; all files compile.
- `pytest tests/test_hooks_cli.py tests/test_convo_miner.py` → **188 passed, 1 skipped** — our alias tests and `develop`'s `+project` / `..` / length tests pass together.
- PR diff remains scoped to the 6 feature files (no `develop` changes attributed to this PR).

nikkunikku:
### Update — synced with `develop` (36d0c06)

Merged current `develop` in and resolved five conflicts.

**Dropped the FTS5 autoheal work from this branch.** It carried a commit for the mine-path FTS5 self-heal (#1596); #1928 landed the same fix upstream in the meantime, and that version is a superset — it passes `progress=logger.info` so the heal banner can't corrupt the MCP JSON-RPC stdout stream, and it adds a full-chain heal test. Took `develop` wholesale on `palace.py` and `tests/test_miner_fts5_validation.py`. Those were the only files that commit touched, so nothing wing-related was lost and the diff drops from 8 files to 6 — now scoped to just the wing-alias change.

**Other resolutions.** `convo_miner.py` and `configuration.md` were adjacency conflicts (both sides appended unrelated content at the same insertion point) — union. In `_ingest_transcript`, kept both sides: develop's write-routing gate runs first, then this branch's alias-aware wing resolution.

**One behavior fix worth a look.** The daemon submit path hardcoded `"wing": "sessions"` while the spawn fallback honored the alias map — so enabling the hooks daemon made `wing_aliases.json` silently inert, which is the exact pooling this PR exists to fix. Both paths now use the resolved wing. `test_ingest_transcript_daemon_opt_in_submits_job` asserted the hardcoded value, so I updated it and added daemon-path alias coverage. Flagging it explicitly since I changed an existing assertion — if that hardcode was deliberate for a daemon-side reason I haven't seen, this is the hunk to revisit.

`ruff check` and `ruff format --check` are clean; the wing-alias tests pass. Remaining full-suite failures on my machine are environmental (SQLite 3.51 rejects the direct FTS5 shadow-table writes some repair tests use to synthesize corruption) and are identical to a clean `develop` checkout.

PR #1620, opened 2026-05-25
fix(hallways): paginate compute_hallways_for_wing to survive large wings

### LINKED ISSUE #1619: compute_hallways_for_wing crashes (too many SQL variables) on wings > ~32k drawers
## Summary
The within-wing hallways post-mine step added in #1558 fetches the entire wing in a single `col.get(where={"wing": wing})`. ChromaDB binds one SQL variable per matching id, so on a wing larger than SQLite's `SQLITE_MAX_VARIABLE_NUMBER` (32766) the call raises `InternalError: ... too many SQL variables` inside chromadb. The exception is caught (the mine still completes), but the wing's hallway graph **silently never builds** — on precisely the large wings that most benefit from hallway navigation. Cross-wing tunnel promotion (#1565), which builds on hallways, is starved for those wings too.

## Environment
- MemPalace 3.3.6, chromadb 1.x (rust bindings), macOS (Apple Silicon). Reproduces independently of the embedding model.

## Reproduction
1. Grow a palace so one wing exceeds ~32,766 drawers.
2. Mine into that wing so the post-mine `compute_hallways_for_wing` step runs.
3. The hallway step logs a warning + traceback and the wing's hallways are not computed.

Confirmed threshold: a **42,062**-drawer wing crashes; a **29,629**-drawer wing succeeds.

## Traceback (abridged)
```
compute_hallways_for_wing: collection.get failed for <wing>
  File "mempalace/hallways.py", line 203, in compute_hallways_for_wing
    results = col.get(where={"wing": wing}, include=["metadatas"])
  File "mempalace/backends/chroma.py", line 1132, in get
    raw = self._collection.get(**kwargs)
  File ".../chromadb/api/rust.py", line 440, in _get
    rust_response = self.bindings.get(...)
chromadb.errors.InternalError: Error executing plan: Internal error:
  error returned from database: (code: 1) too many SQL variables
```

## Root cause
`hallways.py:203` does an unbatched `col.get(where={"wing": wing})`. This is the same `SQLITE_MAX_VARIABLE_NUMBER` hazard already fixed by paginating the fetch in `miner.py` (#851) and `regenerate_closets` (#1073, #1107), both of which carry explicit comments about the 32766 limit. The new #1558 code path didn't inherit the pattern. Sibling 3.3.6 features are fine: `dynamics.py` doesn't fetch drawers, and `palace_graph.py` tunnels already paginate via `limit`/`offset`.

## Proposed fix
Mirror the established pattern: walk the collection in `batch_size=5000` pages via `count()` + `get(limit=, offset=)` and filter to the wing client-side. Validated against a real 42,062-drawer wing — previously crashed; now completes in ~3.7s computing 5,525 hallways. (PR to follow.)

## Severity
Non-fatal (the mine completes) but a silent capability loss that scales with palace size — the hallway graph, and the tunnels promoted from it, never build for large wings.

davidglidden:
## What
Paginate `compute_hallways_for_wing`'s drawer fetch so it survives wings larger than SQLite's `SQLITE_MAX_VARIABLE_NUMBER` (32766).

Fixes #1619.

## Why
The within-wing hallways post-mine step (#1558) fetches the whole wing in a single `col.get(where={"wing": wing})`. ChromaDB binds one SQL variable per matching id, so once a wing exceeds ~32k drawers the call raises `InternalError: ... too many SQL variables`. The exception is caught (the mine completes) but the wing's hallway graph **silently never builds** — on exactly the large wings that benefit most; cross-wing tunnel promotion (#1565) is starved for them too. This is the same hazard already fixed by paginating in `miner.py` (#851) and `regenerate_closets` (#1073/#1107); the new #1558 path just didn't inherit the pattern.

## How
Walk the collection in `batch_size=5000` pages via `count()` + `get(limit=, offset=)` and filter to the wing client-side — mirroring the existing idiom in `miner.py`'s status counter. Sibling features are unaffected (`dynamics.py` doesn't fetch drawers; `palace_graph.py` tunnels already paginate via limit/offset).

## Validation
- Against a real **42,062-drawer** wing: previously raised "too many SQL variables"; now completes in ~3.7s computing 5,525 hallways.
- `tests/test_hallways.py`: mock updated for the `count()` + paginated-`get` path; **all 21 tests pass**, `ruff` clean.

## Scope
One function (`compute_hallways_for_wing`) + its test. No behavior change for wings under the limit.

### DIFF
--- a/mempalace/hallways.py
+++ b/mempalace/hallways.py
@@ -198,16 +198,31 @@ def compute_hallways_for_wing(
 
     min_count = max(1, int(min_count))
 
-    # 1. Query drawers for this wing.
+    # 1. Query drawers for this wing. Paginate the fetch — a single
+    #    get(where={"wing": wing}) binds one SQL variable per matching id and
+    #    trips SQLite's SQLITE_MAX_VARIABLE_NUMBER (32766) inside chromadb once
+    #    the wing exceeds ~32k drawers, crashing the post-mine hallway step on
+    #    exactly the large wings that most need it. Mirrors the paginated fetch
+    #    already used in miner.py (#851) and closet_llm.py (#1073): walk the
+    #    collection in batches of 5000 and filter to this wing client-side.
+    metadatas: list = []
     try:
-        results = col.get(where={"wing": wing}, include=["metadatas"])
+        total = col.count()
+        batch_size = 5000
+        offset = 0
+        while offset < total:
+            batch = col.get(limit=batch_size, offset=offset, include=["metadatas"])
+            batch_metas = (batch or {}).get("metadatas") or []
+            if not batch_metas:
+                break
+            metadatas.extend(m for m in batch_metas if (m or {}).get("wing") == wing)
+            offset += len(batch_metas)
     except Exception:
         logger.warning(
             "compute_hallways_for_wing: collection.get failed for %s", wing, exc_info=True
         )
         return []
 
-    metadatas = (results or {}).get("metadatas") or []
     if not metadatas:
         return []
 
--- a/tests/test_hallways.py
+++ b/tests/test_hallways.py
@@ -27,10 +27,16 @@ def _use_tmp_hallway_file(monkeypatch, tmp_path):
 
 
 def _fake_collection(drawers):
-    """Build a MagicMock collection whose .get() returns the given drawer set."""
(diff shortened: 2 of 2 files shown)


igorls:

PR #1630, opened 2026-05-26
fix(layers): order L1 wake-up by recency (filed_at) so it surfaces the latest moments

### LINKED ISSUE #1629: L1 wake-up ranks by an importance field nothing populates → silently returns oldest moments
## Summary

`MemoryStack` L1 (`layers.py`, the `mempalace wake-up` context) is documented to *"prefer high importance, recent filing"*, but in practice it surfaces the **oldest** moments in a wing — neither important nor recent.

## Root cause

`Layer1.generate()` scores each drawer by `importance` / `emotional_weight` / `weight`, then sorts:

```python
# Score each drawer: prefer high importance, recent filing
importance = 3
for key in ("importance", "emotional_weight", "weight"):
    ...
scored.sort(key=lambda x: x[0], reverse=True)
```

But **no ingest path writes any of those keys.** On a real palace (45,969 drawers), 0 rows carry `importance`, `emotional_weight`, or `weight` in `embedding_metadata`. `emotional_weight` appears in the codebase only in `dialect.py`, where it is *read* (`.get("emotional_weight", 0.5)`), never written as drawer metadata.

So `importance` defaults to `3` for every drawer, every entry ties, and the stable sort leaves them in insertion order → **oldest first**. The "recent filing" half of the docstring is never honored: `filed_at` is not consulted at all.

**Effect:** a scoped `mempalace wake-up --wing <wing>` returns the *oldest* N moments of the wing — the opposite of the intended "essential story".

## Suggested fix

`filed_at` is present on every drawer (ISO-8601, so lexically chronological). Add it as the secondary sort key — `sort by (importance, filed_at) desc`. Importance stays primary for the day a scoring pass populates it; `filed_at` becomes the effective signal today, making "recent filing" true with data already present.

PR incoming from `davidglidden:fix/l1-recency-ordering`.

davidglidden:
Fixes #1629.

## Why

`layers.py` L1 (`mempalace wake-up`) is documented to *"prefer high importance, recent filing"*, but it surfaces the **oldest** moments in a wing. Root cause (#1629): it sorts by `importance`/`emotional_weight`/`weight`, but no ingest path writes those fields — 0 of 45,969 drawers in a real palace carry any of them — so every drawer ties at the default `3` and the stable sort collapses to insertion order (oldest first). `filed_at` was never consulted, despite the docstring's promise.

## What

Add `filed_at` (present on every drawer, ISO-8601 → lexically chronological) as the secondary sort key: `(importance, filed_at)` descending. Importance stays primary for the day a scoring pass populates it; `filed_at` is the effective signal today.

## How to verify

- `mempalace wake-up --wing <a wing spanning more than one day>` — entries are now newest-first instead of oldest-first.
- `pytest tests/test_layers.py` — 44/44 pass.

## What was not changed

- No new metadata is written and no ingest path is touched. This fixes only the *consumer's* ordering, using data already present.
- `importance` remains the primary sort key, so a future scoring pass works unchanged.

## Known limitation (out of scope)

The fetch still scans up to `MAX_SCAN=2000` drawers in collection (insertion) order, so on a very large *unscoped* wing the most-recent drawers may fall outside the cap before sorting. Recency ordering is correct for scoped wings that fit within `MAX_SCAN`. A SQL-side "most-recent-N by `filed_at`" fetch would fix the unscoped case but is a larger change left for a follow-up.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


igorls (approved):
Verified locally against current develop: clean merge, full layers test suite (83 tests) passes, and CI is green across the matrix. The recency tiebreak via filed_at is well-reasoned — importance stays the primary key for when a scoring pass populates it, and filed_at gives a meaningful order today. Thanks!

PR #1631, opened 2026-05-27
fix(embedding): add embed_query/embed_documents to EmbeddinggemmaONNX for ChromaDB 1.5.x

YuRiNjjot:
## Bug

ChromaDB 1.5.x calls `embedding_function.embed_query(input=...)` via keyword argument during `collection.query()`. `EmbeddinggemmaONNX` lacked both `embed_query` and `embed_documents` methods, causing:

```
TypeError: embed_query() got an unexpected keyword argument input
```

whenever semantic search was triggered.

## Fix

This PR adds the two methods required by the ChromaDB EF protocol, using `input` (the ChromaDB kwarg name, noqa A002) so that palace search works correctly with the `embeddinggemma` model.

Also downloads the companion `_data` ONNX file alongside the main model to prevent runtime InferenceSession failures.

## Related

Fixes silent search failures when `embedding_model` is set to `"embeddinggemma"`.

### DIFF
--- a/mempalace/embedding.py
+++ b/mempalace/embedding.py
@@ -191,6 +191,9 @@ class EmbeddinggemmaONNX:
         model_path = hf_hub_download(
             _EMBEDDINGGEMMA_REPO, subfolder="onnx", filename=_EMBEDDINGGEMMA_ONNX
         )
+        hf_hub_download(
+            _EMBEDDINGGEMMA_REPO, subfolder="onnx", filename=_EMBEDDINGGEMMA_ONNX + "_data"
+        )
         tok_path = hf_hub_download(_EMBEDDINGGEMMA_REPO, filename="tokenizer.json")
 
         self._session = ort.InferenceSession(model_path, providers=self._providers)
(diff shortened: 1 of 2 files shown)


igorls:
Maintainer push: added one commit (`48f4eb6`) updating `test_lazy_load_runs_once` to expect **3** `hf_hub_download` calls instead of 2. Your fix correctly adds the `model.onnx_data` external-weights download (the model can't load weights without it), so a single warm-up now legitimately issues 3 downloads. The lazy-load-once invariant still holds — `InferenceSession` and `Tokenizer.from_file` are each still built exactly once. Full `test_embeddinggemma.py` is green locally (10 passed). The `embed_query`/`embed_documents` EF-protocol additions look correct for ChromaDB 1.5.x. Thanks!

mvalentsev:
#1667 independently adds the same `model_quantized.onnx_data` sibling fetch (sidecar only). Flagging the overlap for dedup. This PR also bumps `test_lazy_load_runs_once` for the extra `hf_hub_download` and adds the `embed_query`/`embed_documents` methods, which #1667 doesn't.

Issue #2, opened 2026-04-06
Integrating MemPalace with SoulForge's code intelligence system


proxysoul:
Hi Milla & Ben,

Im Ouail, I built SoulForge, an AI coding agent. When I saw MemPalace I had to reach out because we're solving two halves of the same problem.

My frustration with other agents was the same as yours with memory. They grep the entire codebase on every prompt, stuffing context blindly. So I built a live dependency graph ("Soul Map") that tells the agent exactly where everything is. AST-parsed, PageRank-ranked, rebuilt after every edit. Zero guessing.

But I have no memory system. Every session starts from scratch. That's your whole thing.

I'm going to experiment with integrating MemPalace as the persistent memory layer for SoulForge. Soul Map handles "what is the code right now," MemPalace handles "what happened and why." Feels like they belong together.

PS: Huge fan!! 

Best regards, 
Ouail
----
https://github.com/proxysoul/soulforge

bensig:
Hey Ouail, this is awesome — SoulForge's live dependency graph + mempalace's persistent memory feels like a natural fit. "What is the code right now" vs "what happened and why" is exactly the right framing.

Would love to see what you come up with. If you run into integration questions, feel free to open issues here. And if it makes sense down the road, we'd be open to an official integration (similar to the Hermes PR that's in review right now).

Thanks for reaching out!

proxysoul:
@bensig  a PR has been created & integration on SoulForge is implemented, works flawlessly!

PR #1632, opened 2026-05-27
feat: add Cursor IDE support (hooks, plugin, skill, docs, tests)

undeadindustries:
## Summary

Adds first-class Cursor IDE support alongside the existing Claude Code and Codex flows: auto-save hook, pre-compaction transcript capture, session-start memory recall, plus a Cursor plugin (skill + slash commands + MCP auto-registration). Same privacy model as the existing Claude hooks — local-only, zero external API, strictly additive.

## What's added

- **`hooks/cursor/`** — three hook scripts (`stop`, `preCompact`, `sessionStart`), shared `lib/common.sh`, idempotent `install.sh`, payload schema reference.
- **`.cursor-plugin/` + repo-root `commands/`, `skills/`, `mcp.json`** — plugin manifest, marketplace entry, model-invocable skill, five `mempalace-*` slash commands, MCP server auto-registration.
- **`examples/cursor/`** — full and minimal `hooks.json` examples plus README explaining variant choice.
- **`website/guide/cursor-hooks.md`** — VitePress guide page wired into the sidebar.
- **`README.md`, `CHANGELOG.md`, `hooks/README.md`** — updates noting the new path.
- **129 new tests** — `tests/test_cursor_hooks_shell.py` (75), `tests/test_cursor_hooks_install.py` (19), `tests/test_cursor_plugin_manifest.py` (35).

## Notable design decisions

- **Fail-open hooks**: hook scripts do not use `set -e`; a broken hook can never block the user's conversation.
- **`preCompact` capture strategy**: Cursor's `preCompact` cannot return a follow-up message, so we synchronously mine the transcript before compaction and drop a `pending_save` marker that the next `stop` hook consumes. Guarantees verbatim capture before context compression.
- **Plugin discovery**: verified against the cached Cloudflare plugin that Cursor requires real `commands/`, `skills/`, `mcp.json` at the plugin root. `.cursor-plugin/{commands,skills}` are convenience symlinks back to those canonical locations.
- **bash 3.2 compatibility**: avoids heredoc-in-command-substitution parser bugs; uses `python -c` for JSON; basename-matched entry recognition in `install.sh` so re-installs across paths work.
- **Strictly additive**: no existing files removed, no existing hook behaviour changed, no new runtime dependencies introduced.

## Test plan

- [x] `uv run pytest tests/ --ignore=tests/benchmarks` — 2396 passed, 3 skipped (pre-existing).
- [x] `uv run pytest tests/test_cursor_hooks_shell.py tests/test_cursor_hooks_install.py tests/test_cursor_plugin_manifest.py -v` — 129 passed.
- [x] `uv run ruff check .` — clean.
- [x] `uv run ruff format --check .` — clean (139 files).
- [x] Manual: installed hooks + plugin into local Cursor, verified `stop` hook fires (`cursor_hook.log` populated), MCP server reachable, `@mempalace` skill loads.

## Out of scope / known limitations

- Slash commands (`/mempalace-*`) only appear in the picker once the plugin is published to the Cursor marketplace; local (`~/.cursor/plugins/local/`) symlinked plugins don't surface commands in current Cursor builds. The plugin files are structured correctly for marketplace publication. The MCP server and `@mempalace` skill paths work today.

## Compatibility

- Existing Claude Code and Codex hook flows are untouched.
- No new runtime dependencies.
- Bash 3.2 (macOS default) compatible.


undeadindustries:
Thanks for the thorough review — all five suggestions applied in `df5db57`. Summary:

| # | Severity | File | Fix |
|---|---|---|---|
| 1 | high | `hooks/cursor/lib/common.sh` | Replaced `python3 - <<'PYEOF' ... PYEOF` inside `$(...)` with `python -c '...'` passing the config path as `argv[1]`. Matches the pattern already used in `mempal_parse_stdin` in the same file. |
| 2 | high | `hooks/cursor/install.sh` | Added a `case "$INSTALL_DIR" in /*) ;; *) INSTALL_DIR="$PWD/$INSTALL_DIR" ;; esac` block right after path resolution so relative `--install-dir` values are absolutised before they're written into `hooks.json`. |
| 3 | high | `hooks/cursor/mempal_save_hook_cursor.sh` | Extended the sanitiser case to `''\|*[!0-9]*\|0` so `MEMPAL_SAVE_INTERVAL=0` is coerced to the default (15) and the modulo check never divides by zero. |
| 4 | medium | `hooks/cursor/install.sh` | Inlined `EMPTY_CHECK_PY` as `python -c '...'`. No tmpfile → no leak window if the script is interrupted. |
| 5 | medium | `hooks/cursor/install.sh` | Switched `mktemp -t prefix` to the portable `mktemp "${TMPDIR:-/tmp}/...XXXXXX"` form. |

Two new regression tests were added so we can't backslide:

- `tests/test_cursor_hooks_shell.py::TestSaveHookCounter::test_save_interval_zero_is_coerced_to_default` — guards fix #3.
- `tests/test_cursor_hooks_install.py::TestInstallDirAbsolutePath` (2 cases) — guards fix #2 plus the inverse (absolute paths must be preserved verbatim).

Verification:

- `bash -n` clean on all three edited scripts.
- 132/132 Cursor-specific tests pass (was 129).
- Full suite: 2399 passed, 3 skipped (pre-existing).
- `ruff check` and `ruff format --check` clean.

igorls:
## Review

Unusually careful, genuinely defensive shell work — fail-open on every path, layered kill switches, sentinel-guarded Python JSON parsing with input sanitization, bash 3.2.57 compat throughout, `..`-traversal rejection, atomic counter writes (temp+`mv`), idempotent installer, malformed-JSON-refusing `hooks.json` merge that preserves unrelated entries, basename-guarded uninstall. Privacy is honored and the additive claim holds. The notes below are mostly philosophy and hygiene, not bugs.

### 🟠 The `followup_message` default — why is it needed?

For the **core verbatim guarantee it isn't.** The `stop` hook already does the real capture in the background — `mempalace mine "$(dirname "$MEMPAL_TRANSCRIPT")" --mode convos &` files verbatim drawers. The `followup_message` runs *in addition*; its only distinct contribution is LLM-synthesized curation (entity-aware drawers + an AAAK **diary** entry) — the index/summary layer, not the sacred verbatim layer.

The decisive comparison is the existing canonical hook, `hooks/mempal_save_hook.sh:306-321`. It produces that same curation artifact (`{"decision":"block","reason":"...write a session diary..."}`) — but **gated behind `MEMPAL_VERBOSE`, defaulting to `false`**, with the comment spelling out why:

```
# false = silent mode (default) — save in background, no chat clutter
```

i.e. the house integration deliberately makes the LLM-driven save **opt-in, off by default**, to honor CLAUDE.md's *"Background everything — zero tokens spent on bookkeeping in the chat window."* This PR **inverts that default** — firing `followup_message` unconditionally every Nth turn with no verbose gate. That, not the followup itself, is the tension.

The one scenario where it'd be load-bearing: if Cursor's transcript isn't in a format `convo_miner`/`normalize.py` can read at `stop` time, the background mine captures nothing and the followup is the *only* working capture path. If that's the reason, please say so — and note the `--mode convos` line is then effectively dead.

**Question for you:** is the followup there for the diary/curation layer (→ gate it behind a `MEMPAL_VERBOSE`-style flag, off by default, matching the Claude hook), or because Cursor's transcript isn't machine-minable (→ document that, and the background-mine comment is misleading)? Either way the fix is cheap: adopt the existing `MEMPAL_VERBOSE` gate so Cursor's default matches the Claude hook's default-silent behavior.

### 🟡 Hygiene

- **Hardcoded `"version": "3.3.6"`** in both `.cursor-plugin/plugin.json` and `marketplace.json`. CLAUDE.md makes `version.py` the single source of truth — these will silently drift next release. Generate them, or omit the field (the Antigravity plugin.json omits it entirely).
- **Committed symlinks** (`.cursor-plugin/commands → ../commands`, `.cursor-plugin/skills → ../skills`). On Windows clones with `core.symlinks=false` they materialize as plain text files containing the target → broken plugin. The sibling Antigravity PR shipped real files specifically to avoid this; CI won't catch it (manifest tests skip Windows).
- **Repo-root pollution.** Cursor's discovery forces `commands/`, `skills/`, `mcp.json` into the *repository root* (plugin root == repo root via the local symlink). Coupling a Python project's top level to one IDE's plugin layout is worth questioning — could the plugin live in a subdir with `marketplace.json`'s `source` pointing at it?
- **`preCompact` synchronous mine** is the right call (compaction is irreversible) but blocks the hook and can exceed Cursor's per-hook timeout on a large transcript → Cursor may kill it mid-mine. Incremental mining makes that recoverable; worth a note.

### Cross-cutting (also raised on the Antigravity PR #1633)

- **Unbounded state-file growth.** Per-conversation counter + `.pending` files in `~/.mempalace/hook_state/` are never GC'd — one+ artifact per conversation, forever. A TTL sweep would help. Low severity.
- **Python cold-start vs. perf budget.** Every fire spawns Python ≥1× (parse), often 2× (kill-switch `config.json` read). `sessionStart`/`stop` fire often; ~30–80ms ×2 of interpreter startup makes a 100ms budget optimistic. Worth measuring.
- **Merge collision with #1633.** Both edit `website/.vitepress/config.mts`, `README.md`, `CHANGELOG.md`, `hooks/README.md`. Sequence the merges.

### Minor
- `mempal_infer_wing` is thorough (Windows backslash, lone `/`, trailing slash, underscore-collapse). Note it yields a bare basename (default `cursor_session`) while Antigravity uses `wing_<slug>` — so the same workspace mined from both IDEs lands in different wings. Intentional per-IDE, but a cross-IDE consistency wrinkle.

Mergeable in principle. The only "decide before merge" item is the `followup_message` default; the version-drift and Windows-symlink items are quick hygiene fixes.

undeadindustries:
Thanks for the thorough review, @igorls — all addressed on `feat/cursor-hooks-support` (pushed). Cursor-only scope here; the cross-IDE items are coordinated on the Antigravity branch/PR.

**1. `followup_message` default (decide-before-merge).** Keeping it **on by default**, with a documented opt-out. The deciding evidence for your binary question: `mempalace/normalize.py` has no Cursor parser and Cursor's transcript format is undocumented (`hooks/cursor/STDIN_SHAPE.md`), so the background `mempalace mine --mode convos` is best-effort only and does **not** yet yield clean verbatim drawers. The followup is therefore the load-bearing verbatim-capture path — defaulting it off would make a default Cursor install capture nothing. Users who want the Claude-style silent behaviour set `MEMPAL_CURSOR_SILENT=1` (or `MEMPAL_VERBOSE=false`). I corrected the misleading hook/README/guide comments that implied the mine was primary. **Tracked follow-up:** add a Cursor transcript parser to `normalize.py`; once it lands the default flips to silent to match the Claude hook.

**2. Hardcoded plugin version.** Removed from `plugin.json` + `marketplace.json`; `mempalace/version.py` is the single source of truth and a test now enforces the field stays absent.

**3. Committed symlinks (Windows).** Removed `.cursor-plugin/{commands,skills}` — redundant with the real repo-root `commands/`/`skills/`/`mcp.json` that `source: "."` already serves. Added a no-symlink guard test. The repo-root component layout is required by Cursor's plugin-root discovery (verified against the cached Cloudflare plugin), so I kept it and documented the rationale rather than restructure something I can't test against a live marketplace.

**4. preCompact synchronous-mine timeout.** Documented: the mine is bounded by Cursor's per-hook timeout, and because `mempalace mine` is incremental/append-only a killed mine resumes cleanly next run (no corruption). The pending-save marker is the backstop.

**5. Unbounded state growth.** Added a daily-throttled, Cursor-namespaced TTL sweep (`MEMPAL_STATE_TTL_DAYS`, default 30) after the kill-switch check; touches only `cursor_*.count`/`cursor_*.pending`, never shared logs or `antigravity_*`.

**6. Python cold-start vs perf budget.** Acknowledged as measure-then-decide — no speculative reordering in this PR.

**Deferred (cross-IDE):** `mempal_infer_wing` naming convention and shared-file merge sequencing with #1633 are handled on the Antigravity branch.

Verification: full suite green (2424 passed, 3 skipped), `ruff check` + `ruff format --check` clean, `bash -n` clean on all Cursor scripts. ~30 new tests cover the followup opt-out, the state-file GC (removal/preservation/throttle/kill-switch gating), TTL validation, and the no-symlink/version-absent manifest guards.

PR #1633, opened 2026-05-27
feat: add Antigravity IDE support (hooks, plugin, skill, docs, tests)

undeadindustries:
## Summary

Adds first-class integration with Google's [Antigravity IDE](https://antigravity.google/) as a third sibling to the existing Claude Code and Codex hook integrations.

This is **strictly additive**: no existing files in `main` are restructured. The integration scope is bounded by what Google's [official Antigravity docs](https://antigravity.google/docs/) actually expose — the full audit (with verbatim quotes + URLs + dates + omitted-surface rationale) is in [`hooks/antigravity/INVESTIGATION.md`](hooks/antigravity/INVESTIGATION.md).

### What ships

| Surface         | File / location                                              |
|-----------------|--------------------------------------------------------------|
| Plugin manifest | `.antigravity-plugin/plugin.json` — verified minimal shape   |
| MCP server      | `.antigravity-plugin/mcp_config.json` — registers `mempalace-mcp` stdio |
| Skill           | `.antigravity-plugin/skills/mempalace/SKILL.md` — real file (no symlinks) |
| Stop hook       | `hooks/antigravity/mempal_save_hook_antigravity.sh` — counter + auto-mine |
| PreInvocation hook | `hooks/antigravity/mempal_wake_hook_antigravity.sh` — gated to `invocationNum==1` |
| Installer       | `hooks/antigravity/install.sh` — idempotent, `__PLUGIN_DIR__` substitution, basename-guarded uninstall |
| Documentation   | `website/guide/antigravity.md` + sidebar entry in `config.mts` |
| Standalone examples | `examples/antigravity/{hooks.json,mcp_config.json,README.md}` |
| Tests           | 56 new tests across 3 test files                             |

### What is deliberately NOT shipped (with reasons)

Documented in [`hooks/antigravity/INVESTIGATION.md`](hooks/antigravity/INVESTIGATION.md):

- **`PreCompact` equivalent** — no external Antigravity surface; only the in-process Python SDK has `@hooks.on_compaction`. The `Stop` hook still catches end-of-turn state; auto-compaction mid-turn loses some content but the verbatim transcript ingestion covers long-term recall.
- **Slash-commands / `commands/`** — Antigravity has no `commands/` plugin component. Folded into `SKILL.md`'s `## Common operations` section instead.
- **`rules/`** — would risk colliding with user project rules. Users opt in by dropping their own under `.agents/rules/`.
- **`permissions` field in `plugin.json`** — fabricated by the third-party `antigravity-plugins` community skill; no real Google-shipped plugin uses it. We pin to the verified-minimal shape and have a regression test.
- **Workspace-scoped install by default** — the global install at `~/.gemini/config/plugins/mempalace/` is the canonical UX. Workspace-scoped install is documented in `hooks/antigravity/README.md`.
- **`PreToolUse` / `PostToolUse` hooks** — out of scope for v1, surface is real and noted for future work.

### Constraints honoured

- **bash 3.2.57** (macOS default): no `mapfile`/`readarray`/`declare -A`/`${var^^}`; sentinel-guarded JSON parser uses `sed -n 'Np'` like the existing Claude Code hook.
- **Verbatim guarantee**: wake injection passes `mempalace wake-up` stdout through unchanged via `json.dumps`; never paraphrased.
- **Performance budgets**: hook scripts return `{}` in <500ms on every kill-switch / gate path (test asserts <1.5s on CI), wake hook enforces a 500ms hard timeout on `mempalace wake-up`.
- **Zero new runtime dependencies**: the hooks only use `bash` + `python3` (already required by every existing hook).
- **Privacy**: no telemetry, no external API, no phone-home.
- **Strictly additive**: existing Claude/Codex hooks unchanged; existing tests continue to pass.

### Critical safeguards

- **Stop hook NEVER emits `{"decision":"continue"}`**. That output would force Antigravity into an infinite agent re-execution loop. The save hook's `mempal_emit_stop_pass` hard-codes `printf '{}\n'`. Tested directly in `test_save_hook_never_emits_decision_continue`.
- **`MEMPAL_SAVE_INTERVAL` floored to >= 1**. A user setting `MEMPAL_SAVE_INTERVAL=0` (thinking it disables saves) would otherwise bash-divide-by-zero. Tested in `test_save_hook_floors_zero_save_interval_to_avoid_div_by_zero`.
- **`fullyIdle == false` defers**. Background commands still running means the transcript is in motion; better to skip than ingest a half-finished transcript.
- **Uninstall is basename-guarded AND plugin.json-name-guarded**. Refuses to wipe a directory whose basename isn't `mempalace`, refuses if the directory has no `plugin.json` or names a different plugin. Tested in three separate refusal tests.
- **`__PLUGIN_DIR__` substitution + relative-path absolutization**. The cursor PR review caught that a relative `--install-dir` would bake a relative path into `hooks.json` that Antigravity couldn't resolve at runtime. Tested in `test_relative_install_dir_is_absolutized`.

### Test plan

- [x] `uv run pytest tests/ --ignore=tests/benchmarks -v` → **2314 passed, 3 skipped, 1 unrelated warning**
- [x] `uv run ruff check .` → all checks passed
- [x] `uv run ruff format --check .` → 139 files already formatted
- [x] `bash -n` clean on `lib/common.sh`, save hook, wake hook, `install.sh`
- [x] Local install at `~/.gemini/config/plugins/mempalace/` succeeds
- [x] All 8 expected files present after install with correct executable bits
- [x] Rendered `hooks.json` carries absolute paths (no `__PLUGIN_DIR__` leak)
- [x] Both hooks fire correctly with realistic camelCase JSON payload
- [x] Wing inference picks `wing_mempalace` from `workspacePaths[0]`
- [x] State files all `antigravity_*`-namespaced (no Claude/Codex collision)
- [x] Second `install.sh` run produces byte-identical output (md5 snapshots match)
- [x] Uninstall removes only `mempalace`, leaves all 6 sibling Google plugins (`firebase`, `chrome-devtools-plugin`, `google-antigravity-sdk`, `android-cli-plugin`, `custom-engineering-skills`, `modern-web-guidance-plugin`) untouched
- [x] `mempalace-mcp` is on `$PATH`
- [ ] Smoke-test the rendered plugin against a running Antigravity IDE — confirm MCP server `mempalace` shows in the MCP store, skill `mempalace` appears in the skill list, Stop event fires save hook on conversation end, PreInvocation event fires wake hook on first model call

### New tests (56)

- `tests/test_antigravity_plugin_manifest.py` — **11 tests** — schema contract on `.antigravity-plugin/`, including guards against fabricated fields and symlink leaks.
- `tests/test_antigravity_hooks_shell.py` — **31 tests** — invokes bash hooks via subprocess, asserts `{}` on every failure path, kill-switch coverage (env vars + `config.json` + palace nuke), divide-by-zero floor, traversal rejection, namespacing, wing inference, decision-continue refusal.
- `tests/test_antigravity_hooks_install.py` — **14 tests** — `--dry-run` side-effect-free, real install layout, executable bits preserved, byte-identical idempotent re-runs, basename-match uninstall safety + two further refusal modes, relative-path absolutization, unknown-arg failure. Skipped on Windows.

### Branching

Branched from `origin/main` (per pre-flight constraint), targets `develop` for normal feature flow.

### References

- [`hooks/antigravity/INVESTIGATION.md`](hooks/antigravity/INVESTIGATION.md) — full surface audit
- [`hooks/antigravity/STDIN_SHAPE.md`](hooks/antigravity/STDIN_SHAPE.md) — exact wire format per event with worked examples
- [`hooks/antigravity/README.md`](hooks/antigravity/README.md) — operator-side docs + troubleshooting
- [`website/guide/antigravity.md`](website/guide/antigravity.md) — user-facing guide

Made with [Cursor](https://cursor.com)


igorls:
## Review

Unusually careful, genuinely defensive shell work — fail-open on every path, layered kill switches, sentinel-guarded Python JSON parsing with input sanitization, bash 3.2.57 compat throughout, `..`-traversal rejection, dry-run, idempotent installer with cmp-gated byte-identical re-runs, and a basename-**and**-`plugin.json`-name-guarded uninstall that refuses to wipe anything that isn't a mempalace plugin. Privacy is honored (0600 on logs that echo paths), wake injection is verbatim via `json.dumps`, and `plugin.json` is minimal (`{"name":"mempalace"}`) — no version to drift. The notes below are mostly cross-cutting; few are bugs.

Cleaner philosophy than the sibling Cursor PR (#1632): the Stop hook only background-mines and never emits a token-spending followup — which is the right call under CLAUDE.md's *"Background everything — zero tokens in the chat window."*

### 🟡 Specifics

- **Counter write is not atomic, despite the comment saying it is.** `printf '%s' "$COUNT" > "$COUNTER_FILE"` truncates-then-writes. The Cursor PR's `mempal_write_counter_atomic` (temp + `mv`) is the correct pattern — match it here, or fix the "written atomically" comment. Low impact (concurrent Stop fires are unlikely) but the comment is wrong.
- **Synchronous `python -m mempalace --version` probe on the save path**, in the foreground *before* the mine is backgrounded. If the package import is heavy (ChromaDB et al.), that foreground probe alone can blow the <500ms save budget the brief sets. Move it inside the backgrounded subshell, or cache the result.
- ✅ `mempal_save_interval` octal-strip + zero-floor is correct and well-reasoned. The `kill -0` polling subshell to clean the pending marker (given the sibling-PID constraint that rules out `wait`) is clever and well-documented.

### Cross-cutting (also raised on the Cursor PR #1632)

- **Unbounded state-file growth.** Per-conversation counter + `.pending` files **and a `woke_<conv>` directory** in `~/.mempalace/hook_state/` are never GC'd — one+ artifact per conversation, forever. A TTL sweep would help. Low severity.
- **Python cold-start vs. perf budget.** Every fire spawns Python ≥1× (parse), often 2× (kill-switch `config.json` read). PreInvocation fires before *every* model call before the `invocationNum==1` gate; ~30–80ms ×2 of interpreter startup makes the brief's 100ms startup ceiling optimistic. Worth measuring.
- **Merge collision with #1632.** Both edit `website/.vitepress/config.mts`, `README.md`, `CHANGELOG.md`, `hooks/README.md`. Sequence the merges.

### Minor
- `mempal_infer_wing` yields `wing_<slug>` (default `wing_sessions`), while the Cursor PR uses a bare basename (default `cursor_session`) — so the same workspace mined from both IDEs lands in different wings. Intentional per-IDE, but a cross-IDE consistency wrinkle worth a deliberate decision.

Mergeable in principle — no blockers. The atomic-counter and foreground-`--version`-probe items are the two I'd want addressed; the rest is shared hygiene.

undeadindustries:
Thanks @igorls — pushed `df295bd` addressing the review. Scope here is the **Antigravity branch only**; cross-IDE/Cursor items are flagged for the Cursor PR (#1632) / maintainer coordination as noted below.

### Must-fix

**1. Non-atomic counter write** — Fixed. Added `mempal_write_counter_atomic()` to `lib/common.sh` (same-directory temp file + `mv -f`, an atomic rename on one filesystem, with a direct-write fallback if `mktemp` fails). The save hook now calls it instead of the truncate-then-write `printf > "$COUNTER_FILE"`, and the misleading "written atomically" comment now describes the real mechanism. The temp stays `antigravity_save_count_<conv>.XXXXXX`-namespaced. Tests: counter still increments correctly across fires and leaves no temp behind.

**2. Expensive foreground `--version` probe** — Fixed. Confirmed via `mempalace/cli.py` that the `mine` subparser imports `mempalace.miner` (→ palace → backends → chromadb/onnx) during parser construction, so `mempalace --version` pays the full cold-start import. The probe, the mine, and the pending-marker cleanup now run in **one detached subshell**; the foreground drops the marker, logs, spawns, and returns immediately. New timing test: with a stub interpreter that sleeps 3s on `-m mempalace`, the hook returns in <2s — proving the probe is off the foreground.

   This also **retires the `kill -0` watcher** from the prior gemini fix: marker cleanup is now sequential inside the mine's own subshell, so the sibling-PID `wait` hazard is structurally gone (no `wait`, no `kill -0`, no `MINE_PID` capture).

### Non-blocking hygiene (done in this PR)

**State-file GC** — Added `mempal_state_ttl_days()` (default 30, env `MEMPAL_STATE_TTL_DAYS`, integer-floored/leading-zero-stripped) and `mempal_gc_stale_state()`. The sweep is throttled to **at most once per 24h** via an `antigravity_last_sweep` marker (mtime check), so it adds a single stat to the common path. It removes stale `antigravity_save_count_*`, `antigravity_pending_*`, and `antigravity_woke_*` artifacts older than the TTL; the name globs are specific, so `antigravity_hook.log`, the input/error logs, and the sweep marker itself are never touched. BSD-find (macOS) and GNU-find compatible. Called right after the kill-switch check so a disabled hook touches nothing. Tests cover sweep, daily throttle, stale-marker re-run, TTL validation, and kill-switch gating.

### Measure-first / deferred

- **Python cold-start vs perf budget (PreInvocation)** — measure-then-decide. Phase work above already removes the heaviest foreground cost on the save path. The PreInvocation stdin parse spawns python once before the `invocationNum==1` gate; I'd rather measure that on real hardware than reorder speculatively in this PR.
- **`mempal_infer_wing` cross-IDE naming** (`wing_<slug>` / default `wing_sessions` here vs the bare basename / `cursor_session` in #1632) — this is a cross-IDE alignment decision that shouldn't be unilaterally changed on the Antigravity branch. Deferring to coordinate with the Cursor PR so both land on one convention.
- **Merge-sequencing of shared files** (`website/.vitepress/config.mts`, `README.md`, `CHANGELOG.md`, `hooks/README.md`) — acknowledged; these overlap with #1632 and should be sequenced at merge time rather than pre-empted here.

Full antigravity suite (74 tests) green; full repo suite (2331 passed, 3 skipped) green; `ruff check`/`format` clean; `bash -n` clean on all four shell files; local reinstall is byte-identical idempotent and both hooks return `{}` exit 0 on synthetic camelCase stdin.

igorls:
**Maintainer review** (on `df295bd`) — thanks for the thorough turnaround on the last round: atomic counter via `mempal_write_counter_atomic`, the backgrounded `--version` probe, and the throttled GC sweep all look good, and retiring the `kill -0` watcher is a nice structural simplification. Two things before merge:

1. **The new GC throttle uses GNU-only `date -r FILE`, which breaks on macOS** — the platform you explicitly target (bash 3.2.57). You were careful to make the GC *sweep* BSD/GNU-`find` compatible, but the once-per-24h throttle reads its marker mtime with `date -r "$marker" '+%s'` (the `mempal_gc_stale_state` check), and the pending-marker staleness guard does the same with `date -r "$PENDING_FILE" '+%s'`. On BSD/macOS `date -r` treats its argument as *epoch seconds*, not a file path, so both probes fail open: the throttle never returns early (it runs the 3-pass sweep on **every** Stop instead of once/day), and the pending guard falls through to `rm -f` — defeating the concurrent-save lock you just hardened in this same commit. Linux CI stays green, so the new throttle/marker tests don't catch it. A small portable helper — try `stat -f %m`, fall back to `stat -c %Y`, then `date -r` last — fixes both call sites and makes the "macOS compatible" claim hold.

2. **No normalizer for the Antigravity transcript schema (medium, still open).** The Stop hook runs `mempalace mine … --mode convos`, but `normalize.py` has no Antigravity detector, so an unrecognized `.jsonl` falls to paragraph chunking and stores raw JSON envelopes as drawers. Verbatim is preserved, but the drawers are noisy vs the Claude/Codex/Gemini paths. Either add a `_try_antigravity_jsonl()` detector (mirroring `_try_gemini_jsonl`) + a test, or scope-document the limitation prominently.

The deferred cross-IDE wing-naming and the shared-file merge-sequencing with #1632 are fine to settle at coordination/merge time — agreed those shouldn't be pre-empted here.

PR #220, opened 2026-04-08
fix: allow mempalace init to default to the current directory

2023Anita:
## Summary
- make the init directory argument optional and default it to the current directory
- expose a parser builder so the CLI behavior can be tested directly
- update docs to show the optional init directory form and add regression coverage

Closes #210.

## Validation
- PYTHONPATH=/Users/anita/mempalace pytest --noconftest tests/test_cli.py -q
- python3 -m compileall mempalace

PR #1639, opened 2026-05-28
fix: allow mempalace init to default to current directory

2023Anita:
## Summary

Allow `mempalace init` (without a directory argument) to default to the current directory.

## Root Cause

The `dir` argument was required, forcing users to always type `mempalace init .` even when working in the directory they want to initialize.

## Changes

- `cli.py`: make `dir` positional argument optional (`nargs="?"`, `default="."`)
- `cli.py`: extract `build_parser()` function, add `main(argv=None)` for programmatic use
- `README.md`: show default-dir usage in quickstart
- `tests/test_cli.py`: add `build_parser` import + tests for default and explicit dir parsing

## Verification

```
python -m pytest tests/test_cli.py::TestBuildParser -v
```

## Risk / Compatibility

Low. `dir` remains positional — `mempalace init /path` still works identically. The only behavioral change is that omitting the argument now defaults to `.` instead of showing an error.

Supersedes #220 (closed — branch was stale with merge conflicts).


igorls:

PR #1640, opened 2026-05-28
fix: force UTF-8 on stdin/stdout to handle non-ASCII content on Windows (closes #1488)

### LINKED ISSUE #1488: MCP server fails on Cyrillic/non-ASCII content on Windows due to default cp1252 stdin encoding
## Environment

- OS: Windows 11
- Python: 3.14.3
- mempalace: 3.3.3
- Client: Claude Desktop (latest)
- Locale: ru-RU (system default encoding cp1252)

## Description

All write operations (`mempalace_diary_write`, `mempalace_add_drawer`, `mempalace_kg_add`, `mempalace_create_tunnel`) and search (`mempalace_search`) fail with non-ASCII content (Cyrillic, accented Latin, CJK, etc.) on Windows when the MCP server is spawned by Claude Desktop with the default environment.

The MCP server's stdin reader uses Python's default locale-based encoding (cp1252 on a typical Windows install), and Claude Desktop sends JSON-RPC over stdio with UTF-8-encoded payloads. When non-ASCII bytes arrive, Python falls back to `surrogateescape` and stores undecodable bytes as lone surrogates (`\udcXX`) in the resulting `str`. Any downstream code that calls `.encode()` on that string (default UTF-8) raises `UnicodeEncodeError: 'utf-8' codec can't encode character '\udcXX': surrogates not allowed`. The HuggingFace `tokenizers` library used by the default ChromaDB embedder rejects such strings with `TextInputSequence must be str in query.`

The MCP layer catches the exception and surfaces it as the generic `-32000 Internal tool error` (for writes) or wraps it in the `Search error: ...` JSON payload (for searches). End users see two different symptoms but the root cause is one.

## Reproduction

1. Install on Windows: `pip install mempalace==3.3.3`
2. Configure in Claude Desktop `claude_desktop_config.json` with **default env** (no `PYTHONIOENCODING` set):
   ```json
   "mempalace": {
     "command": "python.exe",
     "args": ["-m", "mempalace.mcp_server"],
     "env": { "MEMPALACE_HOME": "..." }
   }
   ```
3. From Claude, call `mempalace_diary_write` with any Cyrillic content, e.g. `"Тест"`.
4. Observe: `Error: MCP error -32000: Internal tool error`.
5. Call `mempalace_search` with Cyrillic query: `"Search error: TextInputSequence must be str in query."`

## Stack trace (Cyrillic diary_write)

```
File "mempalace/mcp_server.py", line 1678, in handle_request
    result = TOOLS[tool_name]["handler"](**tool_args)
File "mempalace/mcp_server.py", line 947, in tool_diary_write
    f"{hashlib.sha256(entry.encode()).hexdigest()[:12]}"
UnicodeEncodeError: 'utf-8' codec can't encode character '\udc81' in position 5: surrogates not allowed
```

## Diagnostic

- Direct Python use of `chromadb` `collection.add()` and `collection.query()` with Cyrillic **succeeds** when run as a standalone script (because Python is started with proper UTF-8 source decoding).
- Same calls fail when routed through the MCP server spawned via Claude Desktop.
- Setting `PYTHONIOENCODING=utf-8` (or `PYTHONUTF8=1`, or running `python -X utf8`) in the MCP server's env restores correct behavior — Cyrillic writes and searches both work end-to-end.

## Workaround (user-side)

In `claude_desktop_config.json`:

```json
"mempalace": {
  "command": "python.exe",
  "args": ["-X", "utf8", "-m", "mempalace.mcp_server"],
  "env": {
    "MEMPALACE_HOME": "...",
    "PYTHONIOENCODING": "utf-8",
    "PYTHONUTF8": "1"
  }
}
```

Restart Claude Desktop. All non-ASCII operations work.

## Suggested upstream fix

Two complementary changes inside `mempalace`:

1. **Force UTF-8 stdio at server startup.** In `mcp_server.py` near the top (after the existing stdio protection block), add:

   ```python
   import sys
   try:
       sys.stdin.reconfigure(encoding='utf-8', errors='strict')
       sys.stdout.reconfigure(encoding='utf-8', errors='strict')
       sys.stderr.reconfigure(encoding='utf-8', errors='strict')
   except AttributeError:
       pass  # Python < 3.7 — fall back to PYTHONIOENCODING env var
   ```

   This makes the server robust regardless of the client's env-var hygiene.

2. **Defensive encoding for downstream `.encode()` calls.** Audit code paths that call `.encode()` without an explicit codec (e.g. `mcp_server.py:947`) and make them `.encode('utf-8')`. While the surrogate issue is upstream of these calls, the explicit codec is good hygiene and gives a clearer error if the issue ever recurs.

3. **README note**: document the `PYTHONIOENCODING=utf-8` requirement for Windows users of older Python versions, until the upstream fix lands.

## Impact

- All non-English MemPalace users on Windows hit this on first non-ASCII content.
- The wrapped `-32000 Internal tool error` provides no actionable signal — diagnosis requires reading server stderr or attaching a debugger.
- Workaround is one line of env config but completely undiscoverable without reading the traceback.

Happy to submit a PR if you'd like — fix is straightforward.

botbikamordehai2-sketch:
## What

On Windows, Python uses the system locale encoding (e.g. `cp1252`) for `sys.stdin` and `sys.stdout` by default. Claude Desktop sends JSON-RPC payloads over stdio encoded as UTF-8. When non-ASCII bytes (Cyrillic, CJK, accented Latin, etc.) arrive, Python's `cp1252` codec cannot decode them and falls back to `surrogateescape`, storing lone surrogates (`\udcXX`) in the resulting `str`. Downstream code — including the HuggingFace `tokenizers` library used by ChromaDB's default embedder — rejects strings containing lone surrogates, causing:

- Write tools (`mempalace_diary_write`, `mempalace_add_drawer`, etc.) to return `MCP error -32000: Internal tool error`
- `mempalace_search` to return `Search error: TextInputSequence must be str in query.`

## Fix

Call `sys.stdin.reconfigure(encoding='utf-8', errors='replace')` (and the same for `sys.stdout`) immediately after the existing stdio protection block, before any heavy imports or protocol I/O. `reconfigure()` is available on `TextIOWrapper` (Python 3.7+) and is a no-op-safe call — failures are silently ignored so this cannot introduce a new startup failure surface.

The `PYTHONIOENCODING=utf-8` workaround documented in some issues is still valid for advanced users, but this fix makes the server work correctly out of the box with the default Claude Desktop configuration.

Closes #1488


mvalentsev:
"Closes #1488" is already in develop since v3.3.5 (2026-05-10): `main()` in `mempalace/mcp_server.py:2831-2834` calls `stream.reconfigure(encoding="utf-8", errors="replace")` for both `sys.stdin` and `sys.stdout`, no win32 guard. See my comment on #1488 from 2026-05-13.

Base = `main`. CONTRIBUTING.md says PRs target `develop`.

Scope: of 7 changed files only `mempalace/mcp_server.py` (+24/-0) matches the title. The other six (`README.md`, new `data/known_systems.json` (+68), `entity_detector.py`, `palace.py`, `miner.py`, `tests/test_entity_detector.py`) add a Tier 3 known-systems lexicon for entity extraction, unrelated to UTF-8.

igorls:
**Maintainer review.** Confirming @mvalentsev's point above — this is **already fixed on develop** (since v3.3.5), so I'd recommend closing. Commit `35b033d` ("fix(mcp): force UTF-8 on stdio") already calls `reconfigure(encoding="utf-8", errors="replace")` on stdin/stdout inside `main()` (the correct spot, right before the JSON-RPC read loop). `git diff develop...` shows this PR's only effective change is 24 duplicate lines in `mcp_server.py` — the known-systems lexicon / README edits in the GitHub diff are stale-base noise already on develop.

The one non-redundant delta is reconfiguring `sys.stderr` to UTF-8 at import time (`main()` only does stdin/stdout). That changes non-ASCII *log* encoding on Windows, but it isn't part of the #1488 JSON-RPC transport fix. If it's worth keeping, a tiny PR adding just the stderr line on top of current develop would be the clean way; the rest here is a no-op duplicate. Thanks for chasing this either way.

PR #1643, opened 2026-05-28
fix(hooks): use MEMPAL_PYTHON_BIN for mine calls instead of bare mempalace

### LINKED ISSUE #1644: Shell hooks: bare `mempalace` call breaks in harness-launched environments (Claude Code, Codex CLI)
## Summary

Both shell hooks (`mempal_save_hook.sh`, `mempal_precompact_hook.sh`) fail silently in harness-launched environments that do not inherit the user's shell PATH. Every `mine` trigger produces:

```
/path/to/mempal_save_hook.sh: line 295: mempalace: command not found
```

All conversation memories are silently dropped as a result.

## Root cause

Commit 67a0677 switched from `python3 -m mempalace mine` to bare `mempalace`, with the rationale that pipx/uv put the entry point on PATH. This assumption does not hold for harness-launched processes — Claude Code and Codex CLI spawn hooks in a restricted environment that does not inherit the user's interactive shell PATH.

The irony: `MEMPAL_PYTHON` was introduced precisely to handle this case (harness environments without a reliable `python3` on PATH). The commit that removed `-m mempalace` also removed the `PYTHON` variable from `mempal_save_hook.sh`, breaking the very mechanism it was supposed to preserve.

## Affected configuration

Users who set `MEMPAL_PYTHON` in their harness environment (e.g. Claude Code `settings.json`):

```json
{
  "env": {
    "MEMPAL_PYTHON": "/path/to/venv/bin/python3"
  }
}
```

This is the documented workaround for GUI-launched or harness-launched Claude Code that cannot find the right `python3` on PATH. These users now have a working Python resolver but a broken `mempalace` CLI call — silent data loss on every save trigger.

## Reproducer

```bash
# Simulate Claude Code hook invocation (no shell PATH inheritance):
echo '{"session_id":"test","stop_hook_active":false,"transcript_path":"/path/to/session.jsonl"}' \
  | env -i MEMPAL_PYTHON=/path/to/venv/bin/python3 HOME=$HOME \
    bash hooks/mempal_save_hook.sh
# → line 295: mempalace: command not found
```

## Fix

Reuse the already-resolved `MEMPAL_PYTHON_BIN` variable for all four `mine` call-sites in both shell hooks — consistent with every other `python3` call in the same scripts.

See PR #1643.

tehw0lf:
Fixes #1644.

## Problem

Commit 67a0677 switched `python3 -m mempalace mine` to bare `mempalace` in both shell hooks, with the rationale that pipx/uv put the entry point on PATH.

This breaks harness-launched environments (Claude Code, Codex CLI) that do **not** inherit the user's shell PATH — the exact environments `MEMPAL_PYTHON` was introduced to handle. Users with `MEMPAL_PYTHON` set in their harness env (e.g. via Claude Code `settings.json`) see every mine trigger fail silently with:

```
/path/to/mempal_save_hook.sh: line 295: mempalace: command not found
```

This has been silently dropping all conversation memories since the commit landed (~6 weeks ago).

## Fix

Reuse the already-resolved `MEMPAL_PYTHON_BIN` variable (resolution order: `$MEMPAL_PYTHON` env override → `command -v python3` → bare `python3`) for all four `mine` call-sites in both shell hooks. This makes the mine invocations consistent with every other `python3` call in the same scripts.

Users relying on PATH-resolved `python3` are unaffected — `MEMPAL_PYTHON_BIN` falls back to `command -v python3` as before.

## Verified

Tested with `MEMPAL_PYTHON` pointing at a venv Python:
```bash
"$MEMPAL_PYTHON_BIN" -m mempalace mine <transcript-dir> --mode convos
# → MemPalace Mine output, files processed correctly
```


tehw0lf:
Closing — branch needs to target `develop` and be rebased cleanly. Will reopen.

PR #1648, opened 2026-05-29
fix(cli): force UTF-8 when reading/writing .gitignore in init

ALaDingAhmad:
## Summary

`mempalace init` crashes with `UnicodeDecodeError` on Windows when the target project's `.gitignore` contains non-ASCII bytes (e.g. Chinese comments, emoji), even when the file is valid UTF-8.

## Root cause

`mempalace/cli.py::_ensure_mempalace_files_gitignored` calls `Path.read_text()` and `open(path, "a")` without an explicit `encoding=`. Before PEP 686 / Python 3.15, both fall back to `locale.getpreferredencoding()`:

- Linux/macOS: typically UTF-8 — works fine
- **Windows: GBK / CP936 / other legacy MBCS** — crashes on any valid UTF-8 byte that isn't valid in the local code page

Real-world trigger from a Chinese-locale Windows user (auto-init invoked from a hook):

```
File ".../mempalace/cli.py", line 215, in _ensure_mempalace_files_gitignored
    existing = gitignore.read_text() if gitignore.exists() else ""
File ".../pathlib.py", line 1059, in read_text
    return f.read()
UnicodeDecodeError: 'gbk' codec can't decode byte 0x81 in position 37: illegal multibyte sequence
```

The `.gitignore` in question is a perfectly legal UTF-8 file that starts with `# ======== 绝不入库的敏感凭证 ========`.

## Fix

Pin both the read and append to `encoding=\"utf-8\"`. Add `errors=\"replace\"` on the read path as a defensive fallback for any legacy mixed-encoding `.gitignore` already in the wild — we only scan for line membership, so a replacement char in a comment is harmless and still lets init proceed.

## Test plan

- [x] Reproduces locally: \`mempalace init\` on a Windows project whose \`.gitignore\` contains Chinese comments now succeeds; previously aborted with \`UnicodeDecodeError\`
- [x] \`ruff check mempalace/cli.py\` passes
- [x] \`ruff format --check mempalace/cli.py\` passes
- [ ] CI green

## Notes

- 4-line change, no behavior change for ASCII-only \`.gitignore\`
- Affects every non-English Windows user whose project \`.gitignore\` carries native-language comments — silent for English-speaking users, which is likely why this hasn't surfaced upstream before


ALaDingAhmad:
FYI to maintainers: #1600 already covers this `.gitignore` UTF-8 fix as part of a much larger 100-file refactor (34 `open()` / `read_text()` call sites across 12 modules). That PR has been open for 5 days awaiting review.

This PR is intentionally minimal — just the 4 lines that fix the `auto-init` crash path I hit on a Chinese-locale Windows machine. If #1600 lands first, please close this. If #1600 is blocked on review scope, this can be merged independently to unblock Windows users hitting `_ensure_mempalace_files_gitignored` specifically — happy either way.

igorls (approved):
Verified locally against current develop: clean merge, and the init/gitignore tests pass on the integrated result. CI is now green across the matrix. Correct, minimal fix — forcing UTF-8 on both the read and append matches how the rest of init handles files. Thanks!

cddc2729-lab (commented):
- [ ] - - - - [ ] []()@jo

PR #1650, opened 2026-05-29
feat(sqlite): bound read-path statement timeout + connection lifecycle via _connect_ro

apollion69:
## What does this PR do?

Hardens the **read paths** of the SQLite/ChromaDB store against runaway statements, lock waits, and leaked connections, behind a single helper — without changing query behaviour.

A read-only review of the query layer found three gaps:

1. **No statement-timeout protection.** Of the `sqlite3.connect()` sites, only `knowledge_graph.py` set a busy timeout. Every read path (`searcher.py`, the `mcp_server` status read, `backends/chroma.py`, `repair.py`) opened with the default busy timeout and **no runaway-statement ceiling at all**. Against a multi-GB `chroma.sqlite3`, a pathological BM25/IN/metadata scan can run unbounded and block the MCP event loop.
2. **Connections leak.** Several sites use `with sqlite3.connect(...) as conn:`, which commits on exit but never `.close()`s — handles accumulate in a long-lived MCP server.
3. **No regression guard on parameterization.** Parameterization is already clean (no f-string/`%`/`.format()` user values in SQL), but nothing locks that state in.

SQLite has no `SET statement_timeout` (that is Postgres-only). The correct analogs are `PRAGMA busy_timeout` (lock-wait ceiling) plus a `set_progress_handler` wall-clock deadline (runaway-statement ceiling). This PR introduces both via one helper and routes all read paths through it.

**Honest scope:** this is a **safety/correctness** change, **not** a performance optimization. A paired benchmark (below) shows it is **latency-neutral** and **recall-neutral**; the read-tuning PRAGMAs that *could* speed things up gave no measurable gain on the warm path (it is OS-page-cache- and Python/chromadb-bound, not SQLite-I/O-bound) and were therefore left **off by default**.

### Thinking path (project context → this change)

The palace is a long-lived MCP server reading a single large `chroma.sqlite3` on the event loop. The project already treats unbounded resource use as a defect (e.g. the `RAYON_NUM_THREADS` cap in `repair`). The same discipline was missing one layer down at the SQLite connection: a single slow read could stall every tool call with no ceiling and no guaranteed cleanup. The minimal, upstream-shaped fix is to centralize read-connection construction so the ceiling and lifecycle are guaranteed once, at every read site, instead of being re-derived (or forgotten) per call.

## Related Issue

No pre-existing issue; the problem and evidence are documented inline above and in the benchmark results referenced below.

Closes #

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality) — adds a statement-deadline / lock-timeout / lifecycle guarantee; behaviour-preserving on normal queries
- [x] Refactor / code improvement (no behavior change) — read paths routed through one helper
- [ ] Documentation update
- [x] Tests (adding or improving test coverage)
- [ ] CI / infrastructure

## Changes Made

- **`mempalace/_sqlite_ro.py`** (new): `open_ro()` / `connect_ro()`. Applies a safety floor on every read connection — `PRAGMA busy_timeout` (lock-wait ceiling), `PRAGMA query_only = ON` (defense in depth), and a `set_progress_handler` wall-clock statement deadline that aborts a runaway statement instead of letting it run unbounded — and guarantees `.close()` (context manager, plus an `open_ro` drop-in for sites that own their own `try/finally`). `StatementTimeout` is exported as a `sqlite3.OperationalError` subclass. Read-tuning PRAGMAs (`mmap_size` / `cache_size` / `temp_store`) are **opt-in, off by default** (documented why in the module).
- **`mempalace/searcher.py`**: search read connection → `open_ro` (15 s interactive deadline).
- **`mempalace/mcp_server.py`**: status-read connection → `open_ro`.
- **`mempalace/backends/chroma.py`**: 3 read-only (`?mode=ro`) connections → `open_ro`.
- **`mempalace/repair.py`**: 3 read connections → `open_ro` (120 s batch deadline) / `connect_ro` (full-DB `quick_check` runs with no deadline — explicit maintenance op, also fixes a `with sqlite3.connect(...)`-never-closes leak).
- **`tests/test_sqlite_ro.py`** (new, 11 tests): deadline aborts a runaway recursive CTE within tolerance; context manager closes on success and on exception; `open_ro` caller-owns-close; safety-floor PRAGMAs always applied; read-tuning off-by-default + opt-in; `StatementTimeout` type; and an **AST parameterization regression guard** that fails if any read path interpolates a user/query value into SQL (the `",".join(["?"]*n)` IN-list is the one blessed dynamic pattern).

Out of scope (left as-is): write/repair-mutate connections and the `knowledge_graph.py` writer.

## How to Test

```bash
# Unit + the converted modules' existing suites
uv run pytest tests/test_sqlite_ro.py tests/test_searcher.py tests/test_repair.py tests/test_backends.py -v
# Full suite
uv run pytest tests/ -v
# Lint
ruff check .
```

Behaviour-equivalence and latency were measured against a copied snapshot of a ~2.6 GB live store (never the live store), comparing this branch to its base commit:

- **Latency (paired, 5 alternating rounds, p50 medians):** vector 97.9 → 107.9 ms, union 123.0 → 125.1 ms — all deltas inside the ±~40 ms run-to-run noise band (sign test n.s.). No regression; no speedup.
- **Recall:** a deterministic ordered-result-id equivalence check. The result tail on high-fanout queries has inherent run-to-run non-determinism from chromadb's approximate-NN (measured on the **base** branch: 6 runs, 0–6/30 divergence, mean 2.9; top-5 always stable). This branch is statistically indistinguishable from base → recall-neutral.

**Local suite status:** the converted modules and the new helper tests pass (`test_sqlite_ro.py` + `test_searcher.py` + `test_repair.py` + `test_backends.py` → 193 passed). The full `pytest tests/ --ignore=tests/benchmarks` run is **2266 passed**; the only failures are in `test_palace_locks.py` (inter-process file locking) and `test_init.py` (the leaked-`PYTHONPATH` filter test) — both **confirmed to fail identically on the base commit** in this dev environment (WSL2 file locking + the editable-via-`PYTHONPATH` dev setup), i.e. **not introduced by this change**. Zero failures in any module this PR touches. (`tests/benchmarks` requires extra deps and was not collected.)

## Checklist

- [x] I have included a thinking path that traces from project context to this change (see above)
- [x] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [ ] If this change affects the UI, I have included before/after screenshots (N/A — no UI)
- [x] I have considered and documented any risks above (latency-neutral, not a speedup; runaway statements now raise `OperationalError` instead of hanging — the intended behaviour; tuning PRAGMAs off by default to avoid perturbing FTS5 candidate ordering)
- [x] I will address all reviewer comments before requesting merge
- [x] No hardcoded paths (helper takes the db path from the caller; tests use `tmp_path`)
- [x] Linter passes (`ruff check .` → all checks passed)

## AI Disclosure

**AI tool used:** Claude Code (Anthropic Claude Opus).

**Prompt / approach:** Driven as a "measure → find the bottleneck → fix with before/after numbers, preserve behaviour/API" task. The agent profiled the warm read path, discovered it was **not** SQLite-bound (page-cache/Python-bound), so it framed the change as safety hardening rather than a speedup; built a paired alternating benchmark plus a deterministic recall-equivalence harness to prove latency- and recall-neutrality (rejecting an early single-run reading as noise); converted the read sites through one helper; and added an AST guard to lock in the existing no-injection posture. Two iterations were driven by data: a 50k-opcode progress-handler interval was raised to 1M after it measurably slowed the FTS5/union path, and the read-tuning PRAGMAs were disabled after they showed zero gain but perturbed result ordering.

### DIFF
--- a/mempalace/_sqlite_ro.py
+++ b/mempalace/_sqlite_ro.py
@@ -1,0 +1,210 @@
+"""Centralized read-only SQLite connection helper for MemPalace.
+
+SQLite has no ``SET statement_timeout`` (that is Postgres-only). The correct
+analogs are:
+
+- ``PRAGMA busy_timeout`` -- a *lock-wait* ceiling (how long to wait for a
+  writer's lock before giving up). Bounds contention with the miner/repair.
+- ``set_progress_handler`` + an abort return -- a *runaway-statement* ceiling
+  (wall-clock deadline for a single statement). Bounds a pathological
+  BM25/IN/metadata scan that would otherwise run unbounded against the
+  multi-GB ``chroma.sqlite3`` and block the MCP event loop.
+
+This helper centralizes both ceilings, guarantees the connection is closed on
+every exit path (success or exception), enforces ``query_only`` as defense in
+depth, and applies read-tuning PRAGMAs (``mmap_size`` / ``cache_size`` /
+``temp_store``) sized for the large store.
+
+Behaviour-preserving: callers run exactly the same SQL and get exactly the same
+rows in the same order. The only new behaviour is that a statement exceeding its
+wall-clock deadline now raises :class:`StatementTimeout` instead of running
+unbounded. Normal queries (well under the deadline) see no behavioural change.
+"""
+from __future__ import annotations
+
+import sqlite3
+import time
+from contextlib import contextmanager
+from typing import Iterator, Optional
+
+# --- Defaults -----------------------------------------------------------------
+
+# Lock-wait ceiling. Matches SQLite's historical 5 s default; explicit so the
+# value is visible and tunable rather than implicit.
+DEFAULT_BUSY_TIMEOUT_MS = 5_000
+
+# Runaway-statement wall-clock ceiling for interactive read surfaces
+# (search, status, vector-segment lookup). Batch/repair callers pass a
+# larger value explicitly.
+DEFAULT_DEADLINE_S = 15.0
+
+# Progress handler fires every N virtual-machine opcodes and calls
+# time.monotonic(). Lower = finer deadline granularity but more per-opcode
+# overhead -- and FTS5/union scans burn opcodes fast, so an aggressive value
+# measurably slowed the union path (p50 +8%, p95 +63%) in the first AFTER run.
+# For a 15-120 s wall-clock deadline, checking the clock every ~1M opcodes is
+# more than frequent enough; it cuts the callback rate ~20x and erases the
+# union regression while still aborting a true runaway well inside the ceiling.
+_PROGRESS_OPS = 1_000_000
+
+# Read-tuning PRAGMAs (mmap_size / cache_size / temp_store) are DISABLED by
+# default (0 = "don't set"). Two measured reasons, from the paired benchmark in
+# projects/active/mempalace-ab/bench/harden-sqlite/:
+#   1. No latency benefit. The warm read path is served from the OS page cache
+#      and is Python/chromadb-bound, not SQLite-I/O-bound; 5 alternating rounds
+#      showed the tuning delta inside the run-to-run noise band (+/-~40 ms p50).
+#   2. They perturb results. The upstream FTS5 candidate query has no
+#      deterministic ORDER BY before its LIMIT; mmap/cache/temp_store change the
+#      b-tree traversal order, which changes which candidates survive the LIMIT
+#      -> different (and run-to-run unstable) recall on high-fanout queries.
+#      A read-safety change must be recall-neutral, so we do not pay an ordering
+#      risk for a benefit that does not exist.
+# Callers that benchmark a STATIC snapshot may still opt in by passing explicit
+# mmap_size / cache_size_kib. The live read path uses 0/0.
+DEFAULT_MMAP_SIZE = 0  # bytes; 0 = leave SQLite default (no PRAGMA mmap_size)
+DEFAULT_CACHE_SIZE_KIB = 0  # 0 = leave SQLite default (no PRAGMA cache_size)
+DEFAULT_TEMP_STORE_MEMORY = False  # leave PRAGMA temp_store at its default
+
+
+class StatementTimeout(sqlite3.OperationalError):
+    """A read statement exceeded its wall-clock deadline and was aborted."""
+
+
+def _install_deadline(conn: sqlite3.Connection, deadline_s: Optional[float]) -> None:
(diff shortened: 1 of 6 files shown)


apollion69:
Addressed all three review findings (pushed `39ca136`):

**`_sqlite_ro.py:120` — fragile URI interpolation** ✅
`_build_uri` now uses `Path(db_path).absolute().as_uri()` instead of `f"file:{db_path}?mode=ro"`, so Windows drive letters / backslashes and POSIX paths containing `?` or `#` (valid filename chars) are percent-encoded correctly before `?mode=ro` is appended. Added `test_build_uri_handles_special_chars` (a `od?d#name.sqlite3` file opens and reads back).

**`_sqlite_ro.py:70` — `StatementTimeout` never raised** ✅
Good catch — it was dead. Rather than delete it (the typed timeout is the point of the PR), the deadline is now stored on the connection and a `_RODeadlineConnection` / `_RODeadlineCursor` pair translates a deadline-tripped `OperationalError` ("interrupted") into `StatementTimeout`. Both the `conn.execute(...)` shortcut and the `conn.cursor().execute(...)` surface are covered (callers use both — e.g. `backends/chroma.py`). Any `OperationalError` raised *before* the deadline (lock, read-only violation) passes through unchanged. Tests: `test_deadline_abort_raises_statement_timeout_specifically` (both surfaces) and `test_non_timeout_operationalerror_not_mistranslated`.

**`tests/test_sqlite_ro.py:175` — AST guard gap** ✅
The guard now follows one hop of indirection: it collects `name = <f-string|str-binop>` assignments and resolves `conn.execute(sql)` where `sql` is such a variable, so `sql = f"… {user} …"; conn.execute(sql)` no longer bypasses it. Added `test_guard_catches_variable_assigned_fstring_sql` proving the bypass is now flagged. (Went with AST-resolution over Ruff `S608` to keep the targeted allowlist semantics — `row_id_expr`/`filter_sql`/`placeholders` — that the test already encodes.)

Full file: 15 passed, `ruff check` + `ruff format --check` clean.

apollion69:
Gentle nudge on the pending review here — the read-path statement timeout + connection-lifecycle bounds have been running in our fork for a few weeks now with no regressions. Glad to rebase or split it up if that makes review easier.

igorls:

PR #1651, opened 2026-05-29
feat(mining): preserve conversation timestamps from transcript in drawer metadata

### LINKED ISSUE #1649: feat(mining): preserve conversation timestamps from transcript in drawer metadata
## Problem

The conversation miner (`convo_miner.py`) writes a `filed_at` field to every drawer's metadata — but this reflects the *ingest* time, not the *conversation* time. A transcript from three weeks ago mined today gets `filed_at = today`, losing the original temporal context entirely.

Every message entry in Claude Code's JSONL transcript format already carries a `timestamp` field:

```json
{"type": "user", "timestamp": "2026-04-25T22:19:04.027Z", "message": {...}}
```

The miner reads this field for chunking/normalization but never writes it into the stored drawer metadata.

## Impact

- Time-based search ("what did we discuss last week?") is impossible — all drawers look like they were created at ingest time
- The knowledge graph has no reliable temporal anchor for conversation-derived facts
- `mempalace_kg_timeline` and `as_of` queries are less useful because the temporal signal is lost

## Proposed Fix

When mining conversation transcripts, extract the `timestamp` of the first (and optionally last) message in each chunk and write it into the drawer metadata alongside `filed_at`:

```python
batch_metas.append({
    "wing": wing,
    "room": chunk_room,
    ...
    "filed_at": filed_at,           # ingest time (existing)
    "conversation_at": chunk_first_timestamp,  # actual conversation time (new)
})
```

This allows:
- Temporal search filtered by when the conversation actually happened
- KG triples to be anchored to real conversation dates
- Retrospective `as_of` queries that reflect reality

## Notes

- `filed_at` should be kept as-is for backwards compatibility
- `conversation_at` could be ISO 8601 string, same format as the transcript timestamps
- For chunked transcripts, using the timestamp of the first message in the chunk is the most meaningful anchor
- Already-mined drawers would need a re-mine to pick up the new field (acceptable)

tehw0lf:
## Summary

Fixes #1649.

The conversation miner wrote `filed_at` (ingest time) to every drawer's metadata — but this reflects *when mining ran*, not *when the conversation happened*. A transcript from three weeks ago mined today got `filed_at = today`, losing the original temporal context.

Every message entry in Claude Code's JSONL transcript format already carries a `timestamp` field — this PR extracts it and writes it as `conversation_at` alongside the existing `filed_at`.

## Changes

- New `_extract_conversation_timestamp(filepath)` helper reads the JSONL directly and returns the ISO 8601 timestamp of the first user/assistant message
- `conversation_at` is written into drawer metadata when available (optional, backwards-compatible — non-JSONL files and transcripts without timestamps simply omit the field)
- `filed_at` is kept unchanged for backwards compatibility

## Impact

- Time-based search ("what did we discuss last week?") becomes possible
- Knowledge graph facts can be anchored to real conversation dates
- `mempalace_kg_timeline` and `as_of` queries reflect reality instead of ingest time
- Already-mined drawers need a re-mine to pick up the new field (acceptable)

## Test plan

- [ ] All existing `test_convo_miner.py` tests pass (22/22)
- [ ] Verify `conversation_at` appears in drawer metadata after mining a Claude Code JSONL transcript
- [ ] Verify non-JSONL files mine without error (field simply absent)
- [ ] Verify `filed_at` is unchanged

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -14,7 +14,7 @@ import sys
 import hashlib
 import logging
 from pathlib import Path
-from datetime import datetime
+from datetime import datetime, timezone
 from collections import defaultdict
 from typing import Optional
 
@@ -427,7 +427,7 @@ def _extract_conversation_timestamp(filepath: str) -> Optional[str]:
                         try:
                             if ts > 1e11:
                                 ts /= 1000.0
-                            return datetime.fromtimestamp(ts).isoformat()
+                            return datetime.fromtimestamp(ts, tz=timezone.utc).isoformat()
                         except (ValueError, OSError, OverflowError):
                             return None
     except OSError:
@@ -435,7 +435,9 @@ def _extract_conversation_timestamp(filepath: str) -> Optional[str]:
     return None
 
 
-def _file_chunks_locked(collection, source_file, chunks, wing, room, agent, extract_mode, conversation_at=None):
+def _file_chunks_locked(
+    collection, source_file, chunks, wing, room, agent, extract_mode, conversation_at=None
+):
     """Lock the source file, purge stale drawers, and upsert fresh chunks.
 
     Combines the per-file serialization that prevents concurrent agents from
@@ -484,17 +486,17 @@ def _file_chunks_locked(collection, source_file, chunks, wing, room, agent, extr
                 batch_docs.append(chunk["content"])
                 batch_ids.append(drawer_id)
                 meta = {
-                        "wing": wing,
-                        "room": chunk_room,
-                        "hall": _detect_hall_cached(chunk["content"]),
-                        "source_file": source_file,
-                        "chunk_index": chunk["chunk_index"],
-                        "added_by": agent,
-                        "filed_at": filed_at,
-                        "ingest_mode": "convos",
-                        "extract_mode": extract_mode,
-                        "normalize_version": NORMALIZE_VERSION,
-                    }
+                    "wing": wing,
+                    "room": chunk_room,
+                    "hall": _detect_hall_cached(chunk["content"]),
+                    "source_file": source_file,
+                    "chunk_index": chunk["chunk_index"],
+                    "added_by": agent,
+                    "filed_at": filed_at,
+                    "ingest_mode": "convos",
+                    "extract_mode": extract_mode,
+                    "normalize_version": NORMALIZE_VERSION,
+                }
                 if conversation_at:
                     meta["conversation_at"] = conversation_at
                 batch_metas.append(meta)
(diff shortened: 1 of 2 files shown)


igorls:
Maintainer push: added one commit (`d7f5050`) to make this merge-ready:

1. **ruff format** `convo_miner.py` — this was the CI-blocking `ruff format --check` diff (mis-indented `meta = {...}` dict).
2. **UTC timestamps** — the epoch branch used `datetime.fromtimestamp(ts).isoformat()`, which emits *naive local-time* ISO with no offset. That contradicts the helper's own docstring promise of a "lexicographically comparable ISO 8601 value" (naive local times aren't comparable across machines/DST). Changed to `datetime.fromtimestamp(ts, tz=timezone.utc)` so the value is offset-aware UTC. The ISO-string branch still passes through the source value verbatim.
3. **Tests** — added 6 focused tests for `_extract_conversation_timestamp` (ISO passthrough, epoch seconds→UTC, epoch-millis scaling, pretty-printed-JSON abort, no-timestamp, non-JSONL skip).

Full `test_convo_miner.py` green locally (28 passed), ruff format + check clean. The feature itself is a nice addition — thanks! One thing to double-check on your side: the ISO-string branch returns whatever the source recorded (which may be naive); if you want fully uniform UTC you could normalize that branch too, but I left it as-is to avoid changing well-formed source timestamps.

tehw0lf:
Thanks for the thorough review and the maintainer push, @igorls — UTC fix and the new tests are a great improvement!

One follow-up question on the timestamp format: you noted that the ISO-string branch passes through the source value verbatim, which could be naive local time in some edge cases. This got us thinking — would it make more sense to store `conversation_at` as a Unix timestamp (integer) instead of, or in addition to, the ISO string?

Arguments for Unix timestamp:
- Always unambiguous UTC by definition — no timezone/format inconsistency possible
- Numerically comparable and sortable in ChromaDB metadata filters without string parsing
- Compact

We could do either:
1. **Unix timestamp only** (`conversation_at` as `int`) — clean, unambiguous, but less human-readable
2. **Both** — `conversation_at` as ISO string (verbatim, human-readable) + `conversation_at_epoch` as `int` (always comparable)

Happy to implement whichever direction makes sense. What do you think?

tehw0lf:
Following up on the timestamp format discussion: we've now implemented **both** fields in `10b84a3`:

- `conversation_at` — ISO 8601 string (human-readable, directly usable without parsing)
- `conversation_at_epoch` — `int` Unix timestamp (numerically filterable/sortable in ChromaDB metadata, always UTC by definition)

The reasoning: storing only epoch would force every client/query to convert before displaying; storing only ISO would require string parsing for range filters. Both together give callers what they need without any trade-off.

The ISO-string branch now also derives `conversation_at_epoch` via `datetime.fromisoformat()` with UTC normalisation for naive timestamps, so both fields are always consistent with each other.

Tests updated to assert the `(iso, epoch)` tuple return type and verify epoch matches ISO for all three input forms (ISO passthrough, epoch-seconds, epoch-millis).

tehw0lf:
Superseded by #1890, which landed `authored_at` on conversation drawers and wires it through search, ranking and a backfill script. That covers the core goal of this PR — preserving the conversation's own timestamp instead of collapsing everything to ingest time. This branch has also drifted well behind `develop` in the meantime. Closing as no longer needed. Thanks for the reviews and the maintainer push earlier!

PR #1654, opened 2026-05-29
fix(repair): honor --dry-run in repair --mode from-sqlite

potterdigital:
## What does this PR do?

`repair --mode from-sqlite` silently ignored `--dry-run`. `cmd_repair`'s from-sqlite branch never forwarded `args.dry_run` to `rebuild_from_sqlite` (the `max-seq-id` branch does), and `rebuild_from_sqlite` had no `dry_run` parameter. So `mempalace repair --mode from-sqlite --dry-run` did one of two non-preview things:

- with `--archive-existing` (or an existing dest) it ran the **real** rebuild — archiving the live palace and re-embedding every row;
- otherwise it aborted at the destructive-action confirmation prompt.

Neither previews. This matters most for the case from-sqlite exists to handle: a large corrupt palace where a real rebuild re-embeds 100k+ rows and runs for a long time — exactly when you want to look first.

This PR:

- adds a `dry_run` parameter to `rebuild_from_sqlite`; after the normal source validation it prints per-collection row counts and returns them **without** archiving the existing palace, creating collections, or re-embedding. Counts come from `sqlite_drawer_count` — the same SQLite ground-truth helper repair already uses — so the preview matches what a real rebuild upserts.
- wires `dry_run` through `cmd_repair` and skips `confirm_destructive_action` under `--dry-run` (a read-only preview shouldn't require `--yes` or block on a prompt).
- preserves the return contract: a dry run returns a populated `{collection: count}` dict (CLI treats as success); validation refusals still return `{}` exactly as before.

No new dependencies. No behavior change for runs without `--dry-run`.

## How to test

```
mempalace repair --mode from-sqlite --archive-existing --dry-run
# → "DRY RUN — no changes will be made", per-collection would-upsert counts, palace untouched
```

```
python -m pytest tests/test_repair.py tests/test_cli.py -k "dry_run or from_sqlite" -v
```

New tests:
- `test_rebuild_from_sqlite_dry_run_cross_palace_writes_nothing` — dry run creates nothing at dest and reports the counts the existing round-trip test proves a real rebuild produces.
- `test_rebuild_from_sqlite_dry_run_in_place_does_not_archive` — in-place dry run does not move the live palace aside.
- `test_cmd_repair_from_sqlite_dry_run_passes_through_and_skips_confirm` — CLI forwards `dry_run=True` and never calls `confirm_destructive_action`.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — 2270 passed, 3 skipped
- [x] No hardcoded paths

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -828,9 +828,18 @@ def cmd_repair(args):
         # No prompt when source != dest AND dest does not exist (pure
         # extract-into-fresh-dir case is non-destructive to existing
         # palaces).
+        dry_run = getattr(args, "dry_run", False)
+        # A dry run only reads the source SQLite and prints a plan — it
+        # never archives, creates, or writes — so it must not trip the
+        # destructive-action confirmation, which would otherwise block the
+        # preview behind a y/N prompt (or require --yes just to look).
         is_destructive_to_dest = source_path == palace_path or os.path.exists(palace_path)
-        if is_destructive_to_dest and not confirm_destructive_action(
-            "Rebuild from SQLite", palace_path, assume_yes=getattr(args, "yes", False)
+        if (
+            not dry_run
+            and is_destructive_to_dest
+            and not confirm_destructive_action(
+                "Rebuild from SQLite", palace_path, assume_yes=getattr(args, "yes", False)
+            )
         ):
             return
 
@@ -839,6 +848,7 @@ def cmd_repair(args):
                 source_palace=source_path,
                 dest_palace=palace_path,
                 archive_existing_dest=archive_existing,
+                dry_run=dry_run,
             )
         except RebuildPartialError as exc:
             # The error itself was already printed by rebuild_from_sqlite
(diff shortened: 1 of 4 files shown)


fatkobra:
This issue is still present on current `develop` and has now been independently reported again in #2095 and #2133.

The existing implementation here appears to cover the correct contract:

- forward `dry_run` from `cmd_repair`
- do not require destructive confirmation for a read-only preview
- validate the source before reporting a plan
- report per-collection rows without archiving, creating the destination, or re-embedding

The earlier review concern also looks valid: `sqlite_drawer_count(...) or 0` would conflate an unreadable/corrupt collection with a genuinely empty collection.

That path should instead fail explicitly, for example:

```python
count = sqlite_drawer_count(source_palace, collection_name)
if count is None:
    print(
        f"  ERROR: [{collection_name}] row count is unreadable; "
        "dry-run cannot produce a trustworthy rebuild plan."
    )
    return {}

igorls:
Superseded by #2138, which re-applies this dry-run approach on current develop (after single-writer mine-lock wrapping) and fails closed when `sqlite_drawer_count` is unreadable rather than using `or 0`.

igorls:
Superseded by #2138 (merged).

PR #1658, opened 2026-05-30
fix(backend): respect backend.persist_directory from palace mempalace.yaml

w1tc4:
## Problem

`backend.persist_directory` in a palace's `mempalace.yaml` was silently ignored. ChromaDB was **always** initialized at `palace_path` root because every call site hardcoded `path=palace_path` and no code ever read the palace-level YAML's backend section.

Users who configured:

```yaml
backend:
  type: chroma
  persist_directory: ".System_Data"
```

saw `chroma.sqlite3` and the HNSW segment directories keep reappearing in the palace root on every restart — making OneDrive, Dropbox, and other sync tools surface internal DB files alongside user content, and making it impossible to keep the palace directory organized.

## Root cause

`ChromaBackend._client()`, `make_client()`, `_db_stat()`, `_prepare_palace_for_open()`, and `detect()` all received `palace_path` and passed it straight to `chromadb.PersistentClient(path=palace_path)`. There was no code path that read `<palace_path>/mempalace.yaml` at all.

## Fix

Add `ChromaBackend._resolve_persist_dir(palace_path)` which:
1. Reads `<palace_path>/mempalace.yaml` for `backend.persist_directory`
2. Resolves relative paths against `palace_path`
3. Creates the directory if it does not exist
4. Falls back to `palace_path` for unconfigured palaces (fully backwards-compatible)

Wire it into every affected call site:
- `_db_stat` — inode/mtime freshness check
- `_client` — db_path existence check + `PersistentClient` init
- `_prepare_palace_for_open` — HNSW quarantine + blob seq_id fix
- `make_client` — legacy entry point
- `detect` — backend auto-detection
- `mcp_server._get_client` — stale-cache db_path probe (both occurrences)

## Testing

```python
# With backend.persist_directory: ".System_Data" in palace yaml:
ChromaBackend._resolve_persist_dir("D:/OneDrive/palace")
# → "D:\OneDrive\palace\.System_Data"

# Without backend section (existing palaces):
ChromaBackend._resolve_persist_dir("/path/to/palace")
# → "/path/to/palace"  (unchanged)
```

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -2,6 +2,7 @@
 
 import contextlib
 import datetime as _dt
+import functools
 import json
 import logging
 import math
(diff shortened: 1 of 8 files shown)


igorls:
The `room_detector_local` config-preserve change is the solid part; the `persist_directory` wiring needs work before merge:

1. **Blocking — feature is half-wired, breaks search.** `_resolve_persist_dir` is applied to the writer and a few probes, but core readers still hardcode the palace-root `chroma.sqlite3`: `searcher.py:308` (raises `SearchError: No palace database`) and `:424` (BM25 fallback), `palace.py:124`, `cli.py:625/863`, `migrate.py:141/219`, `mcp_server.py:742/2727`, and ~10 sites in `repair.py`. Reproduced: set `backend.persist_directory`, mine → DB lands in the subdir, then `search` reports "no palace". Either route **all** `chroma.sqlite3` / `PersistentClient` construction through `_resolve_persist_dir`, or scope the feature down — as-is it's worse than the bug it fixes for anyone who sets the key.

2. **Blocking — rebase onto develop.** Base predates #1617; `git merge-tree` shows a real conflict in `_prepare_palace_for_open`. After resolving, `_fix_missing_collection_type` must also receive `persist_dir`, or it silently no-ops when the key is set.

3. **Tests.** No tests for `_resolve_persist_dir` or any rewired site. A round-trip test (set `persist_directory` → mine → search/status find the DB) would have caught #1.

4. **CI / perf nits.** `ruff format --check` fails on `mcp_server.py:454` & `:676` (101 chars). `_resolve_persist_dir` also parses YAML on the per-op hot path with no memoization and calls `os.makedirs` from read-only paths (`detect()`); consider caching the resolved dir and moving directory-creation out of read paths.

w1tc4:
- _resolve_persist_dir wired to all ~20 hardcoded sites across chroma.py, searcher.py, migrate.py, repair.py, cli.py,
  mcp_server.py
  - @functools.lru_cache — no per-call YAML parsing
  - makedirs only runs on write paths (create=True)
  - 6 tests, all passing
  - ruff clean
  - No hooks.json change

igorls:

Issue #1663, opened 2026-05-30
Feature request: support configurable embedding model (e.g. embeddinggemma-300m for multilingual)


HZaxx:
## Problem

`mempalace/embedding.py` hardcodes `ONNXMiniLM_L6_V2` (all-MiniLM-L6-v2), an English-only model. For non-English users (Chinese, Japanese, Arabic, etc.), semantic search quality is significantly degraded because the model cannot understand non-English text.

The README already advertises `embeddinggemma-300m` (multilingual, 100+ languages) as an onboarding option, but the current codebase has no mechanism to actually use it.

## Current behavior

```python
# mempalace/embedding.py — hardcoded
def _build_ef_class():
    from chromadb.utils.embedding_functions import ONNXMiniLM_L6_V2
    class _MempalaceONNX(ONNXMiniLM_L6_V2):
        ...
```

## Proposed solution

Allow model selection via config or env var, e.g.:

```yaml
# ~/.mempalace/config.json
{
  "embedding_model": "multilingual"  # or "english" (default)
}
```

Where `"multilingual"` maps to `embeddinggemma-300m` (or a configurable HuggingFace model ID) and `"english"` keeps the current `all-MiniLM-L6-v2`.

The model switch should:
1. Be backward-compatible (default stays all-MiniLM-L6-v2)
2. Trigger a `mempalace migrate` or `mempalace repair` to re-index existing palaces (different model = different vector dims)
3. Auto-download the model on first use

## Context

- README references: `embeddinggemma-300m` (multilingual) and `all-MiniLM-L6-v2` (English-only)
- `mempalace/onboarding.py` mentions model choice in README but not in code
- `mempalace/migrate` CLI command already exists but doesn't handle model migration yet
- ChromaDB `SentenceTransformerEmbeddingFunction` can load arbitrary HuggingFace models, which would be a natural implementation path

## Impact

Non-English users currently get poor semantic search results. For a tool advertised as "the best-benchmarked open-source AI memory system", multilingual support is table stakes.

mvalentsev:
[...]

PR #1670, opened 2026-05-31
fix(cli): add repair rebuild-index alias

jsiu93:
## What does this PR do?

Adds `mempalace repair rebuild-index` as a CLI alias for rebuilding a palace index after changing `MEMPALACE_EMBEDDING_MODEL`.

The alias routes to the existing SQLite rebuild path with archive enabled, so the repair command avoids opening the old Chroma collection with the new embedding function. The embedding mismatch hint now points to the working command and quotes the palace path safely.

Related to #1663.

## How to test

- `.venv/bin/python -m pytest tests/test_cli.py -q`
- `.venv/bin/python -m pytest tests/test_backends.py::test_get_collection_translates_ef_mismatch_to_helpful_error -q`
- `.venv/bin/ruff check mempalace/cli.py mempalace/backends/chroma.py tests/test_cli.py tests/test_backends.py`

## Checklist

- [x] Targeted tests pass
- [x] No hardcoded paths
- [x] Linter passes for changed files

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -8,6 +8,7 @@ import math
 import os
 import pickle
 import re
+import shlex
 import sqlite3
 from collections import defaultdict
 from numbers import Integral
@@ -1800,6 +1801,7 @@ class ChromaBackend(BaseBackend):
             current_model = MempalaceConfig().embedding_model
         except Exception:
             current_model = "unknown"
+        rebuild_cmd = f"mempalace --palace {shlex.quote(palace_path)} repair rebuild-index"
         return (
             f"Embedding model mismatch reading palace at {palace_path!r}.\n"
             f"  Underlying ChromaDB error: {msg}\n"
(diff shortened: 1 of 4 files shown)


mvalentsev:
This command is already documented as if it exists, but it was never added to the CLI. This PR is what introduces it.

The references describe `mempalace repair rebuild-index` as a working recovery step:
- the EF-mismatch hint in `ChromaBackend._explain_ef_mismatch` (`mempalace/backends/chroma.py`)
- the CHANGELOG entry for the embedding-model switch
- the `config.py` and `embedding.py` docstrings

But the subcommand doesn't exist on develop, so following any of them lands on exit 2:

```
$ mempalace --palace /path/to/palace repair rebuild-index
error: unrecognized arguments: rebuild-index
```

So a user who switches `MEMPALACE_EMBEDDING_MODEL`, hits the mismatch, and does exactly what the hint tells them to do currently hits a dead command. This PR closes the gap between the docs and the CLI.

(Routing looks right too: mapping to `--mode from-sqlite --archive-existing` reads from `chroma.sqlite3` directly instead of `get_collection()`, which keeps recovery from re-triggering the same EF-mismatch.)

Issue #1563, opened 2026-05-20
Feature Request: Configuration-driven embedding model switching (no source code edits)


wangtengda0310:
## Problem Statement

Currently, switching embedding models in mempalace requires editing `embedding.py` source code. This is the only way to use models like:

- `BAAI/bge-base-zh-v1.5` (Chinese-optimized, 768d)
- `Qwen3-Embedding` via LM Studio/Ollama (1024d)
- Any other `sentence-transformers` compatible model

This creates significant friction:

1. **Every plugin update requires re-applying patches**
2. **Non-technical users cannot switch models** — editing Python source is a barrier
3. **No discoverability** — Users don't know what models are available or how to enable them
4. **Inconsistent with modern tools** — Most AI frameworks (LangChain, LlamaIndex, etc.) support model switching via configuration

## Current Pain

```python
# Current: Must edit embedding.py
def get_embedding_function(device=None):
    # Hard-coded to ONNXMiniLM_L6_V2
    ef_cls = _build_ef_class()
    ef = ef_cls(preferred_providers=providers)
    return ef
```

```python
# User's workaround: Patch the module
class _LMStudioEmbeddingFunction:
    def __call__(self, input): ...
    def embed_query(self, text): ...
# Then monkey-patch get_embedding_function
```

### 1. Extend `config.json` with embedding configuration

```json
{
  "embedding": {
    "provider": "sentence-transformers",
    "model": "BAAI/bge-base-zh-v1.5",
    "device": "auto"
  }
}
```

Or for API-based models:
```json
{
  "embedding": {
    "provider": "openai-compatible",
    "api_url": "http://localhost:1234/v1/embeddings",
    "model": "qwen3-embedding-0.6b",
    "api_key": ""
  }
}
```

### 2. Built-in provider registry

Support common providers out of the box:

| Provider | Description | Config Example |
|----------|-------------|----------------|
| `onnx` | Default MiniLM (current behavior) | `{"provider": "onnx"}` |
| `sentence-transformers` | HuggingFace/SBERT models | `{"provider": "sentence-transformers", "model": "BAAI/bge-base-zh-v1.5"}` |
| `openai-compatible` | LM Studio, Ollama, etc. | `{"provider": "openai-compatible", "api_url": "...", "model": "..."}` |
| `ollama` | Native Ollama embed | `{"provider": "ollama", "model": "qwen3-embedding"}` |

### 3. Environment variable override

```bash
# Quick switch without editing config
[...]

Issue #1561, opened 2026-05-20
Feature Request: Persist and verify embedding model metadata in palace collections


wangtengda0310:
## Problem Statement

Currently, mempalace does not record which embedding model was used to create a palace. This causes several practical problems:

1. **No way to identify the model from palace data** — Users cannot determine if a palace was created with MiniLM (384d), BAAI (768d), Qwen (1024d), or another model without manually tracking it externally.

2. **Silent failures on model switch** — If a user switches from `embedding_device: "cpu"` to `embedding_device: "api"` (e.g., LM Studio with Qwen), searches return wrong results or errors without clear indication of why.

3. **Backup confusion** — Users with multiple palace backups (e.g., `palace.backup`, `palace-backup-20260520`) cannot tell which backup uses which model without trial and error.

4. **Multi-device/team workflows break** — A palace shared across devices or team members requires everyone to use the exact same embedding model, but there's no mechanism to enforce or verify this.

### Scenario 1: Multi-device workflow
A user works on a Windows PC with LM Studio (Qwen3-Embedding, 1024d) and wants to continue on a MacBook. They need to know the exact model to configure the same setup on the second device.

### Scenario 2: Model upgrade
A user starts with the default ONNXMiniLM_L6_V2 (384d) but later wants better Chinese semantic search. They switch to Qwen3-Embedding (1024d) but forget to rebuild the palace. All searches silently return poor results.

### Scenario 3: Backup archaeology
After several months, a user has multiple palace backups and wants to restore the "best" one. They cannot determine which backup has the most relevant data because they don't know which model each backup used.

### Scenario 4: CI/CD automation
An automated pipeline runs `mempalace mine` and `mempalace search`. If the pipeline environment changes (e.g., from CPU to GPU or from local to API), the pipeline fails silently without clear error messages.

## Current Workaround

Users must manually maintain a `model_info.json` or rely on directory naming conventions (e.g., `palace-qwen-1024/`), which is error-prone and not integrated with mempalace.

### 1. Store embedding metadata in collection metadata

When creating a collection, persist the embedding model information:

```json
{
  "mempalace:embedding_model": "qwen3-embedding-0.6b",
  "mempalace:embedding_dimension": 1024,
  "mempalace:embedding_provider": "lmstudio",
  "mempalace:embedding_device": "api",
  "mempalace:version": "3.3.5"
}
```

For the default ONNX model:
```json
{
  "mempalace:embedding_model": "onnx-minilm-l6-v2",
[...]

Issue #1261, opened 2026-04-28
Feature request: configurable embedding_function (currently hardcoded to all-MiniLM-L6-v2)


DamiCavag:
## Problem

`backends/chroma.py` (around line 536) calls `client.get_or_create_collection(name, metadata={"hnsw:space": hnsw_space})` without passing an `embedding_function`. ChromaDB falls back to its default `all-MiniLM-L6-v2`, which is English-trained.

This blocks adoption for users with non-English content (Spanish, Portuguese, etc.) where multilingual embedders like `paraphrase-multilingual-MiniLM-L12-v2` would significantly improve recall.

## Empirical evidence

Tested v3.3.3 on a Spanish-majority vault sample (~70% ES + 30% EN technical terms), 132 KB markdown across 3 files, 229 drawers indexed:

| Query | Match score | Top-1 correct |
|---|---|---|
| "smoke test exit code grep gated" (mostly EN) | 0.626 | yes |
| "FOCUSED audit pre-pilot" (mixed) | 0.577 | yes |
| "L-22 skill manifest delegation" (mixed) | 0.465 | yes |
| "hub and spoke agents never talk" (mixed) | 0.314 | yes |
| "process bloat watch meta-gate" (mostly ES) | 0.337 | no — concept exists in indexed file but ranked outside top-3 |

Top-1 accuracy: **80%** (4/5) with the English-only embedder. SBERT/MTEB benchmarks suggest a multilingual embedder would lift this to ~90-100% on the same content.

## Suggested API

Option 1 — env var (lowest friction):

```
MEMPALACE_EMBEDDING_MODEL=paraphrase-multilingual-MiniLM-L12-v2
```

Option 2 — field in `~/.mempalace/config.json`:

```json
{ "embedding_model": "paraphrase-multilingual-MiniLM-L12-v2" }
```

Option 3 — CLI flag on `mempalace init`:

```
mempalace init <dir> --embedder multilingual
```

Internally this would translate to passing `embedding_function=SentenceTransformerEmbeddingFunction(model_name=...)` to `get_or_create_collection`.

## Important constraint

The collection's embedder is fixed at creation time (your `EmbedderIdentityMismatchError` enforces this). So the config needs to be honored on `init` / `create_collection` — switching mid-life would require re-indexing.

## Workaround attempted

Local source patch in venv works but isn't sustainable: any `pip install --upgrade mempalace` overwrites it. A fork is heavyweight overhead for a 1-line change.

## Why this matters

[...]

PR #1671, opened 2026-06-01
feat(embedding): add OpenAI-compatible /v1/embeddings backend

### LINKED ISSUE #1559: Feature Request: Support external embedding APIs (LM Studio, Ollama, OpenAI-compatible)
## Summary

Currently, `get_embedding_function()` only supports ONNX Runtime providers (`cpu`/`cuda`/`coreml`/`dml`). It would be very useful to support external embedding APIs like LM Studio, Ollama, or any OpenAI-compatible endpoint.

## Motivation

The default `ONNXMiniLM_L6_V2` model works well for English but has limited performance for Chinese text. Many users (including myself) run local embedding models via LM Studio or Ollama that provide much better multilingual understanding.

For example, Qwen3-Embedding models via LM Studio produce 1024-dim vectors with significantly better Chinese semantic search results compared to the default 384-dim MiniLM.

## Current Workaround

I have to manually patch `embedding.py` to add an `_LMStudioEmbeddingFunction` class:

```python
class _LMStudioEmbeddingFunction:
    def __init__(self):
        self.api_url = os.environ.get("MEMPALACE_EF_API_URL", "http://localhost:1234/v1/embeddings")
        self.model = os.environ.get("MEMPALACE_EF_MODEL", "qwen3-embedding-0.6b")
    
    def __call__(self, input):
        # calls external API
        ...
    
    def embed_query(self, text=None, input=None):
        ...
    
    def embed_documents(self, texts=None, input=None):
        ...
```

This works but breaks on every plugin update.

## Proposed Solution

Add a new `embedding_device` option: `"api"` (or `"external"`), with configuration via:

1. **Environment variables** (preferred for flexibility):
   - `MEMPALACE_EF_API_URL` — API endpoint
   - `MEMPALACE_EF_MODEL` — Model name
   - `MEMPALACE_EF_API_KEY` — Optional API key

2. **Config file** (`~/.mempalace/config.json`):
   ```json
   {
     "embedding_device": "api",
     "embedding_api_url": "http://localhost:1234/v1/embeddings",
     "embedding_api_model": "qwen3-embedding-0.6b"
   }
   ```

## Use Cases

- **LM Studio** — Local GPU-accelerated embedding models
- **Ollama** — `ollama embed` command or API
- **OpenAI API** — `text-embedding-3-small` etc.
- **Any OpenAI-compatible endpoint**

## Additional Context

Tested with:
- MemPalace 3.3.5
- LM Studio 0.3.x with Qwen3-Embedding-0.6B
- Chinese text similarity: 0.82-0.94 cosine similarity for related queries

Would love to see official support for this! Happy to help test or contribute a PR if the approach looks good.

---

**Related**: This would also solve the recurring issue where users want to use different embedding models without invalidating their palace (since the collection dimension would be managed by the external API).

maximilize:
## What does this PR do?

Adds an `embedding_model: "openai-compat"` option that computes embeddings via any OpenAI-compatible `/v1/embeddings` endpoint (LM Studio, llama.cpp, vLLM, Ollama's OpenAI shim, or a self-hosted server) instead of a local ONNX model — useful for larger / multilingual embedders (e.g. Qwen3-Embedding) or GPU offload.

- New `OpenAICompatEmbeddingFunction` (stdlib `urllib` only — **no new dependency**): batches requests, sends `encoding_format: "float"` + a `mempalace/<ver>` User-Agent (avoids the Cloudflare 403 from #1570), validates the response (contiguous `0..n-1` indices + well-formed vectors), and L2-normalizes for the cosine collection. Implements `embed_query` (ChromaDB 1.5 dispatches query embedding through it). `name()` encodes the model id so switching it surfaces as a clear rebuild-index signal. Failures raise a module-specific `EmbeddingAPIError`.
- Endpoint settings are resolved by `MempalaceConfig` as a single source of truth: `embedding_api_url` / `embedding_api_model` / `embedding_api_key`, each overridable via the matching `MEMPALACE_EMBEDDING_API_*` env var.
- **Opt-in, local-first preserved**: the default stays `minilm`; nothing changes unless you select `openai-compat`. When the endpoint is on your machine/LAN no content leaves your network — consistent with the zero-API-by-default principle (the endpoint is the user's explicit choice and may be fully local).
- Mirrors the existing `openai-compat` LLM provider naming. The selector reuses `embedding_model` (not `embedding_device`) because the device axis is for hardware accelerators.

Generalizes the Ollama-only approach in #982 (currently conflicting) to any OpenAI-compatible endpoint — Ollama works via its `/v1` shim. Does **not** touch reranking (#1032).

## How to test

Point at any OpenAI-compatible server via `~/.mempalace/config.json`:

```json
{
  "embedding_model": "openai-compat",
  "embedding_api_url": "http://localhost:1234",
  "embedding_api_model": "<server-side model id>"
}
```

(or the equivalent `MEMPALACE_EMBEDDING_API_*` env vars), then:

```bash
mempalace init <dir> && mempalace mine <dir> && mempalace search "<query>"
```

Unit tests need no server and no network (`urlopen` is mocked):

```bash
uv run pytest tests/test_embedding_api.py -v
```

## Checklist
- [x] Tests pass (`uv run pytest tests/ -v` — 2304 passed, 3 skipped; no network / API keys needed)
- [x] No hardcoded paths

## Related issues
Closes #1559. Related: #756, #1261, #1563, #903, #1561. Generalizes the Ollama-only #982.


maximilize:
Thanks @gemini-code-assist — all three addressed in c3e6692:

- `except` now also wraps `http.client.HTTPException` (BadStatusLine / IncompleteRead) and `ValueError` (the latter subsumes `json.JSONDecodeError` and covers an invalid/missing URL scheme).
- `_vectors_from_response` rejects a non-dict top-level response before `.get()`, so a JSON list / `null` / string raises a clear `EmbeddingAPIError` instead of an unhandled `AttributeError`.
- Added tests for all three. The suggested transport-error and server-error-body tests already existed, so I didn't duplicate them.

---

**For maintainers — design summary (re #1559):**

- **Opt-in.** Default stays `minilm`; nothing changes unless `embedding_model: "openai-compat"` is selected. Zero regression risk.
- **Single source of truth.** Endpoint/model/key resolved by `MempalaceConfig`: `embedding_api_url` / `embedding_api_model` / `embedding_api_key`, each overridable via the matching `MEMPALACE_EMBEDDING_API_*` env var.
- **No new dependency** — stdlib `urllib` only. Stays local (no external-API warning) when the endpoint is on the machine/LAN, consistent with the zero-API-by-default principle.
- Selector reuses `embedding_model` (not `embedding_device`, which is for hardware accelerators) and mirrors the existing `openai-compat` LLM provider naming.
- Generalizes the Ollama-only #982 (currently conflicting) to any OpenAI-compatible `/v1/embeddings` server — Ollama works via its `/v1` shim. Does **not** touch reranking (#1032). Also relates to #756, #1261, #1563, #903, #1561.

Full suite is green and the unit tests need no server/network. Happy to adjust the config surface or naming before merge — wanted to give a concrete, tested proposal rather than just a design sketch.

jhsmith409:
+1 — this is exactly what a shared / multi-machine setup needs.

When several hosts share one vector store (e.g. a qdrant or pgvector collection), every host has to embed with the **identical** model or the vectors are mutually incompatible. Today that means each node runs its own copy of the same local ONNX model. An OpenAI-compatible `/v1/embeddings` option lets the whole fleet point at **one** GPU-hosted embedder (Qwen3-Embedding, etc.), which both guarantees consistency across nodes and offloads embedding from CPU-only hosts.

It would pair naturally with the shared-backend / network-backend efforts (#1963 and the qdrant/pgvector work). Would love to see this land.

cuttie1979:
Hey maintainers — we'd really love to see this land. :wave:

**Use case:** We run several hosts against one shared vector store (Qdrant), and every node has to embed with the *identical* model or the vectors are mutually incompatible. Today that means running a local ONNX model copy on every machine — wasteful on CPU-only hosts and hard to keep consistent. An OpenAI-compatible `/v1/embeddings` option lets the whole fleet point at **one GPU-hosted embedder** (Qwen3-Embedding), guaranteeing vector consistency across nodes while offloading embedding entirely.

**Status:** code review done (all bot findings addressed in `c3e6692`), branch `mergeable`, tests pass, strictly opt-in (`embedding_model: "openai-compat"`) with zero regression risk to the default path. Open since June 1 with no maintainer feedback yet.

Would pair naturally with the shared-backend work (#1963, Qdrant/pgvector backends). @milla-jovovich @bensig @igorls — any chance of a review? Happy to rebase or adjust anything if needed.

igorls:
## Maintainer rebase onto current `develop`

Rebased this branch onto latest `origin/develop` (was ~238 commits behind) and force-pushed to `feat/openai-compat-embeddings` with `--force-with-lease`.

### Rebase result
- **Clean rebase** — the original 2 commits applied with **no conflicts**.
- **+1 post-rebase commit** so the suite is green on current develop:

  | Commit | What |
  |---|---|
  | `d471a9e` | feat: OpenAI-compatible `/v1/embeddings` backend (original) |
  | `f272c84` | fix: harden API error handling (original review follow-up) |
  | `d9a24c7` | post-rebase: opt `test_embedding_api` out of conftest’s stable EF mock + move CHANGELOG entry to Unreleased |

### Why the extra commit was needed
Current `develop` has an autouse fixture in `tests/conftest.py` that stubs `get_embedding_function` for most modules (avoids loading ONNX on Windows). Without opting `test_embedding_api` into `_REAL_EMBEDDING_TEST_MODULES`, the selection-branch tests received `_StableTestEmbeddingFunction` and failed (3 failures). Same pattern as `test_embedding` / `test_embeddinggemma`.

CHANGELOG context had drifted: the feature bullet rebased under released **3.7.0 → Performance**. Moved it to **Unreleased → Features**.

### Verification (local)
```text
uv run pytest tests/test_embedding_api.py tests/test_embedding.py tests/test_embeddinggemma.py -v
# 75 passed
uv run ruff check mempalace/embedding.py mempalace/config.py tests/test_embedding_api.py tests/conftest.py
# All checks passed
```

### Still open (non-blocking for CI, useful before merge)
1. **EF cache lock** — ONNX path uses `_EF_CACHE_LOCK`; openai-compat still does lock-free get/set (double-create race under concurrent first init).
2. **Privacy warning** — mirror `llm_client._endpoint_is_local` when the embedding URL is not loopback/LAN.
3. **CI** — waiting on checks after this push.

@maximilize — sorry for the force-push on your branch; `maintainerCanModify` was on. Happy to adjust anything if you’d rather take the rebase yourself next time.

Closes the “open since June / needs rebase” blocker called out on #1559.

PR #1678, opened 2026-06-02
fix(mcp): defer WAL setup so import no longer recreates ~/.mempalace (#1676)

### LINKED ISSUE #1676: Bug: MCP server module-level mkdir defeats _palace_root_exists() kill switch
## Summary

The kill switch added in #1305 (`hooks_cli._palace_root_exists()`) is silently
defeated by a module-level side effect in `mcp_server.py`. Importing the MCP
server creates `~/.mempalace/wal/` unconditionally — including when the user
has explicitly removed `~/.mempalace/` as the documented kill-switch gesture.

After this, `_palace_root_exists()` returns True again (because the dir
re-exists, even though `config.json` and the rest of the palace state are
absent), and subsequent hook fires resume writing: autosave, mining, state
files.

## Reproducer

```bash
# 1. Confirm kill switch is engaged
rm -rf ~/.mempalace
python -c "from mempalace.hooks_cli import _palace_root_exists; print(_palace_root_exists())"
# False  ✅ kill switch active

# 2. Touch the MCP server (any MCP tool call does this via the running server)
python -c "import mempalace.mcp_server"

# 3. Kill switch is now defeated
ls -la ~/.mempalace/
# drwx------ ... wal/    ← recreated at import time
python -c "from mempalace.hooks_cli import _palace_root_exists; print(_palace_root_exists())"
# True  ❌ kill switch silently disengaged
```

For users who run the MCP server alongside hooks, step 2 happens on every
session-start. The kill switch never holds across more than a single session.

## Root cause

`mempalace/mcp_server.py` (lines 390-391 of `develop` at the time of writing):

```python
_WAL_DIR = Path(os.path.expanduser("~/.mempalace/wal"))
_WAL_DIR.mkdir(parents=True, exist_ok=True)
```

This runs at module import. `parents=True` creates `~/.mempalace/` as a side
effect. There's no check against `_palace_root_exists()` — the WAL setup
predates the kill switch (#1305 was 2026-05-06) and was never updated to
respect the same contract.

The hooks-side fix from #1305 audited every disk touch in `hooks_cli.py`
behind `_palace_root_exists()`. The MCP server side has the same disk touches
(WAL dir, WAL file, palace state on first tool call) and none of them check.

## Evidence from production

Two independent reproductions on Linux bots:

**Bot A (2026-06-01):**
- 19:39:00 UTC — `mv ~/.mempalace ~/.mempalace.disabled-...` (kill switch engaged)
- 19:56:13 UTC — diary write through running MCP server
- 19:56:13 UTC — `~/.mempalace/wal/` recreated (matched the diary write to the second)
- 20:00 UTC – next day 15:00 UTC — 1,184 then +231 chunks ingested by the
  un-paused autosave hook before the issue was diagnosed

**Bot B (2026-06-02):**
- 15:56:00 UTC — `~/.mempalace` restored from aside (for an unrelated reason)
- 15:56:00 UTC — first MCP tool call after restore
- `stat -c '%y' ~/.mempalace` confirmed recreation matched the tool call
  timestamp on a host that had only `locks/` and `wal/` subdirs (no
  `config.json`) — confirming the recreation is via the WAL mkdir path, not
  any restore/init flow

## Related

- #1305 added the hook-side kill switch but didn't cover the MCP server.
- #1510 was the test-isolation manifestation of this same pattern (test
  suite hit `_palace_root_exists()` returning True because something else
  had created `~/.mempalace`). Fixed in the test layer; runtime path
  unchanged.
- #504 asked for kill-switch guidance more generally; the runtime durability
  question lives downstream of that.

## Suggested fix

Move the WAL setup out of module scope into a lazy initializer called from
each write path. Importing the module no longer touches disk; the
kill-switch contract is preserved without `mcp_server.py` needing to know
about it.

A test asserting no top-level disk side effect from `import
mempalace.mcp_server` (when `~/.mempalace` is absent) would catch the next
regression of this shape — the existing test in #1547 covers the hook side
but not import-time MCP behavior.

## Environment

- mempalace 3.3.5 (latest on PyPI as of this report)
- Confirmed identical code path on `develop` HEAD
- Python 3.9, Linux (Amazon Linux 2023, arm64)
- Reproduces on a stock `uv tool install mempalace` setup
- Deployment: MCP server + hooks running together via OpenClaw agent runtime
  (not Claude Code); the kill-switch contract matters more in this shape
  because the MCP server is long-lived across many hook fires

mvalentsev:
## What does this PR do?

Fixes #1676. Importing `mempalace.mcp_server` recreated `~/.mempalace`, which defeated the kill-switch added in #1305.

The write-ahead-log setup ran at module scope:

```python
_WAL_DIR = Path(os.path.expanduser("~/.mempalace/wal"))
_WAL_DIR.mkdir(parents=True, exist_ok=True)   # creates ~/.mempalace as a side effect
```

So merely importing the server (which happens on every session start) ran `mkdir(parents=True)` and recreated `~/.mempalace`, even after a user removed it to engage the documented kill-switch (`hooks_cli._palace_root_exists()` returns `PALACE_ROOT.is_dir()`). That silently re-armed the autosave/mining hooks the user had disabled.

This moves the WAL directory setup into a lazy `_ensure_wal()` that runs from the write path (`_wal_log`) instead of at import. Importing the module no longer touches disk; the directory is created on the first real write, when the palace is being written to anyway.

The WAL is intentionally **not** gated on `_palace_root_exists()`: by the time a write reaches it, the ChromaDB/KG layer is already recreating the palace regardless, so gating would only drop audit records without preventing recreation. Whether MCP writes should honor the kill-switch at runtime is the broader question in #504.

Two intentional, verified side notes:

* The WAL directory `0o700` permissions are now re-asserted on each write (previously on each import); the WAL file is still created atomically with `0o600` by `_wal_log`.
* Importing the server no longer raises if `$HOME` is read-only; WAL failures are logged and never crash a tool call, consistent with the existing WAL contract.

## How to test

New regression tests:

```
python -m pytest tests/test_mcp_server.py -k TestImportKillSwitchSafety -v
```

* `test_import_does_not_recreate_palace_root` imports the server in a subprocess with a clean `$HOME` and asserts `~/.mempalace` is not created (fails on the old code, passes now).
* `test_wal_log_creates_dir_lazily_on_first_write` asserts the directory is created on first write with the expected `0o700`/`0o600` permissions.

Manual check, using a throwaway `$HOME` so your real palace is untouched:

```
HOME=$(mktemp -d) python -c "
from mempalace.hooks_cli import _palace_root_exists
import mempalace.mcp_server
print('palace recreated at import:', _palace_root_exists())  # False with this PR
"
```

## Checklist

* [x] Tests pass (`python -m pytest tests/ -v`)
* [x] No new hardcoded paths (WAL path is pre-existing and HOME-relative)
* [x] Linter passes (`ruff check .`)

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -387,21 +387,51 @@ def _refresh_vector_disabled_flag() -> None:
 # This provides an audit trail for detecting memory poisoning and
 # enables review/rollback of writes from external or untrusted sources.
 
-_WAL_DIR = Path(os.path.expanduser("~/.mempalace/wal"))
-_WAL_DIR.mkdir(parents=True, exist_ok=True)
-try:
-    _WAL_DIR.chmod(0o700)
-except (OSError, NotImplementedError):
-    pass
-_WAL_FILE = _WAL_DIR / "write_log.jsonl"
-# Atomically create WAL file with restricted permissions (no TOCTOU race).
-# os.open with O_CREAT|O_WRONLY and mode 0o600 creates the file if absent
-# or opens it if present, both in a single syscall.
-try:
-    _fd = os.open(str(_WAL_FILE), os.O_CREAT | os.O_WRONLY, 0o600)
-    os.close(_fd)
-except (OSError, NotImplementedError):
-    pass
+_WAL_FILE = Path(os.path.expanduser("~/.mempalace/wal")) / "write_log.jsonl"
+_WAL_INITIALIZED_DIR = None
+
+
+def _ensure_wal() -> None:
+    """Create (and re-harden) the WAL directory lazily, on the first write.
+
+    This must NOT run at import time: a user who removed ``~/.mempalace`` has
+    engaged the documented kill-switch (``hooks_cli._palace_root_exists()``,
+    #1305), and recreating the directory just by importing this module would
+    silently re-arm the autosave/mining hooks they disabled (#1676). Creating
+    it on the first real write keeps the kill-switch contract intact.
+
+    It is deliberately not gated on ``_palace_root_exists()``: by the time a
+    write reaches here the palace is already being recreated by the ChromaDB/KG
+    layer regardless, so gating would only drop audit records, not prevent
+    recreation. Runtime kill-switch enforcement for MCP writes is the broader
+    question tracked in #504.
+
+    Hardening is attempted once per directory and the path cached in
+    ``_WAL_INITIALIZED_DIR`` regardless of outcome (keyed on the path, so a
+    test repointing ``_WAL_FILE`` re-initialises), so a persistent failure on a
+    restricted filesystem does not retry on every write. ``mkdir`` runs only
+    when the initial ``chmod`` raises ``FileNotFoundError`` (EAFP). The parent
+    ``~/.mempalace`` keeps its umask mode, like the other palace directories;
+    the WAL file is created atomically with mode 0o600 by ``_wal_log``.
+    """
+    global _WAL_INITIALIZED_DIR
+    wal_dir = _WAL_FILE.parent
+    if _WAL_INITIALIZED_DIR == wal_dir:
+        return
+    try:
+        wal_dir.chmod(0o700)
+    except FileNotFoundError:
+        try:
+            wal_dir.mkdir(parents=True, exist_ok=True)
+            wal_dir.chmod(0o700)
+        except (OSError, NotImplementedError):
+            pass
+    except (OSError, NotImplementedError):
+        pass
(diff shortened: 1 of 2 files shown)


mvalentsev:
Reworked `_ensure_wal()` to drop the redundant per-write `mkdir` (`ddbb5d4`).

mvalentsev:
@gemini-code-assist `_ensure_wal()` now caches the initialised directory in `_WAL_INITIALIZED_DIR`, so the per-write `chmod` is skipped after the first init (keyed on the directory, so tests that repoint `_WAL_FILE` re-initialise). Pushed as `0d4b404`.

/gemini review

PR #1680, opened 2026-06-03
fix(hallways): paginate drawer fetch to avoid SQLite variable overflow on large wings (#1619)

### LINKED ISSUE #1619: compute_hallways_for_wing crashes (too many SQL variables) on wings > ~32k drawers

milla-jovovich:
compute_hallways_for_wing fetched the whole wing in a single
col.get(where={"wing": wing}). ChromaDB binds one SQL variable per matched id,
so on a wing larger than SQLITE_MAX_VARIABLE_NUMBER (32766) the call raised
"too many SQL variables" inside chromadb. The exception was caught, so the mine
completed — but the wing's hallway graph silently never built, and the
cross-wing tunnels promoted from it were starved, on exactly the large wings
that benefit most from navigation. Confirmed threshold: a 42,062-drawer wing
crashed; a 29,629-drawer wing succeeded.

Replace the single where-get with the established pagination pattern: count()
+ get(limit=5000, offset=...) filtered to the wing client-side — matching
miner.status, palace.regenerate_closets, and palace_graph.build_graph, which
already paginate to dodge the same 32766 limit.

Tests:
- test_hallways_pagination: a collection whose where-get raises (simulating the
  overflow) while count() + paginated get works — RED before, GREEN after.
- test_hallways: _fake_collection updated to the paginated API; existing
  hallway tests are unchanged in behavior.

Closes #1619.

### DIFF
--- a/mempalace/hallways.py
+++ b/mempalace/hallways.py
@@ -198,16 +198,32 @@ def compute_hallways_for_wing(
 
     min_count = max(1, int(min_count))
 
-    # 1. Query drawers for this wing.
+    # 1. Query drawers for this wing. Paginate the fetch and filter to the
+    #    wing client-side: a single get(where={"wing": wing}) binds one SQL
+    #    variable per matched id and overflows SQLite's
+    #    SQLITE_MAX_VARIABLE_NUMBER (32766) on wings > ~32k drawers (#1619).
+    #    Mirrors the established pagination in miner.status / palace /
+    #    palace_graph.
+    metadatas: list = []
     try:
-        results = col.get(where={"wing": wing}, include=["metadatas"])
+        total = col.count()
+        batch_size = 5000
+        offset = 0
+        while offset < total:
+            batch = col.get(limit=batch_size, offset=offset, include=["metadatas"])
+            batch_metas = (batch or {}).get("metadatas") or []
+            if not batch_metas:
+                break
+            metadatas.extend(
+                m for m in batch_metas if isinstance(m, dict) and m.get("wing") == wing
+            )
+            offset += len(batch_metas)
     except Exception:
         logger.warning(
-            "compute_hallways_for_wing: collection.get failed for %s", wing, exc_info=True
+            "compute_hallways_for_wing: collection fetch failed for %s", wing, exc_info=True
         )
         return []
 
-    metadatas = (results or {}).get("metadatas") or []
     if not metadatas:
         return []
 
--- a/mempalace/hallways.py
+++ b/mempalace/hallways.py
@@ -179,7 +179,12 @@ def compute_hallways_for_wing(
 
     Args:
         wing: wing name to scan.
-        col: ChromaDB collection — must support ``.get(where=..., include=...)``.
+        col: ChromaDB collection — must support ``.count()`` and paginated
+            ``.get(limit=..., offset=..., include=...)``. The fetch is filtered
+            to ``wing`` client-side rather than via ``.get(where={"wing": ...})``,
+            which binds one SQL variable per matched id and overflows SQLite's
+            ``SQLITE_MAX_VARIABLE_NUMBER`` on large wings (#1619). Fake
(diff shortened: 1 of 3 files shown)


igorls:
Review notes, updated after a real large-wing smoke test:

- CI is red because `lint` fails on formatting, not because pytest is failing. The log shows `ruff format --check .` would reformat `tests/test_hallways_pagination.py`. Running `uv run ruff format tests/test_hallways_pagination.py` should clear the failing check.
- The new regression test is useful, but it is still a mocked 3-drawer collection. It verifies that `compute_hallways_for_wing()` avoids the `get(where=...)` path; it does not itself prove the >32k ChromaDB failure case.
- I ran a real local ChromaDB smoke test against this PR code. With 33,000 and 42,062 drawers in one wing, the old `col.get(where={"wing": ...})` path reproduced `too many SQL variables`, while the new paginated `count()` + `get(limit=..., offset=...)` path successfully built the hallway. For the 42,062-drawer case, the returned hallway had `co_occurrence_count == 42062`.
- Minor follow-up: `mempalace/hallways.py` still documents `col` as needing `.get(where=..., include=...)`, but this PR changes the contract to `count()` + paginated `get(limit=, offset=, include=...)`. Worth updating the docstring so future fake collections/backends implement the right shape.

So the original bug is real, and this PR fixes the direct large-wing ChromaDB hallway-computation path; the remaining CI failure is just formatting.

PR #1684, opened 2026-06-03
feat(integrations): Hermes memory provider (revives #3, ABC-current + dim-fix)

raman325:
Revives [#3](https://github.com/MemPalace/mempalace/pull/3) (`feat: Hermes memory provider integration`), which was closed for going stale per @bensig:

> *"This has been open a while and conflicts with main. The Hermes integration space has also evolved. If you'd like to continue, a rebase + update to match the current codebase would be needed."*

This PR is that rebase + update. Targets `develop` per CONTRIBUTING.md. Original 4 commits from @ZK-Snarky are preserved verbatim at the base; everything beyond is additive across 16 focused commits.

## What this adds

A drop-in MemPalace memory provider for [Hermes](https://github.com/NousResearch/hermes-agent) (NousResearch/hermes-agent [#6323](https://github.com/NousResearch/hermes-agent/issues/6323)). One command after `pip install mempalace`:

```bash
mempalace hermes install
```

Files land in `~/.hermes/plugins/mempalace/` (the canonical user-installed plugin directory Hermes' `discover_memory_providers()` scans) with a `plugin.yaml` manifest. `~/.hermes/config.yaml` gets `memory.provider: mempalace`. Optional interactive backfill of existing Hermes sessions.

## Why this lives at `mempalace/integrations/hermes/` and not top-level `integrations/hermes/`

Top-level `integrations/openclaw/` exists and the natural read is "put it next to that one." The asymmetry comes from how each integration is consumed at runtime:

- **openclaw** is a Claude Code skill — a single `SKILL.md` that Claude Code reads from the source repo. It doesn't need to be importable Python and it doesn't need to ship in the wheel.
- **The Hermes integration is a Python class** implementing the `MemoryProvider` ABC — Hermes' plugin loader imports the module and calls methods on it. `mempalace hermes install` has to *find* the source files at runtime after a wheel install in order to copy them into `~/.hermes/plugins/mempalace/`.

`pyproject.toml` has `[tool.hatch.build.targets.wheel] packages = ["mempalace"]`. Anything at the top level (including a hypothetical `integrations/hermes/`) is **not in the wheel**. Verified by building locally: with the integration outside the package, `pip install mempalace && mempalace hermes install` fails to locate the source for every user without a source checkout — silently breaking the install flow this PR sells.

Placing the integration inside the package (`mempalace/integrations/hermes/`) makes the standard subpackage layout work. The install command resolves it via `importlib.util.find_spec("mempalace.integrations.hermes")`; the same code path holds in source-tree, editable, and wheel installs; no custom hatchling config needed. Verified with `uv build` — the wheel contains:

```
mempalace/integrations/__init__.py
mempalace/integrations/hermes/__init__.py
mempalace/integrations/hermes/backfill.py
mempalace/integrations/hermes/README.md
```

The alternative would be `[tool.hatch.build.targets.wheel.force-include]` mapping top-level `integrations/hermes/` into the wheel under a different name. That preserves visual symmetry with openclaw but adds custom hatch config and a name-mangling step. Happy to switch if the symmetry matters more than the simpler standard layout.

## What's new vs. PR #3 (the ABC has evolved)

| Method | Before | After |
|---|---|---|
| `initialize` | `(config: dict)` | `(session_id, **kwargs)` with cron-context guard |
| `sync_turn` | `(turn: dict)` | `(user_content, assistant_content, *, session_id, messages)` |
| Tool surface | `get_tools()` with embedded `handler` attribute | `get_tool_schemas()` + `handle_tool_call(name, args) -> str` |
| `on_pre_compress` | returned `list[dict]` | returns `str` (hint into the compression summary prompt) |
| Class | plain `MempalaceProvider` | inherits `MemoryProvider` (ABC stub fallback for non-Hermes envs) |

Opt-in hooks added: `on_session_switch`, `on_delegation`, `on_memory_write`, `on_turn_start`, `queue_prefetch`, `save_config`, `post_setup`.

The provider is **inactive** under `agent_context in {"cron", "flush"}` or `platform == "cron"` so system-generated turns don't corrupt the user representation (matches the discipline in built-in providers like honcho). The `_cron_skipped` flag is cleared at the start of every non-cron `initialize()` so a cron init doesn't permanently poison the instance.

## Tool surface — 27 of 30 documentable MCP tools

The provider exposes the full agent-facing mempalace MCP surface: all 27 tools across Search & Browse, Knowledge Graph, Palace Graph (incl. explicit tunnels), Write/Diary, and the silent-checkpoint ack. The remaining 3 tools — `mempalace_sync`, `mempalace_hook_settings`, `mempalace_reconnect` — are intentionally **omitted**. They're host/admin operations that belong to the MCP host owner, not the conversation loop, and exposing them to the agent would be confusing at best and destructive at worst.

Two changes were required to make this surface actually reachable:

1. **Drop the `_initialized` gate from `get_tool_schemas()`.** Hermes' memory manager snapshots `get_tool_schemas()` at provider-registration time, **before** `initialize()` runs (`agent.memory_manager._register_provider`, registration-time dispatch-table build). The gate caused 0 tools to be registered with the dispatch router even though `get_all_tool_schemas()` (queried each turn for system-prompt assembly) returned the full set — the model saw the tools but dispatch couldn't route to them, surfacing as `Unknown tool: mempalace_status` in real openclaw use. Schemas describe interface; readiness belongs in `handle_tool_call`.
2. **Dispatch the passthrough tools via name-derivation.** A `_dispatch_mcp_passthrough` helper derives `tool_<name>` from `mempalace_<name>` (with one explicit remap for `mempalace_traverse` → `tool_traverse_graph`) and forwards to `mempalace.mcp_server`. The refactor also pulled `handle_tool_call`'s cyclomatic complexity below the C901 limit (was 31 with an if-chain).

The OpenClaw memory skill (`integrations/openclaw/SKILL.md`) was also stale relative to this surface (only 19 tools documented). Parity PR: #1719.

## The embedding-dimension fix

Three earlier Hermes-side PRs (NousResearch/hermes-agent [#5671](https://github.com/NousResearch/hermes-agent/pull/5671), [#12203](https://github.com/NousResearch/hermes-agent/pull/12203), [#9761](https://github.com/NousResearch/hermes-agent/pull/9761)) all called `chromadb.PersistentClient.get_or_create_collection(...)` directly without passing `embedding_function=`. ChromaDB bound its default 384-dim function to the collection; users with palaces built on `bge-m3` (1024-dim) or `embeddinggemma-300m` would hit a hard dimension mismatch on the next write — `@Motokiyo` flagged this in the hermes-agent issue thread.

This PR routes ChromaDB access (both live writes and `backfill.py`) through `mempalace.backends.chroma.ChromaBackend.get_or_create_collection`, which binds the canonical embedding function from `mempalace.embedding.get_embedding_function()`. Existing palaces import without rebuild. The end-to-end test `test_initialize_opens_chroma_via_backend` asserts this construction path; `test_status_tool_counts_seeded_drawers` confirms the provider reads back data written via raw `chromadb.PersistentClient` (the regression check).

## Review-feedback rounds

The PR went through two automated review rounds since opening. All findings are addressed in the commits below.

**Round 1 — Gemini Code Assist (4 findings):**
- `backfill.py` instantiated `PersistentClient` per exchange without `embedding_function=` → reintroduced dim-mismatch + per-turn perf hit. Now goes through `ChromaBackend` once per `backfill()` call.
- `_tool_diary_read` loaded the whole `diary.jsonl` to keep the tail → `collections.deque(f, maxlen=n)`.
- Worker exited on stop signal before draining the queue → loop now waits for both stop + empty.

**Round 2 — Copilot (11 findings):**
- `_initialized` set True even when backend init failed → now reflects actual readiness.
- `KnowledgeGraph` instantiated and never closed in 3 sites → wrapped in `try/finally: kg.close()`.
- `wing` config field documented as "default wing" but always overridden by keyword classification → honored when set.
- Used `session_id` as the `room` value (high-cardinality, inconsistent with backfill) → stable `room: "conversations"`, session_id moved to dedicated metadata field.
- `_classify_wing` used substring matching (`kw in text`) → word-boundary regex.
- CLI install command: dropped Hermes-presence guards from PR #3, used deprecated `os.system`, missed `--upgrade`, assumed POSIX venv layout, hard-coded `~/.mempalace/palace` for backfill ignoring `MEMPALACE_PALACE_PATH`, line-based YAML edit broke on `memory: ~` / inline comments / non-2-space indent → rebuilt with cross-platform venv detection, `subprocess.run` + `--upgrade`, palace-path resolver honoring env vars, `yaml.safe_load` / `safe_dump` round-trip with atomic writes via tmp + rename, and `--hermes-home '.' '/'` rejected.
- `datetime.utcnow()` in backfill → `datetime.now(timezone.utc)`. 16-char doc_id from text-prefix hash → 32-char hash over full content + timestamp.

**Round 3 — Copilot (10 findings):**
- Parallel-path gaps I missed in Round 2: `_mine_session` skipped `_normalize_content`; `on_session_end` and `on_memory_write` skipped the readiness gate; `backfill.classify_wing` skipped the word-boundary fix. All closed; new tests pin the regressions.
- Module docstring + `integrations/hermes/README.md` said "first match wins" and listed `MEMPALACE_COLLECTION_NAME` — both stale after the env-precedence and collection-name-removal fixes. Aligned.
- **Packaging:** `[tool.hatch.build.targets.wheel] packages = ["mempalace"]` meant top-level `integrations/` didn't ship in the wheel → moved to `mempalace/integrations/hermes/` (rationale above).
- `_update_hermes_config_yaml` returned `(bool, str)` where `False` collapsed "already set" and "couldn't parse" → now `(status, message)` with `"updated"` / `"noop"` / `"error"`; install exits non-zero only on `"error"`.
- `_tool_status` / `_list_wings` / `_list_rooms` materialized every drawer's metadata for counts → `STATUS_SCAN_LIMIT = 5000` cap, with a `"truncated"` field on the response so the model knows the breakdown is sampled.

## Tests

**62 new tests** across two files; full mempalace suite **2,340 passed, 3 skipped, 0 failed in ~68s** locally (3 skips are pre-existing).

`tests/test_hermes_integration.py` (40 tests, reuses `palace_path` / `seeded_collection` / `seeded_kg` / autouse `_isolate_home` from `tests/conftest.py`):

- ABC contract shape (`name`, `is_available`, config schema, tool-schema visibility before/after init)
- Cron-context guard (parameterized across `agent_context=cron`, `agent_context=flush`, `platform=cron`)
- Provider remains inactive after `_initialized=False` (worker queue isn't fed by `on_session_end` / `on_memory_write` — Round 3 regression coverage)
- `_normalize_content` flattens Anthropic list-shaped `content` without persisting the `repr`
- `_match_wing_by_keywords` respects word boundaries (substring would route `"said"` to `wing_ai`)
- `backfill.classify_wing` matches the live provider's classification (no live/backfill divergence)
- `sync_turn` end-to-end: enqueues, worker drains, drawer lands with `source=hermes` and the configured wing
- Tool handlers against seeded data: `mempalace_status`, `_list_wings`, `_list_rooms`, `_search`, KG add+query, diary roundtrip
- Config precedence: env vars > `$HERMES_HOME/mempalace.json` > defaults; empty env vars treated as unset
- `collection_name` intentionally absent from the schema (regression test)
- Session-switch bookkeeping (`reset=True` clears turn counter; `reset=False` preserves it)
- Shutdown drains the worker thread

`tests/test_hermes_install_cli.py` (22 tests, addresses Copilot's "no install coverage" finding):

- Path resolution: explicit `--hermes-home` / `$HERMES_HOME` / default precedence
- Refuses `--hermes-home .` and `--hermes-home /` (rather than silently installing into CWD)
- Venv interpreter detection across POSIX (`venv/bin/python3`, `python`) and Windows (`venv\Scripts\python.exe`), falls back to `sys.executable`
- `MEMPALACE_PALACE_PATH` env-var precedence + empty-value handling
- `_atomic_write_text` creates parents and leaves no `.tmp` after success
- `_update_hermes_config_yaml`: missing file, malformed YAML, scalar `memory: ~`, already-set, replace existing — each returns the correct `"updated"` / `"noop"` / `"error"` status

## Commits

20 commits since `upstream/develop` — preserved ZK-Snarky's 4 plus 16 mine:

```
fix(hermes-cli):   distinguish noop vs error in YAML edit return
fix(packaging):    ship Hermes integration as mempalace.integrations.hermes
perf(hermes):      cap metadata scans in status / list_wings / list_rooms
docs(hermes):      align config docs with actual precedence + drop stale env var
fix(hermes):       close parallel-path consistency gaps from Copilot round 2
fix(hermes-cli):   rebuild install command for safety + palace consistency
fix(hermes):       doc_id collisions + naive datetime in backfill
fix(hermes):       word-boundary wing keyword match + honor configured wing + stable room
fix(hermes):       close KnowledgeGraph handles in all three call sites
fix(hermes):       tighten the verbatim-persistence contract
style(hermes):     hoist mempalace internal imports to module top
fix(hermes):       drain queue on shutdown + tail-read diary file
fix(hermes):       route backfill writes through ChromaBackend
test(hermes):      deepen coverage with palace/KG fixtures from conftest
test(hermes):      add provider smoke tests
feat(hermes):      rebuild integration for current MemoryProvider ABC
fix:               correct 12 bugs found in Codex audit (KG API, prefetch shape, …)  [@ZK-Snarky]
fix:               remove unused imports and rename ambiguous variable (ruff lint)   [@ZK-Snarky]
fix:               cache ChromaDB client to prevent SQLite lock races                [@ZK-Snarky]
feat:              Hermes memory provider integration                                [@ZK-Snarky]
```

## Known follow-ups (not blocking)

- `ChromaBackend.close()` on provider `shutdown()`: would release the SQLite file lock so the same Python process can reopen the palace cleanly (matters for tests and gateway-reload scenarios). Skipped here because there's no public close method on the backend yet; needs an upstream API.
- `prefetch()` is synchronous on the hot path. `queue_prefetch()` is not yet overridden; a background-prewarm rewrite would match the discipline in `honcho`. Significant enough to warrant its own PR.
- For palaces above `STATUS_SCAN_LIMIT`, the wing breakdown is sampled. A SQLite-sidecar counter would give exact figures cheaply; mempalace doesn't expose one yet.

## Cross-references

- Revives #3 (this repo)
- Closes (from this side) the upstream invitation in NousResearch/hermes-agent#6323 to publish as a standalone-installable provider
- Supersedes the dim-mismatch-blocked NousResearch/hermes-agent PRs #5671, #12203, #9761
- Companion PR: #1719 — catches up `integrations/openclaw/SKILL.md` to the same 27-tool surface

cc @ZK-Snarky for awareness — happy to fold in further changes you want.


xg-gh-25:
This is impressively thorough integration work — the dimension-mismatch fix alone is worth merging. Three observations from our production memory architecture:

**1. The cron-context guard is exactly right.** We learned this the hard way: system-generated turns (maintenance, backfills, health checks) will absolutely poison your user representation if you don't gate them. Your `agent_context in {"cron", "flush"}` check mirrors our approach. One edge case to watch: if a cron job *starts* a session, then a user *continues* that session later, does the `_cron_skipped` flag persist? Your "cleared at the start of every non-cron initialize()" suggests you've already caught this.

**2. The ChromaDB embedding-function routing is the right architecture.** Going through `ChromaBackend.get_or_create_collection()` instead of raw `chromadb.PersistentClient` prevents the dimension-mismatch footgun and keeps the embedding function centralized. We do the same. The only gap I'd watch for: if someone manually creates a palace with one embedding model, then switches models in config later, ChromaDB will silently accept writes with mismatched dimensions *until* it hits a query. You might want a startup dimension-check that compares the configured model's dimension against existing collections and fails fast if they don't match.

**3. The `STATUS_SCAN_LIMIT = 5000` cap is pragmatic.** Exact counts are expensive at scale; sampled breakdowns are fine for the model. We use a similar pattern. One refinement: if you're returning `"truncated": true`, also return the *sample coverage* (e.g., `"scanned": 5000, "total_estimated": 12000`) so the model knows how partial the view is. Otherwise it might assume the breakdown is complete.

**The pkg structure debate (mempalace/integrations/hermes/ vs top-level):** Your reasoning is sound — the integration has to be importable Python for `mempalace hermes install` to work post-wheel. The symmetry with `integrations/openclaw/` isn't worth the hatchling force-include complexity. The current layout is correct.

**One follow-up question:** The PR mentions `prefetch()` is synchronous on the hot path and `queue_prefetch()` isn't overridden yet. How does this impact session-switch latency in practice? If Hermes switches sessions mid-conversation, does the first user turn in the new session block on a cold palace load, or is there a warmed cache?

Excellent work overall — the test coverage (62 new tests) and the round-trip review discipline are production-grade.

---
*This comment reflects patterns from SwarmAI's memory architecture. Discussion: [T-MEM](https://github.com/xg-gh-25/SwarmAI/discussions/4)*

igorls:
This is one of the more substantial agent-support contributions, and the Hermes direction is worth preserving. I reviewed it during the community merge pass, but did not include it in the clean integration branch because it still needs a rebase and some architectural tightening.

Main blockers I see:

- It conflicts with current `develop`, especially around `mempalace/cli.py`.
- Some paths bypass existing backend/write helpers, which risks embedding-function/dimension drift and duplicated locking/dedup behavior.
- The provider/backfill split needs to keep routing, normalization, and metadata conventions identical between live and historical ingest.
- Large-palace status/list operations should use the current fast paths or bounded pagination rather than full metadata scans.
- Packaging should be reliable from a wheel install, not dependent on top-level files being adjacent to the installed package.

A mergeable revision would probably be smaller in two PRs: first the Hermes MCP/provider core with tests, then backfill/install docs. The core idea is good; I just want it to land through the same local-first, verbatim, backend-safe paths the rest of MemPalace now uses.

raman325:
closing this in favor of the two linked PRs!

PR #1702, opened 2026-06-06
feat(migrate): mempalace migrate-wings — normalize legacy wing names (#1675 follow-up)

igorls:
## What

Adds `mempalace migrate-wings [--dry-run] [--yes]` — a one-time, idempotent migration that normalizes legacy wing names (strips leading/trailing separators) so palaces built before the wing-name rule (#1675) keep their memories discoverable under the new name.

The problem: a Claude-Code path-encoded dir like `-home-user-proj` derived `_home_user_proj`. The normalized derivation strips that to `home_user_proj`, so new mining/diary writes land on a *different* wing than the existing drawers — the history **splits** (it isn't lost, just stranded under the old name). This closes that split.

## How (design verified against the codebase)

- **IDs left untouched.** Drawer/closet IDs embed the wing as an opaque prefix that is **never decoded back into a wing** (grep-verified — nothing splits a wing out of an ID), and mining idempotency keys on `source_file` (`prefetch_mined_set`), not on recomputed IDs. So the migration only re-keys the **metadata `wing` field** in place: closet `→drawer_id` pointers stay valid (no document rewrite) and future mining still skips already-mined files (no duplicates). This avoids a risky extract-rebuild.
- **Stores updated:** drawer metadata + closet metadata (`wing`), and the `topics_by_wing` registry (re-keyed, merging on collision). **Tunnels** resolve via existing read-time normalization (`palace_graph._normalize_wing`) and need no rewrite.
- **Backend-agnostic** (uses the `get`/`update` collection abstraction), **idempotent**, **dry-run-able**, with collision **merges** reported.
- The leading/trailing strip is applied by the migration itself, so it's correct whether or not the running build's `normalize_wing_name` already carries the #1675 change.

## Tests

`tests/test_migrate_wings.py`: pure planner coverage (strip leading/trailing, no-op on clean wings, ignore empty/non-string/all-separator, collision→same target) + hermetic backend integration (relabel, merge-into-existing, dry-run changes nothing, idempotency). 8 tests; `test_migrate.py` (11) and `test_cli.py` (69) still pass.

## Merge ordering

- Pairs with **#1675** (the derivation change) — useful independently (leading-underscore wings already exist today and are rejected by MCP writes), and aligns old data once #1675 lands.
- **Depends on #1700** for green `develop`: this branch is cut from the currently-red `develop`, so its CI will show the **pre-existing #1245 failures** (the `search_memories` C901 + 2 `diary_write` tests) that #1700 fixes — *not* anything from this change. Once #1700 merges, I'll rebase and CI goes green. The migration files themselves are lint-clean and fully tested.

### DIFF
--- a/docs/recovery/wing-name-migration.md
+++ b/docs/recovery/wing-name-migration.md
@@ -1,0 +1,63 @@
+# Recovery: legacy wing names split after the normalization change
+
+**Companion to #1675.** `normalize_wing_name` now strips leading and trailing
+separators, so a path-encoded project dir like `-home-user-proj` derives the
+wing `home_user_proj` instead of `_home_user_proj`. Palaces mined before that
+change filed drawers under the old, separator-padded name. New mining and diary
+writes land on the new name, so the two no longer meet — the history is
+**split**, not lost. `mempalace migrate-wings` re-unites them.
+
+## Symptom
+
+After upgrading, a project that used to surface its memories returns less than
+expected, and `mempalace status` shows two wings for one project — e.g. both
+`_home_user_proj` (old drawers) and `home_user_proj` (newly mined). MCP writes
+to the padded wing may also have been rejected, since `sanitize_name` does not
+accept a leading underscore.
+
+## Recovery
+
+Preview first — this never modifies anything:
+
+```bash
+mempalace migrate-wings --dry-run
+mempalace migrate-wings --dry-run --palace /path/to/palace
+```
+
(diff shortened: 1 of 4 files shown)


igorls:
Rebased onto `develop` (now includes #1700), so the inherited-red caveat above no longer applies. Local gate on the rebased branch: `ruff check` + `ruff format --check` clean, `test_migrate_wings.py` (8) green. Also added `docs/recovery/wing-name-migration.md` covering when/how to run `migrate-wings`.

Issue #1660, opened 2026-05-30
Windows: plugin hooks.json bash wrapper mangles paths, causing Stop/PreCompact hooks to fail


w1tc4:
## Environment
- OS: Windows 11
- Shell: PowerShell (Claude Code runs hooks via PowerShell)
- Claude Code version: 2.1.157
- MemPalace plugin version: 3.3.6

## Problem

The Stop and PreCompact hooks in the plugin's `hooks.json` use a bash wrapper:

```json
"command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/mempal-stop-hook.sh\""
```

On Windows, `CLAUDE_PLUGIN_ROOT` expands to a path like:
```
C:\Users\Rachael\.claude\plugins\cache\mempalace\mempalace\3.3.6
```

When this is passed to `bash`, the backslashes are interpreted as escape sequences, producing a mangled path:
```
C:UsersRachael.claudepluginscachemempalacemempalace3.3.6/hooks/mempal-stop-hook.sh
```

This causes every Stop and PreCompact hook invocation to fail with:
```
Stop hook error: Failed with non-blocking status code: /bin/bash:
C:UsersRachael.claudepluginscachemempalacemempalace3.3.6/hooks/mempal-stop-hook.sh: No such file or directory
```

## Root Cause

The `.sh` wrapper scripts are just thin pass-throughs:

```sh
#!/usr/bin/env bash
mempalace hook run --hook stop --harness claude-code
```

They add no value on Windows (and break things) — the underlying `mempalace` command is available directly on PATH.

## Fix

Replace the bash wrapper calls in `hooks.json` with direct mempalace commands:

```json
{
  "Stop": [{
    "hooks": [{
      "type": "command",
      "command": "mempalace hook run --hook stop --harness claude-code",
      "timeout": 30
    }]
  }],
  "PreCompact": [{
    "hooks": [{
      "type": "command",
      "command": "mempalace hook run --hook pre-compact --harness claude-code",
      "timeout": 90
    }]
  }]
}
```

This works cross-platform since `mempalace` is on PATH after installation regardless of OS.

## Workaround

Manually edit the cached `hooks.json` at:
```
%USERPROFILE%\.claude\plugins\cache\mempalace\mempalace\<version>\hooks\hooks.json
```

Note this reverts on plugin update.

Related: PR #1658 is open from this user's fork for a separate `persist_directory` fix.

PR #1713, opened 2026-06-06
fix: call Claude plugin hooks directly on Windows

Bortlesboat:
## Summary
- preserve the Claude plugin Stop, PreCompact, and SessionEnd wrapper scripts
- pass each expanded plugin path to bash as positional data, normalize Windows backslashes inside bash, then execute the wrapper
- add regression coverage pinning the cross-platform bridge for all three lifecycle events

## Why
Issue #1660 reports that a Windows `CLAUDE_PLUGIN_ROOT` reaches bash as a backslash path. When the path is parsed directly as bash source, the backslashes are consumed as escapes and every hook fails.

This revision follows maintainer guidance: it keeps the wrappers (including the stale-binary protection being developed in #1690), retains SessionEnd's background behavior, and only changes how the wrapper path crosses the PowerShell-to-bash boundary.

## Verification
- `uv run pytest tests/test_claude_plugin_hook_config.py -q` (`8 passed`)
- `uv run ruff check tests/test_claude_plugin_hook_config.py`
- `uv run ruff format --check tests/test_claude_plugin_hook_config.py`
- `python -m json.tool .claude-plugin/hooks/hooks.json`
- `git diff --check`

## AI disclosure
OpenAI Codex assisted with the conflict resolution and regression update; the final diff and test results were reviewed locally.


igorls:
Thanks for tackling the Windows plugin hook path issue. The underlying problem is real: `bash "${CLAUDE_PLUGIN_ROOT}/..."` can be fragile when the plugin root expands to a Windows path.

I did not merge this version during the agent-support pass because #1690 adds stale-binary protection in the wrapper scripts, and replacing the wrappers with direct `mempalace hook run` commands would bypass that guard. There is also the `pre-compact`/`precompact` mismatch noted in the existing review.

A mergeable follow-up could take one of two shapes:

- Keep the wrapper scripts, but make `hooks.json` invoke them in a Windows-safe way that preserves the wrapper's version/runner selection logic.
- Or move the stale-binary/version guard into `mempalace hook run` itself, then direct commands become safe.

Either way, please keep the SessionEnd hook in mind too; current `develop` has Stop, PreCompact, and SessionEnd coverage.

Bortlesboat:
Reworked in 8ca2595 following the maintainer guidance: the current develop branch is merged, all three lifecycle wrappers are preserved, Windows backslashes are normalized only after the expanded path enters bash as positional data, and SessionEnd keeps its background wrapper. The focused config suite passes 8/8 with Ruff, format, JSON, and diff checks clean.

PR #1717, opened 2026-06-06
fix(mcp): drop top-level anyOf from diary_write schema

### LINKED ISSUE #1711: mempalace_diary_write input_schema has top-level anyOf — Anthropic API rejects the whole tools array (400)
## Summary

The `mempalace_diary_write` MCP tool advertises an `input_schema` with an
`anyOf` at the **top level**. The Anthropic Messages API forbids
`oneOf`/`allOf`/`anyOf` at the top level of any tool's `input_schema`, and it
validates the entire `tools` array as a unit. So a single client that registers
the mempalace MCP server alongside Anthropic models gets **every** request
rejected with a 400 — not just `diary_write` calls. The whole session is dead
on arrival.

```
400 invalid_request_error:
tools.77.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level
```

(`tools.77` is just wherever mempalace's tool happened to land in that client's
tool list.)

## Environment

- mempalace **3.4.0** (latest, `uv tool install mempalace`), entry point `mempalace-mcp`
- macOS, Python via uv
- Surfaced through an MCP host (Oh My Pi / omp) that forwards MCP tool schemas
  straight to the Anthropic Messages API without sanitizing them. Clients that
  *do* strip top-level combinators (e.g. some other harnesses) mask the bug,
  but the schema is still invalid per Anthropic's contract.

## The offending schema

Captured live from `mempalace-mcp` via a `tools/list` JSON-RPC call (30 tools
total; exactly one is affected):

```json
{
  "type": "object",
  "properties": {
    "_i": { "type": "string" },
    "agent_name": { "type": "string", "description": "..." },
    "entry": { "type": "string", "description": "..." },
    "topic": { "type": "string", "description": "..." },
    "wing": { "type": "string", "description": "..." },
    "content": { "type": "string", "description": "Alias for 'entry' ..." }
  },
  "required": ["agent_name", "_i"],
  "anyOf": [
    { "description": "{required: [\"entry\"]}" },
    { "description": "{required: [\"content\"]}" }
  ],
  "additionalProperties": false
}
```

The `anyOf` is clearly meant to express *"require either `entry` or `content`"*
(it lines up with the `content`-alias work in #1700). Two problems:

1. **It's invalid for Anthropic** — top-level `anyOf` is rejected, taking down
   the entire request.
2. **It doesn't even do anything** — each branch contains only a `description`
   string, no `required` (or any) constraint, so it enforces nothing. It's a
   no-op that happens to be toxic.

## Reproduction

```bash
printf '%s\n' \
 '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"probe","version":"0"}}}' \
 '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
 '{"jsonrpc":"2.0","id":2,"method":"tools/list"}' \
 | mempalace-mcp \
 | python3 -c 'import json,sys; [print(t["name"], list(t["inputSchema"])) for l in sys.stdin if l.strip() for t in (json.loads(l).get("result",{}).get("tools") or []) if any(k in t.get("inputSchema",{}) for k in ("anyOf","oneOf","allOf"))]'
# -> mempalace_diary_write [...'anyOf'...]
```

Then registering this server with any Anthropic-backed MCP host that forwards
schemas verbatim yields the 400 above on the first turn.

## Expected

`diary_write`'s schema should not use a top-level `anyOf`/`oneOf`/`allOf`, so it
is accepted by the Anthropic Messages API (and stays valid JSON Schema for the
"either entry or content" intent).

## Suggested fix

Drop the top-level `anyOf` entirely. The "one of entry/content required" rule is
already enforced at the handler/dispatch layer (the `content`→`entry` remap from
#1700), so the schema doesn't need to encode it. If you want to keep an
advisory hint, move it out of the top level — e.g. mention it in the field
`description`s, or express it with a top-level-legal construct. The simplest,
fully-compatible schema is just:

```json
{
  "type": "object",
  "properties": { "...": "..." },
  "required": ["agent_name", "_i"],
  "additionalProperties": false
}
```

## Workaround (for affected users, until this ships)

A thin stdio proxy in front of `mempalace-mcp` that strips top-level
`anyOf`/`oneOf`/`allOf` from `tools/list` results before they reach the model.
Requests pass through untouched; only the advertised schema is cleaned. Happy to
share the script if useful.

Thanks for mempalace — glad to test a fix.

Hexecu:
## What does this PR do?

`mempalace_diary_write` declared a top-level `anyOf` in its `input_schema` to require either `entry` or `content`. Anthropic's Messages API rejects any tool whose input schema has a top-level `anyOf`/`oneOf`/`allOf` and returns a 400 for the whole `tools` array, so every MCP session failed to start (#1711). The same breakage was reported on Codex.

The either/or constraint is already enforced at dispatch: `content` is remapped to `entry` before the handler runs, and a missing value returns `-32602`. Dropping the combinator restores API compatibility without weakening validation.

Added a `tools/list` regression test asserting no tool schema exposes a top-level `anyOf`/`oneOf`/`allOf`.

Closes #1711

## How to test

```
uv run pytest tests/test_mcp_server.py -v
```

`test_no_tool_schema_uses_top_level_combinator` fails on the previous schema and passes with the fix. Existing dispatch tests (`test_diary_write_content_aliases_entry`, `test_missing_required_returns_32602_with_param_name`) confirm the entry/content behaviour is unchanged.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -2662,13 +2662,10 @@ TOOLS = {
                     "description": "Alias for 'entry' — accepted because add_drawer uses 'content'. Provide either 'entry' or 'content'; 'entry' wins if both are given.",
                 },
             },
-            # agent_name is always required; 'entry' or its alias 'content' must
-            # be present (the server remaps content->entry at dispatch).
+            # 'entry' (or its alias 'content') is enforced at dispatch, not via a
+            # top-level anyOf: Anthropic rejects schemas with a top-level
+            # anyOf/oneOf/allOf and drops the whole tools array (400).
             "required": ["agent_name"],
-            "anyOf": [
-                {"required": ["entry"]},
-                {"required": ["content"]},
-            ],
         },
         "handler": tool_diary_write,
     },
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -537,6 +537,20 @@ class TestHandleRequest:
         assert "mempalace_add_drawer" in names
         assert "mempalace_kg_add" in names
 
+    def test_no_tool_schema_uses_top_level_combinator(self):
+        """Anthropic's Messages API rejects a tool whose input schema has a
+        top-level anyOf/oneOf/allOf and drops the entire tools array with a
+        400, killing the session (#1711). Cross-tool constraints must be
+        enforced at dispatch instead.
+        """
+        from mempalace.mcp_server import handle_request
+
+        resp = handle_request({"method": "tools/list", "id": 2, "params": {}})
+        for tool in resp["result"]["tools"]:
+            schema = tool["inputSchema"]
+            for keyword in ("anyOf", "oneOf", "allOf"):
+                assert keyword not in schema, f"{tool['name']} schema has top-level {keyword}"
+
     def test_null_arguments_does_not_hang(self, monkeypatch, config, palace_path, seeded_kg):
         """Sending arguments: null should return a result, not hang (#394)."""
         _patch_mcp_server(monkeypatch, config, seeded_kg)


Hexecu:
For reference, the tool schema shape Anthropic documents is always a top-level `type: object` with `properties`/`required` — there's no top-level combinator in any of the examples:
https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/define-tools

A top-level `anyOf` falls outside that shape, which is why the API rejects the whole `tools` array with a 400 rather than just ignoring the offending tool. Moving the either/or check to the dispatch layer keeps the same guarantee while staying within the documented schema.

gnyrd:
Independent confirmation, plus a third failure mode for the evidence pile.

Client-side behavior differs by client. Claude Code (macOS, current build) does not 400 the whole tools array like the Messages API does in #1711. It silently drops just the offending tool: after upgrading an editable install to v3.4.0, 29 of 30 mempalace tools registered and mempalace_diary_write was the only one missing, with no error surfaced. A direct stdio tools/list probe of the same server shows it advertising all 30 tools, diary_write schema keys [type, properties, required, anyOf]. The silent-drop variant is nastier to notice than the 400, and the hooks path (hooks_cli calling tool_diary_write in-process) keeps auto-save diaries working, which masks the breakage for most users.

Verification of this PR: both hunks apply cleanly onto the v3.4.0 tag. With the patch applied, the probe reports no top-level combinators on any tool and diary_write schema keys [type, properties, required]. Dispatch-level content-to-entry aliasing and the signature-based missing-parameter diagnostic cover the dropped constraint, as the PR comment says.

Carrying this as a local patch on v3.4.0 until it lands in a release. Would be good to see it merged.

moeen-dotcom:
Confirming this is still broken on **v3.4.0** (latest release) and that this PR's diff fixes it.

Independent repro on **Claude Code (macOS, current build)**: the client doesn't 400 the whole tools array the way the Messages API does in #1711 — it **silently drops just `mempalace_diary_write`** from the registry while every other tool registers fine. `mempalace_diary_read`, `status`, `add_drawer` etc. all work; only `diary_write` is missing, with no error surfaced. It's invisible to clients, so users burn multiple sessions chasing "stale process / reconnect / restart" before finding the real cause (the root-level `anyOf` is the only top-level combinator in the whole `TOOLS` registry, hence the only tool that disappears). This matches @gnyrd's report above.

The fix is correct and behavior-preserving — the `entry`/`content` choice is already enforced at dispatch (`content`→`entry` remap), so the schema combinator was redundant. The added regression test guarding all tools against top-level combinators is a nice touch.

Is there anything blocking the merge here, or a timeline? There are currently three duplicate issues (#1711, #1728, #1773) and multiple identical PRs, so a lot of people are hitting this. Happy to help if a rebase or extra test is needed. Thanks for maintaining mempalace!

Hexecu:
Hi @igorls 👋 — just a gentle nudge on this one. Three users have independently confirmed the bug and verified the fix works. There are also duplicate issues (#1711, #1728, #1773) pointing to the same root cause, so it seems like a decent number of people are hitting this on v3.4.0. Happy to rebase or adjust anything if needed. Thanks for maintaining mempalace!

PR #1719, opened 2026-06-07
docs(openclaw): catch up SKILL.md with 8 newer MCP tools

raman325:
## Summary

The OpenClaw memory skill (`integrations/openclaw/SKILL.md`) was last expanded against the mempalace MCP surface that existed when it was originally introduced (in PR #491, via @wanikua's earlier PR #207). At that point mempalace exposed 19 agent-facing `tool_*` functions. Since then 13 additional MCP tools have landed, and the skill has been out of step — agents loading the skill see the older subset and have to fall back to `npx mcporter call ...` (or just miss the tool entirely) for anything not listed.

This PR documents the 8 agent-facing tools the skill was missing, bringing it to 27 of mempalace's 30 documentable MCP tools.

### Tools added

**Search & Browse**
- `mempalace_list_drawers` — paginated drawer listing with optional wing/room filters
- `mempalace_get_drawer` — fetch a single drawer by ID (full verbatim content + metadata)

**Palace Graph**
- `mempalace_create_tunnel` — create an EXPLICIT cross-wing tunnel
- `mempalace_list_tunnels` — list all explicit tunnels
- `mempalace_delete_tunnel` — remove an explicit tunnel
- `mempalace_follow_tunnels` — walk explicit tunnels from a room

**Write / Session**
- `mempalace_update_drawer` — mutate drawer content and/or relocate to a different wing/room
- `mempalace_memories_filed_away` — acknowledge the silent auto-save checkpoint

### Tools intentionally NOT added

`mempalace_sync`, `mempalace_hook_settings`, and `mempalace_reconnect` are host/admin operations — not memory operations an agent should be calling on its own. The Hermes MemoryProvider plugin landing in MemPalace/mempalace#1684 makes the same exclusion call.

### Version

Bumped `3.3.0` → `3.4.0` — additive tool surface, no breaking changes to any existing tool docs in the file.

## Test plan

- [ ] Skill loads in OpenClaw (frontmatter still parses)
- [ ] Agent can call `mempalace_update_drawer` natively without `npx mcporter call` fallback
- [ ] Agent can call `mempalace_create_tunnel` / `mempalace_follow_tunnels` natively
- [ ] No regression in existing tool descriptions (only additions + one one-word clarification on `find_tunnels`)

### DIFF
--- a/integrations/openclaw/SKILL.md
+++ b/integrations/openclaw/SKILL.md
@@ -87,9 +87,9 @@ You have access to a local memory palace via MCP tools. The palace stores verbat
 - `mempalace_traverse` — Walk from a room, find connected ideas across wings
   - `start_room` (required): room to start from
   - `max_hops`: connection depth (default 2)
-- `mempalace_find_tunnels` — Find rooms that bridge two wings (implicit overlap)
-  - `wing_a`, `wing_b` (required)
-- `mempalace_create_tunnel` — Create an EXPLICIT cross-wing tunnel between two locations. Use when you notice content in one project relates to another (e.g. API design in `project_api` connects to schema in `project_database`).
+- `mempalace_find_tunnels` — Find rooms that bridge two wings via *implicit* overlap (rooms whose drawers naturally share content across wings — discovered, not declared)
+  - `wing_a`, `wing_b`: optional filters; omit both to scan all wing pairs
+- `mempalace_create_tunnel` — Create an *explicit* cross-wing tunnel: a user/agent-declared link between two locations. Use when you notice content in one project relates to another (e.g. API design in `project_api` connects to schema in `project_database`).
   - `source_wing`, `source_room`, `target_wing`, `target_room` (required)
   - `label`: short description of the relationship
   - `source_drawer_id`, `target_drawer_id`: anchor to specific drawers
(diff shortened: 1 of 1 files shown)


raman325:
Thanks for the reviews. Round 1 response (commit ca38ae8):

**Addressed:**

- **@gemini-code-assist — `find_tunnels` params are optional, not required.** Verified against `mempalace/mcp_server.py:1277` (`def tool_find_tunnels(wing_a: str = None, wing_b: str = None)`). Updated to `wing_a`, `wing_b: optional filters; omit both to scan all wing pairs`. This was a pre-existing doc bug that's been wrong since the file was introduced — happy to catch it here.
- **@Copilot — implicit vs. EXPLICIT casing/jargon.** Lowercased "EXPLICIT", italicized *implicit*/*explicit*, and added inline definitions: implicit = "rooms whose drawers naturally share content across wings — discovered, not declared"; explicit = "a user/agent-declared link between two locations."
- **@Copilot — `memories_filed_away` line too dense.** Split into description + `Returns:` + `When to call:` sub-bullets, matching the structure agents scan when picking tools.

**Not changing:**

- **@Copilot — annotate every param as `(required)` or `(optional)`.** The file's existing convention (in the 8 unmodified tool entries) is to mark `(required)` explicitly and leave optional params unmarked, with defaults shown as `(default X)`. See `mempalace_search`, `mempalace_check_duplicate`, `mempalace_kg_query`, etc. Adding `(optional)` everywhere would split the file into two conventions; I'd rather match what's already there. If the project wants the inverse convention, that's a separate sweep across all 27 entries.

Issue #1722, opened 2026-06-07
Bug: Benchmark/test data loaded into user wing pollutes semantic search


FMembodiment:
# Bug: Benchmark/test data loaded into user wing pollutes semantic search

## Summary

During MemPalace setup or development testing, approximately 55,000–70,000 benchmark and test dataset entries (ShareGPT, JSONL eval files, language config files) were loaded into the same wing/room namespace as real user data. This caused all semantic searches to return irrelevant results from these test datasets instead of actual user memories.

## Environment

- MemPalace version: (please check)
- OS: Windows 10
- Storage backend: ChromaDB (`chroma.sqlite3`)
- Palace path: `C:\Users\<user>\.mempalace\palace`

## What happened

After setup, semantic searches consistently returned irrelevant content from ShareGPT conversations, benchmark eval data, and language JSON files — even for very specific queries about personal data.

Example: searching for `"GG Sauna Dachdecker Kunde Webdesign"` (specific client names and business context) returned results from yoga retreat conversations and coding job descriptions from ShareGPT training data.

**Scope of contamination:**
- Total embeddings before cleanup: **70,466**
- Noise embeddings: **55,375** (78.6%)
- Real user data: **15,091** (21.4%)

**Noise source files identified:**
- `results_mempal_hybrid_v4_llmrerank_session_*.jsonl`
- `results_mempal_hybrid_v4_held_out_session_*.jsonl`
- `results_mempal_raw_session_*.jsonl`
- `id.json`, `de.json`, `en.json`, `fr.json`, `es.json`, `it.json`, `ru.json`, `tr.json`, `ja.json`, `ko.json` (language config files)

All noise was stored in the same `wing: FMEmbodiment, room: general` namespace as real user data.

## Impact

- Semantic search was effectively broken — similarity scores for real user data were indistinguishable from noise (~0.38–0.50 cosine distance)
- The AI assistant could not reliably retrieve client data, business context, or personal memories
- Required manual database surgery to fix (direct SQLite3 access + batch DELETE)

## Workaround applied

Direct ChromaDB SQLite cleanup script:
```python
conn = sqlite3.connect("~/.mempalace/palace/chroma.sqlite3")
cursor = conn.cursor()

NOISE_PATTERNS = ["results_mempal_", ".jsonl", "id.json", "de.json", ...]

noise_ids = set()
for pattern in NOISE_PATTERNS:
    cursor.execute(
[...]

Issue #1662, opened 2026-05-30
Feature request: add mempalace_mine tool to MCP server


w1tc4:
## Problem

Mining is currently only available as a CLI command (`mempalace mine`). The MCP server does not expose a mine/index tool, which means any MCP client (Desktop Commander, LM Studio, Aionui, etc.) has no way to trigger indexing from within a conversation.

Users who want to mine new files or re-index their palace from a non-Claude-Code environment have no option — they have to switch to a terminal or to Claude Code specifically.

## Requested tool

Add a `mempalace_mine` MCP tool that wraps the CLI mine functionality:

```json
{
  "name": "mempalace_mine",
  "description": "Index files into the palace from a source directory",
  "parameters": {
    "source": "Path to directory to mine (e.g. Agent_Logs or My_Memory)",
    "wing": "Wing to file drawers under",
    "mode": "Optional: 'convos' for conversation files",
    "agent": "Optional: agent name for conversation mode"
  }
}
```

## Why this matters

Claude Code has a plugin system with slash commands (`/mempalace mine`) and hooks that auto-mine on session end. Other MCP clients have no equivalent — the only way to get conversations into the palace from Desktop Commander, Aionui, or LM Studio is to manually run a terminal command. A `mempalace_mine` MCP tool would make the palace first-class everywhere, not just in Claude Code.

mvalentsev:
Open PR for this: #1692 (earlier attempt: #1509).

PR #1729, opened 2026-06-08
feat(mcp): add mempalace_delete_by_source bulk-cleanup tool (#1722)

arnoldwender:
## What and Why

#1722 reports benchmark/eval data (ShareGPT dumps, `results_mempal_*.jsonl`, language config JSON) getting mined into the same wing as real user data — in the reporter's palace, **55k of 70k embeddings (78%) were noise**, making semantic search return yoga-retreat conversations for queries about real clients. The only recourse today is hand-rolling `DELETE` against `chroma.sqlite3`.

This adds the **immediate** ask from that issue: an MCP tool to bulk-remove drawers by their `source_file`. (Requests #2 ingestion-time validation and #3 namespace isolation are deliberately left out — separate design calls.)

## Change Summary

- `mcp_server.py` — new `tool_delete_by_source(source_file, dry_run=True)` + `mempalace_delete_by_source` TOOLS entry + header doc line.
  - Matching is exact on `source_file` and pushed to the backend via `delete(where={"source_file": ...})` — the same idiom `miner.py`, `diary_ingest.py`, `format_miner.py` and `palace.py` already use. **No client-side id list, so the SQLite "too many variables" limit cannot be hit** regardless of how many drawers share the source.
  - Dry-run counting uses the existing `_fetch_all_metadata` paginator, so it stays accurate past the 10k `get()` truncation (the reporter had 55k matches).
  - Defaults to a dry run: returns `match_count` + a distinct `(wing, room)` sample so the caller can confirm blast radius before committing. `dry_run=false` commits (irreversible). Absent source is an idempotent no-op, not an error. Commit path writes a WAL audit entry and invalidates the metadata cache.
- `tests/test_mcp_server.py` — `TestDeleteBySource`: dry-run-doesn't-delete, commit-deletes-only-matching-source, idempotent-no-match, empty-source-rejected, registered-and-dispatchable (defaults to dry-run through `handle_request`).
- `website/reference/mcp-tools.md` — reference section (keeps `test_readme_claims` doc-sync green).

## Test Plan

- [x] `ruff check .` + `ruff format --check` clean (ruff 0.15.14, matching CI)
- [x] `pytest tests/ --ignore=tests/benchmarks` — **2435 passed, 5 skipped**
- [x] New `TestDeleteBySource` — 5 passed
- [x] Manually traced: dry-run on a seeded palace reports the right count and leaves all drawers; commit removes only the matching source

Related to #1722

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -15,6 +15,7 @@ Tools (read):
 Tools (write):
   mempalace_add_drawer      — file verbatim content into a wing/room
   mempalace_delete_drawer   — remove a drawer by ID
+  mempalace_delete_by_source — bulk-remove all drawers mined from one source_file
 
 Tools (maintenance):
   mempalace_reconnect       — force cache invalidation and reconnect after external writes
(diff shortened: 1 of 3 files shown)


arnoldwender:
Thanks for the review @gemini-code-assist — both suggestions addressed in `402ee73`:

1. **`source_file` normalization + type guard** (`mcp_server.py`): good catch on the consistency gap. `tool_add_drawer` strips lone surrogates from `source_file` before storing, so exact matching would silently miss rows mined from non-ASCII paths arriving via a cp1252 stdin (#1488). The handler now applies `strip_lone_surrogates` to the input before building the `where` filter, and guards with `isinstance(source_file, str)` so a non-string returns the clean error instead of `AttributeError`. Added a test that adds a drawer with a lone-surrogate path and confirms deletion still matches, plus a non-string-rejection test.
2. **Default `wing`/`room` to `""` in the sample** — applied, consistent with the rest of the file.

CI re-running on `402ee73`. Ready for re-review.

arnoldwender:
Rebased on latest `develop` (post-3.4.1, +84 commits). Conflict resolution only — no logic changes to the tool:

- `mcp_server.py`: the new `tool_mine`/`_capture_fd_stdout` from #1662 landed in the same slot between `tool_delete_drawer` and `tool_sync`; kept both functions and both `TOOLS` dict entries (`mempalace_mine` + `mempalace_delete_by_source`).
- `website/reference/mcp-tools.md`: kept both reference sections; intro count is now `34` (33 after #1662 merged, +1 for this tool). `test_readme_claims` parity verified — 34 dict keys / 34 doc sections, no orphans.

Full suite green locally on `.venv` (2861 passed, 5 skipped); `ruff check`/`format` clean. Ready for re-review.

PR #1743, opened 2026-06-09
fix(codex): make marketplace metadata installable

pussikill:
## Summary

Fix the Codex marketplace metadata so the MemPalace plugin can be installed from a marketplace source with the current Codex plugin ingestion schema.

## Root cause

The repository root already contains `.codex-plugin/plugin.json`, so the marketplace entry should point at the repository/plugin root. It currently points at `./.codex-plugin`, which makes Codex treat `.codex-plugin` itself as the plugin root and then fail to find a valid manifest at the expected location.

The marketplace entry also uses `policy.authentication: "NONE"`, which is not accepted by current Codex marketplace parsing. Codex expects `ON_INSTALL` or `ON_USE`.

Finally, the Codex plugin manifest currently declares MCP servers inline. Current Codex plugin validation expects `mcpServers` to resolve to a `.mcp.json` file.

## Changes

- Point `.agents/plugins/marketplace.json` at `./` instead of `./.codex-plugin`.
- Change marketplace authentication policy from `NONE` to `ON_INSTALL`.
- Move the MCP server declaration into root `.mcp.json` and reference it from `.codex-plugin/plugin.json`.
- Remove the unsupported `hooks` manifest field so the plugin passes current Codex validation. The hook files remain in place for hook loader support/convention-based handling.
- Preserve the Codex hook wrapper and mark it executable.

## Validation

Ran JSON validation and the current Codex plugin validator against the repository root:

```bash
jq . .agents/plugins/marketplace.json >/dev/null
jq . .codex-plugin/plugin.json >/dev/null
jq . .mcp.json >/dev/null
python3 /home/xdev/.codex/skills/.system/plugin-creator/scripts/validate_plugin.py .
```

Result:

```text
Plugin validation passed: /tmp/mempalace-codex-pr
```

## Notes

This intentionally does not duplicate the plugin bundle under `plugins/mempalace/`. Since this repository already has `.codex-plugin/plugin.json` at its root, `source.path: "./"` keeps a single source of truth for the plugin files.


pussikill:
Closing this draft PR; opened prematurely while preparing a fork-local fix.

Issue #1714, opened 2026-06-06
All write tools fail (diary_write / add_drawer / etc.) — ChromaDB compaction error: Rust type u64 (as SQL type INTEGER) is not compatible with SQL type BLOB


wiroon:
Hi! On my Mac (macOS 26.6, Python 3.13, mempalace plugin 3.0.14), every write-side MCP tool — mempalace_diary_write, mempalace_add_drawer, the whole lot — fails on a palace I've been mining into for the last couple of months. 

Read tools (mempalace_status, search, list_*) all work fine. On mempalace==3.1.0 the writes just returned "Connection closed" with no clue why. After upgrading to mempalace==3.4.0 (chromadb 1.5.9) I now get a real error: Error in compaction: … mismatched types; Rust type 'u64' (as SQL type INTEGER) is not compatible with SQL type BLOB — looks like ChromaDB metadata where column 0 is being read as u64 but stored as BLOB, classic version/schema drift.

mempalace repair (default legacy mode) bails out with exactly the same error before it can do anything. The palace is 5 GB / 696,471 drawers / 3.97 GB chroma.sqlite3, so I'm a bit hesitant to nuke it with repair --mode from-sqlite --archive-existing blind. 
Two questions: 
(1) is this a known regression and is from-sqlite rebuild the recommended path, or is there a less-destructive fix? 
(2) Could the MCP server surface the inner ChromaDB error to the client instead of silently dropping the stdio pipe (like 3.1.0 did)? 

That alone would have saved me a few hours thinking it was a connection / network problem. Happy to attach a --dry-run log or sha256 of the corrupt sqlite if helpful — read side of the tool has been doing great work, so I'd love to get the write side back. 🙏

LocallyInsaneDB:
Confirming this on **Windows 11 / WSL2 (Ubuntu), mempalace 3.4.0**, behind a corporate TLS-inspection proxy.

Same fingerprint: `mempalace_add_drawer` (and `diary_write`) failed with a bare **`MCP error: Connection closed`** — the inner ChromaDB error was swallowed and the stdio pipe dropped, exactly as you describe for 3.1.0. Read tools (`search`, `status`, `repair-status`) worked.

**`repair --mode from-sqlite --archive-existing` worked here.** It archived the corrupt palace to `palace.pre-rebuild-<ts>` and rebuilt cleanly:

```
[mempalace_drawers] done: 1277 rows
[...]

PR #1745, opened 2026-06-09
fix(mcp): surface Chroma compaction write failures with repair guidance (#1714)

rodboev:
## Summary

Write-side MCP tools could fail as an opaque disconnect when ChromaDB hit a compaction/corruption error, leaving users with “Connection closed” instead of the underlying storage problem. This PR catches that write failure shape and surfaces the inner Chroma error with actionable repair guidance.

## Fix

- `mempalace/mcp_server.py`: detect the Chroma compaction/corruption failure shape on MCP write tools, preserve the inner error text, and return a structured application error that points users at `mempalace repair --mode from-sqlite --archive-existing`.
- `tests/test_mcp_server.py`: add focused coverage for `mempalace_add_drawer`, `mempalace_delete_drawer`, `mempalace_update_drawer`, and `mempalace_diary_write` so write failures surface as structured tool errors instead of opaque disconnects.

## Why It Matters

Users can distinguish a real storage corruption problem from a network or client issue and are pointed at the existing repair workflow immediately. This saves debugging time without pretending the underlying Chroma corruption has been fixed in-process.

## Test plan

- [x] `D:\Repos\mempalace\.venv\Scripts\python.exe -m pytest tests/test_mcp_server.py -v`
- [x] `D:\Repos\mempalace\.venv\Scripts\python.exe -m ruff check .`
- [x] `D:\Repos\mempalace\.venv\Scripts\python.exe -m ruff format --check .`
- [x] `D:\Repos\mempalace\.venv\Scripts\python.exe -m pytest tests/ -v`

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -874,6 +874,32 @@ def _safe_meta(meta):
     return meta if isinstance(meta, dict) else {}
 
 
+def _is_chroma_compaction_corruption(exc_msg: str) -> bool:
+    msg = exc_msg.lower()
+    return (
+        "not compatible" in msg
+        and "sql type blob" in msg
+        and ("mismatched types" in msg or "rust type" in msg)
+    )
+
+
+def _write_error_payload(exc: BaseException) -> dict:
+    message = str(exc)
+    payload = {
+        "success": False,
+        "error": message,
+        "error_class": type(exc).__name__,
+    }
+    if _is_chroma_compaction_corruption(message):
+        payload["repair_guidance"] = (
+            "Detected Chroma compaction corruption. "
+            "Run `mempalace repair --mode from-sqlite --archive-existing` "
+            "to rebuild from chroma.sqlite3."
+        )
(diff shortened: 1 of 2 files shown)


mvalentsev:
The four write tools already end in `except Exception: return {"success": False, "error": str(e)}` on this PR's own base, so the inner ChromaDB text was already surfacing as a tool error; the delta is the repair hint plus two pre-check spots. What this does not touch is the reported symptom itself: the second report in #1714 (2026-06-09, mempalace 3.4.0) confirms a bare "Connection closed" with the stdio pipe dropped, and a handler-level except is not what fixes a dropped pipe. The summary describes fixing an opaque disconnect; the diff adds guidance text around errors that were not opaque.

The tests say the same: all four use an invented error string ("Rust type 'u64' and corresponding SQL type BLOB are not compatible") that shares three substrings with the real message from the issue ("Rust type 'u64' (as SQL type INTEGER) is not compatible with SQL type BLOB") but never matches it verbatim, so the matcher goes green without ever seeing the text it was written for. Written from the issue text, not from a reproduced failure.

rodboev:

Issue #1665, opened 2026-05-30
Wing-filtered `search` raises `Error finding id` when HNSW drifts from sqlite after bulk delete


marcuss:
## Summary

After a bulk delete operation via direct SQL on `chroma.sqlite3` (followed by `mempalace repair --yes`), some IDs persist in the HNSW vector index that no longer exist in the `embeddings` table. Subsequent `mempalace search` calls that apply a `where=` filter (e.g. `--wing nextasy`) raise ChromaDB's `Error finding id` from the query path. Unfiltered queries succeed because they're satisfied before the HNSW walk reaches stale entries.

## Reproduction (real-world steps that hit it)

A maintenance pass purged ~370k low-value drawers and rebuilt with `mempalace repair --yes`:

1. Identify drawer IDs to delete via SQL:
   ```sql
   SELECT id FROM embedding_metadata WHERE key='wing' AND string_value LIKE 'sessions';
   ```
2. Direct `DELETE` from `embeddings`, `embedding_metadata`, `embedding_fulltext_search` (etc.) — ~370k rows, multiple wings.
3. `VACUUM` the sqlite. Confirm row counts match expected survivors (~2,256).
4. `mempalace repair --yes` — output: `Re-filed 2256/2256 drawers`. Looks fine.
5. Later, additional drawers added via the `PreCompact` hook + curated extractor — `embeddings` row count grows to 2,311.
6. **Now** any query with `where={"wing": "<some-wing>"}` filter raises:
   ```
   Error executing plan: Internal error: Error finding id
   ```
   Unfiltered queries return correctly.

Verification of the drift after step 5:
```
sqlite3 ~/.mempalace/palace/chroma.sqlite3 "SELECT COUNT(*) FROM embeddings;"
# → 2311
mempalace repair --yes
# → Re-filed 2256/2256 drawers
# → 2311 - 2256 = 55 orphan ids in HNSW with no metadata/document rows
```

Re-running `mempalace repair --yes` a second time consolidates HNSW back to 2256 entries and the wing-filtered query succeeds.

## Why the wing filter exposes it (but unfiltered queries don't)

[...]

PR #1750, opened 2026-06-10
fix(searcher): use filter-fallback in CLI search path for HNSW drift (#1665)

rodboev:
## Summary

`search()` (CLI path) and `search_memories()` (MCP/API path) diverged in their handling of filtered HNSW failures. `search_memories` already routes through `_query_drawers_with_filter_fallback` (added for #1245/#1035), which retries unfiltered and post-filters in Python when ChromaDB raises `Error finding id` on a `where=`-scoped query. `search()` did not — its `except Exception` block at `searcher.py:381` caught the error and re-raised as `SearchError`, so wing-filtered CLI searches failed with a cryptic message while the equivalent MCP call recovered cleanly.

The fix routes `search()` through the same helper, making CLI and MCP behavior consistent. No new infrastructure: the helper already handles the unfiltered-retry, Python post-filter, and warning log.

## Fix

Replace the inline `col.query(**kwargs)` call in `search()` (`searcher.py:379`) with `_query_drawers_with_filter_fallback(col, kwargs, query, n_results, wing, room)`. The helper signature accepts the same kwargs dict and the existing wing/room locals, so no refactoring is needed around the call site.

Add a regression in `tests/test_searcher.py`: mock a collection whose `query` raises only when `where=` is present, call `search()` with a wing filter, assert no `SearchError` is raised and results are returned.

## Related work

Open PRs #951, #1005, #1463, and #1598 address overlapping search-fallback surfaces. This PR is the narrowest fix: it routes the CLI `search()` path through the existing `_query_drawers_with_filter_fallback` helper that `search_memories()` already uses. #1005 and #1598 both subsume this change as part of broader scope; if either lands first, this PR is redundant.

## Test plan
- [x] `python -m pytest tests/test_searcher.py -v` — 31 passed
- [x] `ruff check .` — clean
- [x] `ruff format --check .` — clean

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -376,7 +376,7 @@ def search(query: str, palace_path: str, wing: str = None, room: str = None, n_r
         if where:
             kwargs["where"] = where
 
-        results = col.query(**kwargs)
+        results = _query_drawers_with_filter_fallback(col, kwargs, query, n_results, wing, room)
 
     except Exception as e:
         print(f"\n  Search error: {e}")
--- a/tests/test_searcher.py
+++ b/tests/test_searcher.py
@@ -413,3 +413,45 @@ class TestSearchCLI:
         captured = capsys.readouterr()
         assert "[1]" in captured.out
         assert "[2]" in captured.out
+
+    def test_search_wing_filter_fallback_on_hnsw_drift(self, fake_palace_path, capsys):
+        """search() gracefully falls back to unfiltered + post-filter when
+        a filtered query fails due to HNSW/SQLite mismatch (orphan IDs).
+
+        Simulates: a filtered query raises "Error finding id", but an
+        unfiltered query succeeds. The fallback retries unfiltered,
+        post-filters in Python by wing, and returns results without
+        raising SearchError.
+        """
+        mock_col = MagicMock()
+        mock_col.metadata = {"hnsw:space": "cosine"}
+
+        # First call (filtered): raises, simulating orphan ID drift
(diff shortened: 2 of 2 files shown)


mvalentsev:
The fix works, but check the #1665 thread before review: four open PRs already cover this same failure (#1005 delegates CLI search to `search_memories` and removes the CLI/API split entirely, #1598 hardens CLI and API together, #951 carries a `get()`-based variant, #1463 a BM25 fallback). This makes a seventh remedy on the cluster, and the description mentions none of them. Choosing between copying the fallback into the second path and collapsing the two paths is a maintainer call.

rodboev:
Fair point. I should have acknowledged those PRs in the description. Updated it to reference #951, #1005, #1463, and #1598 and note the overlap. This one is the narrowest surgical option for the CLI-only gap; if any of the broader PRs land first it's redundant.

mvalentsev:
@rodboev Just one request, please verify problem/fix by actually running MemPalace, rather than reviewing the code theoretically.

rodboev:
Ran it locally against a small test palace. CLI search works fine with and without wing filters on a healthy index. The tricky part is reproducing the actual drift scenario from #1665, since it requires orphaned HNSW IDs left behind after a bulk delete. The fix itself is straightforward routing: `search()` now goes through the same `_query_drawers_with_filter_fallback` helper that `search_memories()` already uses, so the fallback logic is shared instead of duplicated. I can set up a targeted repro if you'd like, but it would need a palace large enough for the bulk-delete path to leave stale IDs in the index.

rodboev:

Issue #1380, opened 2026-05-06
Hash embedding mode should use lexical/BM25 search fallback


vanachterjacob:
## Problem
When `embedding_device=hash` / lexical fallback is configured, search can return low-signal vector hits instead of obvious lexical matches.

Observed query:

```text
mempalace reindex database hnsw embeddings
```

The MCP/vector path returned unrelated entries such as stopword/i18n files and NUL-byte-like documents, while SQLite FTS/BM25 found relevant HNSW/reindex rows immediately.

## Root Cause
Hash embeddings are a local lexical fallback, but the search path still treats them like dense semantic embeddings. That makes vector distance a weak ranking signal for some real queries. The SQLite FTS fallback also selected candidates without `ORDER BY rank`, so low-rowid partial matches could crowd out better candidates before BM25 reranking.

## Expected Behavior
When hash/lexical embedding mode is active, search should use the SQLite BM25 path intentionally. FTS candidate selection should retrieve relevant candidates first before local BM25 reranking.

## Proposed Fix
- Route MCP and CLI search to SQLite BM25 when `embedding_device=hash` or `lexical`
- Add `ORDER BY rank` to the SQLite FTS candidate query
- Preserve the existing vector path for normal ONNX/dense embedding mode
- Surface the fallback reason in structured search results

vanachterjacob:
Draft fix proposed in #1381: https://github.com/MemPalace/mempalace/pull/1381

jphein:
@vanachterjacob — adjacent prior art on the structural fallback piece, in case useful: [#1005](https://github.com/MemPalace/mempalace/pull/1005) (open, MERGEABLE, all 6 CI green) implements the *machinery* you're asking for in a more general shape, then leaves the trigger conservative. Specifically it adds:

- `_sqlite_fallback_and_scope` — ranks the sqlite candidate pool with the same Okapi-BM25 used in `_hybrid_rank` and tops up the hits list when the vector path underdelivers (sparse HNSW, filter-planner failure, async-flush drift).
- `available_in_scope` — sqlite-authoritative scope count via paginated `col.get(include=[], ...)`, so a query into a wing with N drawers can show that N even when vector returned 0.
- `warnings: [...]` in the response shape — exactly the "surface the fallback reason in structured search results" you're asking for. Each fallback emits a structured warning describing what triggered (vector_underdelivered, available > vector_hit_count, filter-planner failure).
[...]

PR #1751, opened 2026-06-10
fix(searcher): route hash/lexical embedding mode to SQLite BM25 (#1380)

rodboev:
## Summary

When `embedding_device=hash` or `embedding_device=lexical` is configured, `search_memories` was querying the HNSW index with a hash-derived (non-semantic) vector, returning low-signal hits — stopword files, NUL-byte documents — while SQLite FTS/BM25 immediately surfaced relevant results for the same query. The SQLite BM25 path (`_bm25_only_via_sqlite`) already existed for the HNSW-diverged case (#1222); hash/lexical mode just wasn't routing to it.

Two changes: (1) detect `embedding_device in {"hash", "lexical"}` in `search_memories` and route to `_bm25_only_via_sqlite` with `fallback_reason="embedding_device=<device>"`, and (2) add `ORDER BY rank` to the FTS5 candidate SELECT so better BM25 candidates reach the local reranker first instead of low-rowid partial matches crowding them out.

## Fix

Added `_is_lexical_device()` helper at `searcher.py:44` and a routing block in `search_memories` after strategy validation that reads `MempalaceConfig().embedding_device`, calls `_bm25_only_via_sqlite`, and sets the structured `fallback_reason` key. The hash/lexical check only runs when `vector_disabled` is not already set, so the existing HNSW-diverged code path is unchanged.

Added `ORDER BY embedding_fulltext_search.rank` to the FTS5 candidate SELECT at `searcher.py:514`, which uses the built-in FTS5 `rank` column (BM25-style relevance) to pre-sort candidates before the Python Okapi-BM25 reranker.

## Test plan
- [x] `python -m pytest tests/test_searcher.py -v` — 39 passed (11 new)
- [x] `ruff check .` — clean
- [x] `ruff format --check .` — clean

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -43,6 +43,18 @@ class SearchError(Exception):
 
 _TOKEN_RE = re.compile(r"\w{2,}", re.UNICODE)
 
+_LEXICAL_DEVICES = frozenset({"hash", "lexical"})
+
+
+def _is_lexical_device(device: str) -> bool:
+    """Return True when the configured embedding device is hash/lexical mode.
+
+    Hash and lexical devices produce non-semantic vectors; routing to the
+    HNSW index with those vectors returns meaningless hits. Use SQLite
+    BM25 instead.
+    """
+    return (device or "").strip().lower() in _LEXICAL_DEVICES
+
 
 def _first_or_empty(results, key: str) -> list:
     """Return the first inner list of a query result field, or [].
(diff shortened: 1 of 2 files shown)


mvalentsev:
`embedding_device=hash` does not exist on develop: it is not in `_PROVIDER_MAP`, so `_resolve_providers` logs an unknown-device warning and returns the CPU provider (`embedding.py:86`), which means real embeddings and working vector search. The device only exists inside draft PR #1381, where vanachterjacob (also the reporter of #1380) implemented it five weeks ago together with the BM25 routing for both the MCP and CLI paths, the `ORDER BY rank` change, and the `fallback_reason` plumbing. This PR reimplements part of that, without the device itself, and does not mention it.

Net effect on develop: a user with `embedding_device: hash` in config.json gets working vector search today (cpu fallback), and with this PR the same config silently downgrades the MCP search path to BM25-only. The `ORDER BY rank` line is the one piece that does something useful here, and it is already in #1381.

rodboev:

PR #1762, opened 2026-06-10
feat: add append param to update_drawer

### LINKED ISSUE #1761: Add `append` option to update_drawer (avoid re-sending the full drawer body on incremental updates)
## Summary

`update_drawer` always **replaces** a drawer's entire body. When a tool
maintains a long-lived drawer and updates it incrementally — appending a new
finding, flipping a status line, adding a log entry — it has to re-send the
**whole** body on every edit just to tack on a few lines.

For an LLM agent driving the MCP tool, that whole-body round-trip is paid out
of the agent's context window on *every* incremental update, which gets
expensive fast for drawers that are several KB and updated many times over a
session.

## Current behavior

`tool_update_drawer(drawer_id, content=..., wing=..., room=...)` sets
`new_doc = sanitize_content(content)` — a full replace. There is no
append/patch affordance, so an incremental update must reconstruct and
re-transmit the complete prior body plus the delta.

## Proposed direction

One minimal approach: add an `append: bool = False` parameter. When
`append=True` (and `content` is given), concatenate onto the existing body
instead of replacing:

```python
new_doc = old_doc + sanitized if append else sanitized
```

The handler **already fetches `old_doc`** (`col.get(..., include=["documents", ...])`),
so this is a small read-modify-write — no extra round-trip — plus one schema
property. `append` defaults to `False`, so existing callers are unaffected.

### Alternatives (maintainer's call)

I'm not attached to the exact surface — a couple of other shapes might fit the
project better:

- A separate `append_to_drawer` / `mempalace_append_to_drawer` tool, keeping
  `update_drawer` replace-only.
- A richer **section patch** that can target and replace a named section
  (e.g. update just a "status" block) rather than only tail-append.

Tail-append is the smallest change that removes the re-transmit cost; the
section-patch is more powerful but bigger. Happy to follow whichever the
maintainers prefer.

## Context

- Affects any client that maintains living/append-mostly drawers via the MCP
  `update_drawer` tool.
- A PR implementing the minimal `append` form (with tests) is drafted and
  linked below, but treat it as a starting point for the API discussion rather
  than a fixed proposal.

_(PR drafted — link below.)_

fkberthold:
## Summary

Adds an `append: bool = False` parameter to `tool_update_drawer` (and its MCP
schema entry). When `append=True` and `content` is provided, the content is
**concatenated onto the existing drawer body** instead of replacing it, so a
caller updating a long drawer can send only the delta rather than
re-transmitting the whole body on every edit.

`append` defaults to `False` — existing callers and the replace-in-full
behavior are unchanged. The handler already fetches the existing document, so
this is a minimal read-modify-write with no extra round-trip:

```python
new_doc = old_doc + sanitized if append else sanitized
```

Motivation and the alternatives considered (a separate `append_to_drawer`
tool, or a richer section-patch) are in the linked issue — this PR implements
the smallest form and is happy to follow a different surface if preferred.

Closes https://github.com/MemPalace/mempalace/issues/1761

## Changes

- `mempalace/mcp_server.py`
  - `tool_update_drawer(...)` gains `append: bool = False`; content path now
    appends-or-replaces based on the flag.
  - WAL log records the `append` flag.
  - `mempalace_update_drawer` MCP schema gains an `append` boolean property;
    tool description notes the delta-append behavior.
- `tests/test_mcp_server.py`
  - `test_update_drawer_append` — appends a delta and asserts the original body
    survives and the delta is concatenated (`old_doc + sanitize_content(delta)`).
  - `test_update_drawer_replace_is_default` — pins that the default
    (`append=False`) still replaces, guarding the back-compat path.

## Verification

Test-first (RED → GREEN; visible in the commit order on this branch):

- RED commit adds `test_update_drawer_append`, which fails against the current
  code with `TypeError: tool_update_drawer() got an unexpected keyword argument
  'append'`.
- GREEN commit adds the parameter; the test passes.

```
$ pytest tests/ -q
889 passed, 106 deselected, 1 warning in ~56s     # 1 warning is a pre-existing,
                                                   # unrelated fact_checker runpy warning
$ ruff check mempalace/mcp_server.py
All checks passed!
```

## CONTRIBUTING.md adherence

- Conventional commits (`test:` for the failing test, `feat:` for the param).
- `ruff` clean on the changed file (100-char limit).
- Branched from and targeted at `develop`; pushed to a personal fork.
- Tests run without API keys or network access.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1895,8 +1895,21 @@ def tool_list_drawers(wing: str = None, room: str = None, limit: int = 20, offse
         return {"error": str(e)}
 
 
-def tool_update_drawer(drawer_id: str, content: str = None, wing: str = None, room: str = None):
-    """Update an existing drawer's content and/or metadata."""
+def tool_update_drawer(
+    drawer_id: str,
+    content: str = None,
+    wing: str = None,
+    room: str = None,
+    append: bool = False,
+):
+    """Update an existing drawer's content and/or metadata.
+
+    When ``append`` is True and ``content`` is given, the content is
+    concatenated onto the existing drawer body instead of replacing it, so a
+    caller updating a long drawer need only send the delta rather than
+    re-transmitting the whole body on every edit. ``append`` defaults to
+    False, preserving the prior replace-in-full behavior.
+    """
     global _metadata_cache
 
     if content is None and wing is None and room is None:
@@ -1916,9 +1929,10 @@ def tool_update_drawer(drawer_id: str, content: str = None, wing: str = None, ro
         new_doc = old_doc
         if content is not None:
             try:
-                new_doc = sanitize_content(content)
+                sanitized = sanitize_content(content)
             except ValueError as e:
                 return {"success": False, "error": str(e)}
+            new_doc = old_doc + sanitized if append else sanitized
 
         new_meta = dict(old_meta)
         if wing is not None:
@@ -1949,6 +1963,7 @@ def tool_update_drawer(drawer_id: str, content: str = None, wing: str = None, ro
                 "new_wing": new_meta.get("wing", ""),
                 "new_room": new_meta.get("room", ""),
                 "content_changed": content is not None,
+                "append": append,
                 "content_preview": new_doc[:200] if content is not None else None,
             },
         )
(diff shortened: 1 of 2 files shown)


igorls:

Issue #1748, opened 2026-06-10
mempalace_status times out on large palaces under Claude Desktop — full-metadata pagination instead of the existing sqlite aggregation path


KeilerHirsch:
## Summary

On a large palace (~182k drawers), `mempalace_status` consistently times out
in Claude Desktop (hard 240s MCP tool-call limit). Root cause: `tool_status`
fetches the **entire collection metadata** through paginated
`col.get(limit=1000, offset=N)` calls just to compute wing/room counts.
The pure-sqlite aggregation reader that already ships for the #1222
fallback (`_tool_status_via_sqlite`) produces the same breakdown in ~0.3s.

## Environment

- mempalace 3.3.6, chromadb HttpClient mode (chroma server owns the palace)
- Palace: 182,580 drawers, 9 wings
- Windows 11, Python 3.13
- Hosts: Claude Desktop (fails, 240s tool timeout) and Claude Code (succeeds,
  but silently takes minutes)

## Measurements

Per-page latency of `col.get(include=["metadatas"], limit=1000, offset=N)`
against a local chroma server (sqlite OFFSET scan cost grows with depth):

| offset  | latency |
|---------|---------|
| 0       | 0.78s   |
| 20,000  | 0.67s   |
| 60,000  | 0.95s   |
| 120,000 | 1.48s   |
| 180,000 | 1.52s   |

→ 183 pages ≈ **3–4 minutes** per `mempalace_status` call (the 5s metadata
cache TTL means practically every call pays this).

Same aggregation via one read-only sqlite query
(`SELECT em.string_value, COUNT(*) ... GROUP BY em.string_value`):
**0.32s** — the code for this already exists as `_tool_status_via_sqlite`.

py-spy stack of the hanging server (captured mid-call):

```
read (httpcore\_backends\sync.py:128)
...
get (Collection.py:161)
_fetch_all_metadata (mempalace\mcp_server.py:624)
_get_cached_metadata (mempalace\mcp_server.py:648)
tool_status (mempalace\mcp_server.py:765)
handle_request (mempalace\mcp_server.py:2566)
```

## Why it matters

`mempalace_status` is the documented wake-up call of the memory protocol
("ON WAKE-UP: Call mempalace_status..."), so on Claude Desktop a large
palace is effectively unreachable: the first protocol step always dies at
the 240s ceiling. The failure is invisible on hosts with longer timeouts —
they just burn minutes per status call.

## Proposed fix

[...]

PR #1764, opened 2026-06-10
fix(status): route tool_status through sqlite aggregation to avoid 240s timeout on large palaces (#1748)

rodboev:
## Summary

`mempalace_status` times out on palaces with 182k+ drawers because `tool_status()` paginates all metadata through `col.get(limit=1000, offset=N)` — 183 pages at O(n*offset) cost per page, totaling 3-4 minutes. The pure-sqlite aggregation reader `_tool_status_via_sqlite()` already ships and produces the same wing/room breakdown in ~0.3s via `GROUP BY`, but was only used when `_vector_disabled` is set. This routes `tool_status` through the sqlite reader unconditionally.

## Fix

`tool_status()` now calls `_tool_status_via_sqlite()` for all palaces, stripping `vector_disabled`/`vector_disabled_reason` keys when vectors are healthy and setting `backend` from `_selected_backend_name()`. The sqlite reader uses a `mode=ro` connection and never touches the HNSW segment, so it is safe alongside both PersistentClient and HttpClient deployments.

`list_wings`, `list_rooms`, and `get_taxonomy` still use `_get_cached_metadata` — these are lower-frequency tools and can be migrated in a follow-up.

## Test plan
- [x] `python -m pytest tests/test_mcp_server.py -v -k status`
- [x] Verified: patched `tool_status` no longer calls `_fetch_all_metadata` or `_get_cached_metadata`

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1002,45 +1002,15 @@ def _tool_status_via_sqlite() -> dict:
 
 
 def tool_status():
-    # Run the safe sqlite/pickle probe before we touch chromadb. In the
-    # #1222 failure mode, opening the persistent client to call .count()
-    # can segfault — short-circuit to a pure-sqlite path when divergence
-    # is detected so status stays reachable.
-    db_exists = _backend_db_exists()
     _refresh_vector_disabled_flag()
 
-    if _vector_disabled:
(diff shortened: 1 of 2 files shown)


rodboev:

Issue #1716, opened 2026-06-06
Self-repair false positive: empty link_lists.bin (all-layer-0 HNSW) quarantined as corruption, self-perpetuating loop


JeanBaptisteRenard:
## Summary

`_hnsw_link_lists_is_usable_for_payload()` (`mempalace/backends/chroma.py:74-95`, v3.3.6) misclassifies a **legitimate 0-byte `link_lists.bin`** as structural corruption, causing a self-perpetuating quarantine loop on small/low-fanout HNSW indexes.

## Details

The check treats `data_level0.bin > 1024 bytes AND link_lists.bin == 0 bytes` as corruption. But a zero-byte `link_lists.bin` is a valid hnswlib serialization when **all materialized elements sit on layer 0** — hnswlib only serializes link lists for elements with level > 0, which is the normal shape for small indexes.

## Observed loop (production palace, mempalace 3.3.6 / chromadb 1.5.9)

1. Segment quarantined as `<uuid>.drift-<ts>` with `data_level0.bin` = 167,600 bytes, `link_lists.bin` = 0 bytes.
2. Self-repair rebuilds the segment → the rebuilt **live** segment has the byte-identical shape (167,600 / 0).
3. Next cold-start health check quarantines it again → loop. We accumulated **20 quarantine dirs / 221 MB** before noticing.

The trigger that reaches the broken check in our case is the mtime-stale path (sqlite mtime ≫ HNSW flush mtime, beyond the 300s threshold) falling through to the link-lists check — not the bloat-ratio guard.

Cadence amplifier: with a `docker exec` per-process MCP model, the once-per-palace-per-process `_quarantined_paths` gate resets on every process spawn, so the false positive re-fires per session, independent of any ingestion.

## Expected behavior

A 0-byte `link_lists.bin` alongside a plausible `data_level0.bin` should be considered healthy (or at minimum cross-checked against the index element levels / header) before quarantining.

## Suggested direction

Treat `link_lists.bin == 0` as corrupt only when the HNSW header indicates elements with level > 0 (or when `data_level0.bin` is inconsistent with the sqlite-side count). Happy to provide the full segment metadata if useful.

williamblair333:
Hit this on a small real-world corpus (a few thousand notes) — the quarantine loop reproduces, and the index never stabilizes across cold starts.

[...]

PR #1765, opened 2026-06-10
fix(backends): accept zero-byte link_lists.bin for small all-layer-0 HNSW indexes (#1716)

rodboev:
## Summary

Small HNSW indexes where all elements land on layer 0 legitimately produce a 0-byte `link_lists.bin` (hnswlib only serializes link data for elements with level > 0). The health check `_hnsw_link_lists_is_usable_for_payload()` unconditionally treated this as corruption, triggering a quarantine on every cold start. Since self-repair rebuilds the segment in the same shape, the quarantine re-fired each session, accumulating `.drift-*` directories in a self-perpetuating loop.

## Fix

Introduce `_HNSW_ZERO_LINK_LISTS_DATA_CEIL` (512 KB). When `data_level0.bin` is below this ceiling, a zero-byte `link_lists.bin` is accepted as a valid all-layer-0 graph rather than flagged as an interrupted persist. The 512 KB threshold covers ~200 elements at typical embedding dimensions (384-dim, M=16), well within the statistical envelope where all elements on layer 0 is expected. Above the threshold, the existing corruption detection is unchanged.

## Test plan
- [x] `python -m pytest tests/test_hnsw_payload_health.py -v`
- [x] Verified: small segment (data < 512 KB) with 0-byte link_lists is not quarantined
- [x] Verified: large segment (data > 512 KB) with 0-byte link_lists is still quarantined when stale

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -82,10 +82,14 @@ def _hnsw_link_to_data_ratio(seg_dir: str) -> Optional[float]:
 def _hnsw_link_lists_is_usable_for_payload(seg_dir: str) -> bool:
     """Return False when a non-trivial HNSW payload lacks usable link lists.
 
-    A missing or empty link_lists.bin is acceptable only for a fresh/empty
-    segment. Once data_level0.bin has real payload, a zero-byte link_lists.bin
-    is not a harmless async-flush shape: ChromaDB can later hand the broken
-    graph to hnswlib and crash in native code.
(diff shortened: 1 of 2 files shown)


mvalentsev:
@rodboev Why aren't you checking existing PRs before opening new ones? This isn't about this PR specifically, it applies to the whole batch (2 days). It looks like spam.

rodboev:
I checked PRs and didn't see any competing ones. Diving in again to see where the conflict is.

rodboev:

Issue #1712, opened 2026-06-06
HNSW link_lists.bin bloat: silent idempotency exception causes unbounded file growth


orthopal:
[hnsw_bloat_issue.zip](https://github.com/user-attachments/files/28664375/hnsw_bloat_issue.zip)

## What happened

After ~10k agent sessions, the HNSW index grew to **9.6 TB logical / 575 GB real** on disk (a sparse file), causing Python to SIGSEGV on the next load. The culprit is in `tool_add_drawer` in `mcp_server.py`.

## Root cause

The idempotency pre-check swallows all exceptions silently:

```python
try:
    existing = col.get(ids=idempotency_probe_ids, include=[])
    if existing.ids:
        return {"success": True, "reason": "already_exists", "drawer_id": drawer_id}
except Exception:
    logger.debug("Idempotency pre-check failed …", exc_info=True)
    # falls through to upsert() ← BUG
```

When the HNSW index is under stress (e.g. Rosetta 2 translation, high session count), `col.get()` raises. The exception is swallowed and execution falls through to `col.upsert()`. Each upsert on a partially-corrupt index inserts duplicate nodes via relative seeks in `link_lists.bin`, and the seek drift compounds across sessions — sparse file grows without bound.

## Fix

Return an error instead of falling through:

```python
except Exception as e:
    return {"success": False, "error": f"idempotency check failed: {e}"}
```

This is safe: the caller sees the failure and the HNSW index is never written to.

A patch and a standalone reproduction script are attached (`hnsw_bloat_issue.zip`). The repro uses random embeddings (no model/network needed) and shows the unguarded case growing while `_HNSW_BLOAT_GUARD` keeps it at zero.

## Note on `_HNSW_BLOAT_GUARD`

The existing guard (`hnsw:batch_size=50_000, hnsw:sync_threshold=50_000`) helps a lot by deferring flushes, but it doesn't protect against the exception-swallow path. Both fixes together close the loop.

## Environment

- macOS (Apple Silicon, Rosetta 2 context)  
- ChromaDB 1.5.x  
- mempalace 3.3.0

PR #1766, opened 2026-06-10
fix(mcp): return error on idempotency check failure instead of falling through to upsert (#1712)

rodboev:
## Summary

`tool_add_drawer()` swallowed exceptions from the idempotency pre-check (`col.get()`) and fell through to `col.upsert()`. When the HNSW index was under stress, each swallowed exception led to a duplicate write that compounded seek drift in `link_lists.bin`, producing unbounded sparse-file growth (reported: 9.6 TB logical / 575 GB real after ~10k sessions).

## Fix

The `except Exception` block at `mcp_server.py:1445` now returns `{"success": False, "error": "idempotency check failed: {e}"}` instead of silently continuing. Log level upgraded from `debug` to `warning`. The caller sees the failure and can retry; the HNSW index is never written to when the read path is unhealthy.

This is a two-line change. The existing `_HNSW_BLOAT_GUARD` (batch/sync thresholds) helps defer flushes but does not protect against the exception-swallow path; both fixes together close the loop.

## Test plan
- [x] `python -m pytest tests/test_mcp_server.py -v -k idempotency`
- [x] Verified: when `col.get()` raises, `tool_add_drawer` returns `success: False` without calling `col.upsert`
- [x] Verified: normal add and idempotency-hit paths are unchanged

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1462,8 +1462,13 @@ def tool_add_drawer(
         existing = col.get(ids=idempotency_probe_ids, include=[])
         if existing.ids:
             return {"success": True, "reason": "already_exists", "drawer_id": drawer_id}
-    except Exception:
-        logger.debug("Idempotency pre-check failed for %s", idempotency_probe_ids, exc_info=True)
+    except Exception as e:
+        logger.warning(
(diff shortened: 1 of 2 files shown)


rodboev:

Issue #1595, opened 2026-05-23
mempalace-health.py: header.bin parsed as uint32 — trillion-element HNSW corruption silently passes all checks


williamblair333:
## Summary

`mempalace-health.py` parses `header.bin` using `struct.unpack_from('<I', ...)` (unsigned 32-bit int). When the HNSW index is corrupted by the chromadb-hnswlib 1.5.x `updatePoint` type-confusion bug (chroma-core/chroma#4460), `cur_element_count` in the header is written as a garbage value in the trillions (observed: `7,198,365,188,096`). A uint32 read silently wraps this to `0`, so the health check passes with no warning.

## Observed behaviour

```
Vector search is offline. HNSW index has 1,056 entries; SQLite has 467,791 — 
466,735 drawers (100%) are invisible to semantic search. Only BM25 fallback is running.
```

Despite this, `mempalace-health.py` reported `OK` because:

```python
# Current code — WRONG
max_elements = struct.unpack_from("<I", hdr, 12)[0]   # uint32
cur_elements = struct.unpack_from("<I", hdr, 20)[0]   # uint32
# 7,198,365,188,096 & 0xFFFFFFFF == 0 → silently passes
```

The downstream check `if max_elements > 0 and cur_elements > max_elements` also never fires because both values read as `0`.

## Root cause

`header.bin` stores element counts as 64-bit integers (hnswlib header layout). Parsing with `<I` (32-bit) truncates the value. With the chromadb-hnswlib 1.5.x corruption, the 64-bit value is a float-interpreted-as-int64 (C++ type confusion), landing in the trillions — well above uint32 max.

## Fix

Parse as signed int64 and add a sanity cap:

```python
SANE_ELEMENT_CAP = 10_000_000

hdr = (seg_dir / "header.bin").read_bytes()
max_elements, cur_elements = struct.unpack_from("<qq", hdr, 8)[0], struct.unpack_from("<qq", hdr, 16)[0]
if max_elements > SANE_ELEMENT_CAP or cur_elements > SANE_ELEMENT_CAP:
    issues.append(
        f"CRIT [{col_name}]: HNSW header has astronomical values "
        f"(max={max_elements:,} cur={cur_elements:,}) — "
        f"type-confusion corruption (chroma-core/chroma#4460). "
        f"Delete segment HNSW binaries and run `mempalace repair`."
    )
    return issues
```

## Environment

- mempalace: v3.3.5 (installed from GitHub SHA `60d460b3`)
[...]

PR #1767, opened 2026-06-10
fix(backends): detect corrupt HNSW binary headers with astronomical element counts (#1595)

rodboev:
## Summary

When the chromadb-hnswlib 1.5.x `updatePoint` type-confusion bug (chroma-core/chroma#4460) corrupts an HNSW segment, `cur_element_count` in the binary header is written as a garbage value in the trillions. The existing health check reads only from `index_metadata.pickle` and never validates the binary header in `data_level0.bin`, so `repair-status` and the MCP startup probe report OK even when vector search is functionally offline.

## Root cause

`hnsw_capacity_status()` at `mempalace/backends/chroma.py:651` derives its element count from `_hnsw_element_count()` (line 540), which reads `index_metadata.pickle` and counts `id_to_label` entries. The pickle can be stale or absent while the binary header carries the corrupt value. `_segment_appears_healthy()` at line 291 only checks pickle format bytes and payload file size ratios, never the binary header fields. A segment with a trillion-element header passes all existing checks.

## Fix

Adds `_read_hnsw_binary_header()` to parse the hnswlib header from `data_level0.bin` as unsigned 64-bit integers and validate against a 50M sanity cap. Integrates the check into both `hnsw_capacity_status()` (returns `status="corrupt"` with a diagnostic message) and `_segment_appears_healthy()` (returns False so quarantine catches it). New test file covers valid, truncated, missing, and corrupt header scenarios.

The cap of 50M is well above any realistic palace (the largest known production palace has ~500K drawers) and well below the trillions observed in the type-confusion corruption.

## Test plan
- [x] `python -m pytest tests/test_hnsw_binary_header.py tests/test_hnsw_payload_health.py tests/test_hnsw_capacity.py -v`

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -9,6 +9,7 @@ import os
 import pickle
 import re
 import sqlite3
+import struct
 from collections import defaultdict
 from numbers import Integral
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


rodboev:

PR #1792, opened 2026-06-12
fix(searcher): scope _expand_with_neighbors by parent_drawer_id (#1580)

### LINKED ISSUE #1580: searcher: _expand_with_neighbors stitches unrelated chunks across MCP drawers sharing empty source_file
## Summary

After #1540, `tool_add_drawer(content=<oversized>, source_file=None)` stores every chunk with `source_file=""`. Across multiple such calls, every chunk shares the same empty `source_file` key. A search hit on chunk 0 of drawer A can then pull "neighbors" that are actually chunks from drawer B — `searcher._expand_with_neighbors` queries by `source_file + chunk_index` and ignores `parent_drawer_id`.

Pre-#1540 this could not happen because oversized MCP content was filed as one drawer (the bug #1539 fixed). Post-#1540, the chunks are siblings under a shared empty `source_file`.

## Scope

- Triggered by: `mempalace_add_drawer` with `content` over `chunk_size` (default 800) and no `source_file`.
- Effect: `_expand_with_neighbors` returns text that interleaves chunks from unrelated drawers as if they were sequential context.
- Not affected: direct `get_drawer(drawer_id)` returns correct verbatim content; the underlying drawers are stored correctly. The corruption is purely at the neighbor-expansion / search-results step.

## Reproduction

```python
from mempalace.mcp_server import tool_add_drawer
# Two unrelated oversized pastes, no source_file
r1 = tool_add_drawer(wing="repro", room="paste", content="AAAA" * 250)
r2 = tool_add_drawer(wing="repro", room="paste", content="BBBB" * 250)
# A search that hits chunk_index=0 of r1 expands into chunk_index=1 — which
# may be r2's chunk 1 rather than r1's chunk 1.
```

## Fix shape (one line)

In `mempalace/searcher.py:_expand_with_neighbors` (around line 212-217), either:

**Option A — narrowest:** skip neighbor expansion when `source_file` is empty.

```python
if not src or not isinstance(chunk_idx, int):
    return {"text": matched_doc, "drawer_index": chunk_idx, "total_drawers": None}
```

`src` is already truthy-checked here; adding `or src == ""` makes the guard explicit.

**Option B — preferred:** when `parent_drawer_id` is present in metadata, add it to the `$and` filter so chunks from the same logical drawer group are siblings:

```python
filters = [{"source_file": src}, {"chunk_index": {"$in": target_indexes}}]
parent_id = matched_meta.get("parent_drawer_id")
if parent_id:
    filters.append({"parent_drawer_id": parent_id})
neighbors = drawers_col.get(where={"$and": filters}, ...)
```

Option B preserves neighbor-expansion for the chunked MCP path. Option A loses the feature for that path but is safer if we'd rather not invent a new scoping rule.

## Tests to add

- `tool_add_drawer` oversized content from two distinct drawers without `source_file`, then assert that `_expand_with_neighbors` on chunk 0 of drawer A never returns a document that was originally part of drawer B.

## Out of scope

- Whether `source_file` should default to a per-drawer synthetic key on the MCP path (would change the metadata contract).
- Whether `_expand_with_neighbors` should use `parent_drawer_id` as the primary scope key on the chunked path (larger refactor).

cc @mvalentsev (surfaced during the review of #1540 — merging that PR now and tracking this here so we don't block the fix).

ggettert:
Fixes #1580.

## Summary

Implements **Option B** from the issue: scope `_expand_with_neighbors`
by `parent_drawer_id` (in addition to `source_file`) when the matched
chunk carries one.

## Why this is bigger than the one-line guard relaxation

Pre-fix, the helper bailed early when `source_file` was empty:

```python
if not src or not isinstance(chunk_idx, int):
    return {"text": matched_doc, "drawer_index": chunk_idx, "total_drawers": None}
```

That guard *masked* the bug on the MCP-chunked path (where every chunk
has `source_file=""`) by silently disabling neighbor expansion for it.
So the user-visible symptom on `develop` today is "neighbor expansion
quietly doesn't work for `tool_add_drawer` oversized content"; the
*real* leak only surfaces if two drawers ever share a non-empty
`source_file`.

This PR:

1. Relaxes the early-return: bail only when *both* `source_file` and
   `parent_drawer_id` are missing.
2. Adds `parent_drawer_id` to the `$and` filter (neighbor query *and*
   the `total_drawers` count query) when it's present on the matched
   chunk.
3. Net effect: neighbor expansion is restored for the MCP-chunked path
   *and* hardened so cross-drawer leakage is impossible even if
   `source_file` collides.

## Live verification

Beyond unit tests, I built three harnesses:

**1. Adversarial collision** (two drawers, shared non-empty
`source_file`, distinct `parent_drawer_id`):

| | DEVELOP | BRANCH |
|---|---|---|
| Expansion contains ALPHA-MARKER | ✅ | ✅ |
| Expansion contains BETA-MARKER | ❌ leak | ✅ scoped out |
| `total_drawers` reported | 6 (combined) | 3 (correct) |

**2. Natural MCP path** (oversized `tool_add_drawer`, no
`source_file` — Igor's exact repro):

| | DEVELOP | BRANCH |
|---|---|---|
| Neighbor expansion runs | ❌ short-circuits | ✅ runs, scoped |
| `total_drawers` reported | None | 2 (correct) |

**3. Regression fingerprint** (diary chunks + small drawers + closet
hits + MCP chunked drawers, 3 query patterns):

`diff develop.json branch.json` — *zero diff*. Existing file-backed
and closet-boosted paths are byte-identical pre/post.

## Test coverage

`tests/test_closets.py::TestDrawerGrepExpansion`:
- `test_expand_scopes_by_parent_drawer_id_when_present` — pins the
  #1580 repro (two drawers, shared empty `source_file`, scoped result).
- `test_expand_without_parent_drawer_id_keeps_source_file_scope` —
  pins the legacy diary path keeps source_file-only scope.

Full suite: 259/259 in `tests/test_closets.py` + `tests/test_mcp_server.py`.
Ruff clean.

## Out of scope

- Whether `source_file` should default to a per-drawer synthetic key
  on the MCP path (the issue calls this out explicitly).
- Whether to invalidate / repair palaces that already accumulated
  leak-prone metadata (no migration needed; the fix is purely at the
  query layer).

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -270,18 +270,32 @@ def _expand_with_neighbors(drawers_col, matched_doc: str, matched_meta: dict, ra
     """
     src = matched_meta.get("source_file")
     chunk_idx = matched_meta.get("chunk_index")
-    if not src or not isinstance(chunk_idx, int):
+    parent_id = matched_meta.get("parent_drawer_id")
+    if not isinstance(chunk_idx, int):
         return {"text": matched_doc, "drawer_index": chunk_idx, "total_drawers": None}
+    # Issue #1580: chunks filed via ``tool_add_drawer`` with no ``source_file``
+    # all share ``source_file=""`` and a per-call ``parent_drawer_id``. Without
+    # the parent scope, ``_expand_with_neighbors`` would happily stitch chunk 1
+    # of an unrelated drawer onto chunk 0 of the matched drawer. When the
+    # matched chunk carries a ``parent_drawer_id``, narrow the neighbor query
+    # to that logical drawer so cross-drawer leakage cannot happen. Diary /
+    # file-backed chunks have no ``parent_drawer_id`` and rely on the
+    # ``source_file`` filter, which is unambiguous because every chunk in that
+    # path shares one real, non-empty ``source_file``. If neither scope key is
+    # available, neighbor expansion would be unbounded — fall back to the
+    # matched drawer alone.
+    if not src and not parent_id:
+        return {"text": matched_doc, "drawer_index": chunk_idx, "total_drawers": None}
+    base_filters: list[dict] = []
+    if src:
+        base_filters.append({"source_file": src})
+    if parent_id:
+        base_filters.append({"parent_drawer_id": parent_id})
 
     target_indexes = [chunk_idx + offset for offset in range(-radius, radius + 1)]
     try:
         neighbors = drawers_col.get(
-            where={
-                "$and": [
-                    {"source_file": src},
-                    {"chunk_index": {"$in": target_indexes}},
-                ]
-            },
+            where={"$and": base_filters + [{"chunk_index": {"$in": target_indexes}}]},
             include=["documents", "metadatas"],
         )
     except Exception:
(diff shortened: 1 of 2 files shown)


ggettert:
Closing — I missed PR #1628 when scoping this work. #1628 also closes #1580 and implements the same `parent_drawer_id` scoping fix, covers both `_expand_with_neighbors` and the sibling `search_memories` enrichment loop via a shared `_scoped_source_filter` helper, and already has test coverage for both paths plus the empty-string edge case. Apologies for the duplicate noise.

PR #1794, opened 2026-06-12
fix(hooks): replace mapfile with while-read for bash-3.2 compatibility

MrAshRhodes:
mapfile is a bash-4 builtin; macOS /bin/bash is 3.2. Every Stop hook printed "command not found" to stderr, which Claude Code surfaces as a Stop-hook error. A while-read loop is equivalent and silent on bash 3.2.


igorls:

PR #1812, opened 2026-06-15
fix(mcp): exit orphaned stdio server when parent dies

jasonmaxxxon:
## Summary

Adds a parent-death watchdog for the existing MCP stdio server.

When a client process (Claude / Codex / etc.) crashes, force-quits, or restarts without closing stdio cleanly, `mempalace-mcp` can outlive its parent and keep ChromaDB / HNSW / SQLite file handles open. The existing 8 h idle watchdog catches this eventually; this watchdog catches the orphaned process within roughly 5 s.

The earlier `diary_write` schema fix is already present in `develop` via upstream commit `b1537be`, so this rebased PR now contains only the orphan watchdog slice.

### How the watchdog works

POSIX reparents an orphaned child to PID 1. A daemon thread polls `os.getppid()` every 5 s and exits the process on first PPID == 1 observation.

It handles both cases:
- **died (PPID was N)** — parent was alive at startup, then died later
- **absent at startup** — parent was already gone when child began executing

Escape hatch: `MEMPALACE_MCP_PARENT_WATCHDOG=0` for launchd / supervised contexts that supply their own liveness.

## Test plan

- [x] `uv run python -m py_compile mempalace/mcp_server.py`
- [x] `uv run pytest -q tests/test_mcp_parent_watchdog.py` — 7 passed
- [x] `uv run pytest -q tests/test_mcp_stdio_protection.py tests/test_mcp_server.py -k 'ColdStart or PYTHONPATH or idle or diary_write or schema or handle_request'` — 40 passed, 158 deselected
- [x] `uv run ruff check mempalace/mcp_server.py tests/test_mcp_parent_watchdog.py`
- [x] `uv run ruff format --check mempalace/mcp_server.py tests/test_mcp_parent_watchdog.py`
- [x] `uv run pytest tests/ -q --ignore=tests/benchmarks` — 2860 passed, 5 skipped, 1 warning

## Out of scope

HTTP/SSE / Streamable HTTP shared-server transport (one mempalace serving multiple clients) — separate slice. This PR only stabilizes the existing stdio surface.

Local client config changes (`MEMPALACE_MCP_IDLE_HOURS=1` in `~/.claude.json` and `~/.codex/config.toml`) are personal-env-only and not part of this PR.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -3720,6 +3720,51 @@ def _maybe_eager_warmup_embedder() -> None:
         )
 
 
+_MCP_PARENT_WATCHDOG_ENV = "MEMPALACE_MCP_PARENT_WATCHDOG"
+
+
+def _start_parent_death_watchdog() -> None:
+    """Exit immediately if the parent (client) process dies.
+
+    The MCP stdio protocol carries no application-level liveness signal:
+    when a client crashes, force-quits, or restarts without closing stdio
+    cleanly, the child server can outlive its parent indefinitely — still
+    holding ChromaDB / HNSW / SQLite file handles. The idle watchdog only
+    catches this 8 h later; this thread catches it within seconds.
+
+    POSIX reparents an orphaned child to PID 1 (init / launchd). The
+    watchdog polls ``getppid()`` every 5 s and exits the process as soon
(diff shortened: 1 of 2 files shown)


igorls:

PR #1813, opened 2026-06-16
Perf/project mine prefetch skip check

thismilktea:
## What does this PR do?

  Extends the paginated prefetch pattern from `convo_miner` to `project miner`
  by adding a bulk-prefetched skip fast-path for large project mine runs, while
  preserving the lock-held `file_already_mined()` re-check before writes.

  ## Changes

  ### `palace.py`
  - Added `prefetch_project_mined_mtimes(collection)` — returns `dict[str,
  set[float]]` mapping source files to valid stored mtimes
  - Extracted `_metadata_has_current_normalize_version` and
  `_stored_mtime_matches_current` as shared helpers to keep skip-check semantics
  aligned across miner paths

  ### `miner.py`
  - Removed the lockless per-file pre-check from `process_file()`; the lock-held
  authoritative check remains
  - Added prefetch integration in `_mine_impl()` with threshold gating
  (`PROJECT_MINE_PREFETCH_THRESHOLD = 200`)
  - Outer loop now:
    - skips immediately on prefetched `source_file + mtime` hits
    - falls through to `process_file()` on mtime mismatch or when prefetch is
  disabled

 ### `tests/test_miner.py`
  - Added 5 tests covering:
    - prefetch helper correctness
    - mtime hit skip
    - mtime mismatch bypass
    - small-set threshold bypass
    - `--limit` compatibility
  - Patched affected limit tests to force the non-prefetch path where prefetch 
  is not the behavior under test

  ## Design decisions

  - `dict[str, set[float]]` instead of `set[str]` — project miner must preserve 
  mtime change detection
  - Two-tier design:
    - outer prefetch = fast-path skip

 ## Why

  Project mining previously performed already-mined skip checks file-by-file via
  collection reads.  
  For large, skip-heavy project mine runs, that creates repeated query overhead.

  The codebase already uses paginated prefetch for conversation mining. This 
  change applies the same idea to project mining, while preserving 
  project-miner-specific semantics around `source_mtime`, `normalize_version`, 
  and lock-held re-checks before mutation.

  ## Verification

  Ran:

  ```bash
  uv run ruff check .
  uv run pytest tests/test_miner.py -v -k "prefetch"
  uv run pytest tests/test_miner.py -v -k "prefetch_project_mined_mtimes or
  mine_prefetch or mine_limit"
  ```

  ## Checklist
  - [x] Targeted miner tests pass
  - [x] No hardcoded paths
  - [x] Linter passes (`uv run ruff check .`)

    - lock-held `file_already_mined()` = final authority
  - Threshold-gated to avoid full-collection scans for small file sets


igorls:

PR #1821, opened 2026-06-18
feat(hooks): add a budget-safe SessionEnd save hook for clean exits (#1341)

### LINKED ISSUE #1341: SessionEnd hook missing: short sessions lose memories on clean exit
## Problem

Mem Palace only persists session content when the `Stop` hook crosses the
hardcoded `SAVE_INTERVAL = 15` exchange threshold, or when `PreCompact` fires.

If a session ends cleanly before either of those, nothing is saved. Useful
short sessions (a quick decision, a 5-message debug thread, a "what was the
plan again" check-in) are silently dropped from the palace.

## Repro

1. Start Claude Code in a project where Mem Palace is installed.
2. Have a 5 to 10 exchange conversation that produces something worth
   remembering.
3. `/exit` (or close the terminal) cleanly.
4. Search the palace afterwards. The session is not there.

## Evidence in the source

- `.claude-plugin/hooks/hooks.json` registers only `Stop` and `PreCompact`.
- `mempalace/hooks_cli.py` line 5 docstring says: "Supported hooks:
  session-start, stop, precompact". No `session-end`.
- `mempalace/hooks_cli.py` line 17: `SAVE_INTERVAL = 15` is a module
  constant, not read from `MempalaceConfig` or env.
- `hook_stop` (line 551) gates the save on `since_last >= SAVE_INTERVAL`,
  so sub-threshold turns never flush.

## Proposed fix

Add a `SessionEnd` hook that always flushes, regardless of `SAVE_INTERVAL`.

Concretely:

1. `hooks.json`: register `SessionEnd` pointing to a new
   `mempal-session-end-hook.sh` (a 3 line wrapper, same shape as the
   existing stop / precompact wrappers, calling
   `mempalace hook run --hook session-end --harness claude-code`).
2. `hooks_cli.py`: add `hook_session_end(data, harness)`. Mostly a copy of
   `hook_stop` minus the threshold gate, so it always calls
   `_save_diary_direct` + `_ingest_transcript` if there is anything new
   since the last checkpoint.
3. `hooks_cli.py` dispatch dict (line ~703): add
   `"session-end": hook_session_end`.

Claude Code's `SessionEnd` is best-effort and non-blocking, so this is
purely additive. It cannot delay or break session termination.

## Bonus ask (separate issue is fine)

Make `SAVE_INTERVAL` configurable via `MempalaceConfig`, e.g.
`hook_save_interval`, defaulting to 15. Some users want shorter intervals
for chatty sessions, some longer for focused work.

Happy to send a PR for the SessionEnd part if it sounds right to you.

### LINKED ISSUE #1814: Add a SessionEnd hook for a deterministic mine on clean exit (plugin only wires Stop + PreCompact)
## Summary

The plugin's `hooks/hooks.json` wires auto-save mining on **`Stop`** and **`PreCompact`**, but there is **no `SessionEnd` hook**. On a clean session exit that never triggered a compaction, there is no guaranteed final mine of the just-ended conversation. A deterministic `SessionEnd` mine would close that gap.

## Why this matters

- `PreCompact` only fires when a compaction actually occurs — short sessions never compact, so it never runs for them.
- `Stop`-driven capture is interval/throttle-based, so the last few exchanges before the user exits can go uncaptured.
- `SessionEnd` is the natural place for a guaranteed final flush. It cannot *block* the model (Claude Code disallows that on `SessionEnd`), but it **can** run `mempalace mine` on the transcript programmatically — no model interaction required, fully deterministic.

## Request

Add a `SessionEnd` entry to `hooks/hooks.json` that runs a deterministic mine of the session transcript (equivalent to the existing pre-compact mine), so a clean exit always flushes the final window.

## Current workaround

Users can wire `SessionEnd` → the precompact mine script in their own `settings.local.json`. It works, but it lives outside the plugin and is easy to lose on reinstall / hard to discover.

---

_Environment: Claude Code on Windows; plugin installed via marketplace._

mvalentsev:
## What does this PR do?

Closes #1341. Closes #1814 (a duplicate of #1341).

Short sessions that exit cleanly can still lose memories: MemPalace saves only
when the `Stop` hook crosses `SAVE_INTERVAL` (15 human messages) or when
`PreCompact` fires, so a short session that ends without ever compacting is
dropped from the palace.

A `SessionEnd` hook that mines synchronously gets killed before it can save.
From the Claude Code hooks reference:

> SessionEnd hooks have a default timeout of 1.5 seconds. [...] Timeouts set on
> plugin-provided hooks do not raise the budget.

(https://code.claude.com/docs/en/hooks)

A `mempalace` start measures ~8s cold and ~1.7s warm, both past that 1.5s
window, and a plugin-shipped `timeout` cannot widen it. So the hook backgrounds
its work and returns immediately (foreground ~0.01s, measured), and the handler
spawns the capture as a detached child that outlives the exiting session:

- `.claude-plugin/hooks/hooks.json`: register `SessionEnd`.
- New wrappers `hooks/mempal_session_end_hook.sh` and the plugin
  `.claude-plugin/hooks/mempal-session-end-hook.sh` background the dispatch.
- `mempalace/hooks_cli.py`: `hook_session_end()` validates `transcript_path`,
  then flushes regardless of `SAVE_INTERVAL` (an in-process diary checkpoint,
  then a detached transcript ingest and project mine), then clears the
  per-session marker.
- `mempalace/cli.py`: add `session-end` to the `hook run` choices.

This stays inside both the 1.5s budget and the project's "hooks under 500ms /
background everything" principle. `SessionEnd` has no decision control, so the
hook only ever saves.

**Scope:** Claude Code only. The handler is harness-agnostic
(`mempalace hook run --hook session-end --harness <name>`). Cursor has a
`sessionEnd` event that is not wired yet (follow-up); Codex has none today;
Antigravity has none and saves via `Stop`.

**Windows:** verified on POSIX. On Windows only the mine grandchild (spawned
with detached-process flags) is designed to break away from the session; the
backgrounded hook process and the in-process diary checkpoint are best-effort,
and the teardown path has no Windows CI coverage yet.

Builds on the request and design sketch in #1341 and the first implementation
attempt in #1738; the new piece is the budget-safe backgrounding, since that
earlier attempt mined synchronously and the 1.5s plugin budget reaps it before
it can save.

## How to test

```
uv run pytest tests/test_hooks_cli.py tests/test_cli.py tests/test_hooks_shell.py tests/test_hooks_bash_compat.py tests/test_claude_plugin_hook_config.py -v
uv run ruff check .
uv run ruff format --check .
```

246 passed, 1 skipped locally. The shell tests prove backgrounding: the
foreground returns before a 2s worker finishes, and the detached worker still
completes. End to end, a 5-message sub-threshold session was captured and found
via `mempalace search`, and a unit test confirms the diary checkpoint is visible
via `diary_read`.

## Checklist

- [x] No hardcoded paths

### DIFF
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -41,9 +41,10 @@ After installing the plugin, run the init command to complete setup (installs th
 
 ## Hooks
 
-MemPalace registers two hooks that run automatically:
+MemPalace registers three hooks that run automatically:
 
 - **Stop** -- Saves conversation context every 15 messages.
+- **SessionEnd** -- Runs one final save in the background on a clean exit, so short sessions that never hit the Stop interval or a compaction are still captured.
 - **PreCompact** -- Preserves important memories before context compaction.
 
 Set the `MEMPAL_DIR` environment variable to a directory path to automatically run `mempalace mine` on that directory during each save trigger.
--- a/.claude-plugin/hooks/hooks.json
+++ b/.claude-plugin/hooks/hooks.json
@@ -12,6 +12,17 @@
         ]
       }
     ],
+    "SessionEnd": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/mempal-session-end-hook.sh\"",
+            "timeout": 10
+          }
+        ]
+      }
+    ],
     "PreCompact": [
       {
         "hooks": [
--- a/.claude-plugin/hooks/mempal-session-end-hook.sh
+++ b/.claude-plugin/hooks/mempal-session-end-hook.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+# MemPalace SessionEnd Hook — thin wrapper calling the Python CLI.
+#
+# Claude Code documents a default SessionEnd hook timeout of 1.5s, and
+# "timeouts set on plugin-provided hooks do not raise the budget"
+# (https://code.claude.com/docs/en/hooks). A cold `mempalace` start alone
+# exceeds 1.5s, so the final mine must NOT run in the foreground — it would be
+# killed before it saved anything. Unlike the foreground Stop/PreCompact plugin
+# wrappers, this one backgrounds the hook and returns immediately; the detached
+# child finishes the save after the session has exited. All logic lives in
+# mempalace.hooks_cli for cross-harness extensibility.
+run_mempalace_hook() {
+  if command -v mempalace >/dev/null 2>&1; then
+    exec mempalace hook run "$@"
+  fi
+
+  MEMPAL_PYTHON_BIN="${MEMPAL_PYTHON:-}"
+  if [ -z "$MEMPAL_PYTHON_BIN" ] || [ ! -x "$MEMPAL_PYTHON_BIN" ]; then
+    MEMPAL_PYTHON_BIN="$(command -v python3 2>/dev/null || echo python3)"
+  fi
+  if "$MEMPAL_PYTHON_BIN" -c "import mempalace" >/dev/null 2>&1; then
+    exec "$MEMPAL_PYTHON_BIN" -m mempalace hook run "$@"
+  fi
+
+  if command -v python >/dev/null 2>&1 && python -c "import mempalace" >/dev/null 2>&1; then
+    exec python -m mempalace hook run "$@"
+  fi
+
+  echo "MemPalace hook error: could not find a runnable mempalace command or module" >&2
+  exit 1
+}
+
+# Capture stdin (the SessionEnd JSON) before backgrounding — the parent's
+# stdin is gone once we return. Forward it to the detached worker, which runs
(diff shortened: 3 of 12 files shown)


mvalentsev:
Addressed in 3a2954c: `hook_session_end` now runs `transcript_path` through `_validate_transcript_path` before any ingest or diary write, so a traversal or wrong-suffix path is rejected while the independent project mine still runs. Covered by `test_session_end_rejects_invalid_transcript_path`.

/gemini review

mvalentsev:
Reordered in 4054f61: the diary checkpoint now runs before the detached mines, so it no longer contends with a background mine for the palace lock (same order as `hook_stop`). The `transcript_path` validation from the previous round stays in place.

mvalentsev:
Closing this in favor of #1820, which now carries the budget-safe backgrounding implementation: my commit 10e1450 was cherry-picked there with `-x` (authorship and source SHA preserved), and #1820 closes both #1341 and #1814.

Thanks @fatkobra for folding it in.

PR #1824, opened 2026-06-18
perf(status): aggregate wing/room counts in the backend instead of client-side fetch

gavinmcfall:
## Problem
`mempalace_status` builds its wing/room breakdown by pulling **every** drawer's metadata to the client (`_get_cached_metadata`) and counting in Python. That's O(n) memory and time.

On a **418,997-drawer** pgvector palace this took **~344s** and OOM-killed a 2Gi pod. Since the MCP wake-up protocol calls `status` first, large palaces effectively can't wake up.

## Change
- Add an optional collection hook `count_by_metadata_field(field) -> {value: count}`.
- Implement it on the pgvector backend as a single `SELECT metadata->>field, count(*) ... GROUP BY 1`.
- `tool_status` uses it when the collection provides it; backends without it keep the existing client-side path unchanged (duck-typed, fully backward compatible).

## Result
Same 418k palace: `status` drops from **~344s → ~0.4s**, flat memory. Counts verified identical to the client-side tally (7 wings / 50 rooms; e.g. `my_other_repos`=169,587).

## Notes
- Backward compatible — chroma/qdrant/sqlite paths are untouched.
- `list_wings` / `list_rooms` have the same client-side pattern and could adopt this hook in a follow-up; kept this PR focused on the wake-up path.

### DIFF
--- a/mempalace/backends/pgvector.py
+++ b/mempalace/backends/pgvector.py
@@ -1058,6 +1058,21 @@ class PgVectorCollection(BaseCollection):
             return 0
         return self._client.count_rows(self._table)
 
+    def count_by_metadata_field(self, field: str) -> dict[str, int]:
+        """Return ``{value: count}`` for a top-level metadata key, aggregated in
+        SQL. Lets callers (e.g. the status overview) tally by ``wing`` / ``room``
+        without streaming every drawer's metadata to the client — O(1) memory and
+        a single query instead of O(n). ``NULL`` values bucket under ``"unknown"``."""
+        self._ensure_open()
+        if not self._table_exists():
+            return {}
+        sql = (
+            f"SELECT metadata->>%s AS k, count(*) AS c "
+            f"FROM {_quote_identifier(self._table)} GROUP BY 1"
+        )
+        rows = self._client._execute(sql, [field], fetch=True)
+        return {(r[0] if r[0] is not None else "unknown"): int(r[1]) for r in rows}
+
     def lexical_search(self, *, query: str, n_results: int = 10, where: Optional[dict] = None):
         _validate_where(where)
         pushdown = None if _requires_local_filter(where) else where
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -1073,13 +1073,23 @@ def tool_status():
         "backend": _selected_backend_name(),
     }
     try:
-        all_meta = _get_cached_metadata(col)
-        for m in all_meta:
-            m = m or {}
-            w = m.get("wing", "unknown")
-            r = m.get("room", "unknown")
-            wings[w] = wings.get(w, 0) + 1
-            rooms[r] = rooms.get(r, 0) + 1
+        grouper = getattr(col, "count_by_metadata_field", None)
+        if callable(grouper):
+            # Efficient path: backends that can aggregate server-side (e.g.
+            # pgvector via SQL GROUP BY) tally wing/room counts without
+            # streaming every drawer's metadata to the client. On a large
+            # palace the client-side fetch below is O(n) memory and can take
+            # minutes (or OOM); this is a single query.
+            wings.update(grouper("wing"))
+            rooms.update(grouper("room"))
+        else:
+            all_meta = _get_cached_metadata(col)
+            for m in all_meta:
+                m = m or {}
+                w = m.get("wing", "unknown")
+                r = m.get("room", "unknown")
+                wings[w] = wings.get(w, 0) + 1
+                rooms[r] = rooms.get(r, 0) + 1
     except Exception as e:
         logger.exception("tool_status metadata fetch failed")
         result["error"] = str(e)


messelink:
Hey @gavinmcfall — thanks for filing this, the SQL is essentially the right shape. A couple of things have landed in adjacent threads since you opened this PR that you might not have seen:

- **#1835** (@fatkobra) proposes the canonical design for this kind of aggregation: method name `facet_counts(field, where=None)`, capability token `supports_metadata_facets`, default raises `UnsupportedCapabilityError` (matching the existing `lexical_search` precedent). Worth a read — same intent as your `count_by_metadata_field`, more rigorous fit with the existing backend interface.
- **#1868** (@Prayaksh) implements that spec for Qdrant — active review, looks close to merge.

I'd suggest aligning this PR's surface area with #1835's spec (rename `count_by_metadata_field` → `facet_counts`, add `where=None`, switch to the capability-token + `UnsupportedCapabilityError` default). Your SQL is good; it's mainly the wrapper.

Two concrete additions from building the same thing for our pgvector deployment that might be worth folding in:

1. **`where` parameter** to match the spec. Pgvector form: thread the existing `_where_to_sql` translator into the GROUP BY query — keeps array/object semantics consistent with the `query`/`get` paths.
2. **`EmbeddingCollection` explicit forwarder.** When I added `count_by_metadata_*` to `BaseCollection`, the wrapped pgvector implementation was shadowed — `EmbeddingCollection.__getattr__` doesn't fire for methods inherited from `BaseCollection` (Python MRO resolves them first). Without an explicit forwarder on `EmbeddingCollection`, the call silently returns the base default. The `distance_metric` block in `embedding_wrapper.py` has the same shape.

Happy to send a follow-up PR with these on top of yours once the design is settled, or you can pick them up directly — whatever's easier.

messelink:
Cross-link follow-up to my earlier comment: the wire-byte half I mentioned (the `with_document` skip for metadata-only fetches, closing #1840's documented follow-up) is now filed as #1892. Disjoint from your PR's aggregation work — different surface area (scroll_rows parameter + get_all_metadata override vs. your `count_by_metadata_field`), and the two can land in either order. The alignment question for your PR re: #1835's spec / #1868's `facet_counts` naming is still the main open thread here, separate from the wire-byte work.

igorls (changes requested):
Thanks for this, and especially for the concrete incident data — a 419k-drawer pgvector palace taking ~344s and OOM-killing a 2Gi pod on `status` is exactly the kind of real-world pressure that justifies server-side aggregation. The SQL here is sound and injection-safe (bound `%s` field, `_quote_identifier` on the table, `NULL → "unknown"` matching the client-side ground truth).

The blocker isn't correctness — it's interface convergence. This PR predates the agreed design in #1835, which a parallel PR (#1868) now implements as the canonical contract:

- a `BaseCollection.facet_counts(field, where=None) -> dict[str, int]` method (default raises `UnsupportedCapabilityError`, mirroring `lexical_search`),
- a `supports_metadata_facets` capability token on the backend,
- all four call sites (`tool_status`, `tool_list_wings`, `tool_list_rooms`, `tool_get_taxonomy`) checking the token and falling back to client-side counting.

This PR instead adds `count_by_metadata_field(field)` (no `where`, duck-typed via `getattr`/`callable`, only `tool_status`), which would entrench a second, non-conforming aggregation interface — and it textually conflicts with #1868 (both rewrite the same wing/room loop in `tool_status`).

Proposed path so your work lands cleanly:
1. We merge #1868 first (it establishes the base method, the capability token, and all four call sites).
2. Rebase this PR on top as the **pgvector** implementation of `facet_counts`: rename `count_by_metadata_field` → `facet_counts(field, where=None)`, add the `where → SQL WHERE` mapping, register `supports_metadata_facets` on `PgVectorBackend`, drop the `tool_status` edit entirely (those call sites already live in #1868), and add unit tests for the new method.

After that rebase this shrinks to roughly one method on `pgvector.py` + one token line, with no `mcp_server.py` changes and no conflict. The SQL, the NULL bucketing, and the incident motivation all carry straight over. Would you be up for reshaping it that way? Happy to help line it up once #1868 is in.

One thing worth confirming for your scenario: the wing/room tools already short-circuit through a `_sqlite_taxonomy()` cache when present, so it's worth verifying the OOM path actually reached the client-side counting branch (i.e. the SQLite cache wasn't populated) so the fix lands where the incident occurred.

messelink:
@gavinmcfall — thank you for the original SQL and the concrete incident data; both carry forward. Filed #2038 as a take-over of this PR's territory, implementing the rebase path @igorls proposed above: renamed to `facet_counts` per the merged #1868 contract, registered `supports_metadata_facets` on `PgVectorBackend`, no `tool_status` edits, and added unit tests. Attribution kept in the PR body.

Filing separately rather than pushing to your branch since the changes are substantial and you may still want to iterate on your own version. If maintainers prefer to close this in favor of the new PR that seems reasonable to us; if you'd rather reshape this one instead, let us know and we can withdraw the new PR.

gavinmcfall:
Agreed — #2038 is the right shape and supersedes this. The `facet_counts` + `supports_metadata_facets` contract from #1835/#1868 is a better fit than my `count_by_metadata_field`, and threading it through the already-merged `tool_status` wiring is cleaner than my edits there. Apologies for leaving the review feedback unanswered. Thanks for carrying the incident data forward — happy to see it close this way. Closing in favour of #2038.

Issue #1801, opened 2026-06-14
Long-lived stdio server stops answering tools/list after extended uptime (small frames still work)


JeanBaptisteRenard:
## Summary

A long-lived `mempalace.mcp_server` stdio process stops answering **`tools/list`** after extended uptime (days), returning a JSON-RPC error `{"code":0,"message":""}`, while `initialize`, `ping`, and `tools/call` on the **same** process keep working. Restarting the process fixes it immediately. The process stays alive and the socket stays open, so neither a TCP/port check nor a restart-on-exit policy detects it — the client (Claude Code, via `mcp-proxy`) just reports `MCP error 0` and loads **zero** tools.

## Environment

- mempalace **3.3.6** (installed from the v3.3.6 tag tarball), Python 3.12
- Run as one long-lived stdio backend behind `mcp-proxy 0.12.0 --stateless` (mcp SDK 1.27.2), HTTP at `127.0.0.1:8765/mcp`
- Backend launched as: `python3 -m mempalace.mcp_server --palace /data/palace`

## Reproduction / evidence (all observed live)

Through the ~3-day-old process:

| Method | Result |
|---|---|
| `initialize` | ✅ normal result |
| `ping` | ✅ `{}` |
| `tools/call` (e.g. `list_wings`) | ✅ reaches and runs the handler |
| **`tools/list`** | ❌ `{"jsonrpc":"2.0","id":N,"error":{"code":0,"message":""}}`, deterministic |

Controls that isolate it to **process uptime**, not config:

- A **freshly started** `python3 -m mempalace.mcp_server --palace …` driven directly over stdio returns all 30 tools fine.
- Two **fresh** `mcp-proxy` instances (one `--stateless`, one session-mode) over fresh backends both return 30 tools — so it is not the proxy, not `--stateless`, not the tool schemas.
- `TOOLS` is intact (30 entries, each with `description` + `input_schema`); the `tools/list` dispatch in `mcp_server.py` is a trivial comprehension over it.
- `docker restart` of the backend → `tools/list` returns 30 tools immediately.

The only differentiator is uptime, and only the **largest response frame** (`tools/list`, all 30 schemas) fails while small frames survive.

## Suspected cause (hypothesis)

[...]

PR #1859, opened 2026-06-23
fix(palace): process-wide mine_palace_lock re-entrancy so the HTTP transport can write

trongnguyenbinh:
## Problem

When MemPalace runs as a remote MCP server over the HTTP transport (`mempalace-mcp --transport http`), **reads work but writes fail** with:

```
palace ... is held by PID <self> (mempalace-mcp --transport http); wait for it to finish or stop the holder before retrying
```

The server reports the palace as held by *itself*, so `add_drawer` / `update_drawer` never succeed over HTTP. (Over stdio it's fine because everything runs on one thread.)

## Root cause

1. The HTTP server takes a long-lived per-palace **writer-lease** at startup via `_acquire_mcp_writer_lock()` → `mine_palace_lock(...).__enter__()` (`mcp_server.py`). That holds the flock for the process lifetime and records the lock as held **on the acquiring thread**.
2. `mine_palace_lock` re-entrancy is **thread-local** (`palace.py`: `_palace_lock_holders = threading.local()`, checked via `_held_by_this_thread`).
3. `ThreadingHTTPServer` dispatches each request on a **different worker thread**. A write handler runs on a worker thread where the thread-local guard is empty, so it tries to re-acquire the flock and hits `BlockingIOError` (the process already holds it via the lease fd) → `MineAlreadyRunning`.

Reads don't take the lock, so they're unaffected. This is distinct from the multi-*process* races in #1229/#974 — it is a multi-*thread*, single-process self-conflict specific to the threaded HTTP transport (#1801).

## Fix

Make the re-entrancy record **process-wide** instead of thread-local: a module-level set guarded by a `threading.Lock`, tagged with pid so a forked child does not inherit re-entrant credit. A write from any thread of the process that already holds the lease now passes through without re-acquiring.

Safe because:
- flock is per-process — two threads can't independently hold it anyway;
- HTTP writes are serialized by `_HTTP_REQUEST_LOCK`;
- fork-safety preserved via the pid check;
- same-thread nesting (`miner.mine` → `ChromaCollection.upsert`) still works;
- cross-process protection unchanged (`MineAlreadyRunning` still raised between processes).

## Testing

- New regression test `test_mine_palace_lock_reentrant_across_threads_same_process` (fails on the old thread-local code, passes now).
- `tests/test_palace_locks.py` → 12 passed; `test_chroma_collection_lock.py` + `test_mine_lock_lifecycle.py` + `test_mcp_http_transport.py` → 21 passed. No regressions.
- `ruff check` + `ruff format` clean.
- **Verified live:** deployed this branch to a remote `--transport http` server; `add_drawer` over MCP now succeeds (previously errored) and `status` reflects the new drawer.

### DIFF
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -969,9 +969,28 @@ def _validate_palace_fts5_after_mine(palace_path: str) -> None:
 # "we hold it" semantically — the child must reacquire. The pid check clears
 # stale state so a forked child correctly hits the fcntl path. Access is guarded
 # by ``_palace_lock_guard`` because the set is now shared across threads.
+#
+# Fork safety: ``_palace_lock_guard`` is a real ``threading.Lock``, so a child
+# forked while another thread held it would inherit it locked (the holder thread
+# does not exist in the child) and deadlock on the next acquire. An at-fork
+# handler (registered below) replaces the guard with a fresh unlocked lock and
+# clears state in the child, which must reacquire the flock anyway.
 _palace_lock_guard = threading.Lock()
 _palace_lock_pid = None
 _palace_lock_keys = set()
+
+
+def _reset_palace_lock_state_after_fork() -> None:
+    """Reset lock state in a forked child to avoid an inherited-locked deadlock."""
+    global _palace_lock_guard, _palace_lock_pid, _palace_lock_keys
+    _palace_lock_guard = threading.Lock()
+    _palace_lock_keys = set()
+    _palace_lock_pid = os.getpid()
+
+
+# Availability: Unix (no-op elsewhere — Windows has no fork()).
+if hasattr(os, "register_at_fork"):
+    os.register_at_fork(after_in_child=_reset_palace_lock_state_after_fork)
 
 
 def _holder_keys_locked():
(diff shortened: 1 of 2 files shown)


trongnguyenbinh:
Good catch on the fork-deadlock risk. Fixed in 0a7c94b: registered an `os.register_at_fork(after_in_child=...)` handler that replaces `_palace_lock_guard` with a fresh unlocked `threading.Lock` and clears the holder state in the child (which must reacquire the flock anyway). Guarded with `hasattr(os, "register_at_fork")` for Windows.

ruff clean; `tests/test_palace_locks.py` (12) + chroma-collection-lock / mine-lock-lifecycle / mcp-http-transport (21) still pass.

igorls (approved):
Reviewed and verified locally against current develop: 22 palace/lock tests pass, the thread-local→process-wide rename is complete (no dangling references), and the merge with the new _chroma_palace_helper is clean. The diagnosis is precise (ThreadingHTTPServer dispatches writes on worker threads where the thread-local guard is empty) and the fix is sound: flock is per-process so the process is the correct re-entrancy boundary, HTTP writes are already serialized by _HTTP_REQUEST_LOCK, and the os.register_at_fork child reset prevents an inherited-locked deadlock. This unblocks remote HTTP-transport writes (relevant to #1877). Thanks!

PR #1879, opened 2026-06-26
Add opt-in MCP audit telemetry

GanizaniSitara:
## Opt-in MCP audit telemetry

Adds opt-in, off-by-default per-request audit telemetry to the in-process HTTP transport. No behaviour change and no disk I/O unless an operator sets `MEMPALACE_AUDIT=1`.

- Appends a per-request JSONL event (method, tool, latency, fingerprinted args, result summary) to `~/.mempalace/service_logs/mcp_audit.jsonl` (override via `MEMPALACE_AUDIT_FILE`); log created `0600`, directory `0700`, and an existing log best-effort re-tightened to `0600`.
- Privacy-first: sensitive content (search responses, drawer text, context) is stored as a length + truncated SHA-256 fingerprint (`sha256_16`), never raw, unless `MEMPALACE_AUDIT_SEARCH_RESPONSE_TEXT=1` is also set — and even then the raw text is bounded by `MEMPALACE_AUDIT_SEARCH_RESPONSE_MAX_CHARS` (default 2000, parsed defensively so a bad value can't crash startup).
- Zero overhead when disabled: `handle_request` takes a fast-path and builds no event. Telemetry construction/write is wrapped so a failure can never disrupt the request it observes.
- `mempalace audit` summarises the log (methods, tools, search empty-rate, latency p50/p95, top queries, recent errors).

Sits behind the transport's existing Host/Origin policy and the optional `MEMPALACE_MCP_HTTP_TOKEN` bearer auth.

> The read-only `/dashboard` endpoint that was originally part of this PR has been **dropped** per maintainer feedback — this PR is now audit-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,31 @@ cross-wing navigation, drawer management, and agent diaries. Installation
 and the full tool list:
 [mempalaceofficial.com/reference/mcp-tools](https://mempalaceofficial.com/reference/mcp-tools.html).
 
+### Audit telemetry (opt-in)
+
+Opt-in per-request audit telemetry for the HTTP transport, **off by default** —
+no behaviour change and nothing is written to disk unless you enable it:
+
+- **Audit telemetry** (`MEMPALACE_AUDIT=1`) — appends a per-request JSONL event
+  (method, tool, latency, fingerprinted args, result summary) to
+  `~/.mempalace/service_logs/mcp_audit.jsonl` (override with
+  `MEMPALACE_AUDIT_FILE`). Sensitive content is stored as a length plus a
+  truncated SHA-256 fingerprint (first 16 hex chars), never raw, unless
+  `MEMPALACE_AUDIT_SEARCH_RESPONSE_TEXT=1` is also set — and even then the raw
+  text is bounded by `MEMPALACE_AUDIT_SEARCH_RESPONSE_MAX_CHARS`. Summarise the
+  log with `mempalace audit`.
+
+| Env var | Default | Effect |
+|---|---|---|
+| `MEMPALACE_AUDIT` | off | Enable per-request audit telemetry |
+| `MEMPALACE_AUDIT_FILE` | `~/.mempalace/service_logs/mcp_audit.jsonl` | Audit log path |
+| `MEMPALACE_AUDIT_SEARCH_RESPONSE_TEXT` | off | Also store raw search-response snippets (default: fingerprint only) |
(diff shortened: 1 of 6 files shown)


igorls (changes requested):
Thanks for this — the **audit telemetry** half is genuinely well done, and the privacy-first design fits the project precisely: opt-in and off by default, zero overhead when disabled, fingerprint-only by default with raw content gated behind a second explicit env var, and `0600`/`0700` perms on the log. Opt-in local auditability like this is welcome.

The one thing we can't take is the **dashboard**. MemPalace intentionally ships no UI/dashboard surface — that's a standing design decision, separate from (and stricter than) the telemetry/privacy stance in CLAUDE.md. So could you split this PR down to just the audit feature and drop the dashboard?

The dashboard code looks cleanly separable already:
- `_DASHBOARD_ENABLED`, `_DASHBOARD_CSS`, `_dashboard_data()`, `_dashboard_html()`, `_dashboard_response_body()`
- the `GET /dashboard` route block in the handler
- the `MEMPALACE_DASHBOARD` env var + the two README rows + config-table entry
- `test_dashboard_endpoint_renders_activity` / `test_dashboard_disabled_returns_404`

Everything else — `audit.py`, the `mempalace audit` CLI summary, and the per-request audit logging — is exactly what we'd like to land. Drop the dashboard half and I'll get the audit half reviewed and merged. Appreciate the careful work on the privacy model here.

GanizaniSitara:
Thanks @igorls — done. I've dropped the dashboard entirely (the `MEMPALACE_DASHBOARD` env var, the `GET /dashboard` route, the render/CSS code, the dashboard tests, and the now-unused `import html`) and rebased this down to a single audit-only commit.

`audit.py`, the `mempalace audit` CLI, and the per-request logging are unchanged in intent. While here I also folded in the automated-review suggestions on the audit path:

- defensive parsing of `MEMPALACE_AUDIT_SEARCH_RESPONSE_MAX_CHARS` so a bad value can't crash startup, plus a bounded default (2000) so stored text is never unbounded even with raw capture on;
- a `try/except` around telemetry build/write so it can never disrupt the request it observes;
- a true fast-path in `handle_request` when auditing is off (no timer, no event);
- an honest `sha256_16` field name for the truncated fingerprint.

Full suite green, ruff clean. Ready when you are.

GanizaniSitara:
Rebased the audit-only branch onto current upstream/develop (94e72ba). Revalidated on Windows: 368 passed, 3 skipped across test_mcp_server.py and test_cli.py; Ruff passes on all changed files. The dashboard remains excluded as requested.

Issue #1876, opened 2026-06-26
use local sqlite as source of truth


blinkinglight:
**What problem does this solve?**
currently mempalace_status request remote qdrant to get data, so very slow, takes minutes to resolve, to get one by one wings and other things

**What's the proposed solution?**
use sqlite to get wings and other needed data

**Alternatives considered:**
-

blinkinglight:
same with mempalace_list_wings - takes 10 minutes to resolved from remote qdrant

mvalentsev:
The slow status / list_wings against a remote Qdrant backend is already tracked. #1796 fixed the O(n²) part (status re-scrolled the whole collection per page) and shipped in 3.5.0. #1835 then added server-side facet counts for wing/room stats (O(1)); it merged after 3.5.0, so it's on develop, not in 3.5.0 yet. If you're on 3.5.0 you already have the #1796 fix, with the facet one to follow.

Issue #1874, opened 2026-06-26
store vectors in local sqlite db so they vector db could be swapped easily


blinkinglight:
**What problem does this solve?**
if there is 100k+drawers it takes forever to reembed on cheap gpu 

**What's the proposed solution?**
store embedded vectors in sqlite, and push them to vectors db on change 

**Alternatives considered:**
-

trifo-tech:
+1 to this. We've already implemented this pattern for 5,184 drawers — storing vectors in sqlite-vec (vec_search.db) with EmbeddingGemma (384-dim).

Results vs ChromaDB:
- Search: 0.003s (vec) vs 2-5s (ChromaDB via MCP)
- WSL restart: zero corruption (ChromaDB HNSW regularly breaks — see #1845)
- No server, no deadlocks

Our labeling and PCA pipelines (nightly cron jobs) run against vec_search.db and have been rock-solid for weeks. The last piece missing is native MCP server support — right now MCP tools (`mempalace_search` etc.) still go through ChromaDB while scripts use sqlite-vec, creating an awkward dual-backend split.

If the MemPalace team is considering backend pluggability, we'd be happy to test an sqlite-vec MCP transport.

mvalentsev:
Swappable backends already landed in #1679: chroma stays the default, with sqlite_exact, qdrant and pgvector selectable. The "store vectors in local sqlite" half is covered by sqlite_exact today (local SQLite + numpy vectors, no extra dependency), and a dedicated sqlite-vec backend with vec0 ANN search is in progress in #1886. The "keep sqlite as a cache and push to the vector db on change" part is different - that makes sqlite the backend itself rather than a sync layer in front of another store.

PR #1886, opened 2026-06-27
feat(backends): add sqlite_vec backend with vec0 ANN search

### LINKED ISSUE #1845: MCP tools hang forever after an interrupted mine — chromadb 1.x compactor cannot backfill WAL into the HNSW segment (CLI unaffected)
## Summary
After a `mempalace mine` is interrupted/killed mid-write, **every MCP tool call hangs indefinitely** (`mempalace_status`, `mempalace_search`, `mempalace_reconnect`, `mempalace_add_drawer`) — no timeout, no error. The **CLI keeps working** (`mempalace status`, `mempalace repair-status` return in seconds). The hang **survives full reboots** because the broken state is on disk. Root cause is a damaged HNSW segment writer + pending rows in chromadb's `embeddings_queue` that the chromadb-1.x compactor cannot apply.

## Environment
- mempalace **3.4.0**, chromadb **1.5.9** (Rust core / "compactor")
- Windows 11, Python 3.13, onnxruntime-gpu, embedding model `embeddinggemma` (384-dim), backend `chroma`
- Palace size ~268k drawers (~1.4 GB sqlite + ~450 MB HNSW)

## Symptom
- MCP: any tool that opens the chroma collection hangs forever. The harness logs `Tool 'mempalace_status' still running (720s elapsed)`. The hang is in **native code** — it ignores SIGINT.
- CLI: `mempalace status` and `mempalace repair-status` work fine (they read sqlite counts and **never open a chromadb client**).

## Root cause
An interrupted `mine` leaves unflushed rows in chroma's `embeddings_queue` plus a damaged HNSW **segment writer**. Any operation that opens the collection (`count()`, query, add) makes the chromadb-1.x **compactor** try to backfill the queue into the HNSW segment, which fails:

```
chromadb.errors.InternalError: Error executing plan:
  Error sending backfill request to compactor:
  Failed to apply logs to the hnsw segment writer
```

In the long-lived MCP server this surfaces as an **infinite hang**; called directly from Python it raises the error above.

## How to confirm (read-only)
```python
import chromadb
c = chromadb.PersistentClient(path=r"<HOME>/.mempalace/palace")
for col in c.list_collections():
    print(col.name, col.count())   # -> InternalError: ...hnsw segment writer
```
Note: `mempalace repair-status` passes (sqlite vs HNSW counts match within flush-lag) because it **never opens the chroma client** — so it does **not** catch a broken segment writer.

## What it was NOT (ruled out, each with evidence)
- **Stale lock files** — cleaned `~/.mempalace/locks/mine_palace_*.lock` + `hook_state/mine_pids/*.pid`; still hangs. (Separate leak: 2200+ empty `locks/*.lock` accumulate.)
- **HNSW corruption** — `repair-status` OK; `link_lists.bin` tiny (~2 MB), no blowup.
- **HF Hub network** — model is cached locally; `HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1` still hangs.
- **GPU/CUDA/TensorRT** — `embedding_device: cpu` still hangs (the hang is *before* embedding).
- **`max_seq_id` poison** — values sane (`max_seq_id == embeddings count`).

## Fix that worked
`repair --mode from-sqlite` rebuilds the HNSW from the sqlite ground truth, bypassing the broken client/segment:

```bash
# In-place fails on Windows (see gotcha below), so rebuild into a new path
# reading the live sqlite read-only:
mempalace --palace <HOME>/.mempalace/palace_rebuilt repair \
          --mode from-sqlite --source <HOME>/.mempalace/palace --yes
mempalace --palace <HOME>/.mempalace/palace_rebuilt palace set-embedder --model embeddinggemma
# then point config.json "palace_path" at palace_rebuilt and restart the MCP server
```
Result: rebuilt 269,491 rows; `count()` OK, `repair-status` **divergence 0**, end-to-end MCP `status`/`search`/`add_drawer` work again.

### Windows gotcha
In-place `repair --mode from-sqlite --archive-existing` fails with:
```
PermissionError: [WinError 32] ... data_level0.bin is used by another process
```
because the **running MCP server holds the HNSW segment files open**, so the archive-rename can't proceed. Either stop the MCP server first, or rebuild into a separate path from the live sqlite (read-only) as above.

## Suggested hardening
1. **Fail fast, don't hang.** Wrap collection-open / compaction with an operation deadline (and a SQLite `busy_timeout`) so a stuck compactor returns a structured error instead of hanging the whole MCP server forever.
2. **Surface the real cause.** Detect the `Failed to apply logs to the hnsw segment writer` compactor error and point the user to `repair --mode from-sqlite` instead of an opaque infinite hang.
3. **mine crash-safety.** Make queue/WAL application atomic or recoverable so an interrupted `mine` can't leave an unbackfillable segment; on startup, detect a dirty queue and offer a guarded rebuild.
4. **Lock hygiene.** `mine_palace_lock` (palace.py ~L945–984) never liveness-checks the recorded holder PID (no `os.kill(pid,0)` / lease) and leaks empty `locks/*.lock`. Add PID-liveness/lease + cleanup-on-release.
5. **Deeper `repair-status`.** Since it never opens the chroma client, it can't see a broken segment writer — consider an optional guarded `count()` probe so the broken state is detectable without a 6-hour investigation.

trifo-tech:
sqlite_vec Backend (Jun 27)

     New local storage backend powered by sqlite-vec — fast ANN vector search without external servers.

     - 1050 lines, full BaseBackend + BaseCollection contract
     - vec0 virtual table for O(log n) ANN cosine search
     - WAL mode + mmap_size=0 (WSL-safe), FTS5 lexical search
     - Auto-detects Python version for site-packages fallback
     - Per-collection vec0 tables, extension loading disabled after use
     - Migration tool: 1607 drawers, 11.5 MB (vs 55 MB ChromaDB)
     - 4/4 conformance tests pass
     MCP Server Refactoring (Jun 28)

     Replaced hardcoded ChromaDB references with pluggable backend system:

     - Conditional Chroma imports with ImportError fallback
     - _get_client: ChromaBackend.make_client() → get_backend_for_palace()
     - _palace_db_path(): backend-agnostic DB path helper
     - CollectionNotFoundError: backend-agnostic exception
     - All chroma.sqlite3 references → _palace_db_path()
     Reviews

     - Linus Torvalds 👓 (self-review): PASSED
     - Gemini Code Assist 🤖: 7 suggestions, all resolved
     - Claude Opus 4 : validated MCP refactor, all fixes applied
     - Ornith 35B (MIT): generated initial refactoring

     Fixes #1845, connected #1874 #1876

### DIFF
--- a/mempalace/backends/__init__.py
+++ b/mempalace/backends/__init__.py
@@ -40,6 +40,7 @@ from .milvus import MilvusBackend, MilvusCollection
 from .pgvector import PgVectorBackend, PgVectorCollection
 from .qdrant import QdrantBackend, QdrantCollection
 from .sqlite_exact import SQLiteExactBackend, SQLiteExactCollection
+from .sqlite_vec import SQLiteVecBackend, SQLiteVecCollection
 from .registry import (
     available_backends,
     detect_backend_for_path,
@@ -79,6 +80,8 @@ __all__ = [
     "QueryResult",
     "SQLiteExactBackend",
     "SQLiteExactCollection",
+    "SQLiteVecBackend",
+    "SQLiteVecCollection",
     "UnsupportedCapabilityError",
     "UnsupportedFilterError",
     "UnsupportedMaintenanceKindError",
--- a/mempalace/backends/registry.py
+++ b/mempalace/backends/registry.py
@@ -211,6 +211,7 @@ def _register_builtins() -> None:
     from .pgvector import PgVectorBackend
     from .qdrant import QdrantBackend
     from .sqlite_exact import SQLiteExactBackend
+    from .sqlite_vec import SQLiteVecBackend
 
     # Use setdefault semantics so a caller that pre-registered for tests wins.
     if "chroma" not in _registry:
@@ -221,6 +222,8 @@ def _register_builtins() -> None:
         _registry["qdrant"] = QdrantBackend
     if "sqlite_exact" not in _registry:
         _registry["sqlite_exact"] = SQLiteExactBackend
+    if "sqlite_vec" not in _registry:
+        _registry["sqlite_vec"] = SQLiteVecBackend
     if "pgvector" not in _registry:
         _registry["pgvector"] = PgVectorBackend
 
--- a/mempalace/backends/sqlite_vec.py
+++ b/mempalace/backends/sqlite_vec.py
@@ -1,0 +1,1180 @@
+"""SQLite-vec ANN backend for MemPalace.
+
+Fast approximate nearest-neighbour search via sqlite-vec (Alex Garcia).
+Uses vec0 virtual table with cosine distance — 0.003s search vs 2-5s
+for brute-force exact cosine. Ideal for WSL where ChromaDB HNSW
+regularly corrupts across restarts.
+
+Database file: sqlite_vec.sqlite3
+"""
+
+from __future__ import annotations
+
+import contextlib
+import json
+import logging
+import os
+import re
+import sqlite3
+import struct
+import sys
+import threading
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, ClassVar, Optional, Union
+
+import numpy as np
+
+from .base import (
+    BackendClosedError,
+    BaseBackend,
+    BaseCollection,
+    CollectionNotInitializedError,
+    DimensionMismatchError,
+    EmbedderIdentity,
+    GetResult,
+    HealthStatus,
+    LexicalHit,
+    LexicalResult,
+    MaintenanceResult,
+    PalaceNotFoundError,
+    PalaceRef,
+    QueryResult,
+    UnsupportedFilterError,
+    _IncludeSpec,
+)
+
+logger = logging.getLogger(__name__)
+
+_DB_FILENAME = "sqlite_vec.sqlite3"
+_TOKEN_RE = re.compile(r"\w{2,}", re.UNICODE)
+_SUPPORTED_OPERATORS = frozenset(
+    {"$eq", "$ne", "$in", "$nin", "$and", "$or", "$contains", "$gt", "$gte", "$lt", "$lte"}
+)
+_EMBEDDING_DIM = 384  # EmbeddingGemma ONNX
+
+
+# ── sqlite_vec loader ────────────────────────────────────────────────────
+
+
+def _load_sqlite_vec():
+    """Load sqlite_vec extension, trying multiple paths."""
+    try:
+        import sqlite_vec
+
+        return sqlite_vec
+    except ImportError:
+        pass
+    # Fallback: auto-detect mempalace site-packages
+    tools_base = Path.home() / ".local/share/uv/tools/mempalace/lib"
+    if tools_base.exists():
+        for py_dir in sorted(tools_base.iterdir(), reverse=True):
+            if py_dir.name.startswith("python"):
+                sp = str(py_dir / "site-packages")
+                if sp not in sys.path:
+                    sys.path.insert(0, sp)
+                try:
+                    import sqlite_vec
+
+                    return sqlite_vec
+                except ImportError:
+                    continue
+    raise ImportError(
+        "sqlite_vec is required for sqlite_vec backend. Install it with: uv pip install sqlite-vec"
+    )
+
+
+_sqlite_vec_module = None
+_has_vec = False
+
+
+def _ensure_sqlite_vec():
+    """Load sqlite_vec on first use, not at import time."""
+    global _sqlite_vec_module, _has_vec
+    if _sqlite_vec_module is not None:
+        return _sqlite_vec_module
+    _sqlite_vec_module = _load_sqlite_vec()
+    _has_vec = True
+    return _sqlite_vec_module
+
+
+# ── Helpers ──────────────────────────────────────────────────────────────
+
+
+def _utcnow() -> str:
+    return datetime.now(timezone.utc).isoformat()
+
+
+def _json_dumps(obj: Any) -> str:
+    return json.dumps(obj or {}, ensure_ascii=False, separators=(",", ":"), sort_keys=True)
+
+
+def _json_loads(text: str | None) -> dict:
+    if not text:
+        return {}
+    try:
+        value = json.loads(text)
+    except json.JSONDecodeError:
+        return {}
+    return value if isinstance(value, dict) else {}
+
+
+def _as_vector_array(vector: list[float]) -> np.ndarray:
+    arr = np.asarray(vector, dtype=np.float32)
+    if arr.ndim != 1 or arr.size == 0:
+        raise ValueError("embedding must be a non-empty 1D vector")
+    return arr
+
+
+def _encode_vector(vector: list[float]) -> bytes:
+    arr = _as_vector_array(vector)
+    return struct.pack(f"{len(arr)}f", *arr)
+
+
+def _decode_vector(blob: bytes | None) -> list[float]:
+    if not blob:
+        return []
+    return list(struct.unpack(f"{len(blob) // 4}f", blob))
+
+
+def _tokenize(text: str) -> list[str]:
+    if not text:
+        return []
+    return _TOKEN_RE.findall(text.lower())
+
+
+def _bm25_scores(query: str, documents: list[str], k1: float = 1.5, b: float = 0.75) -> list[float]:
+    """Simple BM25 scoring for lexical search fallback."""
+    query_terms = set(_tokenize(query))
+    if not query_terms:
+        return [0.0] * len(documents)
+    n_docs = len(documents)
+    avgdl = sum(len(_tokenize(d)) for d in documents) / max(1, n_docs)
+    scores = []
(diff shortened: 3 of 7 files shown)


igorls (changes requested):
Thanks for this — a corruption-resistant `vec0` backend is a genuinely useful direction, and the backend core (CRUD, ANN, FTS5 lexical, embedder-identity, maintenance) is well-structured and passes the conformance suite. There are a few blocking issues to resolve before it can land, though:

**1. `upsert` of an existing id crashes (breaks re-mine).** After the `ON CONFLICT` updates the `documents` row (rowid preserved), the follow-up `INSERT OR REPLACE INTO doc_vec_<coll> (rowid, embedding)` is rejected by the vec0 virtual table with `UNIQUE constraint failed`. Since `upsert` is the primary write path for `miner`, `convo_miner`, `diary_ingest`, `format_miner`, and `mcp_server`, re-running `mine` over existing content would crash — this violates the "incremental only / append-only re-ingest" design contract. The vec row needs to be `DELETE`d then `INSERT`ed rather than `INSERT OR REPLACE`d. (The conformance suite missed this because it only upserts brand-new ids — worth adding an "upsert existing id" case.)

**2. The `mcp_server.py` rewiring regresses the default Chroma backend (56 test failures).** `_get_client()` was changed from `ChromaBackend.make_client(...)` (a Chroma client) to `get_backend_for_palace(...)` (a backend object), but `_get_collection()` still calls `client.get_collection(_config.collection_name, ...)` positionally while the new backend `get_collection` is keyword-only → `TypeError`. Confirmed these tests pass on `develop` and fail on the branch, so they're introduced here, not pre-existing.

**3. `NameError: _ChromaNotFoundError`.** The import `from chromadb.errors import NotFoundError as _ChromaNotFoundError` was removed but a live `except _ChromaNotFoundError:` reference remains (~`mcp_server.py:1056`) — a guaranteed crash when that path is hit. (ruff flags this as F821.)

**4. `detect()` regresses the #1896 magic-header fix.** It uses bare `os.path.isfile(...)`; the sibling backends now read the 16-byte `b"SQLite format 3\x00"` header to reject 0-byte markers. The filename is distinct so there's no multi-backend collision, but please match the siblings' pattern. See the merged #1896.

**5. Optional-dependency declaration is missing.** `pyproject.toml` has no `sqlite_vec` extra and no `[project.entry-points."mempalace.backends"]` entry (the other four backends have both), and the loader falls back to a hardcoded author-specific site-packages path. A user selecting this backend has no declared way to install `sqlite-vec`. Please add the optional extra + entry point so it stays local-first and installable.

Minor: ruff `check`/`format` currently fail (F821, E731, W293, E402); the migration script has hardcoded personal paths; and a dedicated `tests/test_sqlite_vec_backend.py` (parity with `test_sqlite_exact_backend.py`) would have caught the upsert bug.

Happy to re-review once the write-path crash and the Chroma regression are sorted — those two are the critical ones.

trifo-tech:
Thanks for the thorough review, @igorls! All five issues resolved in 869cd66:

     ✅ 1. upsert crash — Changed INSERT OR REPLACE → DELETE + INSERT on vec0 table.

     ✅ 2. MCP regression — Reverted _get_client to ChromaBackend.make_client. No MCP refactoring needed
     — the existing non-Chroma path already routes to our backend.

     ✅ 3. NameError — Restored _ChromaNotFoundError import.

     ✅ 4. detect() header — Now reads 16-byte SQLite format 3 header.

     ✅ 5. pyproject.toml — Added entry point + optional dep.

     Tested: palace_get_collection with backend='sqlite_vec' works end-to-end. Backend conformance 4/4
     pass. Ready for re-review!

trifo-tech:
"Friendly ping @igorls — all five issues from your review have been resolved. Happy to rebase if
     anything has changed since. Thanks! "

igorls (changes requested):
Thanks for the ping. I re-reviewed the current head (`869cd66`) and the earlier Chroma/MCP regression does look fixed, and the optional extra/entry point now installs cleanly via `uv run --extra sqlite_vec`.

There are still a few blockers before this should land:

1. Filtered ANN queries can incorrectly return no results. `SQLiteVecCollection.query()` fetches only `max(n_results * 3, 10)` vec0 candidates, then applies `where` / `where_document` in Python. If the nearest candidates do not match the filter, matching rows just beyond that window are never considered. I reproduced this with 30 nearer non-matching rows and 10 farther matching rows: `query(..., n_results=5, where={"group": "b"})` returned `ids=[[]]` even though 10 matching documents existed. The backend needs to either push filters into candidate selection where possible or continue expanding the ANN candidate window until it can satisfy `n_results` or exhaust the collection.

2. `delete_collection()` has broken vec0 lifecycle handling. A collection created but never written has no `doc_vec_<collection>` table yet, so deleting it raises `OperationalError: no such table: doc_vec_empty_collection`. For a populated collection, deletion removes rows from the vec0 table but leaves the virtual table and `vec0_table_created:<collection>` metadata behind. Recreating the same collection with a different embedding dimension then fails with a sqlite-vec dimension mismatch because the old vec0 table is reused.

3. `delete_collection()` does not match the compatibility shape used by the other built-in backends and repair helpers. The existing call sites pass `(palace_path: str, collection_name)`, but sqlite_vec expects a `PalaceRef`, so `backend.delete_collection(str(path), "drawers")` fails with `AttributeError: 'str' object has no attribute 'local_path'`. It also lacks the `create_collection()` / `get_or_create_collection()` helpers exposed by the other in-tree backends.

Verification I ran:

- `uv run --extra sqlite_vec python -m pytest tests/test_backend_conformance.py tests/test_sqlite_exact_backend.py -q` -> `31 passed`
- Focused MCP cache tests -> `3 passed, 1 skipped`
- Direct sqlite_vec repro scripts for existing-id upsert, filtered ANN, and collection delete/recreate lifecycle
- `uv run --extra sqlite_vec ruff check mempalace/backends/sqlite_vec.py mempalace/mcp_server.py scripts/migrate-chroma-to-sqlite-vec.py tests/test_backend_conformance.py` still fails with lint issues
- `uv run --extra sqlite_vec ruff format --check ...` reports three files would be reformatted

So the previous big MCP issue is resolved, but the sqlite_vec collection/query lifecycle still needs another pass.

trifo-tech:
All three issues addressed in [63beea7](https://github.com/MemPalace/mempalace/pull/1886/commits/63beea7670566673d4c299ec14e387ab79d65a73):

1. **Filtered ANN** — expanding window loop, bounded by collection size, 20-attempt safety limit
2. **delete_collection** — handles empty collections, cleans vec0 metadata, drops table for clean re-create
3. **API compatibility** — accepts PalaceRef/str, added create_collection and get_or_create_collection helpers

- 31/31 backend conformance ✅
- 21/21 targeted verification ✅  
- ruff lint+format clean ✅

Ready for re-review.

trifo-tech:
Rebased onto latest upstream/develop (v3.6.0). 

- Resolved pyproject.toml conflict (Milvus + sqlite_vec both kept)
- All 166/167 backend conformance tests pass (1 failure is pre-existing ChromaDB embedding mismatch in test environment, not caused by sqlite_vec)
- Clean diff against upstream

Ready for re-review whenever you have time. Thanks! 🦋

trifo-tech:
Hi @igorls @milla-jovovich — just a gentle nudge from a production user. 👋

We've been running MemPalace with the ChromaDB backend at ~3,900 drawers for over a month. The HNSW divergence hits us **daily — sometimes 2-3 times per day.** Every time, vector search silently dies, and we have to run a full `repair --mode from-sqlite` rebuild (5-7 minutes of downtime).

We've added `flock`-based write locks on all 14 cron jobs, and we'll try `sync_threshold=1` as an extra precaution. But these are band-aids — the root cause is ChromaDB's async HNSW flush, which this `sqlite_vec` backend would eliminate entirely.

It's been about a month since the last review, and we'd really appreciate any update — even a quick "we're swamped, will get to it" or "we've decided to go a different direction." Either way, knowing where we stand would help us plan. If there's anything we can do — testing, rebasing, tackling the remaining review items — we're more than happy to help move this forward.

Thanks for building such a great project. 🙏

trifo-tech:
Rebased onto upstream `develop` (MemPalace 3.7.1 + RFC 001) and re-aligned the backend to the merged storage-backend contract:

- `SQLiteVecBackend` / `SQLiteVecCollection` now implement `BaseBackend` / `BaseCollection` with `PalaceRef`, typed `QueryResult` / `GetResult`, and capability tokens (`supports_embeddings_passthrough`, `supports_embeddings_out`, `supports_metadata_filters`, `supports_lexical_search`, `local_mode`, …).
- `require_namespace_support()` honored per RFC 001 §4.4 (no silent namespace drop).
- Registered via entry point + `sqlite_vec = ["sqlite-vec>=0.1.9"]` optional dependency.
- Added to `tests/test_backend_conformance.py` (`_LOCAL_BACKENDS`); conformance suite passes locally (7 passed).
- Squashed the earlier WIP commits into one clean commit; PR is now MERGEABLE.

Happy to iterate on review feedback.

Issue #1889, opened 2026-06-28
Conversation drawers lose chronology: authored date discarded at ingestion


JosefAschauer:
**What problem does this solve?**

Conversation drawers carry only `filed_at` (ingest time) as their date. A bulk
`mine` therefore stamps every drawer with ~the same timestamp, collapsing the real
chronology of the underlying sessions into a single instant. Search ranks purely by
embedding similarity with no recency dimension, and the recency-window fallback
orders by ingest order — so after a re-mine "recent" no longer means "recently
written."

Concrete failure mode: a query about the current state of some work surfaced a
high-similarity chunk from a session several days old (describing a problem that was
later fixed), with nothing to signal it was stale. The chronological signal needed to
disambiguate "was true then" from "is true now" had been discarded at ingestion —
even though every Claude Code / Codex JSONL line carries an ISO-8601 `timestamp`.

In short: memory has chronology; the convos miner was throwing it away.

**What's the proposed solution?**

Two parts. The first is a small, benchmark-neutral foundation (PR opened, see below);
the second is a behavioral change that needs maintainer input and benchmark runs.

1. **Recover + surface the authored date (foundation).**
   - Derive `authored_at` from the max per-line `timestamp` in each transcript and
     store it as drawer metadata (falls back to `filed_at` when unavailable).
   - Surface `authored_at` in search results so consumers can see when content was
     actually written vs. ingested.
   - Break *exact* hybrid-score ties toward the more recently authored drawer
     (benchmark-neutral — only reorders exact ties).

2. **Chronology-aware retrieval (for discussion).**
   - An optional, weighted recency prior in `_hybrid_rank` (decay on `authored_at`),
     validated against the longmemeval benchmark so it doesn't regress relevance.
   - Make the recency-window fallback order by `authored_at` instead of ingest order.
   - Separate "narrative recall" (raw transcript chunks — inherently point-in-time)
     from "current facts" (the KG layer with `kg_invalidate`/`kg_add`), so
     current-state queries can prefer non-superseded facts. The supersedence case
     above is not fully solved by recency alone; the KG layer already models it but
     the transcript-RAG layer can outrank it.

**Alternatives considered:**

- *Datetime metadata alone, no ranking change* — necessary but insufficient: it makes
  dates visible but doesn't change what surfaces first.
- *Rebuild memory on a full RAG engine* — far heavier ingestion/latency for agent
  memory and still doesn't solve supersedence out of the box; the gap is a retrieval
[...]

PR #1890, opened 2026-06-28
feat(convo): preserve authored timestamp from transcripts

JosefAschauer:
## What

Recover and surface the **authored date** of conversation drawers, which was being discarded
at ingestion.

Conversation drawers only carried `filed_at` (ingest time), so a bulk `mine` collapsed every
drawer to a single instant and the chronological signal was lost — even though each Claude
Code / Codex JSONL line already carries an ISO-8601 `timestamp`. The recency-window fallback
and any date-aware consumer then saw ingest order, not when content was written.

Foundation for #1889. The broader chronology-aware retrieval (weighted recency prior,
recency-window ordering by `authored_at`) remains for discussion there.

### `created_at` vs `authored_at`

| Field | Meaning |
|-------|---------|
| `filed_at` / result `created_at` | When the drawer was **mined** (written to the palace). |
| `authored_at` | When the content was **written**, from transcript timestamps. Survives re-mining. |

`created_at` is intentionally left as-is (renaming would be a breaking change); `authored_at`
is added as a peer. Documented in `docs/authored-at.md`.

## Changes

- **`convo_miner`**: `_extract_authored_at()` derives the per-file max line `timestamp` and
  stores it as `authored_at` metadata; falls back to `filed_at` when absent.
- **`searcher`**: surfaces `authored_at` in search results, and breaks **exact**
  hybrid-score ties toward the more recently authored drawer (ISO strings sort
  chronologically; missing dates sort oldest). Only exact ties are reordered → benchmark-neutral.
- **`layers`**: CLI `search` output shows an `authored:` date line per result, so the field
  is visible end-to-end (not just in the result dict / MCP output).
- **`scripts/backfill_authored_at.py`**: in-place migration to adopt `authored_at` on
  existing palaces — metadata only (no re-embedding), idempotent, dry-run by default. Needed
  because re-mining skips files already mined at the current `NORMALIZE_VERSION`.
- **`docs/authored-at.md`**: field semantics + backfill paths.
- **tests**: `_extract_authored_at`, the tie-break, and backfill integration tests.

## Why it's safe

- `authored_at` is additive metadata; drawers without it fall back to `filed_at` at read
  time — no migration required for correctness.
- The tie-break only affects candidates with identical blended scores, so it cannot change
  relevance ordering or benchmark results.
- The backfill updates metadata only; embeddings are untouched and it is idempotent.

## Testing

```
pytest tests/test_backfill_authored_at.py tests/test_convo_miner_unit.py \
       tests/test_hybrid_search.py tests/test_searcher.py tests/test_layers.py
# 147 passed
ruff check   # clean on changed files
```

Refs #1889

### DIFF
--- a/docs/authored-at.md
+++ b/docs/authored-at.md
@@ -1,0 +1,45 @@
+# Authored date (`authored_at`)
+
+Conversation transcripts carry a per-line ISO-8601 `timestamp` (both Claude Code and
+Codex JSONL). The miner records the most recent one per file as the drawer's
+**`authored_at`** — when the content was actually written.
+
+This is distinct from the ingest date:
+
+| Field | Meaning |
+|-------|---------|
+| `filed_at` / result `created_at` | When the drawer was **mined** (written to the palace). A bulk re-mine collapses these to a single instant. |
+| `authored_at` | When the underlying content was **written**, recovered from the transcript timestamps. Survives re-mining. |
+
+`authored_at` is surfaced in search results (and shown in the CLI `search` output), and is
+used as a deterministic tie-break in hybrid ranking: candidates with identical scores order
+with the more recently authored drawer first. Drawers without per-line timestamps (e.g.
+markdown) fall back to `filed_at`.
+
+## Backfilling existing memory
+
+New mines populate `authored_at` automatically. Drawers mined before this feature only have
+`filed_at`. Re-mining does **not** fix them — the scanner skips files already mined at the
+current `NORMALIZE_VERSION`. Two options:
+
+1. **In-place backfill (recommended — no re-embedding).** `scripts/backfill_authored_at.py`
+   reads each convos drawer's source transcript and updates only the `authored_at` metadata.
+   Idempotent and safe to re-run; embeddings are untouched.
+
+   ```bash
+   python scripts/backfill_authored_at.py \
+       --palace ~/.mempalace/palace \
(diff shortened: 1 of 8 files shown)


JosefAschauer:
Thanks for the review — both findings were valid and are fixed in 57e4d9d:

1. **`_extract_authored_at` TypeError** — now guards `isinstance(ts, str)` before comparing, so a non-string `timestamp` on a malformed/foreign JSONL line is skipped instead of crashing the mine.
2. **Tie-break missed the top-level shape** — good catch. The `search_memories` path (MCP / Claude Code) carries `authored_at` at the top level of each hit, while the candidate-union path nests it under `metadata`; the key now reads both, so the tie-break is no longer a no-op on the programmatic path. My original test only covered the nested shape — added a top-level-shape test that fails before this fix.

150 tests pass, ruff clean.

igorls (approved):
Reviewed and verified locally against current develop: 150 tests pass (convo_miner, searcher, hybrid, layers, backfill). The searcher tie-break is genuinely benchmark-neutral — authored_at is only a secondary sort key, so it reorders strictly within exact score ties. Additive metadata with filed_at fallback, idempotent metadata-only backfill, and docs. Solid foundation for #1889. (Pushed a ruff format fix for the 4 files CI flagged.) Thanks for the careful, well-documented work!

PR #1895, opened 2026-06-28
feat(graph): auto-populate the associative graph from mined sessions

### LINKED ISSUE #1894: Conversation mining leaves the graph layers starved (no entities populated)
**What problem does this solve?**

MemPalace ships two graph layers — the temporal knowledge graph (`kg_add`/`kg_query`/
`kg_invalidate`) and the associative hallway/tunnel graph — but **conversation mining never
populates them**. `compute_hallways_for_wing` consumes an `entities` drawer-metadata field
(`hallways._parse_entities`), and the convos miner never sets it. So for the most common
ingestion path — mined CLI sessions — the associative graph is **empty**, and everything
built on it (entity navigation, cross-wing tunnel recommendations) has nothing to work with.
The graph layers are well-designed but **starved**.

This matters because of a failure mode that pure semantic recall can't fix on its own:

- Search ranks by embedding similarity. With no associative or chronological structure, a
  high-similarity chunk describing something that was **later changed/superseded** outranks
  the current state, with nothing to signal it's stale. (We hit exactly this: a query about
  the current state of some work surfaced a days-old chunk describing a problem that had
  since been fixed, presented as if current.)
- The KG models supersedence (temporal validity + `invalidate`) and hallways model
  association — both are the right tools — but if nothing populates them at ingest, the raw
  semantic layer is all retrieval has, and it can't tell "was true then" from "is true now."

In short: the structure needed to disambiguate current vs. superseded, and to navigate by
related concepts, already exists — it's just never fed from mined sessions.

**What's the proposed solution?**

Auto-populate the associative graph at mine time with a **no-LLM, structural** entity
extractor (PR linked below), respecting MemPalace's lean, zero-service, benchmark-honest
ethos:

- Extract structural tokens already unambiguous in dev/CLI transcripts — author-quoted code
  spans, URLs, file paths, qualified identifiers, CamelCase/snake_case symbols. No wordlists,
  no NLP model, precision-biased so prose doesn't pollute the graph.
- Write them to the existing `entities` metadata field and trigger `compute_hallways_for_wing`
  after a convos mine (the project-file path already does this) — so mined sessions finally
  build hallways.
- Backend-safe by construction: this is drawer metadata + the existing SQLite/JSON graph
  stores; it does not touch the vector-backend abstraction (chroma/sqlite_exact/qdrant/pgvector).
- Expose it on both surfaces: hallways are already queryable via the `list_hallways` MCP tool;
  add a `mempalace hallways` CLI command for parity.

**Alternatives considered:**

- *Opt-in LLM extraction for true KG triples (facts).* Higher quality and would populate the
  temporal KG directly, but needs an LLM call per session — against the zero-cloud default.
  Worth a separate, off-by-default follow-up; this issue/PR deliberately scopes to the no-LLM
  associative graph.
- *Retrieval routing only* (prefer KG/hallways for "current state" queries). Valuable, but
  pointless until the graph is populated — this is the prerequisite.
- *Heavier NLP (spaCy NER).* Better recall but adds a heavy dependency that cuts against the
  lean ethos.

JosefAschauer:
## What

Auto-populate the associative (hallway) graph from mined conversations.

`compute_hallways_for_wing` consumes an `entities` drawer-metadata field, but the convos
miner never set it — so mined sessions produced an **empty** associative graph and starved
the entity-navigation / tunnel features built on top. See #1894 for the product gap (and the
stale-vs-superseded failure mode that motivates feeding MemPalace's graph layers).

## Changes

- **`mempalace/entities.py`** — a no-LLM **structural** entity extractor: author-quoted code
  spans, URLs, file paths, qualified dotted identifiers, CamelCase and snake_case symbols.
  No wordlists, no NLP model; precision-biased (prose yields nothing). Output is the
  `;`-joined form `hallways._parse_entities` expects.
- **`mempalace/convo_miner.py`** — set `entities` metadata per chunk, and run
  `compute_hallways_for_wing` after a convos mine (mirroring the project-file path).
  Hallways are computed *before* `_validate_palace_fts5_after_mine`, which opens a direct
  sqlite connection to the Chroma DB that can invalidate the live collection handle on some
  Chroma builds (observed `'RustBindingsAPI' object has no attribute 'bindings'` on
  chromadb 1.5.9). Best-effort: a hallway failure never breaks a mine.
- **`mempalace/cli.py`** — `mempalace hallways [--wing] [--limit]` lists the associative
  graph, for CLI parity with the existing `list_hallways` MCP tool.
- **tests** — extractor precision/ranking/dedup, `entities` populated at mine time, and the
  CLI command.

## Backend safety

No vector-backend code is touched. This is drawer metadata plus the existing hallway store;
it works identically across chroma / sqlite_exact / qdrant / pgvector.

## Verified end-to-end

Mining 3 sessions that share entities produces drawers with populated `entities` and a
15-edge hallway graph (`MemoryStack ↔ ChromaBackend`, `ChromaBackend ↔ search_memories`, …),
listed by `mempalace hallways`. Sessions previously produced zero entities/hallways.

```
pytest tests/test_entities.py tests/test_cli_hallways.py tests/test_convo_miner_unit.py \
       tests/test_convo_miner.py tests/test_hallways.py    # 94 passed
ruff check   # clean
```

## Scope / follow-ups

Deliberately scoped to the **no-LLM associative graph**. Populating the temporal **KG**
(subject–predicate–object facts) wants an opt-in, off-by-default LLM extractor — a natural
follow-up discussed in #1894.

Closes #1894

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1001,6 +1001,21 @@ def cmd_migrate_wings(args):
     )
 
 
+def cmd_hallways(args):
+    """List within-wing entity hallways (the auto-built associative graph)."""
+    from .hallways import list_hallways
+
+    rows = list_hallways(getattr(args, "wing", None))
+    if not rows:
+        print("No hallways yet — they are built from drawer entities when you mine.")
+        return
+    rows.sort(key=lambda h: h.get("co_occurrence_count", 0), reverse=True)
+    print(f"  {len(rows)} hallway(s):")
+    for h in rows[: args.limit]:
+        label = h.get("label") or f"{h.get('entity_a', '?')} <-> {h.get('entity_b', '?')}"
+        print(f"    {label}")
+
+
 def cmd_status(args):
     from .miner import status
 
@@ -1977,6 +1992,9 @@ def main():
     )
     p_migrate_wings.add_argument("--yes", action="store_true", help="Skip the confirmation prompt")
 
+    p_hallways = sub.add_parser("hallways", help="List entity hallways (associative graph)")
+    p_hallways.add_argument("--wing", default=None, help="Filter to one wing")
+    p_hallways.add_argument("--limit", type=int, default=50, help="Max hallways to show")
     p_status = sub.add_parser("status", help="Show what's been filed")
     p_status.add_argument(
         "--backend",
@@ -2061,6 +2079,7 @@ def main():
         "repair-status": cmd_repair_status,
         "migrate": cmd_migrate,
         "migrate-wings": cmd_migrate_wings,
+        "hallways": cmd_hallways,
         "status": cmd_status,
     }
     dispatch[args.command](args)
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1011,7 +1011,7 @@ def cmd_hallways(args):
         return
     rows.sort(key=lambda h: h.get("co_occurrence_count", 0), reverse=True)
     print(f"  {len(rows)} hallway(s):")
-    for h in rows[: args.limit]:
+    for h in rows[: max(0, args.limit)]:
         label = h.get("label") or f"{h.get('entity_a', '?')} <-> {h.get('entity_b', '?')}"
         print(f"    {label}")
 
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -20,6 +20,7 @@ from typing import Optional
 from .collision_scan import assert_no_collisions
 from .ids import ID_RECIPE, make_convo_drawer_id, make_convo_sentinel_id
 from .normalize import normalize
+from .entities import entities_metadata
 from .palace import (
     NORMALIZE_VERSION,
     SKIP_DIRS,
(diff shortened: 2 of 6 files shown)


JosefAschauer:
Thanks — all three addressed in 2f05532:

1. **Semicolon in entity** (high): `_clean` now strips `;` out of tokens, so a URL query string or backtick code span can't split the `;`-joined `entities` field. Added a test.
2. **Leading-underscore snake_case**: `_SNAKE` now allows an optional leading/trailing `_?`, so `_extract_authored_at` etc. are matched in plain text (previously only caught when quoted in backticks). Added a plain-text test.
3. **Negative `--limit`**: clamped with `max(0, ...)` so a negative value shows nothing instead of slicing from the end. Added a test.

53 tests pass, ruff clean.

igorls (approved):
Verified locally against current develop: 82 targeted + 291 related tests pass, ruff clean. Purely additive, local-only (regex extractor, no model/network), idempotent on re-mine (symmetric hallway IDs, preserved L7 fields), crash-safe (atomic temp-file persist, fail-soft skip), and batch-level rather than per-hook so the perf budgets don't apply. Per maintainer call, the structural code-symbol entities are accepted as a complementary signal alongside the project miner's nominal entities. I pushed the ruff format fix and resolved the trivial additive conflicts with the merged authored_at work. Thanks!

PR #1902, opened 2026-06-29
fix(mcp): crash logging + result-size caps; allow long kg_invalidate objects

jonyskids:
Two small robustness fixes to the MCP server, both independent.

**1. Harden the MCP server against silent death (`mcp_server.py`)**
The server could exit on an unhandled exception with no record of why. This adds:
- `sys.excepthook` that logs the full traceback before the process dies
- upgraded the per-request catch from a one-line `logger.error` to `logger.exception` (full stack)
- a default logfile at `~/.mempalace/mcp_server.log` — attached **only if that directory already exists**, so it never creates the palace root on import (respects the kill-switch invariant)

**2. Cap oversized query results (`mcp_server.py`)**
`kg_query`/`kg_timeline` could return very large payloads (observed ~675K chars) that overflow a caller's context. This caps results (default 100 facts) with an explicit `truncated`/`total`/`notice`, plus a universal response-size backstop (~50KB). All thresholds and the logfile path are env-overridable (`MEMPALACE_KG_RESULT_CAP`, `MEMPALACE_MAX_RESPONSE_CHARS`, `MEMPALACE_LOG_FILE`). Truncation is never silent.

**3. Allow long object strings in `tool_kg_invalidate`**
Removes an overly tight length limit that rejected legitimate long fact objects.

No KG/diary data or semantics changed. `ruff check` clean; MCP test suite passes (200).

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -4070,6 +4070,74 @@ except (ValueError, TypeError):
     MAX_RESPONSE_CHARS = 50000
 
 
+def _fit_response(result, *, tool_name: str) -> tuple[object, str]:
+    """Return ``(payload, serialized_text)`` guaranteed to fit ``MAX_RESPONSE_CHARS``.
+
+    Degrades *structurally*, never textually. Slicing already-serialized JSON at
+    a character offset hands the caller text that no longer parses — a worse
+    failure for a program calling ``json.loads`` than for a human reading it,
+    and it can cut a drawer's stored text mid-word. Instead:
+
+    * dict payloads carrying a list: drop whole items off the end of the largest
+      list until the payload fits, reporting the omission in the same
+      ``total`` / ``truncated`` / ``notice`` shape ``_cap_facts`` already uses;
+    * anything else (or a payload still too large once the list is exhausted):
+      replaced with a structured "response too large" object.
+
+    Either way the response parses as valid JSON at every size.
+    """
+    text = json.dumps(result, indent=2, ensure_ascii=False)
+    if len(text) <= MAX_RESPONSE_CHARS:
+        return result, text
+
+    original_chars = len(text)
+
+    if isinstance(result, dict):
+        lists = [(k, v) for k, v in result.items() if isinstance(v, list) and v]
+        if lists:
+            key, items = max(lists, key=lambda kv: len(kv[1]))
+            # Serialized length grows monotonically with the item count, so the
+            # largest prefix that still fits can be found by bisection.
+            best: Optional[tuple[dict, str]] = None
+            lo, hi = 1, len(items)
+            while lo <= hi:
+                mid = (lo + hi) // 2
+                trial = dict(result)
+                trial[key] = items[:mid]
+                # Keep a sibling count field consistent with what actually ships
+                # rather than leaving it describing the untruncated list.
+                if isinstance(trial.get("count"), int) and trial["count"] == len(items):
+                    trial["count"] = mid
+                trial["total"] = len(items)
+                trial["truncated"] = True
+                trial["notice"] = (
+                    f"{len(items) - mid} more item(s) omitted (showing first {mid} of "
+                    f"{len(items)}); the full response was {original_chars} chars, over "
+                    f"the {MAX_RESPONSE_CHARS}-char cap — narrow your query"
+                )
+                trial_text = json.dumps(trial, indent=2, ensure_ascii=False)
+                if len(trial_text) <= MAX_RESPONSE_CHARS:
+                    best = (trial, trial_text)
+                    lo = mid + 1
+                else:
+                    hi = mid - 1
+            if best is not None:
+                return best
+
+    payload = {
+        "error": "response_too_large",
+        "tool": tool_name,
+        "size_chars": original_chars,
+        "cap_chars": MAX_RESPONSE_CHARS,
+        "truncated": True,
+        "notice": (
+            f"Response was {original_chars} chars, over the {MAX_RESPONSE_CHARS}-char "
+            "cap, and could not be reduced by dropping whole items. Narrow your query."
+        ),
+    }
+    return payload, json.dumps(payload, indent=2, ensure_ascii=False)
+
+
 def _cap_facts(results: list, *, narrow_hint: str) -> tuple[list, Optional[str]]:
     """Cap a fact list to ``KG_RESULT_CAP``; return (kept, truncation_notice).
 
@@ -7004,18 +7072,14 @@ def handle_request(request):
                 result = _decorate_mcp_tool_result(
                     tool_name, TOOLS[tool_name]["handler"](**tool_args)
                 )
-            text = json.dumps(result, indent=2, ensure_ascii=False)
             # Backstop against any tool returning an oversized payload that
             # would overflow the caller's context (one real incident: a broad
             # kg_query returned ~675K chars). Per-tool caps (e.g. _cap_facts)
             # handle the common cases gracefully; this is the last line of
-            # defense for everything else. Truncation is announced, not silent.
-            if len(text) > MAX_RESPONSE_CHARS:
-                text = (
-                    text[:MAX_RESPONSE_CHARS]
-                    + f"\n\n... [response truncated: {len(text)} chars exceeded the "
-                    + f"{MAX_RESPONSE_CHARS}-char cap; narrow your query]"
-                )
+            # defense for everything else. _fit_response drops whole items or
+            # substitutes a structured object, so the result stays valid JSON
+            # at every size, and truncation is announced rather than silent.
+            result, text = _fit_response(result, tool_name=tool_name)
             return {
                 "jsonrpc": "2.0",
                 "id": req_id,
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -193,8 +193,22 @@ def _init_logging() -> None:
     # MEMPALACE_LOG_FILE is operator-supplied and opt-in; this is a
     # local-first server (CLAUDE.md design principle), so no path
     # sanitization — the operator's process UID is the trust boundary.
+    # When unset, default to ~/.mempalace/mcp_server.log so an unhandled
+    # crash leaves a persistent record on disk — the MCP client swallows
+    # stderr, so stderr-only is a silent-death black hole.
+    #
+    # Critical: only attach the default file handler if ~/.mempalace ALREADY
+    # exists. The module must never create the palace root on import — a
+    # missing ~/.mempalace is the documented kill-switch (#1676;
+    # hooks_cli._palace_root_exists()) for disabling autosave/mining hooks,
+    # and recreating it would silently re-arm them. When the dir is absent we
+    # stay stderr-only, which matches the kill-switch intent.
     log_file = os.environ.get("MEMPALACE_LOG_FILE", "").strip()
     file_handler: logging.Handler | None = None
+    if not log_file:
+        default_log_dir = Path(os.path.expanduser("~/.mempalace"))
+        if default_log_dir.is_dir():
+            log_file = str(default_log_dir / "mcp_server.log")
     file_handler_error: Exception | None = None
     if log_file:
         try:
@@ -237,6 +251,27 @@ def _init_logging() -> None:
 
 _init_logging()
 logger = logging.getLogger("mempalace_mcp")
+
+
+def _log_uncaught(exc_type, exc_value, exc_tb) -> None:
+    """sys.excepthook: log the full traceback before the process dies.
+
+    The MCP client does not surface a crashed server's stderr, so an
+    unhandled exception escaping ``main()`` (or raised during the import /
+    startup path) otherwise leaves no record — the silent-death failure
+    mode. Routing it through ``logger`` ensures the persistent logfile
+    (see ``_init_logging``) captures a full stack trace. KeyboardInterrupt
+    is delegated to the default hook so Ctrl-C stays clean.
+    """
+    if issubclass(exc_type, KeyboardInterrupt):
+        sys.__excepthook__(exc_type, exc_value, exc_tb)
+        return
+    logger.critical(
+        "Uncaught exception — server is exiting", exc_info=(exc_type, exc_value, exc_tb)
+    )
+
+
+sys.excepthook = _log_uncaught
 
 
 def _get_result_ids(result) -> list:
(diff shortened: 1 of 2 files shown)


jonyskids:
Assessment of the three points, verified against the code:

**1. Potential `TypeError` in `tool_kg_query`/`tool_kg_timeline` if results aren't lists** — already guarded. `_cap_facts` opens with `if not isinstance(results, list) or len(results) <= KG_RESULT_CAP: return results, None`, so a non-list value is returned unchanged rather than raising. No change needed here.

**2. `threading.excepthook` for background threads** — valid. The server runs a daemon thread (`mcp-idle-watchdog`), and `sys.excepthook` does not cover exceptions raised in threads, so a crash there would be silent — the exact failure mode this PR targets. Will add `threading.excepthook`.

**3. `BrokenPipeError`/`ConnectionResetError` on stdout → infinite logging loop** — partially. Logging is attached to `StreamHandler(sys.stderr)` while the MCP protocol writes to stdout, so a broken stdout pipe will not recurse through stderr logging — there's no infinite loop in this architecture. The valid part: a client disconnect currently surfaces as a full-traceback "crash" rather than a clean shutdown. Will wrap the stdout write to exit quietly on `BrokenPipeError`/`ConnectionResetError`.

#2 and the clean-disconnect handling from #3 will follow in a commit on this branch.

igorls:

jonyskids:
Rebased onto current `develop` — merges cleanly now.

One thing worth flagging from the rebase: `_run_stdio_loop` has been substantially rewritten since this was opened (EOF / `OSError` / `BrokenPipeError` handling, write-stall watchdog). Rather than replay my version of the loop, I took `develop`'s wholesale and kept only the surviving delta — `logger.error(f"Server error: {e}")` → `logger.exception(...)`, so the per-request catch records a full traceback rather than a one-line message.

I also dropped the outer fatal-wrap `try/except` that was in the original commit. Keeping it would have meant re-indenting all of the new EOF/broken-pipe handling, and the `sys.excepthook` added in this same commit already logs a full traceback for anything escaping the loop. Same coverage, smaller diff against `develop`.

The other two pieces rebased unchanged: the default `~/.mempalace/mcp_server.log` handler (still gated on the directory already existing, per the #1676 kill-switch) and the `KG_RESULT_CAP` / `MAX_RESPONSE_CHARS` caps at the dispatch chokepoint.

Local test runs on the rebased branch:

- `tests/test_mcp_server.py` — 370 passed
- `tests/test_mcp_server_eof.py` — 2 passed
- `-k kg` across `tests/` — 75 passed

CI hasn't triggered — looks like it needs maintainer approval to run workflows on a fork PR.

Also happy to split this into two PRs if that's easier to review — the `kg_invalidate` fix and the MCP hardening are independent, and the first one is a three-line change.

igorls:
Thanks for rebasing, and for flagging that `_run_stdio_loop` had been substantially rewritten underneath this — that's the kind of note that saves a reviewer an hour.

Most of this I'm happy with. Two parts I want to call out as good rather than just wave through:

- Defaulting the logfile to `~/.mempalace/mcp_server.log` **only when that directory already exists**, so the module never creates the palace root on import, is precisely right. A missing `~/.mempalace` is the documented kill-switch, and quietly recreating it would re-arm autosave for someone who deliberately turned it off. Thank you for reasoning that through rather than just calling `mkdir(parents=True)`.
- `_cap_facts` degrades the way I'd want: it drops whole facts and surfaces `truncated` / `total` / `notice`, so every fact that does come back is still intact and the caller is told what it isn't seeing.

My one real note is the `MAX_RESPONSE_CHARS` backstop at the dispatch chokepoint:

```python
text = text[:MAX_RESPONSE_CHARS] + f"\n\n... [response truncated: ...]"
```

Because it slices the already-serialized JSON at an arbitrary character offset, what reaches the client is text that no longer parses. Any caller doing `json.loads` on the tool result gets a decode error rather than a smaller result, so the failure mode for a program is worse than the one for a human reading it — and it can also cut a drawer's stored text mid-word, which is the one thing this project tries hardest never to do.

I don't want to lose the backstop; a 675K-char response overflowing someone's context is a genuinely worse outcome, and you're right that the per-tool caps can't cover everything. What I'd prefer is that it degrade structurally instead of textually — drop or elide whole items until the payload fits and report how many went, in the same shape `_cap_facts` already uses, so the response stays valid JSON at every size. If that's awkward at the chokepoint because the result is opaque by then, returning a structured "response too large, narrow your query" object in place of the payload would also be fine, and is a smaller change.

Everything else here looks good to me.

jonyskids:
You're right, and the JSON-validity argument is the one that settles it — slicing a serialized payload optimises for the human reading the tail and breaks the program calling `json.loads`, which is exactly backwards.

Pushed `021a060`, which replaces the slice with `_fit_response()`. It takes the structural route you preferred, with the structured object as the fallback for when the payload is genuinely opaque:

- **dict payloads carrying a list** — bisect for the largest prefix of the biggest list that still fits, and report the omission in the same `total` / `truncated` / `notice` shape `_cap_facts` already uses. Bisection is safe here because serialized length grows monotonically with the item count.
- **anything else**, or a payload still over the cap once the list is exhausted — replaced with `{"error": "response_too_large", "tool": ..., "size_chars": ..., "cap_chars": ..., "truncated": true, "notice": ...}`.

Either way the response parses at every size, and every item that ships is byte-identical to what the handler produced.

One thing I added beyond your note, so it doesn't slip through unflagged: when the payload has a sibling `count` field matching the untruncated list length, it's updated to match what actually ships. Otherwise `kg_query` would come back with `count: 200` next to 40 facts — internally inconsistent in a way that's arguably worse than the original bug, since nothing in the payload contradicts it. If you'd rather the chokepoint stayed strictly shape-agnostic and left `count` alone, easy to drop.

New tests in `TestOversizedResponseBackstop` (9): pass-through under the cap, valid JSON over it, surviving items being byte-identical rather than sliced, `count`/`facts` agreement, the unshrinkable-payload path, drawer text never cut mid-value, non-dict payloads, and an end-to-end assertion that the dispatch chokepoint's `text` field always `json.loads`.

Local runs on the branch: 381 passed across `tests/test_mcp_server.py` + `tests/test_mcp_server_eof.py`, 75 passed for `-k kg`.

Left as a separate commit so the delta against your review is visible — happy for it to be squashed on merge.

jonyskids:
CI caught two things on the rebased branch, both mine. Fixed in `4c35201`.

**`ruff format`** — a test signature from the original commit was over the line limit. Reformatted; `ruff format --check` and `ruff check` are both clean now.

**Three failures in `TestColdStartDiagnostics`** — `test_log_file_unset_attaches_only_stream_handler`, `test_log_file_empty_string_attaches_only_stream_handler`, and `test_host_root_logger_config_survives_import`, all asserting `StreamHandler` only and all getting `StreamHandler,FileHandler`.

The cause is worth spelling out because it isn't obvious from the diff. `_run_main` inherits the session environment, and `conftest` redirects `HOME` to a session temp dir. The default-logfile handler attaches only when `~/.mempalace` already exists — so as soon as any earlier test in the suite creates the palace root under that temp `HOME`, those three start seeing a second handler. They pass in isolation and fail in a full run. They were latently dependent on ambient `HOME`; this PR's change is what made it observable.

I fixed the dependency rather than the assertion. `_clean_home_env(tmp_path, with_palace=...)` pins `HOME` — plus `USERPROFILE` / `HOMEDRIVE` / `HOMEPATH` so Windows `expanduser` follows — to a directory whose `.mempalace` state is explicit. The three cases now run against a `HOME` with no palace root, which is the condition they were always meant to describe.

Two tests added for branches nothing covered:

- `test_log_file_unset_attaches_file_handler_when_palace_exists` — the other half: an existing `~/.mempalace` does get the default log, and it lands inside the palace root.
- `test_log_file_unset_never_creates_the_palace_root` — the #1676 kill-switch. A missing `~/.mempalace` has to survive importing the server, or the default-logfile path would silently re-arm hooks the operator deliberately disabled. That reasoning was only a comment before; it's enforced now.

Full suite locally: **4313 passed, 31 skipped**, with `--ignore=tests/benchmarks` to match the workflow. I'd been verifying against `tests/test_mcp_server.py` alone, which is exactly why the order-dependence got past me — running the whole suite is what surfaces it.

One note for whenever the 3.9 job gets re-run: its earlier failure was the same three assertions, not a version issue, and coverage came in at 82.14%.

jonyskids:
Checking in on this one. CI still hasn't run on `4c35201` — fork PRs need a maintainer to approve the workflow, and there are no check runs recorded against that sha, so the branch has been sitting since 16 Aug waiting on that rather than on anything from my side.

`develop` has moved a fair way since the 15 Aug rebase (3.8.0 and 3.9.0 both shipped, and `mcp_server.py` has taken ~44 commits). Happy to rebase again onto current `develop` — say the word and I'll push it, or if you'd rather review against the branch as it stands, that works too.

One thing worth flagging for sequencing: I have follow-up work that builds directly on `_fit_response`/`_cap_facts` from this PR — a `predicate` filter for `kg_query` so a hub entity with thousands of facts stays queryable, a summary mode for results too large to return whole, and recent-first ordering so the cap keeps live facts rather than an arbitrary prefix. It's stacked on this branch and I'd rather not open it until this lands. No rush on my end, just don't want it to look like a surprise second front.

jonyskids:
Rebased onto current `develop` and force-pushed. 167 commits had landed since the August rebase, so this is a substantial reconciliation rather than a refresh — details below, since one of the resolutions is a judgement call on your code and I'd rather flag it than have you find it.

**`_run_stdio_loop` — no conflict this time.** I read today's version before touching anything, given what happened at the last rebase. It has since grown the startup-preflight thread, eager-embedder warmup, the idle-exit and write-stall watchdogs, `OSError`/EOF disconnect handling and the `BrokenPipeError` write guard. This PR's only surviving delta in that function is the one-line `logger.error(f"Server error: {e}")` → `logger.exception("Server error handling request")` swap, and its context is still verbatim, so it applied clean. The rebased function keeps your current structure — no outer fatal-wrap try/except reintroduced.

**The one substantive conflict: `tool_kg_query`.** You rewrote it to bucket rows into `active_facts` / `historical_facts` / `future_facts` alongside the flat `facts` list, plus the `resolved_from` / `candidates` fallbacks. This PR had replaced the whole return with a capped `facts` payload, so the two collided head-on.

I resolved it by **capping at the source, before bucketing** — `_cap_facts(results, …)` runs first and the capped list feeds `facts`, all three buckets, `count` and the `resolved_names` check; `total` / `truncated` / `notice` are set on your `payload` dict. Your resolution logic and candidate fallback are otherwise untouched.

The reasoning: the four lists are views of the same rows, so capping only `facts` would leave the serialized response as large as it was before and defeat the point of the commit.

**The consequence, stated plainly:** above the cap, `active_facts` is no longer exhaustive — it holds the live facts among the first N rows, not every live fact. That is a behaviour change to code you just wrote, and if you'd rather have it a different way (cap per bucket, or leave the buckets uncapped and accept the size) say so and I'll redo it.

**A related weakness this PR does not fix.** Because the cap takes the first N rows in query order, a truncated result can drop live facts in favour of invalidated ones. I have a follow-up commit that sorts recent-first so the cap keeps live facts, but it sits outside this PR's scope and I didn't want to widen the diff mid-review. It's queued behind this one along with a `predicate` filter for `kg_query` and a summary mode for results too large to return whole — all of which build on `_fit_response`/`_cap_facts` from here, which is why they've been waiting.

**Other conflict** was mechanical: `tests/test_mcp_server.py` had your `test_2288_grouped_graph_stats_count_distinct_room_instances` and this branch's `TestOversizedResponseBackstop` both appended at EOF. Both kept, yours first.

**Verification.** Full suite via CI's exact invocation (`pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-fail-under=80`): 4882 passed, 53 skipped, 0 failed, coverage 81.96%. Every skip is environment-gated and none are in this PR's area — 22 rust exact backend (native extension not built), 15 live pgvector conformance, 11 milvus, 2 format-miner (striprtf/markitdown), 1 Qdrant, 1 pgvector, 1 `hooks_cli` tmp_path case. Nothing skipped in `test_mcp_server.py` or `test_knowledge_graph.py`. `ruff format --check` (251 files) and `ruff check` are both clean.

Since no existing test drives `kg_query` past the cap, I also ran 250 rows through the rebased function directly: `count=100`, `total=250`, `truncated=True`, and `active + historical + future == len(facts) == 100`, serializing as valid JSON through the `_fit_response` chokepoint.

**One thing worth knowing about headroom:** the payload shape stores every row twice, once in `facts` and once in a bucket, so a capped 100-fact response already serializes to roughly 39KB against the 50KB `MAX_RESPONSE_CHARS` backstop. Not something this PR introduces, but the margin is a lot thinner than when the cap was sized.

CI still hasn't run on this branch — it'll need your workflow approval again for the new head.

jonyskids:
Rebased onto current `develop` (`25203ed`) and force-pushed. 77 commits had landed since the 8 Sep base, and one of them changes the shape of this PR entirely: `1beed5a` split `mcp_server.py` into a package. Both files this branch touches — `mempalace/mcp_server.py` and `tests/test_mcp_server.py` — no longer exist, so all four commits came back as modify/delete rather than content conflicts. Nothing could be replayed; every hunk had to be placed into a fragment by hand against today's code.

Two of this PR's deltas turned out to be obsolete, and I dropped both rather than reintroduce them.

**`_run_stdio_loop` — dropped, you already landed it.** `runtime.py:145` now reads `logger.exception("Server error")`, and it does better than this branch did: it also returns `-32603` instead of silently continuing, and skips the response for notifications. That one-line swap was this PR's only surviving delta in that function, so **this PR no longer touches `_run_stdio_loop` at all.**

**`tool_kg_timeline` — dropped, your pagination supersedes it.** You gave it real `limit`/`offset` paging with `timeline_total`, bounded by `_MAX_RESULTS`, so the caller can page rather than being handed a truncated prefix. That is strictly better than `_cap_facts` there. I removed the timeline cap and narrowed the `KG_RESULT_CAP` comment to `kg_query`, which is now its only consumer.

**The judgement call, and it is on something you already approved.** You signed off on the default `~/.mempalace/mcp_server.log` handler as written. I narrowed it: it now attaches **only when the root logger is unconfigured**, i.e. the standalone `mempalace-mcp` process. Your `_init_logging` docstring states "with `MEMPALACE_LOG_FILE` unset the root logger is not touched at all" — unconditional defaulting makes that false for any embedding host whose user happens to have a palace. Measured on this branch, with a host app owning root and `~/.mempalace` present:

- unconditional (as approved): `nhandlers=3`
- scoped to standalone (as pushed): `nhandlers=2`

`test_host_root_logger_config_survives_import` asserts `nhandlers=2`. Unpinned, it would have been ambient-`HOME`-dependent — which is the same order-dependence that bit me in August, showing up in your test this time. The `_clean_home_env` pinning from `889347c` is still there, but it is now belt-and-braces rather than the fix. The standalone path is where the client swallows stderr, so the crash-logging purpose is unaffected. Say the word and I'll put it back to unconditional.

**Placement is also mine to defend.** The split package has no obvious home for cross-cutting helpers, so I put each next to its only consumer: `KG_RESULT_CAP` + `_cap_facts` in `tools_kg.py`, `MAX_RESPONSE_CHARS` + `_fit_response` in `protocol.py` beside `handle_request`. If you'd rather they sat together in `_guards.py`, that's a move, not a rewrite. Tests followed the same rule — the `kg_invalidate` cases into `tests/mcp/test_kg.py`, `TestColdStartDiagnostics` and `TestOversizedResponseBackstop` into `tests/mcp/test_protocol.py`.

**`tool_kg_query` applied unchanged** from the 9 Sep resolution — `_cap_facts` runs before bucketing, so `facts`, all three buckets, `count` and the `resolved_names` check derive from one capped set. The consequence I flagged then still stands: above the cap, `active_facts` holds the live facts among the first N rows, not every live fact.

**Verification.** CI's exact ubuntu invocation (`pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10`), `HOME` pinned to a scratch dir: **5428 passed, 53 skipped, 0 failed, coverage 82.57%**. Every skip is environment-gated — 22 rust exact backend (native extension not built), 15 live pgvector conformance, 11 milvus (`pymilvus` absent), 2 format-miner (`striprtf` / markitdown), 1 Qdrant live, 1 pgvector live, 1 `hooks_cli` tmp_path case. **Zero skips in `tests/mcp/`**, so nothing in this PR's area is being skipped past. `ruff format --check` (316 files) and `ruff check` both clean.

No test drives `kg_query` past the cap, so I ran 250 mixed live/expired/future rows through the rebased function directly: `count=100`, `total=250`, `truncated=True`, `active + historical + future == len(facts) == 100`, serializing to 24,269 chars through `_fit_response` and parsing as valid JSON.

**CI has still never run on this PR** — no check runs against any head sha since it was opened. Could you approve the workflow for `1d4c5d2`? That's the only thing outstanding; everything above is a local measurement and I'd rather have your matrix confirm it, particularly 3.9 and Windows, given how much moved underneath this.

PR #1903, opened 2026-06-30
feat(kg): pluggable knowledge-graph storage backend (BaseKGStore)

t425132:
## What

A small storage-backend seam for the knowledge graph, mirroring the existing pluggable vector-backend pattern (`mempalace/backends/registry.py`). The KG can now be backed by something other than the bundled local SQLite store, selected at runtime — with **zero change to existing behavior** (the default remains the local `KnowledgeGraph`).

- **New `mempalace/kg_store.py`:**
  - `BaseKGStore(ABC)` — the 9-method KG contract (`add_entity`, `add_triple`, `query_entity`, `query_relationship`, `invalidate`, `seed_from_entity_facts`, `timeline`, `stats`, `close`), matching the current `KnowledgeGraph` signatures exactly.
  - `get_kg_store(db_path=None, *, explicit=None, backend=None)` — resolves `explicit` / `backend` / `$MEMPALACE_KG_BACKEND` → default `"sqlite"`. The default **deferred-imports** `KnowledgeGraph` (no import-time cost, no circular import).
  - Lazy, memoized entry-point discovery over a `mempalace.kg_backends` group (logs load failures via `logger.exception`, validates `issubclass(BaseKGStore)`) — the same idiom as `backends/registry.py`.
- `KnowledgeGraph` now subclasses `BaseKGStore` (`name = "sqlite"`); method bodies unchanged.
- `mcp_server._get_kg` and `fact_checker` construct the KG via `get_kg_store(...)` instead of `KnowledgeGraph(...)` directly — same caching, same paths.

## Why

To point the KG at a shared/remote store (e.g. a hosted Postgres) so multiple agents can share one knowledge graph, without forking the engine. MemPalace already supports this for *vector* backends via entry points; this brings the KG to parity. The seam is intentionally minimal and additive — it forces no new dependency and changes nothing for existing single-user/local installs.

A concrete second backend (an Aurora/Postgres KG store via the RDS Data API) was built out-of-tree to validate it: a conformance suite runs the same operations against both the SQLite `KnowledgeGraph` and the external backend and asserts identical observable output. Happy to share it.

## Backward compatibility

- Default path unchanged: `get_kg_store()` → local `KnowledgeGraph`, same behavior; the sqlite import stays deferred (no startup cost, no circular import).
- `__abstractmethods__` is empty (all 9 implemented), so `KnowledgeGraph` instantiates exactly as before.
- Existing `_get_kg` caching / `fact_checker` behavior preserved; 3 `test_mcp_server.py` tests had their construction monkeypatch repointed to the factory (assertions unchanged).

## Tests

- New: `tests/test_kg_store_contract.py` (ABC contract), `tests/test_kg_registry.py` (registry + `issubclass` guard + discovery), `tests/test_kg_routing.py` (mcp_server/fact_checker route through the factory).
- Full KG/MCP suite green locally (402 passed); `ruff check` / `ruff format --check` clean.

## Notes

I modeled the registry on `backends/registry.py` (logging + `issubclass` validation) but kept it minimal (no `unregister`/reset) — happy to fully mirror that module if you'd prefer the consistency. Cut from `develop`; glad to adjust naming or structure.

### DIFF
--- a/mempalace/fact_checker.py
+++ b/mempalace/fact_checker.py
@@ -197,14 +197,14 @@ def _check_kg_contradictions(text: str, palace_path: str) -> list:
         return []
 
     try:
-        from .knowledge_graph import KnowledgeGraph
+        from .kg_store import get_kg_store
 
         # KG lives alongside the palace collection; mcp_server uses the
         # same convention (see _kg init). Pass ``db_path`` — the previous
         # code passed a nonexistent ``palace_path`` kwarg which raised
         # TypeError, silently swallowed by the outer except and rendered
         # the entire KG-check path dead.
-        kg = KnowledgeGraph(db_path=os.path.join(palace_path, "knowledge_graph.sqlite3"))
+        kg = get_kg_store(db_path=os.path.join(palace_path, "knowledge_graph.sqlite3"))
     except Exception:
         # KG unavailable (brand-new palace, corrupted DB, etc.) — skip.
         return []
--- a/mempalace/kg_store.py
+++ b/mempalace/kg_store.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import os
 from abc import ABC, abstractmethod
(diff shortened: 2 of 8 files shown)


t425132:
Thanks for the review! Addressed all four points in 360bca0:

- Added `from __future__ import annotations` — matches the repo convention (20+ modules use it) and keeps PEP 604 unions working on the `requires-python >= 3.9` floor.
- Gave `BaseKGStore` a default `__init__(db_path=None, **kwargs)` plus `__enter__`/`__exit__`, so alternate backends share an init contract and get `with` support (mirrors `KnowledgeGraph`, which already implements the context-manager protocol).
- Switched the abstract signatures to explicit `X | None` unions.
- Corrected `_get_kg`'s return annotation to `BaseKGStore` (it now returns pluggable backends via `get_kg_store`).

`ruff check .` / `ruff format --check` are clean and the KG/MCP suite is green locally.

igorls:

PR #1907, opened 2026-07-01
docs: note MiniLM is the default without onboarding

soheilfallah:
## What

The Requirements section lists `embeddinggemma` as the model onboarding
recommends, but doesn't mention what happens if onboarding is skipped. This
adds one sentence clarifying the fallback and how to switch.

## Why

Installing via `uv tool install` and running `mine` **without** onboarding
silently falls back to `all-MiniLM-L6-v2`, because `embedding_model` defaults
to `"minilm"` when `config.json` has no entry (see `config.py`
`embedding_model` property). A user expecting the documented multilingual
default gets an English-only index with no notice.

## Change

Documents the fallback and the two supported ways to get the multilingual
model:
- set `MEMPALACE_EMBEDDING_MODEL=embeddinggemma` (or run onboarding) **before**
  the first mine, and
- `mempalace repair rebuild-index` to switch an existing palace.

Docs-only; no code changes.


soheilfallah:
Thanks for the review. Addressed in `eb5c4b4`: split the bullet into three explicit cases (skip → `minilm` default / choose multilingual before first `mine` / switch an existing palace) and made it clear that `MEMPALACE_EMBEDDING_MODEL=embeddinggemma` must be set **for the rebuild** — `mempalace repair rebuild-index` falls back to `minilm` otherwise. Good catch on that last point.

igorls:

PR #1909, opened 2026-07-01
fix(normalize): strip full slash-command envelope + ANSI escapes (#1333)

### LINKED ISSUE #1333: Claude Code transcript noise: <local-command-*> tags and ANSI escapes survive strip_noise()
## What happened?

After mining a Claude Code session into the palace, drawers contain raw `<local-command-stdout>` envelopes, bare `<command-args></command-args>` remnants, and ANSI escape sequences from Bash tool output. These pollute search results, bloat embeddings (each escape is several BPE tokens), and produce drawers that are mostly noise.

Real samples from a normal mining run (`ESC` denotes the literal `\x1b` byte stored in drawer content):

```
> <local-command-stdout>Set defaultPermissionMode to ESC[1mdefaultESC[22m
Disabled auto-compact Disabled Remote Control for all sessions</local-command-stdout>
```

```
> <command-message>clear</command-message>
<command-args></command-args>
```

```
ESC[38;2;153;153;153m├ESC[39m mempalace_add_drawer ESC[38;2;153;153;153m├ESC[39m mempalace_check_duplicate ...
```

In one palace I inspected, ~85% of drawers in `wing=sessions` landed in `room=technical` and many were dominated by ANSI bytes from `/context` / `/help` output, or were the empty `<command-args></command-args>` remnant of slash-command invocations.

## What did you expect?

The existing `strip_noise()` machinery in `mempalace/normalize.py` already strips `<system-reminder>`, `<command-message>`, and `<command-name>`. The same hygiene should apply to:

1. The rest of the slash-command envelope Claude Code injects (`<local-command-caveat>`, `<command-args>`, `<local-command-stdout>`).
2. ANSI control sequences (CSI + OSC) that Claude Code's `Bash` tool faithfully preserves from terminal output.

## How to reproduce:

1. Run any Claude Code session that uses slash commands (`/clear`, `/effort`, `/help`, plugin commands) and any Bash tool calls producing colored output (most modern CLIs).
2. Wait for the mempalace Stop hook to mine the transcript, or run manually:
   ```
   mempalace mine ~/.claude/projects/<your-folder> --mode convos --wing sessions
   ```
3. Inspect the result, e.g.:
   ```
   mempalace_list_drawers wing=sessions
   ```
   Drawers will contain ANSI-laden content and surviving `<local-command-*>` tag envelopes.

## Environment:

- OS: macOS 25.3.0 (Darwin)
- Python: 3.14
- MemPalace: 3.3.3 (verified `develop` tip has identical relevant code path in `mempalace/normalize.py`)
- Harness: Claude Code

## Root-cause sketch

Claude Code wraps every slash-command invocation in a five-tag envelope:

```
<local-command-caveat>...</local-command-caveat>
<command-name>...</command-name>
<command-message>...</command-message>
<command-args>...</command-args>
<local-command-stdout>...</local-command-stdout>
```

`_NOISE_TAGS` in `mempalace/normalize.py` covers `command-name` and `command-message`; the other three slip through. Separately, `Bash` tool_result content is captured verbatim — including ANSI bytes — and `strip_noise()` does not currently sweep them.

Fix: extend `_NOISE_TAGS` with `command-args`, `local-command-caveat`, `local-command-stdout`, and add CSI + OSC ANSI strippers applied after tag removal in `strip_noise()`. Each ANSI pattern is anchored on the literal ESC byte so user prose that mentions e.g. `[1m]` by name stays intact. PR coming separately.

KeilerHirsch:
## Summary

`strip_noise()` in `mempalace/normalize.py` only stripped three of the tags Claude Code injects (`<system-reminder>`, `<command-message>`, `<command-name>`). Real transcripts also carry the rest of the slash-command envelope plus raw ANSI escapes from Bash-tool output, all of which leaked into drawers verbatim — bloating embeddings (each escape is several BPE tokens) and polluting search. This is #1333.

Closes #1333.

## What changed

- **Four missing envelope tags** added to `_NOISE_TAGS`: `command-args`, `local-command-caveat`, `local-command-stdout`, `local-command-stderr`.
  - `local-command-stderr` was **not** in the #1333 root-cause sketch (which listed five tags) but is present in real transcripts — I enumerated every `<(local-)?command-*>` tag across a live `~/.claude/projects` corpus and all six appear.
- **Tag-body regex** now halts at the next opening tag of the same name instead of the first blank line, so a **multi-paragraph** `<local-command-stdout>` block (tables, stack traces) is removed whole. The line-start anchor still prevents a dangling tag from eating neighbouring messages.
- **ANSI stripping** (CSI + OSC), anchored on the literal ESC byte (`0x1b`) — a control char that never appears in legitimate prose — so text that merely *names* a sequence like `ESC[0m` survives. ReDoS-safe by construction (disjoint/negated classes, no nested overlapping quantifiers).
- **`NORMALIZE_VERSION` 2 → 3** so already-mined drawers rebuild on the next pass (convo mining gates on the version, not mtime).

## Relationship to prior PRs

This supersedes the two stale attempts and folds in the review feedback each got:

- **#1334** was flagged in review as *action-required* for **not bumping `NORMALIZE_VERSION`** — done here.
- **#1701** is `CONFLICTING` and 3+ weeks stale; it also **misses `local-command-stderr`**. This PR rebases clean on `develop` and adds that tag.

The multi-paragraph-halt regex approach matches #1701's, credited in spirit.

## Tests

12 new cases in `tests/test_normalize.py` (`TestStripNoiseClaudeCodeEnvelopeAndAnsi` + the per-tag loop): multi-paragraph stdout, stderr, empty `command-args` remnant, caveat, ANSI CSI SGR, ANSI OSC hyperlink, and a prose-naming-survives verbatim check.

- [x] `pytest tests/test_normalize.py` — 159 passed
- [x] `pytest tests/test_convo_miner.py tests/test_miner.py` — 134 passed, 3 skipped
- [x] `ruff check mempalace/normalize.py mempalace/palace.py tests/test_normalize.py` — clean

### DIFF
--- a/mempalace/normalize.py
+++ b/mempalace/normalize.py
@@ -45,6 +45,17 @@ _NOISE_TAGS = (
     "system-reminder",
     "command-message",
     "command-name",
+    # Rest of the slash-command envelope Claude Code injects around every
+    # slash-command invocation (#1333). Before this only the middle three tags
+    # were covered, so the caveat, args, and captured stdout/stderr streams
+    # leaked into drawers verbatim. ``local-command-stderr`` is included even
+    # though the #1333 root-cause sketch listed only five tags — it is emitted
+    # for commands that write to stderr and was found in real transcripts
+    # alongside the other five.
+    "command-args",
+    "local-command-caveat",
+    "local-command-stdout",
+    "local-command-stderr",
     "task-notification",
     "user-prompt-submit-hook",
     "hook_output",
@@ -53,11 +64,15 @@ _NOISE_TAGS = (
 
 def _tag_pattern(name: str) -> "re.Pattern[str]":
     # Opening tag must begin a line (optionally after a `> ` blockquote marker,
-    # since _messages_to_transcript prefixes lines with `> `). Body is lazy but
-    # forbidden from crossing a blank line, so a dangling open tag can't span
-    # multiple messages. Closing tag eats optional trailing whitespace + newline.
+    # since _messages_to_transcript prefixes lines with `> `). The body is lazy
+    # and halts at the next opening tag of the same name, so a dangling open tag
+    # can't span past the next same-tag block — while still allowing blank lines
+    # *inside* the tag (multi-paragraph <local-command-stdout> output, tables,
+    # stack traces; #1333). Combined with the line-start anchor this keeps a
+    # stray tag from eating neighbouring messages. Closing tag eats optional
+    # trailing whitespace + newline.
     return re.compile(
-        rf"(?m)^(?:> )?<{name}(?:\s[^>]*)?>" rf"(?:(?!\n\s*\n)[\s\S])*?" rf"</{name}>[ \t]*\n?"
+        rf"(?m)^(?:> )?<{name}(?:\s[^>]*)?>" rf"(?:(?!<{name}\b)[\s\S])*?" rf"</{name}>[ \t]*\n?"
     )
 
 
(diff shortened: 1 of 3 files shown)


jrzmurray:
Found a truncation-safety edge case in `_ANSI_CSI_RE` while adapting this for local use — flagging it since it'd affect this PR as submitted, though it's unlikely to come up often in real data.

```python
_ANSI_CSI_RE = re.compile(r"\x1b\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]")
```

Both middle groups are zero-or-more, and the final-byte class (`0x40-0x7E`) covers nearly every ASCII letter. That means a genuinely truncated `ESC[` (no parameters ever written, e.g. because the writer got cut off mid-stream) immediately followed by ordinary real text is indistinguishable from a valid *empty-parameter* CSI sequence — the regex happily treats the next real letter as the "final byte" and eats it.

Confirmed empirically:

```python
>>> _ANSI_CSI_RE.sub("", "before \x1b[this has words after")
'before his has words after'   # the leading "t" of "this" is gone
```

This needs a genuinely truncated escape sequence to land directly against real prose with zero digits in between, which is a narrow window — real Bash tool output being truncated mid-write does happen (a backgrounded process's redirected stdout racing its reader), but the exact conjunction here is uncommon. Not a blocker, just worth hardening defensively since the fix is small: require at least one parameter byte (`+` instead of `*`) and drop the intermediate-byte class, since real SGR/cursor codes are overwhelmingly just digits and `;`:

```python
_ANSI_CSI_RE = re.compile(r"\x1b\[[0-9;]+[\x40-\x7e]")
```

Tradeoff: a truly bare, zero-param CSI (e.g. `ESC[H`, cursor-home with no row/col) would no longer be stripped — left as harmless unstripped noise instead. Verified this closes the repro above while every real-world SGR/cursor-with-param shape I tried still strips correctly.

(Also noticed neither this PR nor #1334 strip the "Fe/Fs" single-byte escape sequences — e.g. `ESC c`, RIS/full-reset, which watch-mode dev tools like `tsx watch`/`nodemon` emit before reprinting on a file-change restart. Only CSI/OSC are covered. Separate, smaller gap if useful — happy to share that pattern too if it'd help.)

PR #1910, opened 2026-07-01
fix(palace): make per-file mine_lock block indefinitely on Windows

KeilerHirsch:
## Summary

The per-file `mine_lock` (`mempalace/palace.py`) is a **blocking** lock: a second writer for the same source file must *wait*, not fail — that's what prevents the delete+insert interleave that produces duplicate drawers. On POSIX `fcntl.flock(LOCK_EX)` waits indefinitely, which is correct. On Windows `msvcrt.locking(LK_LOCK, ...)` only retries ~10 times at 1s intervals and then raises `OSError(EDEADLOCK)`.

A per-file mine of a large transcript (hundreds of MB, minutes of embedding work) routinely runs far longer than 10s, so on Windows a contending waiter **crashed instead of waiting** — the exact opposite of the lock's contract, and platform-divergent behaviour from the same call site.

`mine_lock` wraps a full per-file mine in 7 call sites (`miner.py`, `convo_miner.py`, `diary_ingest.py`, `format_miner.py`, `closet_llm.py`, `palace_graph.py`), so this affects every mining path under concurrency on Windows.

## Fix

Re-issue `LK_LOCK` on the contention errno (`EDEADLOCK`/`EDEADLK`/`EACCES`) so a blocking acquire truly blocks, matching the POSIX contract. Because each `LK_LOCK` call already sleeps ~10s internally before raising, the retry loop is **not** a busy-spin — it's a 10s-granularity block. Non-contention errors (e.g. `EBADF`) still propagate immediately so a genuinely bad handle can't loop forever. The non-blocking path (`LK_NBLCK`, used by `mine_palace_lock` cleanup) is unchanged and still returns `False` on contention.

## Tests

3 new cases in `tests/test_palace_locks.py` that patch a fake `msvcrt` into `sys.modules` and force `os.name == "nt"`, so the Windows path is exercised deterministically **on any host** (incl. Linux/macOS CI):

- `test_windows_blocking_lock_retries_past_lk_lock_timeout` — two simulated `LK_LOCK` timeouts then success; asserts it returns `True`, not raises.
- `test_windows_blocking_lock_reraises_non_contention_error` — `EBADF` propagates.
- `test_windows_nonblocking_lock_returns_false_on_contention` — non-blocking still returns `False`.

- [x] `pytest tests/test_palace_locks.py` — 15 passed
- [x] `ruff check mempalace/palace.py tests/test_palace_locks.py` — clean

### DIFF
--- a/tests/test_palace_locks.py
+++ b/tests/test_palace_locks.py
@@ -621,7 +621,8 @@ class _FakeMsvcrt:
     def locking(self, fileno, mode, nbytes):
         self.calls += 1
         if self.calls <= self.fail_times:
-            raise OSError(errno.EDEADLOCK, "Resource deadlock avoided")
+            deadlock_errno = getattr(errno, "EDEADLOCK", errno.EDEADLK)
+            raise OSError(deadlock_errno, "Resource deadlock avoided")
 
 
 def test_windows_blocking_lock_retries_past_lk_lock_timeout(tmp_path, monkeypatch):
--- a/tests/test_palace_locks.py
+++ b/tests/test_palace_locks.py
@@ -1,4 +1,4 @@
-"""Tests for mine_palace_lock — the per-palace non-blocking mine guard.
+"""Tests for mine_palace_lock â€” the per-palace non-blocking mine guard.
 
 Covers the fix for the runaway mine fan-out described alongside issues
 #974 and #965: if N copies of `mempalace mine` are spawned concurrently
@@ -10,6 +10,7 @@ against *different* palaces must still be free to run in parallel.
 from __future__ import annotations
 
 import multiprocessing
+import errno
 import os
 import threading
 import time
@@ -29,15 +30,15 @@ from mempalace.palace import (
 
 
 def _get_mp_context():
-    """Always use ``spawn`` — ``fork`` deadlocks under modern Python.
+    """Always use ``spawn`` â€” ``fork`` deadlocks under modern Python.
 
     The parent (pytest + chromadb + onnxruntime) is multi-threaded by the time
     these tests run. ``fork`` snapshots that state into the child without the
     threads that hold the locks, which Python 3.13 explicitly warns about and
     which deadlocks the CI runners. macOS additionally forbids
     fork-without-exec via CoreFoundation. ``spawn`` re-imports the package in
-    the child (slower, but safe) and inherits ``os.environ`` — including the
-    monkeypatched ``HOME`` — which is all these lock-file tests need.
+    the child (slower, but safe) and inherits ``os.environ`` â€” including the
+    monkeypatched ``HOME`` â€” which is all these lock-file tests need.
     """
     return multiprocessing.get_context("spawn")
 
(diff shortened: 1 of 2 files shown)


KeilerHirsch:

The only merge conflict was in `test_palace_locks.py`, where current `develop` has the newer orphaned per-source-file lock reaper coverage. I kept that entire current test block and the Windows `LK_LOCK` retry regressions together; production `palace.py` merged cleanly.

- `tests/test_palace_locks.py`: 22 passed

GitHub now reports the PR mergeable again. Pre-rebase head is preserved as `archive/2026-08-29/pr1910-pre-rebase`.

KeilerHirsch:
macOS CI root cause found and fixed in `b7f521b`.

This was a **test-harness portability bug**, not a production locking failure: `_FakeMsvcrt.locking()` raised `errno.EDEADLOCK` directly, but Python/macOS exposes `EDEADLK` and no `EDEADLOCK`, so both simulated-Windows contention tests failed before `_lock_mine_lock_file()` was meaningfully exercised.

The fake now uses `getattr(errno, "EDEADLOCK", errno.EDEADLK)`. Production code already builds its contention errno set defensively from whichever names the host defines.

Fresh local verification:
- 3 Windows-lock target tests: pass
- same 3 tests with `errno.EDEADLOCK` deliberately removed from the Python process (macOS-style symbol surface): pass
- full `tests/test_palace_locks.py`: **22 passed**
- `ruff check` + `ruff format --check`: clean

The previous macOS run failed only these two tests with `AttributeError: module 'errno' has no attribute 'EDEADLOCK'`; Linux/Windows and the rest of the matrix were green. New CI is triggered by this push.

tests/test_palace_locks.py
```diff
@@ -333,7 +334,7 @@ def test_write_lock_holder_writes_utf8_bytes_for_non_ascii_argv(tmp_path, monkey
     monkeypatch.setattr(
         sys,
         "argv",
-        ["mempalace", "mine", "café/北"],
+        ["mempalace", "mine", "cafÃ©/åŒ—"],
```
AmirF194, line 337: I diffed this PR's test file against `origin/main` at the byte level, not just the rendered diff.

`tests/test_palace_locks.py` was re-saved with a UTF-8-as-CP1252 double-encoding. Every pre-existing em dash in the file (15 occurrences, none of them touched by this PR's own changes) turned into a three-character mojibake sequence (codepoints U+00E2 U+20AC U+201D), and the two Unicode test literals on this line and at lines 361/366 (`café/北`) got the same treatment, so this line no longer contains the original CJK character.

```
$ python3 -c "print(bytes.fromhex('e28094').decode('cp1252').encode('utf-8'))"
b'\xc3\xa2\xe2\x82\xac\xe2\x80\x9d'   # matches all 15 corrupted lines in this file
$ git show origin/main:tests/test_palace_locks.py | grep -n 'café|北'
336:        ["mempalace", "mine", "café/北"],
360:            raise UnicodeEncodeError("cp1252", "北", 0, 1, "not representable")
365:    monkeypatch.setattr(sys, "argv", ["mempalace", "mine", "北"])
```

`test_write_lock_holder_writes_utf8_bytes_for_non_ascii_argv` computes its expected value from `sys.argv` at run time, so it still passes with the corrupted string, it just no longer exercises the regression #1435 was actually filed against (a CJK path segment). CI is green here, so nothing catches this.

Worth a re-save of this file as UTF-8 and restoring the original `café/北` literals before merge, independent of the Windows locking fix itself, which I did not find a problem with.

PR #1922, opened 2026-07-04
feat(miner): support .mempalaceignore as the mining ignore file

kostadis:
## What

Adds a project-scoped `.mempalaceignore` convention to the miner. If a `.mempalaceignore` file exists at the project root, mining (and the sync/prune path) read `.mempalaceignore` throughout the tree and never consult `.gitignore`. Otherwise behavior is unchanged — `.gitignore` is used exactly as today.

## Why

Today the mining ignore source is hardcoded to git (`is_gitignored`, `load_gitignore_matcher`, `respect_gitignore`). "What should mining ignore?" and "what does git ignore?" are not always the same question:

- Mine a config/notes directory that git ignores.
- Exclude generated content from the palace that git tracks.

`.mempalaceignore` gives users an ignore surface that's independent of git, while costing nothing for anyone who doesn't use it.

## Design

- **Project-scoped, not per-directory.** `detect_ignore_filename(project_root)` decides the convention once at the root. A per-directory fallback (try `.mempalaceignore`, else `.gitignore` in each dir) was deliberately avoided — it lets one tree mix conventions confusingly (a root `.mempalaceignore` masks the same-dir `.gitignore` but not a sibling-dir one). One convention per project keeps the contract obvious.
- **Same matcher engine.** `.mempalaceignore` uses the identical gitignore-syntax matcher (`GitignoreMatcher`); only the filename read changes.
- **Ingest and prune agree.** `scan_project` and `sync._ancestor_matchers` both honor the detected filename, so a file blocked on the way in is also pruned on the way out.

## Backward compatibility

Fully additive — no behavior change for any project without a `.mempalaceignore`:

- `GitignoreMatcher.from_dir` / `load_gitignore_matcher` gained an optional `filename=` argument that **defaults to `.gitignore`**; existing callers are untouched.
- The public `respect_gitignore` parameter and the `--no-gitignore` CLI flag are preserved. `--no-mempalaceignore` is added as an alias of `--no-gitignore`.
- `.mempalaceignore` is added to `SKIP_FILENAMES` so it isn't itself ingested.

(Naming note: I kept the existing `is_gitignored` / `load_gitignore_matcher` / `respect_gitignore` names and the `GitignoreMatcher` class to avoid a breaking rename. If you'd prefer neutral names — `is_ignored` / `respect_ignore` — I'm happy to rename with back-compat aliases.)

## Tests

Adds coverage in `tests/test_miner.py`:

- `detect_ignore_filename` prefers `.mempalaceignore`, falls back to `.gitignore`
- `scan_project` respects `.mempalaceignore`
- project-mode: a root `.mempalaceignore` makes `.gitignore` (incl. nested) inert
- git-mode fallback preserved when no `.mempalaceignore`
- `respect_gitignore=False` still disables ignore matching

All new tests pass; ruff check + format clean.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1799,8 +1799,10 @@ def main():
     p_mine.add_argument("--wing", default=None, help="Wing name (default: directory name)")
     p_mine.add_argument(
         "--no-gitignore",
+        "--no-mempalaceignore",
+        dest="no_gitignore",
         action="store_true",
-        help="Don't respect .gitignore files when scanning project files",
+        help="Don't respect .mempalaceignore / .gitignore files when scanning project files",
     )
     p_mine.add_argument(
         "--include-ignored",
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -150,6 +150,7 @@ SKIP_FILENAMES = {
     "mempal.yaml",
     "mempal.yml",
     ".gitignore",
+    ".mempalaceignore",
     "package-lock.json",
     "pnpm-lock.yaml",
     "yarn.lock",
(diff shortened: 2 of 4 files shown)


mvalentsev:
Heads-up: this overlaps #154, which implements the same #102 ask (a mempalace-specific ignore file) through the same seam: an optional `filename=` on `GitignoreMatcher.from_dir` / `load_gitignore_matcher`.

Differences worth a maintainer call before either lands:
- Name: `.mempalace-ignore` (#154) vs `.mempalaceignore` here.
- #154 layers it over `.gitignore` per-directory; this PR is project-scoped either/or (a root `.mempalaceignore` disables `.gitignore` tree-wide).
- This PR also fixes the `sync.py` prune path; #154 is ingest-only.

Disclosure: #154 is mine, so I'm flagging rather than judging. Up to the maintainers which shape to land.

kostadis:
## Code Review Fix Applied

Addressed the Gemini Code Assist feedback: `detect_ignore_filename(root)` is now cached in `matcher_cache` keyed by `("ignore_filename", root)` at `mempalace/sync.py:75-78`, avoiding redundant `is_file()` calls per source file.

## Analysis of Alternative (#154)

I examined PR #154 (by @mvalentsev) which implements the same issue (#102). Heres the comparison:

| Dimension | This PR (#1922) | PR #154 |
|---|---|---|
| **File name** | `.mempalaceignore` | `.mempalace-ignore` |
| **Scope** | Project-scoped either/or — root `.mempalaceignore` supresses `.gitignore` tree-wide | Per-directory layering — `.mempalace-ignore` loads alongside `.gitignore` in every dir |
| **When honored?** | When `respect_gitignore=True` (same gate as `.gitignore`) | Always — even with `--no-gitignore` |
| **Prune path** | Fixed (`sync.py` uses same convention) | Not covered (ingest only) |
| **Performance** | 1 `load_gitignore_matcher` call per dir | 2 calls per dir |
| **Conflict resolution** | Either/or — one ignore rule set active at a time | Layered — `.mempalace-ignore` has final say |
| **detect_ignore_filename** | Yes — decides convention once at root | No — always loads both files |

### Key tradeoff

This PR is **simpler and more performant** (1 matcher per dir, prune path covered), but its **either/or** — a root `.mempalaceignore` makes `.gitignore` invisible everywhere. Users who want *both* git exclusions *and* mining-specific exclusions simultaneously cannot do that.

PR #154 lets `.mempalace-ignore` **layer on top** of `.gitignore` (applied after, so it wins on disagreement), and it works **even with `--no-gitignore`**. This is more flexible but costs 2× matcher calls and doesnt cover the prune path.

Happy to adapt either direction — this is a maintainer call on which shape fits the project philosophy.

igorls:

Issue #1927, opened 2026-07-05
Bulk-ingest paths don't validate/sanitize embedded NUL bytes before reaching ChromaDB


jrzmurray:
## Summary

Documents mined by the bulk-ingest paths (`convo_miner`, `miner`, `format_miner`, `sweeper`, `diary_ingest`) reach `ChromaCollection.upsert()`/`add()` without any check for embedded NUL characters (U+0000). A NUL-containing document is well-formed UTF-8 — unlike a lone surrogate, it doesn't raise on `.encode()` — but handing it to ChromaDB's SQLite/FTS5 layer can corrupt the FTS5 inverted index for the **whole collection**, not just fail to store that one row. Reported upstream at [chroma-core/chroma#7388](https://github.com/chroma-core/chroma/issues/7388), with a minimal reproduction independent of mempalace.

This is routine, not exotic, input: a Bash tool_result whose output was captured via a background process (`command > logfile 2>&1 &`) can end up with a contiguous run of NUL bytes if the reader races the writer (or from genuine NUL-delimited/binary output, e.g. `find -print0`). Mining real Claude Code session transcripts will hit this.

## The gap, precisely

`config.py::sanitize_content` **already validates against this** — it raises `ValueError("content contains null bytes")`. But per the `_sanitize_documents_for_chromadb` docstring in `backends/chroma.py` (added for #1235, lone-surrogate sanitization):

> `#1235` fixed this for the MCP write tools via `sanitize_content`, but the bulk ingest paths (miner, convo_miner, sweeper, diary_ingest) build documents without routing through that helper and reach this backend directly.

The exact same gap applies here: the bulk-ingest paths never call `sanitize_content`, so its NUL check never runs for mined content — only for the interactive MCP write-tool path. `_sanitize_documents_for_chromadb` (the chokepoint #1235 added specifically to close this class of gap for *all* callers) strips lone surrogates but doesn't touch NUL bytes.

## Reproduction

```python
from mempalace.backends.chroma import ChromaCollection

class _Fake:
    def upsert(self, **kwargs): self.kwargs = kwargs

fake = _Fake()
col = ChromaCollection(fake)
col.upsert(documents=["clean\x00doc"], ids=["1"])
print(fake.kwargs["documents"])  # -> ['clean\x00doc'] -- reaches chromadb unsanitized
```

[...]

Issue #1925, opened 2026-07-05
test_repair.py: 2 tests fail on SQLite 3.53.2 — FTS5 corruption message wording mismatch in _errors_are_isolated_fts5


jrzmurray:
## Summary

Two tests in `tests/test_repair.py` fail on this environment (SQLite 3.53.2 / macOS arm64) because `PRAGMA quick_check` reports a different error message string for the same class of FTS5 corruption than what `_errors_are_isolated_fts5`'s regex and the tests' assertions expect:

```
tests/test_repair.py::test_maybe_autoheal_fts5_index_heals_isolated_corruption
tests/test_repair.py::test_rebuild_index_preflight_autoheals_isolated_fts5_then_proceeds
```

Both construct their corrupt-palace fixture via the same helper (`_make_fts5_palace(tmp_path, corrupt=True)`), and both assert or rely on `repair._errors_are_isolated_fts5(errors)` returning `True` for the resulting `quick_check` output.

## Reproduction

```bash
git clone https://github.com/MemPalace/mempalace.git
cd mempalace
python3 -m venv .venv && source .venv/bin/activate
pip install -e ".[dev]"
pytest tests/test_repair.py::test_maybe_autoheal_fts5_index_heals_isolated_corruption -v
```

## Actual vs expected error string

```
AssertionError: assert (['fts5: corruption found reading blob 137438953474 from table "embedding_fulltext_search"'] and False)
 +  where False = _errors_are_isolated_fts5(['fts5: corruption found reading blob 137438953474 from table "embedding_fulltext_search"'])
```

`_FTS5_MALFORMED_RE` in `mempalace/repair.py` matches `malformed inverted index for FTS5 table` — the message this SQLite build actually produces for the same corruption is `fts5: corruption found reading blob N from table "..."`, which doesn't match.

The second test fails with the same root cause one level up — `test_rebuild_index_preflight_autoheals_isolated_fts5_then_proceeds` expects the preflight to classify the corruption as isolated-and-healable and proceed; instead it hits the `ABORT: SQLite-layer corruption detected` banner because `_errors_are_isolated_fts5` returns `False` for this message shape.

## Environment

- Python 3.13.7 (Clang 17.0.0), macOS-15.3.1-arm64
- `sqlite3.sqlite_version` (stdlib, linked): `3.53.2`
- chromadb `1.5.9`
- mempalace `develop` @ `da5a48c` (also reproduces on this exact commit with no other changes)

[...]

PR #1928, opened 2026-07-05
mine: auto-heal isolated FTS5 corruption instead of forcing manual repair

### LINKED ISSUE #1926: mine doesn't auto-heal isolated FTS5 corruption the way repair does
## Summary

`mempalace mine` aborts with the `ABORT: SQLite-layer corruption detected` banner on an isolated FTS5 inverted-index corruption — the specific case `maybe_autoheal_fts5_index` (`repair.py`) already exists to fix in place, and does fix, when reached via `mempalace repair`. That helper is wired into `cmd_repair`'s preflight, but **not** into `mine`'s own post-mine validation (`palace._validate_palace_fts5_after_mine`), so a normal `mine` forces a separate manual `mempalace repair` step for a corruption class that's already known to be safely self-healable.

## Reproduction

1. Mine a real project directory in `--mode convos` that happens to include a document with an embedded NUL character (routine in captured Bash tool output — see the related issue below for how one gets there). All files file successfully; the corruption is only detected in the post-mine `PRAGMA quick_check`.
2. Observe the abort banner:
   ```
   ABORT: SQLite-layer corruption detected before repair rebuild.
   ...
       - malformed inverted index for FTS5 table main.embedding_fulltext_search
   ```
3. Run `mempalace repair --yes` against the same palace. Its own output shows the auto-heal already exists and works:
   ```
   Isolated FTS5 inverted-index corruption detected; attempting an
   in-place rebuild from the intact content table before aborting.
   FTS5 index rebuilt from intact content; quick_check is clean.
   ```

So the fix already exists in the codebase — it's just not reachable from `mine` directly, only from the separate `repair` command.

## Expected

`mine` should attempt the same in-place auto-heal before aborting, exactly as `repair` already does, so an isolated (fully recoverable) FTS5 corruption doesn't require a second manual command.

## Root cause detail (traced in source)

`palace.py::_validate_palace_fts5_after_mine` runs `PRAGMA quick_check` unconditionally at the end of every mine and raises `MineValidationError` on any error — including the isolated-FTS5 case that `repair.py::maybe_autoheal_fts5_index` is specifically designed to recover from in place (rebuilding the FTS5 index from the intact content shadow table, no data loss).

## Related

- Root cause of the corruption itself (why FTS5 gets into this state): a NUL byte in mined content corrupting ChromaDB's FTS5 index — tracked as a separate mempalace issue (input validation) and reported upstream at [chroma-core/chroma#7388](https://github.com/chroma-core/chroma/issues/7388).
- Possibly related to #1599, though that issue's repro involves concurrent/live writers — mine reproduced this with a single uninterrupted process, no concurrency, so the trigger conditions may be broader than "concurrent writers" alone.

A PR with the fix (wiring the existing auto-heal into `mine`'s validation path) is incoming, linked to this issue.

jrzmurray:
Fixes #1926

## Summary

`mempalace mine` aborts with the `ABORT: SQLite-layer corruption detected` banner on an isolated FTS5 inverted-index corruption — the specific case `maybe_autoheal_fts5_index` already exists to fix in place. That helper is wired into `cmd_repair`'s preflight, but not into `mine`'s own post-mine validation (`palace._validate_palace_fts5_after_mine`), so `mine` forces a manual `mempalace repair` for a corruption class that's already safely self-healable.

This PR wires the same auto-heal call into `_validate_palace_fts5_after_mine`, before it raises `MineValidationError`.

## Real-world repro that motivated this

Mining a real Claude Code project directory (29 files: 3 session transcripts + 23 subagent transcripts + 3 tool-result dumps, `--mode convos`) into a fresh palace deterministically triggered this exact corruption after all 29 files filed successfully:

```
ABORT: SQLite-layer corruption detected before repair rebuild.
...
    - malformed inverted index for FTS5 table main.embedding_fulltext_search
```

No concurrent process, no interrupted/killed mine — a single, uninterrupted run. Root-caused to a NUL byte in one document (see #1927 and [chroma-core/chroma#7388](https://github.com/chroma-core/chroma/issues/7388) for that side of it) — this PR is the self-heal half of the fix, independent of that root cause.

Running `mempalace repair --yes` against the affected palace confirmed the auto-heal path clears it cleanly (visible in its own output: `Isolated FTS5 inverted-index corruption detected... FTS5 index rebuilt from intact content; quick_check is clean.`) before proceeding to a full rebuild. That's the same primitive this PR calls from `mine` directly, so a normal mine self-heals instead of requiring the separate manual step.

## Verification

- Re-ran the exact same real-world mine against a fresh palace with this patch applied: no abort, `Files processed: 29`, `Done.` `PRAGMA quick_check` confirmed clean (`[('ok',)]`) afterward, and `mempalace search` returned correct results.
- Full test suite: `3214 passed, 20 skipped` — no new failures introduced. (Two unrelated pre-existing failures in `test_repair.py` reproduce identically on unpatched `develop`; filed separately as #1925.)
- The existing `tests/test_miner_fts5_validation.py` suite (17 tests asserting `mine` must not silently succeed on FTS5 corruption) still passes unmodified — `maybe_autoheal_fts5_index` returns the *remaining* errors after the heal attempt, so `MineValidationError` still raises whenever the corruption isn't the isolated, fully-healable case. This patch only changes behavior when the heal has verifiably and fully cleared the corruption.

## Test plan
- [x] `pytest tests/test_miner_fts5_validation.py` — 17 passed
- [x] `pytest tests/` (full suite) — 3214 passed, 20 skipped, 2 pre-existing unrelated failures (see #1925)
- [x] Manual: reproduced the original corruption against real transcript data, confirmed the patched `mine` self-heals and `search`/`wake-up` work correctly afterward

### DIFF
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -951,7 +951,11 @@ def _validate_palace_fts5_after_mine(palace_path: str) -> None:
 
     errors = sqlite_integrity_errors(palace_path)
     if errors:
-        errors = maybe_autoheal_fts5_index(palace_path, errors)
+        # progress=logger.info, not the default print: this runs inside the
+        # MCP server process too (mcp_server.tool_mine -> miner.mine), where
+        # stdout is the JSON-RPC transport -- a stray print() here would
+        # corrupt the protocol stream and crash the connection.
+        errors = maybe_autoheal_fts5_index(palace_path, errors, progress=logger.info)
     if errors:
         raise MineValidationError(palace_path, errors)
 
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -928,12 +928,19 @@ def _validate_palace_fts5_after_mine(palace_path: str) -> None:
     Reuses the same primitive that `cmd_repair` already runs as preflight so the
     operator sees the same recovery banner regardless of which command surfaces
     the bug.
+
+    An isolated FTS5 inverted-index corruption (the common case after a
+    killed-mid-write mine, #1596) is auto-healed in place first via the same
+    `maybe_autoheal_fts5_index` rebuild `cmd_repair` already runs as its own
+    preflight step — so a normal `mine` self-heals the recoverable case
+    instead of forcing the operator to run `mempalace repair` by hand for a
+    derived index that regenerates from intact content.
     """
     if resolve_backend_name(palace_path) != "chroma":
         return
 
     # Defer-import: keeps the repair module graph out of mine's hot import path.
-    from .repair import _close_chroma_handles, sqlite_integrity_errors
+    from .repair import _close_chroma_handles, maybe_autoheal_fts5_index, sqlite_integrity_errors
 
     # Pass the live singleton so the writer's cached PersistentClient actually
     # gets closed and WAL flushes before the read-only sqlite3 re-open.
@@ -943,6 +950,8 @@ def _validate_palace_fts5_after_mine(palace_path: str) -> None:
     _close_chroma_handles(palace_path, backend=_DEFAULT_BACKEND)
 
     errors = sqlite_integrity_errors(palace_path)
+    if errors:
+        errors = maybe_autoheal_fts5_index(palace_path, errors)
     if errors:
         raise MineValidationError(palace_path, errors)
 
--- a/mempalace/repair.py
+++ b/mempalace/repair.py
@@ -638,13 +638,24 @@ def print_sqlite_integrity_abort(palace_path: str, errors: list[str]) -> None:
     print("    6. Re-run `mempalace repair --yes`.")
 
 
-# quick_check labels a corrupt FTS5 inverted index like:
+# quick_check's wording for a corrupt FTS5 inverted index differs by SQLite
+# version — both forms describe the same recoverable condition:
 #   "malformed inverted index for FTS5 table main.embedding_fulltext_search"
-# That specific failure is recoverable in place: the index is derived from the
+#     (older SQLite)
+#   "fts5: corruption found reading blob N from table \"embedding_fulltext_search\""
+#     (SQLite >= ~3.5x, confirmed on 3.53.2 / Python 3.13.7 — the exact
+#     message this repo's own test fixture produces on that build)
+# Either failure is recoverable in place: the index is derived from the
 # intact ``embedding_fulltext_search_content`` shadow table, so rebuilding it
 # restores full-text search without touching any drawer rows. Concurrent
-# killed-mid-write mines are the usual cause (#1596).
-_FTS5_MALFORMED_RE = re.compile(r"malformed inverted index for FTS5 table", re.IGNORECASE)
+# killed-mid-write mines are the usual cause (#1596). A regex matching only
+# the older phrasing would silently decline to auto-heal on newer SQLite —
+# the exact failure this repo's own test suite caught (test_repair.py's two
+# auto-heal tests failed on this machine until this pattern was widened).
+_FTS5_MALFORMED_RE = re.compile(
+    r"malformed inverted index for fts5 table|fts5:\s*corruption found",
+    re.IGNORECASE,
+)
 
 
 def _errors_are_isolated_fts5(errors: list[str]) -> bool:
--- a/tests/test_miner_fts5_validation.py
+++ b/tests/test_miner_fts5_validation.py
@@ -150,17 +150,25 @@ def test_helper_raises_on_page_mangled_sqlite(tmp_path):
     assert "malformed" in combined or "quick_check failed" in combined
 
 
-def test_helper_raises_on_fts5_segment_corruption(tmp_path):
-    """The reporter-shaped failure: FTS5 inverted index malformed, main pages OK."""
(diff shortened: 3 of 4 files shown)


jrzmurray:
Pulled this back to draft for a bit — running the suite against a recent SQLite build (3.53.2 / Python 3.13.7) turned up a real gap in the first commit: `_errors_are_isolated_fts5` only recognized one wording of the FTS5 corruption message,

```
malformed inverted index for FTS5 table
```

but this SQLite version reports the identical isolated-FTS5 condition with different wording:

```
fts5: corruption found reading blob N from table "embedding_fulltext_search"
```

Since the classifier didn't match that phrasing, `maybe_autoheal_fts5_index` silently declined to heal on any machine running a sufficiently recent SQLite, falling straight through to the hard-abort path — the exact failure this PR exists to avoid. Second commit widens the regex to match both.

That fix also exposed that a few tests in `test_miner_fts5_validation.py` had been passing for the wrong reason: they manufacture the exact "reporter-shaped" isolated-FTS5 corruption and asserted `mine()` must raise — true only because the classifier bug prevented auto-heal from ever engaging on this SQLite version. With the classifier fixed, that corruption is now correctly auto-healed and `mine()` succeeds instead, so those tests needed updated expectations (not their fixtures — the fixtures were fine). Also added a couple of end-to-end companion tests proving the full `mine()`/`mine_formats()` chain — not just the standalone validator — actually heals and succeeds for the isolated case.

One implementation note if it's useful during review: while updating the two exception-passthrough tests (`test_full_chain_raises_through_mine_impl`, `test_mine_impl_does_not_print_partial_summary_on_validation_error`), I initially tried swapping to genuinely non-isolated (page-level) corruption to keep exercising a real raise. That corruption was severe enough that ChromaDB's own Rust bindings panicked just opening the file for the re-mine's `get_collection()` call — a native panic, not a catchable Python exception, before the validator ever ran. Switched those two to a monkeypatched raise instead, since what they're actually testing is the exception-passthrough/banner-suppression behavior in `_mine_impl`, not that arbitrary real corruption reliably reaches it.

Full suite (3218 tests) passes clean on this branch now, `ruff check`/`format` clean. Marking ready for review.

nikkunikku:
**Corroborating this from a production palace — this fix is the right call.**

Hit this exact loop independently on a ~1.4 GB palace (159,065 embedding rows / 1.59M metadata rows). A mine was killed mid-write and left an isolated `malformed inverted index for FTS5 table main.embedding_fulltext_search`. Because `_validate_palace_fts5_after_mine` raised instead of healing, **every subsequent mine re-detected it and aborted** — the hook log showed **278** `malformed inverted index` detections and **278** `ABORT: SQLite-layer corruption detected` banners before I intervened.

A few data points that support the approach here:

- **Base data was fully intact** — `embeddings`, `embedding_metadata`, and the `embedding_fulltext_search_content` shadow table all read fine; only the inverted index was malformed. `PRAGMA quick_check`/`integrity_check` reported *only* the FTS5 line, i.e. exactly the `_errors_are_isolated_fts5` case.
- Running `maybe_autoheal_fts5_index` against it rebuilt the index in place and `quick_check` returned `ok` — no drawer rows touched, full-text `MATCH` queries working again immediately. So wiring it into the mine path (this PR) would have prevented the loop entirely.
- Interesting detail: the first detection fired on a *different* wing than the one that was actually mid-write, and the log itself noted "the corruption may pre-date the mine" — so the post-mine validator is catching pre-existing damage, which is exactly why healing (not just guarding the current mine) is what breaks the loop.

**One suggestion on tests:** the fixture-based FTS5-corruption tests `pytest.skip` on SQLite builds that refuse direct FTS5 shadow-table writes (related to #1925), so on those builds the new behavior isn't actually exercised. Two build-independent tests that stub the integrity primitives cover the wiring deterministically everywhere:

```python
def test_validator_suppresses_raise_when_autoheal_clears(tmp_path, monkeypatch):
    palace = tmp_path / "palace"
    _build_palace_with_drawer(palace)
    fts_err = ["malformed inverted index for FTS5 table main.embedding_fulltext_search"]
    monkeypatch.setattr("mempalace.repair.sqlite_integrity_errors", lambda _p: fts_err)
    heal_calls = []
    def _fake_heal(palace_path, errors, **_):
        heal_calls.append((palace_path, tuple(errors)))
        return []  # healed
    monkeypatch.setattr("mempalace.repair.maybe_autoheal_fts5_index", _fake_heal)
    assert _validate_palace_fts5_after_mine(str(palace)) is None
    assert heal_calls == [(str(palace), tuple(fts_err))]

def test_validator_still_raises_when_autoheal_cannot_clear(tmp_path, monkeypatch):
    palace = tmp_path / "palace"
    _build_palace_with_drawer(palace)
    errs = ["database disk image is malformed"]
    monkeypatch.setattr("mempalace.repair.sqlite_integrity_errors", lambda _p: errs)
    monkeypatch.setattr(
        "mempalace.repair.maybe_autoheal_fts5_index",
        lambda palace_path, errors, **_: errors,  # not cleared
    )
    with pytest.raises(MineValidationError):
        _validate_palace_fts5_after_mine(str(palace))
```

Happy to open these as a follow-up or fold them in if useful.

Issue #1818, opened 2026-06-17
MCP server should detect a peer writer on the same palace (concurrent stdio writers corrupt the Chroma backend)


corrm:
## Summary

A long-lived `mempalace-mcp` stdio server does **not** detect when another `mempalace-mcp` process is already a live writer on the *same* palace. When two or more writers are active against one Chroma palace, the database corrupts — typically surfacing as `malformed inverted index for FTS5 table main.embedding_fulltext_search` (and HNSW `link_lists.bin` damage). This has corrupted my palace repeatedly (4 corrupt snapshots over ~2 weeks).

The proposal: have the MCP **writer** do what the `mine` path already does — detect a peer process on the same palace and refuse/warn instead of silently becoming writer #2.

## Environment

- mempalace 3.4.1 (uv tool install), ChromaDB backend (default)
- Linux, Python 3.13
- MCP client: Claude Code (stdio server)
- Palace: single global `~/.mempalace/palace`, `chroma.sqlite3` ~3.4 GB

## How two writers happen in practice

Two independent, easy-to-hit triggers:

1. **Concurrent sessions.** Each Claude Code session spawns its own `mempalace-mcp` stdio process; all open the one global palace.
2. **Double registration (the subtle one).** mempalace is provided globally by the **plugin** (server `mempalace` → `mcp__plugin_mempalace_mempalace__*`). If a user *also* adds `mempalace` to a project's `.mcp.json`, Claude Code treats them as two distinct servers and launches **two** `mempalace-mcp` processes in that single session. I had this in 7 project `.mcp.json` files without realizing it doubled the writers.

## Why the existing mitigations don't fully cover it

The codebase clearly knows about concurrent-writer corruption and has real defenses:

- `mine_palace_lock` (`palace.py`) — cross-process `fcntl` flock, comment cites the HNSW corruption issues it fixes.
- `ChromaCollection._write_lock` (`backends/chroma.py`) wraps `add`/`upsert`/`update`/`delete` in that lock.
- `_get_client()` (`mcp_server.py`) reconnects the `PersistentClient` on inode/mtime change to catch external writes.

But the gap remains:

- The flock serializes the **write critical section**; it cannot invalidate *another live process's* in-memory `PersistentClient` state (HNSW segment / FTS write buffers). Chroma's `PersistentClient` is single-process by design.
[...]

PR #1932, opened 2026-07-06
fix(repair): treat SQLITE_BUSY as contention, not corruption, in sqlite_integrity_errors

elitedevs:
## Problem

`sqlite_integrity_errors()` runs `PRAGMA quick_check` with no explicit `busy_timeout`. Any peer writer holding the write lock longer than Python's 5 s connect default makes quick_check fail with `database is locked`, and every caller treats that as palace corruption.

The worst-hit caller is the MCP startup integrity gate (#1818): while a long batch write is in flight (curator pass, batch mine — routine on a rollback-journal palace), every new MCP server process reports the palace as corrupt and refuses to serve. Clients retry, each retry spawns another server process that fails the same gate — we hit this in production with 11 accumulated MCP processes and a fully bricked palace during a healthy 15-minute batch write. `mempalace_status` reported `PRAGMA quick_check failed: database is locked` as an integrity failure with 0 drawers visible, while the CLI on the host read all 38k drawers fine.

Contention is not corruption.

## Fix

Set `PRAGMA busy_timeout = 15000` on the quick_check connection — same pattern already used by the sqlite fast paths in `mcp_server.py`. Transient writers are waited out; a genuinely wedged lock (>15 s) still surfaces, which is correct for the repair preflight (it must not proceed to `delete_collection()` when it cannot verify).

## Test

`test_sqlite_integrity_errors_waits_out_transient_writer_lock` holds `BEGIN EXCLUSIVE` from a peer connection for 7 s — deliberately over the 5 s implicit default that masked the bug, under the new 15 s timeout. Verified it fails on `develop` without the fix and passes with it. Full `tests/test_repair.py` suite passes (91), ruff clean.

## Note for maintainers

Longer term you may want the integrity gate to distinguish "couldn't verify (BUSY)" from "verified corrupt" — the repair preflight should abort on both, but the MCP startup gate arguably shouldn't brick clients on the former. This PR keeps the contract unchanged and just makes the check contention-tolerant.


igorls:

PR #1939, opened 2026-07-06
fix: re-key hallways/tunnels on wing-name migration so the graph follows renamed wings

### LINKED ISSUE #1938: bug: wing-name migration re-keys drawers but leaves hallways/tunnels on the old wing
## Summary

`migrate.migrate_wing_names` normalizes the `wing` metadata on drawers, closets, and `topics_by_wing`, but never rebuilds the associative graph. After a migration, `hallways.json` and `tunnels.json` still reference the pre-normalization wing, so the graph describes a wing that holds no drawers.

## Reproduction

Against `main` (`9815f0a`), an isolated test palace (via `MEMPALACE_PALACE_PATH`):

```
BEFORE:  drawers[_alpha, beta]  hallways[_alpha, beta]  tunnels[_alpha, beta]
migrate_wing_names(): '_alpha' -> 'alpha', 2 drawers
AFTER:   drawers[alpha, beta]   hallways[_alpha, beta]  tunnels[_alpha, beta]
```

Seed two `_alpha` drawers (`entities="Alice;Bob"`) and two `beta` drawers (`entities="Alice;Carol"`), run `compute_hallways_for_wing` for each, `entity_tunnels_for_wing` to derive the shared-`Alice` tunnel, then `migrate_wing_names(confirm=True)`. Drawers move to `alpha`; the `_alpha` hallway and the `_alpha↔beta` tunnel remain. (A fix PR with a regression test will follow.)

## Root cause

`migrate_wing_names` (`mempalace/migrate.py`) calls `_apply_wing_updates` (drawers + closets) and `_apply_topics_by_wing_renames`, then returns. The module imports neither `hallways` nor `palace_graph`, so no recompute runs. Hallway ids are `hash(wing, entity_a, entity_b)` (`hallways.py`) and tunnel endpoints embed the wing (`palace_graph.py`), so a rename orphans every affected edge.

## Impact

`list_hallways` / `find_tunnels` / `follow_tunnels` return edges keyed to wings that no longer exist. Any wing-name normalization inherits this — the #1675 leading-separator fix (merged) and migrate-wing (#1453).

## Expected

After a wing-name migration, hallways/tunnels for the affected wings are rebuilt (or re-keyed) so the graph matches the drawers. The natural fix is a graph-rebuild step invoked after the re-key — a small `rebuild_graph_for_wings`-style primitive that purges the affected records, recomputes hallways, and re-derives tunnels for the touched wings.

ggettert:
## Summary

Fixes #1938. `migrate_wing_names` re-keyed drawer/closet/`topics_by_wing` metadata but never touched the associative graph, so `hallways.json` and `tunnels.json` kept pointing at the pre-normalization wing.

## What changed

- `hallways.rekey_hallway_wings(rename_map)` — rewrites hallway `wing` + regenerates ids; merges records that collapse onto the same `(wing, entity-pair)`.
- `palace_graph.rekey_tunnel_wings(rename_map)` — rewrites tunnel endpoint wings + regenerates canonical ids; drops tunnels whose endpoints coincide after the rename; merges id collisions. Covers entity and topic tunnels (shared store).
- `migrate_wing_names` calls both after the metadata re-key, using the same rename map, and reports counts.
- Extracted `_hallway_label` so freshly-computed and re-keyed/merged records format identically (no behavior change).

## Why re-key, not recompute

A wing-name migration is a pure rename — drawer membership and entity co-occurrence are unchanged. Re-keying is sufficient and preserves the #1578 dynamics fields (strength/stability/last_activated/access_count); a recompute from drawers would reset them. Collision merges keep the stronger/more-recent dynamics.

## Tests

- Unit: rename regenerates ids + preserves dynamics; collision merges counts/rooms/dynamics; untouched wings/tunnels left alone; self-referential tunnels dropped.
- Integration: seed `_alpha`+`beta` drawers, build hallways/tunnels, run migration → graph moves to `alpha`; two-wings-into-one merges to a single hallway with summed count.

## Known limitations

- The re-key runs after the drawer/closet/`topics_by_wing` re-key without a shared transaction across the four stores. If the graph step fails mid-way, the drawer re-key is intentionally **not** rolled back (a successful data migration shouldn't be undone because a derived index failed); the failure is surfaced as a warning and the operation is **re-runnable** — a subsequent migrate re-keys any orphaned graph records. Full cross-store atomicity is out of scope for this fix.

Fixes #1938.

### DIFF
--- a/mempalace/hallways.py
+++ b/mempalace/hallways.py
@@ -182,6 +182,18 @@ def _parse_entities(value) -> list[str]:
     return result
 
 
+def _hallway_label(entity_a: str, entity_b: str, count: int, rooms: list) -> str:
+    """Human-readable hallway label. Single source of truth so freshly-computed
+    and re-keyed/merged records format identically."""
+    room_summary = ", ".join(rooms[:3]) if rooms else "(no room tags)"
+    if len(rooms) > 3:
+        room_summary += f", +{len(rooms) - 3} more"
+    return (
+        f"{entity_a} \u2194 {entity_b} (co-occur in {count} drawers across "
+        f"{len(rooms) or 'no'} room{'s' if len(rooms) != 1 else ''}: {room_summary})"
+    )
+
+
 def _hallway_id(wing: str, entity_a: str, entity_b: str) -> str:
     """Deterministic id derived from wing + sorted entity pair.
 
@@ -341,9 +353,6 @@ def compute_hallways_for_wing(
             continue
         entity_a, entity_b = key
         rooms = sorted(pair_rooms.get(key, set()))
-        room_summary = ", ".join(rooms[:3]) if rooms else "(no room tags)"
-        if len(rooms) > 3:
-            room_summary += f", +{len(rooms) - 3} more"
         record = {
             "id": _hallway_id(wing, entity_a, entity_b),
             "wing": wing,
@@ -351,7 +360,7 @@ def compute_hallways_for_wing(
             "entity_b": entity_b,
             "co_occurrence_count": count,
             "rooms": rooms,
-            "label": f"{entity_a} ↔ {entity_b} (co-occur in {count} drawers across {len(rooms) or 'no'} room{'s' if len(rooms) != 1 else ''}: {room_summary})",
+            "label": _hallway_label(entity_a, entity_b, count, rooms),
             "created_at": created_at,
             "created_by": "auto",
         }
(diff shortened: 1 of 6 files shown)


ggettert:
Rebased this onto current `develop` after v3.6.0 and updated the graph re-key path to pass the explicit `MempalaceConfig(palace_path=...)` through to hallway/tunnel side-files. Local focused tests and full GitHub CI are green.

Issue #1497, opened 2026-05-13
Question: should MemPalace recommend a single-writer/gateway pattern for multi-agent setups?


isha-sharad:
Hi. We are trying to run MemPalace in a repo where multiple AI agents may be active at the same time:

- Codex Desktop
- Claude Code
- parallel subagents
- occasional manual/session mining
- Stop / PreCompact style hooks

Our main concern is not retrieval quality, but write safety.

## What feels broken / risky

The current architecture seems easy to make unsafe when more than one writer path exists for the same palace:

1. multiple clients can write through MCP at the same time
2. hooks can trigger writes or mining
3. manual `mempalace mine` can run while MCP is active
4. more than one `mempalace-mcp` process can accidentally point at the same palace
5. Chroma/HNSW corruption issues appear in several related reports

Relevant issues we looked at include #1229, #1092, #1202, #1253, #1443, #1454, #1096, and #941.

We may be misunderstanding the current architecture, but our reading is: `mcp-proxy` helps avoid each client spawning its own server, but it does not by itself guarantee that all palace mutations are serialized.

## What we are considering

We are considering a stricter "single writer" deployment pattern:

```text
Codex / Claude / subagents
        |
        v
MemPalace gateway MCP
        |
        +-- read tools -> shared mempalace-mcp/proxy
        |
        +-- write tools -> queue
                         |
                         v
                 single writer worker
                         |
                         v
                 one palace directory
```

The idea would be:

- agents still decide what is worth saving
- write tools enqueue jobs instead of mutating the palace directly
- one worker processes queued writes sequentially
- bulk mining and repair commands use the same lock
- hooks never call `mempalace mine`, `mempalace-mcp`, or direct palace writes
- Codex/Claude only connect to the gateway, not raw `mempalace-mcp`

This is only an option we are exploring. It is heavier operationally, and we would prefer not to build it if there is a simpler supported approach.

## Questions

1. Is this single-writer/gateway model a reasonable direction for multi-agent use?
2. Is there already a recommended way to safely run MemPalace with multiple agents writing to the same palace?
3. Does `mcp-proxy` plus one long-lived `mempalace-mcp` process already serialize writes enough, or can concurrent MCP tool calls still race internally?
4. Should `mempalace mine`, hooks, and MCP writes share a palace-wide lock?
5. Has anyone solved this in practice with Chroma HTTP mode, a separate Chroma server, or another deployment pattern?
[...]

Issue #1172, opened 2026-04-24
PreCompact hook unconditionally blocks /compact, making it unusable


chrisreeves48:
## Summary

The `PreCompact` hook in `hooks_cli.py:hook_precompact()` outputs `{"decision": "block"}` on every invocation, with no conditional logic and no pass-through path. Because Claude Code cancels the `/compact` action outright when a PreCompact hook blocks (the user never sees a mid-conversation reminder the way Stop blocks surface), there is no mechanism for the user or the assistant to ever get past the hook. The net effect: any user with the mempalace plugin enabled cannot ever successfully run `/compact`.

## Root cause

`mempalace/hooks_cli.py:193-216` (as of v3.0.14):

```python
def hook_precompact(data: dict, harness: str):
    """Precompact hook: always block with comprehensive save instruction."""
    parsed = _parse_harness_input(data, harness)
    session_id = parsed["session_id"]

    _log(f"PRE-COMPACT triggered for session {session_id}")

    # Optional: auto-ingest synchronously before compaction (so memories land first)
    mempal_dir = os.environ.get("MEMPAL_DIR", "")
    if mempal_dir and os.path.isdir(mempal_dir):
        try:
            ...
            subprocess.run([sys.executable, "-m", "mempalace", "mine", mempal_dir], ...)
        except OSError:
            pass

    # Always block -- compaction = save everything
    _output({"decision": "block", "reason": PRECOMPACT_BLOCK_REASON})
```

The docstring is explicit: *"Precompact hook: always block with comprehensive save instruction."* But there is no corresponding mechanism for letting compaction proceed afterward — the block message says *"Save everything to MemPalace, then allow compaction to proceed"* but there is no path to "allow."

## Why this differs from the Stop hook

The `Stop` hook has conditional blocking logic at `hook_stop()` (~line 134): *"block every N messages for auto-save."* When Stop blocks, the assistant receives a system-reminder mid-conversation, responds with MemPalace tool calls, and the next Stop doesn't block. That loop works.

`/compact` can't benefit from that loop because it's a local slash command — when the PreCompact hook blocks, the compact operation cancels immediately, so the assistant never gets a turn to respond and unblock.

## Reproduction

1. Enable the `mempalace@mempalace` plugin in Claude Code.
2. Start any session; have any amount of conversation.
3. Run `/compact`.
4. Compact fails with:
   ```
[...]

PR #1940, opened 2026-07-06
feat(hooks): add hooks.transcript_mining config knob

Lcstyle:
## Summary

Adds a `hooks.transcript_mining` config knob (env: `MEMPALACE_HOOKS_TRANSCRIPT_MINING`) that lets operators disable the in-hook transcript-mining subprocess while keeping the save-reminder / silent-diary-checkpoint behavior. Zero-breaking-change default (mining stays on).

## Motivation

Multi-agent deployments (multiple concurrent Claude Code / Codex sessions filing through the same palace) currently have to choose between:

1. **Full hooks on** — every hook fire can spawn a transcript-mine subprocess. Under 8+ concurrent sessions, mines pile up in the daemon queue (#1497) and interactive MCP writes time out waiting behind them. Or, without a daemon, hooks contend directly for the palace storage lock.
2. **`hooks.auto_save: false`** — turns off both the reminder AND the mine. Operator loses the assistant-nudge that drives interactive saves.

The gap: users want reminders **without** the mine. Downstream has been carrying out-of-tree patches for months:

- @anastasiiaanfimova's [`hooks_cli` mine-off gist patch](https://github.com/MemPalace/mempalace/issues/1497#issuecomment-4458410561) — literally an early `return` in `_maybe_auto_ingest` / `_mine_sync`.
- @jphein's [`palace-daemon`](https://github.com/techempower-org/palace-daemon) at the gateway layer — same principle, gateway-scale ("For anything heavy (transcript ingestion), append to a plain text file from the hook and let a batch process pick it up later").

This PR upstreams the pattern as a first-class config knob so no fork is required.

## Design

- **New property** `MempalaceConfig.hooks_transcript_mining` (defaults to `True` → matches pre-PR behavior).
- **Env override** `MEMPALACE_HOOKS_TRANSCRIPT_MINING={false,0,no,true,1,yes}`.
- **Master gate** `hooks.auto_save=false` still masks it (backwards compat: existing `auto_save: false` deployments stay silent).
- **Env override wins over the auto_save mask** — mirrors how `MEMPALACE_HOOKS_AUTO_SAVE` already behaves.

Belt-and-suspenders guards short-circuit at the top of each mining entry point rather than at each call site — one source of truth, no risk of a future caller forgetting the outer gate:

- `_maybe_auto_ingest` (async background MEMPAL_DIR mine)
- `_mine_sync` (synchronous precompact mine)
- `_ingest_transcript` (transcript → convo drawer ingest)

## Composes with #1826's daemon

With mining out of the hook path, the daemon queue only holds ~200ms interactive writes — the daemon becomes viable for multi-session deployments without the mine-starvation problem that motivated the workarounds above.

Recommended multi-agent config unlocked by this PR:

```json
{
  "hooks": {
    "auto_save": true,
    "silent_save": false,
    "transcript_mining": false,
    "daemon": true
  }
}
```

- Stop/PreCompact still fire and emit save reminders
- Assistant responds by calling `mempalace_diary_write` / `mempalace_add_drawer`
- Interactive writes serialize through the daemon (safe under N concurrent sessions)
- No mining subprocess is ever spawned by a hook fire
- Bulk mining runs out-of-band via a scheduled `mempalace mine --daemon` when no session is active

## Tests

- **Config truth table** (8 tests in `tests/test_config.py`): default / from-config / four env overrides / `auto_save`-mask / env-wins-over-mask.
- **Guard tests** (4 tests in `tests/test_hooks_cli.py`): each entry point verified as no-op when `transcript_mining=false`, plus a regression test that mining still fires when `transcript_mining=true` (default).
- **All 275 existing tests** in `test_hooks_cli.py` + `test_config.py` continue to pass.

```
275 passed, 1 skipped in 8.80s
```

## Test plan

- [ ] `uv run pytest tests/test_config.py tests/test_hooks_cli.py -v`
- [ ] `uv run ruff check .` + `uv run ruff format --check .` clean
- [ ] Manual: set `hooks.transcript_mining: false` on a live deployment, confirm hooks fire but no `_spawn_mine` subprocess appears; confirm assistant still gets Stop reminder and calls MCP tools; confirm scheduled `mempalace mine` still populates the palace at 04:00.

## Deliberately out of scope

- **`PRECOMPACT_BLOCK_REASON` wiring** — the constant is defined at `hooks_cli.py:110` but never emitted (`hook_precompact` at line ~1353 always does `_output({})`). Wiring it up would revisit #955 / #1172 (PreCompact-block-caused-deadlock) which were closed with a specific "don't block on PreCompact" fix. That's a separate design decision better handled in a follow-up PR that specifically considers those closed issues.

Related: #1497 (multi-writer safety discussion), #1826 (write daemon), #1828 (daemon hardening).

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -486,6 +486,36 @@ class MempalaceConfig:
         return hooks.get("auto_save", True)
 
     @property
+    def hooks_transcript_mining(self):
+        """Whether hooks may spawn detached transcript-mining subprocesses.
+
+        Transcript mining (via ``_ingest_transcript`` / ``_maybe_auto_ingest``
+        / ``_mine_sync``) is heavyweight. Under multi-writer deployments
+        (multiple concurrent Claude Code / Codex sessions filing through
+        the same palace), an in-hook mine can starve interactive MCP writes
+        queued behind it in the daemon (#1497), or contend with concurrent
+        processes for the palace storage lock.
+
+        When ``False``, hooks emit reminders (governed by ``hooks.silent_save``)
+        but never spawn mining. Bulk mining can then be scheduled out-of-band
+        via a systemd timer / cron job that runs when no session is active —
+        the pattern @anastasiiaanfimova ships as an out-of-tree ``mine off``
+        patch and @jphein's palace-daemon replicates at gateway layer.
+
+        Master ``hooks.auto_save`` still gates this: if auto-save is off,
+        this is always effectively off.
+
+        Env override: ``MEMPALACE_HOOKS_TRANSCRIPT_MINING``.
+        """
+        env_val = os.environ.get("MEMPALACE_HOOKS_TRANSCRIPT_MINING")
+        if env_val is not None:
+            return env_val.lower() not in ("false", "0", "no")
+        if not self.hooks_auto_save:
+            return False
+        hooks = self._file_config.get("hooks", {})
+        return hooks.get("transcript_mining", True)
+
+    @property
     def topic_wings(self):
         """List of topic wing names."""
         return self._file_config.get("topic_wings", DEFAULT_TOPIC_WINGS)
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -588,7 +588,15 @@ def _maybe_auto_ingest():
     target gets its own PID slot, so distinct targets never block each
     other but a re-fire of the same target while the previous one is
     still running is silently skipped.
+
+    Respects ``hooks.transcript_mining`` — when disabled, returns
+    immediately without spawning any mine subprocess.
     """
+    try:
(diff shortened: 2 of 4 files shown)


Lcstyle:
Reporting dogfood experience with this knob, in case it helps the PR land.

We have been running this change as a local patch on several hosts since it was
opened. The operational need it addresses is real: on hosts running many
concurrent agent sessions, hook-triggered mining is the dominant write source, and
without a switch there is no way to separate "record the session" from "mine the
transcript" once the latter becomes expensive.

No regressions observed in that time. The knob defaults to existing behaviour, so
adopting it is a no-op for anyone who does not set it.

Would be glad to see this merged, and can provide before/after mine timings from a
large palace if that would help.

PR #1941, opened 2026-07-06
feat(hermes): scan-based dedup safety nets for session_end and pre_compress

raman325:
## What does this PR do?

Follow-up to #1915 (now merged), restoring the `on_session_end` / `on_pre_compress` capture that PR removed — this time as **scan-based safety nets** that file only turns `sync_turn` missed, instead of blind re-filing that duplicated every turn.

Rebased onto `develop` now that #1915 has landed, so every commit here is safety-net work and the diff stands on its own. Ready for review.

cc @GoXLd — you offered to help with this follow-up; review and testing very welcome.

### Design

- **`sync_turn` stays the primary filing path**, and now stamps a `turn_fp` fingerprint into drawer metadata: `make_turn_fingerprint(normalize(raw user message))`, computed from the raw `messages` snapshot Hermes passes — never from the cleaned positional args. This is the correlation key that survives representation differences (skill injection, tool traffic, think-stripping make exact-text matching structurally impossible — verified against the Hermes source).
- **`on_session_end` / `on_pre_compress`** enqueue the raw window plus a snapshot of `/branch` lineage; the worker segments it into turns anchored at real user messages (tool-result-only user messages are not anchors), scans the session's (and ancestors') existing drawers into a multiset of `(turn_fp, text)` **pairs**, and files only uncovered segments.
- **Atomic pair consumption, three passes**: exact `(fp, text)` matches for every segment first (so a donor arm can never steal a drawer an exact match needs), then fingerprint groups, then exact-text groups. Every skip consumes exactly one physical drawer across all indexes — over-skipping (data loss) is structurally impossible. When same-fingerprint occurrences outnumber their drawers, **all are filed**: bounded duplication of a covered turn over any possible loss.
- **Loss-safe fallbacks everywhere**: empty session_id → blind-file (the shared `unknown` bucket would false-dedup across sessions); scan failure → blind-file with a warning; `session_end` uses a bounded blocking enqueue (no retry exists after session end); loss paths log at `warning` like the duplication paths.
- **Branch lineage**: `/branch` appends the parent session to the scan scope (transcript carries forward); `/resume` clears it (its `parent_session_id` is the session being *left* — stale lineage would false-dedup); rewind keeps it.
- Interrupted turns — which Hermes deliberately never syncs (NousResearch/hermes-agent#15218) — are captured by the safety net, honoring verbatim/100%-recall.

### Integration with the post-#1915 Chroma serialization (`94a41ee`)

`develop` picked up `fix(hermes): stop second Chroma client racing the filing worker` after #1915 merged, which put every access to the long-lived Chroma client behind `self._collection_lock`. This branch is rebased on top of that, and the safety net now honors the same invariant:

- `_file_turn` takes the lock **per retry attempt**, with the transient-lock backoff sleep held *outside* the lock — the lock only serializes in-process access, so holding it while waiting out a cross-process sqlite lock would stall readers for no benefit.
- `_scan_filed` runs under the lock for the whole scan, released before the filing loop (the lock is not reentrant and `_file_turn` re-takes it).
- The wake-up cache refresh at `session_end` is routed through the worker queue **after** the safety-net task, so that reader can never race the safety net's writes. It clears `_wake_up_done` at enqueue time so waiters observe "refresh pending" across the whole queued interval, not just while the refresh runs.

### Known, accepted limit (documented + pinned by test)

Fingerprint coverage cannot see occurrences that compression removed from the window: a byte-identical prompt ("continue") recurring after its synced occurrence was compressed away gets fp-skipped, and if the recurrence was interrupted, its partial *assistant* output is not captured (the user's words are already stored). Undecidable with occurrence counting alone — Hermes message dicts carry no per-message identity. `test_truncated_window_fp_coverage_is_accepted_limit` pins this so any future correlation improvement surfaces loudly.

## How to test

```
uv run pytest tests/test_hermes_integration.py -v   # 90 tests; the dedup suite starts at "Dedup safety net"
uv run pytest tests/ -v --ignore=tests/benchmarks   # full suite
```

The interesting tests: `test_session_end_does_not_refile_synced_turns`, `test_session_end_skips_tool_turns_covered_by_sync` (the representation-mismatch case), `test_interrupted_then_retried_identical_prompt_keeps_both` (the greedy-steal regression), `test_contended_fingerprint_files_all_occurrences` (file-all-when-contended policy), `test_double_session_end_is_idempotent`, `test_branch_session_end_skips_parent_filed_turns`.

## Checklist
- [x] Tests pass (`uv run pytest tests/ --ignore=tests/benchmarks`)
- [x] No hardcoded paths


GoXLd (commented):
Ran the full suite on 325d407..c4fc9ad: 90/90 hermes tests, 3313 passed / 20 skipped full suite, ruff clean.

Went through the three-pass dedup algorithm in _file_missing_exchanges line by line — the exact-pair-before-fingerprint-donor ordering and the atomic pair-consumption invariant hold up under the contention scenarios (test_contended_fingerprint_files_all_occurrences, test_interrupted_then_retried_identical_prompt_keeps_both).

Also cross-checked the on_session_switch lineage tracking against the actual hermes-agent source (hermes_cli/cli_commands_mixin.py): the reason="branch"/reason="resume" handling matches the real call sites exactly, including that /resume's parent_session_id is the session being left (not an ancestor) — good catch keeping that out of the lineage. And the "interrupted turns never sync" premise is real: run_agent.py's _sync_external_memory_for_turn returns early on interrupted, referencing hermes-agent#15218 — so this safety net earns its keep.

No correctness issues found. LGTM.

raman325:
CI failure appears to be unrelated to this PR

PR #1942, opened 2026-07-06
feat(hermes): backfill, install command, and docs

raman325:
Second half of the #1684 split per @igorls — backfill, the `mempalace hermes install` command, and docs. (Moved here from raman325#1 so it's visible upstream.)

**Stacked on #1941.** The tool-turn fix below reuses #1941's `_segment_turns` / `_compose_exchange_text`, so this has a real code dependency on the dedup branch and sits on `feat/hermes-filing-dedup`. #1915 has merged and the stack is rebased onto current `develop`; GitHub can only diff against a branch in this repo, so the commit list shows #1941's commits too — **only the last four are new here** (`61bdb2f`, `c02e2f6`, `974c1ff`, `8453bda`, everything from `feat(hermes): backfill, install command, and docs` onward). Merge #1941 first and this collapses to just those four.

## What this adds

One command after `pip install mempalace`:

```bash
mempalace hermes install
```

Files land in `~/.hermes/plugins/mempalace/` (the canonical user-installed plugin directory Hermes' `discover_memory_providers()` scans) with a `plugin.yaml` manifest. `~/.hermes/config.yaml` gets `memory.provider: mempalace`. The command pip-installs mempalace into the Hermes venv, and optionally backfills existing Hermes session exports into the palace (`--skip-backfill` to opt out, `--yes` for scripted installs).

Also: `mempalace/integrations/hermes/README.md` (install, hooks, tool surface, and why the ChromaBackend path avoids dimension mismatches) and a top-level README pointer. `pyyaml` becomes a runtime dependency for the `config.yaml` edit.

## How the original review blockers were addressed

- **Live/historical ingest parity** — `backfill.file_exchange` routes through `convo_miner.file_conversation_exchange()`, the exact write path the live provider's `_file_turn` uses, so historical and live drawers carry identical routing, normalization, and metadata. `authored_at` is stamped from the session file's mtime, so date filters reflect when the conversation happened, not when the backfill ran.
- **One wing-routing implementation** — `backfill.classify_wing` delegates to the provider's `_match_wing_by_keywords` instead of maintaining a hand-synced copy; a parity test guards that the delegation survives the `spec_from_file_location` loading mode the install flow uses.
- **No backend bypass** — backfill opens its collection once via `ChromaBackend.get_or_create_collection()` (same construction path as the runtime provider), never a raw `PersistentClient`.
- **Re-run idempotency** — each session file is guarded with `palace.file_already_mined(..., extract_mode="exchange")` — the same per-source check the convo miner uses — so re-running `mempalace hermes install` doesn't duplicate previously backfilled exchanges. Regression test runs backfill twice and asserts the drawer count holds.

## Review findings addressed

Both of @GoXLd's findings are fixed — details in the commit message for `8453bda`. Tool-using turns now segment through the live provider's `_segment_turns` (final answer + tool markers preserved, `_normalize_content` flattening, one implementation); the `.jsonl` parser unwraps `hermes sessions export`'s session-per-line shape while still accepting message-per-line transcripts; and `file_exchange` composes drawer text via `_compose_exchange_text` and keeps assistant-only preamble segments. Regression tests use the reported repro shapes verbatim.

## Known gap, deliberately left for a fast-follow

Live turns use `source_file="hermes-session:{id}"` while backfill uses the real session file path, so a session that was partially live-filed and later backfilled does not cross-dedup.

#1941 (directly beneath this PR) supplies the machinery to close it: `turn_fp` fingerprints and the `(turn_fp, text)` pair-scan (`_scan_filed` / `_file_missing_exchanges`). Backfill could compute `turn_fp` per exchange and consult the `hermes-session:{id}` scan before filing, closing the overlap for sessions whose id is derivable from the transcript filename. That is kept out of this PR to hold its scope to backfill/install/docs — it is a natural follow-up once both land, not a blocker for either.

## Tests

28 tests in `tests/test_hermes_install_cli.py`: the path-resolution helpers (`--hermes-home` guardrails, venv interpreter discovery, palace path precedence), atomic `config.yaml` editing (including the scalar-`memory:` and malformed-YAML cases), wing-routing parity, and the backfill re-run guard.

```
uv run pytest tests/test_hermes_install_cli.py tests/test_hermes_integration.py -v   # 118 tests
uv run pytest tests/ -v --ignore=tests/benchmarks                                    # full suite
```


GoXLd:
Ran the full suite (82/82 hermes tests, 3301 passed / 20 skipped full suite, ruff clean) — but coverage doesn't reach the two issues below; both reproduce with a few lines of code.

**1. `_messages_to_exchanges` drops the assistant's real answer on any tool-using turn.**

It pairs `messages[i]` (user) with `messages[i+1]` only if `role == "assistant"` — but a tool-calling turn looks like `user → assistant(tool_use) → user(tool_result) → assistant(final text)`. The function grabs the tool_use message (usually no `"text"` block) as "the" assistant side and never looks past the tool_result to the real final answer:

```python
_messages_to_exchanges([
    {"role": "user", "content": "check the weather and tell me if I need an umbrella"},
    {"role": "assistant", "content": [{"type": "tool_use", "id": "1", "name": "get_weather", "input": {}}]},
    {"role": "user", "content": [{"type": "tool_result", "tool_use_id": "1", "content": "rainy"}]},
    {"role": "assistant", "content": [{"type": "text", "text": "Yes, bring an umbrella - it's rainy."}]},
])
# -> [{'user': '...', 'assistant': ''}]   # the real answer is gone
```

`file_exchange()` only guards on `user_msg` being non-empty, so these land in the palace with an empty assistant side. Since Hermes is a coding agent, most real turns involve tool calls — this isn't an edge case, it's the common case, and it's the exact "verbatim always" guarantee this project is built around. #1941's `_segment_turns` already solved this correctly (fold everything between real user anchors into one assistant side); backfill needs the same treatment, or should reuse that helper.

**2. The `.jsonl` parser doesn't match the shape `hermes sessions export` actually produces.**

`export_session`/`export_all` (`hermes_state.py`) write one JSON object **per session** per line (`{**session, "messages": [...]}`) — confirmed against `hermes_cli/main.py`'s `export` action. `parse_session_file`'s `.jsonl` branch treats each line as a message dict directly (`_messages_to_exchanges(messages)` over the raw parsed lines), so a real export file yields zero exchanges silently:

```python
f.write_text(json.dumps({"id": "sess-123", "messages": [
    {"role": "user", "content": "let's plan the Q3 roadmap"},
    {"role": "assistant", "content": "here is the plan"},
]}) + "\n")
parse_session_file(f)  # -> []
```

Notably the `.json` (non-jsonl) branch already unwraps this shape (`data.get("messages", data.get("turns", []))`) — the `.jsonl` branch just never got the same treatment.

Neither path is covered by a test that exercises real message shapes — `test_backfill_rerun_does_not_duplicate` only uses flat `{"role": "user", "content": "<plain string>"}` pairs.

GoXLd:
Quick follow-up on those **_two_** — went and dug through the history a bit out of curiosity, since neither one looked like something that would've slipped in during a rebase xD (`xg-gh-25` and `igorls` on #1684) 

Mentioning it mostly so it's not read as "something #1942 broke" — it's more like a blind spot that's been there since day one. Given #1941 already built exactly the right primitive for problem 1 (`_segment_turns`), and problem 2 is a one-line unwrap once the `.jsonl` case is treated like the `.json` case already is, neither should be a big lift - happy to take a pass at a patch if that's useful.

raman325:
thanks for the review! I think it's easier/quicker to attempt a patch myself and to get another round of reviews from you. working on it now!

raman325:
Both confirmed and fixed in 93448dc — thank you for the repros, they landed as regression tests almost verbatim, and thanks especially for checking the history so this didn't read as rebase fallout.

1. **Tool-turn answer loss**: `_messages_to_exchanges` now delegates to the live provider's `_segment_turns`, exactly as you suggested — the final answer, tool_use markers, and tool results all fold into the turn's assistant side via `_normalize_content`. One segmentation implementation, same rule this PR already applied to wing routing. (This adds a real dependency on #1941, so the branch is stacked on `feat/hermes-filing-dedup` now.)
2. **`.jsonl` shape**: the parser unwraps session-per-line objects (`messages`/`turns` keys) exactly like the `.json` dict branch, and still accepts message-per-line transcripts.

While in there, two adjacent parity gaps got the same treatment: `file_exchange` now composes drawer text via `_compose_exchange_text` (byte-identical to live filing — which is what #1941's exact-text dedup arm compares against, so this also sets up the backfill↔live cross-dedup fast-follow), and assistant-only preamble segments are kept instead of dropped.

Your offer to patch was appreciated — the fix wanted to land with the stacking decision it forced, but if you want to take the cross-dedup fast-follow (backfill computing `turn_fp` + consulting the `hermes-session:{id}` scan) once this stack merges, it's yours.

vavush:
Hey @igorls — check-in now that #1915 and #2215 landed. The remaining stack (#1941 + #1942) is now the only path to a usable provider: v3.7.0 shipped without it, so `pip install mempalace` still exposes nothing to Hermes users until these two land. Both are rebased on current develop, GoXLd's July reviews are addressed, and the diff is scoped (4 files dedup nets / 9 files install+backfill+docs). One heads-up: #1941's CI was flagged as failing earlier this week — raman325 says it looks unrelated to the PR; worth confirming it's green before merge. Anything else needed from the stack?

raman325:
Rebased onto `develop` (090a26d), stacked on the rebased #1941. The `mempalace/cli.py` additions moved into a new `mempalace/cli/cmd_hermes.py` fragment (registered before `parser.py` in the load order) with the `hermes install` subparser and dispatch in `mempalace/cli/parser.py`. No logic changes; hermes, ids, and cli test suites pass locally, ruff clean.

PR #1950, opened 2026-07-07
feat: per-drawer salience (dynamics extended to drawers)

### LINKED ISSUE #1921: feat: per-drawer salience dynamics for retrieval frequency and recency
## Problem / motivation

Agent-driven "dreaming" pipelines need a first-class way to identify drawers that
are valuable enough to keep versus cold enough to forget. The planned `prune`
task wants a value signal like:

```text
v(d) = usage-frequency × recency-of-use × KG-degree
```

Today the pipeline can compute KG-degree, but it cannot ask MemPalace "how often
has this drawer actually been retrieved?" or "when was it last useful?" As a
result, prune must proxy usage with weaker signals: redundancy, KG source-degree,
ephemeral marker text, or host-specific session-store evidence.

A native per-drawer usage counter would make salience-based forgetting a
first-class MemPalace capability, not a host-coupled heuristic. It would also
benefit general ranking/reinforcement: hot drawers can be surfaced, cooled
drawers can be de-emphasised, and consumers can make explicit salience decisions.

## Current state

`mempalace/dynamics.py` already implements the right cognitive math:

- Hebbian potentiation: strength grows on co-access.
- Ebbinghaus exponential decay: strength fades with time since last activation.
- Cepeda spacing effect: stability grows with spaced reinforcement.

The existing fields are:

```text
strength: float
stability: float
last_activated: str
access_count: int
```

`initialize_dynamics_fields(connection, *, now=None)` safely backfills missing
fields, and `potentiate(connection, ...)` increments `access_count`, refreshes
`last_activated`, and grows `strength` / `stability` using the existing model.

The limitation is scope: the module docstring says it is "Living-connection math
for halls + tunnels", and callers live in `hallways.py` / `palace_graph.py`.
Those are palace-graph **edges**. Drawers are not currently treated as
strengthened/decayed entities.

On drawer write, `mempalace_add_drawer` stores metadata such as:

```text
wing
room
source_file
added_by
filed_at
id_recipe
```

There is no drawer-level `access_count`, and `filed_at` is filing time rather
than last retrieval time. The `mempalace_search` path returns search results but
does not currently potentiate drawer metadata when a drawer is surfaced.

## Proposed change

Extend the existing dynamics model from connections to drawers.

1. **Store dynamics fields in drawer metadata**
   - Add `strength`, `stability`, `last_activated`, and `access_count` to drawer
     metadata.
   - Use `filed_at` as the natural fallback for `last_activated` when lazily
     initialising old drawers.
   - Keep the fields additive and default-safe: existing drawers without these
     fields are neutral until first read/backfill.

2. **Potentiate on retrieval**
   - When `mempalace_search` returns/surfaces a drawer, call the same dynamics
     helper on that drawer's metadata.
   - A returned drawer increments `access_count`, updates `last_activated`, grows
     `strength`, and grows `stability` only when reinforcement is spaced, exactly
     like existing hall/tunnel dynamics.
   - For chunked drawers, update either the logical parent salience or all
     physical chunks consistently; prefer a logical parent-level salience if the
     search result is rejoined before being returned to the client.

3. **Expose a read path**
   - Include salience fields in `mempalace_search` / `mempalace_get_drawer`
     results, or add a small read helper such as `mempalace_drawer_salience`.
   - Consumers should be able to rank or filter by `access_count`,
     `last_activated`, decayed `strength`, and/or an already-decayed salience
     snapshot.

This reuses code MemPalace already has instead of inventing a second scoring
model for drawers.

## API sketch

Additive metadata shape:

```jsonc
{
  "id": "drawer_id",
  "wing": "copilot-mempalace",
  "room": "dreaming",
  "filed_at": "2026-07-03T12:34:56",
  "strength": 1.15,
  "stability": 1.2,
  "last_activated": "2026-07-03T20:12:00+00:00",
  "access_count": 3
}
```

Possible search result shape:

```jsonc
{
  "text": "...",
  "wing": "...",
  "room": "...",
  "id": "drawer_id",
  "distance": 0.31,
  "salience": {
    "strength": 1.15,
    "stability": 1.2,
    "last_activated": "2026-07-03T20:12:00+00:00",
    "access_count": 3
  }
}
```

Optional dedicated read tool:

```text
mempalace_drawer_salience(
    wing: str | None = None,
    room: str | None = None,
    limit: int = 100,
    order_by: "strength" | "access_count" | "last_activated" = "strength",
) -> {
  "drawers": [
    {"id": "...", "wing": "...", "room": "...",
     "strength": 1.15, "stability": 1.2,
     "last_activated": "...", "access_count": 3}
  ]
}
```

Implementation note: if search needs to remain strictly read-only in some modes,
the potentiation write could be guarded behind config or queued through the
existing write path. The important semantic event is "drawer was surfaced to a
consumer", not "query was issued".

## Alternatives considered

- **Use the host session store as the usage oracle.** This works for Copilot
  dreaming prototypes, but it couples MemPalace pruning to one harness's
  telemetry schema. MemPalace should know which drawers it returned without
  requiring external session databases.

- **Keep proxy-only salience.** Redundancy, KG source-degree, and text markers are
  useful hints, but they are lossy. A drawer can be frequently retrieved without
  having high KG degree, and a highly connected drawer can be stale or unused.

- **Add a separate drawer scoring model.** This would duplicate
  `dynamics.py`. Reusing the existing Hebbian/Ebbinghaus/Cepeda implementation
  keeps connection and drawer reinforcement semantics aligned.

## Backward compatibility

Additive only:

- Existing drawers continue to work without salience fields.
- `initialize_dynamics_fields` can lazily backfill missing drawer fields, using
  `filed_at` as the drawer analogue of connection `created_at`.
- Existing search callers can ignore the extra metadata.
- Existing hall/tunnel dynamics remain unchanged.

## Scope / non-goals

- This proposal does **not** define pruning policy or delete any drawers.
- It does **not** make search ranking automatically salience-first; exposing the
  signal is enough for consumers to opt in.
- It does **not** require raw embedding export or host telemetry access.
- It does **not** replace KG-degree; it supplies the missing usage-frequency and
  recency terms so consumers can combine them explicitly.

et1975:
## What does this PR do?
Adds per-drawer salience by reusing the existing dynamics model (Hebbian potentiation, Ebbinghaus decay, and Cepeda spacing) for drawer metadata as well as graph connections. Drawer reads now expose a lazy-decayed `salience` block from copied metadata, with `filed_at` used as the drawer fallback for `last_activated` initialization.

This is backward-compatible and default-safe: metadata fields are additive, search/get read paths remain read-only by default, and retrieval potentiation is opt-in only via `MEMPALACE_SALIENCE_POTENTIATE=true`. When enabled, search updates drawer metadata only (no re-embedding), skips writes on read-only servers or when the MCP writer lock is unavailable, and keeps chunked logical drawers consistent across all physical chunks.

Also adds the read-only `mempalace_drawer_salience` tool and wires it through `TOOLS` and `service.READ_TOOLS`.

Closes #1921

## How to test
- `uv run pytest tests/test_drawer_salience.py -v` → 10 passed
- `uv run pytest tests/test_dynamics.py -v` → 25 passed
- `uv run pytest tests/test_searcher.py -v` → 45 passed
- `uv run pytest tests/test_closets.py -v` → green after the `_parent_drawer_id` regression fix
- `uv run pytest tests/ --ignore=tests/benchmarks` → 3268 passed; 3 unrelated pre-existing develop failures (2 tracked in #1925, 1 antigravity install)
- `uv run ruff check .` → clean
- `uv run ruff format --check .` → clean

Note: the second commit restores `_parent_drawer_id` scrubbing from public search hits, exposes the logical drawer id as `drawer_id`, and uses that field for potentiation; `tests/test_closets.py` covers the regression.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/dynamics.py
+++ b/mempalace/dynamics.py
@@ -6,10 +6,10 @@ spacing effect: stability grows when reinforcement is spaced rather than
 massed.
 
 This module is pure. No I/O, no DB, no chromadb. It operates on plain
-dicts (hall records, tunnel records) and mutates them in place. Callers
-in ``hallways.py`` and ``palace_graph.py`` invoke these functions; the
-math lives here in one place so both connection kinds share identical
-semantics.
+dicts (hall records, tunnel records, drawer metadata records) and mutates
+them in place. Callers in ``hallways.py`` and ``palace_graph.py`` invoke
+these functions; the math lives here in one place so both connection kinds
+and drawer salience share identical semantics.
 
 Schema fields added to hall + tunnel records (all default-safe — existing
 records without them work via ``initialize_dynamics_fields``):
@@ -100,6 +100,44 @@ def initialize_dynamics_fields(connection: dict, *, now: Optional[datetime] = No
     connection.setdefault("last_activated", created_at)
     connection.setdefault("access_count", 0)
     return connection
+
+
+def initialize_drawer_dynamics_fields(
+    drawer_metadata: dict, *, now: Optional[datetime] = None
+) -> dict:
+    """Populate dynamics fields on drawer metadata, using ``filed_at`` as creation time.
+
+    Drawer metadata historically has ``filed_at`` rather than connection-style
+    ``created_at``. This adapter keeps the shared dynamics math unchanged while
+    preserving the drawer metadata shape: ``created_at`` is only supplied as a
+    temporary fallback and is not left behind when absent from the input.
+    """
+
+    missing_created_at = "created_at" not in drawer_metadata
+    if missing_created_at and drawer_metadata.get("filed_at"):
+        drawer_metadata["created_at"] = drawer_metadata["filed_at"]
+
+    initialize_dynamics_fields(drawer_metadata, now=now)
+
+    if missing_created_at:
+        drawer_metadata.pop("created_at", None)
+
+    return drawer_metadata
+
+
+def drawer_salience(drawer_metadata: dict, *, now: Optional[datetime] = None) -> dict:
+    """Return lazy-decayed salience for drawer metadata without mutating input."""
+
+    record = dict(drawer_metadata or {})
+    initialize_drawer_dynamics_fields(record, now=now)
+    if _parse_iso(record.get("last_activated")) is not None:
+        apply_decay(record, now=now)
+    return {
+        "strength": float(record.get("strength", DEFAULT_STRENGTH)),
+        "stability": float(record.get("stability", DEFAULT_STABILITY)),
+        "last_activated": record.get("last_activated"),
+        "access_count": int(record.get("access_count", 0)),
+    }
 
 
 # ─────────────────────────────────────────────────────────────────────────────
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -53,7 +53,7 @@ import hmac  # noqa: E402
 import sqlite3  # noqa: E402
 import threading  # noqa: E402
 import time  # noqa: E402
-from datetime import date, datetime  # noqa: E402
+from datetime import date, datetime, timezone  # noqa: E402
 from pathlib import Path  # noqa: E402
 from typing import Optional  # noqa: E402
 from urllib.parse import urlparse  # noqa: E402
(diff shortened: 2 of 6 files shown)


igorls:

et1975:
@igorls  rebased on latest

et1975:
@igorls rebased again

Issue #1908, opened 2026-07-01
MCP tools hang after interrupted mine (chromadb-1.x compactor); 3.5.0 retest: lock/peer-writer guard confirmed working


KeilerHirsch:
# MCP tools hang forever after an interrupted `mine` — chromadb-1.x compactor can't backfill WAL into the HNSW segment (CLI unaffected)

## Summary
After a `mempalace mine` is interrupted/killed mid-write, **every MCP tool call hangs indefinitely** (`mempalace_status`, `mempalace_search`, `mempalace_reconnect`, `mempalace_add_drawer`) — no timeout, no error. The **CLI keeps working** (`mempalace status`, `mempalace repair-status` return in seconds). The hang **survives full reboots** because the broken state is on disk. Root cause is a damaged HNSW segment writer + pending rows in chromadb's `embeddings_queue` that the chromadb-1.x compactor cannot apply.

Originally observed on **mempalace 3.4.0**; see the "Update" section below for a retest on **3.5.0**.

## Environment
- mempalace **3.4.0** (original report), chromadb **1.5.9** (Rust core / "compactor")
- Windows 11, Python 3.13, onnxruntime-gpu, embedding model `embeddinggemma` (384-dim), backend `chroma`
- Palace size ~268k drawers (~1.4 GB sqlite + ~450 MB HNSW)

## Symptom
- MCP: any tool that opens the chroma collection hangs forever. The harness logs `Tool 'mempalace_status' still running (720s elapsed)`. The hang is in **native code** — it ignores SIGINT.
- CLI: `mempalace status` and `mempalace repair-status` work fine (they read sqlite counts and **never open a chromadb client**).

## Root cause
An interrupted `mine` leaves unflushed rows in chroma's `embeddings_queue` plus a damaged HNSW **segment writer**. Any operation that opens the collection (`count()`, query, add) makes the chromadb-1.x **compactor** try to backfill the queue into the HNSW segment, which fails:

```
chromadb.errors.InternalError: Error executing plan:
  Error sending backfill request to compactor:
  Failed to apply logs to the hnsw segment writer
```

In the long-lived MCP server this surfaces as an **infinite hang**; called directly from Python it raises the error above.

## How to confirm (read-only)
```python
import chromadb
c = chromadb.PersistentClient(path=r"<HOME>/.mempalace/palace")
for col in c.list_collections():
    print(col.name, col.count())   # -> InternalError: ...hnsw segment writer
```
[...]

PR #1956, opened 2026-07-07
fix: bounded retry before MCP writer lease goes permanently read-only

KeilerHirsch:
## Problem

`_acquire_mcp_writer_lock()` decides read-write vs read-only for an entire MCP server process on its *very first* mutating tool call, and per its own docstring that decision is permanent for the process lifetime once it goes read-only ("Once a server starts read-only it stays read-only for its lifetime; restarting is the safe way to become the writer").

In practice, a peer writer is very often a short-lived `mempalace mine` kicked off by a SessionStart/ingest hook — not a competing interactive session. If the very first mutating tool call in a session happens to land inside that mine's lock window, the *entire* session goes read-only for `mempalace_kg_add` / `mempalace_add_drawer` / `mempalace_diary_write` / etc., even though the peer releases the lock moments later. The only way back is a full client reconnect.

I hit this live: a normal search-heavy session raced a background session-ingest mine on its first knowledge-graph write and got permanently locked out for the rest of the session, well after the ingest had already finished.

## Fix

Add a short, bounded retry around the *first* acquisition attempt only: default 5s, polled every 0.5s, configurable via `MEMPALACE_MCP_WRITER_LOCK_RETRY_SECONDS`. Setting it to `0` reproduces the exact pre-existing immediate-fail behavior (single probe, no sleep).

Scope, deliberately narrow:
- Only affects the `MineAlreadyRunning` branch (a transient peer-writer collision). Setup failures (any other exception acquiring the lock) still fail open immediately, unchanged.
- Does **not** touch the "once read-only, stays read-only for the rest of the process lifetime" contract — that's still exactly as strict as before. This only widens the window for the *initial* decision.
- Does not change `mine_palace_lock()` itself or its use by the `mine` CLI path, which is intentionally non-blocking for a different reason (so hook-spawned miners don't pile up as waiting workers).

## Tests

4 new tests in `tests/test_mcp_server.py`, using a fake monotonic clock (no real sleeping in the suite):
- retry succeeds when the peer releases within the grace window
- retry window expires and the process still goes permanently read-only (existing contract preserved)
- `RETRY_SECONDS=0` reproduces the old single-probe, no-sleep behavior exactly
- env var parsing (default / valid float / negative clamps to 0 / invalid falls back to default)

Full suite: `pytest tests/test_mcp_server.py tests/test_palace_locks.py` → 280 passed, 0 failed, 3 skipped (pre-existing, unrelated). `ruff check` + `ruff format --check` clean on both touched files.

## Related

Not a fix for #1908 (that was the startup connection-refusal / compactor path) — this is a separate, narrower issue in the peer-writer guard added for #1818.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -377,6 +377,35 @@ _MCP_WRITER_LOCK_FAILED = False
 _MCP_WRITER_LOCK_ERROR = ""
 _MCP_ALLOW_PEER_WRITER_ENV = "MEMPALACE_MCP_ALLOW_PEER_WRITER"
 
+# A peer writer is very often a short-lived `mempalace mine` kicked off by a
+# SessionStart/ingest hook, not a competing interactive session. Without a
+# grace period, an MCP server whose *very first* mutating tool call happens
+# to land inside that mine's lock window goes permanently read-only for its
+# entire process lifetime (see the "once read-only, stays read-only"
+# contract on _acquire_mcp_writer_lock) even though the peer releases the
+# lock moments later — the only way back is a full client reconnect. Retry
+# the initial acquisition for a short, bounded window before committing to
+# that permanent decision. This does not change behavior once a server has
+# already decided read-write or read-only — only the first attempt gets the
+# grace period, and only against MineAlreadyRunning (a transient peer-writer
+# collision), not against setup failures (those still fail open immediately,
+# unchanged).
+_MCP_WRITER_LOCK_RETRY_SECONDS_ENV = "MEMPALACE_MCP_WRITER_LOCK_RETRY_SECONDS"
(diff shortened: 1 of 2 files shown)


KeilerHirsch:
Closing as superseded by #1960, which removes the sticky read-only latch entirely: `_acquire_mcp_writer_lock` now re-attempts the non-blocking flock on every mutating call, so a server that starts read-only self-heals into the writer the moment the peer exits.

I verified #1960 also covers the scenario that motivated this PR (first mutating call colliding with a short-lived session-ingest mine): the call that hits the collision window is refused transiently, and the next mutating call after the mine exits promotes the server to writer — no restart, no permanently read-only session. A bounded retry on top of the per-call re-attempt would actually regress the legitimate peer-writer case (every refused write would stall for the grace period), so this approach is obsolete. Thanks to @Evgen197310 for the cleaner fix!

PR #1959, opened 2026-07-07
feat(convo_miner): incremental mining + duplicate detection + per-file wing resolution

jrzmurray:
Three related convo-mining features that turned out not to be independently separable, bundled into one PR — see note on scope below.

## Incremental mining (opt-in, on by default after a real-corpus benchmark)

Full re-normalize + re-chunk of a transcript on every re-mine gets expensive for large, frequently-appended sessions. Adds a cursor mechanism: `_compute_convo_cursor` anchors on the trailing exchange of a Claude Code JSONL transcript at mine time; a later re-mine of a grown file can resume from that anchor (`_incremental_reparse`) and re-chunk only the new tail instead of the whole file, purging only the drawers at or after the resume point. Falls back to a full re-mine whenever anything about the fast path doesn't hold (non-JSONL format, no stored cursor, schema version mismatch, cursor validation failure) — always safe. Gated by `incremental_mining_enabled` (`MempalaceConfig`), defaulted to `True` after measuring against a real corpus (see the commit for numbers).

## Duplicate detection (opt-in, off by default)

Bulk mining had no cross-file duplicate detection at all — a duplicated export or copy-pasted conversation appearing across two files was stored twice with no way to know. Adds an opt-in batched cosine-similarity check per upsert batch (`find_near_duplicates`, one query per batch via ChromaDB's `query_texts`, not one per chunk) that flags near-matches with `possible_duplicate_of`/`duplicate_similarity` metadata rather than skipping them — nothing is silently dropped. A second, stricter, separately opt-in threshold (`duplicate_drop_enabled`/`duplicate_drop_threshold`, clamped to never go below the flagging threshold) additionally *drops* only near-certain duplicates rather than just flagging them, for users who want that.

## Per-file wing resolution for multi-project sweeps

A single `mempalace mine --mode convos` sweep across multiple projects' transcript directories previously filed everything under one wing. Adds per-file wing resolution so each transcript lands in the wing matching its own originating project, with a `WING_RESOLUTION_VERSION` schema bump that forces a one-time reclassification pass over already-filed drawers (mirroring how `NORMALIZE_VERSION` already forces rebuilds after a normalization change).

## Why bundled instead of three PRs

Each of these was going to be its own PR, but duplicate detection's stricter-threshold commit and per-file wing resolution both turned out to structurally depend on incremental mining's shared batch-upsert refactor (`_upsert_chunk_batch`, extracted from what used to be inline in `_file_chunks_locked`) — they reference the function and its metadata-dict shape directly, not just nearby code. Incremental mining in turn depends on the mtime-aware re-mining mechanism from #1957 — splitting further would have meant either stacking three PRs sequentially or re-deriving logic against stale code shapes rather than replaying the actual commits. Bundling was the option that kept every commit's content faithful to what was actually authored and tested together.

Full suite passes clean (3361 passed; the only 2 failures in the full run are `test_repair.py`'s pre-existing FTS5 message-wording gap, confirmed present on plain `develop` today too — see #1928 — unrelated to this PR). `ruff check` / `ruff format --check` clean.


fatkobra (changes requested):
Blocking: the persisted source mtime is not tied to the bytes that were parsed. `_read_source_file()` happens before the palace lock, but the write path samples `os.path.getmtime(source_file)` later. If the file changes in that window, stale chunks are committed with the new mtime, and the next sweep can treat the file as unchanged and permanently miss the appended or edited content.

Please capture the file stat before and after the read, retry or fall back when it changes during the read, and pass the validated mtime into the locked write path rather than sampling it there. A regression test should append to the file between read and write and verify that the new bytes are picked up during this or the next run.

fatkobra (changes requested):
Thanks for the follow-up. Threading the timestamp captured from the open file descriptor through the full-rebuild, incremental, sentinel, and project-mining write paths fixes the original direct “read old bytes, then sample a newer path mtime” window.

However, the read is not actually validated as a stable snapshot. Both `_read_source_file()` and `_read_text_no_follow()` call `fstat()` once before `f.read()`. The source can change while or immediately after it is being read, so the captured timestamp is not guaranteed to identify the bytes returned by `read()`.

There is also a concrete remaining lost-update case: `_is_unchanged_since_last_mine()` and `file_already_mined()` consider mtimes less than 1 ms apart equal. If the source is appended after the read but within 1 ms of the pre-read `fstat()`, the old content can be committed with the earlier timestamp and every later sweep can treat the final file as unchanged, permanently missing that append.

Please use a stable source fingerprint—preferably `(st_dev, st_ino, st_size, st_mtime_ns)`—captured both before and after the read. If the fingerprint changes, retry a bounded number of times or decline that file for this pass. Persist and compare that fingerprint exactly, while retaining compatibility with existing `source_mtime` metadata, rather than relying on the 1 ms tolerance.

Please also add the requested regression test that injects a source change during the read or between the read and commit, including a sub-millisecond mtime change, and verifies that the new bytes are included either in the current pass or the next pass.

Once those cases are covered, my original concern is resolved.

igorls:

Issue #1949, opened 2026-07-07
Semantic search cannot recall Chinese-dense drawers on sqlite_exact (minilm recall + n_results*3 pool cutoff; MCP cannot reach BM25 union path)


felixmurphybox-dot:
## Summary

On a ~1200-drawer palace using the `sqlite_exact` backend (v3.5.0), semantic search (`mempalace_search` via MCP) fails to return certain drawers even for near-verbatim title queries, while `get_drawer` / `list_drawers` / FTS rows are all fine. After a two-agent investigation with numeric reproduction, this looks like a recall/design limitation rather than data corruption, plus one small unreported chroma-hardcoding bug found along the way. Filing both here and asking what fix you would recommend.

## Environment

- mempalace 3.5.0 (PyPI), Python 3.14, macOS (Apple Silicon)
- Backend: `sqlite_exact` (migrated from chroma on 2026-07-07 by read-only extract + re-embed, after the #1845 HNSW hang class; `config.json "backend": "sqlite_exact"`, no env overrides)
- Embedder: `minilm` (all-MiniLM-L6-v2, 384d) — unchanged across the migration (avoided embeddinggemma due to #1858 hang on Apple Silicon)
- Corpus: 1205 document rows, heavily Chinese/mixed-language technical notes

## Symptom

A long, Chinese-dense drawer (4 chunks) is never returned by `mempalace_search`, even when the query is the drawer title almost verbatim. `get_drawer` by ID works; the chunks exist in `docs_fts`; embeddings verified byte-correct (re-embedding each chunk's own text with the same minilm model gives cos(stored, fresh) = 1.0000).

## Diagnosis (numbers from the production embedder against all 1205 vectors)

1. **English-only embedder vs Chinese corpus**: with minilm, the near-verbatim Chinese title query scores best-chunk similarity ~0.32 against its own drawer, ranking **193 / 1205** — while unrelated drawers score 0.53+. Even scoped to the drawer's wing+room (71 docs) it ranks 26-42.
2. **Candidate pool cutoff hides it**: `search_memories` over-fetches only `n_results * 3` vector candidates (`searcher.py`, `total_before_filter = len(drawer_results.documents)`), so rank 193 can never enter a 9/15/30-sized pool. `max_distance` is not the blocker (distances ~0.7 pass the default 1.5).
[...]

PR #1964, opened 2026-07-08
fix(search): expose MCP candidate union

iroiro147:
## Summary
- expose `candidate_strategy` on the `mempalace_search` MCP tool so callers can opt into backend BM25/vector union search
- preserve the selected strategy across the transient-index retry path
- when union runs under `max_distance > 0`, compute vector distances for lexical hits from stored embeddings before admitting them, so BM25 rescue works with the default MCP threshold without weakening the distance guarantee

Refs #1949.

## Tests
- `uv run ruff format --check mempalace/searcher.py mempalace/mcp_server.py tests/test_sqlite_exact_backend.py tests/test_hybrid_candidate_union.py tests/test_mcp_server.py`
- `uv run ruff check mempalace/searcher.py mempalace/mcp_server.py tests/test_sqlite_exact_backend.py tests/test_hybrid_candidate_union.py tests/test_mcp_server.py`
- `git diff --check`
- `uv run pytest tests/test_sqlite_exact_backend.py tests/test_hybrid_candidate_union.py tests/test_mcp_server.py::TestSearchTool tests/test_hnsw_capacity.py::test_bm25_fallback_returns_matches tests/test_hnsw_capacity.py::test_bm25_fallback_filters_by_wing -q` (67 passed)
- `uv run pytest tests/test_searcher.py tests/test_hybrid_search.py -q` (54 passed)
- `uv run pytest tests/ -q --ignore=tests/benchmarks` (3259 passed, 20 skipped, 2 failed)

The two full-suite failures are baseline failures on clean `origin/develop` in `tests/test_repair.py` around FTS5 corruption message wording; I reproduced the same two failures there before opening this PR.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -2373,6 +2373,7 @@ def tool_search(
     max_distance: float = 1.5,
     min_similarity: float = None,
     context: str = None,
+    candidate_strategy: str = "vector",
 ):
     limit = max(1, min(limit, _MAX_RESULTS))
     try:
@@ -2383,6 +2384,9 @@ def tool_search(
         return {"error": str(e)}
     # since/before are validated inside search_memories (shared
     # parse_window), which returns the same {"error": ...} shape.
+    candidate_strategy = candidate_strategy or "vector"
+    if not isinstance(candidate_strategy, str) or candidate_strategy not in {"vector", "union"}:
+        return {"error": "candidate_strategy must be one of ('vector', 'union')"}
     # Backwards compat: accept old name
     # Backwards compat: convert old similarity scale (higher=stricter) to
     # distance scale (lower=stricter). Similarity 0.8 → distance 0.2.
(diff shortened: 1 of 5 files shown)


igorls:

iroiro147:
@igorls done — rebased onto current `develop` at `af7bca77`, conflicts resolved, and it merges cleanly now.

Two notes from the rebase, since develop moved a lot under this PR:

- The searcher-side union plumbing that landed on develop since this was opened integrates cleanly with this change; the remaining substance here is the **MCP surface exposure** (`candidate_strategy` parameter with eager validation) plus the **load-and-check path for BM25-only candidates under `max_distance`** — instead of skipping them outright, union mode loads their stored embeddings and admits only those whose computed distance satisfies the threshold, so the vector-distance guarantee holds while the rerank pool still widens. The docstring now states that behaviour explicitly.
- I kept develop's `lang:` stop-word documentation and the shared since/before validation comment intact alongside the new validation.

Locally: the three touched test files pass (418 tests) on a clean Python 3.11 env with chromadb. CI should confirm shortly. Happy to adjust anything for the 3.8.0 cycle — and thanks for the nudge rather than a close.

Issue #1963, opened 2026-07-08
Epic: concurrent-writer HNSW corruption — unified tracking for the recurring-divergence cluster (357, 1092, 1581, 1799, 1845, 1888, chroma#6975, …)


trek-e:
# Epic: concurrent-writer HNSW corruption — unified tracking for the recurring-divergence cluster

## What this is

A tracking issue for **one architectural problem that currently surfaces as ~25 separate bug reports**: mempalace's deployment topology spawns N concurrent ChromaDB writers (MCP server + per-session stop-hook subprocesses + manual CLI mines + multi-session daemons), but ChromaDB's local `PersistentClient` is fundamentally single-process. When writers contest, die mid-compaction, or are interrupted, the in-memory HNSW state is lost, `embeddings_queue` backs up, and the next operation either SIGBUSes (Rust bindings on diverged state), hangs forever, or silently erodes. Recovery is also broken at scale.

I have personally hit this 4-5 times now. Each recurrence is the same architectural mismatch expressing itself through a different symptom.

This issue exists to (a) tie the fragmented reports to one root cause, (b) track the multi-PR fix pipeline, and (c) give maintainers a single decision point on the architectural direction.

## Root cause

**chroma-core/chroma#6975** (OPEN): `PersistentLocalHnswSegment._persist()` only flushes the in-memory HNSW tree to `data_level0.bin` when `sync_threshold` (default 1000 records) is crossed. Below that, writes survive only as `embeddings_queue` WAL rows. **If a different writer crosses the threshold with different records before yours do, your writes are silently purged** when `max_seq_id` advances past them.

This is ChromaDB's durability model, not a mempalace bug. The only safe deployment is **one process owning the ChromaDB connection**.

## Forensics from one incident (2026-07-07, ~1M-drawer palace)

Three independent smoking guns, all consistent with multi-writer contention:

| Evidence | What it proves |
|---|---|
| `acquire_write` table has **712 rows, all `lock_status=1`** | Every write lock ChromaDB ever took is still marked held. None released cleanly. |
| `embeddings_queue` holds **717 unconsumed rows** spanning seq_id 956348→1001406 | WAL is backing up; nobody is consuming it. |
| Vector-segment max seq_id (1000691) vs metadata-segment max (1001406) = **gap of 715, matching the 717-row queue** | Conclusive: queued writes are not being applied to the HNSW segment. |
[...]

Issue #1888, opened 2026-06-27
Peer-writer lock (#1818/#1823) is incompatible with the multi-session daemon topology (#1270) + external hook/manual mines


marcoaperez:
### Summary

The new peer-writer guard shipped in 3.5.0 (#1818 / #1823) acquires `mine_palace_lock` and holds it for the **entire process lifetime**. For users running the **multi-session daemon** model — i.e. the topology that the still-open PR #1270 is meant to enable — this guard is, as far as I can tell from reading the code, **not usable**: it locks out hook-driven and manual `mempalace mine` runs, and it doesn't fit a daemon that services many sessions on different threads. The result is that the one 3.5.0 feature that would actually fix our root cause (concurrent writers corrupting the HNSW index) is the feature we can't adopt.

I'm filing this as a friendly heads-up: **#1270 (open) and #1818 (merged) are on a collision course.** If the guard stays as-is, daemon-topology users can't move to 3.5.0.

### My setup (a concrete instance of #1270)

- MemPalace via `pipx`, palace `~/.mempalace`, embedder `embeddinggemma` (CPU).
- A single long-lived **daemon** (a LaunchAgent) that `import`s `mempalace.mcp_server.handle_request` and serves every editor session (Claude Code, Claude Desktop, WARP) through a tiny stdio↔unix-socket **bridge**. This is byte-for-byte the design in **PR #1270**. It exists precisely to avoid the concurrent-stdio-writer HNSW corruption that #1818 also targets.
- Auto-save **hooks** that shell out to `mempalace mine --mode convos`, plus periodic **manual** `mempalace mine` of docs. These are separate processes from the daemon.

I just hit an HNSW divergence (index held 21 elements vs 16,551 in sqlite; vector search fell back to BM25) — exactly the corruption class #1818 wants to prevent — so I very much *want* this protection. I just can't enable it without breaking the rest.

### The incompatibility (code references, v3.5.0 / `develop`)

1. **The guard holds the per-palace mine lock for the whole process lifetime.** `_acquire_mcp_writer_lock()` enters `mine_palace_lock(_config.palace_path)` and registers an `atexit` to release it — so it's held until the process exits. `mempalace/mcp_server.py:299`.

[...]

PR #1966, opened 2026-07-08
fix(mcp): release the writer lease after a mutating-idle window (#1888)

JasonAiassist:
## Problem

The per-palace writer lease (#1818/#1823) is held for the **whole MCP process lifetime**. As #1888 reports, one interactive session therefore starves every other writer on the palace for hours:

- hook-driven and manual `mempalace mine` runs exit with `MineAlreadyRunning`
- queued-writes daemon jobs fail against the held flock
- a second session's mutating tools are refused with `-32001` until the first session ends

We hit this in production daily: multiple Claude Code sessions plus lifecycle hooks against one palace produced recurring sqlite→HNSW divergence (362 drawers in one day) before 3.5.0's lease, and with the lease the starvation above.

## Fix — cooperative lease

Release the lease once no mutating tool has run for `MEMPALACE_MCP_WRITER_LEASE_IDLE_S` seconds (default 300; `0` restores the legacy hold-until-exit behavior).

The critical correctness piece: the release also runs `_force_chroma_cache_reset()`, so the next mutating call reopens the palace from disk. Stale in-memory HNSW state is *why* the lease was lifetime-scoped; dropping every cached Chroma handle at release time is what makes a re-acquire safe. Re-acquisition rides the existing self-heal retry path (#1960), so the first mutating tool after a release transparently wins the lease back — no restart, no user action.

Implementation notes:
- A mutating-tool in-flight counter plus last-mutation timestamp (guarded by a leaf `threading.Lock`) drive the idle clock; the lease is never released mid-write.
- The idle-exit watchdog thread performs the release check and now also starts when only the lease release is enabled (`MEMPALACE_MCP_IDLE_HOURS=0` no longer disables it).
- The atexit hook is registered once and reads the *current* lease, replacing the per-acquire lambda that would go stale across release/re-acquire cycles.

## Tests

8 new tests in `tests/test_mcp_server.py` covering: idle release + cache reset, no release before the window / while a mutation is in flight / when disabled via env, env fallback on invalid values, re-acquire after release, in-flight tracking, and atexit idempotency. Full suite: **3267 passed, 20 skipped**; ruff check/format clean.

## Scope

Part of the #1963 concurrent-writer cluster (tactical tier — no architectural change; fully compatible with both the queued-writes daemon and the #1270 bridge direction). Fixes the starvation half of #1888; the thread-local re-entrancy half was already fixed by the process-wide holder set in `palace.py`.

Follow-up candidate (separate PR if wanted): `service.run_mcp_tool` executes write handlers without taking `mine_palace_lock`, so daemon-executed `mcp_tool` jobs could still interleave with a live lease holder once something starts submitting them.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -44,6 +44,7 @@ except (OSError, AttributeError):
 sys.stdout = sys.stderr
 
 import argparse  # noqa: E402  (deferred until after stdio protection above)
+import contextlib  # noqa: E402
 import json  # noqa: E402
 import logging  # noqa: E402
 import re  # noqa: E402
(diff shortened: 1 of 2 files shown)


fatkobra (changes requested):
Blocking: resetting Chroma caches only when the lease is released does not make lease handoff safe. A local read can reopen and cache a PersistentClient while this process is lease-free; a peer can then write; and when this process reacquires the lease, the first mutation reuses that stale client because no reset occurs after acquisition.

There is also a handoff race because the OS lease is released before `_force_chroma_cache_reset()` completes, so a new local mutator can acquire the lease and begin using a client that the previous releaser is about to reset.

Please reset and reopen after every successful lease acquisition before the first mutation, and coordinate release, reset, and reacquisition under one process-wide lifecycle guard. Add a regression covering: release -> local read reopens -> peer writes -> reacquire -> mutate.

igorls:

Issue #1552, opened 2026-05-19
Subprocess hook architecture causes ChromaDB cold-init deadlock + STATUS_ACCESS_VIOLATION segfaults on Windows (multi-session)


cdewet78:
﻿## Summary

When MemPalace is invoked as a short-lived subprocess from a Claude Code Stop hook (one subprocess per response), under concurrent multi-session usage on Windows, the chromadb client cold-init wedges indefinitely while holding HNSW segment file locks. A meaningful percentage of subprocess invocations also exit with `STATUS_ACCESS_VIOLATION` (Windows exit code `3221225477` / `0xC0000005`). Both symptoms cause cascading session hangs that present to the user as "Calling mempalace..." spinners that never complete.

The underlying mempalace MCP server (long-running, in-process) handles drawer reads/writes fine. The bug only surfaces under the **repeated cold-init subprocess** pattern triggered by hook-based auto-save.

## Environment

- **OS:** Windows 11 Pro 10.0.26200 (build 26200)
- **mempalace:** 3.3.5 (also reproduces against 3.3.1)
- **chromadb:** 1.5.8 (also reproduces with 1.5.9 from fresh install)
- **onnxruntime:** 1.26.0
- **Python:** Tested on **3.14.3** (Windows installer, system) AND **3.12.10** (winget user-scope install) â€” **same wedge in both**, so this is not a 3.14 compatibility issue
- **Concurrent Claude Code sessions:** 5 typical, 2-3 minimum for repro
- **Palace size at repro:** ~6250 drawers, 38 MB chroma.sqlite3, 2 HNSW segment dirs

## Symptom

1. User starts multiple Claude Code sessions. Each session installs `mempalace` MCP server via `mempalace.mcp_server` stdio entry.
2. A separate per-session Stop hook (Claude Code feature) fires on every response, invoking `mempalace hook run --hook stop --harness claude-code` as a fresh subprocess.
3. Most subprocess invocations complete in 1-3 seconds. ~20% of them either:
   - **Hang indefinitely** on chromadb cold init â€” process exists with 0 CPU, ~3-4 MB working set, hours later still not exited
   - **Segfault** with exit code `3221225477` (`STATUS_ACCESS_VIOLATION`)
4. Wedged subprocesses hold the per-palace lock (`~/.mempalace/locks/mine_palace_*.lock`) indefinitely.
5. Subsequent hook invocations from other sessions deadlock on the held lock â†’ cascade.
[...]

PR #1968, opened 2026-07-08
fix(mcp): exit stdio watchdog on orphanhood, not idle time

supersuperltd:
The #1552 idle watchdog exits the server after MEMPALACE_MCP_IDLE_HOURS (default 8 h) even when the stdio client is still connected. Claude Desktop keeps servers attached but silent overnight and never respawns a stdio server it saw die, leaving a persistent 'Server disconnected' banner.

Fix: the stdio watchdog now exits only when the process is orphaned — parent PID check on POSIX (ppid 1 or reparented), parent-handle poll on Windows (OpenProcess/WaitForSingleObject), so the Windows file-handle goal of #1552 stays covered. Clean disconnects were already handled by the stdin-EOF break in the read loop. The HTTP transport keeps the plain idle-timeout exit (no parent to watch; supervisors respawn it). MEMPALACE_MCP_IDLE_HOURS is unchanged: 0 disables, other values set the HTTP threshold and polling cadence.

Tests: 6 new tests in TestIdleExitWatchdog; full test_mcp_server.py suite passes (277). Verified end-to-end with real subprocesses: with a 3.6 s timeout, a live-parent server survived 12 s of idleness; an orphaned one exited within ~2 s.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -324,11 +324,16 @@ _kg_by_path: dict[str, KnowledgeGraph] = {}
 _kg_cache_lock = threading.Lock()
 _palace_flag_given: bool = bool(_args.palace)
 
-# MCP server idle auto-exit (#1552).  Stale MCP servers from ended Claude
+# MCP server stale auto-exit (#1552).  Stale MCP servers from ended Claude
 # Code sessions do not self-terminate, accumulating ChromaDB/HNSW file
-# handles on Windows.  When MEMPALACE_MCP_IDLE_HOURS is set (or defaults
-# to 8 h), a background daemon thread exits the process once no request
-# has been handled for that long.  Set to 0 to disable.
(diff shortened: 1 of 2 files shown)


igorls:

supersuperltd:
Thanks Igor, and no worries about the turnaround.

I'd rather not pick this one back up, so please go ahead and close it. We've stepped back from active development on our fork, so I won't be rebasing.

For what it's worth, the underlying bug still looks live on develop: the stdio watchdog still exits on idle time alone, so a live-but-silent Claude Desktop connection can still be killed overnight and Desktop won't respawn it. The fix here is self-contained, an orphanhood check gating the stdio exit with HTTP keeping the plain idle timeout, so if it's useful to you or anyone else please feel free to lift the commit directly. No attribution needed.

Thanks for the project.

Issue #1799, opened 2026-06-13
Concurrent sessions: MCP write tools (add/delete/update_drawer) bypass mine_palace_lock and corrupt the palace


hypyc95:
## Summary

Running **two or more concurrent Claude Code sessions** with the MemPalace plugin can corrupt the palace, because the **MCP server's write tools bypass the per-palace mine lock**. The backend is an embedded `chromadb.PersistentClient` over a single `chroma.sqlite3` (plus HNSW `.bin` segment files); when an MCP write runs concurrently with a `mempalace mine` (or with another session's MCP write), the writes race and corrupt the FTS5 index and/or the HNSW vector segment.

> **Correction (edited):** an earlier version of this issue claimed the auto-save *hook* writes take no palace-wide lock. That was wrong — the Stop/PreCompact hooks call `mempalace mine … --mode convos`, which **does** take `mine_palace_lock`, so hook-driven mines are already serialized against each other (the loser raises `MineAlreadyRunning`). The real gap is the MCP write tools, below.

## Environment

- mempalace 3.4.0 (PyPI), chromadb 1.5.9
- Windows 11, Python 3.13
- Claude Code with the MemPalace plugin (MCP server + Stop/PreCompact auto-save hooks)
- User routinely runs multiple Claude Code sessions at once (so multiple `mempalace-mcp` servers run concurrently)

## Root cause (from reading the installed package)

- **The mine path is correctly locked.** `miner.mine` and `convo_miner.mine_convos` both wrap their write in `mine_palace_lock(palace_path)` (non-blocking; the loser raises `MineAlreadyRunning`). So two concurrent `mempalace mine` invocations against one palace cannot corrupt each other.
- **The MCP write tools are NOT locked.** In `mcp_server.py`, `tool_add_drawer` → `col.upsert(...)`, `tool_delete_drawer` → `col.delete(...)`, and `tool_update_drawer` → `col.add(...)` call Chroma directly. `mine_palace_lock` does not appear anywhere in `mcp_server.py`. So an MCP write races against a concurrent mine or against another session's MCP write.
- **HNSW `.bin` segment files live outside SQLite's transaction**, so a partially-written vector index is not protected by SQLite atomicity even when the SQLite write is fine.
- `hooks.auto_save = false` is **not** a sufficient mitigation: each session's `mempalace-mcp` server keeps the DB open and its write tools remain callable independent of the hooks.

## Symptoms observed (one mining session, multi-session machine)

[...]

Issue #1581, opened 2026-05-21
HNSW corruption from concurrent ChromaDB clients (MCP server + hook subprocesses + auto-ingest mine) within single Claude Code session


KVSocial:
## Environment

- mempalace 3.3.0
- chromadb 1.5.8
- Python 3.13
- Claude Code (one session, single user, single workstation)

## Summary

Inside a **single** Claude Code session, mempalace's own design opens **multiple concurrent ChromaDB PersistentClient instances** against the same `~/.mempalace/palace` directory. ChromaDB's Rust-backed HNSW writer is not multi-process safe — the result is `chromadb.errors.InternalError: Error in compaction: Error constructing hnsw segment reader: Error loading hnsw index`, after which the palace is unreadable. `mempalace repair --yes` returns "Cannot recover — palace may need to be re-mined from source files."

I initially assumed multi-Claude-session contention. It is not. The user confirmed they only ever run one Claude Code session against this project. The concurrent writers are all spawned by mempalace itself.

## Concurrent writers (within ONE Claude Code session)

1. **MCP server subprocess** — long-lived stdio process started by Claude Code on session start. Opens ChromaDB on first tool call, keeps it open until session end.
2. **`Stop` hook** — fires after every assistant turn (`mempalace hook run --hook stop --harness claude-code`). New Python process, opens its own ChromaDB client, writes save state, exits.
3. **`PreCompact` hook** — fires on context compaction. New process, opens ChromaDB.
4. **`SessionStart` hook** — fires on startup, can spawn background `mempalace mine` subprocess (auto-ingest). Auto-ingest is long-running and keeps its own ChromaDB client open.
5. **`PostToolUse` hooks** if configured.

All of (1)–(5) open independent `chromadb.PersistentClient` instances on `~/.mempalace/palace`. The hook log shows `Skipping auto-ingest: mine already running` recurring at every save — the auto-ingest subprocess does not always finish before the next stop hook fires.

## Evidence

```
~/.mempalace/hook_state/hook.log:

  File "/home/steven/.local/share/uv/tools/mempalace/lib/python3.13/site-packages/mempalace/backends/chroma.py", line 204, in upsert
    self._collection.upsert(**kwargs)
  File "/home/steven/.local/share/uv/tools/mempalace/lib/python3.13/site-packages/chromadb/api/models/Collection.py", line 530, in upsert
    self._client._upsert(...)
[...]

PR #1969, opened 2026-07-08
feat(cli): warn in mempalace status when hooks are active but daemon off

trek-e:
## What does this PR do?

Surfaces the at-risk "hooks active but daemon off" topology in `mempalace status`, the cheapest adoption nudge for the opt-in daemon.

The daemon (`mempalace/daemon.py`) serializes hook writes through a single ChromaDB PersistentClient, eliminating the concurrent-open pressure that produces recurring HNSW divergence (epic #1963; cluster: #357 #1092 #1581 #1799 #1845). It is **off by default**, requires an explicit `mempalace daemon start`, and is never surfaced in onboarding or status — so users whose stop-hooks are actively saving have no signal that they're in the at-risk topology until divergence recurs (4-5 times, in the reporter's case).

This PR adds a best-effort warning to `cmd_status`:

- When `daemon.get_client_if_running(palace_path)` reports the daemon isn't running
- AND `~/.mempalace/hook_state/hook.log` shows activity within the last 7 days

…print a clear pointer at `mempalace daemon start`. The warning never raises and never blocks the status output — a probe that fails just means we can't tell, so we stay silent rather than crying wolf.

## How to test

```bash
uv run pytest tests/test_cli_daemon_warning.py -v        # the 5 new tests
uv run pytest tests/test_cli.py -v                       # no regression on cmd_status callers
```

Five new tests cover the seams:

- `test_no_warning_when_daemon_running` — daemon up → no warning (hook saves serialized)
- `test_warning_when_hooks_active_daemon_off` — the at-risk state → warn + actionable hint
- `test_no_warning_when_hooks_never_ran` — no hook.log → nothing to warn about
- `test_no_warning_when_hook_activity_stale` — hook.log older than 7 days → effectively inactive
- `test_no_warning_when_daemon_probe_raises` — best-effort contract: probe failure must not break status

## Checklist

- [x] Tests pass (`uv run pytest tests/test_cli_daemon_warning.py tests/test_cli.py` → 87 passed)
- [x] No hardcoded paths
- [x] Linter passes on changed files (`uvx --from "ruff==0.15.14" ruff check` + `ruff format --check` clean)

## Scope

Intentionally narrow — one helper + one call site + tests. Part of epic #1963. Follow-ups (deferred to keep this PR focused):

- Onboarding offer to enable + start the daemon after hooks are set up
- A `mempalace daemon install-launchd` (macOS) / systemd (Linux) helper so the daemon survives reboots — without this, the daemon dies on reboot and hooks silently fall back to the direct fail-fast path

🤖 This PR was prepared with AI assistance per CONTRIBUTING.md.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -33,6 +33,7 @@ import os
 import sys
 import shlex
 import argparse
+import time
 from pathlib import Path
 
 from .config import MempalaceConfig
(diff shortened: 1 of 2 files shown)


igorls:

PR #1971, opened 2026-07-08
fix(palace): pair the mine_palace_lock holder-set update with its release (#1970)

### LINKED ISSUE #1970: bug: mine_palace_lock holder set can outlive the flock on an async interrupt (silent two-writer window)
**What happened?**

`mine_palace_lock()` records the hold with `_mark_held(palace_key)` before the `try:` that would release it:

```python
# mempalace/palace.py, mine_palace_lock (~L1167)
_write_lock_holder(lf)
_mark_held(palace_key)          # records the hold
try:
    yield
finally:
    _mark_released(palace_key)  # only released here
```

So if an async exception (`KeyboardInterrupt`/SIGINT, or one injected into a thread) hits the small window after `_mark_held()` and before the `try:`, `_mark_released()` never runs and `palace_key` is left in the process-wide `_palace_lock_keys` set. The outer `finally:` still drops the flock and closes the fd. Now the holder set and the OS lock disagree: in memory we still "hold" the palace, but the flock is free.

`_held_by_this_process()` is process-wide, so the next `mine_palace_lock()` in this process just passes through and writes without the flock. Nothing stops another process from grabbing it at the same time, so you get two writers into one palace, which is the HNSW corruption the lock is there to stop.

**What did you expect?**

The holder set shouldn't outlive the flock. Once the flock is dropped, `_held_by_this_process()` should read `False` again so the next acquire really locks.

**How to reproduce:**

develop @ `a20d769b`, POSIX. The window is a couple of bytecodes, so the script raises in it directly instead of racing a real signal:

```python
import fcntl
import hashlib
import os
import tempfile

import mempalace.palace as palace
from mempalace.palace import mine_palace_lock

d = tempfile.mkdtemp()
p = os.path.join(d, "palace")
os.makedirs(p, exist_ok=True)
key = hashlib.sha256(os.path.normcase(os.path.realpath(p)).encode()).hexdigest()[:16]
lock_path = os.path.join(os.path.expanduser("~"), ".mempalace", "locks", f"mine_palace_{key}.lock")

# Fire an interrupt in the window: right after _mark_held() records the hold,
# before the try: that guarantees _mark_released().
_real_mark_held = palace._mark_held


def _interrupt_in_window(k):
    _real_mark_held(k)       # the real hold is recorded in _palace_lock_keys
    raise KeyboardInterrupt  # ...then interrupted, before entering the try


palace._mark_held = _interrupt_in_window
try:
    with mine_palace_lock(p):
        pass
except KeyboardInterrupt:
    pass
finally:
    palace._mark_held = _real_mark_held

print("in-memory holder set still thinks we hold it:", palace._held_by_this_process(key))

# The outer finally already dropped the flock. A fresh non-blocking acquire
# proves the OS lock is free (flock competes across open file descriptions, so
# this raises BlockingIOError if the lock is still held):
with open(lock_path, "r+b") as f2:
    try:
        fcntl.flock(f2, fcntl.LOCK_EX | fcntl.LOCK_NB)
        print("OS flock is actually FREE (fresh acquire succeeded)")
        fcntl.flock(f2, fcntl.LOCK_UN)
    except BlockingIOError:
        print("OS flock still held")
```

Output:

```
in-memory holder set still thinks we hold it: True
OS flock is actually FREE (fresh acquire succeeded)
```

**Proposed fix**

Move `_mark_held()` inside the `try:` so it always pairs with `_mark_released()`:

```python
_write_lock_holder(lf)
try:
    _mark_held(palace_key)
    yield
finally:
    _mark_released(palace_key)
```

`_mark_released()` is a `set.discard`, so releasing a key that was never added does nothing, and the normal path is unchanged. Happy to send a small PR with a regression test that fires the interrupt in the window if you want it.

**Environment:**
- OS: Linux (POSIX flock path; the Windows `msvcrt` branch is structured the same way)
- Python version: 3.11+
- MemPal version: develop @ `a20d769b`

Low severity in practice: you need an async interrupt in a tiny window and a process that catches it and keeps writing. Worth fixing anyway, because when it does hit, the corruption is silent.

mvalentsev:
Fixes #1970.

## What does this PR do?

`mine_palace_lock()` recorded the process-wide hold with `_mark_held(palace_key)` *before* the `try:` whose `finally:` runs `_mark_released()`. An async exception (a `KeyboardInterrupt`/SIGINT, or one injected into a thread) landing in the gap after `_mark_held()` and before the `try:` skips `_mark_released()`, so `palace_key` is left in the `_palace_lock_keys` set. The outer `finally:` still frees the flock, so the in-memory holder set outlives the OS lock: `_held_by_this_process()` then reports a hold the flock no longer backs, the next re-entrant acquire in this process passes through and writes without the flock, and another process is free to acquire it at the same time. That is two writers into one palace, the HNSW-corruption race the lock is there to stop.

The fix moves `_mark_held()` to the first line inside the `try:`, so it pairs with `_mark_released()` on every exit. `_mark_released()` is a `set.discard`, so releasing a key that was never added is a no-op, and the normal acquire/release path is unchanged.

## How to test

`tests/test_palace_locks.py::test_holder_set_not_orphaned_by_interrupt_after_mark_held` runs the real `_mark_held`, then raises in the window as a signal would, and asserts the holder set is unchanged afterwards. It fails on the current code (the key is stranded) and passes with the fix.

```
uv run pytest tests/test_palace_locks.py tests/test_chroma_collection_lock.py
```

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -1165,8 +1165,14 @@ def mine_palace_lock(palace_path: str):
                 ) from exc
         # Record our own identity for any later contender's diagnostic message.
         _write_lock_holder(lf)
-        _mark_held(palace_key)
+        # Mark the hold from inside the try so it always pairs with
+        # _mark_released. If it sat before the try, an async exception
+        # (a SIGINT/KeyboardInterrupt) landing in the gap would orphan
+        # palace_key in the holder set while the outer finally frees the
+        # flock, so the in-memory hold would outlive the OS lock and a later
+        # re-entrant acquire would pass through and write without the flock.
         try:
+            _mark_held(palace_key)
             yield
         finally:
             _mark_released(palace_key)
--- a/tests/test_palace_locks.py
+++ b/tests/test_palace_locks.py
@@ -17,6 +17,7 @@ import sys
 
 import pytest
 
+import mempalace.palace as palace_mod
 from mempalace.palace import (
     _write_lock_holder,
     MineAlreadyRunning,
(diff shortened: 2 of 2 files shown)


mvalentsev:
The red `lint` check isn't from this PR. `ruff format --check .` flags `tests/test_mcp_server.py`, which went in unformatted with #1960 and isn't touched here. #1965 fixes that, so this should go green once it lands.

PR #1977, opened 2026-07-09
fix(lock): recover same-process stale palace lock

### LINKED ISSUE #1924: mine_palace_lock flock FD leak under long-lived MCP server → writes self-deadlock (MineAlreadyRunning cites own PID)
### Summary
Under the long-lived MCP server (`mempalace.mcp_server`), `mine_palace_lock`'s advisory flock FD leaks open. Subsequent writes from the same server fail with `MineAlreadyRunning` citing **the server's own PID** (self-deadlock). Every write after the first degrades or fails.

### Environment
- mempalace 3.4.1
- Python 3.11 (macOS, `/Library/Frameworks/Python.framework/Versions/3.11`)
- Used as an MCP server via [`mempalace-pi`](https://github.com/daniel-ospina/mempalace-pi) (long-lived stdio server) inside the [pi coding agent](https://github.com/earendil-works/pi-coding-agent)

### Reproduce
1. Start `python3 -m mempalace.mcp_server` (or run via an MCP host that keeps it alive).
2. Issue several write tool calls (`mempalace_add_drawer`, `mempalace_kg_add`, etc.) over the server's lifetime.
3. After the first 1–2 writes, subsequent writes fail:
   ```
   palace ... is held by PID <SERVER_PID> (-c mempalace.mcp_server ...); wait for it to finish
   ```
   The cited PID is the **server's own PID** — the server holds its own lock and can't re-acquire it.

### Root cause (traced)
- `palace.py:1043 mine_palace_lock` acquires `fcntl.flock(LOCK_EX|LOCK_NB)` (line ~1122) on a lock file at `~/.mempalace/locks/mine_palace_<sha256(palace_path)[:16]>.lock`.
- The lock is held for the **entire server lifetime** (by design — `mcp_server.py:main()` wraps the request loop in `with mine_palace_lock(...)`; the comment cites the trade-off as "blocks concurrent ChromaDB access, MCP-MCP + MCP-CLI"). This is intentional for the server's own writes.
- However, the write **tool calls** within the server acquire the same `mine_palace_lock` again. The re-entrancy guard is a thread-local marker (`_held_by_this_thread`, line ~1086). Under the long-lived server, this marker **desyncs from the OS flock state** — the OS-level flock is held (FD leaked open), but the thread-local marker doesn't reflect that the *current tool call's* re-acquire should be a no-op. So `flock(LOCK_NB)` returns `BlockingIOError` → `MineAlreadyRunning`, citing the server's own PID (which the lock file records as the holder).

### Workaround (verified)
```bash
rm -f ~/.mempalace/locks/mine_palace_*.lock
```
Unlinking the directory entry works despite the leaked FD: the leaked FD keeps the stale inode open (its flock persists), but a new `open()` at the same path creates a **fresh inode** whose flock succeeds. Writes resume immediately. This is fragile (manual, race-prone) but confirms the FD-leak diagnosis.

### Proposed fix
Detect `MineAlreadyRunning` where the holder PID == the current process's own PID (self-deadlock) and auto-recover:
- If `holder_pid == os.getpid()` (or `holder_pid` is in the current process tree) → unlink the stale lock file and retry once.
- Safety: this only ever clears a lock the current process itself leaked; it never clears a genuinely separate `mine`'s lock.

(We have implemented a host-side mitigation for this in our agent harness — a pi extension that pre-clears self-deadlocked locks before write tool calls. Filing upstream so the root cause can be fixed at the source. Happy to open a PR if this approach sounds right.)

### Related
This also manifests as a second `mcp_server` instance failing to start with `"MemPalace MCP server is already running for this palace."` (`mcp_server.py:3813`) — same lifetime-lock design.

fatkobra:
## What does this PR do?

Closes and Fixes #1924 by adding narrow same-process stale-lock recovery to `mine_palace_lock()`.

When lock acquisition fails and the lock file says the holder PID is this same process, the process has self-deadlocked on a stale/leaked flock FD. The patch unlinks that pathname and retries once. This matches the verified workaround from the issue while keeping the safety boundary tight.

## Safety properties

- Only recovers when the recorded holder PID equals `os.getpid()`.
- Never clears a separate process's lock.
- Retries only once.
- Declines pathname-unlink recovery on Windows, where open byte-locked files cannot be handled the same way.
- Preserves existing process-wide re-entrancy for normal nested writes.

## How to test

Added `tests/test_palace_lock_self_deadlock.py` covering:

- holder PID parsing;
- POSIX stale same-process inode replacement and retry;
- refusal to clear another process's holder identity.

Run:

````
    python3 -m ruff format mempalace/palace.py tests/test_palace_lock_self_deadlock.py
    python3 -m ruff check mempalace/palace.py tests/test_palace_lock_self_deadlock.py
    python3 -m pytest tests/test_palace_lock_self_deadlock.py -q
    python -m pytest tests/ -v
````

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -1073,119 +1073,156 @@ def _write_lock_holder(lock_file) -> None:
         pass
 
 
+def _holder_pid_from_message(holder: str) -> int | None:
+    match = re.search(r"\bPID\s+(\d+)\b", holder or "")
+    if not match:
+        return None
+
+    try:
+        return int(match.group(1))
+    except ValueError:
+        return None
+
+
+def _holder_is_current_process(holder: str) -> bool:
+    return _holder_pid_from_message(holder) == os.getpid()
+
+
+def _recover_self_deadlocked_palace_lock(lock_path: str, holder: str) -> bool:
+    """Recover a stale same-process palace lock by replacing its pathname.
+
+    This is intentionally narrow: only a lock file whose recorded holder PID is
+    the current process can be cleared. We never unlink a separate writer's
+    lock. On POSIX this recovers the common leaked-FD shape: the old inode keeps
+    its stale flock, while a new open at the same path can acquire a fresh lock.
+    Windows cannot reliably unlink an open byte-locked file, so we decline
+    recovery there and preserve the original error.
+    """
+
+    if not _holder_is_current_process(holder):
+        return False
+
+    if os.name == "nt":
+        return False
+
+    try:
+        os.unlink(lock_path)
+    except FileNotFoundError:
+        return True
+    except OSError:
+        logger.debug("Self-deadlocked palace lock unlink failed", exc_info=True)
+        return False
+
+    return True
+
+
 @contextlib.contextmanager
 def mine_palace_lock(palace_path: str):
     """Per-palace non-blocking lock around the full `mine` pipeline.
 
-    The per-file `mine_lock` only protects delete+insert interleave for a
-    single source; it does not prevent N copies of `mempalace mine <dir>`
-    from being spawned concurrently by hooks. When that happens, each copy
-    drives ChromaDB HNSW inserts in parallel against the same palace,
-    which (combined with chromadb's multi-threaded ParallelFor) can
-    corrupt the HNSW graph and produce sparse link_lists.bin blowups.
-
-    The lock file is keyed by sha256(palace_path) so mines against
-    *different* palaces can still run in parallel — we only serialize
-    writes into the same palace, which is the correctness boundary.
-
-    The key is derived from a fully normalized form of the path:
-    `realpath` resolves symlinks and `..` segments, and `normcase` folds
-    case on Windows (which has a case-insensitive filesystem). Without
-    normcase, `C:\\Palace` and `c:\\palace` would hash to different keys
-    on Windows and let two concurrent mines touch the same on-disk palace.
-
-    Non-blocking: if another `mine` is already writing to this palace,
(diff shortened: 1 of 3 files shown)


mvalentsev:
@fatkobra the failure mode this recovers should already be gone on develop: #1859 made the re-entrancy marker process-wide, so a second acquire in the same process passes through instead of failing with its own PID. I couldn't reproduce the self-deadlock on current develop; the new test only hits it by taking the flock through a handle opened outside `mine_palace_lock`. Is there still a path to "flock fails and holder PID == getpid()" on current code?

The unlink also worries me: after `os.unlink` the old inode stays locked while a new open creates a fresh one, so a process that opened the path before the unlink ends up locking a different file, and two mines can write the same palace at once. And since `_write_lock_holder` is best-effort, with PID reuse the same-PID check can match a lock held by a live foreign process.

Also heads-up: this rewrites the same lines as #1971 (which moves `_mark_held` inside the inner try), so whichever lands second needs a rebase that keeps that.

fatkobra:
Thanks @mvalentsev  — I agree with this concern too.

I’m going to stop this PR unless I can reproduce the self-deadlock on current `develop` through a real `mine_palace_lock()` path. The test I added forces the condition by taking the flock outside `mine_palace_lock()`, which does not prove the current code can still hit `holder PID == os.getpid()` after #1859 made the re-entrancy marker process-wide.

I also agree that unlinking the lock file is too risky because it can create two lock inodes for the same palace. That is worse than the original failure because it can allow two writers.

I’ll close this instead of trying to land the unlink approach.

PR #1985, opened 2026-07-10
fix: make _HTTP_REQUEST_LOCK write-only so reads run concurrently

### LINKED ISSUE #1984: Global _HTTP_REQUEST_LOCK serializes all requests; a slow read blocks concurrent calls (HTTP transport)
**Summary**

On the HTTP transport, `_HTTP_REQUEST_LOCK` wraps the entire `handle_request()` dispatch, so every tool runs one-at-a-time even though the server is a `ThreadingHTTPServer(daemon_threads=True)`. A single slow read therefore starves all concurrent requests.

**Impact**

With the Qdrant backend, aggregate reads like `list_wings`/`status`/`get_taxonomy` full-scan the collection (~180s on a large palace). While one runs it holds the global lock, so concurrent `search`/`initialize`/`tools/list` calls block until it finishes and time out. In practice the MCP client fails to connect while an aggregate call is in flight.

**Root cause**

The lock's own comment scopes its purpose to preventing concurrent Chroma/HNSW mutation, and the Chroma safety paths are gated by `_is_chroma_backend()`. So it's a Chroma/SQLite/HNSW-era safeguard that, on the thread-safe Qdrant backend, serializes reads with no shared embedded index to protect. Qdrant's client carries its own re-entrant lock and the server handles concurrent requests transactionally.

**Suggested fix**

Make the lock write-only: acquire `_HTTP_REQUEST_LOCK` only for mutating tools; reads plus `initialize`/`ping`/`tools/list` run lock-free. Two read paths touch shared module state and need their own narrow guards rather than the global lock — the `_metadata_cache` store in `_get_cached_metadata`, and the `_collection_cache` rebuild in `_get_collection`. The cross-process writer flock (`_MCP_WRITER_LOCK`) stays unchanged, so write-vs-write remains serialized. PR attached.

**Verification**

Authored and concurrency-tested on v3.5.0: a 5-assertion test passed first try — a `search` returned in 3.75s while a `list_wings` full-scan was still running (pre-fix this hung); 5 concurrent searches all <5s; concurrent aggregate reads race-free; two concurrent writes still serialize with no corruption; a read during an in-flight write returned in 0.196s. The attached PR ports that fix onto `develop`.

JasmineBlake-Author:
Fixes #1984.

`_HTTP_REQUEST_LOCK` wrapped the whole `handle_request()` dispatch, serializing every tool despite `ThreadingHTTPServer(daemon_threads=True)`. Its purpose (per its comment) is preventing concurrent Chroma/HNSW mutation, and the Chroma guards are gated by `_is_chroma_backend()` — so on the Qdrant backend it needlessly serialized reads, and a slow read (e.g. a `list_wings` full-scan) starved concurrent searches and caused MCP connect timeouts.

**Change**
- Acquire `_HTTP_REQUEST_LOCK` only for mutating tools; reads and `initialize`/`ping`/`tools/list` run lock-free.
- Add a `_metadata_cache` guard (paired store in `_get_cached_metadata`; the fetch stays outside the lock).
- Add a `_collection_cache` rebuild guard in `_get_collection` (double-checked; warm cache-hit stays lock-free).
- Leave the cross-process writer flock (`_MCP_WRITER_LOCK`) unchanged — write-vs-write still serializes.

Chroma's HNSW-mutation hazards remain covered where `_is_chroma_backend()` applies; this only stops the lock from serializing reads on backends that don't need it.

**Port note (please review):** the fix was authored against v3.5.0, whose `mcp_server.py` is ~400 lines shorter than `develop` HEAD. Two judgment calls while porting:
- The mutating-tool set reuses `develop`'s canonical `_MUTATING_TOOLS` (`_WRITE_TOOLS = _MUTATING_TOOLS | {"mempalace_reconnect"}`) rather than the hand-rolled list from v3.5.0. This automatically picks up `mempalace_kg_supersede` (a mutator that exists on `develop` but not in the v3.5.0 list) and avoids a parallel set that would drift. `mempalace_reconnect` is added on top because `tool_reconnect()` closes and rebuilds the shared backend handle, which must not race a concurrent lock-free read — `develop` deliberately keeps it out of `_MUTATING_TOOLS`, but under today's wrap-everything lock it is serialized, and this preserves that.
- The second `_HTTP_REQUEST_LOCK` acquisition (the `_warmup_with_lock` eager-embedder-warmup thread) is intentionally left holding the lock — it mutates embedder/HNSW state and is not a per-request read path.

**Verification — please note honestly:** authored and concurrency-tested on **v3.5.0** (search 3.75s while `list_wings` still scanning — pre-fix hung; 5 concurrent searches <5s; aggregate reads race-free; 2 concurrent writes serialize; read-during-write 0.196s). This PR **ports** that fix onto `develop` HEAD; the port is verified by inspection + `py_compile`, **not** re-run against the concurrency suite on `develop`. Maintainer verification against your test setup is welcome.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -322,6 +322,9 @@ _READ_ONLY = bool(getattr(_args, "read_only", False)) or os.environ.get(
 
 _kg_by_path: dict[str, KnowledgeGraph] = {}
 _kg_cache_lock = threading.Lock()
+# Narrow guards so read paths do not depend on the (now write-only) global lock.
+_metadata_cache_lock = threading.Lock()
+_collection_cache_lock = threading.RLock()
 _palace_flag_given: bool = bool(_args.palace)
 
 # MCP server idle auto-exit (#1552).  Stale MCP servers from ended Claude
@@ -1041,23 +1044,27 @@ def _get_collection(create=False):
                 ):
                     _collection_open_error = None
                     return _collection_cache
-                _collection_cache = None
-                _collection_cache_backend = None
-                _collection_cache_palace = None
-                if _collection_cache is None:
-                    from .palace import get_collection as palace_get_collection
+                with _collection_cache_lock:
+                    # Double-checked locking: another thread may have rebuilt
+                    # the handle while we waited on the lock.
+                    if not (
+                        _collection_cache is not None
+                        and _collection_cache_backend == backend_name
+                        and _collection_cache_palace == _config.palace_path
+                    ):
+                        from .palace import get_collection as palace_get_collection
 
-                    _collection_cache = palace_get_collection(
-                        _config.palace_path,
-                        collection_name=_config.collection_name,
-                        create=create,
-                        backend=backend_name,
-                    )
-                    _collection_cache_backend = backend_name
-                    _collection_cache_palace = _config.palace_path
-                    _collection_open_error = None
-                    _metadata_cache = None
-                    _metadata_cache_time = 0
+                        _collection_cache = palace_get_collection(
+                            _config.palace_path,
+                            collection_name=_config.collection_name,
+                            create=create,
+                            backend=backend_name,
+                        )
+                        _collection_cache_backend = backend_name
+                        _collection_cache_palace = _config.palace_path
+                        _collection_open_error = None
+                        _metadata_cache = None
+                        _metadata_cache_time = 0
                 return _collection_cache
             except (BackendMismatchError, KeyError) as exc:
                 logger.warning("backend open failed for %s: %s", _config.palace_path, exc)
@@ -1360,8 +1367,9 @@ def _get_cached_metadata(col, where=None):
         return _metadata_cache
     result = _fetch_all_metadata(col, where=where)
     if where is None:
-        _metadata_cache = result
-        _metadata_cache_time = now
+        with _metadata_cache_lock:
+            _metadata_cache = result
+            _metadata_cache_time = now
     return result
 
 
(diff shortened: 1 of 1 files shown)


fatkobra (changes requested):
Blocking: classifying `mempalace_reconnect` as a write only serializes it against other writes. Reads never acquire `_HTTP_REQUEST_LOCK`, so reconnect can run concurrently with an in-flight query/get and close the shared backend, clear Chroma's process-global System cache, and replace KG handles while that read is still using them.

Reconnect is a lifecycle operation and needs to be exclusive against both readers and writers. Please use a reader/writer lifecycle lock, with ordinary operations holding a shared lease and reconnect holding the exclusive lease, or use a generation/refcounted swap that waits for old readers to drain. Add a regression with a deliberately slow read overlapping reconnect.

mvalentsev:
Making the lock write-only lets reads run against writes, not only against other reads.

I replaced the handlers for `mempalace_list_wings` and `mempalace_add_drawer` with a 2s sleep on both branches and fired two requests at once. develop serializes every combination at about 4s. This branch: read+read 2.2s (the intended win), write+write 4.0s, and read+write 2.0s, so a query is in flight against the same backend handle while a mutation commits. That's the case the comment on the old lock was there to prevent, and it isn't limited to `mempalace_reconnect`.

A reader/writer lock gets the same throughput without that: shared lease for reads, exclusive for `_MUTATING_TOOLS` plus reconnect. `_metadata_cache_lock` as written doesn't add anything either, since the readers of `_metadata_cache` don't take it.

There's no coverage of the new policy at the moment. A test that overlaps a slow read with a write would pin the intended behaviour either way.

igorls:

PR #1989, opened 2026-07-10
fix(embedding): preload CUDA/cuDNN DLLs before the first GPU session

KeilerHirsch:
## Problem

pip-installed CUDA/cuDNN runtimes (the `nvidia-*-cu12` wheels) live inside site-packages, which is **not on the Windows DLL search path**. The result is the worst kind of failure — a silent one:

- `CUDAExecutionProvider` reports *available*, `describe_device()` says GPU
- `InferenceSession` creation fails to load `cublasLt64_12.dll` / `cudnn64_9.dll` (LoadLibrary error 126)
- onnxruntime **silently falls back to CPU** — every embedding runs ~10x slower, ~2 GB extra host RAM per process, and nothing in the logs says why

Field-observed on Windows 11 + RTX 4060 + `onnxruntime-gpu==1.26.0` + cu12 wheels (326k-drawer palace). Until now the only fix was a hand-rolled `sitecustomize.py` in the venv calling `onnxruntime.preload_dlls()` — which silently disappears on every `uv tool upgrade mempalace`.

## Fix

Call `onnxruntime.preload_dlls()` once per process from `_resolve_providers`, **only when a GPU provider is actually selected**:

- `preload_dlls()` ships with onnxruntime >= 1.21 and resolves the pip-wheel DLLs explicitly; it is a no-op where nothing needs fixing. Older builds are skipped via `getattr`.
- Lock-guarded one-shot (`_PRELOAD_LOCK`), so a warmup thread racing the first tool call cannot double-preload.
- CPU resolutions never pay the preload.
- Failures log a one-shot warning and reproduce today's fallback behavior instead of raising.

## Tests

Four new tests in `tests/test_embedding.py` (preloads once per process; CPU path skips; tolerates missing `preload_dlls` on old builds; swallows preload crashes). The autouse fixture inerts the one-shot by default so existing tests that fake CUDA availability don't run the real preload on CI.

`tests/test_embedding.py` + `test_embedding_wrapper.py` + `test_embedder_identity.py`: **55 passed**. `ruff format --check` + `ruff check`: clean.

### DIFF
--- a/mempalace/embedding.py
+++ b/mempalace/embedding.py
@@ -101,6 +101,47 @@ _EF_CACHE: dict = {}
 _EF_CACHE_LOCK = threading.Lock()
 _WARNED: set = set()
 
+# One-shot guard for _preload_gpu_dlls: the DLL search path only needs to be
+# fixed up once per process, before the first GPU InferenceSession. The lock
+# keeps concurrent resolvers (e.g. a warmup thread racing the first tool
+# call) from both paying the preload.
+_PRELOAD_DONE = False
+_PRELOAD_LOCK = threading.Lock()
+
+
+def _preload_gpu_dlls(ort) -> None:
+    """One-shot ``onnxruntime.preload_dlls()`` before the first GPU session.
+
+    pip-installed CUDA/cuDNN runtimes (the ``nvidia-*-cu12`` wheels) live
+    inside site-packages, which is not on the Windows DLL search path.
+    Without the preload, ``CUDAExecutionProvider`` reports *available* but
+    session creation fails to load ``cublasLt64_12.dll``/``cudnn64_9.dll``
+    (LoadLibrary error 126) and onnxruntime silently falls back to CPU —
+    ``describe_device()`` says GPU while every embedding runs ~10x slower.
+    ``preload_dlls()`` (onnxruntime >= 1.21) resolves the wheel DLLs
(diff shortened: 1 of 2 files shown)


igorls:

KeilerHirsch:
Yep, definitely still want this one landed. I’ll rebase onto current develop and get CI green.

I’ve been distracted by some rather deep debugging, forensics, and measurement/provenance rabbit holes lately — apparently that problem class follows me around 😅. So MemPalace has been waiting for me rather than forgotten.

Ironically I’m currently back in GPU/runtime forensics locally, so this PR is still very relevant. Thanks — I’ll pick it back up.

KeilerHirsch:

The provider resolver has grown model-aware auto-selection since this PR was opened. The rebase keeps that current behavior intact: the existing `_AUTO_PROVIDER_DENYLIST` still decides whether an auto provider is eligible, and DLL preloading now runs only after an eligible accelerator is actually selected. Explicit accelerator requests keep the existing current-develop path and preload before session construction.

- `tests/test_embedding.py tests/test_embedding_wrapper.py tests/test_embedder_identity.py`: 66 passed (2 existing deprecation warnings)

GitHub now reports the PR mergeable again. Pre-rebase head is preserved as `archive/2026-08-29/pr1989-pre-rebase`.

Issue #1646, opened 2026-05-29
Feature: SSE/HTTP transport mode for mempalace-mcp (multi-agent safe)


mjjy-chen:
## Problem

When multiple AI agent profiles (e.g., Hermes Agent with 9 independent profiles) each start their own `mempalace-mcp` process via stdio, all 9 processes concurrently read/write the same ChromaDB (SQLite + HNSW). ChromaDB is not designed for multi-process concurrent access, which causes:

1. **HNSW/sqlite drift** — SQLite stays current but HNSW index becomes stale, eventually triggering quarantine (`Quarantined corrupt HNSW segment`). This happened 3 times in our setup (drift-20260522, drift-20260525, drift-20260528).
2. **SIGSEGV crashes** — v3.3.2 (#1000) fixed the quarantine recovery, but the root cause (concurrent access) remains.
3. **No sharing** — Each profile runs its own process, so there's no way for multiple agents to safely share the same palace without corrupting it.

The current workaround is removing mempalace from all but one profile, which means 8 worker agents lose memory storage entirely. Daily `mempalace repair --mode legacy` helps recover drift, but doesn't prevent it.

## Proposed Solution

Add an **SSE (Server-Sent Events) / Streamable HTTP transport mode** for `mempalace-mcp`, similar to how many MCP servers now support both stdio and SSE:

```
# Start as a persistent HTTP server
mempalace-mcp --transport sse --port 8765

# Or in config
mcp_servers:
  mempalace:
    url: http://localhost:8765/sse
```

This would:
- Run **one** mempalace-mcp server process that serializes all writes
- Allow N agent profiles to connect via HTTP/SSE (read-only or full access)
- Eliminate HNSW drift caused by concurrent ChromaDB access
- Make mempalace safe for multi-agent setups (Hermes profiles, Claude Code worktrees, Codex parallel agents, etc.)

## Implementation Notes

- The MCP SDK (`fastmcp` or `mcp` Python package) already supports SSE transport — this is mostly a config/entry-point change
- Could add a `--transport stdio|sse` flag and `--port` option to the CLI
- SQLite WAL mode could be enabled for the single-server case to improve write throughput
- Consider a `--readonly` client flag for profiles that only need search/query

## Environment

- MemPalace v3.3.5 on WSL2 (Ubuntu)
- Hermes Agent v0.14.0 with 9 profiles (default + 8 workers)
- 129 drawers, 21 wings, ChromaDB backend
[...]

PR #1992, opened 2026-07-10
feat(mcp): opt-in write forwarding to the daemon (MEMPALACE_MCP_FORWARD_WRITES)

JLujan10:
## Problem

Multiple concurrent MCP stdio sessions against one palace still contend on the per-palace writer lease: the first session to write holds it for its **entire process lifetime**, so every other session's mutating tools refuse until that terminal closes. The self-healing retry (recently landed) fixes the *stale-latch* half, but while the holder lives, concurrent writers are still locked out.

This is the standing pain in #1646 (N agent profiles, one palace) and #1497 (single-writer/gateway question). My own setup hits it daily: multiple git-worktree Claude Code sessions, where each agent team's *last* task writes its diary — so the first team to finish locks all the others out for the rest of the day.

## Approach

The write daemon (`mempalace.daemon`) already exists as an opt-in queued writer with an `mcp_tool` job kind. This PR wires the MCP server to it, strictly opt-in:

**`MEMPALACE_MCP_FORWARD_WRITES`** =
- unset (default): behavior unchanged.
- `prefer` (or bare truthy): mutating tool calls are submitted to a *running* daemon as `mcp_tool` jobs; if no daemon is reachable, fall back to today's direct path.
- `require`: forward, and refuse (`-32004`) when no daemon is reachable — the server then **never** acquires the writer lease.

With the daemon as the single palace writer, any number of stdio sessions plus CLI/hook daemon jobs converge on one writer — no lease contention, no corruption risk from a second long-lived Chroma client's stale HNSW/FTS state. Complements the HTTP transport (#1801): stdio remains Claude Code's default wiring, and forwarding also covers hosts that can't switch transports.

## Design notes

- **Scope**: only tools in `_MUTATING_TOOLS` that `service.classify_tool()` calls `write` are forwarded — `run_mcp_tool` is kept as the single arbiter. Maintenance tools (`mempalace_mine`/`mempalace_sync`) keep the direct path (the daemon has dedicated job kinds for them, wired via the CLI `--daemon` flag). Side observation while aligning the sets: `mempalace_kg_supersede` is in `_MUTATING_TOOLS` but absent from `service.WRITE_TOOLS` — possibly drift worth a look; under this PR it simply falls through to the direct path.
- **Gate ordering**: the read-only-mode and sqlite-integrity gates are applied *before* forwarding (a read-only team server must refuse, not forward); the peer-writer gate is what forwarding replaces.
- **Fallback safety**: `prefer` falls back to direct **only when the job was never submitted**. After submission the daemon owns the write; a wait/transport failure returns an error rather than retrying in-process, because re-applying a non-idempotent mutation (`diary_write`) could double-write.
- **Result parity**: the daemon's job-transport keys (`stdout`/`stderr`/`exit_code`) are stripped so the forwarded result matches what the in-process handler returns; `_decorate_mcp_tool_result` is applied the same way.
- The diary `content`→`entry` alias mapping is hoisted above the preflight (behavior-neutral for the direct path — preflight gates key on tool name only) so forwarded jobs carry resolved arguments.
- `MEMPALACE_MCP_FORWARD_TIMEOUT` (seconds, default 120) bounds the job wait — an MCP client shouldn't inherit the daemon client's 1-hour default.

## Testing

- New `tests/test_mcp_write_forwarding.py` — 17 tests through `handle_request` with a fake daemon client, mirroring the peer-writer guard test style: off-by-default, prefer/require daemon-up/down, no-fallback-after-submission (the double-write guard), failed-job surfacing, read tools never forwarded, maintenance tools keep the direct path, alias-before-forwarding, read-only gate precedence, mode parsing.
- `tests/test_mcp_server.py`, `test_mcp_stdio_protection.py`, `test_mcp_http_transport.py`, `test_daemon.py`: **325 passed** (no regressions).
- `pre-commit` (ruff + ruff format) clean.

Happy to adjust naming/defaults to whatever direction you prefer for the #1646 family.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -407,6 +407,158 @@ def _truthy_env(name: str) -> bool:
     return os.environ.get(name, "").strip().lower() in {"1", "true", "yes", "on"}
 
 
+# MCP write forwarding (opt-in; #1646 / #1497).
+#
+# When MEMPALACE_MCP_FORWARD_WRITES is set, mutating tool calls are submitted to
+# the long-lived write daemon (``mempalace.daemon``) as ``mcp_tool`` jobs
+# instead of executing in-process — so this MCP server never acquires the
+# per-palace writer lease, and any number of concurrent stdio sessions (plus
+# CLI/hook daemon jobs) converge on the daemon as the single palace writer.
+_MCP_FORWARD_WRITES_ENV = "MEMPALACE_MCP_FORWARD_WRITES"
+_MCP_FORWARD_TIMEOUT_ENV = "MEMPALACE_MCP_FORWARD_TIMEOUT"
+_MCP_FORWARD_TIMEOUT_DEFAULT = 120.0
+_mcp_forward_mode_warned = False
+
+
+def _mcp_forward_mode() -> str:
+    """Resolve MEMPALACE_MCP_FORWARD_WRITES to ``prefer``, ``require``, or ``""``.
+
+    - unset/empty: off — current direct-write behavior.
+    - ``prefer`` (or a bare truthy value): forward writes to a running daemon;
+      fall back to the direct path when no daemon is reachable.
+    - ``require``: forward writes; refuse when no daemon is reachable. This
+      server then never takes the palace writer lease.
+    """
+    global _mcp_forward_mode_warned
+    raw = os.environ.get(_MCP_FORWARD_WRITES_ENV, "").strip().lower()
+    if not raw:
+        return ""
+    if raw in {"prefer", "require"}:
+        return raw
+    if raw in {"1", "true", "yes", "on"}:
+        return "prefer"
+    if not _mcp_forward_mode_warned:
+        logger.warning(
+            "%s=%r not recognized (expected 'prefer' or 'require'); write forwarding stays OFF",
+            _MCP_FORWARD_WRITES_ENV,
+            raw,
+        )
+        _mcp_forward_mode_warned = True
+    return ""
+
+
+def _mcp_forward_error(req_id, tool_name: str, message: str):
+    return {
+        "jsonrpc": "2.0",
+        "id": req_id,
+        "error": {
+            "code": -32004,
+            "message": message,
+            "data": {"tool": tool_name},
+        },
+    }
+
+
+def _mcp_maybe_forward_write(req_id, tool_name: str, tool_args: dict):
+    """Forward a mutating tool call to the write daemon when enabled.
+
+    Returns a complete JSON-RPC response when the call was handled here
+    (forwarded, refused, or failed), or ``None`` to fall through to the direct
+    dispatch path.
+
+    Scope: only tools that are both in ``_MUTATING_TOOLS`` and classified
+    ``write`` by the daemon's own allowlist (``service.classify_tool``) are
+    forwarded. Maintenance tools (``mempalace_mine`` / ``mempalace_sync``) keep
+    the direct path — the daemon has dedicated job kinds for those (wired via
+    the CLI ``--daemon`` flag), and ``run_mcp_tool`` would reject them.
+
+    Fallback safety: ``prefer`` mode falls back to the direct path ONLY when
+    the job was never submitted. After submission the daemon owns the write; a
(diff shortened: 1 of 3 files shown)


fatkobra (changes requested):
Blocking: `mempalace_kg_supersede` is in the MCP server's mutating-tool set, but it is absent from the daemon routing classifier. As a result, even with `MEMPALACE_MCP_FORWARD_WRITES=require`, this mutation falls through to local execution and the MCP process can still become a direct palace writer.

Please use one canonical source of truth for mutating/write classification, or add this tool to `WRITE_TOOLS`. A table-driven test should assert that every member of `_MUTATING_TOOLS` is classified as `write`, unless it is explicitly documented and handled as an exclusive maintenance operation.

JLujan10:
@fatkobra Thank you — confirmed, and fixed in 7e1ad2d (branch also rebased onto current develop). \`mempalace_kg_supersede\` was added to \`_MUTATING_TOOLS\` after this branch was cut, and the gate's \`classify_tool(...) != "write" → return None\` fall-through turned that drift into exactly the silent direct write this feature exists to prevent.

What changed:

- **Concrete fix**: \`mempalace_kg_supersede\` added to \`service.WRITE_TOOLS\` — it now forwards like every other KG mutation.
- **Canonical classification**: the forwarding scope is now derived as \`_MUTATING_TOOLS − service.MAINTENANCE_TOOLS\` (reusing the daemon's own maintenance set rather than restating it). A \`classify_tool()\` mismatch on anything in that scope is treated as classifier **drift**, not a routing decision: **refused under \`require\`** (-32004, naming the missing-classification defect), warn + documented direct fallback under \`prefer\`.
- **Table-driven test** as requested: \`test_every_mutating_tool_is_classified_for_forwarding\` asserts every \`_MUTATING_TOOLS\` member is classified \`write\` unless it's in the documented \`MAINTENANCE_TOOLS\` exemption. Verified it fails on the pre-fix tree (catches \`kg_supersede\`). Two behavior tests pin the drift branch under both modes.
- **One more hole closed in the same class**, found while fixing: under \`require\`, maintenance tools (\`mempalace_mine\`/\`mempalace_sync\`, both in \`_MUTATING_TOOLS\`) previously fell through to in-process execution and could take the writer lease the mode promises never to take. They're now refused under \`require\` with a pointer to the CLI \`--daemon\` path (dedicated job kinds); \`prefer\` keeps the direct path.
- **Adopted #2027**: \`MEMPALACE_MCP_FORWARD_WRITES\` now parses through \`write_routing.parse_write_routing_policy\` (\`legacy_boolean=True\`), so the MCP transport speaks the same \`direct\`/\`prefer\`/\`require\` vocabulary as the staged hook/CLI rollout. The routing *decision* stays transport-specific since an MCP server discovers daemon availability by submitting rather than probing — happy to move it onto \`choose_write_route\` if you'd rather.

375 tests passing across \`test_mcp_write_forwarding.py\` (now 22), \`test_write_routing.py\`, \`test_mcp_server.py\`, and \`test_hook_write_routing.py\`.

fatkobra (approved):
Thanks — I reviewed `7e1ad2d`, and this addresses my blocking finding.

The important parts are now covered:

- `mempalace_kg_supersede` is classified through `service.WRITE_TOOLS`;
- the parity test guards every `_MUTATING_TOOLS` member, apart from the
  documented maintenance exemptions;
- classifier drift fails closed under `require`;
- the `prefer` fallback is explicit and logged;
- maintenance operations no longer fall through to direct execution under
  `require`.

Although `_MUTATING_TOOLS` and `WRITE_TOOLS` remain separate sets, the CI
parity invariant plus the runtime fail-closed behaviour satisfy the intent
of my request.

I also agree that the MCP submit/wait path has sufficiently different
transport semantics that reusing the shared policy parser without forcing
the decision through `choose_write_route()` is reasonable here.

My blocking concern is resolved.

One non-blocking documentation cleanup: the PR description still contains
the pre-fix statements that `mempalace_kg_supersede` falls through, that
maintenance tools retain the direct path, and that the forwarding suite has
17 tests. Please update those sections to reflect `7e1ad2d`.

igorls:

PR #1994, opened 2026-07-11
feat(hooks): hooks.ignore_paths config to exempt project dirs from capture

JasonAiassist:
## Problem

Ephemeral worker sessions (objective-loop agents spawning dozens of nested harness sessions per hour) each fire session-start/stop/session-end ingests. On my store that meant hundreds of ingest writes per hour, several of which died mid-write (segfault in the HNSW add path) and re-diverged the index within hours of a repair (same failure family as #1888 / #1966).

## Change

A new `hooks.ignore_paths` config lets an operator exempt listed project directories from capture entirely:

- Matched by cwd path-prefix, with a `transcript_path` dash-encoded fallback for hooks that receive no cwd.
- The hook still returns valid JSON immediately for exempted paths, so the harness never sees an error.
- 38 lines of unit tests covering prefix matching, the transcript-path fallback, and the passthrough case; full suite green (147 tests).

Running live on my instance since 2026-07-10 — the agent-farm directories no longer generate ingests, and the store has stayed convergent since.

## Note

Stacked on #1966 — this branch includes that PR's commit (`327c28e`) because it was cut from the same local build; once #1966 merges to `develop`, this PR reduces to the single `feat(hooks)` commit. Happy to rebase instead if you'd prefer.

### DIFF
--- a/mempalace/hooks_cli.py
+++ b/mempalace/hooks_cli.py
@@ -1353,6 +1353,36 @@ def hook_precompact(data: dict, harness: str):
     _output({})
 
 
+def _ignored_cwd(data: dict) -> bool:
+    """True when the session's cwd falls under ``hooks.ignore_paths`` (config).
+
+    Ephemeral worker sessions (e.g. objective-loop agents spawning dozens of
+    nested harness sessions per hour) each fire session-start/stop/session-end
+    ingests; ``hooks.ignore_paths`` lets an operator exempt those project dirs
+    from capture instead of flooding the store. Matches the hook payload's
+    ``cwd`` by path-prefix and, as a fallback, the dash-encoded project-dir
+    form inside ``transcript_path``.
+    """
+    try:
+        cfg = json.loads((PALACE_ROOT / "config.json").read_text())
+        paths = (cfg.get("hooks") or {}).get("ignore_paths") or []
+    except Exception:
+        return False
+    if not paths:
+        return False
+    cwd = str(data.get("cwd") or "")
+    transcript = str(data.get("transcript_path") or "")
+    for raw in paths:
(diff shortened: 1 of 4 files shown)


fatkobra (changes requested):
Blocking: 
the transcript-path fallback uses an unanchored substring check on a lossy slash-to-dash encoding. For example, ignored cwd `/home/dev/foo` encodes to `-home-dev-foo`, which also appears inside the transcript directory for sibling project `/home/dev/foobar`. Path components containing dashes can create additional encoding collisions. This can silently suppress capture for an unrelated project.

Please parse the known project-directory component from the transcript path and compare a complete encoded path token with explicit boundaries, using the same normalization on both sides. Add regressions for `foo` versus `foobar` and for paths whose slash/dash encodings collide.

mvalentsev:
Adding to the substring problem above: the prefix check has a Windows gap as well.

With `ignore_paths: ["C:\\Users\\dev\\foo"]`, that exact directory is skipped through the `cwd == p` branch, but a session in `C:\Users\dev\foo\sub` is captured, because `cwd.startswith(p + "/")` never matches a backslash path. So on Windows the setting covers the top directory only and quietly misses everything under it, which is the opposite failure from the one on Linux but just as invisible. Normalizing separators on both sides and comparing whole components would take care of this and the dash-encoding case in the same place.

Separate from the feature itself: the diff also carries the writer-lease rework from #1966, with `_acquire_mcp_writer_lock` rewritten and `_writer_lease_activity`, `_writer_lease_idle_seconds` and the idle-release env knob added. A config knob for hook capture and a change to when the MCP process releases the palace writer lease carry very different risk, and someone reviewing by the title won't be expecting the second one.

igorls:

PR #1995, opened 2026-07-11
feat(sweep): add --wing/--room to classify swept drawers (#1207, #1979)

### LINKED ISSUE #1207: Add --wing flag to sweep command for proper drawer classification
## Summary

`mempalace sweep` currently has no `--wing` flag, so all swept drawers land in a `?/?` wing/room. This is fine for single-wing palaces but leaks into search results once a palace contains multiple wings — orphan `?/?` results show up alongside properly-tagged wings in `mempalace search` output.

## Use case

Common pattern (recovery + nightly cron):

1. Primary ingestion: `mempalace mine ~/my-project --mode convos --wing my_project`
2. Recovery from killed terminal OR nightly catch-up: `mempalace sweep ~/.claude/projects/-home-user-my-project`

After step 2, palace contains:

- `my_project/technical` — original `mine` drawers (file-level chunks)
- `?/?` — sweep drawers (message-level, valuable but unlabeled)

Both layers are complementary (different abstraction depth, both useful for recall) — but the wing label is missing from the second.

## Proposed

```bash
mempalace sweep <target> [--wing <name>] [--room <name>]
```

- If `--wing` not provided → current `?/?` fallback (preserves backward compat)
- If `--wing` provided → tag new drawers identically to `mine --wing <name>`
- Optional `--room` to override default room classification

## Why this matters for cron-based deployments

The natural cron schedule is **sweep nightly** (catches every new convo turn from the day) + **mine weekly** (refreshes file-level summaries). Without `--wing`, every nightly sweep grows the `?/?` wing instead of keeping the project wing canonical and searchable.

## Workaround today

Tagging via direct ChromaDB access — fragile, version-coupled to ChromaDB schema, no migration path.

## Repro / discovery

Discovered while recovering from a killed `mempalace mine --mode convos --wing X` terminal. Followed README pattern:

```bash
# After accidental terminal close mid-mine:
mempalace sweep ~/.claude/projects/-home-user-project
# → "+48560 new, 0 already present, 134826 skipped (< cursor)"
# All 48560 new drawers landed in `?/?` instead of the X wing
```

mempalace 3.3.3, plugin install via `claude plugin install --scope user mempalace`.

### LINKED ISSUE #1979: Add taxonomy metadata to sweep drawers for transcript catch-up
## Summary

`mempalace sweep` is useful for message-level catch-up on appended/growing transcript files, but swept drawers are currently stored without `wing` / `room` metadata. As a result, they show up as `? / ?` in `mempalace status` and search results, even when the same transcript was originally mined under a proper wing.

This makes the practical `mine + sweep` workflow hard to use in multi-wing palaces.

Related older issue: #1207. I am opening this separately because the issue is not only the absence of a `--wing` flag, but also the broader taxonomy / metadata behavior of swept drawers when `sweep` is used as the catch-up path for growing transcripts.

## Why this matters

For transcript ingestion, there appear to be two complementary paths:

```text
mine  = file/exchange-level ingestion, classified with wing/room
sweep = message-level catch-up / recovery ingestion, idempotent per message
```

That split makes sense. However, if `sweep` writes message-level evidence without taxonomy metadata, users have to choose between:

- `mine`: classified, but not reliable for appended/growing transcript catch-up;
- `sweep`: reliable catch-up, but unclassified as `? / ?`.

For multi-project or multi-wing palaces, the unclassified `? / ?` drawers are hard to filter, inspect, and reason about through CLI or MCP clients.

## Minimal reproduction on mempalace==3.5.0

1. Create a small Claude-style JSONL transcript.
2. Run:

```bash
mempalace mine <transcript-dir> --mode convos --wing test_wing
```

Observed: one drawer is filed under `test_wing / general`.

3. Append a new message to the same transcript file.
4. Run the same `mine` command again.

Observed:

```text
Files processed: 0
Files skipped (already filed): 1
Drawers filed: 0
```

The appended message is not filed by the second `mine` run.

5. Run:

```bash
mempalace sweep <transcript.jsonl>
```

Observed:

```text
Swept ...: +3 new, 0 already present, 0 skipped (< cursor).
```

The appended message is now searchable, but search/status classify the swept drawers as:

```text
? / ?
```

`mempalace status` shows the split:

```text
WING: ?
  ROOM: ?           3 drawers

WING: test_wing
  ROOM: general     1 drawer
```

6. Run `sweep` again.

Observed:

```text
+0 new, 1 already present, 2 skipped (< cursor)
```

So `sweep` is idempotent with its own message-level drawer IDs. The issue is not repeated duplicate writes. The issue is missing taxonomy metadata on useful swept drawers.

## Code-level observation

In the current sweeper path, message metadata includes fields such as:

```text
session_id
timestamp
message_uuid
role
source_file
filed_at
ingest_mode=sweep
```

But it does not include:

```text
wing
room
```

That explains why swept drawers appear as `? / ?`.

## Proposed behavior

Add explicit taxonomy support to `mempalace sweep`:

```bash
mempalace sweep <target> --wing <wing> --room <room>
```

Suggested behavior:

1. `--wing` is optional for backward compatibility.
2. `--room` is optional, but when omitted MemPalace should use a stable default such as `transcript`, `sessions`, or `general` instead of leaving the room unset.
3. The values should be threaded into both single-file sweep and recursive directory sweep.
4. Swept drawers should include metadata like:

```json
{
  "wing": "test_wing",
  "room": "transcript",
  "ingest_mode": "sweep",
  "source_file": "...",
  "session_id": "...",
  "message_uuid": "...",
  "timestamp": "..."
}
```

## Suggested tests

Please consider adding tests for:

- `sweep file.jsonl --wing X --room Y` stores `wing=X`, `room=Y`;
- `sweep <dir> --wing X --room Y` applies the same metadata to every swept file;
- repeated `sweep` remains idempotent;
- `mempalace status` no longer reports swept drawers as `? / ?` when taxonomy is provided;
- `mempalace search --wing X` can find swept drawers under that wing;
- MCP search filters can see swept drawers under the chosen taxonomy.

## Documentation request

It would also help to document the intended relationship between `mine` and `sweep` for transcript workflows:

```text
mine  = classified file/exchange-level ingestion
sweep = classified message-level catch-up / recovery ingestion
```

Right now the two layers are technically complementary, but operationally hard to manage because only one of them carries taxonomy metadata.

mvalentsev:
Fixes #1207
Fixes #1979

## What does this PR do?

`mempalace sweep` writes one verbatim drawer per message for transcript catch-up, but stored them with no `wing`/`room` metadata. Swept drawers therefore showed up as `?/?` in `mempalace status` and were invisible to `mempalace search --wing`. In a multi-wing palace the intended two-layer workflow (`mine` for file-level chunks, `sweep` for appended-transcript catch-up) was hard to use, because only the `mine` layer carried taxonomy.

This adds optional `--wing` / `--room` to `sweep`, matching how `mine --wing` classifies:

- `mempalace sweep <target> --wing my_project` files swept drawers under `my_project`. `--room` defaults to the miners' fallback room `general`, so swept drawers sit alongside mined ones and are reachable via `search --wing my_project`.
- `--room <name>` sets the room. `--room` without `--wing`, or with a blank one, is a no-op and prints a warning, since a room needs a wing to nest in.
- With no `--wing`, behavior is unchanged: drawers stay unclassified (`?/?`), so existing callers and the hook/cron path are unaffected.

Both the single-file `sweep()` and the recursive `sweep_directory()` thread the values through, and so does a sweep routed through the daemon (`--daemon`, or a `prefer`/`require` CLI write-routing policy): the job payload carries both values and `service.run_sweep` hands them to the sweeper.

## How to test

```bash
# classified under a wing/room
mempalace --palace /tmp/p sweep session.jsonl --wing my_project --room chat
mempalace --palace /tmp/p status                              # my_project / chat, not ?/?
mempalace --palace /tmp/p search "pricing" --wing my_project  # finds swept drawers

# --room defaults to general
mempalace --palace /tmp/p2 sweep session.jsonl --wing my_project
mempalace --palace /tmp/p2 status                             # my_project / general

# unchanged without --wing
mempalace --palace /tmp/p3 sweep session.jsonl
mempalace --palace /tmp/p3 status                             # ?/?
```

`tests/test_sweeper.py` adds `TestSweeperTaxonomy` and `TestSweeperCLI` covering: wing+room stamped, room defaulting to `general`, no-wing staying unclassified, blank/whitespace wing ignored, `sweep_directory` threading, idempotency preserved, a classified sweep still carrying the `source_dir_ino` that #2368 records, a wing-scoped `sync` reading classified swept drawers and not unclassified ones, the CLI wiring, and the `--room`-without-`--wing` warning. `TestSweeperDaemonRoute` covers the daemon route: the job payload the CLI submits, the same warning on that route (and none for a blank `--room`, or for a sweep that stops on a routing error first), and the worker stamping wing/room for a file and for a directory target. In `tests/test_cli_write_routing.py`, the two sweep tests that pin the exact job payload and the exact `sweep()` call now include `wing`/`room`.

### Notes

- Drawer IDs (`sweep_<session>_<uuid>`) do not depend on wing/room, so idempotency and the timestamp cursor are unchanged; re-running `sweep --wing X` is still a no-op.
- Wing/room are passed through to metadata the same way `mine --wing` does. The CLI miners do not run `sanitize_name`; that validation lives at the MCP write-tool boundary, and wing/room are opaque metadata compared by equality, so there is no new injection surface.
- The cursor makes sweep resume-safe, not re-classifying: `--wing` applies to newly-swept messages, so re-sweeping an already-ingested transcript under a different wing will not retag messages already below the session cursor.
- Sibling ingestion paths (`convo_miner`, `format_miner`, `diary_ingest`, `miner`) already stamp `wing`/`room`; `sweep` was the only one that did not.
- A classified swept drawer is in scope for `sync --wing <wing>` like a mined one: its `source_file` takes part in root auto-detection and it can be pruned as gitignored or missing. An unclassified swept drawer stays outside a wing-scoped sync, and a sync with no wing already read swept drawers before this change.

## Checklist

- [x] No hardcoded paths


mempalace/sweeper.py
```diff
@@ -215,6 +231,16 @@ def sweep(jsonl_path: str, palace_path: str, source_label: Optional[str] = None)
...
     collection = get_collection(palace_path, create=True)
+
+    # Optional taxonomy (#1207, #1979): when a wing is given, swept drawers are
+    # classified under wing/room like ``mine --wing`` (room defaults to the
+    # miners' fallback room "general") so message-level catch-up is searchable
+    # instead of stranded as ``?/?`` in status and search. With no wing we stamp
+    # nothing, preserving the historical unclassified behavior for existing callers.
+    wing = (wing or "").strip() or None
+    room = (room or "").strip() or None
+    taxonomy = {"wing": wing, "room": room or "general"} if wing else {}
```
igorls: **[bug]** Wing and room are only `.strip()`'d, then written straight into drawer metadata. They never pass `config.sanitize_name`. MCP search/list filters run names through `_sanitize_optional_name` / `sanitize_name`, so a value this path will store (`foo/bar`, `_hidden`, over-length, `?`) cannot be used as an MCP `--wing`/`wing=` filter — the drawers are filed and then unreachable by the search path this feature exists to serve. Tests cover both flags, only-wing→`general`, only-room warning, blanks, directory threading, and same-wing idempotency counts; they do not reject invalid names.

**Suggestion:** After the blank-as-unset strip, run `sanitize_name(wing, "wing")` and `sanitize_name(room or "general", "room")`. Raise `ValueError` (CLI: print and exit 1) on invalid input rather than filing a name MCP cannot query. Add a test that `--wing 'foo/bar'` / leading `_` / empty-after-strip does not write metadata.

mempalace/sweeper.py
```diff
@@ -299,6 +325,7 @@ def _flush():
             "source_file": source_label or jsonl_path,
             "filed_at": datetime.now().isoformat(),
             "ingest_mode": "sweep",
+            **taxonomy,
```
igorls: **[bug]** Classification is baked into every metadata dict that reaches `collection.upsert`. Sweep still rewrites every message with `timestamp >= cursor` — including IDs already present. Chroma/backend upsert replaces the metadata document; it does not merge. Consequences: (1) `sweep --wing proj` then a later `sweep` with no `--wing` drops `wing`/`room` from the latest message(s) in each session; (2) `sweep --wing A` then `sweep --wing B` retags those same boundary drawers to B; (3) `get_palace_cursor` treats a backend error as `cursor=None`, so a degraded lookup re-upserts the entire session and can strip or retag every swept drawer. `test_sweep_with_taxonomy_is_still_idempotent` only asserts `drawers_added == 0` on a second same-wing run; it does not assert metadata is stable.

**Suggestion:** For IDs already present, either skip the upsert when document+taxonomy are unchanged, or fetch existing metadata and keep `wing`/`room` unless this run explicitly supplies a wing. Never write `taxonomy={}` over a drawer that already has a wing. Add tests: classified then unclassified must not strip the last-timestamp drawer; classified A then classified B must not retag already-present IDs (or, if retag is intentional, document it and do it for the whole session, not just the cursor tie).

mempalace/sweeper.py
```diff
@@ -215,6 +231,16 @@ def sweep(jsonl_path: str, palace_path: str, source_label: Optional[str] = None)
     * ``drawers_upserted`` — total writes = added + already_present.
     """
     collection = get_collection(palace_path, create=True)
+
+    # Optional taxonomy (#1207, #1979): when a wing is given, swept drawers are
```
igorls: **[suggestion]** The six-line block above the `.strip()` calls restates the new docstring, cites #1207/#1979, and narrates why `?/?` used to happen. Comments should say why a non-obvious constraint exists, not replay the design.

**Suggestion:** Delete the block. The `taxonomy = ... if wing else {}` line plus the docstring Args section already carry the contract.

PR #1999, opened 2026-07-13
feat: MCP streamable-HTTP proxy with production hardening (circuit breaker, watchdog, monitoring)

jjcav84:
## What does this PR do?

Adds a production-grade MCP streamable-HTTP proxy and operational tooling to `deploy/proxy/`. MemPalace's `--transport http` mode speaks plain JSON over HTTP (`BaseHTTPRequestHandler`, `Connection: close`, no SSE), which means MCP clients that expect the streamable-http transport protocol (POST/GET/DELETE `/mcp` with `Mcp-Session-Id` and `text/event-stream`) cannot connect directly. This proxy bridges that gap.

### What's included

**Proxy** (`mempalace-mcp-proxy.py`):
- Streamable-HTTP bridge: translates POST/GET/DELETE `/mcp` to MemPalace's plain JSON HTTP
- Persistent `httpx.AsyncClient` with connection pooling (20 max conns, 10 keepalive)
- Circuit breaker (3 consecutive failures to open, 30s to half-open, probe to close)
- Retry with exponential backoff for transient failures (5xx, timeouts, connect errors)
- Bearer token forwarding (respects `MEMPALACE_MCP_HTTP_TOKEN`)
- `/health` endpoint — tests upstream with actual `tools/list` call, reports circuit state + sessions + uptime
- `/metrics` endpoint — Prometheus-style counters (requests, errors, circuit state, sessions, uptime)
- Session management with TTL cleanup (30min default)
- Structured logging with request IDs
- Graceful shutdown on SIGTERM/SIGINT
- All config via env vars — no hardcoded paths or IPs

**Watchdog** (`mempalace-watchdog.sh`):
- Detects hangs (process alive but not responding to JSON-RPC), not just crashes
- Auto-restart with rate limiting (5/hour max, 60s cooldown between restarts)
- Force-kill (SIGKILL) if SIGTERM doesn't work
- Waits up to 20s for server to become responsive after restart

**Monitor** (`mempalace-monitor.sh`):
- Proactive health checks every 5min (proxy `/health` + MCP `tools/list` + optional SSH)
- Desktop notifications (macOS `osascript` / Linux `notify-send`)
- Auto-remediation via launchd/systemd reload after N consecutive failures

**Deployment templates**:
- macOS launchd plist (`com.mempalace.proxy.plist` — RunAtLoad + KeepAlive)
- Linux systemd units for proxy and watchdog
- Environment file template (`proxy.env.example`)

### Why this is needed

The core MemPalace server is excellent at storage and retrieval but has no operational layer for production deployments. Running it unattended requires:
- Connection pooling (otherwise every request opens a new connection)
- Retry logic (single transient failure shouldn't kill a request)
- Circuit breaker (prevent cascading failures when upstream is down)
- Health endpoint (distinguish "listening on a port" from "actually working")
- Auto-restart on hang (process can be alive but stuck — `Restart=on-failure` doesn't catch this)
- Metrics (observability for monitoring systems)

This package adds all of that **without modifying the MemPalace core** — it's a transparent forwarding layer that respects the "local-first, zero external API" design principle. No telemetry, no phone-home, no data leaves the proxy-to-server path.

## How to test

```bash
# 1. Install dependencies
pip install aiohttp httpx

# 2. Start MemPalace server
mempalace serve --host 127.0.0.1 --port 8765

# 3. Start the proxy
UPSTREAM_URL=http://127.0.0.1:8765/mcp python deploy/proxy/mempalace-mcp-proxy.py

# 4. Test health
curl http://127.0.0.1:8766/health

# 5. Test MCP through proxy
curl http://127.0.0.1:8766/mcp -X POST -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}'

# 6. Run tests
python -m pytest tests/test_proxy.py -v
```

### Verified against a live deployment

Tested against a MemPalace server with 425K drawers across 4 wings:
- 36 MCP tools accessible through the proxy
- Circuit breaker verified: opens on 3 consecutive failures, closes on recovery
- Watchdog verified: 4-second recovery from process kill to fully responsive
- No hardcoded paths or IPs (passes PR checklist)

## Checklist
- [x] Tests pass (8 tests in `tests/test_proxy.py`)
- [x] No hardcoded paths (verified by `TestNoHardcodedPaths` test class)
- [x] Linter passes (`ruff check deploy/proxy/mempalace-mcp-proxy.py`)
- [x] Formatter passes (`ruff format --check deploy/proxy/mempalace-mcp-proxy.py`)
- [x] No telemetry or analytics (respects "local-first, zero external API" principle)
- [x] No summarization of user content (proxy is transparent forwarding only)


fatkobra (changes requested):
Blocking: 
the proxy retries every upstream POST, including non-idempotent JSON-RPC `tools/call` mutations. If the upstream commits an add/update/delete/diary operation and the response is then lost or returns a 5xx, the proxy automatically replays the mutation. There is no end-to-end idempotency key or server-side deduplication that makes this safe.

Please parse the JSON-RPC method and tool and retry only requests proven to be read-only, or add an idempotency token that the upstream persists and deduplicates. The regression should simulate "the first attempt commits, but its response is lost" and verify that the mutation is not executed twice.

jjcav84:
@fatkobra I rebased this onto the latest develop. The retry logic still limits write operations (tools/call add/update/delete/diary) to a single attempt and only retries read-only methods/tools; the 'response lost after commit' regression remains in tests/test_proxy.py. tests/test_proxy.py passes (27 passed). Ready for another review when you have time.

fatkobra (approved):
Thanks for the rebase and follow-up. The default-deny retry classifier, single-attempt treatment of unknown/mutating requests, and response-lost regression are the right overall shape.

fatkobra:
@mvalentsev any feedback here?

mvalentsev:
@fatkobra Yes, four:

1. This breaks CI on `develop`.

2. `_READ_ONLY_TOOLS` still holds a writer: `mempalace_hook_settings` persists `silent_save` / `desktop_toast` through `config.set_hook_setting()`. Idempotent per value, so not the double-add case, but not read-only either. Same for `mempalace_reconnect`.

3. The breaker never sees a 5xx. Line 259: a 5xx on the final attempt falls into `circuit_record_success()`, and mutating calls always run `max_attempts=1`. Three 503s in a row leave the circuit `closed` and `requests_success` at 3. `test_mutating_tool_not_retried_on_5xx` passes either way.

4. `_is_retry_safe` parses inside the `try` but calls `req.get()` outside it, so a batch array or a bare scalar raises `AttributeError` and the client gets a `500` instead of a JSON-RPC error.

fatkobra (changes requested):
`mempalace_memories_filed_away` consumes state by unlinking `~/.mempalace/hook_state/last_checkpoint`. If the first call deletes the file and its response is lost, the proxy retries and the second call returns the different “quiet” result. The original acknowledgement is lost.

`mempalace_hook_settings` also writes configuration whenever arguments are supplied, so retry safety is request-sensitive rather than tool-name-only. `mempalace_reconnect` is likewise documented by the MCP implementation as not write-free and should not be treated as proven retry-safe without a separate idempotency analysis.

Please remove these tools from unconditional retry eligibility, or make classification request-sensitive and based on explicit retry-safety metadata. Add response-lost regressions proving:

- `mempalace_memories_filed_away` is attempted once;
- `mempalace_hook_settings` with update arguments is attempted once; and
- maintenance/reconnect calls are not retried unless their idempotency contract is explicitly established.

There is also a separate circuit-breaker issue. A final upstream 5xx currently falls through to `circuit_record_success()` and increments `requests_success`. When that 5xx has a JSON body, the handler then returns the payload without preserving the upstream status, so the client receives HTTP 200. Persistent JSON-bodied 5xx responses can therefore reset the failure count instead of opening the circuit.

Please classify every final 5xx as a circuit failure, increment the failed rather than successful metric, and preserve the upstream HTTP status in both JSON and SSE responses. Add an end-to-end handler regression showing that repeated JSON-bodied 5xx responses remain 5xx downstream and open the circuit after the configured threshold.

Finally, the Streamable HTTP endpoint currently performs no Origin validation. The MCP transport specification requires invalid Origin headers to be rejected with HTTP 403 to prevent DNS rebinding. The configured upstream token is injected by the proxy and does not authenticate callers to the proxy itself.

Please add an allowed-Origin/Host policy at the inbound boundary, reject invalid origins before processing JSON-RPC, and add inbound client authentication for deployments that are not strictly loopback-only. Include tests for a valid origin, an invalid origin receiving 403, and a non-browser request with no Origin under the intended local-client policy.

jjcav84:
Hi maintainers — this PR is rebased onto the latest  and the proxy test suite passes. Ready for review.

jjcav84:
@fatkobra addressed the three review points:\n\n1. Removed , , and unconditional  from the read-only retry allowlist.  is now request-sensitive: only retried when called with no arguments (status query); any arguments disable retries.\n2. Every upstream 5xx now counts as a circuit failure (), increments , and preserves the upstream HTTP status in both JSON and SSE responses. Added a regression showing repeated 5xx opens the circuit.\n3. Added , , and  inbound validation: invalid Origins get 403, non-browser clients must use an allowed Host, and deployments can require a Bearer token.\n\n passes (38 tests).

jjcav84:
@fatkobra addressed the three review points in this push:

1. Retry safety: removed `mempalace_memories_filed_away`, `mempalace_reconnect`, and unconditional `mempalace_hook_settings` from the read-only retry allowlist. `mempalace_hook_settings` is now request-sensitive — retried only with no arguments (status query); any arguments disable retries. Added regressions for each case.
2. Circuit breaker / 5xx: every upstream 5xx now counts as a failure (`circuit_record_failure`), increments `requests_failed`, and preserves the upstream HTTP status in both JSON and SSE responses. Added a regression showing repeated 5xx opens the circuit.
3. Origin/Host: added `ALLOWED_ORIGINS`, `ALLOWED_HOSTS`, and `INBOUND_TOKEN` validation. Invalid Origins get 403, non-browser clients must use an allowed Host, and deployments can require an inbound Bearer token.

`tests/test_proxy.py` passes (38 tests).

jjcav84:
@fatkobra added the requested end-to-end handler regressions:

- Repeated JSON-bodied 5xx responses now return HTTP 503 downstream and open the circuit after the configured threshold.
- Origin/Host validation is exercised through the real aiohttp request path, confirming 403 for an invalid origin and 202 for both a valid Origin and a non-browser request with an allowed Host.

`tests/test_proxy.py` passes (40 tests), including the new `TestEndToEnd` class.

jjcav84:
@fatkobra all points from the latest `CHANGES_REQUESTED` review are addressed and the requested end-to-end regressions are in. Full suite passes. Could you re-review and merge when ready? @milla-jovovich @igorls also welcome to weigh in.

fatkobra (changes requested):
Thanks — I re-reviewed the latest implementation, and my three previous
concerns are resolved:

- retry safety is now determined from the request;
- upstream 5xx responses contribute to circuit failure while preserving the
  original status;
- inbound Origin, Host, and token validation is in place.

One existing blocker remains before I can approve: the JSON request-shape
issue identified by @mvalentsev.

`_is_retry_safe()` assumes that successfully parsed JSON is a mapping and
calls `.get()` on it. A valid JSON-RPC batch array, or scalar JSON such as
`null`, a string, or a number, can therefore escape as an unhandled
exception and produce a raw HTTP 500.

Please validate the parsed shape before retry classification.

The proxy should either:

- support JSON-RPC batch arrays correctly; or
- reject unsupported shapes with a structured JSON-RPC `-32600 Invalid
  Request` response.

It should never turn a client-provided valid JSON value into an unhandled
server error.

Please add real HTTP-handler coverage for:

- a batch array;
- `null`;
- a string or number;
- a valid object missing expected JSON-RPC fields;
- and a normal single-request object.

Once that existing finding is resolved, I will be comfortable approving.

jjcav84:
Hi @fatkobra — the JSON shape blocker should now be resolved:

- `_is_valid_jsonrpc_request` now returns a JSON-RPC error code: `-32700` for unparseable JSON, `-32600` for valid JSON that is not a valid JSON-RPC object.
- `handle_mcp_post` returns a structured `-32600` response for non-object shapes including `null`, strings, numbers, and batch arrays.
- `_is_retry_safe` only proceeds after validation passes, so it never calls `.get()` on a non-mapping.
- Added real HTTP-handler regression tests for: batch array, `null`, string, number, invalid object missing `jsonrpc`, and a normal single request.

Verification:
- `pytest tests/test_proxy.py` — 47 passed
- `ruff check deploy/proxy/mempalace_mcp_proxy.py tests/test_proxy.py` — clean

fatkobra (approved):
Thanks — I reviewed `e281db6`, and the remaining JSON request-shape blocker
is resolved.

The proxy now validates the parsed JSON value before retry classification or
request-field access. Malformed JSON returns `-32700`, while valid non-object
values—including batch arrays, `null`, strings, and numbers—return a
structured `-32600 Invalid Request` response and are not forwarded upstream.

The real aiohttp handler regressions cover those invalid shapes, and a normal
single JSON-RPC request object remains accepted and forwarded.

Together with the earlier request-sensitive retry classification, final-5xx
circuit accounting and status preservation, and Origin/Host/inbound-auth
validation, my blocking concerns are resolved.

Approved.

jjcav84:
@mvalentsev thanks for the four points — all are addressed in the current push:

1. **CI on develop**: rebased onto latest `develop`, full suite passes (4515 passed, 31 skipped). `tests/test_proxy.py` — 47 passed.

2. **`_READ_ONLY_TOOLS` held a writer**: `mempalace_hook_settings` moved to `_CONDITIONALLY_READ_ONLY_TOOLS` and is now request-sensitive — retried only with no arguments (status query); any arguments disable retries. `mempalace_reconnect` removed from the allowlist entirely. Regressions added for each case.

3. **Breaker never saw 5xx**: every upstream 5xx now calls `circuit_record_failure()` (line 336), increments `requests_failed`, and preserves the upstream HTTP status in both JSON and SSE responses. Regression: repeated 5xx opens the circuit.

4. **`_is_retry_safe` AttributeError on non-dict JSON**: `_is_valid_jsonrpc_request` now validates the parsed shape before `_is_retry_safe` runs. Non-object JSON (batch arrays, `null`, strings, numbers) returns a structured `-32600 Invalid Request` response. `_is_retry_safe` only proceeds after validation passes. Real HTTP-handler regressions cover each case.

@fatkobra has approved. Could you re-review when you have a moment?

jjcav84:
Hi @igorls @milla-jovovich — this PR was approved by @fatkobra on Aug 15 and rebased onto latest develop on Aug 22. The full test suite passes (4512 passed, 31 skipped) and ruff is clean. It's been production-hardened through 3 review rounds (idempotent retries, circuit breaker on 5xx, Origin/Host validation, JSON-RPC shape validation). Ready for merge when you have a moment.

jjcav84:
This PR was approved by @fatkobra on Aug 15 and has been rebased onto latest develop. Full test suite passes (4927 passed, 53 skipped). Ruff clean. It's been pending merge for ~4 weeks. Could a maintainer merge when convenient? @milla-jovovich @igorls

igorls:
Reviewed head `10b10fde81c68ee848e8c765f7d04ec1f64f3bc1`. I recommend changes before merging; four issues remain:

1. **[P1] Validate Host independently of Origin** — `deploy/proxy/mempalace_mcp_proxy.py:463-464`. Matching attacker-controlled headers such as `Host: attacker.example:8766` and `Origin: http://attacker.example:8766` bypass `ALLOWED_HOSTS`. With the default empty `INBOUND_TOKEN`, a DNS-rebinding page can reach the localhost proxy, which attaches its configured upstream credentials. I reproduced HTTP 200 and forwarding of a mutation request to a mocked upstream using these headers. Require an independently trusted Host before accepting a same-origin request, and add this case to the handler regressions.

2. **[P1] The entry point crashes on Python 3.14** — `deploy/proxy/mempalace_mcp_proxy.py:774`. Running `python deploy/proxy/mempalace_mcp_proxy.py` raises `RuntimeError: There is no current event loop in thread 'MainThread'` at `asyncio.get_event_loop()`, before the listener starts. The repository explicitly supports Python 3.14. Initialize the loop through the application lifecycle and cover actual entry-point startup.

3. **[P1] Quote the watchdog's systemd environment assignment** — `deploy/proxy/mempalace-watchdog.service:30`. Systemd parses the unquoted assignment as `MEMPALACE_START_CMD=mempalace`, discarding `serve --host 0.0.0.0 --port 8765`. Confirmed with `systemd-analyze verify`, which reports those arguments as invalid environment assignments. On recovery, the script kills the server and launches the bare CLI instead of restarting it. Use `Environment="MEMPALACE_START_CMD=mempalace serve --host 0.0.0.0 --port 8765"`.

4. **[P2] Preserve writable palace storage for the replacement server** — `deploy/proxy/mempalace-watchdog.service:35-38`. The server spawned by the watchdog inherits `ProtectSystem=strict`, but this unit has no writable exception for the palace. After the command fix, a palace under `/var/lib/mempalace` still cannot initialize or perform normal database writes. The existing server unit supplies `ReadWritePaths` and `StateDirectory`. Prefer restarting the actual server service, or preserve its storage permissions and configuration in the replacement process. This finding is based on tracing the unit and child-process configuration, not a live sandbox test.

Light MCP compatibility also needs to be explicit in the documentation. The existing path is `client -> light MCP over stdio (palace_query / palace_exec / palace_coordinate) -> full HTTP hub`; this PR adds `client -> HTTP proxy -> full HTTP hub`. The proxy forwards `tools/list` unchanged, so clients connecting through it receive the full hub catalog, not the three consolidated light tools. The light server currently only supports stdio and cannot be used directly as this proxy's HTTP upstream.

There is no direct code conflict with the existing light MCP path, and merging this optional proxy does not replace it. Deploying the watchdog can affect that path because it manages the same upstream hub. Please document this scope; if the intended feature is light MCP over HTTP, it requires integration with the light dispatcher and an end-to-end test asserting the three-tool catalog and correct dispatch.

Validation: `python -m pytest tests/test_proxy.py -q --override-ini addopts=''` passed all 47 tests. Startup and authorization were checked separately as described above; the passing handler suite does not cover those failures.

jjcav84:
Thanks @igorls — all four issues are addressed in the latest push (b3a2573):

1. **[P1] Host validated independently of Origin** — `_is_inbound_request_allowed()` now checks `Host` first, before Origin. A DNS-rebinding page with `Host: attacker.example:8766` and `Origin: http://attacker.example:8766` is rejected with 403 because the Host is not in `ALLOWED_HOSTS`. Added regression tests: `test_dns_rebinding_attacker_host_and_origin_rejected` and `test_valid_origin_with_disallowed_host_rejected`.

2. **[P1] Python 3.14 entry point** — replaced `asyncio.get_event_loop()` (which raises `RuntimeError` on 3.14 when no loop exists) with signal registration inside the async startup lifecycle via `get_running_loop()`.

3. **[P1] Watchdog systemd Environment quoted** — `Environment="MEMPALACE_START_CMD=mempalace serve --host 0.0.0.0 --port 8765"` so systemd does not discard the arguments.

4. **[P2] Watchdog writable storage** — added `StateDirectory=mempalace` and `ReadWritePaths=/var/lib/mempalace /var/log/mempalace` so the replacement server can write under `ProtectSystem=strict`.

5. **Light MCP compatibility** — documented in the proxy README that this proxy forwards `tools/list` unchanged (full catalog, not the three consolidated light tools), and that the watchdog manages the same upstream hub process.

Full test suite passes (49 tests in test_proxy.py). Ruff clean. Ready for re-review.

jjcav84:
Verified at head `b3a2573`: all four review issues are addressed — Host is validated independently before Origin, the Python 3.14 `get_event_loop()` crash is fixed via `get_running_loop()` in the async lifecycle, the watchdog unit's `Environment` assignment is quoted, and writable palace storage is preserved. The retry path also gates on `_is_retry_safe()` — non-idempotent `tools/call` mutations get a single attempt, only proven read-only tools/methods retry — with regression coverage in `test_proxy.py`. 13 checks green. Ready for re-review @igorls @fatkobra.

igorls (changes requested):
Re-reviewed at b3a2573. Thanks for working through retry classification, 5xx accounting, JSON shape validation and the Host-before-Origin ordering; those all look right now. Three things still block:

1. **The entry point doesn't start on any Python version.** `_setup_signals(app._runner)` at line 789 raises `AttributeError: 'Application' object has no attribute '_runner'`. I reproduced it on 3.12 and 3.14. Please delete the custom signal code rather than patching it. `web.run_app` already handles SIGINT/SIGTERM. Your `add_signal_handler` calls replace aiohttp's handlers with one that only closes the httpx client, so the process wouldn't exit on SIGTERM. Please also add a test that actually starts `main()`.
2. **Non-loopback binds are open.** The Host allowlist only stops browsers. With `HOST=0.0.0.0` and no `INBOUND_TOKEN`, `curl -H 'Host: 127.0.0.1:8766' http://<lan-ip>:8766/mcp` gets a 200, and the proxy forwards `add_drawer` upstream with the hub's bearer token. The hub refuses a non-loopback bind without a token (`http.py:950`), and the proxy should do the same. Compare tokens with `hmac.compare_digest`, not `!=`. `/health` and `/metrics` also skip every inbound check: from the LAN, `Host: evil.example` gets a 200 that shows the upstream URL.
3. **The watchdog conflicts with `deploy/mempalace-server.service`.**
   - Its health check has no token by default, so it gets a 401 and pkills a healthy server.
   - The replacement it starts (`serve --host 0.0.0.0`) has no `MEMPALACE_MCP_HTTP_TOKEN`, so the hub refuses to bind. Meanwhile systemd restarts its own copy, and the two race for the port and the writer lease.
   - `ReadWritePaths=/var/log/mempalace` names a directory nothing creates.

   Please restart the real unit (`systemctl restart mempalace-server`) and check `/healthz`, or drop the watchdog from this PR.

On scope: the hub already does Host/Origin/token checks and has `/healthz` and `/statusz`. The actual streamable-HTTP gap I can see is that `GET /mcp` returns 404 and `DELETE /mcp` returns 501, where the spec allows 405. I'd rather fix that in `mcp_server/http.py` than keep a second process holding the hub token. Happy to take that as a separate small PR.

If the proxy stays:
- Keep it clearly as a `deploy/` example.
- Have its tests use `pytest.importorskip("aiohttp")` rather than adding aiohttp to everyone's dev dependencies.
- Document `ALLOWED_HOSTS`, `ALLOWED_ORIGINS` and `INBOUND_TOKEN`.
- Update the README's "no health endpoint / no auto-restart" rationale, which is out of date.

jjcav84:
@igorls all three blockers addressed in `1f6dab4`:

1. **Entry point** — the custom signal code is deleted entirely (no `app._runner` access, no handler replacement). `web.run_app` owns SIGINT/SIGTERM and `on_cleanup` closes the httpx client through its graceful shutdown. New `TestEntryPoint` subprocess tests actually start `main()`: one asserts the listener comes up and SIGTERM exits cleanly, one asserts the non-loopback refusal, one asserts the opt-out env.

2. **Non-loopback binds** — `main()` now refuses `HOST=0.0.0.0` (or any non-loopback) without `INBOUND_TOKEN`, same rule as the hub, with `PROXY_ALLOW_INSECURE_NO_TOKEN=1` as the documented opt-out for trusted fronting layers. Token comparison uses `hmac.compare_digest`. `/health` and `/metrics` now pass through `_is_inbound_request_allowed`, so the upstream URL is no longer disclosable from the LAN.

3. **Watchdog dropped** — `mempalace-watchdog.sh` + `mempalace-watchdog.service` removed from the PR (restarting the real unit is the right answer; the proxy's own units stay). `mempalace-monitor.sh` now carries `PROXY_TOKEN` as a Bearer on its `/health` and `/mcp` checks.

Scope items: tests use `pytest.importorskip("aiohttp")`/`importorskip("httpx")` and the pyproject/uv.lock dep changes are reverted — the proxy stays a `deploy/` example with its own runtime deps. README documents `ALLOWED_HOSTS`, `ALLOWED_ORIGINS`, `INBOUND_TOKEN`, and the non-loopback rule, and the stale capability claims are corrected (the hub already has `/healthz` and `/statusz`).

`tests/test_proxy.py` — 48 passed. Ruff clean.

Happy to take the `GET /mcp` → 405 / `DELETE /mcp` fix as the separate small PR you suggested — that one belongs in `mcp_server` proper.

PR #2005, opened 2026-07-14
feat(chroma): opt-in HTTP client/server mode for multi-process palaces

### LINKED ISSUE #832: [Feature] Support standalone ChromaDB HTTP Client to prevent SQLite concurrency crashes
### 🐛 Problem / Issue
When running `mcp_server` and `mempalace mine` simultaneously (or running multiple concurrent miners), users frequently encounter fatal crashes such as:
`chromadb.errors.InternalError: Error in compaction: Failed to apply logs to the hnsw segment writer`

This occurs because `ChromaBackend` is currently hardcoded to unconditionally instantiate `chromadb.PersistentClient()`. According to ChromaDB's local persistence architecture, the embedded SQLite database and HNSW segment writers are inherently **not thread/process-safe** for concurrent writers. As a result, running the MCP server and sync routines synchronously triggers destructive segment locks, inevitably corrupting the index data. 

### 💡 Solution / Changes
To support advanced workflows, MemPalace needs the ability to connect to a standalone Chroma Server via HTTP (`chroma run`), which handles concurrent queries and index serialization safely.

**Changes introduced in this PR:**
1. Modified `config.py` to support `chroma_server_host` and `chroma_server_port` properties via environmental variables or `config.json`.
2. Modified `ChromaBackend.get_collection()` in `backends/chroma.py`. If a `host` configuration is provided, the backend elegantly bypasses local file bindings (e.g. `chmod`, directory creations) and switches to an isolated `chromadb.HttpClient`. 

> **Important Usage Note for users testing this PR:**
> If users utilize a global Web Proxy (e.g., `HTTP_PROXY`), they should be reminded to correctly configure `NO_PROXY=localhost,127.0.0.1` locally, otherwise `httpx` underlying `HttpClient` might inadvertently route internal `localhost` traffic to the proxy, resulting in 502 Bad Gateways.

#### 1. `mempalace/config.py`
```diff
@@ -174,6 +174,22 @@
         return self._file_config.get("hall_keywords", DEFAULT_HALL_KEYWORDS)
 
+    @property
+    def chroma_server_host(self):
+        """Host for ChromaDB standalone server. If set, HttpClient is used."""
+        env_val = os.environ.get("MEMPALACE_CHROMA_HOST")
+        if env_val:
+            return env_val
+        return self._file_config.get("chroma_server_host")
+
+    @property
+    def chroma_server_port(self):
+        """Port for ChromaDB standalone server."""
+        env_val = os.environ.get("MEMPALACE_CHROMA_PORT")
+        if env_val:
+            return int(env_val)
+        return self._file_config.get("chroma_server_port", 8000)
```

#### 2. `mempalace/backends/chroma.py`
```diff
@@ -7,6 +7,7 @@
 
 import chromadb
 
+from ..config import MempalaceConfig
 from .base import BaseCollection
 
 logger = logging.getLogger(__name__)
@@ -71,18 +72,24 @@
     """Factory for MemPalace's default ChromaDB backend."""
 
     def get_collection(self, palace_path: str, collection_name: str, create: bool = False):
-        if not create and not os.path.isdir(palace_path):
-            raise FileNotFoundError(palace_path)
-
-        if create:
-            os.makedirs(palace_path, exist_ok=True)
-            try:
-                os.chmod(palace_path, 0o700)
-            except (OSError, NotImplementedError):
-                pass
-
-        _fix_blob_seq_ids(palace_path)
-        client = chromadb.PersistentClient(path=palace_path)
+        cfg = MempalaceConfig()
+        host = cfg.chroma_server_host
+
+        if host:
+            client = chromadb.HttpClient(host=host, port=cfg.chroma_server_port)
+        else:
+            if not create and not os.path.isdir(palace_path):
+                raise FileNotFoundError(palace_path)
+
+            if create:
+                os.makedirs(palace_path, exist_ok=True)
+                try:
+                    os.chmod(palace_path, 0o700)
+                except (OSError, NotImplementedError):
+                    pass
+
+            _fix_blob_seq_ids(palace_path)
+            client = chromadb.PersistentClient(path=palace_path)
         if create:
             collection = client.get_or_create_collection(collection_name)
```

### LINKED ISSUE #1096: Feature request: support `chromadb.HttpClient` mode for multi-process palace access
## Summary

Add an opt-in backend mode to `ChromaBackend` that uses `chromadb.HttpClient` (talking to a user-managed `chroma run` server) instead of `chromadb.PersistentClient`. MemPalace today hardcodes `PersistentClient`, which is not safe when more than one process (e.g. multiple Claude Code sessions, each spawning `mempalace.mcp_server`, plus the mining CLI) writes to the same palace concurrently.

## Background — real-world incident

On 2026-04-21 a palace I run as a single user under four concurrent Claude Code sessions (→ four `mempalace.mcp_server` processes, plus `mempalace mine` firing from the auto-save Stop hook) corrupted its HNSW index:

- Collection had ~4,800 embeddings.
- `link_lists.bin` grew to **2.5 TB apparent / ~1.8 TB allocated** — roughly 1000× expected size — filling the entire disk.
- `chromadb` queries began segfaulting (exit 139) against that store.
- Recovery required killing all writers, deleting the HNSW segment directory, and letting Chroma rebuild from the (intact) SQLite embeddings table.

This is consistent with the two known failure modes that compound under concurrent writers:

1. Chroma’s own [system constraints doc](https://cookbook.chromadb.dev/core/system_constraints/) states `PersistentClient` is not process-safe.
2. The bug fixed in #544 (convo_miner using `add` instead of `upsert`) produced duplicate HNSW inserts per mining run, with "terabytes" and "segfault" called out by name in the PR.

Fixing #544 prevents duplicate inserts during mining, but it does not address the underlying multi-writer race in `PersistentClient`: any two concurrent writers (two MCP servers, or one MCP server + one `mempalace mine`) can still corrupt the HNSW files. The official Chroma recommendation for multi-process access is `chroma run` + `HttpClient`.

## Proposal

Add a config surface for HTTP mode:

```jsonc
// ~/.mempalace/config.json
{
  "palace_path": "/Users/.../.mempalace/palace",
  "collection_name": "mempalace_drawers",
  "backend": {
    "mode": "http",             // "persistent" (default) | "http"
    "host": "127.0.0.1",
    "port": 8000,
    "ssl": false
  }
}
```

Implementation sketch for `backends/chroma.py`:

- Extend `ChromaBackend.__init__` to read the `backend.mode` field.
- In `_client` (line 423) and `make_client` (line 481), branch on mode:
  ```python
  if mode == "http":
      return chromadb.HttpClient(host=cfg.host, port=cfg.port, ssl=cfg.ssl)
  _fix_blob_seq_ids(palace_path)
  return chromadb.PersistentClient(path=palace_path)
  ```
- In HTTP mode, skip `_fix_blob_seq_ids` and the inode/mtime freshness cache (they are filesystem-level concerns — the server owns them).
- Gate capabilities: HTTP mode should advertise `local_mode=False`.

Documentation additions:

- A short "Multi-process palaces" section explaining when to use HTTP mode.
- A sample `launchd` plist (macOS) / `systemd` unit (Linux) to run `chroma run --path ~/.mempalace/palace --port 8000` as a user-level service.

## Backwards compatibility

- Default remains `persistent` — zero change for single-process users.
- The MCP server and mining CLI both go through `backends.chroma.ChromaBackend`, so a single config flip covers all call sites.
- `repair` (`cli.py::cmd_repair`) should detect HTTP mode and either refuse (with a clear message pointing at Chroma’s server-side rebuild story) or proxy the rebuild through the server.

## Why this matters for MemPalace specifically

MemPalace is explicitly designed to be invoked from background hooks (`mempal_save_hook.sh` → `mempalace mine`) concurrently with interactive MCP servers. That usage pattern is the *exact* case Chroma says `PersistentClient` does not support. Supporting `HttpClient` removes a latent corruption surface that is very hard for end users to diagnose after the fact.

## Offer

Happy to submit a draft PR if the maintainers would accept this direction.

---

*Filed by a MemPalace user after resolving the 1.8 TB HNSW bloat incident described above. Full recovery writeup + chromadb 0.6.3 vs 1.5.8 behavior notes available on request.*

ahabshamaa:
## What does this PR do?

Adds an opt-in HTTP client/server mode for the ChromaDB backend, as requested in #832 and #1096 — and implements the "ChromaDB HTTP server mode" option named in the architectural decision epic #1963.

**Problem.** Embedded `chromadb.PersistentClient` is not process-safe. When more than one process writes the same palace — multiple MCP server instances, stop-hook mines firing alongside an interactive session, a manual `mempalace mine` — writers contest the same sqlite/HNSW files and can corrupt the index (the #1963 cluster). Worse, a blocked writer waits on the file lock with no timeout, so palace calls can hang the MCP server indefinitely.

**Fix.** With `MEMPALACE_CHROMA_MODE=http`, one standalone `chroma run` server exclusively owns the palace directory and every MemPalace process is a thin HTTP client constructed in one place (`mempalace/palace_client.py`). The server serializes writes where ChromaDB actually supports concurrency; clients get:

- **Health probe** — construction hits `/api/v2/heartbeat` (3 s timeout) and raises `PalaceBackendUnreachableError` naming host:port instead of hanging.
- **Hard per-operation timeout** — every wire call runs under a 15 s ceiling (`MEMPALACE_OP_TIMEOUT`) via a `TimeoutProxy` wrapping the client and every collection obtained from it; timeouts raise `PalaceOperationTimeoutError` naming the operation.
- **One error shape** — transport failures mid-call (`httpx.TransportError`, `ConnectionError`) are translated into the same structured unreachable error the probe gives; the MCP server returns these as structured tool errors, never hangs, never retries (the bound already happened).

**Default unchanged.** `embedded` remains the default: zero change and no external service for single-process installs. Config surface: `MEMPALACE_CHROMA_MODE`, `CHROMA_HOST`/`CHROMA_PORT` (default `127.0.0.1:8801`), `MEMPALACE_OP_TIMEOUT`.

**Embedded-only mechanisms gated off in HTTP mode** (each is correct for a process that owns the files, wrong for a thin client):

- the pre-open repair pass — client-side repair against a live server's files is a corruption risk;
- the inode-keyed client cache and mtime reconnect — file identity is meaningless over HTTP, and the server bumps mtime on every write;
- the HNSW capacity probe — #1222 guards an embedded-only segfault; left on it would wrongly route search to BM25;
- the collection `_write_lock` flock — #974/#965 guarded embedded multi-threaded corruption; kept on it turns legitimate concurrent writers into `MineAlreadyRunning` errors;
- the MCP peer-writer lease (#1818) — in HTTP mode it would force peer servers read-only, the same incompatibility #1888 reports for daemon topologies.

Mining passes still take `mine_palace_lock` for application-level atomicity in both modes.

**Docs:** `docs/chroma-client-server.md` — topology, config, error contract, launchd + systemd service examples, heartbeat/backup/vacuum runbook (#1096 asked for exactly this).

**Relationship to the daemon+bridge direction (#1270/#1976):** complementary, not competing — this is opt-in, changes nothing by default, and gives multi-process users a supported deployment today at the layer where ChromaDB itself supports concurrency.

Closes #832. Closes #1096. Related to #1963, #1581, #1799, #1888.

## How to test

```bash
python -m pytest tests/ -v                      # suite forces embedded mode; 17 new tests in tests/test_palace_client.py
# live smoke test:
chroma run --path /tmp/palace-http --port 8801 &
MEMPALACE_CHROMA_MODE=http mempalace init ~/somewhere && MEMPALACE_CHROMA_MODE=http mempalace mine ~/somewhere
# kill the server mid-session → structured PalaceBackendUnreachableError, no hang
```

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — 3284 passed; the 2 failures in `test_repair.py` (`_errors_are_isolated_fts5` string matching) fail identically on clean `develop` in this environment
- [x] No hardcoded paths

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ non-default backend is opt-in.
 
 | Backend | Mode | Install | Namespaces | Lexical | Configure with |
 | ------- | ---- | ------- | :--------: | :-----: | -------------- |
-| `chroma` _(default)_ | Local (embedded) | bundled | – | ✓ | – |
+| `chroma` _(default)_ | Local (embedded) · Server opt-in | bundled | – | ✓ | `MEMPALACE_CHROMA_MODE=http` |
 | `sqlite_exact` | Local (exact) | bundled | – | ✓ | – |
 | `milvus` | Local (Lite) · Server opt-in | `mempalace[milvus]` | ✓ | ✓ | `MEMPALACE_MILVUS_URI` |
 | `qdrant` | Server (REST) | bundled | ✓ | ✓ | `MEMPALACE_QDRANT_URL` |
@@ -125,6 +125,13 @@ Select with `--backend <name>`, `MEMPALACE_BACKEND=<name>`, or
 `"backend": "<name>"` in `config.json`. See
 [Storage backends](/guide/configuration#storage-backends) for connection
 variables, namespace behavior, and deployment notes.
+
+When more than one process shares a Chroma palace (multiple MCP servers,
+hook-triggered mines), embedded mode is unsafe — set
+`MEMPALACE_CHROMA_MODE=http` and run one standalone `chroma run` server
+that owns the palace directory; all MemPalace processes become thin HTTP
+clients with structured errors and hard timeouts. See
+[docs/chroma-client-server.md](docs/chroma-client-server.md).
 
 ## Quickstart
 
--- a/docs/chroma-client-server.md
+++ b/docs/chroma-client-server.md
@@ -1,0 +1,213 @@
+# Chroma HTTP mode: client/server topology for multi-process palaces
+
+Opt-in transport for the ChromaDB backend (#832, #1096): one standalone
+`chroma run` server exclusively owns the palace directory, and every
+MemPalace process (MCP servers, hooks, mining CLI) is a thin HTTP client.
+Enable it with `MEMPALACE_CHROMA_MODE=http`.
+
+## Why
+
+Embedded `chromadb.PersistentClient` is not process-safe: when more than
+one process writes the same palace — multiple MCP server instances,
+stop-hook mines firing alongside an interactive session, a manual
+`mempalace mine` — writers contest the sqlite file lock and can corrupt
+the HNSW index (see the incident cluster tracked in #1963). Worse, a
+blocked writer waits on the file lock with no timeout, so palace calls
+can hang the MCP server indefinitely.
+
+In HTTP mode, one server process owns the files and serializes writes;
+concurrency is handled where ChromaDB supports it. Every client call is
+health-probed and timeout-bounded, so failures surface as structured
+errors instead of hangs.
+
+Single-process installs need none of this: the default remains
+`embedded`, with zero configuration and no external service.
+
+## Topology
+
+- **One server**: `chroma run --path <palace-path> --host 127.0.0.1
+  --port 8801`, typically managed as a user service (examples below). It
+  must be the **only** process that opens the files under the palace
+  directory.
+- **Thin clients**: every MemPalace process constructs its Chroma client
+  through [`mempalace/palace_client.py`](../mempalace/palace_client.py) —
+  the single construction point. In HTTP mode nothing outside that module
+  may build a Chroma client; that rule is what prevents a second embedded
+  writer from re-appearing.
+- The client returned by `make_http_client()` is wrapped in a
+  `TimeoutProxy` that also wraps every collection obtained from it, so the
+  whole call graph is timeout-guarded without call-site changes.
+
+## Config surface
+
+| Variable | Default | Meaning |
+| --- | --- | --- |
+| `MEMPALACE_CHROMA_MODE` | `embedded` | Set `http` to enable client/server mode. |
+| `CHROMA_HOST` | `127.0.0.1` | Server address the clients connect to. |
+| `CHROMA_PORT` | `8801` | Server port. |
+| `MEMPALACE_OP_TIMEOUT` | `15` (seconds) | Hard per-operation ceiling on every wire call. |
+
+While the server is running, never open the same palace in embedded mode
+(e.g. offline repair tooling): two writers on the same sqlite/HNSW files
+is the corruption scenario HTTP mode exists to remove. Stop the server
+first.
+
+## Error contract
+
+Palace failures in HTTP mode are **structured errors, never hangs**. If a
+call blocks forever, that is a bug in this contract, not expected
+behavior. The two shapes (both in `mempalace/palace_client.py`):
+
+- `PalaceBackendUnreachableError` — the server did not answer. Raised by
+  the construction-time health probe (`GET /api/v2/heartbeat`, 3 s
+  timeout, error names `host:port`) **and** by in-flight calls whose
+  transport dies mid-request (`httpx.TransportError` / builtin
+  `ConnectionError` are translated into this same shape), so consumers see
+  one error whether the server died before or during the call.
+- `PalaceOperationTimeoutError` — a single operation exceeded the
+  `MEMPALACE_OP_TIMEOUT` ceiling. The error names the operation (e.g.
+  `collection[memories].query`) and the backend address.
+
+The MCP server catches both and returns a structured tool error with a
+recovery hint. Errors are not retried internally — the probe/timeout
+already bounded the wait; a retry would just double it.
+
+## Running the server as a user service
+
+### macOS (launchd)
+
+`~/Library/LaunchAgents/com.example.chroma-server.plist` (adjust the
+`chroma` binary path — `which chroma` — and the palace path):
+
+```xml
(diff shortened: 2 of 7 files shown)


fatkobra (changes requested):
Blocking: this is a timeout on the caller's wait, not a timeout or cancellation of the Chroma operation. The implementation explicitly leaves a timed-out call running and then reports failure. For `add`, `upsert`, `update`, or `delete`, the caller can retry while the original request later commits, producing duplicate, reordered, or overlapping effects.

Repeated hangs also occupy all eight executor workers. Subsequent healthy calls can remain queued and time out before they ever start, making the failure self-sustaining.

Please use transport-level connect/read/write/pool timeouts enforced by the HTTP client. An ambiguous write timeout must be reported as "outcome unknown" and reconciled with an idempotency key or status check rather than treated as a clean failed operation. Add tests for commit-after-timeout and executor saturation.

igorls:

PR #2007, opened 2026-07-14
fix(chroma): persist and safely recover memory indexes

phileokairos-pixel:
## What changed

- Reduce the HNSW batch and sync checkpoint from 50,000 to 10,000 records.
- Recognize collections below their first sync threshold as healthy instead of quarantining their preallocated HNSW segment.
- Report that below-threshold state as expected and healthy.
- Rebuild FTS5 and vacuum SQLite after a `from-sqlite` recovery closes Chroma's database handles.
- Add regression coverage for collection creation, cold-start quarantine, capacity status, and SQLite recovery cleanup.

## Why

The 50,000-record guard prevented ordinary palaces from ever writing `index_metadata.pickle`. After restart, the cold-start safety check interpreted the preallocated segment as an interrupted write and quarantined healthy HNSW data. Separately, large recovery upserts could leave Chroma's derived FTS5 index malformed even when every source row survived.

## Impact

Medium-sized palaces now persist a durable semantic index without returning to Chroma's high-frequency default flush behavior. Legitimately unflushed small collections stay usable, while unknown or structurally corrupt segments remain subject to quarantine. SQLite recovery now completes the full text-index repair automatically.

## Validation

- `189 passed` across backend, HNSW capacity, HNSW payload, and repair tests.
- `4 passed` in the focused post-commit regression run.
- A real recovery rebuilt 22,492 rows, passed SQLite `quick_check` and `integrity_check`, produced durable HNSW metadata, and returned semantic plus BM25 search results.

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -114,10 +114,12 @@ def _hnsw_payload_appears_sane(seg_dir: str) -> bool:
 # cycle compounding the next. Result: link_lists.bin grows into hundreds of
 # GB sparse, after which `status`/`search`/`repair` segfault.
 #
-# Setting large batch and sync thresholds at collection creation defers
-# persistence until a single large batch completes, breaking the resize+
-# persist feedback loop. Empirically validated on a 39,792-drawer rebuild
-# (palace 376 MB, link_lists.bin 0 bytes, no segfault) in 2026-04.
+# Setting moderately large batch and sync thresholds at collection creation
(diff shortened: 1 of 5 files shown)


phileokairos-pixel:
Superseded by a narrower rebased PR. Upstream develop already contains the HNSW threshold/quarantine fix, so the replacement keeps only the missing FTS5 cleanup after SQLite recovery.

PR #2008, opened 2026-07-14
fix(repair): rebuild FTS5 after SQLite recovery

phileokairos-pixel:
## What changed

- Run the existing FTS5 rebuild and SQLite vacuum after `rebuild_from_sqlite` finishes its Chroma upserts.
- Move the success return until after cleanup completes.
- Ensure Chroma's backend closes before cleanup so SQLite can obtain the exclusive lock required on Windows.
- Add a regression test proving the cleanup hook runs after a successful SQLite recovery.

## Why

Large bulk recovery upserts can leave Chroma's derived `embedding_fulltext_search` FTS5 index malformed even when every source row survives. The legacy rebuild path already performs this cleanup, but the corruption-recovery path that reads directly from SQLite did not.

That allowed `rebuild_from_sqlite` to report success while a later `PRAGMA quick_check` still failed with a malformed FTS5 inverted index.

## Impact

SQLite-based recovery now completes the same derived-index cleanup as the legacy rebuild. If FTS5 rebuild or vacuum fails, recovery no longer prints a successful completion message.

## Validation

- `92 passed` in `tests/test_repair.py` on the latest upstream `develop` branch.
- A real recovery rebuilt 22,492 rows and both SQLite `quick_check` and `integrity_check` returned `ok` afterward.

### DIFF
--- a/mempalace/repair.py
+++ b/mempalace/repair.py
@@ -1480,13 +1480,20 @@ def rebuild_from_sqlite(
             else:
                 print(f"    done: {upserted} rows in {cname}")
 
-        print(f"\n  Rebuild complete. {sum(counts.values())} total rows.")
-        if archive_path is not None:
-            print(f"  Original palace archived at: {archive_path}")
-        print(f"{'=' * 55}\n")
-        return counts
     finally:
         backend.close()
+
+    # Bulk Chroma upserts can leave the derived FTS5 index internally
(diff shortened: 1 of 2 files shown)


mvalentsev (commented):
My mistake: I should not have reviewed a draft, and the point I raised was already covered in this thread. Sorry for the noise.

Issue #1573, opened 2026-05-21
HNSW quarantine never re-arms after in-place palace writes — _quarantined_paths gate skips mtime-change reconnects


C-LaForest:
## Summary

The HNSW segment quarantine (`quarantine_stale_hnsw` / `quarantine_invalid_hnsw_metadata`) runs once per palace per process, gated by the class-level set `ChromaBackend._quarantined_paths`. After an external in-place write drifts the index, neither `mempalace_reconnect` nor the `_get_client()` auto-reconnect re-runs quarantine — so new drift is never isolated and wing-scoped searches fail with chromadb `Internal error: Error finding id` until a full process restart.

## Detail

`ChromaBackend._prepare_palace_for_open()` gates quarantine on `_quarantined_paths` membership ("fires once per palace per process").

`_get_client()` detects palace change via inode AND mtime, and rebuilds the client on either — but it only `discard()`s the palace from `_quarantined_paths` on `inode_changed`. A full `repair` replaces `chroma.sqlite3` (new inode) so quarantine re-arms. But `closet_llm`, `mine`, and `compress` write in place — same inode, mtime bumped only. That reconnect path rebuilds the chromadb client but does NOT clear the gate, so `make_client()` -> `_prepare_palace_for_open()` skips quarantine.

`tool_reconnect()` clears `_client_cache`, the chromadb `SharedSystemClient` cache, and the KG cache — but never touches `_quarantined_paths` either.

Net: the long-running MCP server quarantines exactly once, at first open. Every drift after that (closet regeneration is the common trigger) is permanent for the life of the process. `mempalace repair` or a container restart are the only remedies; a restart severs live MCP client sessions (the SSE transport reconnects but the MCP `initialize` handshake does not, so subsequent tool calls return `-32602`).

## Reproduce

1. Long-running `mempalace.mcp_server` against a palace.
2. External `closet_llm` / `mine` / `compress` writes (in place, same chroma.sqlite3 inode) that drift an HNSW segment.
3. Wait past quarantine staleness threshold (300s).
4. Wing-scoped search whose ANN path crosses the drifted segment -> `Internal error: Error finding id`.
5. `mempalace_reconnect` -> still fails. Only a process restart clears it.

## Suggested fix

Re-arm the gate whenever the palace is detected changed on disk, not only on inode change:

- in `_get_client()`, `_quarantined_paths.discard(palace_path)` on `mtime_changed` / `mtime_appeared` as well as `inode_changed`;
[...]

PR #2026, opened 2026-07-15
fix(mcp_server): reset chromadb System cache on staleness reconnect (#2002)

### LINKED ISSUE #2002: _get_client() reconnect leaves the stale Chroma segment live: a peer writer's persist destroys already-indexed records (index count goes backwards)
## Summary

`_get_client()` correctly detects that a peer process wrote the palace (the inode/mtime guard) and rebuilds the Chroma client. But rebuilding the client is not sufficient: chromadb caches its `System`, and therefore the live in-memory `PersistentLocalHnswSegment`, keyed by path. `make_client()` therefore hands back **the same stale segment**.

That segment keeps its outdated view of the index. On its next `_persist()` it writes that view over the on-disk index, silently destroying records other writers had already indexed. The persisted element count goes **backwards**.

This sits under the epic in #1963, and is a distinct, independently fixable defect from the `sync_threshold` / WAL durability problem in chroma-core/chroma#6975. It does not require a threshold crossing to lose data: it loses records that were already safely on disk.

## Reproduction

Two processes, which is the normal mempalace deployment (a long-lived writer plus a per-session MCP server):

1. P1 (long-lived) files one drawer, then holds its cached collection.
2. P2 (peer) files 4 drawers and persists. On-disk index is now at 4.
3. P1 files one more drawer, crossing its batch threshold, so it persists.

Observed, `hnsw:sync_threshold=2`, chromadb 1.5.x rust bindings, mempalace 3.5.0:

```
before P2:          sqlite=1  hnsw=0   divergence=1
after  P2 writes 4: sqlite=5  hnsw=4   divergence=1
after  P1 persists: sqlite=6  hnsw=3   divergence=3   <-- index count went 4 -> 3
```

P2's indexed records are gone from the index. They remain in sqlite, so the palace reports a growing divergence, vector reads get disabled by the capacity guard, and search silently degrades to BM25.

In our production palace this cost roughly 63% of new drawers: over one day, sqlite grew by 107 drawers while the index grew by 40.

## Root cause

`mempalace/mcp_server.py`, in `_get_client()`:

```python
if _client_cache is None or inode_changed or mtime_changed:
    _refresh_vector_disabled_flag()
    if inode_changed or mtime_changed:
        ChromaBackend._quarantined_paths.discard(_config.palace_path)
    _client_cache = ChromaBackend.make_client(_config.palace_path)   # same System, same segment
```

`SharedSystemClient` caches the `System` by path identifier, so the stale segment survives the client rebuild.

## Fix

`_force_chroma_cache_reset()` already does exactly the right thing: it closes the cached client, drops the backend client cache, and calls `SharedSystemClient.clear_system_cache()`. It simply is not called on this path. `repair.py` calls `clear_system_cache()` in two places, so the trap is already known elsewhere in the codebase.

```diff
     if inode_changed or mtime_changed:
         ChromaBackend._quarantined_paths.discard(_config.palace_path)
+        # Rebuilding the client is not enough: chromadb caches its System, and
+        # therefore the live in-memory HNSW segment, keyed by path. make_client()
+        # would hand back the SAME stale segment, which on its next _persist()
+        # overwrites the on-disk index and destroys records other writers had
+        # already indexed.
+        _force_chroma_cache_reset()
     _client_cache = ChromaBackend.make_client(_config.palace_path)
```

With the fix, the same harness ends at `sqlite=6 hnsw=6 divergence=0`. Nothing is lost.

## Verification

- The reproduction above fails before the change and passes after.
- Existing suite: 125 passed, 0 new failures (2 unrelated failures in `test_antigravity_hooks_shell.py` reproduce on the unpatched tree).
- Deployed to a live ~2.2k-drawer palace: a concurrent-writer control (worker plus a peer client, 9 drawers interleaved) indexed 8 immediately with the 9th flushing on the next write, versus the previous behaviour where the peer's writes were destroyed.

Happy to raise this as a PR with a regression test if useful. Note that this does not remove the need for the single-writer direction in #1963 and #1976; it removes one concrete way that data is lost today.

colorpanda82:
Fixes #2002. Part of the #1963 concurrent-writer epic; independent of the #1976 daemon/bridge work.

Thanks @fatkobra for the go-ahead on a focused PR.

## Problem

`_get_client()` detects a peer writer's inode/mtime change on `chroma.sqlite3` and rebuilds the client via `ChromaBackend.make_client()`. But chromadb caches its `System`, and therefore the live in-memory `PersistentLocalHnswSegment`, keyed by path. So the rebuilt "fresh" client is handed back the same stale segment, which on its next `_persist()` writes its outdated index over the on-disk one, destroying records other writers had already indexed.

Observed on a live multi-writer palace: the persisted HNSW count went backwards (4 to 3). Over one day, sqlite grew by ~107 drawers while the index grew by ~40, so roughly 63% of new drawers never reached the index and vector search silently degraded to BM25.

## Fix

Call the existing `_force_chroma_cache_reset()` on the staleness path, before `make_client()`, so chromadb rebuilds the segment from the on-disk state. It is guarded by the existing `inode_changed or mtime_changed` check, so first-open is unaffected.

`_force_chroma_cache_reset()` already closes the client, drops the backend's per-palace cache, and calls `SharedSystemClient.clear_system_cache()`. It simply was not being called on this path (it is already used in the missing-DB and repair paths). `clear_system_cache()` is the accepted workaround for this class of path-keyed staleness in chromadb (cf. chroma-core/chroma#2536, chroma-core/chroma#5843).

## Test

`test_get_client_resets_chroma_system_cache_on_reconnect` asserts the reset runs before `make_client()` on an mtime reconnect. It fails on `develop` (order is `['make_client']`) and passes with the fix (`['reset', 'make_client']`). The existing `test_get_client_rearms_quarantine_on_reconnect` (#1573) and the inode/mtime invalidation tests still pass.

## Scope

This removes one concrete data-loss path on the staleness reconnect. It does not make multi-writer safe on its own; chromadb `PersistentClient` is single-process by design, which is what #1976 / #1270 address at the architecture level.

## Manual multi-process reproduction

With a long-lived writer plus a genuinely separate process:

1. long-lived writer indexes 3 records,
2. separate process indexes 2,
3. long-lived writer indexes 1.

Without the fix the persisted index count regresses; with the fix it ends at 6 (monotonic). Happy to add this as a `@pytest.mark.slow` integration test if you'd prefer it in-tree; I kept it out of the suite to keep CI deterministic.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -998,6 +998,13 @@ def _get_client():
         _refresh_vector_disabled_flag()
         if inode_changed or mtime_changed:
             ChromaBackend._quarantined_paths.discard(_config.palace_path)
+            # #2002: a peer process changed chroma.sqlite3 on disk. chromadb
+            # caches its System (and the live HNSW segment) keyed by path, so
+            # make_client() below would hand back the STALE segment, which then
+            # persists its outdated index over the peer's writes, driving the
+            # persisted count backwards. Drop chromadb's shared cache first so
+            # make_client() rebuilds the segment from the on-disk state.
+            _force_chroma_cache_reset()
         _client_cache = ChromaBackend.make_client(_config.palace_path)
         _collection_cache = None
         _collection_cache_backend = None
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -4176,6 +4176,51 @@ class TestStructuredErrors:
             "_get_client should call _prepare_palace_for_open on reconnect"
         )
 
+    def test_get_client_resets_chroma_system_cache_on_reconnect(
+        self, monkeypatch, config, palace_path, kg
+    ):
+        """``_get_client`` must clear chromadb's path-keyed System/HNSW cache
+        (via ``_force_chroma_cache_reset``) *before* calling ``make_client`` on an
+        inode/mtime reconnect. Otherwise chromadb hands back the stale in-memory
+        HNSW segment, which persists its outdated index over a peer writer's
+        on-disk changes, driving the persisted count backwards (#2002)."""
+        _patch_mcp_server(monkeypatch, config, kg)
+        from mempalace import mcp_server
+        from mempalace.backends.chroma import ChromaBackend
+
+        _client, _col = _get_collection(palace_path, create=True)
+        del _client
+
+        # Prime the cache.
+        mcp_server._get_collection()
+
+        # Simulate a peer writer touching chroma.sqlite3 on disk.
+        old_mtime = mcp_server._palace_db_mtime
+        monkeypatch.setattr(mcp_server, "_palace_db_mtime", old_mtime - 10.0)
+
+        order: list[str] = []
+        real_reset = mcp_server._force_chroma_cache_reset
+        real_make = ChromaBackend.make_client
+
+        def spy_reset():
+            order.append("reset")
+            real_reset()
+
+        @staticmethod
+        def spy_make(path):
+            order.append("make_client")
+            return real_make(path)
+
+        monkeypatch.setattr(mcp_server, "_force_chroma_cache_reset", spy_reset)
+        monkeypatch.setattr(ChromaBackend, "make_client", spy_make)
+
+        mcp_server._get_client()
+
+        assert order == ["reset", "make_client"], (
+            "_get_client must reset chromadb's system cache BEFORE reopening the "
+            "client on a staleness reconnect (#2002)"
+        )
+
     def test_call_kg_retries_after_concurrent_close(self, monkeypatch):
         """A KG closed mid-handler must trigger a one-shot retry with a fresh
         instance — not surface a -32000 to the MCP client."""
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -2707,12 +2707,16 @@ class TestDeleteBySource:
     def test_dry_run_reports_closet_match_count(self, monkeypatch, config, palace_path, kg):
         """Dry run surfaces the closet blast radius (#1722) without deleting."""
         self._seed(monkeypatch, config, palace_path, kg)
-        closets_col = self._seed_closets(palace_path)
+        self._seed_closets(palace_path)
         from mempalace.mcp_server import tool_delete_by_source
+        from mempalace.palace import get_closets_collection
 
         result = tool_delete_by_source("results_mempal_hybrid_v4_session_1.jsonl")
         assert result["dry_run"] is True
         assert result["closet_match_count"] == 2
+        # Re-acquire: the staleness reconnect drops chromadb's path-keyed System
+        # cache (#2002), so a handle taken before the call is dead by now.
+        closets_col = get_closets_collection(palace_path, create=False)
         # Nothing removed — all three closets still present.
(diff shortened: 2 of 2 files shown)


fatkobra (approved):
Reviewed the reconnect path, `_force_chroma_cache_reset()`, and the new regression test.

The fix is correctly placed: the full Chroma system-cache reset runs only after an inode/mtime change is detected and before `ChromaBackend.make_client()` reconstructs the client. First-open behavior is unchanged.

The test is also meaningful: it uses a real palace, primes the client cache, simulates a peer-write mtime change, and verifies the required `reset -> make_client` ordering.

This is a focused and appropriate fix for #2002. It does not claim to make multi-writer Chroma safe generally, and it complements the daemon/single-writer work in #1976 rather than duplicating it.

Approved from my side.

Non-blocking notes:

- A future test could assert the internal `SharedSystemClient.clear_system_cache()` call directly.
- `ChromaBackend._client()` has its own inode/mtime reopen path; it may deserve a separate investigation for long-lived non-MCP callers rather than expanding this PR.
- GitHub flags hidden/bidirectional text in the touched files; the added hunk looks normal, but a control-character scan would be useful.

colorpanda82:
Following up on the three non-blocking notes.

**Hidden/bidirectional text.** Ran a byte-level scan over the full diff. No bidi controls (U+202A-202E, U+2066-2069), no zero-width or BOM (U+200B-200D, U+2060, U+FEFF), and no other invisible/format or non-ASCII control characters, in the added lines or the surrounding context. The only non-ASCII codepoint in the whole patch is a single U+2014 em dash in a test docstring (`tests/test_mcp_server.py`), which is neither hidden nor bidirectional, so the flag does not correspond to anything in this patch. Glad to swap that em dash for ASCII if you'd prefer the touched file stay pure ASCII.

**Direct assertion of the cache reset.** Kept the test at the behavioral `reset -> make_client` ordering level to avoid coupling to chromadb internals, but I agree an explicit assertion on `SharedSystemClient.clear_system_cache()` is a reasonable strengthening. Happy to add it here, or fold it into the follow-up test mentioned below, whichever you prefer.

**`ChromaBackend._client()` reopen path.** Confirmed on `develop`: `_client()` (`chroma.py` L2044-2064) has its own inode/mtime reopen that reconstructs `PersistentClient` without a system-cache reset (`_force_chroma_cache_reset` does not appear in `chroma.py`), so long-lived non-MCP callers holding a backend across a rebuild would be exposed to the same #2002 class. Filed #2028 to track it separately rather than expand this PR.

fatkobra:
Thanks for the thorough follow-up.

This addresses all three of my non-blocking notes:

- The byte-level Unicode scan is sufficient. The `U+2014` em dash is normal visible punctuation, so there is no need to replace it.
- I agree that the current behavioral `reset -> make_client` ordering test is the right level for this PR. Since `_force_chroma_cache_reset()` already exists and this change adds it to the missing reconnect path, a direct assertion against Chroma’s internal `SharedSystemClient` would add coupling without being necessary for this regression.
- Filing #2028 separately for `ChromaBackend._client()` is exactly the right scope. That path has a wider backend-level blast radius and deserves its own caller analysis, reproduction, and tests rather than expanding #2026.

No further changes requested from my side. My approval stands.

PR #2031, opened 2026-07-15
feat(repair): add `reconcile` mode — re-index only the HNSW-missing drawers

KeilerHirsch:
## What

Adds `mempalace repair --mode reconcile`: a surgical alternative to a full `--mode from-sqlite` rebuild for the common **small flush-lag divergence** — the case where the HNSW metadata pickle lags `chroma.sqlite3` by a few hundred drawers (a crash-truncated pickle, or the un-flushed tail after a killed mine), so `hnsw_capacity_status` disables vector search even though the index binary is otherwise sound.

Instead of re-embedding the whole palace (`from-sqlite` re-embeds all N drawers — minutes on a large palace), reconcile diffs the sqlite embedding IDs against the HNSW `id_to_label` keys and re-upserts **only** the missing drawers through the same locked `ChromaCollection.upsert` primitive the live MCP write path uses.

## How

- **`backends/chroma.py`** — new `_hnsw_id_to_label_keys(palace, seg_id) -> set[str]` exposes the indexed-ID set from `index_metadata.pickle` (via the existing allowlist unpickler); `_hnsw_element_count()` becomes a thin wrapper over it (behaviour + existing tests unchanged).
- **`repair.py`**
  - `_sqlite_embedding_ids()` — the IDs that *should* be indexed (collection-scoped, no `embedding_metadata` join so 0-metadata drawers aren't dropped). Returns `None` (not an empty set) on a missing DB or a sqlite lock/error, with a `busy_timeout`, so a held mine can never masquerade as "nothing missing".
  - `compute_missing_ids()` = `sqlite_ids − hnsw_id_to_label_keys`.
  - `reconcile_index()` — preflight `sqlite_integrity`, compute the gap, fetch just those rows (`extract_via_sqlite(only_ids=...)`, chunked `IN`), upsert in batches, re-probe, and report honestly whether the pickle re-flushed (never claims a closed divergence on a legacy palace whose large `hnsw:sync_threshold` hasn't triggered a persist).
  - `ReconcileNotSafe` + caps (**5000 absolute / 2% relative**): refuse a large gap **before** `get_collection` opens the live segment, and point at `from-sqlite`. Reconcile opens the live vector segment (unlike from-sqlite, which reads sqlite directly), so a #1222-scale gap must not be handed to a native resize on open.
  - `extract_via_sqlite()` gains an optional chunked `only_ids` path (the full-scan path is byte-for-byte unchanged).
- **`cli.py`** — `repair --mode reconcile [--dry-run]` dispatch.

## Safety / scope

- Non-destructive: additive upsert only, never delete-and-recreate; `chroma.sqlite3` is untouched as the source of truth. Propagates `MineAlreadyRunning` (clean retry) rather than racing a mine on the HNSW segment.
- **Reconcile is for a flush-lag gap.** A genuinely corrupt / undersized HNSW (the #1222 / #1308 class, where `count`/`upsert` fault natively) can still crash on open even at a *small* divergence that clears the caps — an uncatchable native access violation the size caps cannot predict. That is a `from-sqlite --archive-existing` case (reads sqlite, never opens the failing segment), and the docstring says so. Found empirically against a real ~292k palace whose 0.16% divergence hid exactly this: `--dry-run` detected the gap cleanly, but the live segment access-violated on `count()`.

## Tests

24 new tests in `tests/test_repair_reconcile.py`:

- **unit** (synthetic sqlite + pickle, no chromadb open): `_hnsw_id_to_label_keys` (keys / missing / tampered-pickle), `_sqlite_embedding_ids`, `compute_missing_ids` (gap / no-gap / None-on-absent-pickle / None-on-unreadable-sqlite / negative-divergence), `extract_via_sqlite` only_ids (filter / chunk-over-limit / empty / full-scan-unchanged).
- **orchestration** (mocked backend): gap → upserts exactly the missing ids; batching; no-op; bails when pickle absent; cap trips (absolute + fraction, before opening); dry-run; `MineAlreadyRunning` propagates.

`test_hnsw_capacity.py` + `test_repair.py` (126) still green after the `_hnsw_element_count` refactor; `test_cli.py -k repair` (19) green; ruff clean.

## Validation at scale

`repair --mode reconcile --dry-run` on a 292,052-drawer palace detected the exact **472-drawer gap (0.16%)**, and the fraction cap correctly did **not** trip.

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -643,16 +643,16 @@ class _SafePersistentDataUnpickler:
             return _Restricted(f).load()
 
 
-def _hnsw_element_count(palace_path: str, segment_id: str) -> Optional[int]:
-    """Return the element count chromadb thinks the HNSW segment holds.
+def _hnsw_id_to_label_keys(palace_path: str, segment_id: str) -> Optional[set[str]]:
+    """Return the set of embedding IDs the HNSW segment's pickle knows about.
 
     Reads ``index_metadata.pickle`` via a tight-allowlist unpickler and
-    counts ``id_to_label`` entries. This is the count chromadb consults
-    when sizing/loading the HNSW index on next open — distinct from
-    hnswlib's internal ``cur_element_count`` in the binary files. For
-    #1222's divergence check this is the number that matters, because it
-    is what gets compared against ``count() * resize_factor`` when
-    chromadb decides whether to resize HNSW on load.
+    returns the KEYS of its ``id_to_label`` map — the chroma embedding IDs
+    (string UUIDs) currently present in the flushed HNSW index. This is the
+    authoritative "what is indexed" set: :func:`_hnsw_element_count` counts
+    it for #1222's divergence check, and :mod:`mempalace.repair`'s reconcile
+    mode diffs it against the sqlite embedding IDs to find exactly which
+    drawers are missing from the index.
 
     Uses :class:`_SafePersistentDataUnpickler` rather than chromadb's own
     ``PersistentData.load_from_file`` so the probe works even when
@@ -680,11 +680,29 @@ def _hnsw_element_count(palace_path: str, segment_id: str) -> Optional[int]:
         else:
             id_to_label = getattr(pd, "id_to_label", None)
         if isinstance(id_to_label, dict):
-            return len(id_to_label)
+            return set(id_to_label.keys())
         return None
     except Exception:
-        logger.debug("_hnsw_element_count failed for %s", pickle_path, exc_info=True)
+        logger.debug("_hnsw_id_to_label_keys failed for %s", pickle_path, exc_info=True)
         return None
+
+
+def _hnsw_element_count(palace_path: str, segment_id: str) -> Optional[int]:
+    """Return the element count chromadb thinks the HNSW segment holds.
+
+    Thin wrapper over :func:`_hnsw_id_to_label_keys` — counts the
+    ``id_to_label`` entries in ``index_metadata.pickle``. This is the count
+    chromadb consults when sizing/loading the HNSW index on next open —
+    distinct from hnswlib's internal ``cur_element_count`` in the binary
+    files. For #1222's divergence check this is the number that matters,
+    because it is what gets compared against ``count() * resize_factor``
+    when chromadb decides whether to resize HNSW on load.
+
+    Returns ``None`` when the pickle is absent (fresh / never-flushed
+    segment) or unreadable. Callers treat ``None`` as "unknown".
+    """
+    keys = _hnsw_id_to_label_keys(palace_path, segment_id)
+    return len(keys) if keys is not None else None
 
 
 # Divergence threshold: chromadb's HNSW flushes asynchronously, so HNSW
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -2202,6 +2202,26 @@ def cmd_repair_status(args):
     repair_status(palace_path=palace_path)
 
 
+def _run_reconcile_mode(palace_path: str, collection_name: str, dry_run: bool) -> None:
+    """Dispatch ``repair --mode reconcile``: surgical HNSW re-index of only the
+    drawers missing from the index, with clean non-zero exits for the
+    too-large-gap and mine-in-progress cases."""
+    from .palace import MineAlreadyRunning
+    from .repair import ReconcileNotSafe, reconcile_index
+
+    try:
+        reconcile_index(palace_path, collection_name, dry_run=dry_run)
+    except ReconcileNotSafe as exc:
+        print(f"\n  {exc}")
+        sys.exit(1)
+    except MineAlreadyRunning as exc:
+        print(
+            f"\n  {exc}\n  A mine is holding the palace; "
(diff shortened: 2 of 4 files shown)


fatkobra (changes requested):
Blocking: reconcile does not hold one palace writer lease across the operation. The missing-ID snapshot and SQLite extraction happen before any lease is acquired, and locking occurs separately inside each `col.upsert()` batch. A normal writer can therefore change SQLite/HNSW between the diff, extraction, individual batches, and post-verification.

The reconcile set can then describe an older generation than the state being modified: its safety cap no longer applies to the current state, already-repaired IDs can be processed again, and stale extracted rows can overwrite a newer concurrent update.

Please hold the same per-palace writer lease for the complete reconcile transaction: integrity preflight/diff -> SQLite extraction -> every upsert -> final verification. Nested collection writes should use the existing re-entrant behavior or an explicit already-locked primitive. Add an interleaving-writer regression at each phase boundary.

KeilerHirsch:
Addressed in f7e2c23 — thanks for the precise failure analysis, it was exactly right.

**The lease.** `reconcile_index` now takes ONE `mine_palace_lock(palace_path)` for the complete transaction — integrity preflight, missing-ID diff, SQLite extraction, every upsert batch, and the final verification — and delegates the unchanged body to `_reconcile_under_lease`. Nothing runs outside the lease anymore, so the reconcile set always describes the generation being modified. Nested `ChromaCollection.upsert` acquisitions pass through the lock's existing process-wide re-entrancy (#1859); no new locking primitive was introduced. A foreign holder now raises `MineAlreadyRunning` at entry, before any read — `_run_reconcile_mode` in cli.py already maps that to a clean non-zero exit, so the CLI contract is unchanged (it just fires earlier and more reliably now).

**Interleaving-writer regressions at each phase boundary**, as requested. The foreign writer is simulated with an independent file handle on the lock file — Windows byte-range locks and POSIX `flock` conflict per-handle even within one process, so no subprocess is needed and the tests run identically on both CI platforms:

- `test_reconcile_holds_one_writer_lease_across_all_phases` — at preflight, diff, extract, every upsert, and verify, a foreign writer's non-blocking acquire must FAIL, and must succeed again once the transaction ends (lease released).
- `test_reconcile_nested_real_write_lock_reenters_under_the_lease` — runs the REAL `ChromaCollection._write_lock` → `mine_palace_lock` nesting under the outer lease (only the innermost chromadb collection is mocked): re-entrant pass-through, no self-deadlock, foreign writer still refused during the upsert.
- `test_reconcile_refuses_at_entry_when_a_writer_holds_the_palace` — with a foreign writer holding the palace, reconcile refuses before the preflight ever reads.
- `test_reconcile_dry_run_diff_runs_under_the_lease` — the dry-run report also comes from a lease-consistent snapshot.
- `test_reconcile_releases_lease_when_a_phase_raises` — a mid-transaction failure never strands the lease (non-vacuous: asserts the lock file exists and is free).

TDD order: the boundary tests were written first and failed against the previous commit (no lease held at any phase; entry not refused), then went green with the fix. Full `test_repair_reconcile.py` (28 tests) + `test_repair.py` + `test_palace_locks.py`: 138/138 passing, ruff clean.

fatkobra (approved):
Thanks — f7e2c23 resolves my blocking concurrency concern.

`reconcile_index` now acquires the palace writer lease before the first palace read and retains that same lease across the integrity preflight, missing-ID snapshot, SQLite extraction, every upsert batch, and final verification. That closes the stale-generation/interleaving window from my original review.

The regression coverage also addresses the requested cases directly: it probes the real OS lock at each phase boundary, exercises the real `ChromaCollection._write_lock` → `mine_palace_lock` nesting, verifies refusal before preflight when another writer owns the palace, covers dry-run consistency, and confirms that exceptions release the lease.

I am satisfied that the original issue is resolved and am approving the implementation.

One mechanical item remains before merge: the current lint job reports that `ruff format --check .` fails. Please apply the formatter and get that check green.

igorls:

KeilerHirsch:
Definitely still interested in landing this — I’ll rebase it onto current develop and get CI green.

I’ve been temporarily buried in a couple of other projects and forensic rabbit holes that turned out to be at least as important as MemPalace 😅. Apparently I’ve developed a habit of finding measurement and provenance problems in other projects too.

So MemPalace work has been paused, not abandoned. 3.8.0 sounds good — thanks for picking this back up!

KeilerHirsch:

One non-mechanical conflict was worth calling out: current `develop` includes the later sparse-metadata recovery fix that drives `extract_via_sqlite()` from `embeddings` with a `LEFT JOIN`. The old reconcile helper would have reintroduced the zero-`embedding_metadata` row loss when it factored extraction into the `only_ids` path. I reproduced that regression first, then changed the shared helper to preserve sparse rows for both full-scan and reconcile extraction and added a dedicated `only_ids` regression.

- `tests/test_repair_reconcile.py` + `tests/test_palace_locks.py`: 49 passed
- focused `test_repair.py` extraction/HNSW set: 7 passed
- `test_cli.py -k repair`: 26 passed, 1 skipped

GitHub now reports the PR mergeable again. Pre-rebase head is preserved on my fork as `archive/2026-08-29/pr2031-pre-rebase`.

Issue #1835, opened 2026-06-20
perf(qdrant): server-side facet aggregation for O(1) wing/room stats


fatkobra:
**What happened?**

Follow-up to #1796 and the direction set by #1664.

#1796's fix (single-scroll get_all_metadata()) makes Qdrant's bulk metadata
read O(n) instead of O(n²), which is correct but still linear in collection
size — every call to mempalace_status / mempalace_list_wings /
mempalace_get_taxonomy walks every drawer's metadata to build wing/room
counts client-side.

Qdrant's REST API exposes a [facet endpoint](https://qdrant.tech/documentation/concepts/facet/)
that computes per-field value counts server-side without returning the
underlying points — true O(1) (from the client's perspective) stats for
exactly the metadata.wing / metadata.room aggregation tool_status,
tool_list_wings, and tool_get_taxonomy need.

**Proposed shape**

- New BaseBackend capability token, e.g. supports_metadata_facets.

- New optional BaseCollection method, e.g. facet_counts(field: str, where: Optional[dict] = None) -> dict[str, int],
with no default implementation (raises UnsupportedCapabilityError,
matching the existing pattern for lexical_search).

- QdrantCollection implements it via POST /collections/{name}/facet.

- mcp_server.py's wing/room-counting call sites (tool_status, tool_list_wings, tool_list_rooms, tool_get_taxonomy) check for the
capability and use it when available, falling back to
get_all_metadata() + client-side counting otherwise — same fallback
pattern already established by #1796.

**Why a separate issue/PR**

This is a new wire capability and a new backend-contract method, not a fix
to existing broken behavior — it deserves its own design discussion (exact
field-name escaping, multi-field facets if Qdrant supports them, how
where-filtered facets map onto Qdrant's filter syntax) rather than riding
along with #1796's bug fix.

cc @YC-AIUSER  who originally suggested this direction on #1796.

messelink:
[...]

PR #2038, opened 2026-07-17
feat(pgvector): implement facet_counts server-side aggregation (#1868 contract)

messelink:
Implements `BaseCollection.facet_counts(field, where, limit)` on `PgVectorBackend`, completing the pgvector side of the aggregation contract merged in #1868 and specified in #1835.

## Motivation

Direct incident data from a production peer-client deployment:
- 419K-drawer pgvector palace, remote clients over TLS + WAN
- `mempalace status` took ~344s and OOM-killed a 2Gi pod (client-side counting materialized ~150 MB of Python dicts)
- With this change: status completes in <1s, wire transfer drops to <100 KB

Same class of pain motivated #1824 originally.

## Approach

Follows the rebase path @igorls proposed in [#1824's CHANGES_REQUESTED review](https://github.com/MemPalace/mempalace/pull/1824#pullrequestreview-4588183373):

1. Method on `PgVectorBackend`: `facet_counts(field, where=None, limit=None) -> dict[str, int]`
2. SQL: `SELECT COALESCE(metadata->>%s, 'unknown') AS k, COUNT(*) FROM <table> WHERE <where_sql> GROUP BY 1 [LIMIT %s]`
   - `COALESCE` matches the `m.get(path, "unknown")` semantics of the existing client-side fallback (byte-for-byte parity)
   - Field bound as parameter; table quoted via `_quote_identifier` for injection safety
   - `where` mapped through the same predicate translator pgvector uses elsewhere
3. `supports_metadata_facets` registered on `PgVectorBackend` so the four MCP call sites (`tool_status`, `tool_list_wings`, `tool_list_rooms`, `tool_get_taxonomy`) take the fast path automatically
4. No `tool_status` / `mcp_server.py` edits — those call sites already switched to `facet_counts` in #1868
5. Tests: 6 new unit tests (empty palace, single wing, multi-wing, `where` filter passthrough, `limit`, NULL bucketing). `EmbeddingCollection` already forwards `facet_counts` since #1898 merged.

## Attribution

- Original problem statement and initial SQL: @gavinmcfall in #1824 (2026-06-18). This PR reuses the SQL shape they got right; the delta is API alignment (`count_by_metadata_field` → `facet_counts`), `supports_metadata_facets` capability registration, and unit tests.
- #1824 has been silent since @igorls's CHANGES_REQUESTED review on 2026-06-28 (30 days). Filing this so the perf work doesn't continue to bit-rot.
- Contract from #1835 (@fatkobra) and #1868 (@Prayaksh).
- Wrapper forwarder from #1898 (merged 2026-07-06) means this drops in without additional plumbing.

Closes #1824 if maintainers agree.

### DIFF
--- a/mempalace/backends/pgvector.py
+++ b/mempalace/backends/pgvector.py
@@ -54,6 +54,7 @@ from .base import (
     PalaceNotFoundError,
     PalaceRef,
     QueryResult,
+    UnsupportedCapabilityError,
     UnsupportedFilterError,
     _IncludeSpec,
 )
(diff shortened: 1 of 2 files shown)


igorls (approved):

PR #2040, opened 2026-07-17
feat(kg): add mempalace_kg_wing_links — area links from shared entities

jonkafton:
## What does this PR do?

Adds a read-only MCP tool, **`mempalace_kg_wing_links`**, that derives *area-to-area* links from the knowledge graph: two wings are linked when they **share (or are related through) the same entity**. This is a stronger signal than the existing `find_tunnels`, which links wings that merely have a room of the same *name* (a specific shared name is meaningful; a generic one like "Notes" is often coincidence). It powers a "Knowledge-graph links" view in the 1Presence 3D memory map.

**How it works**
- New `KnowledgeGraph.current_triples_with_source()` — returns `(subject, object, source_drawer_id)` for every *current* fact that carries drawer provenance. A `source_drawer_id` resolves to a wing (the drawer's `wing` metadata), so each entity can be *located* in one or more wings. Facts without provenance (older, pre-`source_drawer_id`) are skipped.
- `tool_kg_wing_links()` builds `entity → {wings}` from that join, then weights each wing-pair by **specificity**: `1 / log2(1 + n_wings)`, so an entity that appears in many wings (a hub like a person's own name) contributes little — it links everything, so it means little — while an entity shared by just two wings counts full. Each link is labelled with its most-specific shared entity. Returns the top-weighted links.

**Why specificity weighting matters:** without it, a ubiquitous entity would link nearly every wing to every other — the exact hairball this is meant to avoid.

## How to test

```
uv run pytest tests/test_knowledge_graph_extra.py -v
```

Adds tests for the provenance filter (facts without `source_drawer_id` are excluded) and expired-fact exclusion (`valid_to IS NULL` only). I also verified the full join + specificity math against a live in-memory `KnowledgeGraph`: a 2-wing entity outranks a 3-wing entity's pairs (`1/log2(3) = 0.63 > 1/log2(4) = 0.50`), confirming the down-weighting. `ruff check` passes on all three files.

## Checklist
- [x] Tests pass (`tests/test_knowledge_graph_extra.py`; the tool's collection scan needs `chromadb`, which CI provides)
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .` — clean on changed files)

## Notes
- **Read-only** (not in `_MUTATING_TOOLS`); it scans the collection metadata + KG triples, no writes.
- Coverage depends on `source_drawer_id` provenance — facts predating that column don't participate. On a palace with sparse provenance the link set will be small; that's honest (better to draw few real links than many fabricated ones).

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ Usage and tool reference:
 
 ## MCP server
 
-36 MCP tools cover palace reads/writes, knowledge-graph operations,
+37 MCP tools cover palace reads/writes, knowledge-graph operations,
 cross-wing navigation, drawer management, and agent diaries. Installation
 and the full tool list:
 [mempalaceofficial.com/reference/mcp-tools](https://mempalaceofficial.com/reference/mcp-tools.html).
--- a/mempalace/README.md
+++ b/mempalace/README.md
@@ -16,7 +16,7 @@ The Python package that powers MemPalace. All modules, all logic.
 | `dialect.py` | AAAK compression — entity codes, emotion markers, 30x lossless ratio |
 | `knowledge_graph.py` | Temporal entity-relationship graph — SQLite, time-filtered queries, fact invalidation |
 | `palace_graph.py` | Room-based navigation graph — BFS traversal, tunnel detection across wings |
-| `mcp_server.py` | MCP server — 34 tools, AAAK auto-teach, Palace Protocol, agent diary |
+| `mcp_server.py` | MCP server — 37 tools, AAAK auto-teach, Palace Protocol, agent diary |
 | `onboarding.py` | Guided first-run setup — asks about people/projects, generates AAAK bootstrap + wing config |
 | `entity_registry.py` | Entity code registry — maps names to AAAK codes, handles ambiguous names |
 | `entity_detector.py` | Auto-detect people and projects from file content |
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -653,6 +653,20 @@ class KnowledgeGraph:
             "relationship_types": predicates,
         }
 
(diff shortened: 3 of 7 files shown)


mvalentsev (commented):
Heads up on the tool count in the docs: this branch takes `len(TOOLS)` to 37, but README.md still says "36 MCP tools" (website/reference/mcp-tools.md also says 36, and mempalace/README.md says 34, which was already behind before this PR).

Nothing in CI will flag it. `TestToolCount` scans the README for `(\d+)\s+tools`, and that pattern doesn't match the phrase "36 MCP tools", so it loops over an empty list of claims and passes. A green run isn't evidence the counts line up.

jonkafton:
Good catch, thanks @mvalentsev — fixed in 4494146.

- **Counts bumped to 37:** `README.md`, `website/reference/mcp-tools.md`, and the package `mempalace/README.md` module table (that one was already stale at 34 pre-PR, now 37).
- **Documented the tool:** added the `### \`mempalace_kg_wing_links\`` section to `website/reference/mcp-tools.md` — it had none. That was also a live `TestNoUnlistedTools` failure waiting to happen (code_tools − doc_tools = the new tool); it just hadn't surfaced because no checks have reported on this fork commit yet.
- **Fixed the false green you flagged:** `TestToolCount` now matches the `"<n> MCP tools"` phrasing (`(\d+)\s+(?:MCP\s+)?tools`), scans the MCP tools reference alongside the README, and asserts at least one count claim was actually found — so a phrasing the regex misses can't loop over zero claims and pass vacuously again.

Verified locally: `TestToolCount`, `TestReadmeToolsExistInCode`, and `TestNoUnlistedTools` all pass, and `ruff check` is clean on the test file.

igorls:

Issue #1796, opened 2026-06-13
qdrant backend: status / mempalace_status is O(n^2) on large palaces — get(limit, offset) re-scrolls the whole collection per page


YC-AIUSER:
### Environment
- mempalace **3.4.0** (pip), backend = **`qdrant`** (Qdrant **1.18.2**, local `http://localhost:6333`)
- Windows 11, Python 3.12 · palace ≈ 74.5k drawers (`mempalace_drawers`), ≈ 4.9k (`mempalace_closets`)

### Summary
On the `qdrant` backend, both `mempalace_status` (MCP) and `mempalace status` (CLI) hang / time out (>120 s) on a large palace, while `mempalace search` returns in ~1 s and writes succeed. This is **distinct from** the chroma-centric #1748 / #802 / #1664: those route status through the sqlite aggregation path, which non-chroma backends lack — as the #1764 review itself notes ("breaks status for non-ChromaDB backends … they lack `chroma.sqlite3`"). There is currently **no working fast-status path for qdrant**.

### Root cause
Status enumerates all drawer metadata via paginated `col.get(limit, offset)`:
- MCP: `mcp_server._fetch_all_metadata` loops `col.get(include=["metadatas"], limit=1000, offset=offset)` → ~75 pages for 74k.
- CLI: `miner.status` loops `col.get(limit=5000, offset=offset)` → ~15 pages.

On qdrant, `QdrantCollection.get(limit, offset)` has **no native offset**: it calls `_scroll_all()`, which scrolls the **entire** collection (native cursor, `limit=256`/page) then slices `[offset:offset+limit]` in Python. So **every** paginated `get()` re-scrolls the whole collection → O(n²) (MCP ≈ 75 × full_scroll(74k) ≈ 5.5M point reads → >120 s). `search` is unaffected because it uses `query()` (vector index), not `get()`.

### Evidence
A single native cursor scroll (`POST /collections/{c}/points/scroll`, following `next_page_offset`, payload-only, `limit=4096`) over all 74,464 points completes in **~1.4 s** (~55k/s). Qdrant itself is fast; the cost is purely the adapter emulating integer offset by re-scrolling.

### Why existing fixes don't apply
- #802's fix (paginate `col.get`) targeted chroma's "too many SQL variables"; on qdrant that same pagination is what triggers the O(n²).
- #1748 / #1664 / #1764 route status through **sqlite** aggregation, unavailable for qdrant / pgvector / sqlite_exact (per #1764 review).

[...]

Issue #1657, opened 2026-05-29
perf: O(1) read path for large palaces — SQL-aggregate stats, SQLite retrieval module, bounded source-file fetch


trek-e:
**What problem does this solve?**

Read-path operations are O(N) in palace size and become very slow on large palaces. On a ~400K-drawer palace:

- `mempalace status` takes **minutes**. `miner.status()` paginates the entire collection in 5,000-row batches and aggregates wing/room counts in Python (`miner.py:1907-1937`). The MCP `tool_status` / `list_wings` / `list_rooms` / `get_taxonomy` path does the same via `_fetch_all_metadata` (`mcp_server.py:615-652`), materializing hundreds of MB of Python dicts.
- A correct SQL `GROUP BY` aggregation **already exists** (`_tool_status_via_sqlite`, `mcp_server.py:665-734`) but is only used when `vector_disabled=True` — the fast path runs only when the index is broken.
- Per query, closet-boost enrichment calls `drawers_col.get(where={"source_file": ...})` with **no `limit`** (`searcher.py:936-984`) and counts drawers via a full `get()` just to call `len()` (`searcher.py:246-249`) — an unbounded O(M) materialization that scales with the largest source file.
- The BM25/SQLite fallback (`searcher.py:398-634`, 237 lines) reaches directly into ChromaDB's private schema and is mislayered inside the searcher, making the production-scale retrieval path hard to test in isolation.

**What's the proposed solution?**

Three cohesive read-path deepenings (architecture review, performance-at-scale focus):

1. **Metadata aggregation seam.** Promote a single palace-stats module backed by `GROUP BY` on `embedding_metadata` as the *primary* path for all count/taxonomy reads (CLI `status`, `tool_status`, `list_wings`, `list_rooms`, `get_taxonomy`). The Python full-scan becomes dead code. Minutes → milliseconds; the large Python allocation disappears.
2. **Verbatim-over-SQLite retrieval module.** Extract the BM25/SQLite fallback into its own adapter that owns all direct `chroma.sqlite3` schema knowledge behind a small `query → ranked drawer ids` interface, so the searcher orchestrates two adapters (ANN-vector + SQLite-exact) and the production-scale fallback is independently testable.
3. **Bounded source-file drawer access.** Replace the unbounded `get(where=source_file)` calls with a small module: count via SQL `COUNT(*)`, fetch with explicit pagination/limit.

[...]

PR #2043, opened 2026-07-18
Fix O(collection) client-side scans and missing payload indexes on the qdrant network backend

### LINKED ISSUE #2042: Shared qdrant network backend doesn't scale: O(collection) client-side scans hang; required payload indexes never created
## Summary

Setting up a **shared palace across multiple machines** on the qdrant network backend (via `mempalace serve` / `MEMPALACE_BACKEND=qdrant` with a shared namespace) surfaces several operations that assume cheap local ChromaDB reads and instead perform **O(collection) client-side scans** or rely on **payload indexes that are never created**. On a small local palace these are instant; on a shared collection of ~184k drawers they take minutes or fail outright, which makes routine operations (mine, status, auto-save hooks) hang.

Environment: 6 machines sharing one qdrant collection (~184k drawers), all on minilm (384-dim), mempalace 3.6.0.

### 1. Metadata payload indexes are never created on qdrant
`QdrantBackend._ensure_remote_collection()` only creates a payload index on the document field (`_PAYLOAD_DOCUMENT`, `"text"`). It never indexes `metadata.wing`, `metadata.room`, or `metadata.source_file`. Consequences:
- `facet_counts()` on `wing`/`room` fails with **HTTP 400 `No appropriate index for faceting: metadata.wing`**. This breaks the MCP `mempalace_status` tool's fast path (it calls `col.facet_counts("wing"/"room")`) — it errors and falls back to a full scan.
- Metadata-filtered reads (`where={"source_file": ...}` in `file_already_mined` / `_source_file_delete_ids`) scan the whole collection instead of using an index.

**Fix:** create keyword payload indexes on `metadata.{wing,room,source_file}` when ensuring the collection (idempotently, so existing collections get them too).

### 2. `hallways.compute_hallways_for_wing()` scans the entire collection on every mine
It paginates **all** drawers via `col.get(limit, offset, include=["metadatas"])` and filters to the target wing **client-side** (the code comments note this is deliberate). Runs after every `mine`. On a large shared collection this pins a core for minutes per mine, so every `mempalace mine` (and every auto-save hook) hangs.

**Fix:** fetch with a server-side `where={"wing": wing}` filter (bounded to the wing) instead of scanning + client-side filtering.

### 3. `miner.status()` (CLI) scans all metadata client-side
The CLI `status` path paginates all drawers and tallies wing/room in Python. Hangs on a large network backend. (The MCP `mempalace_status` tool already uses `facet_counts` — the CLI should too.)

**Fix:** use `facet_counts` when the backend advertises `supports_metadata_facets`, else fall back to the scan.

### 4. `palace.prefetch_mined_set()` scans the entire collection on every convo-mode mine
`mine_convos` calls `prefetch_mined_set(collection)` once per run to build a `{source_file: mtime}` map for skip decisions — via a full paginated metadata scan. Auto-save hooks run convo-mode mining, so **every hook fire triggers a multi-minute full scan** of the shared collection (and it grows worse over time).

**Fix:** answer the skip check per-file via the indexed `where={"source_file": ...}` lookups that `file_already_mined` already uses (scoped to the files actually being mined), instead of one unbounded scan. A lazy set that preserves the existing `in` / `[]` / `.get` interface keeps the call sites unchanged.

### 5. `qdrant_backend.json` marker is only written on the first write
The marker is written inside `add()`/`upsert()`. Until then, CLI `search`/`status` on a freshly-configured qdrant palace report *"Palace dir exists but has no qdrant_backend.json yet — run mine"*, even though the collection is fully populated and reachable. A read-only client (or a host that hasn't written yet) can't search a perfectly good shared palace.

**Fix:** write the marker when the collection is opened/ensured, not only on first write.

## Impact

Together these make the advertised shared / team network-backend setup impractical beyond a small collection: mine, auto-save hooks, and `status` hang, facet status errors, and freshly-pointed clients refuse to read. The common root cause is operations designed around cheap local ChromaDB reads that don't account for a large remote-backed collection.

I have working fixes for all five and will open a PR referencing this issue.

jhsmith409:
## What & why

Fixes #2042. Setting up a **shared palace across multiple machines** on the qdrant network backend surfaces several operations that assume cheap local ChromaDB reads and instead do **O(collection) client-side scans**, or rely on **payload indexes that are never created**. On a shared collection of ~184k drawers these hang (mine, `status`, auto-save hooks) or fail (facet `status` 400s), which makes the network-backend / team setup impractical at scale.

## Changes

All new paths are **gated on the backend's `supports_metadata_facets` capability** and fall back to the existing local-optimized path, so **ChromaDB behavior is unchanged**.

- **qdrant payload indexes** — `_ensure_remote_collection` now creates keyword indexes on `metadata.wing`, `metadata.room`, `metadata.source_file` (idempotent via the existing 400/409-swallowing `create_payload_index`, and backfills already-existing collections). Fixes `facet_counts()` `HTTP 400 "No appropriate index for faceting"` (which the MCP `mempalace_status` tool hits) and lets metadata-filtered reads use an index.
- **qdrant marker on open** — write `qdrant_backend.json` when a populated collection is opened for write, not only on first upsert, so a read-first client can `search`/`status` a shared palace it hasn't written to yet.
- **`miner.status`** — count wing/room via `facet_counts` on facet-capable backends instead of streaming every drawer's metadata (mirrors the MCP status tool).
- **`palace.prefetch_mined_set`** — on facet-capable backends, resolve skip checks with lazy, cached, indexed per-`source_file` lookups scoped to the files being mined, instead of one full-collection scan that hangs every convo-mode mine (auto-save hooks).
- **`backends/base.py`** — the facet-capability check is hoisted into a single shared `collection_supports_facets()` helper rather than inline copies at each call site.

### Dropped: `hallways.compute_hallways_for_wing`

This branch originally also scoped the hallways drawer fetch to the wing. **That change is no longer part of this PR** — #2477 landed on `develop` first and fixed the same O(collection) scan, and its fix is strictly better:

| | this branch (dropped) | #2477 on develop (kept) |
|---|---|---|
| qdrant/pgvector/milvus | server-side `where={"wing": wing}` | same |
| chroma | full client-side scan, to dodge the `SQLITE_MAX_VARIABLE_NUMBER` overflow on >32k wings (#1619) | also scoped server-side — bounded pages keep the bind count under the limit |

So chroma gets the scoped query too, instead of being excluded from the fix. On merging `develop` I took its version wholesale; the `collection_supports_facets` import and `server_filter` local in `hallways.py` went with it. Nothing else in this PR changed as a result.

## Testing

- Full merged tree: **317 passed / 1 skipped** across `test_hallways*`, `test_cli_hallways`, `test_backends`, `test_backend_conformance`, `test_qdrant_backend`, `test_qdrant_bulk_metadata_scroll`, `test_miner`; **406 passed** across `tests/mcp/`.
- `ruff check mempalace/` clean; `ruff format --check` clean.
- **Validated in production**: this exact logic has been running as runtime patches across a 6-machine fleet sharing one ~184k-drawer qdrant collection. Before: `status` and convo-mode mine hung for minutes; facet `status` 400'd. After: `status` ~0.5s, project mine ~1.2s, convo-mine ~0.5s, facets fast.

## Merge note

`develop` split `mempalace/mcp_server.py` into the `mempalace/mcp_server/` package while this branch was open, so this branch's edits to that module were relocated: the `backends` import to `mcp_server/__init__.py`, the `_supports_metadata_facets` delegation to `mcp_server/_session.py`. No behavior change — the pre-split function body was identical to develop's.

https://claude.ai/code/session_01H32fWQrjdZWPNhfY1gzCgH


jhsmith409:
For reviewers — where this fits with the existing server-side aggregation work:

- The `facet_counts` contract landed in #1868, and #2038 completes it for **pgvector**. This PR fills the **qdrant** side of the same gap: qdrant collections were never getting the `metadata.{wing,room,source_file}` keyword payload indexes that `facet_counts` (and metadata-filtered reads) require, so the MCP `mempalace_status` facet path 400s and CLI `miner.status` still full-scans.
- It therefore also contributes to #1657 (O(1) read path) on the qdrant side — status/taxonomy no longer stream the whole collection.
- Draft #1935 overlaps the incremental-mining metadata angle; happy to coordinate if that lands first.

All new fast paths are gated on `supports_metadata_facets` with a fallback to the existing local path, so ChromaDB is unchanged.

jhsmith409:
Rebased onto current `develop` (`aa89bd8`, post-#2051) — clean, no conflicts, and the diff is unchanged at 183+/18− across the same four files. The only upstream movement in files this PR touches was `miner.py` (#2054 / #2055 `chunk_text` changes), which doesn't overlap the `status` facet path changed here.

Could a maintainer approve the workflow run? No checks have ever reported on this branch — as a fork PR the Actions run needs approval, so `Tests` has never executed, and the `unstable` merge state reflects that rather than any failure. Happy to rebase again or split this into smaller pieces if that would make it easier to review.

jhsmith409:
Rebased onto current `develop` (`1d113e4`, post-#2148) — the branch had gone stale and was showing `CONFLICTING`. It's `MERGEABLE` again at `0a6bdad`.

**Conflict resolution (one hunk, `miner.status()`)** — upstream added the #1222 HNSW-divergence preflight and moved `total = col.count()` down into the scan block; this branch had moved `total` up and inserted the facet fast path at the same spot. Kept both, ordered:

1. HNSW divergence preflight (unchanged from `develop` — it has to stay ahead of `count()`, since that's the call it guards against the SIGSEGV class)
2. `total = col.count()`
3. facet fast path, gated on `supports_metadata_facets`
4. paginated client-side scan fallback

Net effect on ChromaDB is still nil: the preflight runs first exactly as on `develop`, and the facet branch is skipped.

**Also addressed the helper duplication I flagged in the description.** `_backend_supports_facets` is now a single `collection_supports_facets()` in `backends/base.py`, re-exported from the `backends` package, replacing the three inline copies. `mcp_server._supports_metadata_facets` stays as a module-level name (now a one-line delegate) rather than being renamed — `tests/test_mcp_server.py` monkeypatches it in four places, and I'd rather not pull test churn into this PR. Happy to collapse it fully if you'd prefer.

Also fixed a missing blank line in `hallways.py` that `ruff format --check` would have failed on; squashed into the original commit.

**Verification** — I got a clean dev environment this time, so the suite I couldn't run before now has results:

- `pytest` (full suite): **3607 passed, 31 skipped, 106 deselected**
- `ruff check mempalace/`: clean
- `ruff format --check mempalace/`: clean

Diff is now 189+/25− across 7 files (was 184+/18− across 4), in two commits.

`mergeStateStatus` is still `UNSTABLE`, which continues to reflect that no Actions run has ever been approved on this fork branch rather than any failure — a maintainer approving the workflow would give you the real signal.

jhsmith409:
Rebased onto current `develop` (`517a7a0`) — the branch had gone stale again (205 commits behind, `CONFLICTING`/`DIRTY`). Now `f26f5ba`, still two commits.

**Conflicts (two hunks)**

1. `palace.prefetch_mined_set` — upstream #2183 replaced the flat `mined` dict with per-`source_file`/per-`source_mtime` `groups` and a `chunk_total` completeness rule. Kept upstream's scan verbatim and put the facet-capable early return ahead of it, so the fallback path is byte-identical to `develop`.
2. `mcp_server` imports — `develop` added `date_window` to the same import block this branch touched for `collection_supports_facets`. Kept both.

**Follow-on fix while resolving:** `_LazyMinedSet._lookup` (the facet-backend path) predated #2183 and would have reported a mid-file partial as fully mined — exactly the stranding bug #2183 fixed, just on qdrant/pgvector instead of Chroma. It now applies the same rule as the bulk scan: group by stored mtime, count against `chunk_total`, treat a drawer with no `chunk_total` as a legacy row and trust it. Without this the rebase would have re-introduced the bug on the network backends.

**Verification** (rebased branch vs. `develop` at the same commit, clean venvs, same machine):

| | branch `f26f5ba` | `develop` `517a7a0` |
|---|---|---|
| `pytest` | 4310 passed, 4 failed, 31 skipped | 4310 passed, 4 failed, 31 skipped |
| `ruff check mempalace/` | clean | — |
| `ruff format --check mempalace/` | clean (79 files) | — |

The 4 failures are all in `tests/test_palace_graph_limits.py` (`TestFindTunnelRanking`, `TestGraphStatsDetails`) and reproduce identically on unmodified `develop` — pre-existing, unrelated to this PR, which touches no graph code.

Diff is 206+/25− across 7 files. Still no Actions run has ever executed on this fork branch — a maintainer approving the workflow would give you real CI signal.

jhsmith409:
Any chance for a review and incorporation into your next release?  I've been running this on my cluster of 5 machines for over a month now with no issues.  Significantly speeds up use of MemPalace when used across more than one machine.

jhsmith409:
Rebased onto current `develop` (post-3.8.0). No conflicts, and `git patch-id` of the net change is identical before and after the rebase, so nothing was dropped or altered in the move — only the base changed.

**Still integrates cleanly with what 3.8.0 changed underneath it:**

- `mcp_server._supports_metadata_facets` now delegates to the hoisted `backends.base.collection_supports_facets` rather than duplicating the capability probe. Kept as a module-level name so the call sites stay monkeypatchable per-module; all nine call sites are unchanged.
- `_LazyMinedSet` mirrors the `chunk_total` completeness rule from #2183 — a source whose only current-mtime group is short of its `chunk_total` is reported as not mined, so the caller re-mines instead of stranding the remaining chunks.
- In `miner.status`, the facet path sits after the sqlite fast path and the HNSW divergence preflight, so Chroma keeps everything #2314/#2308 added and only non-sqlite backends take the facet route.

**Verification:** full suite green on the rebased branch — 4468 passed, 31 skipped, 0 failures.

**Measured on a live shared Qdrant palace** (252k drawers, 86 wings, minilm):

| build | `mempalace status` |
|---|---|
| 3.8.0 as released | still scanning when cut off at 45s |
| 3.8.0 + this PR | **0.54s** |

For context on why this matters beyond one command: the same O(collection) pattern is on `prefetch_mined_set`, which every convo-mode mine hits, so on a shared network backend the auto-save hooks stall for minutes per invocation. We've been carrying these three changes as out-of-tree monkeypatches against the installed package on a six-host fleet since 3.6.0, re-applying them after every upgrade because `uv tool upgrade` silently reverts them and the only symptom is that the host goes slow again. Happy to split this into per-file PRs if that's easier to review.

jhsmith409:
Re-tested against **v3.9.0** on a live 6-host shared-qdrant cluster (261,889 drawers, 95 wings, ~262k). Short version: this branch still applies cleanly, and the three scans it removes are all still present on `develop` today.

### No rebase needed this time

Rebased onto current `develop` (`e809834`, the 3.9.0 release merge) locally to check: **clean, no conflicts**, and `git patch-id --stable` is byte-identical before and after (`701238554f2e…`) — so a force-push would change nothing. Net diff is unchanged at 206+/25− across the same 7 files. GitHub agrees (`mergeable: true`). I've left the branch as-is rather than pushing a no-op rebase.

I also checked this against what 3.9.0 changed underneath it:

- **#2384 (bound qdrant scroll)** does not overlap. It caps `_scroll_all` where a limit is known; none of the three paths below pass one, so they still scan the whole collection.
- **#2379 (CLI→hub search forward)** is orthogonal — different read path.
- `mcp_server._supports_metadata_facets` still delegates to the hoisted `backends.base.collection_supports_facets`, as reconciled in the 3.8.0 rebase. All call sites unchanged.

### Still unfixed on `develop` as of 3.9.0

- `palace.prefetch_mined_set` — full paginated scan, `collection.get(limit=1000, offset=…)` over every drawer
- `hallways.compute_hallways_for_wing` — scans the entire collection and filters to the wing **client-side**
- `miner.status` — no facet path; falls through to the client-side metadata scan on any non-chroma backend

### Measured: stock 3.9.0 vs this branch, same host, same live collection

| operation | stock 3.9.0 | this PR | speedup |
|---|---|---|---|
| `mempalace status` | **113.9 s** | **0.58 s** | ~196× |
| `compute_hallways_for_wing` (small wing) | **108.6 s** | **0.01 s** | ~10,000× |
| `compute_hallways_for_wing` (larger wing) | **108.8 s** | **0.28 s** | ~390× |
| `prefetch_mined_set` + 5 lookups | **539.8 s** | **0.011 s** | ~49,000× |

Both are completed runs, not timeouts. Identical output in every case — same drawer/wing totals from `status`, same hallway results.

The hallways number is the one I'd flag: stock costs ~109 s **per wing regardless of wing size**, because each call rescans the whole collection. At 95 wings that's ~2.9 hours for a full rebuild, which is why hallway maintenance is effectively unusable on a shared backend today. This branch makes it proportional to the wing.

`prefetch_mined_set` is the one that hurts most in practice: it's on the auto-save/mine hook path, so on a shared backend every conversation save pays a full-collection scan. A single stock call takes **9 minutes** (539.8 s) on this collection; the branch answers in 0.011 s (lazy + indexed lookups, mirroring the #2183 `chunk_total` completeness rule so the fallback semantics are preserved).

### Ask

Any chance of a review for the next release? This has been running on 6 machines against one shared qdrant palace since mid-July with no issues — through 3.6.0, 3.7.1, 3.8.0 and now 3.9.0 — and I've rebased it five times to keep it current. Without it, a multi-host palace is essentially unusable past ~100k drawers: `status` and every conversation save stall for minutes with no error, which is a rough first experience for anyone trying the shared/team backends.

No checks have ever reported on this branch — as a fork PR the Actions run needs maintainer approval, so `Tests` has never executed, and the `unstable` merge state reflects that rather than any failure. Happy to split it into smaller PRs (payload indexes / status / prefetch / hallways are independent) if that would make it easier to land.

jhsmith409:
Rebased onto current `develop` (`d9f0590`) — the branch was 104 commits behind. **No conflicts**, despite the churn in `mcp_server.py` since (the lightweight MCP/PQL server, #2412). Now at `e5c16a5`, still two commits, diff unchanged at 206+/25− across the same 7 files.

**Still needed after #2384.** That PR bounds `_scroll_all()` by `offset + limit`, which makes small-page `get()` calls cheap, but `develop`'s `qdrant.py` still creates only the `document` text index — there is no keyword payload index on `metadata.{wing,room,source_file}`, so `facet_counts()` continues to raise "No appropriate index for faceting" and metadata-filtered reads still full-scan. That gap is what this PR closes; the two changes are complementary.

**Verification** (clean venv, same machine, this branch vs. `develop` at `d9f0590`):

| | `e5c16a5` | `develop` |
|---|---|---|
| `pytest tests/ --ignore=tests/benchmarks` | 4821 passed, 31 skipped | 4821 passed, 31 skipped |
| `ruff check .` | clean | clean |
| `ruff format --check .` | clean | clean |

**Could a maintainer approve the workflow run?** This is worth a moment of context, because the PR now shows a red X that means the opposite of what it looks like:

- Runs `30963714213` (Docker) and `30963714215` (Tests) were queued on `0a6bdad` on 2026-08-05 and sat at `action_required`. Nobody approved them. At 2026-09-04T00:36:57Z — 30 days and two minutes after they were created — both flipped to `failure` **in the same second, with zero jobs and zero check-runs**. Nothing was ever checked out or built; that is the pending-approval retention window expiring, not a test failure.
- The runs on the previous head `f20f093` (queued 2026-08-24) are on the same clock and would have expired the same way around 2026-09-23.
- The push above queued fresh runs (`33825821685`, `33825821737`), which are again waiting at `action_required`.

So since this PR was opened on 2026-07-18, not one job has executed on it. `UNSTABLE` here has always meant "no checks approved", never "checks failed". I'm happy to keep rebasing, and equally happy to split this into smaller pieces — the payload indexes, the `hallways`/`miner` fast paths, and the read-first marker fix are independent and could land separately if that's easier to review.

jhsmith409:
Maintainers: if you're not planning to pick this PR up, please let me know so I can stop keeping it current with `develop`. Thanks!

---

Merged the latest `develop` in (c61d22f) and resolved the conflicts:

- **`mempalace/palace.py` → package split (#2494):** moved `_LazyMinedSet` and the facet-backed early return in `prefetch_mined_set` into `palace/mined.py`, and imported `collection_supports_facets` in `palace/__init__.py`. The lazy lookup still applies the same `chunk_total` completeness rule (#2183) as the bulk scan.
- **`miner.status()`:** both sides added a fast path. The order is now: facet counts (qdrant/pgvector/milvus), then `get_all_metadata()` (#1796), then the paginated scan. `col.count()` is only called on the facet and paginated paths, so `test_does_not_call_count` still passes.

Tests: ruff is clean and the non-benchmark suite passes locally. `tests/benchmarks/test_mcp_light_bench.py::test_schema_reduction_benchmark` fails (68.6% vs a 70% threshold), but it fails the same way on a clean `develop` checkout, so this PR didn't cause it.

PR #2048, opened 2026-07-20
feat(wal): opt-in verbatim payloads so dropped writes stay recoverable

molanov:
## Problem

The WAL replaces every content-bearing param with `[REDACTED N chars]`:

```python
safe_params[k] = f"[REDACTED {len(v)} chars]" if isinstance(v, str) else "[REDACTED]"
```

Sensible as a default. But it removes the WAL's ability to answer the one question it is uniquely placed to answer after a storage-layer fault: **what were we trying to write?**

This is not hypothetical. On a self-hosted palace, a ChromaDB `max_seq_id` watermark desync made `col.add()` report success and persist nothing for **2.5 months** (2026-04-30 → 2026-07-19). `automatically_purge` empties `embeddings_queue`; `seq_id` is `INTEGER PRIMARY KEY` without `AUTOINCREMENT`, so rowids restart at 1, while the watermark still held `1229782938920431664` — every new record looked already-applied and was dropped before reaching the queue. Reads and search kept working the whole time, so nothing looked broken, and `mempalace_diary_write` returned `{"success": true, "entry_id": ...}` throughout.

When it was finally caught, the WAL held a record of all **98** lost calls (81 `diary_write` + 17 `add_drawer`) and the content of **none** of them. Every preview was the literal string `[REDACTED 200 chars]`. The writes were logged and still unrecoverable.

3.6.0's schema-strict dispatcher fixed the *other* half of that incident (callers sending `content`/`title` being silently stripped to an opaque `-32000`). This is the remaining half: making the log useful when storage lies.

## Change

`wal_store_full_payload` (env `MEMPALACE_WAL_STORE_FULL_PAYLOAD`), default `false`. When enabled, redact keys are written verbatim.

## Why a flag rather than always storing payloads

The WAL is already created `0o600` inside a `0o700` directory, so file permissions are not what redaction defends against — it defends against an operator attaching the log to a bug report or support dump. That is a real concern for shared and team palaces, especially now that `mempalace serve` exists. So the conservative default stays and single-user self-hosted installs opt in.

Rejected alternatives:
- *Always store, rely on `0o600`* — simpler, but silently changes what every existing install writes to disk, including team servers.
- *Encrypt with a local-only key* — key management outweighs the benefit for the single-user case this targets.

## Implementation notes

- **Cached per process.** `MempalaceConfig` reads `config.json` in its constructor and the WAL is on the hot path of every write; a per-call read would put disk I/O inside the hook latency budget.
- **Fails closed.** Any error resolving the setting keeps redaction rather than writing memory text to disk because a lookup broke.
- **Lazy import.** `MempalaceConfig` is imported inside the function to preserve this module's no-import-side-effects contract (`test_wal_import_has_no_mcp_server_side_effect`).

## Backward compatibility

Flag off (the default) is byte-identical to current behaviour — `test_wal_redacts_when_flag_disabled` pins that, and the pre-existing redaction tests are untouched and still pass.

## Tests

Five new tests in `tests/test_wal.py`, including the end-to-end regression you'd want here: a real `tool_diary_write` call, then read the WAL back and assert the payload is not `[REDACTED*]`. It covers `entry_preview` specifically — the key that was redacted during the incident window.

Both behavioural tests are non-vacuous — with the guard reverted they fail, and the end-to-end one reproduces the exact symptom:

```
E       AssertionError: ['[REDACTED 55 chars]']
```

Full suite: 3361 passed, 31 skipped. `ruff check` and `ruff format --check` clean.

## Docs

`website/guide/configuration.md` gains a "Write-ahead log payloads" section explaining the tradeoff and when to enable, plus the env-var table row. Warns that the setting must be on *before* it is needed — the WAL cannot retroactively un-redact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Features
 
-- **Opt-in verbatim write-ahead log payloads.** `wal_store_full_payload` (or `MEMPALACE_WAL_STORE_FULL_PAYLOAD`) records write payloads in the WAL instead of replacing them with `[REDACTED N chars]`, so a write lost to a storage-layer fault can be reconstructed from the log. Off by default and fails closed, leaving existing behaviour unchanged; intended for self-hosted single-user palaces, where the WAL is already `0o600` inside a `0o700` directory.
+- **Opt-in verbatim write-ahead log payloads.** `wal_store_full_payload` (or `MEMPALACE_WAL_STORE_FULL_PAYLOAD`) records write payloads in the WAL instead of replacing them with `[REDACTED N chars]`, so a write lost to a storage-layer fault can be reconstructed from the log. Enabling it also lifts the 200-character preview cap that `add_drawer`, `delete_drawer`, `update_drawer`, and `diary_write` applied before logging — otherwise a multi-kilobyte entry would still be recoverable only as its first 200 characters. Off by default and fails closed, leaving existing behaviour byte-identical; intended for self-hosted single-user palaces, where the WAL is already `0o600` inside a `0o700` directory.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Features
+
+- **Opt-in verbatim write-ahead log payloads.** `wal_store_full_payload` (or `MEMPALACE_WAL_STORE_FULL_PAYLOAD`) records write payloads in the WAL instead of replacing them with `[REDACTED N chars]`, so a write lost to a storage-layer fault can be reconstructed from the log. Off by default and fails closed, leaving existing behaviour unchanged; intended for self-hosted single-user palaces, where the WAL is already `0o600` inside a `0o700` directory.
+
 ---
 
 ## [3.6.0] — 2026-07-14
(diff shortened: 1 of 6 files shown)


molanov:
Pushed a follow-up commit (`0971544`) after enabling this on a live palace and reading a real entry back out of the WAL.

**The flag alone did not make writes recoverable.** Every WAL call site passed `value[:200]` before `_wal_log` ever saw it:

```python
"entry_preview": entry[:200],
```

Redaction was hiding a second limit underneath itself — a redacted 200-char slice and a redacted 20 KB entry are the same string in the log, so the cap was invisible. With payloads stored it becomes the binding limit, and recovering the first 200 characters of a multi-kilobyte diary entry is not recovering it. Real entries run to several kilobytes, so as first written this would have logged roughly a tenth of a lost write while reporting the write as captured.

The four WAL previews (`add_drawer`, `delete_drawer`, `update_drawer`, `diary_write`) now route through `wal_payload()`, which returns the value whole when full payloads are enabled and the historical `value[:200]` otherwise. Flag off is still byte-identical: truncate, then redact.

Two more tests, both non-vacuous. Suite now 3363 passed / 31 skipped, ruff clean.

Verified end-to-end on a live palace — `entry_preview` in `write_log.jsonl` now holds the exact entry text, byte-for-byte, where it previously held `[REDACTED N chars]`.

fatkobra (changes requested):
Blocking: storing the complete text does not yet make every WAL entry faithfully recoverable.

The `add_drawer` WAL record omits `source_file`, and the `diary_write` record omits the resolved destination `wing`. Replaying those records can therefore lose source provenance/synchronization behavior or place a diary entry in a different wing from the original request.

The current end-to-end tests establish that the text is no longer truncated or redacted, but they do not establish semantic replay equivalence.

Please log the complete validated request after defaults have been resolved. At minimum, this includes:

- `source_file` for `add_drawer`; and
- the resolved `wing` for `diary_write`.

Please also audit the remaining recoverable operations for other optional/defaulted fields that are currently omitted.

Add round-trip regressions that reconstruct each write from its WAL entry and verify equivalent content, destination, and provenance—not only that the full text appears in the log.

Issue #1593, opened 2026-05-23
miner.py: stop delete-on-remine — verbatim history should never be destroyed by MemPalace miners


milla-jovovich:
So obviously it's a bit awkward to create an issue on a system I conceived. Saying that, Mempalace was more concept than code to begin with, so thank you to the core group of people who've contributed the most from the get go, making it an accessible, working infrastructure with the ability to support all the different systems it would need to run on to be usable. 

@igorls has been my saving grace by maintaining the site while simultaneously engineering it to be more platform agnostic which opened the "palace doors" to people. 

He also made it possible for me to focus on what I'm good at which is architecture and vision, taking the time I need to render blueprints of a concept and truly make it a reality through code. It was then that I was able to create my first major commit "context-aware corpus detection" which is the foundation for what I have been building and what you all will be seeing in the next few weeks unfold into a concrete structure with rules that make it possible for MemPalace to evolve organically with users, storing and retrieving the information that will create and populate the Wings and Rooms of your individual Palaces, breathing life into them in ways that may surprise you.

The problem is, while I've been working on deepening the complexity of how the mechanics need to work in tandem and write the rules the structure needs to follow for all this to happen in the most intuitive way possible, I've ironically missed one of the literal cornerstones of what MemPalace is built on. VERBATIM HISTORY STORAGE. This is literally what started my journey on current memory storage/retrieval options and Ai summaries that take our ideas and turn them into shells without nuance, stripping them of personality and potential ideas that might have seemed unimportant at the time, mistakes even, that we later regret not having record of. That was the birth of MemPalace and through my never ending audits of each new version that been released, I didn't realize until today that this core principle was lost amidst the madness of infrastructure development. The fault is mine of course, so basically, I'm raising an issue that I myself created because, well honestly, I haven't been seeing the forest for the trees. And I deeply apologize for that. So without further ado via Claude's technical observation, here is my issue:

### The architectural violation

MemPalace's design law is that **nothing is ever lost** unless the user explicitly deletes it. From the project's stated purpose:

> MemPalace is an AI memory storage and retrieval system. Not an editing tool.  A memory tool whose main objective is verbatim storage. This is how legacies are sustained.

The current miner violates this. In `mempalace/miner.py:1014`:

```python
# Converts modified-file re-mines from upsert-over-existing-IDs (which hits
# the chromadb 0.6.3 update bug) into a clean delete+insert, bypassing the
# update path entirely.
try:
    collection.delete(where={"source_file": source_file})
[...]

PR #2050, opened 2026-07-20
fix: prevent duplicate drawers when re-mining LLM conversations

### LINKED ISSUE #2044: Duplicate drawers get created when mining LLM conversations from the same source (even if only some conversations are actually new)
**What happened?**
in an attempt to update, ideally regularly, my set of conversations and drawers from LLMs such as ChatGPT or Claude, I used the mining process on an updated export of Claude conversations as exported from Claude itself. Instead of preventing duplicates and only mining the uniquely new/changed conversations, MemPalace created a duplicate of almost every, if not every, conversation and drawer that it had already mined in the first place in addition to the handful of new ones that were changed. Note that even attempting to use the split command does not prevent this problem. Also note that this problem occurs whether or not MCPs or any other access points are used. Purely on the CLI, this is visible. 

**What did you expect?**
instead, the mining process should be able to clearly identify that certain conversations are the exact same verbatim conversations it had already mined. It should save the time and prevent mining, and thus speed up the process and prevent duplicate mining into the memory palace. Ideally a process should be in place or should be available to automatically mine w/o duplicates the conversations on triggers such as new LLM conversation exports going into a folder etc.

My question overall is: if every time you re-mine exports from Claude and ChatGPT, etc. Mempalace duplicates what you already had, how can that result in a useable system to keep your palace up to date? I.e. does Mempalace have an established/normal process to make sure that the palace keeps itself up to date with your LLM conversations/interactions? A process that would prevent duplicates and yet be up to date (or even be manual but consistent)? 

**How to reproduce:**
1. Export conversations from Claude or ChatGPT by standard processes.
2. Use the command line interface to mine the conversations.
3. Have a few more conversations in Claude or ChatGPT and extract an updated version of your conversations from them.
4. Use the command line interface to mine the conversations again.
5. Notice that the memory palace now contains duplicates of almost every single conversation. (and that the mining process took way longer than would be needed for only a handful of conversations that are new.)

**Environment:**
- OS: MacOS 15.6
- Python version: 3.14.5

rohanprichard:
## Problem

When the same LLM conversation is re-exported from Claude, ChatGPT, or other tools, it often lands under a **new filename** (timestamped export bundle, regenerated slug, etc.) even though the conversation content is unchanged. The existing `mined_mtimes` check only recognizes files as already-mined by their **exact path**, so the same conversation re-exported under a new path was always treated as "new" — creating **duplicate drawers** containing identical verbatim content.

## Fix

Adds **content-hash-based deduplication** to `convo_miner.py`:

1. **`prefetch_content_hashes()`** — new function in `palace.py` that bulk-loads `content_hash → source_file` mappings for all drawers already filed at the current schema version, analogous to the existing `prefetch_mined_set()` 
2. **`_normalize_convo_content()`** — extracted normalization into a helper that also registers empty/short files as filed
3. **`_skip_content_duplicate()`** — checks normalized content SHA-256 against the prefetched hash map; if the same hash already exists under a different `source_file`, registers the duplicate and skips it
4. **`content_hash` stored in metadata** — each drawer now records its content hash, and `_register_file()` accepts an optional `content_hash` for the sentinel
5. **Next-run optimization** — registering the duplicate with the sentinel means subsequent runs skip it via the cheap mtime check instead of re-normalizing and re-hashing

## Changes

| File | What |
|------|------|
| `mempalace/convo_miner.py` | Core dedup logic: SHA-256 hash, `_skip_content_duplicate`, `_normalize_convo_content` helper, hash propagation through `_file_chunks_locked` and `_register_file` |
| `mempalace/palace.py` | New `prefetch_content_hashes()` bulk loader |
| `tests/test_convo_miner.py` | `test_mine_convos_skips_same_content_under_new_filename` — verifies duplicate transcript under new filename produces zero extra drawers |

## Test

```bash
uv run pytest tests/test_convo_miner.py::test_mine_convos_skips_same_content_under_new_filename -v
```

Fixes #2044

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -21,7 +21,7 @@ from typing import Optional
 
 from .collision_scan import assert_no_collisions
 from .ids import ID_RECIPE, make_convo_drawer_id, make_convo_sentinel_id
-from .normalize import normalize
+from .normalize import normalize_conversations
 from .entities import entities_metadata
 from .palace import (
     NORMALIZE_VERSION,
@@ -630,35 +630,30 @@ def _is_ai_tool_path(path: Path) -> bool:
     return False
 
 
-def _skip_content_duplicate(
-    dry_run: bool,
-    collection,
-    filepath: Path,
-    content_hash: str,
+def _split_new_and_duplicate_conversations(
+    conversations: list,
     wing: str,
-    agent: str,
-    extract_mode: str,
+    source_file: str,
     mined_content_hashes: dict,
-    progress: str,
-) -> bool:
-    """Register and report a content-duplicate file, if this content_hash is
-    already filed under a different source_file. Returns True if skipped.
+) -> tuple:
+    """Hash each conversation and split them into (new, duplicate) lists.
 
-    Repeated exports from Claude/ChatGPT commonly land under a new filename
-    each run even when the conversation itself is unchanged, so the
-    source_file-keyed mtime skip never recognizes them. Registering the
-    duplicate here means the next run skips it via the cheap mtime check
-    instead of re-normalizing and re-hashing it.
+    A conversation is a duplicate when its hash is already registered under
+    a *different* source_file in the same wing — mining the same transcript
+    into a second wing is a deliberate re-file, not a repeat, so the lookup
+    is scoped to (wing, hash). Returns ([(hash, text), ...] new, [(hash,
+    dup_source_file), ...] duplicates).
     """
-    if dry_run:
-        return False
-    source_file = str(filepath)
-    dup_source = mined_content_hashes.get(content_hash)
-    if dup_source is None or dup_source == source_file:
-        return False
-    _register_file(collection, source_file, wing, agent, extract_mode, content_hash)
-    print(f"  = {progress} {filepath.name[:50]:50} duplicate of {Path(dup_source).name}")
-    return True
+    new_items = []
+    duplicates = []
+    for conversation in conversations:
+        content_hash = hashlib.sha256(conversation.strip().encode("utf-8")).hexdigest()
+        dup_source = mined_content_hashes.get((wing, content_hash))
+        if dup_source is None or dup_source == source_file:
+            new_items.append((content_hash, conversation))
+        else:
+            duplicates.append((content_hash, dup_source))
+    return new_items, duplicates
 
 
 def _is_unchanged_since_last_mine(source_file: str, mined_mtimes: dict) -> bool:
@@ -778,7 +773,7 @@ def _compute_hallways_for_wing_safe(wing, collection, drawers_filed, config=None
         print(f"  (hallways skipped: {exc})")
 
 
-def _normalize_convo_content(
+def _normalize_convo_conversations(
     filepath: Path,
     source_file: str,
     cfg_min_chunk_size: int,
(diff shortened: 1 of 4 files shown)


mvalentsev (commented):
I don't think this closes the case in the issue. The report is about re-mining an updated export: the same conversations plus a few new ones. Claude's privacy export is a single JSON file holding every conversation, and `normalize` joins them with `"\n\n".join(transcripts)`, so the hash here covers the whole bundle rather than a conversation.

I ran it both ways: a 2-conversation export, then a second export with those same two plus one new one under a new filename. develop files 10 drawers, 4 for each old conversation and 2 for the new one. This branch files exactly the same 10, because adding one conversation changed the bundle hash and the skip never fires.

What it does catch is the narrower case where the file is byte-identical and only the name changed.

Hashing per conversation rather than per file would cover the reported case, though that means splitting the transcript before hashing.

One other thing I hit: the hash map isn't scoped to a wing, so mining the same transcript into a second wing skips it there and leaves that wing holding only the `_registry` sentinel.

rohanprichard:
> I don't think this closes the case in the issue. The report is about re-mining an updated export: the same conversations plus a few new ones. Claude's privacy export is a single JSON file holding every conversation, and `normalize` joins them with `"\n\n".join(transcripts)`, so the hash here covers the whole bundle rather than a conversation.
> 
> I ran it both ways: a 2-conversation export, then a second export with those same two plus one new one under a new filename. develop files 10 drawers, 4 for each old conversation and 2 for the new one. This branch files exactly the same 10, because adding one conversation changed the bundle hash and the skip never fires.
> 
> What it does catch is the narrower case where the file is byte-identical and only the name changed.
> 
> Hashing per conversation rather than per file would cover the reported case, though that means splitting the transcript before hashing.
> 
> One other thing I hit: the hash map isn't scoped to a wing, so mining the same transcript into a second wing skips it there and leaves that wing holding only the `_registry` sentinel.

Sorry about that, I misunderstood, will test out and make sure it works for all cases. I'll also add test results here once I push the fix

mvalentsev (commented):
Re-checked the new version and the reported case is fixed. On develop, re-exporting the bundle re-files the old conversations as duplicates. On your branch only the new one gets mined, and the per-wing scoping holds up too.

One heads-up though: this part is already being worked on by @milla-jovovich. #1593 is the design issue and #1628 is her open PR making all four miners additive. #1628 actually defers the "at-ingestion content-hash dedup" and the normalize.py session-splitting to a later follow-up, which is basically what this PR does for the convo miner. It also touches convo_miner.py and normalize.py, and its model keeps every version as a layer instead of skipping. Might be worth flagging your PR on those so it's on her radar, though it's a big track and might not move fast.

fatkobra (changes requested):
Blocking: the revised dedupe still has no durable content-ownership or source-reference model.

Consider this sequence:

1. Mine source A containing conversation X.
2. Mine source B containing the same conversation X.
3. B is skipped as a duplicate and receives only the duplicate/sentinel state.
4. Later, A changes and no longer contains X.
5. Re-mining A purges A’s stored copy of X.
6. B remains unchanged and mtime-current, so it is not reconsidered.

Conversation X now disappears from the palace even though authoritative source B still contains it.

Please make deduplicated content independently content-owned, or retain durable aliases/reference counts for every source that supplies it. One possible design is a deterministic conversation/content ID with separate provenance records for each source, so removing one source cannot delete content still referenced by another.

At minimum, the duplicate metadata must retain per-conversation canonical references, and source deletion/rebuild must promote or re-materialize a surviving alias before deleting the last stored copy.

Please add this regression:

- mine A(X);
- mine B(X);
- change A so it no longer contains X;
- re-mine A;
- verify X remains searchable because B still supplies it, without changing B’s mtime.

PR #2059, opened 2026-07-21
docs: document KG supersede usage

ggettert:
## Summary

- Document `KnowledgeGraph.supersede()` in the knowledge graph and Python/API docs
- Add `mempalace_kg_supersede` to the MCP knowledge graph tool lists
- Show the preferred pattern for replacing a single-valued fact at one temporal boundary

## Testing

- `bun run docs:build`

Part of #2236

### DIFF
--- a/integrations/shared/recall-protocol.md
+++ b/integrations/shared/recall-protocol.md
@@ -41,8 +41,11 @@ question-driven, not reflexive.
 5. **After a substantive session**, record continuity with
    `mempalace_diary_write` (background hooks may already do this — do not
    double-file).
-6. **When a fact changes**, call `mempalace_kg_invalidate` on the old
-   fact, then `mempalace_kg_add` for the new one.
+6. **When a fact changes**, choose the operation that preserves temporal
+   history: use `mempalace_kg_supersede` for single-valued replacements
+   (model, employer, owner, address, current status),
+   `mempalace_kg_invalidate` for facts that ended without replacement,
+   and `mempalace_kg_add` for independent/coexisting facts.
 
 ## Tool selection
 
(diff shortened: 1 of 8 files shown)


integrations/shared/recall-protocol.md
```diff
@@ -41,15 +41,19 @@ question-driven, not reflexive.
 5. **After a substantive session**, record continuity with
    `mempalace_diary_write` (background hooks may already do this — do not
    double-file).
-6. **When a fact changes**, call `mempalace_kg_invalidate` on the old
-   fact, then `mempalace_kg_add` for the new one.
+6. **When a fact changes**, choose the operation that preserves temporal
```
igorls: **[suggestion]** After this change, the canonical recall protocol tells agents to use `mempalace_kg_supersede` for single-valued replacements, matching `PALACE_PROTOCOL` in `mempalace/mcp_server.py` (which forbids hand-rolling invalidate + add because the two facts overlap at the boundary). The sibling file `integrations/shared/coordination-protocol.md` — and its published copy `website/guide/shared-brain.md` — still say "When a fact changes: mempalace_kg_invalidate the old fact, then mempalace_kg_add the new one." That is the exact path the runtime protocol now warns against. The four recall skill copies in this PR stay in sync with each other; this leftover is outside the recall skills but inside `integrations/shared/`.

**Suggestion:** Update the shared-brain memory bullet (and the website copy that quotes it) to the same three-way split: supersede for single-valued replacements, invalidate when a fact ended with no successor, add for independent/coexisting facts. Does not need to block this PR if a follow-up is already planned; it is the remaining contradiction agents will still ingest via Claude.md / shared-brain installs.

PR #2064, opened 2026-07-23
fix(normalize): strip noise blocks with multi-paragraph bodies and indented tags

mazurd-acre:
## Problem

Two escape paths let Claude Code harness chrome leak into mined drawers and pollute search/recall:

1. `<task-notification>` and `<system-reminder>` blocks carrying multi-paragraph payloads (subagent results, recalled memories) were never stripped: the tag-body pattern refused to cross blank lines, so exactly the largest noise blocks survived wholesale.
2. Claude Code emits slash-command chrome with indented tags (`<command-message>`, `<command-args>` after `<command-name>`), which the line-start anchor missed. `command-args` was also missing from the noise list entirely.

Found by mining ~90 real Claude Code sessions: thousands of drawers contained raw task-notification payloads, and semantic recall surfaced harness boilerplate instead of content.

## Fix

- Tag bodies may now span blank lines, guarded by a same-tag-reopen lookahead instead of the blank-line rule: a dangling open tag still cannot merge with a later block and eat the real content between them (the original span-eating protection this rule existed for).
- The line anchor tolerates leading whitespace after the optional `> ` blockquote marker.
- `command-args` added to `_NOISE_TAGS`.

## Testing

- 3 new regression tests (multi-paragraph block, dangling-tag safety, indented slash-command chrome).
- Full test suite green.
- Re-mined the same ~90 real sessions (20k+ drawers): zero noise blocks survive, inline prose mentioning the tags is still preserved.

### DIFF
--- a/mempalace/normalize.py
+++ b/mempalace/normalize.py
@@ -45,6 +45,7 @@ _NOISE_TAGS = (
     "system-reminder",
     "command-message",
     "command-name",
+    "command-args",
     "task-notification",
     "user-prompt-submit-hook",
     "hook_output",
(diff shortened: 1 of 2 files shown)


fatkobra (changes requested):
Blocking: allowing the noise-block body to cross blank lines reintroduces a destructive dangling-tag case.

The negative lookahead prevents an unclosed opener from merging with a later *opening* of the same tag, but it does not prevent that opener from consuming arbitrary transcript content until a later standalone or literal closing tag.

For example, an unclosed `<system-reminder>` followed by legitimate conversation text and a later user/code example containing `</system-reminder>` can cause all intervening content to be removed.

That contradicts the conservative normalization invariant: when pairing is uncertain, original transcript content should survive.

Please bound matching using actual transcript/message boundaries, or replace the cross-paragraph regex with a small parser that strips only demonstrably paired noise blocks.

Please add a regression containing:

1. a dangling opening tag;
2. blank lines;
3. legitimate user/assistant content; and
4. a later closing-tag token without another opening tag.

The legitimate intervening content must remain unchanged.

PR #2068, opened 2026-07-23
feat(mine): route explicit source adapters through the RFC 002 registry (#2062)

### LINKED ISSUE #2062: feat(mine): route explicit source adapters through the RFC 002 registry
## Context

RFC 002's source-adapter specification landed in #990 and its registry/`PalaceContext` scaffold landed in #1014. However, `mempalace mine` still dispatches directly to legacy `--mode` miners. Registered entry-point adapters therefore cannot yet be invoked from the CLI, blocking #1944 and #1484.

Relates to #989.

## Scope

Add `--source <adapter-name>` to `mempalace mine`. When supplied, resolve the adapter through `resolve_adapter_for_source()`, construct the existing `PalaceContext`, and run the adapter ingestion path.

## CLI compatibility

- Existing `mempalace mine <dir>` remains equivalent to `--mode projects`.
- Existing `--mode projects|convos|extract` behavior remains unchanged.
- `--source` and an explicitly supplied `--mode` are mutually exclusive and fail with a clear parser error.
- Unknown or unavailable source adapters fail clearly.

## Acceptance criteria

- A registered fixture adapter can be invoked end-to-end through `mempalace mine <dir> --source fixture`.
- Unknown adapter names produce an actionable error.
- Existing no-flag and `--mode` paths retain their current behavior.
- Targeted tests cover CLI parsing, registry resolution, `PalaceContext` construction, and legacy-mode compatibility.
- CLI help documents `--source` as the RFC 002 extension path.

## Out of scope

- Migrating the legacy filesystem, conversation, or extract miners.
- Adapter-specific configuration/options.
- Changing transform, privacy, incremental-ingest, closet, or routing semantics.
- Any behavior change to existing `--mode` users.

## Follow-on

After this lands, #1944 can add its adapter-specific CLI integration without rebuilding the dispatch layer.

ggettert:
Closes #2062.

## What does this PR do?

RFC 002's source-adapter registry and `PalaceContext` scaffold already exist, but `mempalace mine` still dispatches only to legacy `--mode` miners. This adds the explicit extension path:

```console
mempalace mine <source> --source <adapter-name>
```

The runner resolves the adapter through the registry, constructs `PalaceContext`, and invokes `ingest()`. `--source` and `--mode` are mutually exclusive; omitting both preserves the legacy `projects` default.

### Safety and protocol boundaries

- **Dry runs are inert.** They never open a backend or knowledge graph. Adapters receive empty recording facades, so direct published-surface collection/KG writes are recorded but never persisted.
- The dry-run collection facade preserves RFC 001 reads: `.get()` and `.query()` return typed empty `GetResult` / `QueryResult` values, including the correct query outer dimension.
- **Non-dry-run adapter ingestion holds one palace writer lease** before collection/KG handles open and through the complete iteration and write sequence.
- A competing writer produces a clean nonzero source-adapter CLI exit without a traceback.
- Adapters declaring `supports_incremental` are rejected before ingest until the RFC 002 currentness lifecycle exists. Non-incremental `SourceItemMetadata` is accepted as progress/version information with a warning, avoiding a late partial-ingest failure.
- Unsupported adapter yield types fail explicitly rather than being silently discarded.
- Daemon mine payloads preserve explicit source-adapter dispatch and report `mode: "source"`.

### Scope

This is the #2062 CLI-dispatch foundation. It deliberately does **not** make #1944 or #1484 runnable yet: both adapters declare `supports_incremental`, which requires the still-unimplemented RFC 002 currentness lifecycle.

Out of scope: migrating legacy miners, adapter-specific options, transforms, privacy, incremental-ingestion semantics, closet semantics, and routing semantics.

## How to test

```console
python -m pytest tests/test_cli_source_adapters.py \
  tests/test_daemon.py::test_run_mine_source_adapter_dispatches_through_adapter_runner -q
ruff check .
ruff format --check .
```

Focused source-adapter coverage passes on `b676b2b`; ruff check and formatting pass.

Coverage includes:

- parser validation, registry resolution, `PalaceContext` construction, unknown-adapter errors, and legacy no-flag compatibility;
- dry-run direct collection/KG mutation interception;
- storage invariants for a nonexistent palace, existing empty Chroma directory, and initialized `sqlite_exact` palace (including any WAL/SHM sidecars);
- full-ingest writer leasing and real competing-process CLI contention;
- typed empty dry-run reads;
- non-incremental metadata, unsupported yields, and daemon source-adapter dispatch.

GitHub Actions runs are approval-gated for this fork PR; the focused validation above is the available verification pending approval.

## Checklist

- [x] Focused tests pass
- [x] No hardcoded paths


fatkobra (changes requested):
Blocking: `--dry-run` is not side-effect-free for source adapters.

The loop suppresses only core’s `context.upsert_drawer(result)` call, but it still invokes `adapter.ingest()` with a live `PalaceContext` containing the real drawer collection and knowledge graph.

RFC 002 exposes those objects so adapters can add KG triples, and `PalaceContext` permits adapters to bypass `upsert_drawer()` and interact with the collection directly. Those mutations would still occur under `--dry-run`.

Please either:

- pass genuinely non-mutating read-only/recording collection and KG proxies in dry-run mode; or
- change the execution contract so every adapter mutation is mediated by core.

Please add a fixture adapter that writes directly through both `palace.drawer_collection` and `palace.knowledge_graph`, and verify that a dry run leaves both stores unchanged.

fatkobra (changes requested):
Blocking: the executable RFC 002 loop silently discards every `SourceItemMetadata` value and never calls `adapter.is_current(...)`.

A compliant lazy/incremental adapter therefore cannot use the published protocol: core never performs the currentness decision, so unchanged items can still be extracted and re-upserted, while the metadata handshake has no effect.

Please either implement the complete:

`SourceItemMetadata -> is_current -> skip/extract -> write`

handshake before exposing arbitrary registered adapters through `mempalace mine`, or explicitly reject adapters using that protocol until the core runner supports it. Silently accepting and discarding the metadata makes a conforming adapter appear to work while bypassing its incremental behavior.

Please add a fixture adapter that yields `SourceItemMetadata`, asserts that `is_current()` is called, and verifies that a current item does not produce a drawer write.

ggettert:
Thanks, @fatkobra — addressed in b6fad9e.

- **`--dry-run`:** adapters now receive recording/no-op drawer-collection and knowledge-graph facades, so direct writes through the published `PalaceContext` surface are intercepted rather than persisted. The new fixture verifies those writes do not reach the backing-store doubles.

- **Incremental protocol:** adapters advertising `supports_incremental`, and adapters yielding `SourceItemMetadata`, are now rejected explicitly rather than being silently accepted while currentness metadata is ignored. This takes the review’s fail-fast alternative; full RFC 002 `SourceItemMetadata → is_current() → skip/write` support remains follow-up work.

fatkobra (changes requested):
Thanks for the follow-up. b6fad9e improves both areas from my original review:

- adapters advertising `supports_incremental` are now rejected before ingest instead of silently bypassing their currentness contract; and
- the recording drawer/KG facades prevent direct calls through those published mutation methods from reaching the backing-store doubles during dry-run.

I still cannot approve the current head because three correctness gaps remain.

1. Dry-run still opens the real collection with creation enabled.

`mine_source_adapter()` calls `get_collection(palace_path)` before installing `_DryRunCollectionProxy`. `get_collection()` defaults to `create=True`, so a dry run against a fresh palace can create the backend database/collection and may record the embedder identity before the proxy exists.

The current regression mocks `get_collection()`, so it does not exercise this effect.

Please make dry-run use a genuinely non-creating/read-only collection view, or an empty recording facade when no collection exists. Add an integration regression against a temporary fresh palace and verify that no backend artifact, collection, or embedder metadata is created.

2. The source-adapter path does not hold one palace writer lease across the complete ingest.

The existing project and conversation miners wrap their full non-dry-run operations in `mine_palace_lock(palace_path)`. This new path opens the collection and knowledge graph and runs the adapter without that outer lease.

Per-call collection locking is insufficient because an adapter retains the same local backend handle across multiple writes, another process can interleave between yields, and direct KG mutations do not pass through the collection lock.

Please acquire the palace lease before opening writable collection/KG handles and retain it through the complete adapter iteration and all writes. Dry-run can skip the writer lease once it is genuinely non-mutating.

Please add a regression where another process owns the palace and verify that adapter ingest, collection creation, KG opening, and writes are all refused before they begin.

3. Runtime rejection of `SourceItemMetadata` can leave a partial ingest.

RFC 002 permits adapters without `supports_incremental` to yield metadata while performing a full re-extract, and eager adapters may interleave metadata and drawer records. The runner currently commits each `DrawerRecord` immediately and raises when metadata later appears. A sequence such as `DrawerRecord -> SourceItemMetadata` therefore leaves persisted data behind while the command reports failure.

The pre-ingest rejection for adapters advertising `supports_incremental` is sufficient for this scoped runner. For adapters that do not advertise it, please accept metadata as non-incremental progress/version information and continue the full re-extract, preferably with an explicit warning. Unexpected result types should fail explicitly rather than being silently ignored.

Finally, please clarify the scope/closing language. The two adapters #2062 says this work should unblock, #1944 and #1484, both advertise `supports_incremental`, so the current runner intentionally cannot execute them. This can still land as a partial dispatch foundation, but it does not yet make those motivating adapters runnable or fully close that integration gap.

Once dry-run avoids opening writable storage, the complete ingest participates in the palace lease, and metadata cannot produce partial failure, my original concerns will be resolved.

ggettert:
@fatkobra - Thanks for the detailed follow-up. I’ve addressed the three remaining concerns in `141e16f`.

Dry runs now preserve read access to an existing palace without creating anything new; a missing or uninitialized palace uses an empty recording facade instead. Non-dry-run adapter ingestion holds the palace writer lease before opening collection or KG handles and through the full ingest. And non-incremental metadata is accepted as progress information, rather than failing after records may already have been written.

I also updated the PR description to make the scope explicit: this is the CLI-dispatch foundation, while #1944 and #1484 still need the incremental currentness lifecycle.

fatkobra (changes requested):
Thanks for the follow-up. `141e16f` resolves two of my remaining concerns:

- non-dry-run adapter ingestion now holds one palace lease before collection/KG handle creation and through the complete ingest; and
- non-incremental `SourceItemMetadata` no longer causes a late failure after drawers may already have been stored.

The revised scope language is also accurate about incremental adapters remaining follow-up work.

I still cannot approve because the dry-run open is not actually read-only, and the new lease introduces an unhandled CLI contention path.

1. `get_collection(..., create=False)` does not guarantee a non-mutating open.

For `sqlite_exact`, `create=False` still opens a normal writable SQLite connection and unconditionally runs `_init_schema()`, including `PRAGMA journal_mode=WAL`, DDL, FTS/meta writes, and a commit.

For Chroma, `create=False` still reaches `_client()`, whose cold-open path runs database migrations and HNSW quarantine/rename operations before constructing `PersistentClient`. An existing but uninitialized palace directory can also acquire `chroma.sqlite3` during client construction.

The new fresh-palace regression covers only a path that does not exist, where the backend refuses before opening. It does not cover an existing empty directory or an initialized backend.

Please do not open a backend during dry-run unless the backend exposes a genuinely read-only, non-repairing open contract. For this scoped PR, using the empty recording facade for all dry runs may be the safest option.

Please add regressions for:

- an existing but uninitialized palace directory, verifying no Chroma artifact is created; and
- an initialized `sqlite_exact` palace, verifying that the database and WAL/SHM state are unchanged.

2. Direct CLI contention is no longer handled cleanly.

`mine_source_adapter()` can now raise `MineAlreadyRunning`, but the `source_adapter` branch in `cmd_mine()` catches only `UnknownSourceAdapterError` and `UnsupportedSourceAdapterProtocolError` and returns before the existing `MineAlreadyRunning` handler.

A direct `mempalace mine ... --source ...` invocation can therefore produce a traceback when another writer owns the palace.

Please move source-adapter execution under the same contention guard as the legacy miners, or explicitly catch `MineAlreadyRunning` in that branch. Add a CLI-level regression asserting a clean error, no traceback, and a nonzero exit status.

3. The empty collection facade should preserve the backend result contract.

Its `get()` and `query()` methods currently return plain dictionaries. Real RFC 001 collections return `GetResult` and `QueryResult`, and adapters may use attribute access such as `.ids`.

Please return `GetResult.empty()` and `QueryResult.empty()`—with the appropriate outer query dimension—so an adapter behaves consistently whether the dry run sees an existing collection or an empty facade.

Once those cases are addressed, the original source-adapter dispatch concerns will be resolved.

ggettert:
@fatkobra — thank you for the careful review. The remaining concerns are addressed in `f414bb8`, `ba163b1`, and `b676b2b`.

1. **Dry-run backend opens:** source-adapter dry runs now always use empty recording collection/KG facades and never call `get_collection()` or construct `KnowledgeGraph`. This avoids the writable sqlite_exact open and Chroma cold-open/migration paths entirely. Coverage now verifies all three storage states:
   - a nonexistent palace remains absent;
   - an existing empty directory does not acquire `chroma.sqlite3`; and
   - an initialized `sqlite_exact` palace's artifact set and bytes (including any WAL/SHM sidecars) remain unchanged.

2. **CLI lock contention:** the `--source` branch now catches `MineAlreadyRunning` and exits 1 with the normal clean `mempalace: ...` error. The regression uses a real competing process holding `mine_palace_lock`, rather than mocking the exception, and asserts no traceback.

3. **Typed dry-run reads:** `_DryRunCollectionProxy.get()` and `.query()` now return `GetResult.empty()` and `QueryResult.empty()`; `.query()` preserves the outer dimension and requested embeddings shape.

Coverage also includes the explicit fresh-palace regression, unsupported adapter yields failing explicitly, and daemon source-adapter dispatch through the same runner. The existing fixes remain in place: direct dry-run collection/KG writes are intercepted, incremental adapters are rejected before ingest, non-incremental metadata is accepted with a warning, and non-dry-run ingestion holds the palace lease from handle opening through the full write sequence.

Focused source-adapter tests and ruff pass locally. The PR body now reflects the all-empty-facade dry-run contract and the intentional incremental-adapter scope boundary. Could you please take another look when you have a chance?

fatkobra (approved):
Thanks — I reviewed the latest commits. The three concerns from my previous
review are resolved:

- dry-run no longer opens the real collection or knowledge graph;
- direct `--source` lock contention now exits cleanly;
- the dry-run collection facade returns typed empty `GetResult` and
  `QueryResult` values.

The previously added full-ingest lease and incremental-protocol safeguards
also remain in place.

igorls (approved):

PR #2070, opened 2026-07-24
fix(hooks): make transcript ingest wing configurable

### LINKED ISSUE #2057: Hook transcript ingest: derive a stable per-project wing instead of hardcoded "sessions"
## Problem

The Claude Code hooks' verbatim transcript ingest is hardcoded to `--wing sessions` in `hooks_cli.py` (`_ingest_transcript`, both the daemon-job payload and the `_spawn_mine` argv — still the case in v3.6.0). Meanwhile the diary checkpoint path already derives a per-project wing via `_wing_from_transcript_path()`.

For users who organize their palace by per-project wings (and backfill them with explicit `mempalace mine <project-transcript-dir> --mode convos --wing <project>`), enabling `hooks.auto_save` causes every already-filed transcript to be re-filed into `sessions`: the "already filed" tracking is per-wing, so the same JSONL files are duplicated wholesale. On a multi-project palace this means tens of thousands of duplicate drawers, growing `sessions` into exactly the undifferentiated pile that per-wing scoping is meant to avoid — and cleanup is hard because `delete_by_source` matches a source file across all wings, so purging `sessions` would also destroy the per-project copies.

## Why "sessions" was presumably chosen

The incremental mine needs a **stable** wing per source file (see the #1231 review note about not re-filing the same JSONL into a different wing on every hook fire). `sessions` is trivially stable.

## Proposal

A cwd-derived project wing is just as stable per file: `_wing_from_transcript_path()` already resolves it deterministically from the transcript's recorded cwd, and a given session JSONL never changes project. So the dedup invariant holds.

Suggested shape (backward-compatible):

```json
"hooks": { "transcript_wing": "project" | "sessions" }
```

- `"sessions"` (default): current behavior, no migration surprises.
- `"project"`: `_ingest_transcript` uses `_wing_from_transcript_path(transcript_path)` — same derivation the diary already uses, same `wing_sessions` fallback when no cwd is resolvable.

Happy to send a PR with the config key, the wing routing, and tests if this direction is acceptable.

## Environment

MemPalace 3.6.0, Claude Code hooks (`--harness claude-code`), Linux. Verified with `--dry-run`: files already filed in a project wing are re-planned in full when targeting `sessions` (per-wing tracking).

grtninja:
## Summary

Closes #2057.

Hook-driven transcript mining is currently hardcoded to `--wing sessions`, even though the diary checkpoint path already derives a stable project wing from the transcript JSONL `cwd`/path. This adds a backward-compatible routing policy:

```json
{
  "hooks": {
    "transcript_wing": "sessions"
  }
}
```

- `sessions` remains the default and preserves current behavior.
- `project` derives the stable wing through the existing `_wing_from_transcript_path()` logic.
- `MEMPALACE_HOOK_TRANSCRIPT_WING` overrides file config.
- invalid values fail safe to `sessions` rather than silently relocating memories.

## Implementation

- Add `MempalaceConfig.hook_transcript_wing` with `sessions|project` validation.
- Resolve the transcript wing once for both direct and daemon hook write routes.
- Keep ambiguous project derivation on `sessions` rather than creating a surprise `wing_sessions` migration.
- Include the selected wing in daemon mine dedupe keys so the same source can be intentionally routed to a different wing without sharing an in-flight slot.
- Preserve the existing direct-spawn per-target guard, whose command hash already includes `--wing`.

## Test-first verification

The repair branch was created from current upstream `develop` and used a temporary owner-controlled test-first workflow. The workflow:

1. added the regression tests and observed the pre-fix failures;
2. applied the minimal config/routing change;
3. ran the focused transcript-wing tests;
4. ran `tests/test_hooks_cli.py` and `tests/test_hook_write_routing.py`;
5. ran Ruff, `compileall`, and `git diff --check`;
6. removed its temporary repair script/workflow before committing.

The submitted branch contains only:

- `mempalace/config.py`
- `mempalace/hooks_cli.py`
- `tests/test_hooks_cli.py`

## Risk

Low and opt-in. Existing installations remain on the historical `sessions` wing unless they explicitly select `project`. No storage schema, existing drawer, or default hook behavior is changed.


fatkobra (changes requested):
Blocking: `hooks.transcript_wing=project` derives one wing from the triggering transcript and then mines `path.parent` using that single wing.

That assumption is valid for Claude Code’s per-project transcript directory, but not for Codex’s date-bucketed `sessions/YYYY/MM/DD/` layout. One Codex date directory can contain rollout files from several projects.

A hook triggered by project A can therefore submit the whole date directory with project A’s wing, causing sibling transcripts from projects B or C to be filed or re-filed into A’s wing. The daemon route has the same problem because it also submits `source=path.parent`.

Please do one of the following:

1. scope this directory-level policy to transcript layouts where the parent directory is project-specific;
2. mine only the triggering transcript; or
3. partition files by each transcript’s independently derived project wing before mining/submission.

Please add a regression with two Codex rollout files in the same date directory but with different cwd/project values, and verify that neither file is routed through the other file’s wing.

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

grtninja:
[Removed by author: internal coordination telemetry was posted to this public PR in error.]

ATKabli:
Independent confirmation from a small production fleet (2 nodes, shared pgvector
palace): the `"sessions"` hardcode this PR targets is a real problem, not a
theoretical one. We hit it running MemPalace hooks on two machines against one
shared palace — every node's live transcript ingest converged into the same
`sessions` wing regardless of which agent/machine produced it, because the wing is
hardcoded at both `_ingest_transcript` call sites (the daemon submit payload and the
`_spawn_mine` argv). This PR is the only one of the three open hooks-wing PRs that
actually touches both of those call sites, which matches what we needed to fix.

Where our patch differs: we made the wing an arbitrary operator-named value
(`MEMPALACE_HOOKS_WING` env, falling back to a `hooks_wing` config key, falling back
to the historical `sessions` default) rather than a `sessions`/`project` policy
toggle. For a multi-node fleet where the same repos/transcript layouts get checked
out on more than one machine, `project`-derived wings would still collide across
nodes — we needed each node to declare its own stable wing name (`snake`,
`meiling`), not have one derived from the transcript path. Might be worth a third
policy value here (`transcript_wing: "<name>"`) alongside `sessions`/`project`, or a
parallel env var read at the same two call sites.

One more data point in favor of getting this right at ingest rather than fixing it
up later: we had 1,772 drawers land in the wrong wing before we shipped the fix, and
re-winging them was not a metadata edit — drawer ids embed the wing prefix but the
content hash used to derive the id excludes wing, so moving a drawer to the correct
wing means rewriting its id. That's the same shape of problem the reviewer flagged
on #2071's `migrate-wings --rename` (tunnel endpoints going stale after a wing
rename) — once ids/references bake in a wrong wing, fixing it after the fact is
lossy. Getting the wing right at the two `_ingest_transcript` call sites, before
anything is written, is the cheap fix; anything downstream of that is a migration
with edge cases.

Also flagging: the Codex date-bucket sibling-project concern raised above is real
and would bite an operator running Codex hooks against a shared palace the same way
the `sessions` hardcode bit us — worth resolving before this merges regardless of
which config model wins.

PR #2071, opened 2026-07-24
feat: support canonical hook wing consolidation

Michael-WhiteCapData:
## What changed

- Adds optional `hooks.wing` / `MEMPALACE_HOOK_WING` so hook-created diary checkpoints can use one canonical wing instead of deriving a new wing per project path.
- Extends `mempalace migrate-wings` with repeatable `--rename OLD=NEW` mappings for explicit, in-place wing consolidation.
- Preserves existing normalization behavior when no explicit mappings are supplied. Explicit-map mode touches only named source wings.
- Keeps drawer and closet IDs unchanged while updating their wing metadata through the backend API.
- Rewrites matching explicit tunnel endpoints, regenerates canonical tunnel IDs, resolves resulting ID collisions deterministically, and preserves the remaining tunnel metadata.

## Why

Large long-lived palaces can accumulate split session-history wings from encoded Claude project directories, worktrees, and historical hook derivation changes. Operators need a supported way to prevent recurrence and consolidate existing metadata without direct SQLite edits.

## Verification

- Focused feature/regression suite: 371 passed, 1 skipped.
- Full suite in four collection batches under WSL: 3,411 passed, 31 skipped.
- `ruff check .`: passed.
- `ruff format --check .`: passed.

The full suite was batched because a one-shot collection run hit a CPython segfault in pytest assertion rewriting before executing tests; every test file passed when collected in smaller batches.

### DIFF
--- a/docs/hook-write-routing.md
+++ b/docs/hook-write-routing.md
@@ -110,3 +110,19 @@ Configuration file:
         "hooks": "require"
       }
     }
+
+## Canonical diary wing
+
+By default, hook-created diary checkpoints derive a wing from the session's
+project directory. Operators who keep all session history in one wing can set
+an explicit canonical target:
+
+    {
+      "hooks": {
+        "wing": "sessions"
+      }
+    }
+
+`MEMPALACE_HOOK_WING=sessions` provides the equivalent environment override.
+Transcript mining already accepts an explicit wing independently; this setting
+only controls hook-created diary checkpoints.
--- a/docs/recovery/wing-name-migration.md
+++ b/docs/recovery/wing-name-migration.md
@@ -39,6 +39,18 @@ mempalace migrate-wings           # prompts for confirmation
 mempalace migrate-wings --yes     # no prompt
 ```
 
+To consolidate multiple historical wings into one canonical wing, repeat
+`--rename` and preview the exact counts first:
+
+```bash
+mempalace migrate-wings --dry-run \
+  --rename wing_claude=sessions \
+  --rename wing_codex=sessions
+mempalace migrate-wings --yes \
+  --rename wing_claude=sessions \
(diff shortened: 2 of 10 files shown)


fatkobra (changes requested):
Blocking: explicit `--rename OLD=NEW` is an arbitrary namespace migration, but the implementation only rewrites drawer, closet, and topic metadata.

Explicit palace tunnels are persisted separately with `source_wing` and `target_wing` stored as their endpoints, and their IDs are derived from those endpoints. Read-time wing normalization cannot translate an arbitrary mapping such as `wing_claude=sessions`.

After this migration, drawers may live in `sessions` while existing tunnels still point at `wing_claude`, so navigation can silently break and stale tunnel IDs remain.

Please include explicit tunnels in the migration plan and apply phase:

- rewrite either endpoint that matches a source mapping;
- regenerate the canonical tunnel ID;
- handle resulting ID collisions deterministically;
- preserve the tunnel’s remaining metadata.

Please add regressions for source-only, target-only, both-endpoint, and resulting-ID-collision cases.

The existing “no tunnel rewrite needed” reasoning may be valid for punctuation-only canonicalization, but it does not hold for user-supplied arbitrary renames.

Michael-WhiteCapData:
The tunnel migration request is addressed in f563b628c1361e2de8c40e4f069fb728e2c3f8f4.

I verified that the plan and apply phases rewrite either endpoint, regenerate canonical tunnel IDs, and preserve the surviving record's metadata. Existing destination records win collisions; otherwise the smallest original ID wins deterministically.

The regressions cover source only, target only, both endpoints, collision handling, and a persisted migration. I reran the migration and tunnel suites: 66 passed and one skipped. Ruff and formatting checks passed. This is ready for another look at the earlier blocking review.

PR #217, opened 2026-04-08
Add palace compatibility guardrails

yeager:
## Summary

This PR adds basic compatibility guardrails around local palace access so MemPalace fails cleanly when the local Chroma environment is incompatible with the palace metadata.

## What changed

- add `mempalace/compat.py`
- write `mempalace_meta.json` into the palace with:
  - MemPalace version
  - Chroma version
  - Chroma major version
- check palace compatibility before:
  - `search`
  - `wake-up`
  - `status`
  - `compress`
  - layer-based retrieval/search
  - collection access in `miner.py`
- convert top-level `RuntimeError` failures in `cli.py` into clean CLI error messages

## Why

I hit a local case where MemPalace had indexed successfully, but later access crashed because the effective Chroma environment had changed. This PR does not try to solve every Chroma-side issue, but it does make one bad failure mode much safer:

- instead of touching a palace blindly and risking corruption/segfault-like behavior,
- MemPalace now refuses to proceed when the recorded Chroma major version does not match the current one.

## Behavior

Before:
- a palace created under one Chroma major could be opened under another with no explicit warning
- failures were harder to diagnose and could look like deeper runtime breakage

After:
- palace metadata is written on collection creation/use
- incompatible major-version access fails early with a clear error message

## Notes

This is intentionally conservative.
It is better to stop with a clear message and ask for a rebuild than to touch a palace with a known-incompatible local environment.

PR #2081, opened 2026-07-27
fix(mcp): refuse writes when the served library is no longer installed (#899)

### LINKED ISSUE #899: MCP server silently serves with stale library versions after package upgrade
## Summary

Long-running MCP server processes cache `mempalace` and `chromadb` in `sys.modules` at startup and never reload. Users who upgrade mempalace or its dependencies mid-session continue serving tool calls with stale code, leading to silent data corruption or API mismatch errors. Claude Code users regularly hit this because a single Claude Code session (and its dedicated MCP server) can live for days or weeks via session resume.

## Problem scenario

We experienced this during the `chromadb` 0.6.x → 1.5.7 transition enabled by PR #690 (which removed the `<0.7` upper bound):

1. Claude Code session started on day 1 with `mempalace` + `chromadb 0.6.x` — MCP server imports both into memory
2. User runs `pipx install --force -e .` on day 5 (or any other upgrade path: `pip install -U mempalace`, `uv pip install -U`, etc.) — pipx venv now has `chromadb 1.5.7`
3. **The already-running MCP server is oblivious** — `sys.modules['chromadb']` is still pinned to 0.6.x in memory
4. User continues issuing `mempalace_add_drawer` / `mempalace_diary_write` calls via MCP
5. Those writes go through the 0.6.x code path and produce database rows with legacy format (e.g., `seq_id` as BLOB instead of INTEGER)
6. A fresh process (CLI mining, or a newly-started Claude Code session) reads those rows and crashes during compaction with errors like:
   ```
   chromadb.errors.InternalError: Error in compaction: Error reading from metadata segment reader:
   error occurred while decoding column 0: mismatched types;
   Rust type `u64` (as SQL type `INTEGER`) is not compatible with SQL type `BLOB`
   ```
   or
   ```
   'dict' object has no attribute 'dimensionality'
   ```

There is no warning or error at tool-call time in the stale MCP server — it reports success. The inconsistency is only discovered later, when a fresh process hits the legacy rows.

## Impact

- **Silent data inconsistency** across multi-day Claude Code sessions
- **Bug amplification:** stale-library writes contribute to the compaction crashes that PR #796 is trying to mitigate on the miner side
- **Multi-session blast radius:** one user can have 3+ Claude Code sessions running (on different projects, or via remote-wrapped MCP like SSH-to-raindance). Each session's dedicated MCP server has its own staleness timer, and all of them write to the same palace.
- **Cross-version silent merges:** a single palace can accumulate rows written by different mempalace+chromadb versions over days, with no audit trail or warning.
- **Existing mitigations don't help:** `mempalace migrate` detects read compatibility but not write/compaction compatibility (#722). PR #735 addresses the schema validation gap after migration, but a stale in-memory process bypasses migration entirely.

## Reproducing

1. `pipx install -e .` — note the installed `mempalace` and `chromadb` versions
2. Start Claude Code and make any MCP tool call (forces MCP server startup → imports)
3. On the host, `pipx install --force -e .` or `pip install -U chromadb` — upgrade in the same venv
4. In the existing Claude Code session, call `mempalace_add_drawer` — succeeds using stale library
5. Start a new Claude Code session or run `mempalace mine` from the CLI — crashes on the first write that touches metadata touched by the stale process

## Suggested fixes (not mutually exclusive)

**A. Version check at tool-call time (minimum viable fix)**

On MCP server startup, record the running versions:
```python
import importlib.metadata
_STARTUP_MEMPALACE_VERSION = importlib.metadata.version(\"mempalace\")
_STARTUP_CHROMADB_VERSION = importlib.metadata.version(\"chromadb\")
```

On each tool call, compare against the currently installed versions. If they differ, return a clear error:
```json
{
  \"error\": \"MCP server library is stale (started with mempalace X.Y.Z / chromadb A.B.C, disk has X.Y.W / chromadb A.B.D). Restart the MCP server to pick up the new version.\",
  \"action_required\": \"restart_mcp_server\"
}
```

**B. Refuse writes on mismatch**

If any version mismatch is detected, disable write tools (`add_drawer`, `diary_write`, `kg_add`, `mine` via MCP, etc.) but keep read tools functional. Prevents further inconsistency while still letting the user read their palace.

**C. Self-restart via `os.execv()`**

More aggressive: on detecting staleness, `os.execv(sys.executable, [sys.executable, \"-m\", \"mempalace.mcp_server\"])` to replace the current process with a fresh Python. Risk: loses any in-flight state, not always safe in Python processes with threads or C extensions (chromadb has Rust bindings).

**D. Documentation-only mitigation (until a code fix lands)**

Update README, plugin docs, and PR #340's description with a note about restarting MCP servers after any mempalace or chromadb upgrade. This is what Claude Code users currently need to do manually.

## Related

- **PR #340** — adds `mempalace-mcp` console entry point making the MCP server actually work for pipx/uv users. This bug was latent before #340 because pipx/uv users couldn't run the MCP server at all. Our fix unlocks the door.
- **PR #690** — removed `chromadb<0.7` upper bound, creating the version mismatch gradient users can cross
- **Issue #722** / **PR #735** — post-migration schema validation. Fixes a related but different gap (migrate command completeness), not the in-process staleness
- **PR #796** — batch upserts in miner to prevent compaction crashes. Mitigates the downstream symptom, not the root cause

mvalentsev:
## What does this PR do?

Closes #899.

A long-lived MCP server imports `mempalace` and its storage backend once and serves from those in-memory modules for the life of the process. Upgrading the package on disk mid-session cannot reach it, because Python never reloads `sys.modules`, so the server keeps accepting writes produced by code the user no longer has installed and reports success for each one. Reproduced on `develop` over stdio: a server started at 3.6.0 kept answering `tools/call` after the installed distribution moved, with no warning anywhere, and `ping` still reported the connection healthy.

This adds a dispatch-level gate:

* Mutating tools are refused with JSON-RPC `-32005` once a watched distribution's installed version differs from the one recorded at startup, or once it is gone entirely (reported as `not installed`). The error names both versions, sets `action_required: "restart_mcp_server"`, carries `override_env`, and says that `mempalace_reconnect` reopens the palace but cannot reload Python modules.
* Reads are untouched. The palace is intact at that point, and a user who has just been told to restart still needs `search` and `status`.
* `mempalace_status` gains `library_versions`, so the condition is diagnosable and not only observable as a refusal.
* `MEMPALACE_MCP_ALLOW_STALE_LIBRARY=1` disables it, following `MEMPALACE_MCP_ALLOW_PEER_WRITER`.

### Why refuse rather than warn

#457 is this condition with the volume turned up: upgrading mempalace moved the installed ChromaDB across a storage-format boundary (downwards, 1.5.6 to 0.6.3), and every tool call then failed loudly until the host was restarted. #899 is the quiet version. The writes succeed, in a format the newly-installed library may not read back, and nothing surfaces until a fresh process opens the palace. A warning leaves those writes happening. Restarting is also the remedy #457's reporter and the maintainers already gave there.

The gate sits beside the read-only (`-32003`), SQLite integrity (`-32002`) and diverged-index (`-32004`) gates and refuses the same way, taking `-32005`. Corruption is checked first, so its repair instruction is not replaced by this one. This gate then runs ahead of the diverged-index one: that gate's payload directs the operator to `mempalace repair rebuild-index`, a separate process running the installed code, without mentioning the restart this server needs first. Its probe re-runs per call, so the index verdict still surfaces on the first call after that restart. Both directions of the precedence are covered by tests.

### What it deliberately does not do

It is not a general staleness detector. `importlib.metadata` suppresses failure at both levels it reads: `read_text` swallows `PermissionError` and friends, so unreadable metadata arrives as an empty version, and `FastPath.children` wraps its `os.listdir` in `with suppress(Exception)`, so a search root that will not open arrives as no distribution at all, indistinguishable from a real uninstall. Taken at face value the second one refuses every write on a healthy install whenever site-packages is briefly unreadable, which on a threaded server is what file-descriptor exhaustion looks like. Both are recovered explicitly and fail open, named in `library_versions.unreadable`. A genuine uninstall, seen through a search path that opened cleanly, still refuses.

A third reading of the same module is not a suppression but a cache. `FastPath.search` memoizes each search root's listing against that root's mtime, taken in seconds where this gate compares nanoseconds, so an upgrade whose removal and creation both land inside one timestamp tick, about 15 ms on a Windows clock, leaves the memo naming the `dist-info` that upgrade has already deleted. Its version then reads as empty, the distribution is recorded unreadable and left uncompared, and nothing writes to the root afterwards to move its mtime again, so the gate would stay off for that distribution for the rest of the process. That is the upgrade this is here to catch, so the memo is dropped before each reading rather than trusted.

### Scope

The MCP server only. `mempalace daemon serve` has the same exposure but is opt-in and not yet the default write path (`write_routing.py` states it "changes no caller defaults by itself", and `resolve_write_routing_policy` defaults to `DIRECT`), so it is left for a follow-up.

Watched distributions are `mempalace`, plus `chromadb` when chromadb is the backend actually serving. Those are the ones whose code writes the local on-disk format. chromadb is a hard dependency rather than an extra, so it is installed even for a palace served from pgvector, qdrant or milvus, and watching it there would refuse those users' writes whenever chromadb alone was upgraded, over a library that touches nothing they own. Their own client libraries stay out of it too: those backends keep their format server-side. A backend that cannot be resolved at import counts as chroma, since watching a distribution that turns out not to matter costs a restart and not watching the one that does costs the corruption this is here to prevent.

Related but distinct: PR #217 records the ChromaDB major version in palace metadata and refuses to open a palace written under a different one. That answers "was this palace written by a different library"; this answers "did my library change while I was running". No file overlap.

Docs: the `mempalace_status` return line gains `library_versions`, plus the `sqlite_integrity` it had already fallen behind on, and the write-tools section gains a note about the refusal, including the caveat that the gated set does not coincide with that section.

Thanks @messelink for the report in #899 and for laying out the options. What landed is A plus B from that write-up, down to the `action_required` string, so the commit is co-authored.

## How to test

```bash
uv run pytest tests/test_mcp_server.py -k StaleLibrary -v   # 66 passed
uv run pytest tests/ -q --ignore=tests/benchmarks           # 3467 passed, 31 skipped
```

New coverage in `TestStaleLibraryGate`: version drift and outright removal; reads and the escape hatch staying open; both of the metadata-suppression paths driven against real broken files and a real unreadable directory rather than a patched finder; a genuine uninstall still refusing; the startup baseline surviving `importlib.reload`; cache invalidation on a rewritten `dist-info` and on a same-mtime rewrite; an upgrade seen through a search root whose own mtime never moved, which `importlib`'s memo would otherwise answer from; the excluded working directory being resolved rather than merely recorded; the verdict never being memoized; deduplicated logging; and the gate's place in the preflight order against the read-only and SQLite gates.

Also driven end to end through a live `mempalace-mcp --transport stdio` process, which is worth doing rather than trusting the unit tests: an earlier revision passed them all while a cache window let one post-upgrade write through. Start the server, file a drawer, upgrade or reinstall `mempalace` in the same environment while it is still running, and the next `mempalace_add_drawer` is refused with `-32005` naming both versions, while `mempalace_search` still answers, `mempalace_status` reports `library_versions.stale: true`, and the drawer filed before the upgrade is still retrievable. Restarting clears it.

Cost on an environment with four search-path entries and 100 installed distributions, best of twenty-five rounds and reproduced across three runs: 0.0005 ms on a read tool, which returns on the tool name before touching the filesystem; 0.7 ms on a write with the fingerprint unchanged; 3.5 ms when it moved and the metadata is reread, the larger part of that being the relisting the dropped memo forces. The cache is keyed on that fingerprint and never on elapsed time, because a time window is the gap a post-upgrade write slips through.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -468,6 +468,107 @@ _READ_ONLY_REFUSED_TOOLS = _MUTATING_TOOLS | {
 }
 
 
+# Stale-library write gate (#899).
+#
+# A long-lived MCP server imports mempalace and its storage backend once and
+# then serves from those in-memory modules for the whole life of the process.
+# Upgrading the package on disk mid-session (`pip install -U mempalace`,
+# `uv tool upgrade`, `pipx install --force`) cannot reach it: Python caches
+# modules in sys.modules and never reloads them. The server keeps accepting
+# writes, produced by code the user no longer has installed, and reports
+# success for every one of them.
+#
+# #457 is the same condition with the volume turned up: upgrading mempalace
+# tightened its ChromaDB pin and moved the installed ChromaDB across a
+# storage-format boundary (downwards, 1.5.6 to 0.6.3), the running server kept
+# serving the modules it had already loaded, and every tool call then failed
+# with `no such column: collections.schema_str` until the host was restarted. That one
+# announced itself. #899 reports the quiet variant, where the writes succeed in
+# a format the newly-installed library may not read back and nothing surfaces
+# until a fresh process opens the palace. Either way nothing told the running
+# server to stop first; `mempalace migrate` (#502) only recovers such a palace
+# after the fact.
+#
+# Watched distributions are the ones whose code writes the palace on this
+# machine: mempalace itself, plus chromadb when chromadb is the backend
+# actually serving. Its on-disk format has moved between releases, which is
+# what makes a superseded copy of it dangerous. The networked backends
+# (pgvector/qdrant/milvus) keep their format server-side, so a client-library
+# upgrade does not rewrite local files the same way, and they are deliberately
+# left out rather than gated on a guess.
+#
+# chromadb is a hard dependency rather than an extra, so it is installed even
+# for someone serving from pgvector, and watching it unconditionally would
+# refuse that person's writes over an upgrade to a library that touches nothing
+# they own. The backend is read once here, at import, from the same config the
+# server goes on to serve with. One that cannot be resolved counts as chroma:
+# watching a distribution that turns out not to matter costs a restart, while
+# not watching the one that does costs the silent corruption this exists to
+# prevent.
+#
+# The resolved versions are cached behind a stat-only fingerprint, so the common
+# case reads no metadata at all. That is not the same as free: building the
+# fingerprint lists every search root once and stats the watched metadata files
+# it finds there, and on a normal environment the listing is most of the cost.
+# It is paid only by mutating tools — the refusal checks the tool name before
+# anything else, so a read leaves the filesystem alone entirely. The one
+# exception among reads is mempalace_status, which pays it deliberately, being
+# the surface that reports this state.
+#
+# The cache is invalidated by that fingerprint and never by elapsed time,
+# because a time window is precisely the gap a post-upgrade write slips
+# through. The fingerprint covers what installers actually do: a renamed or
+# removed metadata directory, and a metadata file rewritten in place. It cannot
+# see a rewrite that leaves both the size and the recorded mtime untouched,
+# which needs the replacement to be the same length AND to land inside the
+# filesystem's timestamp granularity — nanoseconds on ext4, but roughly 15 ms
+# on Windows and two seconds on FAT. An upgrade arriving minutes into a session
+# clears that comfortably; an archive restore or a rewrite in the same tick as
+# the previous reading does not, and that case fails open, leaving the gate no
+# worse than its absence.
+# -32004 belongs to the diverged-index gate; this one takes the next free code
+# so a client can tell "restart the server" from "rebuild the index" without
+# parsing the message.
+_STALE_LIBRARY_ERROR_CODE = -32005
+
+
+def _stale_library_watched_dists() -> tuple:
+    """Distributions worth comparing, given the backend this server serves with."""
+    watched = ["mempalace"]
+    try:
+        backend = str(_config.backend).strip().lower()
+    except Exception:
+        # Config trouble is not a reason to narrow the check.
+        logger.debug("stale-library backend could not be resolved", exc_info=True)
+        backend = "chroma"
+    if backend == "chroma":
+        watched.append("chromadb")
+    return tuple(watched)
+
+
+_STALE_LIBRARY_WATCHED_DISTS = _stale_library_watched_dists()
+_MCP_ALLOW_STALE_LIBRARY_ENV = "MEMPALACE_MCP_ALLOW_STALE_LIBRARY"
+# A distribution version is PEP 440 text. This metadata is a file on disk that
+# the server does not own, and its value is quoted back to the client, so its
+# shape is checked before it is echoed anywhere.
+_VERSION_TEXT_RE = re.compile(r"\A[A-Za-z0-9._+!-]{1,64}\Z")
+# Reported in place of a version when a distribution that was installed at
+# startup is no longer installed at all. Not a valid PEP 440 version, so it can
+# never collide with a real one.
+_UNINSTALLED = "not installed"
+_stale_library_cache_lock = threading.Lock()
+_stale_library_cache: dict = {"signature": None, "versions": {}, "errors": {}}
+# What has already been announced to the log, so a persistent condition is
+# reported once rather than on every mutating call. These have their own lock
+# because they are not part of the cached reading and outlive it: the cache is
+# dropped whenever a reading fails, while what was last logged has to survive
+# exactly that. Putting them behind the cache lock would also make the log
+# bookkeeping contend for it on every mutating call.
+_stale_library_log_lock = threading.Lock()
+_stale_library_reported_errors: dict = {}
+_stale_library_reported_drift: list = []
+
+
 def _truthy_env(name: str) -> bool:
     return os.environ.get(name, "").strip().lower() in {"1", "true", "yes", "on"}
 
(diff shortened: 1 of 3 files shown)


messelink:
Thanks @mvalentsev — glad #899's write-up mapped to something you could pick up cleanly, and grateful for the co-author credit.

Heads-up that we've since migrated to pgvector, so we don't exercise the primary chromadb-drift path #899 was written about anymore — happy to soak the mempalace-half of the gate (pipx reinstall triggering the refusal, `library_versions` appearing cleanly on a non-chroma backend) if that's useful, otherwise we'll just watch it merge.

The `sys.path` embedded-NUL detail is the kind of thing that would have taken us a long time to notice on our own — nice catch.

mvalentsev:
Thanks @messelink, just pushed a fix for that: chromadb is now watched only when it is the backend serving, so a pgvector palace will not get writes refused over it. Would be grateful if you get a chance to try it.

Yet another cursed chromadb task in mempalace. My kind of challenge.

messelink:
Tried it — `chromadb` is correctly out of the watch list on a pgvector palace (`('mempalace',)`), and still in it when chroma is the serving backend, so default users keep the coverage. `library_versions` reports cleanly through the status path (`{"stale": false, "serving": {"mempalace": "..."}, "packages": []}`), and no spurious refusal on a healthy install. Cherry-picked onto our working branch across two peers.

Nice touch that a config resolution failure falls back to watching chromadb rather than dropping it — narrowing on the error path is exactly how a safety gate quietly stops being one.

> Yet another cursed chromadb task in mempalace. My kind of challenge.

Ha — better you than me. Enjoy.

igorls (approved):

PR #2082, opened 2026-07-27
fix(entity): guard candidate extraction against ReDoS on base64/minif…

### LINKED ISSUE #2063: Entity extraction regex backtracks catastrophically on base64/minified content — mines pinned for days
## Symptom

`mempalace mine` pins one core at 100% for **days** without completing a single file. py-spy shows it parked on the same chunk across dumps taken hours apart:

```
Thread (active+gil): "MainThread"
    _candidate_entity_words (mempalace/palace.py:535)
    _extract_entities_for_metadata (mempalace/miner.py:1011)
    _build_drawer_metadata (mempalace/miner.py:1370)
    process_file (mempalace/miner.py:1521)
```

with `text` being a base64-like blob (`"RgAAAAAAAAcAgIwAAAA..."`). Observed on v3.6.0 after fixing #2054 — a mine ran **39 hours** on one 5000-char entity-extraction window.

## Root cause

The English candidate pattern

```
\b([A-Z][a-z]+(?:[A-Z][a-z]+|[A-Z]{2,})+|[A-Z][a-z]{1,19})\b
```

has nested ambiguous quantifiers: `(?:[A-Z][a-z]+|[A-Z]{2,})+` can partition a long uppercase run in exponentially many ways. On natural text this never bites, but conversation transcripts and code files routinely contain long whitespace-free mixed-case runs (base64 payloads, minified JS, hashes, data URIs). A single ~5000-char window of such content makes `rx.findall()` effectively never return.

Repro:

```python
import re
rx = re.compile(r"\b([A-Z][a-z]+(?:[A-Z][a-z]+|[A-Z]{2,})+|[A-Z][a-z]{1,19})\b")
blob = "RgAAAAAAAAcAgIwAAAAAAAAOAQEIABAAAAAADAISAAAwAAAAAAgENAAAYAAAAAAACHgAAMAAAAAAAADgEBGAAAAAAAABwCAjAAAAAAAAA4BC6zKVNvr6q9o6tb1" * 40
rx.findall(blob)   # does not return (hours+)
```

## Fix (verified locally)

Sanitize the window before matching — no natural-language entity is a 30+ char whitespace-free token, and the pre-pass is linear:

```python
# in _candidate_entity_words, before the findall loop
text = re.sub(r"\S{30,}", " ", text)
```

Same 5000-char blob: **0.5 ms**, entity output on natural text unchanged (verified: names/products still extracted). Alternatives worth considering upstream: possessive quantifiers (Python 3.11+: `(?:[A-Z][a-z]+|[A-Z]{2,})++`) on the pattern itself, or both.

Related: #2054 (chunk_text O(N*K)) — the two together made large-transcript mines look permanently hung; we hit them back-to-back on the same corpus.

bkr499:
…ied content

The i18n candidate_pattern's nested ambiguous quantifiers `(?:[A-Z][a-z]+|[A-Z]{2,})+` partition long whitespace-free mixed-case runs (base64 blobs, minified code) in exponentially many ways, causing catastrophic backtracking that pins `mempalace mine` at 100% CPU indefinitely.

Add `_strip_long_tokens()` to collapse `\S{30,}` runs to a single space before the candidate regexes run, applied at both entry points: `extract_candidates()` and `palace._candidate_entity_words()` (the per-drawer mining hot path). The guard pattern is linear and cannot itself backtrack. Natural-language extraction is unaffected — proper nouns are far shorter than 30 chars.

Reproduced with `"Xa" + "B"*42 + "1"`; adds regression tests with thread-watchdog timeouts so a regression fails fast instead of hanging.

Fixes #2063.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/entity_detector.py
+++ b/mempalace/entity_detector.py
@@ -272,6 +272,30 @@ SKIP_FILENAMES = {
 # ==================== CANDIDATE EXTRACTION ====================
 
 
+# Proper nouns are short and space-delimited; a whitespace-free run of 30+
+# characters is base64, a minified blob, or a code identifier — never an
+# entity. This pattern is linear (a single greedy quantifier over a character
+# class, no alternation or nesting) so it cannot itself backtrack.
+_LONG_TOKEN_RE = re.compile(r"\S{30,}")
+
+
+def _strip_long_tokens(text: str) -> str:
+    """Collapse whitespace-free runs of 30+ chars to a single space.
+
+    The i18n ``candidate_pattern`` uses nested ambiguous quantifiers
+    (``(?:[A-Z][a-z]+|[A-Z]{2,})+``) that can partition a long
+    whitespace-free mixed-case run in exponentially many ways. Natural
+    language never triggers this, but base64 blobs and minified code embed
+    thousands of whitespace-free characters and pin ``re.findall`` at 100%
+    CPU for hours, hanging ``mempalace mine`` indefinitely (issue #2063).
+
+    Removing such runs before the candidate regexes ever see them drops a
+    5000-char blob from hours to well under a millisecond, with no effect on
+    natural-language extraction (proper nouns are far shorter than 30 chars).
+    """
+    return _LONG_TOKEN_RE.sub(" ", text)
+
+
 def extract_candidates(text: str, languages=("en",)) -> dict:
     """
     Extract all capitalized proper noun candidates from text.
@@ -281,6 +305,10 @@ def extract_candidates(text: str, languages=("en",)) -> dict:
     for English, Latin+diacritics for pt-br, Cyrillic for Russian,
     Devanagari for Hindi). Matches from all languages are unioned.
     """
+    # Guard the candidate regexes against catastrophic backtracking on
+    # base64 / minified content before any pattern runs (issue #2063).
+    text = _strip_long_tokens(text)
+
     langs = _normalize_langs(languages)
     patterns = get_entity_patterns(langs)
     stopwords = _get_stopwords(langs)
--- a/mempalace/palace.py
+++ b/mempalace/palace.py
@@ -26,7 +26,11 @@ from .backends import (
     resolve_backend_for_palace,
 )
 from .backends.embedding_wrapper import EmbeddingCollection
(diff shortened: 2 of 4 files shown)


mvalentsev:
Heads up on the guard's scope. The `\S{30,}` sketch comes from the issue itself, but `\S` matches 65389 of the 65408 codepoints above U+007E, so it also matches Chinese prose: CJK carries no ASCII whitespace, so a whole paragraph is a single run and gets collapsed before any candidate pattern runs.

On current develop:

```python
import re
from mempalace.entity_detector import extract_candidates

text = "张三：方案没问题。张三，明天继续。张三。已经搞定了。团队同意了这个方向。"  # 36 chars, no whitespace at all
extract_candidates(text, languages=("zh-CN",))                           # {'张三': 3}
extract_candidates(re.sub(r"\S{30,}", " ", text), languages=("zh-CN",))  # {}
```

Three other shapes behave the same (a longer variant, 「」 quotes, a numbered list): `{'张三': 3}` before, `{}` after. Nothing recovers it downstream, since `get_entity_patterns(("zh-CN",))` resolves 1 candidate pattern and 0 multi-word patterns, and the known-systems prepass reads the same collapsed text.

The zh tests will not catch this. Both positive ones still pass under the guard (`{'朱宜振': 4}` and `{'张三': 4}`) because their longest whitespace-free runs are 7 and 6 characters, under the threshold. The only input that reaches 30, `test_zh_tw_known_limitation_inline_name_no_boundary`, asserts the name is absent, so it passes too.

Scoping the class to ASCII fixes this without weakening the guard. The exploding pattern is `\b([A-Z][a-z]+(?:[A-Z][a-z]+|[A-Z]{2,})+|[A-Z][a-z]{1,19})\b`, whose classes are ASCII-only, so catastrophic backtracking needs an ASCII run by construction. On the 4920-char blob from the issue both `\S{30,}` and `[!-~]{24,}` give the same result, and `[!-~]` matches nothing above U+007E.

Disclosure: I have #2065 open against the same issue taking that ASCII-scoped route, so read this as a data point rather than a neutral review.

igorls:
Rejected for 3.7.0 spine: CJK-unsafe `\S{30,}` guard; superseded by #2065/#2127 which uses ASCII-only collapse.

PR #2086, opened 2026-07-28
fix(repair): preserve verified temp copy on failed rebuild swap; fail loud on truncated pagination

KeilerHirsch:
## Problem

`rebuild_index()` and `cmd_repair()`'s own inlined rebuild logic both build the new collection into a temporary collection first, verify it, then swap it in for the live collection (`_rebuild_collection_via_temp`, shared by both). Two related gaps in the failure paths around that swap:

**1. Data loss on a failed post-swap re-upload.** Once the live collection is deleted and the fresh replacement is being populated, if that re-upload fails partway through, the previous recovery code deleted the verified temp copy. In `rebuild_index()` it then tried to restore from a pre-rebuild `chroma.sqlite3` file backup — misleading, since the live collection's on-disk HNSW segment directories were already destroyed by the delete that preceded the failure, so restoring only the sqlite file leaves the palace referencing segment UUIDs that no longer exist on disk. In `cmd_repair()` (the actual `mempalace repair` CLI command) the same failure triggered a full-directory `rmtree` + restore-from-backup — which additionally destroyed the just-preserved temp copy in the same step, and printed recovery guidance ("restoring from backup...") that had nothing to do with the temp copy's fate. Either way, the one thing that *was* still fully intact — the verified temp copy — got discarded instead of used.

**2. Silent truncation when offset-based pagination is broken.** `_paginate_ids()`'s fallback path (used when the backend's offset-based `get()` fails) re-fetches the same first page of results with no offset. If that fallback lands exactly on a page boundary, the previous code treated "no new ids in this page" as "we're done" — indistinguishable from "we're stuck and haven't actually seen the rest of the collection." Callers (`scan_palace`, `rebuild_index`, `cmd_repair`) then act on that partial ID list as if it were the complete palace.

## Fix

- Added `_promote_temp_collection()`: recovers a failed live-swap by copying directly from the verified temp collection instead of a stale sqlite file backup or a full-directory rollback. The temp copy is now never deleted once the live collection has already been replaced — it's the only intact data left at that point.
- Both `rebuild_index()` and `cmd_repair()` now call this same promotion helper on a live-swap failure, so the recovery behavior and the printed guidance are consistent regardless of which entry point is used. `cmd_repair()` still makes its pre-repair full-directory backup as a manual last-resort fallback (mentioned in the error text) if promotion itself also fails.
- `_paginate_ids()` now disambiguates the page-boundary case against the collection's own `count()`: if `count()` confirms there's more data than what was collected, it raises loudly instead of returning a silently truncated ID list; if `count()` confirms the collected set genuinely is everything, it completes normally.

## Scope

This covers the two failure-path issues above only. It does **not** address holding a mining lock for the full duration of a rebuild so concurrent writes during a multi-stage rebuild aren't lost — that's a larger, separate change (queuing writes during rebuild rather than blocking them) that needs its own design pass and would follow as a separate PR.

## Tests

17 new/updated unit tests across `tests/test_repair.py` and `tests/test_cli.py`, all against synthetic `MagicMock` fixtures (no real ChromaDB collections or production data involved) — covering the promotion path for both entry points, the truncation-vs-completion disambiguation at the page boundary, and updating the existing failure-recovery integration tests to match the new recovery behavior.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1114,6 +1114,7 @@ def cmd_repair(args):
         _close_chroma_handles,
         _extract_drawers,
         _post_rebuild_cleanup,
+        _promote_temp_collection,
         _rebuild_collection_via_temp,
         check_extraction_safety,
         index_read_recovery_guidance,
@@ -1307,19 +1308,27 @@ def cmd_repair(args):
     except RebuildCollectionError as e:
         print(f"  Repair failed: {e}")
         if getattr(e, "live_replaced", False):
-            print("  Live collection was already replaced; restoring from backup...")
+            temp_name = f"{collection_name}__repair_tmp"
+            print(f"  Attempting recovery: promoting verified copy from '{temp_name}'...")
             try:
                 _close_chroma_handles(palace_path, backend=backend)
-                if os.path.exists(palace_path):
-                    shutil.rmtree(palace_path)
-                shutil.copytree(backup_path, palace_path)
-                print(f"  Restore complete from backup: {backup_path}")
-            except Exception as restore_error:
-                print(f"  Automatic restore failed: {restore_error}")
-                print("  Manual recovery required:")
-                print(f"    1. Remove or rename the broken directory: {palace_path}")
-                print(f"    2. Restore the backup directory to: {palace_path}")
-                print(f"       Backup location: {backup_path}")
+                _promote_temp_collection(
+                    backend,
+                    palace_path,
+                    temp_name,
+                    collection_name,
+                    len(all_ids),
+                    batch_size,
+                    progress=print,
+                )
+                print("  Recovery succeeded: live collection restored from the verified temp copy.")
+            except Exception as promote_error:
+                print(f"  Automatic recovery failed: {promote_error}")
+                print(
+                    f"  The verified pre-swap copy still survives under '{temp_name}' -- do NOT "
+                    f"delete it. Recover manually by promoting it, or restore the full-directory "
+                    f"backup at: {backup_path}"
+                )
         sys.exit(1)
 
     # The bulk delete + re-upsert cycle above leaves the FTS5 inverted index
--- a/mempalace/repair.py
+++ b/mempalace/repair.py
@@ -354,7 +354,7 @@ def _rebuild_collection_via_temp(
             raise RebuildCollectionError(str(exc), live_replaced=live_replaced) from exc
         # The live collection was already deleted (line 294) before this
         # failure. `temp_name` is now the ONLY intact, verified copy of the
-        # data left on disk -- never delete it here (#8). Point the operator
+        # data left on disk -- never delete it here. Point the operator
         # at it instead of destroying the one thing that can still recover.
         raise RebuildCollectionError(
             f"{exc}. The live collection '{collection_name}' was already "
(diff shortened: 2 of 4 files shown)


fatkobra (changes requested):
Blocking: `_paginate_ids()` can still silently return a partial result when
both the offset request and its no-offset fallback fail.

The inner failure path remains:

```python
except Exception:
    break
```

If one or more pages were already collected, that returns a partial ID list
to callers such as `scan_palace()` as though it were complete. If it occurs
on the first page, the function returns `[]`, and `scan_palace()` can report
“Nothing to scan.”

Since this PR's purpose is to fail loudly rather than allow repair code to
act on an incomplete palace, the double-failure path should raise an
actionable error instead of breaking.

There is also a filtered-pagination issue. `_paginate_ids()` accepts
`where`, and `scan_palace(..., only_wing=...)` passes a wing filter, but the
page-boundary disambiguation uses the collection-wide `col.count()`.

For example:

- the selected wing contains exactly 1000 rows;
- other wings contain additional rows;
- offset pagination is broken;
- the fallback correctly returns the complete 1000-row filtered set.

The global count is still greater than 1000, so it is not authoritative
evidence that the filtered result is truncated.

Please either obtain a filter-aware total or treat filtered completeness as
unknown without using the global total as proof of truncation.

Regression tests should cover:

1. one successful page followed by failure of both the offset and fallback
   requests, asserting that the function raises rather than returning the
   partial first page;
2. failure of both requests on the first page, asserting that it does not
   return an apparently valid empty result;
3. exactly 1000 matching filtered rows plus additional non-matching rows,
   asserting that the global collection count is not treated as the filtered
   total.

KeilerHirsch:
Thanks @fatkobra — both are valid, fixed in 9de876d.

**1. Double-failure now raises instead of returning a partial/empty list.**
When both the offset `get()` and the no-offset fallback fail, the branch now raises a `RuntimeError` (chained from the underlying error via `raise ... from`) rather than `break`ing. This covers both sub-cases you called out: a mid-run failure after one or more pages already landed, and a failure on the very first page (which previously returned `[]` and made `scan_palace()` print "Nothing to scan." for a palace it never read).

**2. Global `count()` is no longer used as truncation proof for a filtered listing.**
The count-based disambiguation is now gated on `where is None`. `col.count()` takes no `where` argument, so when `_paginate_ids` is called with a wing filter (as `scan_palace(..., only_wing=...)` does), a global count greater than the collected rows is not authoritative evidence of truncation — the extra rows may live in other wings. For filtered calls, completeness is treated as unknown rather than raising a false positive. (A genuine authoritative check would need a filter-aware total, which chromadb's `count()` doesn't expose; I've left that as a known limitation rather than inventing a second unpaginated count-with-where pass.)

**Regression tests added** (all three fail against the pre-fix code, verified by a stash/pop mutation run):
1. one successful page then double failure → asserts it raises rather than returning the partial first page;
2. double failure on the first page → asserts it raises rather than returning an apparently-valid `[]`;
3. exactly 1000 matching filtered rows plus additional non-matching rows → asserts the global collection count is not treated as the filtered total (`len(ids) == 1000`, no raise).

fatkobra (approved):
Thanks — I reviewed `9de876d`.

The two double-failure cases are correctly fixed: both a first-page failure
and a failure after progress now raise rather than returning an empty or
partial ID list.

Gating the collection-wide `count()` check on `where is None` is also
correct. A global count must not be treated as the filtered total.

Issue #50, opened 2026-04-07
Question: multilingual search support (non-English / multilangual memories)


Jus2Cat:
The default embedding model (`all-MiniLM-L6-v2`) is English-centric. Cross-lingual search (e.g. querying in Russian while memories are stored in English) works partially but with noticeable quality drop.

**Questions:**
1. Is there a plan to support `paraphrase-multilingual-MiniLM-L12-v2` or another multilingual model?
2. Is this configurable already (e.g. via `mempalace.yaml` or env var)?
3. Is the quality degradation on cross-lingual queries a known limitation?

This isn't mentioned in the README. Would be helpful to document even if the answer is "not supported yet".

igorls:
Thanks for this — closing here because it's a better fit for Discussions than the issue tracker. We're reserving Issues for actionable bugs and concrete feature work, and routing questions, ideas, showcases, and open-ended proposals to:

https://github.com/MemPalace/mempalace/discussions

Please feel free to repost there (or link back to this issue) to continue the conversation.

PR #2099, opened 2026-07-29
fix(dialect): whole-word entity match + wire lang_regex into extraction (#50, #53)

KeilerHirsch:
## What

Two independent extraction-accuracy fixes in `dialect.py` (audit Group 8), both directly relevant to non-English (German) content.

### #53 — `encode_entity` substring collision

The fallback loop matched a registered name as a bare **substring** of the query, so a short registered name wrongly claimed a longer, unrelated name's code:

```python
# "Ann" registered  →  encode_entity("Annabelle") returned Ann's code
if key.lower() in name.lower():   # "ann" in "annabelle" == True
```

Fixed to a **whole-word** match (`re.search(r"\b{escape(key)}\b", ...)`): `"Alice"` still matches `"Alice Cooper"`, but `"Ann"` no longer matches `"Annabelle"`. The same substring bug existed in the key-sentence decision-word scoring (`"key"` matched inside `"monkey"`) and is fixed the same way.

### #50 — `lang_regex` loaded but never used

`Dialect.__init__` loaded `self.lang_regex = get_regex()` but `_extract_topics` / `_detect_entities_in_text` / `_extract_key_sentence` never consulted it, so non-English content ran through ASCII-only English patterns:

- The tokenizer `[a-zA-Z][a-zA-Z_-]{2,}` split German words at umlauts (`Größe` → `Gr`/`e`).
- The English-only decision-word list scored German decision sentences at ~0.

Now `__init__` derives `_stop_words`, `_topic_pattern`, and `_decision_words` from `self.lang_regex`, falling back to the English module defaults (`_STOP_WORDS`, a new `_DECISION_WORDS`, and a unicode-aware topic pattern) when a locale omits a key. Added `regex.decision_words` to `en.json` (mirrors the English default) and `de.json` (German). The other 13 locales fall back to the English defaults — no guessed translations.

## Regression fix (caught mid-work by the existing suite)

My first `en.json` `topic_pattern` (`[A-Z][a-z]{2,}|…`) was **more restrictive** than the original hardcode and split `GraphQL` → `Graph`, breaking `test_boosts_technical_terms`. Replaced with a unicode-aware `[^\W\d_][\w-]{2,}` that keeps umlaut words whole **and** preserves CamelCase/snake_case/kebab-case.

## Honest scope notes

- **English behavior:** no change on the existing test suite or normal English prose. The substring→whole-word swap is a deliberate, correctness-improving behavior change that only fires on edge inputs (`"key"` in `"monkey"`); it is not a bit-for-bit no-op, so this is worded as "no observed change on English text," not "identical."
- **Intended side effect:** the unicode-aware `topic_pattern` now treats `getUserById` / `kebab-case-word` as single compound tokens (previously split at each internal capital/hyphen). This is the same "don't split at internal boundaries" property that fixes the `GraphQL` bug.
- **Out of scope (follow-ups, not this PR):** (a) `encode_entity` with an empty-string registered key matches any query via a zero-width `\b\b` — pre-existing (the old substring code had the identical bug), worth a guard later. (b) `Dialect.__init__` calls the module-level `i18n.load_lang()`, mutating process-global locale state; harmless for single-threaded CLI/test use but a latent race if `Dialect` is ever constructed concurrently with different `lang=` (e.g. an MCP server serving multiple locales). Neither is introduced by this PR.

## Tests

`tests/test_dialect_extraction_accuracy.py` — 4 tests (3 demonstrate the regressions RED-before / GREEN-after, 1 positive whole-word check), plus an autouse fixture that restores `i18n`'s process-global current language after each test so `Dialect(lang="de")` does not leak German into later tests. Verified: 3 bug-tests RED before the fix, all GREEN after; broad suite (`-k "dialect or compress or i18n or extract or miner or convo"`) 519 passed / 0 failed; `ruff check` + `ruff format --check` clean.


fatkobra (changes requested):
Blocking: the same substring collision still exists in the user-facing
plain-text compression path.

`encode_entity()` now uses a whole-word match, but `compress()` does not call
that method. It calls `_detect_entities_in_text()`, which still does:

```python
if not name.islower() and name.lower() in text.lower():
```

Therefore, with `entities={"Ann": "X99"}`, compressing text that contains
`Annabelle` can still record `X99` even though Ann was never mentioned.

The new regression tests call `encode_entity()` directly, so they do not
exercise the normal `compress()` path.

Please use one shared entity-boundary helper in both `encode_entity()` and
`_detect_entities_in_text()`, and add a regression test through `compress()`
or `_detect_entities_in_text()` proving that:

- `Annabelle` does not match the registered entity `Ann`;
- `Alice Cooper` still matches the registered entity `Alice`.

It would also be useful for the shared helper to handle registered entities
that begin or end with punctuation, for example `.NET`, `C++`, `C#`, or
`@alice`; explicit `(?<!\w)` / `(?!\w)` boundaries are safer for those than
`\b...\b`.

mempalace/dialect.py
```diff
@@ -347,6 +371,18 @@ def __init__(
         self.aaak_instruction = t("aaak.instruction")
         self.lang_regex = get_regex()
 
+        # Derive locale-aware extraction inputs from lang_regex, falling back to
+        # the English module defaults when a locale omits a key (finding #50:
+        # these were loaded but never consulted, so non-English content ran
+        # through ASCII-only English patterns).
+        _sw = self.lang_regex.get("stop_words")
+        self._stop_words = set(_sw.split()) if _sw else _STOP_WORDS
```
igorls: **[bug]** Every shipped locale has `regex.stop_words`, so the `_STOP_WORDS` fallback never runs. Default `Dialect()` therefore replaces the extraction function-word list (have/been/they/what/because/using/…) with `en.json`'s BM25/conversation filler list. That field is what `get_stopwords()` already uses for search; it was not written as an AAAK topic stoplist. Measured on ordinary prose: "We have been talking about the thing they have made for the project." goes from topics `talking`, `project` to `have`, `been`, `talking`. Existing tests still pass because they lean on capitalized technical terms. The PR's "no observed change on English text" claim is false. The parallel unmerged `fix/dialect-non-latin-topics` branch unions locale stops with `_STOP_WORDS` instead of replacing them.

**Suggestion:** Keep `_STOP_WORDS` for English (optionally union `en.json` extras). For other locales, union or replace deliberately — do not point extraction at the BM25 list unchanged. Do not enlarge `en.json` `regex.stop_words` to compensate; that would change searcher BM25.

mempalace/dialect.py
```diff
@@ -394,8 +430,14 @@ def encode_entity(self, name: str) -> Optional[str]:
...
             return self.entity_codes[name.lower()]
+        # Whole-word match only: a registered name must appear as a full word
+        # in the query, not as a bare substring inside a longer word. Without
+        # this, "Ann" wrongly claims "Annabelle"'s code (finding #53), while a
+        # legitimate multi-word query like "Alice Cooper" must still match
+        # "Alice".
+        name_lower = name.lower()
         for key, code in self.entity_codes.items():
-            if key.lower() in name.lower():
+            if re.search(rf"\b{re.escape(key.lower())}\b", name_lower):
```
igorls: **[suggestion]** `\b{key}\b` is the wrong boundary for several real entity shapes. `\b` is a `\w`/`\W` transition, so it does not fire next to leading/trailing punctuation and it does fire at hyphens. Measured: registered `.NET` / `C++` / `C#` / `@alice` no longer match inside a longer query (old substring did); registered `Ann` still matches `Ann-Marie` and `Jean` still matches `Jean-Pierre`. Concatenated CJK (`田中` in `田中太郎`) also fails. `mempalace/i18n/__init__.py` already documents this `\b` failure and ships `_script_boundary` / `(?<!\w)`-style helpers.

**Suggestion:** Match with `(?<!\w)...(?!\w)` (or the existing script-aware helper) so `.NET`/`C++` still match and hyphenated given names are treated like `Annabelle`, not like `Alice Cooper`. Add those cases next to the Ann/Alice tests.

mempalace/dialect.py
```diff
@@ -451,20 +493,21 @@ def _detect_flags(self, text: str) -> List[str]:
...
-        # Count frequency, skip stop words
+        # Tokenize with the locale topic pattern (keeps umlauts/non-ASCII letters
+        # whole; the ASCII default split German words at the umlaut — finding #50).
+        words = re.findall(self._topic_pattern, text)
+        # Count frequency, skip locale stop words
         freq = {}
         for w in words:
             w_lower = w.lower()
-            if w_lower in _STOP_WORDS or len(w_lower) < 3:
+            if w_lower in self._stop_words or len(w_lower) < 3:
```
igorls: **[suggestion]** After this PR, ja/ko/zh-CN/zh-TW `topic_pattern`s are finally consulted, but they are written as `{2,}` (2-character words are the norm in those scripts) and then dropped by the leftover `len(w_lower) < 3` filter. Korean `토큰`/`집합` and typical 2-character CJK compounds never become topics. The English capitalization `+2` boost is also still applied in every locale, so mixed Hangul/CJK + Latin text still prefers `GraphQL` over script content.

**Suggestion:** Honor the locale pattern's minimum (2 for CJK/Hangul). Restrict the Latin capitalization boost to `lang == "en"` and give non-Latin tokens a small non-ASCII nudge. Cover `Dialect(lang="ko")` / `lang="zh-CN"` so 2-character script tokens survive.

tests/test_dialect_extraction_accuracy.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class TestGermanExtraction:
+    def test_topic_tokenizer_keeps_umlaut_words_intact(self):
+        """#50: German words with umlauts must not be split at the umlaut."""
+        d = Dialect(lang="de")
+        topics = d._extract_topics("Die Größe der Änderung war wichtig für Müller.")
+        # Under the ASCII tokenizer, 'Größe' -> 'Gr'/'e', 'Änderung' -> 'nderung'.
+        # A locale-aware tokenizer keeps them whole.
+        assert not any(t in ("gr", "nderung", "berpr") for t in topics), (
```
igorls: **[suggestion]** The German tokenizer test only forbids split fragments (`gr`, `nderung`, `berpr`). An empty topic list, or topics that omit `größe`/`änderung`/`müller`, still passes. `berpr` is not even in the fixture sentence. There is also no test that English `key` no longer scores inside `monkey`, which the production comment claims to fix.

**Suggestion:** Assert the intact umlaut tokens are present. Add `_extract_key_sentence("The monkey sat. The key insight was X.")` and prove the decision sentence wins.

mempalace/dialect.py
```diff
@@ -347,6 +371,18 @@ def __init__(
         self.aaak_instruction = t("aaak.instruction")
         self.lang_regex = get_regex()
 
+        # Derive locale-aware extraction inputs from lang_regex, falling back to
+        # the English module defaults when a locale omits a key (finding #50:
```
igorls: **[suggestion]** Several new comments restate the diff and embed audit IDs (`finding #50`, `finding #53`) instead of a short why. Same pattern at the `_DECISION_WORDS` header, the `encode_entity` fallback, the topic tokenizer, and the decision-word loop.

**Suggestion:** Keep one short why per site (whole-word so `Ann` ≠ `Annabelle`; locale pattern so umlauts stay intact). Drop the finding numbers and the before/after narration.

Issue #2108, opened 2026-07-29
Single-writer discipline: bound how many daemons exist per palace, and how writers queue when they collide


joshuafontany:
**Is your feature request related to a problem?**

Yes — three problems that share one root: nothing bounds how many writers reach one palace, or how
they behave when they collide.

`write_routing` (#1963) decides **whether** a write takes the daemon path. Nothing decides **how many
daemons may exist**, or **how writers queue** once they arrive. Found running many parallel agent
worldlines against a single palace, across different agents and backends:

1. **Orphaned warm daemons pile up.** Each start binds its own handle on the same HNSW segment, so
   "use the daemon" stops meaning "one writer."
2. **`repair --mode from-sqlite` archives before taking the mine-lock**, so a concurrent mine can
   interleave with the archive step.
3. **Lock contention spams the console.** `mine_palace_lock` is deliberately non-blocking and raises
   `MineAlreadyRunning` so in-process callers can bail — but a harvest of back-to-back
   `mempalace mine` runs against a palace a daemon already holds prints the same
   "held by PID … wait for it to finish" line over and over.

**Describe the solution you'd like**

- A **flock on a per-palace lock file** making the warm daemon a singleton: a second start finds the
  lock held and declines rather than standing a rival writer.
- **Take the mine-lock before the archive** in `repair --mode from-sqlite`.
- A **quiet bounded backoff** for mine-lock contention — one notice plus exponential backoff, with a
  wait budget (`MEMPALACE_MINE_LOCK_WAIT`, default 30s, `0` keeping today's fail-fast), then a clean
  give-up. Matches the deadline-loop idiom already used by the daemon stale-takeover and the
  closet_llm retry. Never spins forever, never spams.

**How this composes with your existing work:** the flock guarantees ONE daemon exists;
`write_routing=require` guarantees every write goes THROUGH it. Together they give single-writer
discipline that holds under load. `prefer` cannot give that alone — its fallback re-opens direct
multi-writer access at exactly the moment contention peaks.

**Describe alternatives you've considered**

`flock` reads as a choice, not the only answer. A PID file with liveness probing, an abstract unix
socket, or a lock row in sqlite would each serve; flock earns it here by releasing automatically when
the holder dies, which matters most in the failure mode that produced the orphans. **If you would
rather stand the singleton on a different mechanism, that answer belongs to you** — I would rather
learn it before the effort goes further.

**Additional context**

[...]

PR #2033, opened 2026-07-15
feat(cli): apply shared daemon write-routing policy

fatkobra:
## What does this PR do?

Builds on the shared routing-policy foundation merged in #2027 and the hook-routing implementation merged in #2030.

Contributes to #1963.

This is the third PR in the staged hook and CLI gateway rollout.

It applies the shared `direct` / `prefer` / `require` policy to routine CLI write operations:

- `mempalace mine`;
- `mempalace sweep`;
- `mempalace sync`;
- the optional post-setup mine run by `mempalace init`.

It also makes `sweep` a first-class daemon job and preserves the one-time project scan performed by `init` when that mine is submitted to the daemon.

### `direct`

Use the existing direct in-process execution path.

### `prefer`

Submit through the local daemon.

Interactive CLI commands are allowed to start the daemon if it is not already running.

### `require`

Submit through the local daemon.

Interactive CLI commands are allowed to start the daemon if it is not already running.

Because CLI commands can start the daemon, both `prefer` and `require` normally select it. The distinction remains important for hook callers, which cannot safely cold-start the daemon.

## Explicit overrides

Force daemon routing:

    mempalace mine ./project --daemon

Force direct execution:

    mempalace mine ./project --direct

The options are mutually exclusive and override environment/config policy.

The same options are available for:

- `init`;
- `mine`;
- `sweep`;
- `sync`.

## Background jobs

`--background` now follows the selected route rather than requiring the literal `--daemon` flag.

For example, this works when CLI routing is configured as `prefer` or `require`:

    mempalace mine ./project --background

A direct route with `--background` exits with an explicit configuration error.

## No ambiguous fallback

Once daemon submission begins, an error never causes the command to rerun directly.

The daemon may have durably accepted the job before the client observed a timeout or transport failure. Retrying the operation directly could duplicate memories or race the accepted job.

## Post-init mine

`mempalace init` already scans the project once to show the user a file-count and size estimate.

When the accepted mine is daemon-routed, this PR forwards that exact scanned file list in the daemon payload. The daemon therefore does not need to walk the project a second time.

## Sweep support

Before this PR, `sweep` was direct-only.

This PR adds daemon service support for:

- single transcript files;
- recursively swept directories;
- normal result summaries;
- partial directory failures;
- lock-contention errors;
- backend and palace isolation.

## Backward compatibility and rollout

The default CLI policy remains `direct`.

Existing users therefore receive no silent execution-topology change.

A supervised Tier 3 deployment can opt in with:

[...]

PR #2109, opened 2026-07-29
fix(repair): take the mine-lock before archiving in repair --mode from-sqlite

joshuafontany:
Addresses #2108.

### Who is asking

Same context as #2107, restated so this PR stands alone. Four instances(extendable) of one system on one machine:
three sovereign "content palaces" (one per sensorium) **plus** whatever install the operator already
runs at `~/.mempalace`. 

**Peers, never satellites.** We consume mempalace as a library from our own
NDJSON holder processes, with the embedding contract inverted — vectors arrive on the
wire, so no model loads in the store process. Everything here comes out of running that shape with
many parallel agent worldlines writing one palace.

### Opening by naming your work, because this composes with it rather than replacing it

**`write_routing` (#1963) decides WHETHER a write takes the daemon path. This decides HOW MANY
daemons may exist, and HOW writers queue once they arrive.**

The two compose into something neither gives alone: the flock guarantees **one** daemon exists;
`write_routing=require` guarantees every write goes **through** it. Together, single-writer discipline
that holds under load. `prefer` cannot reach that by itself, because its fallback re-opens direct
multi-writer access at exactly the moment contention peaks — daemon busy, or daemon starting.

### Three commits

1. **`feat(daemon)`** — a flock on a per-palace lock file makes the warm daemon a singleton. A second
   start finds the lock held and declines rather than standing a rival writer over the same HNSW
   segment. Found with orphaned warm daemons piling up, each holding its own handle.
2. **`fix(repair)`** — take the mine-lock *before* the archive in `repair --mode from-sqlite`, so a
   concurrent mine cannot interleave with the archive step.
3. **`fix(cli)`** — a quiet bounded backoff for mine-lock contention. `mine_palace_lock` stays
   deliberately non-blocking and still raises `MineAlreadyRunning` for in-process callers; this
   collapses a harvest's repeated "held by PID …" spam to one notice plus exponential backoff, bounded
   by `MEMPALACE_MINE_LOCK_WAIT` (default 30s; `0` keeps today's fail-fast with a clean give-up
   message), matching the deadline-loop idiom your daemon stale-takeover and closet_llm retry already
   use. Behaviour otherwise unchanged: it eventually mines, or exits non-zero once the budget is
   spent. Never spins forever, never spams.

### Scope

This PR touches `daemon.py`, `repair.py`, `cli.py`, and their tests.

### On `flock` as the mechanism

`flock` releases automatically when the holder dies, which is the failure mode that produced the
orphans. A PID file with liveness probing, an abstract unix socket, or a lock row in sqlite would
each serve as well — if you would rather stand the singleton on one of those, say so and I will
rebuild it that way.

### Verification

Rebased onto `develop` @ `aa89bd8`.

- `ruff check .` → clean · `ruff format --check .` → 194 files already formatted
- `pytest tests/test_daemon.py tests/test_repair_lock_safety.py tests/test_cli.py` → **124 passed**
- `pytest tests/ --ignore=tests/benchmarks` → **3407 passed, 31 skipped, 5 failed**

**On those 5 failures:** all five are `test_init_filters_sys_path_from_leaked_pythonpath`, and they
fail **identically on clean `develop` in the same worktree** — I ran that control because the test
reacts to *where* it runs rather than to the code under it. Zero regressions here; the cause is a
leaked `PYTHONPATH` in my environment, which is the very thing the test asserts about.

Local ruff reads 0.15.20 against your pinned 0.15.14, so please let CI have the final word on lint.

### DIFF
--- a/mempalace/repair.py
+++ b/mempalace/repair.py
@@ -1471,6 +1471,43 @@ def rebuild_from_sqlite(
             )
             return {}
 
+    # Acquire the single-writer mine-lock BEFORE the archive/rename. The
+    # rebuild upserts into ``dest_palace`` through the same backend write
+    # path that takes ``mine_palace_lock`` per batch; if a daemon or a
+    # concurrent mine already holds it, those upserts used to fail *after*
+    # ``shutil.move`` had already stranded the existing palace aside
+    # (renamed to ``.pre-rebuild-…`` with no rebuilt replacement, leaving a
+    # partial/archived mess). Taking the lock up front makes contention
+    # fail CLEAN — no archive, no partial dest, the palace left untouched —
+    # and runs the whole rebuild as one writer (the inner per-batch
+    # acquires pass through re-entrantly on this thread).
+    from .palace import mine_palace_lock
+
+    with mine_palace_lock(dest_palace):
+        return _rebuild_from_sqlite_locked(
+            source_palace=source_palace,
+            dest_palace=dest_palace,
+            in_place=in_place,
+            batch_size=batch_size,
+        )
+
+
+def _rebuild_from_sqlite_locked(
+    *,
+    source_palace: str,
+    dest_palace: str,
+    in_place: bool,
+    batch_size: int,
+) -> dict[str, int]:
+    """Body of :func:`rebuild_from_sqlite`, run while holding
+    ``mine_palace_lock(dest_palace)`` so the archive/rename and the
+    upserts form one atomic single-writer operation.
+
+    Split out so the lock acquired in :func:`rebuild_from_sqlite` wraps
+    every destructive step (the archive ``shutil.move`` below included).
+    Raising :class:`MineAlreadyRunning` from the wrapping ``with`` happens
+    before this body runs, so a held lock never reaches the archive.
+    """
     archive_path: Optional[str] = None
     if in_place:
         ts = datetime.now().strftime("%Y%m%d-%H%M%S")
@@ -1501,7 +1538,6 @@ def rebuild_from_sqlite(
             )
             return {}
         source_palace = archive_path
-        src_db = os.path.join(source_palace, "chroma.sqlite3")
 
         # In-place only: drop chromadb's process-wide System registry so
         # the new client at dest_palace builds a fresh System. Without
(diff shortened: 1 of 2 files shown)


fatkobra (changes requested):
The repair lock-before-archive change looks correct and valuable: contention is detected before the original palace is renamed, so the palace remains untouched on failure. 

I see blockers in the daemon-singleton commit:

- the singleton is not enforced on Windows: when `fcntl` is unavailable, `_acquire_serve_lock()` returns an unlocked file handle, so the title’s “one daemon per palace” invariant is POSIX-only;
- stale takeover sends `SIGTERM` after one failed health probe based on total daemon age, not duration of unhealthiness, and does not verify that the PID still belongs to the expected daemon;
- the contended `start.lock` path returns an existing client before the later explicit close/finally path, leaving cleanup dependent on interpreter object destruction;
- the new idle timer appears to track queue-job activity only; this needs coordination with #1976 so `/mcp` traffic also keeps the shared daemon alive.

The CLI backoff should be separate. It changes the default from fail-fast to a 30-second wait and overlaps the open CLI-routing work in #2033. I recommend rebasing that idea after #2033 and integrating it into the shared `direct` / `prefer` / `require` policy rather than introducing a second routing policy in `cmd_mine`.

joshuafontany:
Cut back to the repair fix.

The flock singleton withdraws rather than gets patched: your two blockers share one root. Under a kernel-enforced lock the takeover path has nothing to do — the lock releases when the holder dies, so a failed acquire means someone live holds it, and signalling a recorded PID there does harm. A Windows branch leaves that intact. The code wants a deletion plus `filelock`.

The backoff withdraws — I found #1978 afterwards. Your lock tests pin fail-fast deliberately, and I should have searched the closed PRs first.

The contention behind it runs real on our side, so the measurement goes to #2033 as data.

If the singleton earns anything, I'd rather bring it to #1963 as a comment than as a patch beside your in-flight work.

fatkobra (approved):
Thanks for narrowing the PR.

The current 7b49503 resolves my original concern. `mine_palace_lock(dest_palace)` is now acquired before the first destructive archive/rename and remains held through the complete rebuild. A contending writer therefore causes `MineAlreadyRunning` while the original palace is still untouched.

The cross-process regression verifies the important failure properties directly: no archive sibling is created, no partial destination is produced, and the original palace contents remain unchanged.

The daemon-singleton and CLI-backoff concerns from my previous review no longer apply because those changes have been removed.

Approved.

PR #1626, opened 2026-05-26
feat(codex-plugin): bring Codex CLI to Claude parity (30 MCP tools)

trek-e:
## Summary
This PR brings the Codex CLI plugin surface to full parity with the Claude plugin and with the current live MCP runtime surface.

Primary outcomes:
- Align plugin manifests/docs/help text from `19` -> `30` MCP tools
- Keep Claude/Codex plugin identity fields in lockstep via parity tests
- Harden Codex hook wrapper execution/fallback behavior and timeout schema coverage
- Add full environment parity evidence document for reproducible validation

Closes #1625

## What changed
- Codex plugin:
  - update tool-count claims in `.codex-plugin/README.md`, `.codex-plugin/plugin.json`, `.codex-plugin/marketplace.json`
  - add/ensure `.codex-plugin/.mcp.json`
  - add umbrella skill `.codex-plugin/skills/mempalace/SKILL.md`
  - harden hook config + wrapper (`.codex-plugin/hooks.json`, `.codex-plugin/hooks/mempal-hook.sh`)
- Claude plugin:
  - align descriptions/tool-count claims to match Codex (`30` tools)
- Instructions/docs:
  - expand `mempalace/instructions/help.md` MCP section to full 30-tool surface
  - align module doc line in `mempalace/README.md`
  - add full live validation report: `docs/codex-cli-parity-test-results.md`
- Tests:
  - `tests/test_plugin_feature_parity.py`
  - `tests/test_codex_plugin_hook_config.py`
  - `tests/test_codex_plugin_hook_wrappers.py`

## Related / partial prior work
Issues:
- #27
- #184
- #295
- #408
- #462

Open PRs with partial overlap/context:
- #325
- #334
- #410
- #670
- #1187
- #1381

Merged prior context:
- #61, #1077, #1231, #1436, #1470, #1562

## Test evidence
Detailed run log is committed in:
- `docs/codex-cli-parity-test-results.md`

Automated tests executed:
```bash
uv run pytest -v tests/test_plugin_feature_parity.py tests/test_codex_plugin_hook_config.py tests/test_codex_plugin_hook_wrappers.py tests/test_hooks_shell.py
# 27 passed

uv run pytest -v tests/test_plugin_feature_parity.py tests/test_codex_plugin_hook_config.py tests/test_codex_plugin_hook_wrappers.py tests/test_readme_claims.py::TestToolCount::test_readme_tool_count_matches_code
# 25 passed
```

Live environment exercise included:
- plugin marketplace + install verification
[...]

PR #2120, opened 2026-08-01
fix(codex-plugin): resolve hooks from the plugin snapshot root

### LINKED ISSUE #2061: Codex plugin: hooks path "./hooks.json" in .codex-plugin/plugin.json is unresolvable from the snapshot root (new in 3.6.0)
Since 3.6.0, `.codex-plugin/plugin.json` declares `"hooks": "./hooks.json"` (the field did not exist in v3.5.0). Codex CLI resolves that path against the plugin snapshot root — the directory containing `.codex-plugin/` — not against the manifest's own directory. Since `hooks.json` lives at `.codex-plugin/hooks.json`, every Codex session now warns:

```
⚠ failed to read plugin hooks
  config <codex-home>/plugins/cache/<marketplace>/mempalace/3.6.0/hooks.json: No such file or directory (os error 2)
```

This matches the Claude Code convention, where `plugin.json` paths are relative to the plugin root (the directory containing the `.claude-plugin/` folder), not to the manifest folder. The skills field (`"./skills/"`) is presumably affected the same way when the snapshot is the repo root.

Suggested fix: either ship `hooks.json` (and the `hooks/mempal-hook.sh` it references via `${CODEX_PLUGIN_ROOT}`) so the declared paths resolve from the snapshot root, or change the manifest to `"hooks": "./.codex-plugin/hooks.json"`.

Local workaround that silences the warning: drop a neutral `{"hooks": {}}` at the snapshot root.

Environment: Codex CLI (July 2026), mempalace 3.6.0 installed from a marketplace pointing at the repo root, Linux.

Taylor-yuhb:
## Summary

- resolve the Codex hook manifest from the installed plugin snapshot root
- use the documented `${PLUGIN_ROOT}` variable for bundled lifecycle commands
- preserve the packaged hook runner's executable bit
- add regression coverage for component paths and lifecycle execution

## Root cause

Codex resolves component paths in `.codex-plugin/plugin.json` relative to the plugin root, not relative to the manifest directory. MemPalace declared `"hooks": "./hooks.json"`, but the file is packaged at `.codex-plugin/hooks.json`, so a repository-root marketplace snapshot resolves the declaration to a missing file.

The hook commands also referenced the undocumented `${CODEX_PLUGIN_ROOT}` variable and a runner path that does not exist in the unmodified snapshot. The current [Codex plugin contract](https://developers.openai.com/codex/plugins/build#bundled-mcp-servers-and-lifecycle-hooks) provides `${PLUGIN_ROOT}` as the installed plugin root.

## Approach

This PR keeps the current package layout and uses the supported explicit manifest override `./.codex-plugin/hooks.json`. Moving the files to the canonical root-level `hooks/hooks.json` layout is also possible, but would be a broader packaging change; I am happy to adjust to that layout during review if maintainers prefer it.

The existing `"skills": "./skills/"` declaration remains unchanged because the repository snapshot already contains the declared root-level `skills/` directory.

Related: #1626 touches broader Codex parity work but does not correct this manifest-path regression.

## Validation

- focused plugin regression suite: `24 passed`
- `ruff check .`
- `ruff format --check tests/test_codex_plugin_layout.py`
- JSON validation for both plugin manifests
- `bash -n .codex-plugin/hooks/mempal-hook.sh`
- isolated marketplace install with a Codex home path containing spaces; `SessionStart`, `Stop`, and `PreCompact` all resolved to and executed the cached runner

The lock-resolved full suite could not be run on this macOS x86_64 host because the locked `onnxruntime==1.24.4` has no compatible wheel for that platform. CI should provide the full supported-platform matrix; the plugin-focused tests above run without the project dependency environment.

Fixes #2061


pierrefrns:
Live reproduction on Linux with MemPalace 3.9.0 and Codex CLI 0.153.2 confirms the executable-bit part of this fix is still needed.

- the 3.9.0 marketplace checkout tracks `.codex-plugin/hooks/mempal-hook.sh` as `100644`
- the installed Codex plugin snapshot exposes it as mode `0664`
- `.codex-plugin/hooks.json` invokes the runner directly
- direct execution fails with exit 126 / `Permission denied`
- changing only the installed file mode to `0755` makes a `session-start` smoke call return `{}` and route successfully to the native daemon with `write_routing.hooks=require`

No wrapper content change was required. This also reproduces after upgrading an existing installation, because refreshing the plugin snapshot restores the non-executable mode.

Issue #2024, opened 2026-07-15
mine_palace_lock has no staleness timeout: a wedged write inside a long-lived MCP daemon blocks all writes + nightly mining forever


westonale:
## Summary

`mine_palace_lock` (`palace.py`) is an `fcntl.flock`, so it is released only when the
holding open-file-description closes — i.e. when the holder process exits. That is fine
for the CLI miner (short-lived, exits, releases). It is **not** fine for the long-lived
MCP HTTP daemon (`mempalace-mcp --transport http`).

Because ChromaCollection write methods acquire this lock (correctly — it's the guard
against concurrent HNSW inserts), **any** MCP write (`add_drawer`, `kg_add`,
`diary_write`) takes it inside the daemon. If one of those writes wedges, the flock is
held for the daemon's entire lifetime and can never be reclaimed as stale, because the
recorded PID is still very much alive.

Everything downstream then fails with:

```
palace /Users/…/.mempalace/palace is held by PID <daemon> (…mempalace-mcp --transport http);
wait for it to finish or stop the holder before retrying
```

…until a human intervenes. There is no path back on its own.

## Observed

On mempalace 3.5.0 (macOS, chroma backend, ~606k drawers / 3.7 GB palace):

- A write wedged at ~00:35. Daemon sat at **0.0% CPU** (blocked, not busy — only ~1m35s
  CPU total across 21h uptime), holding the flock.
- **All MCP writes failed for ~21 hours.** Reads were unaffected.
- The **03:30 nightly mine also failed** (`launchctl` exit status 1) because it couldn't
  take the same lock. So one leaked write silently killed nightly mining too — that was
  the more damaging half, and it was invisible.
- `lsof` confirmed the daemon still held the lock fd. Recovery required deleting the lock
  file by hand so a fresh `open()` would land on a new inode.

## Why this is inconsistent with the existing design

`hooks_cli.py` already guards the *analogous* failure for mine PID slots:

```python
# Maximum wall-clock hours a mine subprocess is allowed to run before its
# PID slot is treated as stale (even if the process is still alive).  A
# wedged mine — e.g. one that is blocking indefinitely on ChromaDB
# cold-init under concurrent Windows load (#1552) — would otherwise hold
# its slot forever.
_MINE_TIMEOUT_HOURS_ENV = "MEMPALACE_MINE_TIMEOUT_HOURS"
_MINE_TIMEOUT_HOURS_DEFAULT = 2.0
```

[...]

PR #2131, opened 2026-08-02
feat(lock): hand the writer lease between processes instead of holding it for life

Evgen197310:
## The problem

The peer-writer guard (#1818) holds `mine_palace_lock` for the **entire process lifetime**. That is the right correctness boundary — local backends keep in-memory index state that a second long-lived writer would silently invalidate — but the lease is sticky in a way that hurts every multi-process setup:

- The first server to touch a palace owns it until it **exits**. A second Claude Code session is read-only for its whole life, however long the first one sits idle.
- #1960 made a read-only server self-heal, but only when the holder *dies*. Two live sessions still can't share a palace.
- Hook-driven and manual `mempalace mine` are locked out for as long as any MCP server is up (#1888).

The lock has no way to say "I own this, but I'm not doing anything with it."

## What this adds

The lease becomes a **baton**. Two OS locks per palace in `~/.mempalace/locks/`:

| file | meaning |
| --- | --- |
| `mine_palace_<key>.lock` | **ownership** — whoever holds it may write (unchanged) |
| `mine_palace_<key>.want` | **demand** — a contender holds it while queued (new) |

**Contender** — `mine_palace_lock(palace, wait=N)`: try ownership non-blocking; if contended, take the demand lock and poll until ownership lands or `wait` expires. On expiry, the same `MineAlreadyRunning` as today.

**Holder** — a watchdog thread in the MCP server: probe demand (a non-blocking attempt on the demand file), and if a peer is queued, this server has held the lease past a minimum hold, no request is being dispatched, and no write frame is active — close storage handles, release ownership, go read-only until the next write reacquires.

Demand is signalled by *holding a lock*, not by writing a timestamp or a pid: a killed contender stops asking the instant the kernel drops its descriptor, so there is no stale state to garbage-collect and no heartbeat to tune. Queueing on the demand file also serializes contenders, so N idle servers can't stampede one holder into N handoffs.

## Why it can't hand off mid-write

This is the part that would be silent if it were wrong — two processes in one Chroma is exactly the corruption #1818 exists to prevent. Two additions to the re-entrancy layer:

1. **Write-frame depth.** Every executing `mine_palace_lock` body counts, *including* the re-entrant pass-through frames a write takes under an existing lease (those hold no OS lock of their own and trust the lease to still be there — a naive release would let them run unprotected). `lease=True` deliberately does not count: a parked lease is precisely the idle state the handoff targets.
2. **A handoff window.** Depth is checked and the window opened under one lock, and a writer arriving mid-release parks on that condition instead of taking pass-through credit. Without it, a thread that checked ownership microseconds before the release would write with no lock at all.

Reads are covered separately: they serve from cached storage handles *without* taking the palace lock, so closing those handles under a running request would be a use-after-close. The dispatch lock that HTTP already used is now shared with stdio, and the watchdog acquires it non-blocking — a request in flight simply defers the handoff to the next tick.

## Configuration

| variable | default | meaning |
| --- | --- | --- |
| `MEMPALACE_WRITER_HANDOFF_WAIT_SECONDS` | `15` | how long a mutating tool waits for the baton before the usual read-only refusal |
| `MEMPALACE_WRITER_MIN_HOLD_SECONDS` | `5` | minimum ownership time before giving the lease away |
| `MEMPALACE_WRITER_HANDOFF_POLL_SECONDS` | `1` | how often the holder probes for demand |
| `MEMPALACE_WRITER_HANDOFF_DISABLED` | unset | opt out entirely — pre-handoff behavior |
| `MEMPALACE_WRITER_HANDOFF_ENABLED` | unset | opt *in* for the HTTP transport (off there by default) |
| `MEMPALACE_PALACE_LOCK_WAIT_SECONDS` | `0` | let non-MCP callers (`mempalace mine`, hooks) wait for the baton too |

`WAIT > MIN_HOLD + POLL` is a real constraint (a contender must outlast the holder's floor), so it's asserted in a test rather than left to a comment. `MIN_HOLD` is the anti-thrash knob: handing off means reopening storage, so a floor of zero would let two chatty sessions ping-pong the lease and end up slower than the refusal they replaced.

## Compatibility

- **CLI/hooks unchanged by default.** Waiting is opt-in (`MEMPALACE_PALACE_LOCK_WAIT_SECONDS=0`) precisely so N hook-spawned mines keep collapsing to one runner instead of queueing as waiters that then drive parallel HNSW inserts. Operators who would rather have a mine wait than lose its turn can now choose that — the smaller half of #1888.
- **Mixed versions are safe.** A holder running older code never answers the demand probe. The contender waits once, times out, and falls back to exactly today's refusal; a short backoff then keeps one unresponsive peer from making every subsequent write pay the full wait. The feature only engages when both sides run it.
- **HTTP is opt-in** because a writable HTTP server is a deliberate single-writer service that already refuses to start without ownership — handing the palace to a random peer mid-flight should be an operator's decision.
- `daemon.py` now marks its lifetime lease `lease=True`. No behavior change today (it runs no watchdog and is meant to be the sole owner), but without the flag its write-depth would sit at 1 forever and silently refuse any handoff wired up later.

## Testing

- `tests/test_palace_lock_handoff.py` (11) — the protocol: demand signalling, the fail-fast default, handoff refused while a write frame is active, a writer arriving mid-window parking instead of passing through, and two cross-process cases via `spawn`: a waiter receiving the lease from a **live idle** holder, and a waiter correctly timing out against a holder that is **writing**.
- `tests/test_mcp_writer_handoff.py` (11) — the server wiring: release on demand, refusal inside the minimum hold / during dispatch / during a write, the disable switch, the contender's wait and backoff, and the `writer_lease` block in `mempalace_status`.
- Two real MCP servers against one palace, by hand: A writes (takes the lease) → B writes while A is alive and idle (baton moves in ~3 s, no restart) → A writes again (baton returns). After each handoff **both** servers were checked for the other's drawer by exact `get_drawer` *and* semantic search — the point being index consistency, not just that the write succeeded.
- Full suite: `3518 passed, 31 skipped`. One unrelated failure, `test_sqlite_exact_read_only_open_sees_active_writer_wal`, reproduces identically on a clean `develop` checkout (verified with the branch stashed).
- `ruff check` / `ruff format --check` clean.

## Scope

Deliberately not addressed here: a **wedged** write (#2024) still blocks handoff forever — depth never drops, and inventing a timeout for a write that might be legitimately slow belongs in that issue, not this one. Where the daemon owns writes through `require` routing (#1963), this is redundant by construction: nothing queues against the sole owner. It's what keeps direct-writing peers usable everywhere the daemon isn't deployed.

One interaction worth flagging for #1984: the handoff relies on the dispatch lock as a barrier against closing storage under a live request. If that lock is later narrowed to unblock concurrent reads, the watchdog needs to keep *some* barrier — `docs/writer-lease-handoff.md` says so where it explains the rule.

Refs #1888, #1818, #1960, #1963.


Evgen197310:
Note on the last commit (91c9d29): it came out of running this branch on a live palace, not out of the test suite.

After deploying and restarting the MCP servers, the very first `mempalace_status` reported:

```json
"writer_lease": { "held_by_this_server": false, "holder": "PID 3942483 (…/mcp_server.py --palace …)" }
```

…naming a process that had been killed minutes earlier. The lock-file body records whoever took the lock *last*, and the status block was quoting it as the current owner. Nothing was broken about the handoff itself, but the one diagnostic a person reaches for while debugging a stuck write was pointing at a PID that no longer existed.

Fixed by asking the kernel: `palace_lock_is_held()` probes ownership on a fresh descriptor and releases immediately, and the status block now carries `palace_locked` from that probe with `holder: null` when the palace is free. Covered by two tests — one at the palace layer (the probe outlives the recorded identity), one at the status layer (a dead holder is not reported).

For what it's worth, the deployed build was also exercised end to end before that: two real MCP servers on one palace, baton moving across and back while both processes stayed alive, and after each handoff both servers were checked for the other's drawer by exact `get_drawer` *and* semantic search — so index visibility, not just write success.

Evgen197310:
Brought up to date with develop (`mergeable` again). Merged rather than rebased so the branch history stays readable mid-review — happy to rebase if you'd rather have it linear.

Two conflicts, both "both sides added something here": the `writer_lease` block now sits next to `library_versions` in the status decoration, and the handoff watchdog starts alongside the write-stall watchdog on both transports.

One change beyond the conflicts, and it is fallout from the merge rather than new scope. `_HTTP_LOCK_FREE_TOOLS` (#2315) narrowed the dispatch lock this PR leans on, and `_dispatch_stdio_request` (#2312) became the single stdio entry point. Keeping the old "lock around the whole stdio request" would have put a five-minute `mempalace_event_wait` long-poll in front of every handoff — the watchdog would find the barrier held for minutes by a call that touches no storage at all — and wrapping the hub forward would have held it across a network round trip. So the locking policy moved into one shared `_dispatch_locally`: lock-free for the logstream set (every handler in it goes through `_call_logstream`, which is also why those tools are exempt from the writer lease), dispatch lock for anything that can reach Chroma or the KG. Local handling inside `_dispatch_stdio_request` — both the no-hub path and the unreachable-hub fallback — goes through it, so a proxied session keeps the barrier for the requests it answers itself. This is the #1984 interaction I flagged in the description, arriving earlier than expected.

Two tests pin both directions. They probe the handoff from the main thread while a request runs on another, because the dispatch lock is reentrant: probing from the dispatching thread would take its own credit and prove nothing.

`docs/writer-lease-handoff.md` records the exemption and what it means for #1984, and gained a short section saying that where you control the deployment, one `mempalace serve` process is a better answer than several direct writers trading a baton (raised in #2305).

Full suite on this branch: **4481 passed, 42 skipped, 4 failed**. All four reproduce on a clean `upstream/develop` checkout in the same environment: three `TestStaleLibraryGate` cases that depend on `chmod 0o000` denying reads (they are run here as root, so the file is readable regardless) and `test_sqlite_exact_read_only_open_sees_active_writer_wal`, the one already noted in the description. `ruff check` / `ruff format --check` clean.

A later commit (`c35d324`) makes the two lease documents match the branch. I read them against the code before promising the `serve` section in #2305, and five things did not match: the holder's steps were printed in the wrong order (the minimum-hold check runs before the demand probe), the contender's back-off after a timed-out wait was undocumented, hub discovery was described as configuration when it is automatic through the per-palace registry, the lock policy said nothing about the forward deliberately taking no lock, and `MIN_HOLD` appeared in code font where no such symbol exists. `docs/write-routing-policy.md` also still said writable storage reopens "after that owner exits" — accurate before this branch, and the one line someone would quote to conclude the handoff does not exist.

Ordering discussion with @jorgeccastro is in #2305 — we agree the two cover different axes, he is content to rebase his timer on top of this, and the demand mark that would give his timer a real gate is now #2340.

Evgen197310:
**Stepping back from this PR — it's takeable as it stands, and nobody should wait on me.**

Our own deployment stopped needing the handoff, so continuing to carry a 1.5k-line branch
through review would mean maintaining something I no longer run. Better to say so plainly
than to go quiet: two people have arranged their work around this PR (#2305's rebase,
@C-LaForest's adoption in #1888), and that is the part I don't want to leave hanging.

**Where the branch actually is**

- `feat/writer-lease-handoff` @ `c35d324`, **6 ahead / 32 behind** `develop`.
- CI green, but as of 23 Aug against that base. GitHub reports `mergeable: CLEAN`, which
  only means there is no textual conflict — it is not evidence the branch still fits
  current `develop`.
- The one place a refresh needs real attention, not a mechanical rebase: the dispatch-lock
  barrier the watchdog leans on. `_HTTP_LOCK_FREE_TOOLS` (#2315) already forced one rework
  of exactly that (the `_dispatch_locally` policy in the 23 Aug merge), and it is the #1984
  interaction flagged in the description. If that lock is narrowed further, the watchdog
  needs to keep *some* barrier against closing storage handles under a live request;
  `docs/writer-lease-handoff.md` says so where it explains the rule.

**What is not done**

1. Refresh onto current `develop` (see above — 32 commits, and @rubicon's note on #2305
   about `tests/test_palace_graph_limits.py` shows how a stale base manufactures failures).
2. @jorgeccastro's three points on the docs section plus the `pgrep` backup gotcha. Handed
   to him on #2305 — the section is his now.
3. @rubicon's phantom-lease constraint (#2172). Answered there, in scope terms, rather than
   left to be read as "the author thinks it's covered": #2131 is a *consumer* of a
   generation/token contract, not an answer to it.

**Taking it over**

MIT. The fork and the branch stay reachable — cherry-pick the commits or reopen the branch as
your own PR; no need to ask me and no need to negotiate credit. `Allow edits by maintainers`
stays on, so anyone with write on this repo can also push to the branch directly; if someone
else wants push access to the fork branch, say so and I'll add you as a collaborator.

I'm deliberately **not** closing this, because #2305 and #1888 are sequenced behind it and a
closed PR reads as "the idea was rejected" rather than "the author moved on". @igorls, my ask
is to leave it open as takeable work; if you'd rather not carry an unowned PR in the queue,
say so and it can be reopened from the branch whenever someone picks it up.

**On the feature itself, so the record isn't ambiguous**

We ran this branch in production from 2 to 17 Aug with real contending peers — the baton
moved across and back between live servers, and after each handoff both servers were checked
for the other's drawer by exact `get_drawer` *and* semantic search, so index consistency and
not merely write success. It did what it says.

Then on 17 Aug we cut over to a single `mempalace serve` process over loopback HTTP (Qdrant
backend, every client on HTTP, no stdio writers left), and the cross-process lease left our
topology altogether — the shape this branch's own doc section recommends as the better answer
where you control the deployment. That is why I'm stepping back, and it is not a verdict on
the PR: for anyone with direct stdio writers, and specifically for @C-LaForest's `sqlite_exact`
fleet pinned at 3.6.0, this is still the zero-reconfig path, and it is finished work sitting
32 commits from being current.

Thanks @igorls for #2158, and @jorgeccastro for the review on #2305 — measuring the reopen
cost instead of arguing about it settled the one open design question in an afternoon.

PR #2135, opened 2026-08-02
fix(sweep): collapse repeated message UUIDs before upsert

### LINKED ISSUE #2112: sweep: a transcript with a repeated message uuid drops the entire file (duplicate IDs in upsert batch)
## Summary

If a transcript contains the same message `uuid` more than once, `mempalace sweep` aborts on that file and **no messages from it are filed at all** — not just the duplicated one. The batch is rejected wholesale by ChromaDB with `Expected IDs to be unique ... in upsert`.

Repeated UUIDs occur naturally in transcripts that are appended to or rewritten in place, which is the same class of file 3.6.0 started re-filing via "Conversation transcripts are treated as mutable" (#1957). So sweep and mine now disagree about which files they can ingest.

## Reproduction

Three records, one `uuid` repeated:

```python
import json
SID = "00000000-0000-0000-0000-00000000aaaa"
DUP = "11111111-1111-1111-1111-111111111111"
recs = [
  {"type":"user","uuid":DUP,"sessionId":SID,"timestamp":"2020-01-01T00:00:01.000Z",
   "message":{"role":"user","content":"first copy"}},
  {"type":"user","uuid":DUP,"sessionId":SID,"timestamp":"2020-01-01T00:00:02.000Z",
   "message":{"role":"user","content":"second copy, same uuid"}},
  {"type":"assistant","uuid":"22222222-2222-2222-2222-222222222222","sessionId":SID,
   "timestamp":"2020-01-01T00:00:03.000Z","message":{"role":"assistant","content":"ok"}},
]
open("/tmp/mp-dup-repro/transcripts/session.jsonl","w").write(
    "".join(json.dumps(r)+"\n" for r in recs))
```

```
mempalace --palace /tmp/mp-dup-repro/palace sweep /tmp/mp-dup-repro/transcripts
```

## Actual

```
sweeper: existence pre-check failed (Expected IDs to be unique, found duplicates of:
  sweep_00000000-0000-0000-0000-00000000aaaa_11111111-1111-1111-1111-111111111111 in get.);
  counting all batch rows as new (metric may over-count on reruns).
sweeper: sweep failed on /tmp/mp-dup-repro/transcripts/session.jsonl: Expected IDs to be
  unique, found duplicates of: sweep_00000000-...-aaaa_11111111-...-1111 in upsert.
  WARNING: 1 file(s) failed to sweep - see stderr / logs for details.
  Swept 0/1 files: +0 new, 0 already present, 0 skipped (< cursor).
```

`Swept 0/1` — the third message has a unique ID and is still dropped, because the whole batch fails.

## Expected

The file is swept; the duplicated ID collapses to one drawer (last-wins, consistent with the deterministic-ID/idempotency intent). At worst, skip the offending message — not the entire transcript.

## Root cause

`mempalace/sweeper.py`:

- `_drawer_id_for_message()` (~L183-190) returns `f"sweep_{session_id}_{message_uuid}"`. Its docstring says *"Deterministic drawer ID so upserts at the same message are no-ops"* — that gives idempotency **across runs**, but nothing collapses duplicates **within one batch**.
- The message loop (~L272-284) does `batch_ids.append(drawer_id)` with no uniqueness check.
- `_flush()` then calls `collection.get(ids=list(batch_ids), ...)` (~L237) and `collection.upsert(ids=batch_ids, ...)` (~L251). ChromaDB rejects duplicate IDs in both calls.

Two distinct symptoms fall out:

1. **The `get()` failure is caught but degrades the metric** — it logs `counting all batch rows as new (metric may over-count on reruns)`, so `+N new` is silently unreliable whenever this fires.
2. **The `upsert()` failure is fatal for the file.**

## Impact

One repeated UUID anywhere in a transcript costs every per-message drawer for that entire session. It is also deterministic — re-running `sweep` cannot recover the file, so the gap is permanent until the transcript changes or the bug is fixed. Mining still covers the file at chunk granularity, so this shows up as missing per-message recall rather than total absence, which makes it easy to miss.

## Suggested fix

De-duplicate by ID at append time, keeping the last occurrence, e.g. accumulate the batch in a dict keyed by `drawer_id` (or drop a record whose ID is already staged) before `get()`/`upsert()`. Last-wins matches the existing cross-run upsert semantics.

Worth applying the same guard in `_flush()` itself so any future caller is protected.

## Environment

- MemPalace 3.6.0
- chromadb 1.5.9
- Python 3.10, Linux

fatkobra:
## What does this PR do?

- collapse repeated deterministic drawer IDs before the sweeper's backend calls
- send only unique IDs to both the existence check and upsert
- use explicit last-record-wins behavior for repeated message UUIDs
- calculate added, already-present, and upserted counters from the unique rows actually processed
- preserve unrelated messages from the same transcript batch

## Root cause

`sweep()` stages one row for every transcript message, while drawer IDs are deterministic from `session_id + message_uuid`.

When a transcript contains the same UUID more than once, the batch contains duplicate IDs.

ChromaDB rejects duplicate IDs in both:

- `collection.get(ids=...)`
- `collection.upsert(ids=...)`

The existence-check failure was caught and converted into an inaccurate metric, but the later upsert failure aborted the complete file. As a result, unrelated valid messages in the same batch were also lost.

## Implementation

`_flush()` now collapses the staged rows into a dictionary keyed by deterministic drawer ID before either backend call.

For repeated IDs:

- the key retains its original insertion position
- the final occurrence supplies the document and metadata
- only one row for that ID is sent to the backend

This provides last-record-wins behavior consistent with the existing deterministic-ID and upsert semantics.

The returned counters are calculated from the unique rows actually sent to the backend.

## Tests

Adds the reported three-record reproduction:

- two user records share one UUID
- the second user record contains updated content and timestamp
- one assistant record has a distinct UUID

The regression verifies that:

- the sweep completes successfully
- two unique drawers are written
- the repeated UUID stores the final record
- the final record's timestamp is preserved
- the unrelated assistant message is preserved
- returned counters describe the two unique writes

Closes #2112

## How to test

- `python3 -m pytest tests/test_sweeper.py -k "collapses_repeated_uuid_last_record_wins" -q`
- `python3 -m pytest tests/test_sweeper.py -q`
- `python3 -m pytest tests/test_sweeper.py tests/test_cli.py -q`
- `python3 -m ruff check .`

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/sweeper.py
+++ b/mempalace/sweeper.py
@@ -230,28 +230,48 @@ def sweep(jsonl_path: str, palace_path: str, source_label: Optional[str] = None)
         nonlocal drawers_added, drawers_already_present
         if not batch_ids:
             return
-        # Pre-flight: which IDs in this batch are already present?
-        # Upsert is idempotent on data but counts as "added" would lie;
-        # this pre-query makes the metric honest (Copilot PR 998 review).
+        # Collapse repeated deterministic IDs before backend calls.
+        # Rewritten transcripts can contain the same message UUID more than
+        # once, and Chroma rejects duplicate IDs in one get or upsert request.
+        # Reassigning a dict key preserves its first position but replaces its
+        # value, so the final occurrence supplies the document and metadata.
+        deduped: dict[str, tuple[str, dict]] = {}
+        for drawer_id, document, metadata in zip(
+            batch_ids,
+            batch_docs,
+            batch_metas,
+        ):
+            deduped[drawer_id] = (document, metadata)
+
+        unique_ids = list(deduped)
+        unique_docs = [deduped[drawer_id][0] for drawer_id in unique_ids]
+        unique_metas = [deduped[drawer_id][1] for drawer_id in unique_ids]
+
+        # Pre-flight: which unique IDs are already present?
         try:
-            existing = collection.get(ids=list(batch_ids), include=[])
-            # Chroma returns a dict; typed backends return GetResult — the
-            # compat shim makes ``.get("ids")`` work on both.
+            existing = collection.get(
+                ids=unique_ids,
+                include=[],
+            )
+            # Chroma returns a dict; typed backends return GetResult - the
+            # compatibility shim makes .get("ids") work on both.
             present = set(existing.get("ids") or [])
         except Exception as exc:
             logger.warning(
                 "sweeper: existence pre-check failed (%s); "
-                "counting all batch rows as new (metric may over-count on reruns).",
+                "counting all unique batch rows as new "
+                "(metric may over-count on reruns).",
                 exc,
             )
             present = set()
-        new_count = sum(1 for rid in batch_ids if rid not in present)
-        already_count = len(batch_ids) - new_count
+
+        new_count = sum(1 for drawer_id in unique_ids if drawer_id not in present)
+        already_count = len(unique_ids) - new_count
 
         collection.upsert(
-            ids=batch_ids,
-            documents=batch_docs,
-            metadatas=batch_metas,
+            ids=unique_ids,
+            documents=unique_docs,
+            metadatas=unique_metas,
         )
         drawers_added += new_count
         drawers_already_present += already_count
--- a/tests/test_sweeper.py
+++ b/tests/test_sweeper.py
@@ -316,3 +316,104 @@ class TestSweeperDrawerMetadata:
                 "user",
                 "assistant",
             ), f"Drawer missing or wrong role metadata: {m}"
+
+
+class TestSweeperDuplicateMessageIds:
+    def test_sweep_collapses_repeated_uuid_last_record_wins(
+        self,
+        tmp_path,
+    ):
+        from mempalace.palace import get_collection
+        from mempalace.sweeper import _drawer_id_for_message, sweep
+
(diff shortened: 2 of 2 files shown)


fatkobra:
Windows CI appears to have hit an unrelated transient ChromaDB/HNSW failure.

The only failing test was:

`tests/test_closets.py::TestDrawerGrepExpansion::test_hybrid_search_enrichment_isolates_chunks_across_drawers_sharing_source_file`

It failed while opening the test palace's HNSW segment:

This PR changes only `mempalace/sweeper.py` and `tests/test_sweeper.py`. The new Windows regression for this PR passed later in the same run:

`tests/test_sweeper.py::TestSweeperDuplicateMessageIds::test_sweep_collapses_repeated_uuid_last_record_wins PASSED`

The Windows run otherwise completed with 3249 passed, 280 skipped, and 80.97% coverage.

The workflow has `--reruns 2`, but its `--only-rerun` filter currently matches only:

`Failed to apply logs to the hnsw segment writer`

The observed `Nothing found on disk` HNSW variant therefore was not automatically retried.

@igorls could you please rerun the failed Windows job?

igorls (approved):

PR #2141, opened 2026-08-03
fix(chroma): persist EF-free caller-vector schema

### LINKED ISSUE #2121: Caller-supplied vectors: embedding_function=None still embeds, and an EF-less schema cannot coexist with metadata=
#1261 asks for a configurable embedding function. The neighbouring case — a caller that already holds vectors and needs the store to embed nothing — hits two blockers on chromadb 1.5.9.

**`embedding_function=None` does not disable embedding.** chromadb carries two independent functions: the client-level one, and `VectorIndexConfig.embedding_function`, defaulting to a live `DefaultEmbeddingFunction()`. `CollectionCommon._embed()` falls through client → configuration → schema. So `create_collection(name, embedding_function=None)` then `add(documents=...)` stores a 384-dim ONNX embedding, silently. On a collection already holding vectors of another width it raises a dimension error rather than refusing.

**The working shape collides with the HNSW metadata.** Creating with `schema=Schema().create_index(config=VectorIndexConfig(embedding_function=None))` behaves correctly — misuse raises `ValueError: You must provide an embedding function…`, caller vectors land, the shape survives reopen. But `create_collection` rejects `metadata=` and `schema=` together ("Cannot set both collection config and schema simultaneously"), and `backends/chroma.py` passes `metadata=` for `hnsw:space` and the write thresholds.

So a caller-vector collection must give up the HNSW metadata, or those parameters move into the schema — out of the `collection_metadata` table that `_read_sync_threshold` and `distance_metric` read.

Happy to build whichever direction you prefer.

fatkobra:
## What does this PR do?

- create caller-vector Chroma collections with an embedding-function-free schema
- explicitly pass `embedding_function=None` during create and reopen
- move cosine distance and HNSW write settings from legacy metadata into the schema
- fail closed when the client, configuration, or live schema still exposes an embedding function
- retain MemPalace-side guards for adds, upserts, document updates, and text queries
- read persisted `sync_threshold` from either legacy metadata or `schema_str`
- read the actual distance metric from either legacy metadata or the live schema
- add real persistent create/reopen regression coverage

## Root cause

ChromaDB has multiple embedding-function sources:

- the client-level collection embedding function;
- the collection configuration embedding function;
- `VectorIndexConfig.embedding_function` in the collection schema.

The previous revision avoided resolving MemPalace's local embedding function, but it omitted the `embedding_function` keyword when the selected value was `None`.

Chroma consequently attached `DefaultEmbeddingFunction`.

After a three-dimensional caller vector was stored, a raw document-only write attempted to produce a 384-dimensional default embedding and failed with a dimension mismatch.

The schema-level default also survived a fresh reopen.

## Red-before reproduction

Before this revision, a real persistent collection reported:

- client embedding function: `DefaultEmbeddingFunction`;
- configuration embedding function: `DefaultEmbeddingFunction`;
- schema embedding function: `DefaultEmbeddingFunction`.

A document-only raw write attempted embedding and failed because the collection expected dimension 3 but Chroma generated dimension 384.

After reopening the collection with the client-level embedding function explicitly set to `None`, the persisted schema embedding function was still `DefaultEmbeddingFunction`, and the same document-only write again attempted embedding.

## Implementation

Caller-vector creation now uses:

- `Schema().create_index(...)`;
- `VectorIndexConfig(embedding_function=None, ...)`;
- `embedding_function=None` explicitly on the client call.

The HNSW settings are supplied through `HnswIndexConfig`:

- distance space;
- number of threads;
- batch size;
- sync threshold;
- optional construction EF;
- optional maximum neighbors.

Legacy non-caller-vector collection creation continues to use the existing metadata path unchanged.

## Runtime verification

After every caller-vector create or reopen, MemPalace verifies all three live Chroma sources.

The collection is rejected when:

- the client embedding function is active;
- the configuration embedding function is active;
- the schema embedding function is active;
- any source cannot be verified.

This prevents an existing default-EF collection from being silently reopened as a caller-vector collection.

## Persisted HNSW settings

The persistence diagnostic showed that ChromaDB 1.5.9 stores schema-created HNSW settings in:

`collections.schema_str`

while:

`collections.config_json_str`

remains empty.

The divergence guard now reads `sync_threshold` from:

1. legacy `collection_metadata`;
2. the `#embedding` HNSW section in `schema_str`;
3. Chroma's default of 1000 when neither is available.

The explicit-threshold check recognizes both storage layouts.

## Distance metric

Legacy collections continue to expose their distance metric through `hnsw:space` metadata.

Schema-created caller-vector collections expose it through the live `#embedding` vector configuration.

`ChromaCollection.distance_metric` now checks both sources before falling back to L2.

## Schema serialization note

Chroma's serialized JSON represents an EF-free schema embedding function as a legacy descriptor, while the live `VectorIndexConfig.embedding_function` property is `None`.

The implementation therefore does not infer embedding-function safety from the serialized JSON field.

Runtime embedding-function validation uses the live Chroma objects.

`schema_str` is parsed only for persisted HNSW settings.

## Tests

Regression coverage verifies that:

- wrapper adds and upserts require explicit embeddings;
- document updates require explicit embeddings;
- metadata-only updates remain allowed;
- text queries require explicit query embeddings;
- the schema contains no live embedding function;
- caller HNSW options are retained;
- a real persistent collection is EF-free at the client, configuration, and schema levels;
- a raw document-only write is refused before and after reopen;
- explicit three-dimensional vectors can be stored and queried;
- cosine distance survives reopen;
- the schema-backed sync threshold is visible to the divergence guard;
- an existing default-EF collection is rejected in caller-vector mode.

The real path was also exercised manually across separate processes:

1. create the collection through `ChromaBackend`;
2. verify all three embedding-function sources are `None`;
3. store a caller-supplied vector;
4. verify a raw document-only write is refused;
5. close the backend;
6. reopen through a new process;
7. verify the EF-free schema and HNSW settings;
8. query the caller vector;
9. verify document-only writes remain refused.

Closes #2121

## How to test

- `python3 -m pytest tests/test_chroma_caller_vectors.py -q`
- `python3 -m pytest tests/test_chroma_caller_vectors.py tests/test_backends.py tests/test_distance_metric.py tests/test_hnsw_capacity.py tests/test_collection_metric_invariant.py -q`
- `python3 -m ruff check .`

## Checklist
- [x] No hardcoded paths


fatkobra:
Thanks @mvalentsev  for checking this against a live ChromaDB collection.

You are right: the current PR adds a MemPalace wrapper guard, but it does not remove Chroma's schema-level `DefaultEmbeddingFunction`. Therefore it does not close #2121 as currently written.

I am converting this PR to draft and removing the assumption that omitting the client-level embedding function is sufficient.

A corrected implementation needs a live persistent-Chroma regression that proves all of the following:

- the collection schema itself has no embedding function;
- a document-only write fails instead of silently producing a default embedding;
- caller-supplied vectors succeed;
- the collection survives close and reopen with the same behavior;
- cosine distance and the required HNSW write settings remain available after moving configuration into the schema.

Until that test passes against the supported Chroma version, this PR should not claim to close #2121.

fatkobra:
Thanks — I reproduced the live failure before replacing the implementation.

The previous revision produced:

- client EF: `DefaultEmbeddingFunction`;
- configuration EF: `DefaultEmbeddingFunction`;
- schema EF: `DefaultEmbeddingFunction`.

After storing a three-dimensional caller vector, a raw document-only write attempted to create a 384-dimensional default embedding and failed with a dimension mismatch.

The schema-level default survived a fresh reopen even when the client was reopened with `embedding_function=None`.

The revised PR now:

- creates caller-vector collections with `Schema` and `VectorIndexConfig(embedding_function=None)`;
- explicitly passes `embedding_function=None` during create and reopen;
- moves cosine and HNSW write settings into `HnswIndexConfig`;
- verifies the live client, configuration, and schema EF sources after every open;
- rejects legacy default-EF collections rather than treating them as caller-vector collections;
- reads schema-backed `sync_threshold` from `collections.schema_str`;
- reads the distance metric from the live schema when legacy metadata is absent;
- retains the wrapper-level explicit-vector guards as defense in depth.

One persistence nuance came from the end-to-end diagnostic: Chroma serializes the EF-free schema field as a legacy descriptor in `schema_str`, even though the live `VectorIndexConfig.embedding_function` is `None`.

The implementation therefore validates EF safety only through the live Chroma objects and parses `schema_str` only for HNSW settings.

The regression now uses a real persistent collection, closes it, reopens it, queries an explicit caller vector, and verifies that raw document-only writes are refused both before and after reopen.

I also exercised the path manually across two separate Python processes with the persisted sync threshold verified as 512.

fatkobra:
@igorls rebased this branch onto current `develop` and resolved the Chroma
backend conflicts.

Both original commits were preserved:

- `fix(chroma): enforce caller-supplied vector mode`
- `fix(chroma): persist EF-free caller-vector schema`

The rebased implementation preserves current upstream Chroma safety and
lifecycle behavior, including:

- namespace validation;
- HNSW capacity and divergence detection;
- current sync-threshold and stale-index handling;
- process-wide Chroma system-cache invalidation after external disk changes;
- client-cache freshness and invalidation;
- quarantine and recovery behavior;
- read-only SQLite inspection;
- local write locking;
- current backend capability declarations.

It also retains the caller-vector contract:

- caller-vector collections are created with an embedding-function-free
  schema;
- `embedding_function=None` is passed explicitly during creation and reopen;
- client, configuration, and live-schema embedding-function sources are
  verified after every open;
- active or unverifiable embedding-function sources fail closed;
- add, upsert, document-update, and text-query paths require explicit
  vectors where appropriate;
- metadata-only updates remain supported;
- schema-backed HNSW settings and distance metrics survive persistent
  close/reopen cycles.

Validation completed:

- repository-wide Ruff formatting and lint checks;
- the caller-vector, backend, distance-metric, HNSW-capacity, collection
  invariant, and current Chroma safety suites;
- the complete CI-shaped test suite with coverage;
- GitHub CI is green.

Both commits were rebased onto current `develop`, and the rewritten history
was pushed with `--force-with-lease`.

fatkobra:
@igorls 
Rebased both commits onto current `develop` and resolved the `mempalace/backends/chroma.py` conflicts semantically.

The resolution preserves current upstream behavior, including:

- namespace validation
- backend client-freshness restamping after collection opens
- backend-aware `ChromaCollection` wrappers
- write locking and post-write freshness tracking
- current HNSW safety, quarantine, and capacity behavior

It also retains the complete caller-vector contract:

- explicit-vector guards for add, upsert, document update, and text query
- explicit `embedding_function=None` during caller-vector create and reopen
- EF-free `Schema` / `VectorIndexConfig` creation
- fail-closed validation of client, configuration, and live-schema embedding-function sources
- schema-backed `sync_threshold` and distance-metric reads
- persistent create/reopen regression coverage

Both original commits remain separate.

Local validation passed:

- `python3 -m ruff format --check mempalace/backends/chroma.py tests/test_chroma_caller_vectors.py`
- `python3 -m ruff check mempalace/backends/chroma.py tests/test_chroma_caller_vectors.py`
- `python3 -m pytest tests/test_chroma_caller_vectors.py tests/test_backends.py tests/test_distance_metric.py tests/test_hnsw_capacity.py tests/test_collection_metric_invariant.py -q`

The rewritten branch was pushed with `--force-with-lease`.

PR #2143, opened 2026-08-03
fix(search): return distinct passages after closet enrichment

### LINKED ISSUE #2052: Mining files near-identical duplicate chunks into the palace within a single pass (docs + code)
## Environment
- mempalace 3.6.0 (latest on PyPI as of 2026-07-20 — confirmed via `pip index versions mempalace`)
- Windows 11, chromadb backend
- Ingest mode: `projects` (default)

## Summary
Mining a project directory files the *same text* into many drawers within a single mining pass (all created within tens of milliseconds of each other), rather than producing distinct chunks. This happens for both markdown docs and C# source files. It's not chunk-boundary overlap — it's the literal same paragraph/code block repeated as separate drawers.

## Repro data (measured via `mempalace_search` scoped to `source_file`, comparing `total_before_filter` for the file against how many top results are the same text)

| File | Total drawers for file | Duplicate drawers | Ratio |
|---|---|---|---|
| `docs/Calendar-Types-FE-Guide.md` | 48 | 20 | 42% |
| `docs/architecture.md` | 13 | 10 | 77% |
| `docs/Layer-Groups-Hierarchy-FE-Guide.md` | 14 | ~13 | ~93-100% |
| `src/.../LayerService.cs` (C#) | 33 | 30 | 91% |

For each of these, grepping the source file confirmed the duplicated paragraph/section appears **exactly once** in the actual source — the duplication is introduced entirely at mining time, not by the source content.

Example (`LayerService.cs`, `room:src`): querying `"LayerService update raster viewer config validation"` with `source_file` scoped to the exact path, `limit: 10`, `max_distance: 0.8` returns 10 results that are all the *same* `AddLayerAsync` switch-statement chunk, with `created_at` timestamps spanning `2026-07-20T22:45:31.208457` to `.228504` (~20ms apart) — i.e. filed repeatedly in one mining call, not across separate mining runs.

## Impact
1. **Token cost**: scoping search to `room`/`source_file` and tightening `max_distance` doesn't help — the duplicate drawers sit at genuinely low cosine distance from any on-topic query, so they dominate results regardless of threshold.
2. **Information loss**: for the most-affected files (Layer-Groups-Hierarchy-FE-Guide.md, LayerService.cs), the overwhelming majority of a file's drawer budget goes to one repeated chunk, so later sections/methods in the file may not be represented in the palace at all.

## Expected
Each mining pass over a file should produce distinct chunks covering the file's content, with no (or minimal, content-hash-deduped) exact-text repeats within the same pass.

## Not yet tried
- Did not test whether `--max-chunks-per-file` or a fresh `mempalace init --redetect-origin` changes this.
- Only sampled 4 files across `docs/` and `src/`; ratio could be higher or lower elsewhere in a larger corpus.

### LINKED ISSUE #2083: search_memories returns the same passage several times when hits come through closets
### Summary

When a query's hits arrive through closets, `search_memories` returns the same passage several times. Asking for 5 results can give 5 copies of one chunk, so a search for five things shows one.

Measured over eight real questions against a palace of 35 910 drawers: **40 results carried 23 distinct passages (57%)**. The split is clean:

| hits of that query | distinct out of 5 |
|---|---|
| all `drawer+closet` | 1 |
| all plain `drawer`  | 5 |

Language of the query made no difference; `matched_via` did.

### Cause

`searcher.py`, the drawer-grep enrichment pass (v3.5.0, around lines 1222-1278). For every closet-boosted hit, the hit's text is replaced by the keyword-best chunk of that source file plus its neighbours:

```python
query_terms = set(_tokenize(query))
best_idx, best_score = 0, -1
for idx, d in enumerate(ordered_docs):
    s = sum(1 for t in query_terms if t in d.lower())
    if s > best_score:
        best_score, best_idx = s, idx
...
h["text"] = expanded
```

`best_idx` is scored against the query alone, so it is identical for every hit from the same file. Five drawers of one file therefore end up carrying byte-identical text, and nothing deduplicates after the substitution. `hits` was already cut to `n_results` before enrichment, so there is nothing left to promote in their place either.

### Reproduce

Any file with several chunks that a closet points at. Here, `drift-reminder.py` (11 chunks):

```python
r = search_memories(q, palace_path=PALACE, wing="code", n_results=5,
                    collection_name="mempalace_drawers")
print(len({h["text"] for h in r["results"]}))   # 1
```

The five results differ in `distance` and `created_at`, and are genuinely five different drawers; only the rendered text is the same one.

### What worked here

Locally, keeping ranking untouched and changing only what survives to the caller:

1. take a wider pool before enrichment (`n_results * 4`) instead of cutting to `n_results` at that point,
2. memoise the per-source `drawers_col.get` for the duration of one search, so the wider pool costs no extra queries,
3. after enrichment, drop hits whose `(source_file, text)` is already present and cut back to `n_results`.

Result on the same eight questions: 38 results, 38 distinct. Latency unchanged at 0.16-0.20 s per query (the memoisation pays for the wider pool), first result and its similarity identical in every case.

Happy to send this as a PR if the approach looks right.

fatkobra:
## What does this PR do?

- remove the speculative miner-side content deduplication
- keep a wider vector candidate pool until closet enrichment completes
- memoise drawer retrieval by full source path and parent drawer group
- preserve the existing closet hydration and final hybrid ranking
- remove repeated rendered passages after final ranking
- retain the first-ranked occurrence
- fill the requested result set from remaining distinct candidates
- preserve plain drawer repeats and all stored source content

## Root cause

The original report in #2052 was measured through `search_memories()` and showed different drawer IDs and timestamps carrying the same passage.

A two-version real-path investigation traced the same corpus through MemPalace v3.6.0 and current `develop`.

For both versions:

- `chunk_text()` produced distinct chunks;
- the drawer upsert payload was distinct;
- the raw Chroma collection was distinct;
- direct SQLite extraction was distinct;
- direct vector-query results were distinct;
- repeated text first appeared in the final `search_memories()` response.

The duplication is introduced during closet enrichment.

Every closet-boosted hit from one source independently selects the same query-best source chunk and neighbours. The hit keeps its original ID, timestamp, distance, and ranking metadata, but its displayed `text` is replaced with the same expanded passage.

The old implementation had already cut the pool to `n_results`, leaving no unused candidates to promote after those repeated rendered passages were removed.

This is the cause described independently in #2083.

## Why the miner patch was removed

The previous revision deleted any later chunk whose complete text matched an earlier chunk.

That assumption was not supported by the real-path investigation.

It was also unsafe: the real chunker can legitimately emit the same legal notice, code block, or paragraph from two different source positions. Content-only deduplication removed the later valid occurrence.

This revision does not modify mining, stored documents, drawer IDs, or source-position metadata.

## Implementation

The vector path now retains up to:

`n_results * 4`

candidates through closet enrichment.

Drawer retrieval for enrichment is memoised by:

`(full_source_path, parent_drawer_id)`

The parent component preserves the existing #1580 isolation between unrelated logical drawers that share one `source_file`.

After the existing final hybrid rank, results are deduplicated by:

`(full_source_path, rendered_text)`

A duplicate is removed only when either occurrence came through closet enrichment.

Plain drawer results retain their historical behavior, including legitimate exact repeats at different source positions.

The first-ranked occurrence is preserved, and the result list is cut to `n_results` only after rendered-result deduplication.

## Result-count semantics

`n_results` remains a maximum.

When the wider pool contains enough distinct candidates, duplicate enriched hits are replaced by lower-ranked distinct results.

When no additional distinct passage exists, the response may contain fewer than `n_results` entries rather than padding the response with repeated copies.

## Investigation evidence

The comparison covered:

- MemPalace v3.6.0, the version reported in #2052;
- current `develop`;
- Markdown;
- CRLF C#;
- BOM-prefixed Markdown;
- long-line boundary cases;
- an intentional exact-repeat control;
- two unchanged mining passes;
- raw Chroma and direct SQLite inspection.

Across the audited corpus, the miner and raw storage stayed unique. The same source-level passage was duplicated only after closet enrichment.

## Candidate-pool compatibility

For the default `candidate_strategy="vector"` path, the vector pool remains four times wider through closet enrichment and rendered-result deduplication.

The opt-in `candidate_strategy="union"` path preserves its historical candidate contract:

- query `n_results * 3` vector candidates;
- retain the requested top `n_results` vector candidates;
- merge backend lexical candidates;
- perform the final hybrid rank.

This prevents the closet-diversity change from altering lexical-candidate provenance or the existing `matched_via="bm25_backend"` behavior.

## Tests

Regression coverage verifies that:

- the first-ranked closet-rendered occurrence is preserved;
- repeated closet-rendered passages are removed;
- plain drawer repeats remain untouched;
- identical basenames under different full paths do not collide;
- one source/group fetch is reused for repeated hits;
- different parent drawer groups remain isolated;
- the vector query requests a four-times wider pool;
- distinct lower-ranked candidates refill the requested result count;
- `process_file()` preserves legitimate exact repeats from different source positions;
- a real Chroma collection contains unique raw IDs and documents;
- a real source-scoped closet search returns no duplicate rendered passages.

Closes #2052

Closes #2083

## How to test

- `python3 -m pytest tests/test_search_distinct_closet_results.py -q`
- `python3 -m pytest tests/test_search_distinct_closet_results.py tests/test_searcher.py tests/test_closets.py tests/test_hybrid_search.py tests/test_hybrid_candidate_union.py tests/test_search_drawer_id.py tests/test_miner.py -q`
- `python3 -m ruff check .`

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -1198,6 +1198,238 @@ def _candidate_pool_size(n_results: int, date_window_active: bool) -> int:
     return max(min(n_results * 15, 500), n_results)
 
 
+_CLOSET_RESULT_POOL_MULTIPLIER = 4
+_MAX_HYDRATION_CHARS = 10000
+
+
+def _candidate_pool_limits(
+    candidate_strategy: str,
+    n_results: int,
+    date_window_active: bool = False,
+) -> tuple[int, int]:
+    """Return vector-query and pre-enrichment candidate limits.
+
+    Date-window searches retain the wider current-develop pool because the
+    timestamp constraint is applied after retrieval. The normal vector path
+    also remains at least four times wide through closet enrichment. Union
+    mode keeps only the requested top vector candidates before lexical merge.
+    """
+    base_query_limit = _candidate_pool_size(
+        n_results,
+        date_window_active,
+    )
+
+    if candidate_strategy == "union":
+        return base_query_limit, n_results
+
+    widened = max(
+        base_query_limit,
+        n_results * _CLOSET_RESULT_POOL_MULTIPLIER,
+    )
+    return widened, widened
+
+
+def _enrich_closet_hits(
+    hits: list,
+    drawers_col,
+    query: str,
+    stop_words: frozenset = frozenset(),
+) -> list:
+    """Hydrate closet-boosted hits and memoise each source/group fetch."""
+    query_terms = set(
+        _tokenize(
+            query,
+            stop_words,
+        )
+    )
+    source_cache: dict = {}
+
+    for hit in hits:
+        if hit.get("matched_via") == "drawer":
+            continue
+
+        full_source = hit.get("_source_file_full") or ""
+
+        if not full_source:
+            continue
+
+        parent_drawer_id = hit.get("_parent_drawer_id") or None
+        cache_key = (
+            full_source,
+            parent_drawer_id,
+        )
+
+        if cache_key not in source_cache:
+            try:
+                source_drawers = drawers_col.get(
+                    where=(
+                        _scoped_source_filter(
+                            full_source,
+                            parent_drawer_id,
+                        )
+                    ),
+                    include=[
+                        "documents",
+                        "metadatas",
+                    ],
+                )
+            except Exception:
+                logger.debug(
+                    "Neighbor fetch failed for %s",
+                    full_source,
+                    exc_info=True,
+                )
+                source_cache[cache_key] = None
+            else:
+                source_cache[cache_key] = (
+                    list(
+                        getattr(
+                            source_drawers,
+                            "documents",
+                            None,
+                        )
+                        or []
+                    ),
+                    list(
+                        getattr(
+                            source_drawers,
+                            "metadatas",
+                            None,
+                        )
+                        or []
+                    ),
+                )
+
+        cached = source_cache[cache_key]
+
+        if cached is None:
+            continue
+
+        docs, metadatas = cached
+
+        if len(docs) <= 1:
+            continue
+
+        indexed = []
+
+        for index, (
+            document,
+            metadata,
+        ) in enumerate(
+            zip(
+                docs,
+                metadatas,
+            )
+        ):
+            chunk_index = (
+                metadata.get(
+                    "chunk_index",
+                    index,
+                )
+                if isinstance(
+                    metadata,
+                    dict,
+                )
+                else index
+            )
+
+            if not isinstance(
+                chunk_index,
+                int,
+            ):
+                chunk_index = index
+
+            indexed.append(
+                (
+                    chunk_index,
+                    document or "",
+                )
+            )
+
+        indexed.sort(key=lambda pair: pair[0])
+        ordered_docs = [document for _, document in indexed]
+
+        best_index = 0
+        best_score = -1
+
+        for index, document in enumerate(ordered_docs):
+            lowered = document.lower()
+            score = sum(1 for term in query_terms if term in lowered)
+
+            if score > best_score:
+                best_score = score
+                best_index = index
+
+        start = max(
+            0,
+            best_index - 1,
+        )
+        end = min(
+            len(ordered_docs),
+            best_index + 2,
+        )
+        expanded = "\n\n".join(ordered_docs[start:end])
+
+        if len(expanded) > _MAX_HYDRATION_CHARS:
+            expanded = expanded[:_MAX_HYDRATION_CHARS] + (
+                f"\n\n[...truncated. "
+                f"{len(ordered_docs)} total drawers. "
+                "Use mempalace_get_drawer "
+                "for full content.]"
+            )
+
+        hit["text"] = expanded
+        hit["drawer_index"] = best_index
+        hit["total_drawers"] = len(ordered_docs)
+
+    return hits
+
+
+def _dedupe_rendered_hits(
+    hits: list,
+) -> list:
+    """Drop repeated closet-rendered passages while preserving rank order.
+
+    Plain drawer hits retain their historical behavior, including legitimate
+    exact repeats at different source positions. A duplicate is removed only
+    when either occurrence came through closet enrichment.
+    """
+    unique = []
+    first_by_key: dict = {}
+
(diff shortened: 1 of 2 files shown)


fatkobra:
CI found one real compatibility regression in the first search-layer revision.

The four-times wider vector pool had been applied to every candidate strategy. In `candidate_strategy="union"` mode, that allowed a lexical target to enter the initial vector pool before backend lexical merge. The lexical copy was then correctly deduplicated by source/chunk identity, but the surviving hit retained `matched_via="drawer"` instead of the historical `matched_via="bm25_backend"` provenance.

The correction now scopes the four-times wider pool to the default vector/closet path.

Union mode preserves its existing contract:

- vector query: `n_results * 3`;
- pre-merge vector pool: `n_results`;
- backend lexical merge;
- final hybrid rank.

The existing sqlite-exact lexical-union regression and a new pool-limit regression now cover both strategies.

The Windows failure in the previous run was the separate intermittent Chroma error:

The deterministic candidate-strategy regression has been corrected in this push.

fatkobra:
@igorls rebased this branch onto current `develop` and resolved the
`searcher.py` conflict.

The rebased implementation preserves current upstream search behavior,
including:

- `source_file`, `since`, and `before` filtering;
- date-window candidate over-fetching and truncation reporting;
- round-trippable drawer IDs, including chunked diary entries;
- current vector-disabled and backend-lexical fallback behavior;
- the existing search error and result envelope.

It also retains this PR's closet-result diversity fix:

- vector-mode candidates remain wide through closet enrichment;
- date-window searches retain their wider current-develop retrieval pool;
- union mode preserves its pre-merge vector limit and lexical provenance;
- source/group drawer fetches are cached for the duration of one search;
- duplicate rendered passages are removed after final hybrid ranking;
- the first-ranked occurrence is preserved;
- lower-ranked distinct candidates refill the requested result count;
- plain drawer repeats remain untouched.

Validation completed:

- repository-wide Ruff formatting and lint checks;
- the focused closet-diversity, search, hybrid, drawer-ID, closet, and miner
  suites;
- the complete CI-shaped test suite with coverage;
- GitHub CI is green.

The rebased history was pushed with `--force-with-lease`.

PR #2148, opened 2026-08-03
fix(windows): safely repair legacy mojibake with backups

### LINKED ISSUE #1055: [Windows] Encoding repair script for palaces mined before v3.1.0
## Problem

Users who mined content on Windows using mempalace versions before 3.1.0 
end up with corrupted text in their drawers. Characters like `é`, `ó`, `ñ`, 
`—`, `→` are stored as mojibake sequences like `Ã©`, `Ã³`, `Ã±`, `â€"`, `â†'`.

This happens because Python on Windows defaults to CP1252 when reading files, 
causing a latin-1 → UTF-8 double-encoding that corrupts any non-ASCII character.

The v3.1.0 release fixed this for new content (#129, #281), but existing 
palaces are not repaired automatically.

## Who is affected

Any Windows user who:
- Installed mempalace before v3.1.0
- Mined conversations or project files containing non-ASCII characters

This affects Spanish, French, German, Portuguese, Polish, and any other 
language using characters outside basic ASCII. It is not a Spanish-only issue.

## Solution

I wrote a repair script that:
1. Connects directly to your local ChromaDB palace
2. Detects drawers with corrupted encoding
3. Repairs the text using a character-by-character sliding window algorithm 
   that handles mixed damage patterns (some drawers have multiple corruption 
   types in the same text)
4. Updates each drawer in place using `col.update()` — no data is deleted

The script includes a `--dry-run` mode to preview all changes before applying them.

Tested on:
- Windows 10 Home
- Python 3.12
- ChromaDB 1.5.7
- mempalace 3.3.1
- Palace with 39 drawers — 0 errors after repair
- Palace content in Spanish (easily reproducible with any non-ASCII language)

## Script

👉 [mempalace_repair_encoding.py](https://gist.github.com/Cre3b3ep/f5d78005bf952f745a0f0eff030d9d81)

## Usage

```powershell
# Preview what will be repaired (no changes made)
python mempalace_repair_encoding.py --dry-run

# Apply the repair
python mempalace_repair_encoding.py
```

The palace path is hardcoded to the default Windows location 
(`C:\Users\<you>\.mempalace\palace`). Edit line 16 if yours is different.

After running, start a new conversation with your AI so it reloads 
the palace with the repaired content.

## Notes

- The script handles mixed damage patterns in the same drawer
- `DIRECT_FIXES` dictionary covers uppercase accented characters 
  (Ó, É, Ñ, etc.) that the sliding window cannot resolve automatically
- Safe to run multiple times — already-clean drawers are skipped
- Developed by [@Cre3b3ep](https://github.com/Cre3b3ep) with assistance from Claude (Anthropic)
- The sliding window algorithm makes the script language-agnostic —  it repairs bytes, not language patterns. DIRECT_FIXES can be extended for other uppercase accented characters as needed.

fatkobra:
## What does this PR do?

- add a supported repair path for legacy Windows mojibake
- automatically repair only high-confidence UTF-8-as-Windows-1252 signatures
- deliberately leave ambiguous `Ä...` and `Å...` sequences unchanged
- preserve the multilingual clean-text cases raised during review
- print bounded before/after previews for every proposed change
- default to a read-only dry-run
- create a private, non-overwriting backup before any update
- flush original documents to disk before overwriting live rows
- bind each backup to its collection name
- provide a validated restore operation
- hold the palace writer lock during apply and restore

## Detection contract

The original implementation treated any short Windows-1252 character window whose bytes formed valid UTF-8 as damaged.

That was unsafe because legitimate sequences such as:

```text
Å²
É®
É 
```

can also form valid UTF-8 byte sequences.

The revised automatic mode recognizes only the common visible lead families produced by UTF-8-as-Windows-1252 corruption:

```text
Â...
Ã...
â...
ð...
ï...
```

The `Ä...` and `Å...` families are intentionally excluded from automatic repair because they are ambiguous with legitimate Central European and scientific text.

This deliberately favors false negatives over destructive false positives.

The old Windows corruption path can also preserve five byte values that Windows-1252 leaves undefined:

```text
0x81
0x8D
0x8F
0x90
0x9D
```

Those values appear in stored Python strings as invisible C1 control characters.

The revised decoder accepts them only as continuation positions inside an already-recognized high-confidence mojibake sequence. It maps each control code point back to its original raw byte before UTF-8 decoding.

This repairs characters such as `Á`, `Í`, `Ï`, `Ð`, `Ý`, and the closing curly quote `”` without enabling the ambiguous visible `Ä...` or `Å...` lead families.

Standalone control characters are not rewritten.

## Dry-run behavior

Dry-run is the default.

It scans the collection without writing and prints every proposed drawer ID with a bounded before/after preview.

Users can therefore inspect the actual proposed changes before choosing `--apply`.

## Apply and backup safety

Apply requires a backup path internally. When no path is supplied on the command line, a timestamped path beside the palace is generated.

The backup:

- is created with exclusive creation;
- never overwrites an existing file;
- uses owner-only permissions where supported;
- records the target collection name;
- stores each changed drawer ID and its complete original document;
- is flushed and fsynced before the corresponding collection update.

The collection name is resolved through MemPalace backend wrappers, including `ChromaCollection`.

## Restore safety

The same command restores original documents with:

```text
python3 scripts/mempalace_repair_encoding.py \
  --palace /path/to/palace \
  --restore-backup /path/to/backup.jsonl
```

Before the first restore write, the implementation:

- validates the backup header;
- validates every JSONL record;
- rejects a backup belonging to a different collection.

## Locking

Apply and restore hold the normal palace writer lock for the complete operation.

The underlying backend write lock remains re-entrant, so calls through `ChromaCollection.update()` do not deadlock.

## Tests

Regression coverage includes:

- 28 clean multilingual strings;
- the exact `PERÚ`, `Å²`, `CAFÉ®`, and `RÉSUMÉ :` review cases;
- 12 high-confidence damaged strings;
- double-encoded text;
- ambiguous sequences that must remain unchanged;
- idempotence;
- paginated dry-run behavior;
- backup-before-update behavior;
- backup overwrite refusal;
- wrapped Chroma collection-name resolution;
- rejection of a backup for another collection;
- complete-backup validation before restoration;
- a real ChromaDB collection dry-run, apply, verification, and restore path.

The actual command was also exercised manually across separate processes against a disposable persistent palace:

1. seed four clean and two damaged drawers;
2. run the real dry-run;
3. verify that only the damaged IDs are proposed;
4. apply with an explicit backup;
5. validate the collection-bound backup and original documents;
6. reopen the palace in a fresh process;
7. verify every clean and repaired document exactly;
8. restore the backup through the command;
9. reopen again and verify the original damaged documents.

Additional regression coverage verifies:

- all five undefined Windows-1252 continuation-byte values;
- Spanish uppercase `Á` and `Í`;
- `Ï`, `Ð`, and `Ý`;
- both opening and closing curly quotes;
- a mixed row containing visible and invisible corruption;
- one-pass completion through `repair_collection()`;
- a real ChromaDB apply where all selected rows are completely repaired;
- an immediate second dry-run reporting zero remaining changes;
- backup restoration of the original invisible-byte rows.

Closes #1055

## How to test

- `python3 -m pytest tests/test_encoding_repair.py -q`
- `python3 -m ruff check .`

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,13 +30,18 @@ jobs:
           python-version: "3.13"
           cache: 'pip'
       - run: pip install -e ".[dev]"
-      # ChromaDB's rust HNSW core intermittently fails compaction on Windows
-      # ("Failed to apply logs to the hnsw segment writer") regardless of our
-      # code — a long-standing, non-reproducible-on-Linux/macOS flake. Retry
-      # ONLY that specific transient error (via --only-rerun) so real,
+      # ChromaDB's rust HNSW core intermittently fails compaction or reader
+      # initialization on Windows ("Failed to apply logs to the hnsw segment
+      # writer" / "Error creating hnsw segment reader: Nothing found on disk")
+      # regardless of our code — long-standing, non-reproducible-on-Linux/macOS
+      # flakes. Retry ONLY those specific transient errors (via --only-rerun) so real,
       # deterministic failures still fail on the first run. Linux/macOS jobs
       # deliberately run with no reruns so genuine regressions surface there.
-      - run: python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80 --durations=10 --reruns 2 --reruns-delay 5 --only-rerun "Failed to apply logs to the hnsw segment writer"
+      - run: >-
+          python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace
+          --cov-report=term-missing --cov-fail-under=80 --durations=10 --reruns 2
+          --reruns-delay 5 --only-rerun
+          "Failed to apply logs to the hnsw segment writer|Error creating hnsw segment reader: Nothing found on disk"
 
   test-macos:
     runs-on: macos-latest
--- a/mempalace/encoding_repair.py
+++ b/mempalace/encoding_repair.py
@@ -1,68 +1,77 @@
-"""Repair legacy UTF-8 text that was decoded as Windows-1252."""
+"""Safely repair high-confidence UTF-8 mojibake in MemPalace drawers."""
 
 from __future__ import annotations
 
+import json
+import os
+import re
+from pathlib import Path
+from typing import Callable, Iterator, Optional, TextIO, Union
 
-def _byte_for_character(character: str) -> int | None:
+_BACKUP_FORMAT = "mempalace-encoding-repair"
+_BACKUP_VERSION = 1
+_UNDEFINED_CP1252_BYTES = frozenset(
+    {
+        0x81,
+        0x8D,
+        0x8F,
+        0x90,
+        0x9D,
+    }
+)
+
+
+def _cp1252_character(byte_value: int) -> str:
+    return bytes([byte_value]).decode("cp1252")
+
+
+_CONTINUATION_CHARS = "".join(
+    _cp1252_character(byte_value)
+    for byte_value in range(0x80, 0xC0)
+    if byte_value not in _UNDEFINED_CP1252_BYTES
+)
+_CONTINUATION_CLASS = re.escape(_CONTINUATION_CHARS)
+
+# These are the characteristic visible lead characters produced by
+# common UTF-8-as-Windows-1252 corruption:
+#
+# C2/C3 -> Â/Ã
+# E2    -> â
+# F0    -> ð
+# EF    -> ï
+#
+# C4/C5 -> Ä/Å are deliberately excluded because strings such as Å²
+# can be legitimate scientific text.
+_HIGH_CONFIDENCE_RUN = re.compile(
+    rf"(?:"
+    rf"[ÂÃ][{_CONTINUATION_CLASS}]"
+    rf"|â[{_CONTINUATION_CLASS}]{{2}}"
+    rf"|ð[{_CONTINUATION_CLASS}]{{3}}"
+    rf"|ï[{_CONTINUATION_CLASS}]{{2}}"
+    rf")+"
+)
+
+
+def _decode_high_confidence_run(
+    match: re.Match,
+) -> str:
+    candidate = match.group(0)
+
     try:
-        encoded = character.encode("cp1252")
-    except UnicodeEncodeError:
-        codepoint = ord(character)
-        if codepoint <= 255:
-            return codepoint
-        return None
-
-    if len(encoded) != 1:
-        return None
-
-    return encoded[0]
-
-
-def _decode_candidate(
-    text: str,
-    start: int,
-) -> tuple[str, int] | None:
-    for width in (4, 3, 2):
-        segment = text[start : start + width]
-        if len(segment) != width:
-            continue
-
-        raw_values = []
-        for character in segment:
-            value = _byte_for_character(character)
-            if value is None:
-                break
-            raw_values.append(value)
-        else:
-            raw = bytes(raw_values)
-            try:
-                decoded = raw.decode("utf-8")
-            except UnicodeDecodeError:
-                continue
-
-            if len(decoded) == 1 and ord(decoded) >= 128:
-                return decoded, width
-
-    return None
+        return candidate.encode("cp1252").decode("utf-8")
+    except (
+        UnicodeEncodeError,
+        UnicodeDecodeError,
+    ):
+        return candidate
 
 
 def repair_mojibake_once(text: str) -> str:
-    """Repair one layer of UTF-8-as-Windows-1252 mojibake."""
-    output: list[str] = []
-    index = 0
-
-    while index < len(text):
-        candidate = _decode_candidate(text, index)
-        if candidate is None:
-            output.append(text[index])
-            index += 1
-            continue
-
-        decoded, width = candidate
-        output.append(decoded)
-        index += width
-
-    return "".join(output)
+    """Repair one layer of high-confidence UTF-8-as-CP1252 mojibake."""
+    return _HIGH_CONFIDENCE_RUN.sub(
+        _decode_high_confidence_run,
+        text,
+    )
 
 
 def repair_mojibake(
@@ -70,13 +79,18 @@ def repair_mojibake(
     *,
     max_passes: int = 3,
 ) -> str:
-    """Repair repeated mojibake layers until stable."""
+    """Repair repeated high-confidence mojibake layers until stable."""
+    if max_passes < 1:
+        raise ValueError("max_passes must be at least 1")
+
     current = text
 
     for _ in range(max_passes):
         repaired = repair_mojibake_once(current)
+
         if repaired == current:
             break
+
         current = repaired
 
     return current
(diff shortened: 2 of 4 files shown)


mvalentsev:
Hey again @fatkobra! Ran into a few hard ones in your latest batch. Could you run them through e2e in mempalace itself? They all show up on the first real run.

This one: the dry-run flags text that was already fine and `--apply` overwrites it, because `_decode_candidate` reads "É" + "®" as `0xC9 0xAE`, a valid two-byte sequence, and collapses the pair.

```
La canción «PERÚ» abre el disco.  ->  La canción «PERڻ abre el disco.
Buried surface area 1250 Å².      ->  Buried surface area 1250 Ų.
CAFÉ® is a registered mark.       ->  CAFɮ is a registered mark.
```

French typography hits it too, since the capital É lands right before the non-breaking space in `RÉSUMÉ :`. That's 6 of 26 natural es/fr/de/pt/pl/ca strings coming back changed, and `collection.update()` keeps no copy, so those are gone for good.

#2141: against a live chromadb the collection still comes back carrying `DefaultEmbeddingFunction`, so the blocker it closes isn't closed.

#2143: the real `chunk_text` emits no byte-identical chunks on ordinary code or prose, which is why the test has to monkeypatch it to produce any.

Worth putting the earlier PRs through e2e as well, the problem and the fix, not just the fix.

fatkobra:
Thanks @mvalentsev  — this is valid and blocking feedback.

The current detector is fundamentally too broad: it accepts any short CP1252 character window whose bytes happen to decode as one non-ASCII UTF-8 character. That means already-correct text can be changed even when it contains no mojibake marker.

The examples with `É®`, `Å²`, and French typography demonstrate that this is not safe enough for an in-place repair operation. A few special-case exclusions would not be a responsible fix because other valid character combinations could still collide.

I am converting this PR to draft.

Before marking it ready again, I will require:

- a broad multilingual clean-text corpus that must remain byte-identical;
- real corrupted strings representing the pre-3.1 Windows failure;
- a live ChromaDB dry-run/apply/reopen test;
- exact changed-ID and before/after reporting;
- a backup or rollback path before any document update;
- proof that the dry-run selects damaged documents and zero clean documents.

I will also revisit #2141 and #2143 with real end-to-end reproductions rather than mocked versions of the suspected failure.

Thank you for running these against the real application.

fatkobra:
Thanks again — the initial E2E run found one additional wrapper issue before I pushed the revision: the backup records were correct, but the collection field was `null` because the MemPalace `ChromaCollection` wrapper does not expose a public `name` attribute.

That is now fixed and covered as well.

The revised PR now:

- removes the broad sliding-window decoder;
- repairs only the high-confidence `Â`, `Ã`, `â`, `ð`, and `ï` lead families;
- deliberately leaves ambiguous `Ä...` and `Å...` families unchanged;
- preserves all 28 multilingual clean-text cases;
- includes the exact `PERÚ`, `Å²`, `CAFÉ®`, and `RÉSUMÉ :` review examples;
- creates a private, non-overwriting backup before apply;
- records and validates the target collection name through backend wrappers;
- flushes and fsyncs originals before collection updates;
- validates the complete backup before restoration;
- rejects restoration into a different collection;
- provides a tested restore operation;
- holds the palace writer lock during apply and restore.

I reran the complete command against a newly created disposable palace across separate processes:

1. dry-run scanned six rows and proposed only the two damaged IDs;
2. apply updated exactly those two rows;
3. the backup contained the correct collection name and two complete originals;
4. a fresh process read all four clean review strings unchanged and both damaged strings repaired;
5. restore validated and restored two records;
6. another fresh process verified the original damaged strings.

The PR remains draft pending review of the revised safety contract.

mvalentsev:
One thing is still open, from the same place. cp1252 leaves five bytes undefined (`0x81 0x8D 0x8F 0x90 0x9D`) and `_CONTINUATION_CHARS` drops them, so a character whose UTF-8 continuation byte lands there can never match: `Á Í Ï Ð Ý`, and the closing quote `”` while the opening `“` repairs.

On a palace damaged the way a pre-3.1.0 Windows install damaged it, invisible bytes escaped:

```
in the palace   Ã\x81LVARO vive en PARÃ\x8dS. Ã\x8dNDICE: pÃ¡gina 12.
after --apply   Ã\x81LVARO vive en PARÃ\x8dS. Ã\x8dNDICE: página 12.

in the palace   Dijo â€œholaâ€\x9d y se fue.
after --apply   Dijo “holaâ€\x9d y se fue.

Rows scanned: 4 / Documents needing repair: 4 / Documents updated: 4
```

Three of those four rows get rewritten, counted as updated, and are still damaged. A second run does not finish them.

#1055 is a Spanish palace, and the reporter writes there that `DIRECT_FIXES` covers the uppercase accented characters the sliding window cannot resolve. `Á` and `Í` are ordinary Spanish capitals.

The six strings from the last round do survive a real `--apply` now.

fatkobra:
Thanks again @mvalentsev for excellent feedback — this was valid and blocking.

The previous revision deliberately omitted the five byte values that Windows-1252 leaves undefined:

`0x81`, `0x8D`, `0x8F`, `0x90`, and `0x9D`.

That preserved the earlier clean-text safety, but it also meant that old palaces containing those values as invisible C1 controls could only be partially repaired. The row was counted as updated when one visible portion changed even though another damaged sequence remained.

The revised decoder now:

- includes those five invisible controls as valid continuation positions;
- maps them back to their original raw byte values before UTF-8 decoding;
- still recognizes only the existing high-confidence `Â`, `Ã`, `â`, `ð`, and `ï` lead families;
- does not enable the ambiguous visible `Ä...` or `Å...` lead families;
- leaves the previous `PERÚ`, `Å²`, `CAFÉ®`, and `RÉSUMÉ :` clean cases unchanged.

New coverage includes:

- `Á`, `Í`, `Ï`, `Ð`, and `Ý`;
- the exact `Ã\x81LVARO ... PARÃ\x8dS ... Ã\x8dNDICE` example;
- the exact `Dijo â€œholaâ€\x9d` example;
- a four-row fake-collection apply;
- a four-row real ChromaDB apply;
- exact post-apply document verification;
- an immediate second dry-run that must report zero remaining changes;
- backup restoration of all four original damaged rows.

I also exercised the actual command against a disposable persistent palace with four clean and four damaged rows:

1. dry-run selected exactly the four damaged rows;
2. apply updated exactly four rows;
3. a fresh process verified the complete repaired text and no remaining C1 controls;
4. a second real dry-run reported zero repairs;
5. restore recovered all four original damaged documents.

fatkobra:
The current Windows failure is unrelated to this PR's encoding changes.

The only failing test is:

`tests/test_closets.py::TestDrawerGrepExpansion::test_hybrid_search_enrichment_isolates_chunks_across_drawers_sharing_source_file`

It failed inside Chroma with:

The run otherwise completed with 3,357 passed tests, and none of the `tests/test_encoding_repair.py` tests failed.

PR #2148 changes only the encoding-repair module, its command-line script, and its tests. It does not modify the closet/search/HNSW path. The same test and exact HNSW error have also appeared in an earlier Windows CI run.

The workflow currently reruns only failures matching:

`Failed to apply logs to the hnsw segment writer`

so this `Nothing found on disk` variant was not retried.

@igorls  could you please rerun the failed Windows job?

igorls:
> The current Windows failure is unrelated to this PR's encoding changes.
> 
> The only failing test is:
> 
> `tests/test_closets.py::TestDrawerGrepExpansion::test_hybrid_search_enrichment_isolates_chunks_across_drawers_sharing_source_file`
> 
> It failed inside Chroma with:
> 
> `Error creating hnsw segment reader: Nothing found on disk`
> 
> The run otherwise completed with 3,357 passed tests, and none of the `tests/test_encoding_repair.py` tests failed.
> 
> PR #2148 changes only the encoding-repair module, its command-line script, and its tests. It does not modify the closet/search/HNSW path. The same test and exact HNSW error have also appeared in an earlier Windows CI run.
> 
> The workflow currently reruns only failures matching:
> 
> `Failed to apply logs to the hnsw segment writer`
> 
> so this `Nothing found on disk` variant was not retried.
> 
> @igorls  could you please rerun the failed Windows job?

I've been testing from mac recently, will switch to windows to test this better

PR #2150, opened 2026-08-04
fix(cli): record embedder identity on closets too

tonyliu312:
## What does this PR do?

- record the embedder identity on both collections in `RECOVERABLE_COLLECTIONS` instead of only the default drawers collection
- name the collection in every success/failure line so a partial run is readable
- add a regression test asserting neither collection warns after `set-embedder`

`set_palace_embedder_identity()` already takes `collection_name`, but `cmd_palace_set_embedder` never passes it, so it always lands on `mempalace_drawers`. Closets stay in the `unknown` state.

That is more than a stray warning. `_enforce_embedder_identity` exists so "a model swap fails fast — before any query silently returns degraded results". With no recorded identity on closets, that check never engages there: after a model swap drawers raise `EmbedderIdentityMismatchError` while closet vectors keep being compared against embeddings from a different model.

Hit this on a qdrant palace (86k drawers / 70k closets) that predates the identity sidecar. `palace set-embedder --model embeddinggemma` reported success, and every subsequent search still warned on `mempalace_closets`.

## How to test

```
mempalace palace set-embedder --model <your model>
```

Before: one `✓ recorded embedder identity` line, and opening the palace still warns on `mempalace_closets`.
After: one line per collection, no warning from either.

```
pytest tests/test_embedder_identity.py -v
```

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1063,38 +1063,51 @@ def cmd_palace_set_embedder(args):
     """Record (or force-override) a palace's embedder identity (RFC 001).
 
     Resolves the ``unknown`` state for a legacy palace, or records a specific
-    model with ``--model``. It records identity on the palace only; it does not
-    change the configured model — when the two differ it prints how to align
+    model with ``--model``. Both the drawers and the closets collection are
+    recorded — a palace whose closets keep an unknown identity never gets the
+    fail-fast model-swap check that ``_enforce_embedder_identity`` provides.
+    It records identity on the palace only; it does not change the configured
+    model — when the two differ it prints how to align
     ``MEMPALACE_EMBEDDING_MODEL``. ``--force`` overwrites an existing,
     differently-named identity.
     """
     from .backends.base import EmbedderIdentityMismatchError
     from .palace import set_palace_embedder_identity
+    from .repair import RECOVERABLE_COLLECTIONS
 
     config = MempalaceConfig()
     palace_path = os.path.abspath(
         os.path.expanduser(args.palace) if args.palace else config.palace_path
     )
     model = getattr(args, "model", None)
-    try:
-        old, new = set_palace_embedder_identity(
-            palace_path,
-            model=model,
-            force=getattr(args, "force", False),
-            backend=_backend_arg(args),
-        )
-    except EmbedderIdentityMismatchError as exc:
-        print(f"  ✗ {exc}")
-        raise SystemExit(2) from exc
-    if old is None:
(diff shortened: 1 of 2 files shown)


remotejob:
Corroborating this on a different backend, and on a palace that is *not* legacy — the bug reproduces on a `mempalace_closets` collection created minutes earlier by `mempalace compress` on 3.9.0.

**Environment**

- MemPalace 3.9.0 (pipx), Python 3.14.4
- chromadb 1.5.9 (the PR's report was qdrant)
- Palace: 1452 drawers / 1446 closets, model `minilm` (all-MiniLM-L6-v2, 384-dim)

**Sequence**

1. Ran `mempalace compress --wing=<wing>` on a palace that had no `mempalace_closets` collection at all. The run created it and emitted:

   ```
   palace.py:287: EmbedderIdentityUnknownWarning: palace collection 'mempalace_closets'
   has no recorded embedder identity; assuming the current model 'minilm'.
   Run `mempalace palace set-embedder --model <name>` to record it.
   ```

2. Followed that instruction exactly:

   ```
   $ mempalace palace set-embedder --model minilm
     ✓ recorded embedder identity: minilm (dim=384)
   ```

3. The sidecar showed only one collection recorded:

   ```json
   { "mempalace_drawers": { "model_name": "minilm", "dimension": 384 } }
   ```

   Opening closets still warned. So the command the warning itself recommends does not fix the collection the warning is about.

4. Worked around it by calling the library directly, which already supports this:

   ```python
   from mempalace.palace import set_palace_embedder_identity
   set_palace_embedder_identity(
       "/path/to/palace", model="minilm", collection_name="mempalace_closets",
   )
   ```

   After that, both collections are recorded and a real open of closets emits zero `EmbedderIdentityUnknownWarning`.

**One extra data point that may be worth a look**

`_enforce_embedder_identity` is documented to record identity when a collection is "brand-new and empty":

> Record only when the collection is brand-new and empty: recording the *current* model on a legacy palace that already holds vectors from an unknown model would mislabel it […]

Here `mempalace_closets` genuinely was brand-new — `compress` created it via `get_closets_collection(palace_path, create=True)` (`palace.py:350`, called from `cli.py:3058`), which reaches `_enforce_embedder_identity(..., create=True)`. The auto-record still did not fire, and the run warned instead. I did not chase why, so this may be a second, separate gap from the one this PR fixes — the CLI plumbing fix here would mask it either way, since a subsequent `set-embedder` would then cover both collections.

Happy to re-test against this branch if that's useful.

PR #2159, opened 2026-08-05
test(chroma): assert the HNSW defaults rather than a threshold arithmetic

joshuafontany:
Follow-up to the review on #2107.

The HNSW default tests assert what a created collection carries: `batch_size` and `sync_threshold` land on fresh metadata at chromadb's own documented values, through both `ChromaBackend` creation paths. The relationship they assert reads `batch_size <= sync_threshold` — what chromadb permits — in both tests that check it.

Their docstrings state the same ground: the values are chromadb's own defaults, so a collection this backend creates indexes on the same terms as one chromadb creates itself, and `sync_threshold=2` reads as a value chromadb's own parameter validation rejects.

`_hnsw_creation_metadata(options)` is the single place `hnsw_space` resolves, so `get_collection()` holds no second copy of that resolution. The option's path to a created collection is unchanged.

Write amplification belongs to the benchmark suite, where it can measure I/O.

`tests/test_backends.py`: 91 passed. Backend/palace selection: 593 passed, 16 skipped.

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -2410,9 +2410,6 @@ class ChromaBackend(BaseBackend):
                 pass
 
         client = self._client(palace_path)
-        hnsw_space = "cosine"
-        if options and isinstance(options, dict):
-            hnsw_space = options.get("hnsw_space", hnsw_space)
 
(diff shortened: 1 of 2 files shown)


igorls (approved):

PR #1933, opened 2026-07-06
RFC 004: The Replicated Palace

igorls:
# RFC 004: The Replicated Palace

> **Draft.** This is the big one — the arc from a single-hub shared brain to **N equal
> replicas of one palace**, converging over an encrypted mesh, each with a local writer
> and locally-derived indexes. It also lands RFC 003 (the agent coordination logstream)
> and RFC 005 (agent identity & routing), which RFC 004 builds on.
>
> **66 commits · 95 files · +20.8k/−0.4k.** Opened as a draft to review the design and the
> seams before it merges — feedback welcome from anyone. Not asking to merge yet — asking
> for eyes.

---

## What does this PR do?

Today a MemPalace "shared brain" is one hub other machines talk to. When the hub machine
sleeps, every other machine loses recall, capture, and coordination at once — a brain that
stops remembering because *another* brain fell asleep. RFC 004 removes that dependency:

**Each human has ONE palace, replicated in full across every machine they own.** Agents
always talk to `127.0.0.1`; the services converge with each other over the mesh. The hub
stops being a dependency and becomes a *role* (rendezvous, and the home of genuinely
*shared* palaces). Offline is the default posture, not an edge case — **recall reads and
capture writes never block on the network; only convergence may wait.**

The mechanism, one sentence: *N equal replicas of the facts, each with a local writer and
locally-derived indexes, converging through provenance-stamped ops over the mesh, with
origin as the home-of-record for source-bound maintenance.*

Three RFCs land together:
- **RFC 003 — Agent logstream:** append-only, verbatim coordination events (task.request /
  reply / patch.ready) so agents on different machines delegate work and hand off patches.
- **RFC 004 — The Replicated Palace:** the op-log, anti-entropy sync, fold/promote, the v4
  content-pure id migration, and the write-flip cutover.
- **RFC 005 — Agent identity & routing:** `host:agent:project` identity so ops and events
  are attributable and routable across the fleet.

---

## Read this first (suggested review order)

For a PR this size, the docs are the map. Recommended path:

1. **`docs/rfcs/004-replicated-palace.md`** — the design, requirements (R0–R8), and the
   storage/transport split. Start here; everything else is this doc made real.
2. **`docs/rfcs/003-agent-logstream-coordination.md`** — the coordination substrate RFC 004
   rides on.
3. **`docs/rfcs/005-agent-identity-routing.md`** — short; the identity model.
4. **`website/concepts/replicated-palace.md`** + **`website/guide/shared-brain.md`** — the
[...]

PR #2162, opened 2026-08-06
feat: agent logstream — coordination for a multi-machine agent fleet (RFC 003)

igorls:
# Agent Logstream — coordination for a multi-machine agent fleet

Carves the shippable half out of [the RFC 004 draft](https://github.com/MemPalace/mempalace/pull/1933) and lands it on develop for 3.7.0: **RFC 003, the agent coordination logstream**, plus the hub and sync machinery it needs to work across machines.

Agents on different machines can delegate work to each other, hand back patches verbatim, and close the loop with acks — through an append-only event stream that converges between hubs on its own.

**34 files, +8.8k / −54.** No palace migration, nothing removed, every new surface opt-in.

---

## What you get

**Coordination primitives (RFC 003).** Append-only, verbatim events with routing (`stream`, `room`, `from_agent`, `to_agent`, `correlation_id`) and an artifact store for exact content — patches, files, logs — addressed by SHA-256.

Seven MCP tools: `event_append`, `event_list`, `event_wait`, `event_ack`, `artifact_put`, `artifact_get`, `patch_submit`. Matching CLI: `mempalace logstream` and `mempalace artifact`, both `--json` for scripting. `event_wait` long-polls, and the hub serves `GET /logstream/stream` over SSE with `Last-Event-ID` resume.

**The hub becomes a service.** One `mempalace serve` process owns the palace and its writer lease. Stdio MCP servers spawned by agent harnesses now detect a live hub and **auto-proxy every request to it** — plugins and desktop apps join the shared brain with zero client-side reconfiguration (`MEMPALACE_HUB_FORWARD=0` opts out). CLI and hook mines route the same way, so transcript capture doesn't silently stop on a hub machine. Adds an authenticated `/statusz` and a `Host` allowlist for loopback binds behind a tailnet or reverse proxy.

**Cross-machine sync.** Name peers in `peers.json`; hubs exchange version vectors and pull what they're missing every `MEMPALACE_SYNC_INTERVAL` seconds (default 15). Multi-master and idempotent — every replica carries every origin's events, so two machines that never exchanged credentials still converge through a common peer, and a machine offline for a week re-pulls the tail. `GET /sync/peers` (and `mempalace_mesh_peers`) shows the estate: reachability, version vectors, and origins known only through gossip. Bearer tokens live in `peers.json` and are never included in that payload — asserted in tests.

---

## What is deliberately NOT here

This is the part worth reading before testing, because the parent PR's docs promised more.

**Memory does not replicate.** Peer sync covers the logstream — events and artifacts. Each machine's drawers and knowledge graph stay local to that machine. Two synced machines share an inbox and can hand patches back and forth, but they **do not share recall**: ask one what it remembers and you get that machine's palace. If you want one shared memory today, point every agent at a single hub (Remote / Team Server), as before.

Also not included, and staged with the rest of RFC 004: the memory op-log, the v4 content-pure id recipe and its write-flip, `migrate-ids` / `reconcile-ids`, memory read replicas (`mempalace replica`), and the `/snapshot/*` endpoints. `ID_RECIPE` stays `v3` — **nothing about an existing palace changes on upgrade.**

The website docs in this PR were rewritten to say exactly this. The parent branch's `concepts/replicated-palace.md` is held back rather than shipped describing a mesh whose memory doesn't converge.

---

## A bug worth calling out

Found by actually running two hubs, not by the test suite.

`_start_peer_sync_thread()` read `peers.json` **once at startup** and returned early when the file was absent — so the thread never started. A hub booted before `peers.json` was written would never sync, silently and permanently, with no error anywhere.

That is the exact order the setup guide tells you to follow: run the hub, then name the peers. Anyone following it in order would have gotten a fleet that quietly never converged.

Membership is now re-read every round (a malformed file logs once per transition, not every 15s). Regression coverage in `tests/test_logstream_sync.py::TestPeerSyncThreadStartup` — the tests were confirmed to fail against the old behavior before being kept.

---

## How to test

```bash
uv sync --extra dev
```

Then the part that matters — **two hubs on one box, no restart between steps:**

```bash
# two throwaway palaces
mkdir -p /tmp/mpA /tmp/mpB
TOKEN=$(openssl rand -hex 16)

# start both hubs FIRST, with no peers.json (the order the guide uses)
MEMPALACE_MCP_HTTP_TOKEN=$TOKEN mempalace serve --palace /tmp/mpA --host 127.0.0.1 --port 8793 &
MEMPALACE_MCP_HTTP_TOKEN=$TOKEN mempalace serve --palace /tmp/mpB --host 127.0.0.1 --port 8794 &

# now introduce them to each other
echo "{\"peers\":[{\"name\":\"B\",\"url\":\"http://127.0.0.1:8794\",\"token\":\"$TOKEN\"}]}" > /tmp/mpA/peers.json
echo "{\"peers\":[{\"name\":\"A\",\"url\":\"http://127.0.0.1:8793\",\"token\":\"$TOKEN\"}]}" > /tmp/mpB/peers.json

# delegate on A
mempalace --palace /tmp/mpA logstream append --type task.request \
  --stream project/demo --room delegation --from-agent mac-claude \
  --to-agent windows-codex --correlation-id task_1 --body "Fix the flaky test." --json

# wait ~15s, then check B's inbox — no manual sync, no restart
mempalace --palace /tmp/mpB logstream list --to-agent windows-codex --json
```

Verbatim artifact round-trip:

```bash
git diff | mempalace --palace /tmp/mpB artifact put --kind patch --created-by windows-codex --json
mempalace --palace /tmp/mpA artifact get art_... > /tmp/out.patch   # should be byte-identical
```

Note `serve` takes its own `--palace` *after* the subcommand — `mempalace --palace X serve` silently ignores the global one. That shadowing is pre-existing on develop, not introduced here, but it cost me a confusing few minutes and may be worth a follow-up.

What I verified locally on two live hubs: the full delegation loop (request → claimed → patch.ready → ack), byte-identical patch round-trip by sha256, sync in both directions with `origin_replica` provenance intact, CLI `logstream sync` running safely alongside a live hub, and auto-convergence 15s after writing `peers.json` with no restart.

---

## Design-principle compliance

- **Verbatim always** — events and artifacts are immutable and byte-exact; artifacts carry a SHA-256 and round-trip unchanged. Nothing is summarized.
- **Incremental only** — append-only by construction. No migration, no rewrite of existing data; a crash mid-round means the next round re-pulls the tail.
- **Local-first, zero external API** — sync is hub-to-hub over the user's own network. No content leaves machines the user operates.
- **Privacy by architecture** — peer bearer tokens live only in `peers.json` and never appear in events, artifacts, or the estate payload (asserted in tests).
- **Performance budgets** — recall and capture never block on the network; logstream tools dispatch outside the global HTTP request lock so long-polls can't starve the hub.

---

## Checklist

- [x] Tests pass (3782 passed, 31 skipped)
- [x] Linter and formatter clean
- [x] No hardcoded paths
- [x] Docs match shipped behavior (rewritten for this scope)
- [ ] Hands-on testing ← **@milla-jovovich, over to you**

@milla-jovovich — the two-hub walkthrough above is the fastest way to see it work, and the "What is deliberately NOT here" section is the honest boundary of what this does. Would especially value your read on whether the coordination-vs-memory split lands clearly for someone meeting the feature for the first time, and on the protocol snippet in `website/guide/shared-brain.md` that agents get pasted into their instruction files.


igorls:
@milla-jovovich — one fix landed since you were tagged, worth knowing before you test.

Every CLI path carrying artifact and event content read and wrote through Python's text layer, which applies universal-newline translation. A patch piped in from a Windows agent had its CRLF flattened to LF before storage, so it was stored as different bytes than `git diff` produced — under a sha256 of the mangled content, which then wouldn't match on the machine applying it. The `--out` path had the mirror problem: on Windows the text layer expands LF back to CRLF on write, so the file on disk wouldn't match the digest you're told to verify.

It also quietly disarmed the CRLF warning for `kind=patch`: `put_artifact` looks for a carriage return that the read path had already stripped, so the check meant to catch unappliable diffs could never fire on the platform that produces them.

Reads and writes now go through `.buffer` and `read_bytes()`/`write_bytes()`. Verified through real pipes: a 63-byte CRLF patch round-trips at 63 bytes with an unchanged sha256 via stdin, `--out`, and stdout, and the warning fires. Regression coverage in `TestVerbatimNewlines`, confirmed failing against the old behavior first.

Config reads (`peers.json`, `replica.json`, the server registry, the token file) were deliberately left alone — JSON is newline-insensitive and none of them carry user content.

If you have a Windows box in reach, that is the interesting seat for the artifact round-trip. On macOS and Linux the old code happened to work, which is exactly why it survived this long.

igorls:
Correction on the last line: don't worry about sourcing a Windows box, @milla-jovovich — I will cover the Windows seat for the artifact round-trip from ours. The macOS walkthrough in the description is the full ask for you.

PR #2165, opened 2026-08-06
feat(pgvector): push the lexical filter down to Postgres

ATKabli:
## What does this PR do?

`PgVectorCollection.lexical_search` scrolls the **entire table**, documents included, to the client and scores BM25 in Python. Cost is O(table size) bytes over the wire per query. On a remote palace this makes the lexical arm of hybrid search unusable: at ~141k drawers over a WireGuard link we measured **minutes per query**.

This PR pushes the *filter* into Postgres and keeps the *ranking* local and exact.

**This is a redesign of the PR's first version**, in direct response to @mvalentsev's review. The first version ranked server-side (`ts_rank_cd` + SQL `LIMIT`), and his repro proved that changes the **result set**, not just the transport: ts_rank_cd and BM25 disagree on what matters (no IDF, no length normalization), so no window size fixes it. Both commits are kept so the evolution is reviewable; the second commit's message is written against the first.

**What Postgres decides, and what it does not.** Postgres evaluates a *filter* and nothing else: an OR of literal-substring tests (`ILIKE`, trigram-indexable), one per query token, with no `LIMIT` and no server-side ranking. Scoring, ranking and the `n_results` cut all happen client-side, over the same BM25 numbers the local path produces. The pushdown therefore decides only which rows do not need to cross the wire, never which rows win.

That the filter cannot drop a hit is a proof rather than a measurement: every token `_tokenize` emits is a span the regex matched inside `text.lower()`, so it is a literal substring of the lowered document, and a case-insensitive substring test on it is true for every document `_bm25_scores` can score above zero. The filter admits a superset of the scored set; the extra rows score zero and are dropped locally.

Identical scores additionally require the collection's BM25 statistics, not the candidate pool's — IDF is a function of collection size and length normalization a function of mean document length, so a pool scored with pool-derived statistics reorders. Document frequency needs no correction, since a document containing a query term always survives a filter built from those terms. Collection size and mean document length come from one aggregate over a `token_count` column, written on upsert by the same `_tokenize` BM25 uses, backfilled once for existing rows. Verified against a live server: **330 comparisons over an adversarial corpus, asserting identical ids and identical scores against the pre-pushdown implementation, zero mismatches.**

The pushdown declines, and the query takes the exact local path unchanged, when: `where` does not translate fully to SQL (`$or`, comparisons, `$contains`, or any null-valued operand); any row in scope lacks a `token_count`, so the collection statistics would be a guess; or the server returns an error. `UnsupportedFilterError` is re-raised rather than silently degraded. `maintenance_state()` reports `lexical_pushdown` so which path a collection is on is observable rather than inferred.

Measured on 10k drawers (5.2 MB of text): a selective query moves 0.00 MB instead of 5.17 MB and runs 66x faster; a mid-frequency query saves 97.6% of the wire and runs 2x faster; a query whose term appears in every document saves nothing and costs ~13%, which is the bounded worst case.

## How to test

Unit tests run without a database:

```
pytest tests/test_pgvector_backend.py tests/test_live_pgvector_conformance.py tests/test_maintenance_hooks.py
→ 86 passed, 23 skipped
```

The 23 skips are live-server arms; set `MEMPALACE_PGVECTOR_LIVE_DSN` to a scratch database to run them, including the parity suite that asserts identical ids and scores between pushdown and local paths.

Full suite on this branch: **4319 passed, 38 skipped** (develop after the 3.7.1 sync, `906b918`, same environment baseline: 4295 passed, 31 skipped — the 7 extra skips are the new live-server arms). `ruff check .` and `ruff format --check .` both clean on the repo's 0.16.1 pin.

## Checklist

- [x] Follows the design principle: the backend decides transport, never ranking — results are provably identical to the local path
- [x] Tests pass (`python -m pytest tests/ -v`) — 4319 passed, 38 skipped
- [x] No hardcoded paths

CHANGELOG entry under Unreleased / Performance. No new dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 - **Shared-brain rules are `host:harness:project`, declared-idle, and MCP-shape aware.** `mempalace rules` takes `--host --harness --project` (stable lowercase tokens) and optional `--mcp full|light` (default `full`, matching the 45-tool server). The packaged snippet is the only coordination text: compose the identity from the current workspace, arm `logstream watch` only on listen / claim / delegate, write topics on named lanes without filtering the default inbox on them, claim with a lowest-HLC mutex, and use `kg_supersede` for single-valued fact changes. `--mcp light` swaps tool tokens onto the 3-tool triad; prose is identical. `logstream watch --agent` now defaults a sanitized `--state-file` (`:` → `_` under `~/.mempalace/watch/`) so Windows tuple identities do not need a private path overlay.
 
+### Performance
+
+- **pgvector lexical search no longer scrolls the whole table.** `lexical_search` pushed every document over the wire on every call, which made hybrid search unusable on a remote palace of ~141k drawers. Postgres now evaluates a substring filter that is a provable superset of the documents BM25 can score above zero, so only candidate rows cross the wire. Scoring, ranking and the `n_results` cut stay client-side and produce the same ids and the same scores as before. Collections gain a `token_count` column and a `pg_trgm` index; a palace created earlier picks both up on open and is backfilled once. `maintenance_state()` now reports `lexical_pushdown` so an operator can see which path a collection is on. (#2165)
+
 ### Bug Fixes
 
 - **The transcript-path fallback no longer gives every git worktree its own wing.** `_wing_from_transcript_path`'s primary path (reading `cwd` from the JSONL) already collapsed a `<project>/.claude/worktrees/<wt>` segment before deriving the wing; the fallback path, used whenever `cwd` is absent, had no equivalent strip, so the flattened `--claude-worktrees-<wt>` segment survived into the wing name. Applied the same collapse there. (#2388)
--- a/mempalace/backends/pgvector.py
+++ b/mempalace/backends/pgvector.py
@@ -94,6 +94,14 @@ _TOKEN_RE = re.compile(r"\w{2,}", re.UNICODE)
 # fold to "_", and the character set the result must then consist of.
 _SHARED_NAMESPACE_SEPARATOR_RE = re.compile(r"[\s\-./:_]+")
 _SHARED_NAMESPACE_RE = re.compile(r"^[a-z0-9_]*[a-z0-9][a-z0-9_]*$")
+
+# Text-search configuration for the pushed-down lexical arm. 'simple' does no
+# stemming and no stop-word removal, so it tokenizes any script the same way
+# ``_TOKEN_RE`` does — Arabic, CJK and English drawers all stay searchable in
+# one palace. A language config ('english') would stem and drop stop words for
+# English while silently degrading every other language, which breaks the
+# verbatim-recall promise for a mixed corpus.
+_FTS_CONFIG = "simple"
 # Operators that translate to a JSONB containment predicate and so can be
 # pushed down to SQL. Comparisons, $or and $contains stay on the local exact
 # path (Python filtering), mirroring the Qdrant backend's local fallback.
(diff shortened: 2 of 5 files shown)


mvalentsev:
The "Fallback guarantee ... the pushdown can never change or drop results" claim does not hold.

Today's path scores BM25 over the full scroll and slices `hits[:n_results]`. This one ranks by `ts_rank_cd` with `LIMIT n_results` in SQL. Two different rankings pick a different set, not just a different order.

Against a local PostgreSQL 16.2, 60 drawers, at the window `searcher.py:985` actually asks for (`n_results * 3`, so 9 for a 3-hit search):

```
query='alpha beta'   lost=4 gained=4
  unreachable: ['long17', 'long24', 'long25', 'short00']
```

`short00` is "we shipped the alpha and beta change today", an exact match on both terms and the top BM25 hit. `_hybrid_rank` recomputes BM25 over the merged pool, but it cannot rank a drawer the pool never held. Widening the window to 30 over the same 60 drawers loses nothing, so the gap opens once matches outnumber the window, which is the large-palace case this PR targets. CLAUDE.md sets 100% recall as the requirement every search path is measured against.

CI cannot catch this: `pgvector.py:21` states CI runs against the in-memory fake client, and `test_live_pgvector_conformance.py:32` skips the live arms unless `MEMPALACE_PGVECTOR_LIVE_DSN` is set.

Either way the body should not claim the pushdown cannot change results.

ATKabli:
You're right, and the repro made it easy to confirm. ts_rank_cd inside a LIMIT picks a different set than full-scroll BM25 top-n, not just a different order, and no window size fixes that because the two rankings disagree on what matters (no IDF, no length norm in ts_rank_cd). The "cannot change or drop results" line in the body was wrong and I should not have written it.

I'm reworking the PR rather than patching the window: keep BM25 as the only ranking (computed locally with exact collection statistics), push down only a candidate filter that provably over-matches the tokenizer, and prove parity against the scroll path with a live differential run rather than the fake client, which I now realize simulates Postgres with the same Python tokenizer and so could never have caught this. Will update here when it's ready; the current head should not be reviewed further.

ATKabli:
The rework is pushed, as promised above. The branch now does what the reply said it would: Postgres evaluates a filter only (token `ILIKE` tests, no `ORDER BY`, no `LIMIT`), ranking stays client-side BM25 over exact collection statistics from a `token_count` column, and parity is asserted rather than claimed — 330 live comparisons against the pre-pushdown implementation, identical ids and scores, zero mismatches. Where exactness cannot be guaranteed (non-pushdown filters, missing `token_count`, server error) the query takes the unchanged local path. `ts_rank_cd` is gone from the file entirely.

Both commits are kept so the redesign is reviewable against your objection; the PR body is rewritten to match. Rebased onto `906b918` (post-3.7.1 develop): 4319 passed, 38 skipped, ruff clean.

ATKabli:
Rebased onto `develop@c04d423`. The only conflict was `CHANGELOG.md`: 3.9.0 shipped and the `[Unreleased]` block was rewritten around the Rust engine work, so this PR's `### Performance` entry now sits as its own section after `### Features` instead of at the top of `[Unreleased]`. No code conflicted — `mempalace/backends/pgvector.py` has not been touched upstream since this branch's previous base, so the filter-only design from the 2026-08-14 rework is carried over unchanged. Both commits are preserved.

One thing worth flagging that arrived with 3.9.0: `searcher.py`'s `candidate_strategy='union'` path calls `drawers_col.lexical_search(query=, n_results=, where=)`, the same keyword shape this PR keeps, so union merging on a pgvector palace now also gets the pushdown rather than a full scroll.

Full suite as CI runs it (Python 3.13, `pytest tests/ --ignore=tests/benchmarks --cov=mempalace --cov-fail-under=80`): 4887 passed, 61 skipped, coverage 81.96%. `ruff 0.16.1 check` and `format --check` clean. The live pgvector arms remain skipped without `MEMPALACE_PGVECTOR_LIVE_DSN`; the 330-comparison parity run reported earlier has not been re-executed against this rebase.

ATKabli:
Rebased onto `develop@df330d8` after #2167, #2168 and #2170 landed. Conflicts were `CHANGELOG.md` (the `### Performance` entry now sits between `### Features` and `### Bug Fixes`), the constants block at the top of `mempalace/backends/pgvector.py` (the shared-namespace regexes from #2167 now precede `_TOKEN_COUNT_COLUMN`), and `tests/test_pgvector_backend.py`, where the new namespace and reconnect tests from develop sit ahead of the pushdown tests. No logic changed. Full suite passes locally on the rebased tree; the live pgvector parity run still needs `MEMPALACE_PGVECTOR_LIVE_DSN` as noted before.

PR #2170, opened 2026-08-06
fix(hooks): skip harness boilerplate in the checkpoint recent: line

ATKabli:
## What does this PR do?

The checkpoint entry summarizes the session as `...|msgs:N|recent:<msg>|<msg>|...`, built from `_extract_recent_messages`. Transcripts carry a lot of text tagged `role: "user"` that the user never typed, and only two wrappers were filtered: `<command-message>` and `<system-reminder>`.

Everything else leaked. On a session with background agents, local commands or skills, the last ten "user messages" are mostly tooling, so the `recent:` line fills with the same boilerplate prefix every time and the checkpoint records the harness instead of the conversation. The Codex branch was weaker still: it filtered only `<command-message>`.

Two changes.

**1. Nine more markers**, all literal wrappers observed *leading* real Claude Code and Codex transcripts:

| Marker | Source |
|---|---|
| `<command-name>` | slash-command name, when it leads |
| `<command-args>` | slash-command arguments |
| `<local-command-caveat>` | local command output caveat block |
| `<local-command-stdout>` | local command output body |
| `<task-notification>` | background task completion notice |
| `[SYSTEM NOTIFICATION` | unbracketed notification banner |
| `[Request interrupted by user` | interruption record, carries no topic |
| `[Image:` | pasted-image placeholder, no words to summarize |
| `Base directory for this skill:` | skill preamble |

They live in one module constant, `_HARNESS_BOILERPLATE_MARKERS`, with a small `_is_harness_boilerplate` helper, so there is one place to extend when a harness adds a wrapper. Both transcript formats now share the set.

**2. The match is anchored**, `startswith` after `lstrip`, not searched for anywhere in the body. That is the whole correctness argument, and it is what makes widening the list safe: unanchored, every marker added is new false-positive surface.

Position is the discriminator. The harness emits a wrapper **as** the message, so an injection always opens one. A wrapper further in is a human quoting the tooling, which is the same defect class @mvalentsev reported against the L1 salience filter in #2169. Unanchored matching discarded a 7,732-character design message because it explained, at offset 5,287, that events "arrive as `<task-notification>` messages and wake this loop" — even though the leading 200 characters, the only part `_extract_recent_messages` keeps, were pure prose. The skip decision was being made on text nothing would ever read.

Substring matching rather than regex: the wrappers are literal and the hook runs inside the 500ms budget.

### Why `startswith` and not "the leading 200 characters"

A bounded leading window was the obvious fix and it is not the right one. A quote inside the first 200 characters is still a quote, and a window makes the rule a tunable whose false-positive rate is a function of its size. `startswith` has no such knob.

The corpus below says the two score identically on real data, so anchoring costs no recall and removes the knob.

## How to test

```
pytest tests/ -q --ignore=tests/benchmarks     4303 passed, 31 skipped
```

Baseline `origin/develop@906b918` (develop after the 3.7.1 sync) in the same environment: **4295 passed, 31 skipped**, so **+8** tests and no existing test changed. ruff **0.16.1** (the repo pin), Python 3.12.13.

**Validated against real transcripts, not constructed ones.** All three rules were run over every local Claude Code transcript:

```
files=2269  jsonl_lines=239280  role:"user" lines=64376  carrying actual text=4778

rule                                     skipped    kept
anywhere in body (the current code)         1369    3409
leading 200 characters                      1352    3426
anchored, startswith after lstrip           1352    3426

anywhere drops that anchored keeps: 17
window   drops that anchored keeps:  0
anchored drops that window keeps  :  0
```

Note the shape of that corpus: only 4,778 of 64,376 `role: "user"` lines carry message text at all, the rest being `tool_result` blocks. Of the messages that do, more than a quarter are harness injection.

**Anchoring and the 200-character window agree on every one of 4,778 messages**, in both directions. So the anchor gives up nothing and drops the tunable.

All **17** messages the current unanchored rule discards and the anchored rule keeps are genuine prose. Their earliest marker sits between offset **587 and offset 245,389**; not one is inside the window the checkpoint stores:

```
len=  7677  first_marker_at=  5287  head='# /loop - schedule a recurring or self-paced prompt...'
len=  5690  first_marker_at=  3775  head='You are porting local patches of the open-source project mempalace to...'
len=  3811  first_marker_at=  1457  head='Adversarial re-review of OUR OWN upstream PRs + fix implementation...'
len=  2309  first_marker_at=  1069  head='Read-only investigation. Do NOT write any files...'
len=402870  first_marker_at=245389  head='Review this change for security vulnerabilities...'
   ... 12 more, all of the same shape
```

Per-marker position, over the same corpus:

| marker | contains | in first 200 | opens the message |
|---|---|---|---|
| `<task-notification>` | 530 | 499 | 499 |
| `Base directory for this skill:` | 350 | 345 | 323 |
| `<command-message>` | 224 | 218 | 90 |
| `<command-name>` | 221 | 218 | 128 |
| `<command-args>` | 172 | 170 | **0** |
| `<local-command-caveat>` | 133 | 128 | 128 |
| `[Request interrupted by user` | 63 | 62 | 62 |
| `[Image:` | 56 | 54 | 54 |
| `<local-command-stdout>` | 53 | 52 | 52 |
| `[SYSTEM NOTIFICATION` | 21 | 16 | 16 |
| `<system-reminder>` | **5** | **0** | **0** |

Two things in that table are worth saying rather than hiding, since both are markers the list keeps:

- **`<command-args>` never opens a message** in this corpus (0 of 172). Its enclosing slash-command envelope does, via `<command-message>` or `<command-name>`, which is why the union still catches all of them and why the anchored and windowed rules tie.
- **`<system-reminder>` never appears as top-level `role: "user"` text at all** (5 occurrences, all deep inside the body). It was one of the two original markers, and in this path it is effectively dead code. Harmless to keep for harnesses that do lead with it, but it is not doing the work the list implies.

Corpus of one machine, so treat the absolute counts as indicative. The *direction* is not marginal: 17 to 0.

New tests (8): every new marker on the Claude Code string-content shape and the list-content shape, the Codex `event_msg` shape, the 7.7k-character quoting case, prose that paraphrases a marker in words ("the task notification arrived late"), the anchoring unit contract, and a wrapper quoted at an offset that is **inside** the 200 characters the checkpoint keeps, which is the case a windowed rule gets wrong and an anchored one does not.

### Known limitation

`[Image:` is anchored like the rest, so a message that opens with a pasted-image placeholder and then contains the user's actual question is skipped whole. Of the 54 messages in this corpus opening with that prefix, **52 are the placeholder and nothing else**, which is why it is on the list; the other 2 carry real text after it and are the one place this PR has a non-zero false-positive cost. Stripping the placeholder rather than skipping the message would be the better fix and needs its own change.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — 4303 passed, 31 skipped, against 4295 on develop in the same env
- [x] No hardcoded paths

CHANGELOG entry under Unreleased. No new dependencies, no behavior change outside the checkpoint summary.


ATKabli:
Rebased onto `develop@c04d423`. The only conflict was `CHANGELOG.md`, where the FTS5 auto-heal entry landed on the same line in `[Unreleased]`; both entries are kept. `mempalace/hooks_cli.py` has had no upstream changes since this branch's previous base, so the marker list and the `startswith`-after-`lstrip` anchoring are carried over unchanged and the single commit is preserved.

Full suite as CI runs it (Python 3.13, `pytest tests/ --ignore=tests/benchmarks --cov=mempalace --cov-fail-under=80`): 4871 passed, 54 skipped, coverage 81.88%. `ruff 0.16.1 check` and `format --check` clean.

PR #2171, opened 2026-08-06
fix(embedding): clamp/skip out-of-range embeddinggemma token ids (#2114)

### LINKED ISSUE #2114: EmbeddinggemmaONNX crashes on <image_soft_token>: Gather index 262144 out of bounds
## `EmbeddinggemmaONNX` crashes on `<image_soft_token>`: Gather index 262144 out of bounds

**mempalace:** 3.6.0 (also unfixed on `develop` HEAD)
**Python:** 3.14.6 · **onnxruntime:** 1.27.0
**Config:** `embedding_model: embeddinggemma`

### Problem

`onnx-community/embeddinggemma-300m-ONNX`'s tokenizer defines
`<image_soft_token>` as special token id `262144` — one past the last row
(`262143`) of the text-only ONNX export's `embed_tokens` matrix. Any mined
text containing that literal string tokenizes to that id, and
`EmbeddinggemmaONNX.__call__` (`mempalace/embedding.py`) passes it straight
into `session.run()` with no bounds check, aborting the entire mine.

### Repro (no mempalace CLI/palace needed)

```python
import numpy as np
import onnxruntime as ort
from huggingface_hub import hf_hub_download
from tokenizers import Tokenizer

REPO = "onnx-community/embeddinggemma-300m-ONNX"
model_path = hf_hub_download(REPO, subfolder="onnx", filename="model_quantized.onnx")
hf_hub_download(REPO, subfolder="onnx", filename="model_quantized.onnx_data")
tok_path = hf_hub_download(REPO, filename="tokenizer.json")

tokenizer = Tokenizer.from_file(tok_path)
session = ort.InferenceSession(model_path, providers=["CPUExecutionProvider"])

enc = tokenizer.encode("<image_soft_token>")
print(enc.ids)  # [2, 262144, 1]

session.run(None, {
    "input_ids": np.asarray([enc.ids], dtype=np.int64),
    "attention_mask": np.asarray([enc.attention_mask], dtype=np.int64),
})
```

### Output

```
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT :
Non-zero status code returned while running Gather node. Name:'/model/embed_tokens/Gather'
Status Message: indices element out of data bounds, idx=262144 must be within the inclusive range [-262144,262143]
```

### Impact

Any codebase containing the string `<image_soft_token>` (tests, docs,
Gemma-related tooling) aborts the whole `mempalace mine` run partway
through when `embedding_model: embeddinggemma` — there's no per-document
error boundary around `session.run()`.

### Suggested fix

Clamp/filter token ids against the model's actual `embed_tokens` row
count before calling `session.run()`, or wrap the call in a per-document
try/except so one bad document is skipped instead of aborting the run.
Ollama's own conversion code hits the same class of bug and truncates
tokenizer arrays to match embedding rows (`llama/compat/llama-ollama-compat.cpp`,
near the `<image_soft_token>` comment).

### Related

#1663, #1667

belt:
## Summary

Fixes #2114 — `EmbeddinggemmaONNX.__call__` passed token ids straight into
`session.run()` with no bounds check. Any mined text containing a special
token past the model's `embed_tokens` row count (e.g. Gemma's
`<image_soft_token>`, id 262144 on the text-only ONNX export whose matrix
has 262144 rows) aborted the whole mine.

Two commits:

1. **`fix(embedding): skip poisoned documents instead of aborting mine`** —
   retry a failed sub-batch one document at a time and drop only the
   offender (filed as a zero vector), so one bad document no longer takes
   down the run.
2. **`feat(embedding): opt-in clamp for out-of-range embeddinggemma token ids`** — (optional)
   the requested clamp/filter behavior, gated behind
   `--clamp-embedding-tokens` / `MEMPALACE_EMBEDDING_CLAMP_TOKEN_IDS` /
   `embedding_clamp_token_ids` in `config.json`, off by default. When
   enabled and the optional `onnx` package (`mempalace[clamp]`) is
   installed, out-of-range ids are clamped to the last valid row before
   `session.run()` so the affected document gets a real (if lossy)
   embedding instead of the zero-vector fallback.

## Testing

- `tests/test_embeddinggemma.py`: graph-parsing against synthetic ONNX
  models (direct-initializer and quantize/dequantize-hop layouts), missing
  `onnx` package fallback, clamp on/off behavior against a mocked poisoned
  session.
- `tests/test_embedding.py`: cache-key separation, threading through
  `get_embedding_function`.
- `tests/test_config.py`: new `embedding_clamp_token_ids` config property.
- Verified end-to-end against the real cached `embeddinggemma-300m-ONNX`
  model: reproduced the exact `<image_soft_token>` crash from the issue
  and confirmed the clamp path fixes it.

`ruff check` / `ruff format --check` clean. `pytest` scoped to the touched
modules (`test_embeddinggemma.py`, `test_embedding.py`, `test_config.py`,
`test_cli.py`) passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- **Opt-in clamp for out-of-range embeddinggemma token ids.** `mempalace mine --clamp-embedding-tokens` (or `embedding_clamp_token_ids` in `config.json` / `MEMPALACE_EMBEDDING_CLAMP_TOKEN_IDS`) clamps token ids past the model's `embed_tokens` row count (e.g. Gemma's `<image_soft_token>`) to the last valid row before `session.run()`, so the affected document gets a real embedding instead of falling through to the zero-vector, per-document skip fallback. Off by default; requires the optional `onnx` package (`pip install mempalace[clamp]`) to read the row count from the ONNX graph. (#2114)
+
 ---
 
 ## [3.6.0] — 2026-07-14
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -551,6 +551,9 @@ def cmd_mine(args):
     for raw in args.include_ignored or []:
         include_ignored.extend(part.strip() for part in raw.split(",") if part.strip())
 
+    if getattr(args, "clamp_embedding_tokens", False):
+        os.environ["MEMPALACE_EMBEDDING_CLAMP_TOKEN_IDS"] = "true"
+
     if getattr(args, "background", False) and not getattr(args, "daemon", False):
         print("mempalace: --background requires --daemon", file=sys.stderr)
         sys.exit(2)
@@ -568,6 +571,7 @@ def cmd_mine(args):
             "include_ignored": include_ignored,
             "max_chunks_per_file": getattr(args, "max_chunks_per_file", None),
             "redetect_origin": getattr(args, "redetect_origin", False),
+            "clamp_embedding_tokens": getattr(args, "clamp_embedding_tokens", False),
         }
         _submit_daemon_cli_job("mine", payload, args, background=getattr(args, "background", False))
         return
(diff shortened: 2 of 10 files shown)


belt:
Seems already fixed on `develop` branch. Nevermind

PR #2192, opened 2026-08-08
fix(service): restore MEMPALACE_PALACE_PATH after the call that stamped it

joshuafontany:
Each `run_*` entrypoint in `service.py` sets `MEMPALACE_PALACE_PATH` so downstream config resolution sees that call's palace, and none of them put it back.

In the daemon that costs nothing — one call, one process. In any process making two calls it leaks: `config.py` reads this variable with priority over the config file, so the first call's palace silently becomes the second caller's. A tmpdir palace since removed then reads back as `PalaceNotFoundError` from an unrelated code path, which is how this surfaced.

`daemon.py` already saves and restores exactly these keys around its own dispatch. This gives `run_mine`, `run_sync` and `run_diary_write` the same discipline through a small decorator.

Tests cover both directions — a caller whose variable was already set gets it back unchanged, and a caller who had none keeps none. All four fail with the decorators removed.

Service/config/palace selection: 541 passed, 4 skipped.

### DIFF
--- a/mempalace/service.py
+++ b/mempalace/service.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import contextlib
 import io
+import functools
 import os
 import sys
 from typing import Any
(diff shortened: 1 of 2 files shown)


igorls (approved):

Issue #2193, opened 2026-08-08
encoding_repair --apply destroys clean text: the [ÂÃ][continuation] window matches ordinary uppercase Portuguese/Vietnamese/Turkish prose


arnoldwender:
## Summary

`repair_mojibake` rewrites **clean** text when an uppercase `Ã` or `Â` is immediately followed by typographic punctuation. Each hit substitutes the letter *and deletes one character* of user content, and `repair_collection` reports it as a successful repair. On a palace containing no mojibake at all, `--apply` will silently modify drawers.

This collides with the project's first design principle — verbatim always — so I wanted to get it in front of you quickly rather than sit on it.

## Reproduction

```python
from mempalace.encoding_repair import repair_mojibake

for s in [
    "A palavra do dia e «MAÇÃ».",
    "O título e “IRMÃ” em maiúsculas.",
    "AMANHÃ… ninguém sabe.",
    "Sonuç: “HÂLÂ” geçerli.",
]:
    out = repair_mojibake(s)
    print(f"{'DAMAGED' if out != s else 'ok':>8}  {s!r} -> {out!r}")
```

Observed on `develop` @ `c7a026b`, Python 3.12:

```
 DAMAGED  'A palavra do dia e «MAÇÃ».'      -> 'A palavra do dia e «MAÇû.'
 DAMAGED  'O título e “IRMÃ” em maiúsculas.' -> 'O título e “IRMÔ em maiúsculas.'
 DAMAGED  'AMANHÃ… ninguém sabe.'           -> 'AMANHÅ ninguém sabe.'
 DAMAGED  'Sonuç: “HÂLÂ” geçerli.'          -> 'Sonuç: “HÂL\x94 geçerli.'
```

The closing `»` / `”` / `…` is consumed, not just the letter — so the result is one character shorter and, in the quoted cases, structurally unbalanced. The Turkish case replaces two visible characters with an invisible C1 control.

Driving it through the public API on a seeded collection with **zero** mojibake gives `{'scanned': 6, 'changed': 5, 'updated': 5}` — the tool considers the damage a successful repair, so nothing surfaces it to the user.

## Root cause

[...]

PR #2194, opened 2026-08-08
fix(repair-encoding-cli): reconfigure stdio to UTF-8 like the other entry points

arnoldwender:
## What and Why

`mempalace/_stdio.py` states the rule in its own docstring:

> Every console entry point that touches stdio needs to fix this on Windows -- the MCP server, the CLI, the fact_checker `--stdin` mode -- so the reconfigure code lives here in one place.

`cli.py:2160` and `fact_checker.py:321` both apply it. `scripts/mempalace_repair_encoding.py`, added in #2148, does not (`grep -c _stdio` = 0).

That gap lands hardest on this particular entry point. The tool exists for Windows users whose palace carries legacy mojibake, and `_print_change` emits a before/after preview for **every** proposed change. Under the console codepage the tool was written for, that preview cannot be encoded — the lead bytes of the corruption it detects are precisely the characters the codepage rejects — so the run dies before repairing a single drawer.

## Reproduction

With stdout on a legacy codepage, printing the same preview the tool emits:

```
UnicodeEncodeError: 'gbk' codec can't encode character '\xc3'
in position 14: illegal multibyte sequence
```

`U+00C3` is `Ã` — the lead character of the family this tool repairs.

**Expected:** the dry-run lists proposed changes and exits 0.
**Actual:** exits 1 on the first change containing a non-ASCII preview.

## Change Summary

- `scripts/mempalace_repair_encoding.py` — add `_reconfigure_stdio_utf8_on_windows()` (thin wrapper over the shared helper, mirroring `cli.py` / `fact_checker.py`) and call it first thing in `main()`.
- `tests/test_repair_encoding_cli.py` — new file, loading the script by path the way `tests/test_backfill_authored_at.py` does.

`replace` on stdout/stderr, matching the other two console entry points: the preview carries verbatim drawer text that may hold surrogate halves round-tripped from filenames, where `strict` would crash mid-preview.

## Test Plan

- [x] `test_reconfigures_stdio_to_utf8_on_windows` — mirrors the existing assertions in `test_cli.py` and `test_fact_checker.py`, pinning the per-stream errors policy.
- [x] `test_change_preview_survives_a_legacy_console_codepage` — reproduces the actual failure rather than asserting a call happened: a child process gets a legacy codepage for stdout and is asked to print a real mojibake preview.
- [x] **Mutant-checked.** Neutralising the reconfigure call turns the second test red with the genuine `UnicodeEncodeError: 'gbk' codec can't encode character '\xc3'`, so it is pinning the behaviour and not the wiring.
- [x] Full suite: **3861 passed, 31 skipped**.
- [x] `ruff check .` and `ruff format --check .` clean (213 files).

## Notes

Related to #1122 (same exception class, main CLI's help output) but not a duplicate: different entry point, and a different fix — that one can be solved with ASCII-safe static strings, whereas this one prints arbitrary user content and needs the stdio reconfigure.

Found while auditing the encoding-repair surface. I also filed #2193 for a separate, more serious defect in the same module — that one is a correctness question in `_HIGH_CONFIDENCE_RUN` and needs a design call, so I kept it out of this PR to keep this one trivially reviewable.

### DIFF
--- a/scripts/mempalace_repair_encoding.py
+++ b/scripts/mempalace_repair_encoding.py
@@ -134,7 +134,24 @@ def _print_change(
     )
 
 
+def _reconfigure_stdio_utf8_on_windows() -> None:
+    """Decode stdio as UTF-8 on Windows for the encoding-repair CLI.
+
+    Thin wrapper around the shared helper in ``mempalace._stdio``, matching
+    ``cli.py`` and ``fact_checker.py``. stdout/stderr override to ``replace``
+    because every proposed change prints a before/after preview of verbatim
+    drawer text -- under the legacy console codepage this tool is written for,
+    ``strict`` raises on the mojibake lead bytes themselves and aborts the run
+    before a single drawer is repaired.
+    """
+    from mempalace._stdio import reconfigure_stdio_utf8_on_windows
+
+    reconfigure_stdio_utf8_on_windows(stdout_errors="replace", stderr_errors="replace")
+
+
 def main() -> int:
+    _reconfigure_stdio_utf8_on_windows()
+
     parser = build_parser()
     args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


igorls (approved):
Wave 1 for 3.7.0: LGTM.

PR #2195, opened 2026-08-08
fix(repair): hold writer lease across rebuild_index

KeilerHirsch:
## Problem

`rebuild_index()` takes a point-in-time snapshot of the collection and then rebuilds/swaps the live index without holding the palace single-writer lease for the full operation.

A concurrent writer can therefore land after the snapshot and be absent from the rebuilt collection, immediately recreating SQLite/HNSW divergence.

## Fix

Hold `mine_palace_lock(palace_path)` across the complete snapshot → rebuild/swap → cleanup cycle.

The existing rebuild body is moved behind a small helper so the lock boundary stays explicit and the current rebuild/recovery flow is otherwise preserved.

## Tests

Added regression tests proving that:

- ChromaDB is not opened when the writer lease cannot be acquired.
- The rebuild body executes while the writer lease remains held.

Local validation:

- 2 new safety canaries passed
- 21 existing `rebuild_index` tests passed (`105 deselected`)
- `ruff check` passed
- `ruff format --check` passed

Base used for this delta: `c7a026bbd019b5db05d84a5a425b4459784f38c5` (`develop`).

### DIFF
--- a/mempalace/repair.py
+++ b/mempalace/repair.py
@@ -1167,7 +1167,31 @@ def rebuild_index(
         progress(index_read_recovery_guidance())
         return
 
-    backend = ChromaBackend()
+    # Hold the palace writer lease for the complete snapshot -> rebuild/swap
+    # -> cleanup cycle. A writer landing after the snapshot but before the
+    # rebuilt collection becomes authoritative would otherwise be lost from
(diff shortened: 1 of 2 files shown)


igorls (approved):

PR #2199, opened 2026-08-09
fix(extractor): ignore markdown emphasis as emotion

### LINKED ISSUE #2197: general_extractor.py: EMOTION_MARKERS false-positives on markdown-heavy text (`\*[^*]+\*` matches formatting syntax, not emotion)
## Summary

`general_extractor.py`'s `EMOTION_MARKERS` list contains `r"\*[^*]+\*"`, which matches markdown emphasis/bold syntax (`*text*`, and any substring of `**text**`), not emotional content. On real-world engineering conversation transcripts (Claude Code sessions, `--mode convos --extract general`), this single rule dominates the scoring and causes the vast majority of chunks to be misclassified as `emotional`, burying the `decision`/`problem`/`milestone` categories the extractor is presumably most useful for on this kind of corpus.

## Environment

- mempalace 3.6.0 (uv tool install)
- macOS, Python 3.11
- Source: `~/.claude/projects/<slug>/*.jsonl` (Claude Code session transcripts), `mine --mode convos --extract general`

## Root cause

`general_extractor.py`, `EMOTION_MARKERS`:

```python
r"\*[^*]+\*",
```

`ALL_MARKERS` scoring is `max(scores, key=scores.get)` — whichever category has the most keyword hits wins. Markdown-heavy technical writing (`**bold**` used for emphasis in explanations, decisions, etc.) triggers this regex repeatedly per chunk — `**text**` itself contains a matchable `*text*` substring, so a single bolded phrase can register multiple hits. The rule has no relationship to emotional content; it's matching formatting syntax.

## Minimal repro

```python
from mempalace.general_extractor import extract_memories

text = (
    "We decided to use PostgreSQL instead of MongoDB. "
    "**Key reasons:** better transactional guarantees and mature tooling. "
    "The migration script needs testing before **Friday's** deploy. "
    "This fixes the connection-pool exhaustion problem we saw last week."
)
chunks = extract_memories(text)
for c in chunks:
    print(c["memory_type"], "->", c["content"][:60])
```

With `EMOTION_MARKERS` as shipped, this text (a decision + a problem, zero emotional content, two `**bold**` spans) scores higher on `emotional` than on `decision` or `problem`, purely from the two bold spans. Remove `r"\*[^*]+\*"` from the list and it classifies correctly.

## Field measurement (real corpus, not just the synthetic repro)

Ran `mine --mode convos --extract general` (dry-run) against two real project directories under `~/.claude/projects/` (77 and 24 session files respectively).

Isolated test on a single 232k-character session file, classifying the same normalized content with `EMOTION_MARKERS` as shipped vs. with the `r"\*[^*]+\*"` rule removed (all other markers unchanged):

| | `emotional` share | decision / problem / milestone |
|---|---|---|
| Shipped `EMOTION_MARKERS` | 110/180 chunks (61%) | unaffected — identical hit counts with or without the rule |
| Minus asterisk rule | 31/180 chunks (17%) | unchanged |

Full-corpus check (two projects combined, 4552 chunks) after removing only that one rule, with per-marker attribution among the remaining `emotional` chunks:

```
158  i need
  6  i can't
  1  \bworried\b
```

Once the asterisk rule is gone, the residual `emotional` share (305/4552, 6.7%) is small and mostly attributable to `r"i need"`, which is also broad in an engineering-conversation context ("I need to check the logs") but nowhere near as dominant as the asterisk rule was. Final category distribution on this corpus after the fix: `decision` 43%, `milestone` 27.5%, `problem` 22.6%, `emotional` 6.7%.

## Suggested fix

Remove `r"\*[^*]+\*"` from `EMOTION_MARKERS`. It's the only rule in the list with no semantic relationship to emotion — every other entry (`\blove\b`, `i'm scared`, `never told anyone`, etc.) is a real word/phrase; this one is pure markdown syntax.

I did **not** touch `r"i need"` / `r"i can't"` in my local patch — they're broad, but I don't have evidence they dominate the way the asterisk rule did, and narrowing them risks cutting recall on the personal/diary-style conversations this extractor's marker set otherwise looks designed for (e.g. "I need you to know...", "I can't do this anymore"). Flagging them here as a secondary, much smaller observation, not proposing a specific fix.

## Related suggestions (lower confidence, filing here rather than as separate issues since they're all from the same investigation)

1. **`thinking` content blocks are silently dropped, for every `--mode convos` user, not just this bug.** `normalize.py`'s `_extract_content()` has branches for `text`/`tool_use`/`tool_result` but none for `thinking` — extended-thinking content in Claude Code transcripts is parsed by the JSONL loader but never reaches the chunker. On our corpus this was 2.6-12.4% of raw bytes across two projects, and qualitatively looked like the highest-signal content for anyone mining conversation history to recover past reasoning (which is presumably a common motivation for `--extract general`). Worth adding a branch that captures it, possibly gated behind a flag if there's a reason to keep it opt-in.
2. **`CONVO_SKIP_DIRS` could use `subagents` and `memory` alongside the existing `tool-results` entry.** `subagents/` (per-subagent worker transcripts) is the same noise category as `tool-results/` already documented in the code. `memory/` matters if anyone points `mine --mode convos` at a Claude Code project directory that also holds `~/.claude/projects/<slug>/memory/*.md` (per-project persistent memory files, if that mechanism is in use) — on our first dry-run pass, mining a project directory without excluding it pulled in already-curated memory notes as fake "conversations." Both are one-line additions to the existing `set` literal.
3. **`exclude_patterns` (`mempalace.yaml`) isn't respected by `--mode convos`, only by `--mode projects`.** Would be nice if convo mining could reuse the same per-project exclusion config rather than requiring a source patch to add directories like the above without maintaining a fork.

## Context

Related to the writer-lock discussion in #1972 — while testing that fix I ended up building a small pilot to mine our own Claude Code session history into a dedicated wing (motivation: recover past design reasoning/decisions that never made it into permanent docs), which is what surfaced this. As part of the same pilot, independently measured that `tool_use`+`tool_result` content accounts for 80-90% of raw transcript bytes on our two projects — consistent with what another commenter on #1972 (daproject85) described patching locally on their fork (stripping `tool_use`/`tool_result` formatting in `normalize.py`'s `_extract_content`). Not filing that as a separate issue since it sounds like an intentional design tradeoff (formatting tool calls into readable text is presumably useful for other consumers of `_extract_content`), but sharing the numbers here as corroboration.

Happy to share the full analysis script or open a PR for the one-line `EMOTION_MARKERS` fix if that's easier than triaging from this report.

KeilerHirsch:
## Summary

Remove the `r"\*[^*]+\*"` entry from `EMOTION_MARKERS` in `general_extractor.py`.

That regex matches Markdown formatting syntax rather than emotional content, so Markdown-heavy technical prose can accumulate enough `emotional` hits to outrank real decision/problem/milestone markers.

## Reproduction

Added two regression tests:

- Markdown emphasis/bold alone must score `0.0` against `EMOTION_MARKERS`.
- Markdown-heavy technical prose containing a real decision marker must classify as `decision`, not `emotional`.

Before the production change:

- `2 failed, 30 passed`
- two `**bold**` spans produced an `emotional` score of `3.0`
- the decision example was misclassified as `emotional`

The `3.0` score is worth calling out: the regex does not only match the two formatted spans; it can also bridge between adjacent `*` delimiters across ordinary text, making the false-positive behavior even broader than a one-hit-per-span interpretation.

## Fix

Delete only the Markdown asterisk marker from `EMOTION_MARKERS`. No other emotional markers or scoring behavior are changed.

## Verification

- focused extractor suite: `32 passed`
- extractor + convo miner regression suites: `122 passed, 3 skipped`
- `ruff check`: clean
- `ruff format --check`: clean
- `git diff --check`: clean

Fixes #2197

### DIFF
--- a/mempalace/general_extractor.py
+++ b/mempalace/general_extractor.py
@@ -159,7 +159,6 @@ EMOTION_MARKERS = [
     r"i need",
     r"never told anyone",
     r"nobody knows",
-    r"\*[^*]+\*",
 ]
 
 ALL_MARKERS = {
--- a/tests/test_general_extractor.py
+++ b/tests/test_general_extractor.py
@@ -310,3 +310,29 @@ def test_extract_memories_chunk_index_contiguous_across_segments():
     assert indices == list(range(len(memories))), (
         f"chunk_index must be 0..N-1 sequential; got {indices}"
     )
+
+
+def test_markdown_emphasis_does_not_score_as_emotional():
+    """Markdown formatting syntax must not count as emotional content."""
+    text = (
+        "**Key reasons:** transactional guarantees and mature tooling. "
+        "The rollout plan is documented under **Migration notes**."
+    )
+
+    score, _ = _score_markers(text, ALL_MARKERS["emotional"])
+
+    assert score == 0.0
+
+
+def test_markdown_bold_does_not_override_decision_classification():
+    """Markdown-heavy technical prose must not outrank a real decision marker."""
+    text = (
+        "We decided to use PostgreSQL for the new service. "
+        "**Key reasons:** transactional guarantees and mature tooling. "
+        "The rollout details are documented under **Migration notes**."
+    )
+
+    memories = extract_memories(text, min_confidence=0.1)
+
+    assert len(memories) == 1
+    assert memories[0]["memory_type"] == "decision"


igorls (approved):

PR #2200, opened 2026-08-09
fix(palace): reap orphaned per-source-file mine locks

fxinfo24:
## Problem

`_cleanup_mine_lock_file` reclaims a lock correctly on the happy path — but only for the specific lock a `mine_lock` context manager just released. A process that dies before reaching its own `finally` block (SIGKILL, force-quit, host crash) never runs that cleanup, and nothing else in the codebase later revisits the lock file.

Found in the wild: one long-lived installation had **5,636 stale lock files** in `~/.mempalace/locks/`, the oldest several months old, none held by any live process (confirmed via `lsof` before manual cleanup).

I checked this wasn't already covered before writing a fix:
- **#1264** (lock-holder-diagnostics, merged) — identifies *who* holds a live lock and fails loudly on contention. Doesn't reclaim orphans.
- **#1299** (mcp_server missing embedding_function) — different code path (MCP server's own collection-reopen), unrelated.
- The 2026-07-10 outage postmortem comment in `mcp_server.py`'s stdio loop — covers *graceful* client disconnection (stdin EOF/error → clean shutdown → the process's own lock-release runs). Doesn't cover *abrupt* process death, where that shutdown code never gets to run at all.

None of the three address orphan reclamation for a process that's simply gone.

## Fix

`reap_stale_mine_locks()` reuses `_cleanup_mine_lock_file` itself for the actual removal — same nonblocking-flock-reacquire safety mechanism, same Windows/POSIX handling already tested in this file, no duplicated locking logic. A lock is only ever removed after this process re-acquires it via the existing flock, so anything genuinely held by a live process is left untouched regardless of how old it looks by mtime — age is a courtesy throttle (avoids racing a lock that was *just* released and may still be mid-rendezvous with a waiter on the same pathname), not the safety mechanism.

Wired into `mine_lock()` via a throttled opportunistic call (`_maybe_reap_stale_mine_locks`, at most once per 15 minutes) rather than a new background thread, scheduled task, or CLI surface — it piggybacks on the natural cadence of mining rather than adding new infrastructure.

`mine_palace_*.lock` (the newer per-palace lock from the #974/#965 fan-out fix) is explicitly skipped — separate lifecycle, separate holder-identity tracking, not affected by this failure mode.

## Testing

6 new cases in `test_palace_locks.py`:
- removes a genuinely stale + unheld lock
- leaves a young lock alone regardless of hold state
- **core safety property**: a lock held by another process (real cross-process test via `multiprocessing`) is never removed, even when its mtime is backdated past the age threshold
- skips `mine_palace_*`-prefixed locks
- missing lock dir is a no-op
- the opportunistic throttle itself

Full existing `test_palace_locks.py` suite (19 tests) passes unchanged. Broader `tests/ -k 'palace or mine'` run clean — 730 passed, 10 skipped — aside from two pre-existing failures I confirmed are unrelated and present on an unmodified `develop` checkout before touching anything (`test_hnsw_capacity.py` SQLite WAL signature caching, `test_repair.py` FTS5 shadow-table write restriction — both environment/SQLite-build-specific, neither touches locking).

`ruff check` clean on both modified files.

## Compatibility

Purely additive — no existing function signatures changed, no new required config, no new CLI surface. The one behavioral change is that `mine_lock()` now does a cheap, throttled, best-effort directory scan at most once per 15 minutes; failures in that scan are swallowed (logged at debug level) and never propagate to the caller's actual mine operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 - **Embeddings via any OpenAI-compatible `/v1/embeddings` endpoint.** New `embedding_model: "openai-compat"` option computes embeddings on a server (LM Studio, llama.cpp, vLLM, Ollama's OpenAI shim, or a self-hosted endpoint) instead of a local ONNX model — useful for larger/multilingual embedders such as Qwen3-Embedding, or GPU offload. New `OpenAICompatEmbeddingFunction` in [`mempalace/embedding.py`](mempalace/embedding.py) speaks the standard `/v1/embeddings` protocol over stdlib `urllib` (no new dependency), batches requests, re-sorts the response by `index`, and L2-normalizes for the cosine collection. Endpoint settings are resolved by `MempalaceConfig` as a single source of truth — `embedding_api_url` / `embedding_api_model` / `embedding_api_key` in `config.json`, each overridable via the matching `MEMPALACE_EMBEDDING_API_*` env var. The embedding function's `name()` encodes the model id so changing it forces `mempalace repair rebuild-index` (different vector space). Mirrors the existing `openai-compat` LLM provider naming; stays local when the endpoint is on your machine/LAN. (#1559)
 
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


igorls (approved):

PR #2203, opened 2026-08-09
fix(deploy): survive the MCP idle watchdog exit in the systemd unit

### LINKED ISSUE #2201: systemd unit uses Restart=on-failure, but the MCP idle watchdog exits 0, so the server stays down
**What happened?**

`deploy/mempalace-server.service` sets `Restart=on-failure`, but the MCP idle watchdog terminates the process with `os._exit(0)`. systemd does not restart a unit that exits 0 under that policy, so a server installed from the template stops permanently after 8 idle hours and stays down until someone notices.

The chain is four lines, all on `develop` at `8516db7`:

1. `mempalace/mcp_server.py:5507` arms the watchdog in `_run_http_loop`, so this affects the HTTP transport and not just stdio.
2. `mempalace/mcp_server.py:333` sets `_MCP_IDLE_HOURS_DEFAULT = 8.0`.
3. `mempalace/mcp_server.py:5098` terminates with `os._exit(0)`.
4. `deploy/mempalace-server.service:27` sets `Restart=on-failure`, which ignores a clean exit.

Two details make it harder to catch than it should be. The first is that `/healthz` does not reset the idle timer: `_last_request_time` is assigned only inside `handle_request` (`mcp_server.py:4718`), and `do_GET` answers `/healthz` and returns before reaching it (`mcp_server.py:5305`). Since the operating notes in `website/guide/remote-server.md` recommend `/healthz` as a load balancer or Kubernetes liveness probe, an operator can be probing every few seconds and still have the server exit for inactivity, with the probe reporting healthy until the process is gone. The second is that `MEMPALACE_MCP_IDLE_HOURS` is not documented anywhere; `grep -rn "IDLE_HOURS" --include='*.md' .` returns nothing, so there is no knob an operator could have known to set. I have filed that separately as a docs issue.

The Docker path is not affected, because `deploy/docker-compose.server.yml` uses `restart: unless-stopped` on both services, which does restart after a clean exit. This is specific to the systemd template.

**What did you expect?**

That a server deployed from the project's own hardened unit template stays up, or restarts itself when the watchdog fires.

**How to reproduce:**

1. Install `deploy/mempalace-server.service` per the header comment, then `systemctl enable --now mempalace-server`.
2. Leave it with no MCP requests for 8 hours. Polling `/healthz` neither prevents nor delays this.
3. `systemctl status mempalace-server` reports `inactive (dead)` with result `success` and no restart. The journal shows `MCP server idle for 8.0 h (limit 8.0 h); exiting to release file handles.`

Setting `MEMPALACE_MCP_IDLE_HOURS=0.02` in `/etc/mempalace/server.env` reproduces it in about 72 seconds.

**Environment:**
- OS: traced from source on `develop` at `8516db7`; originally hit on macOS 15 (Darwin 25.5.0) running the HTTP transport under launchd, where `KeepAlive=true` masked it by restarting on any exit
- Python version: 3.12
- MemPal version: `develop` at `8516db7` (pyproject 3.7.0); the same code is in the released 3.6.0

**Suggested fix**

The narrow fix is `Restart=always` in the unit plus `MEMPALACE_MCP_IDLE_HOURS=0` in `deploy/server.env.example`, on the grounds that a dedicated always-on server is not the case the watchdog was written for. I have that ready and will open a PR shortly.

There is a broader question I would rather ask than assume. The watchdog's stated purpose is reaping the per-session stdio servers that accumulate ChromaDB and HNSW handles. That reasoning does not obviously carry over to a singleton HTTP service, so it may be that `_run_http_loop` should not arm it by default at all and the 8 hour default should stay stdio-only. That is a behaviour change in the core rather than a deployment fix, so per CONTRIBUTING I am raising it here first instead of sending it unsolicited. Happy to implement whichever you prefer.

Related: #1043 asks for remote MCP deployment and lifecycle documentation, and #1552 is where the watchdog originated.

rubicon:
## What does this PR do?

The idle watchdog terminates the MCP server with `os._exit(0)`. systemd does not restart a unit that exits 0 when the policy is `Restart=on-failure`, so a server installed from `deploy/mempalace-server.service` stops for good after the idle timeout rather than restarting.

This switches the unit to `Restart=always` and sets `MEMPALACE_MCP_IDLE_HOURS=0` in `deploy/server.env.example`. Both, deliberately. Disabling the watchdog is the correct default for this deployment shape, since a dedicated always-on server is not the per-session stdio case it was added for in #1552, and `Restart=always` still covers an existing installation whose env file predates this change.

Closes #2201.

The issue also raises a broader question that this PR intentionally leaves alone: whether `_run_http_loop` should arm the watchdog at all, given the stated purpose is reaping per-session stdio servers. That is a change in the core rather than in the deployment templates, so I asked there first instead of bundling it here.

## How to test

The default timeout is 8 hours, so shorten it. On a host running the unit, set `MEMPALACE_MCP_IDLE_HOURS=0.02` in `/etc/mempalace/server.env`, run `systemctl restart mempalace-server`, and leave it with no MCP requests for about 90 seconds. Polling `/healthz` does not count as activity, so it will not keep the timer alive.

Before this change the unit reaches `inactive (dead)` with result `success`, and does not come back. With `Restart=always` the process returns within `RestartSec=2`. With the env template unchanged from this PR, the watchdog never arms in the first place.

`systemd-analyze verify deploy/mempalace-server.service` will confirm the unit still parses.

## Checklist
- [x] No hardcoded paths

On the first box, I did not run the suite and would rather say so than tick it. This changes only `deploy/mempalace-server.service` and `deploy/server.env.example`, no Python, and `grep -rln 'deploy/\|mempalace-server.service\|server.env' tests/` returns nothing, so no test reads either file. `ruff check .` passes on the branch. Happy to run the full suite if you would like it on the record.

### DIFF
--- a/deploy/mempalace-server.service
+++ b/deploy/mempalace-server.service
@@ -24,7 +24,10 @@ User=mempalace
 Group=mempalace
 EnvironmentFile=/etc/mempalace/server.env
 ExecStart=mempalace serve --host 0.0.0.0 --port 8765
-Restart=on-failure
+# The idle watchdog exits with status 0, which on-failure reads as an
+# intentional stop, so the server would stay down. server.env disables the
+# watchdog; Restart=always covers a deployment that has not set it.
+Restart=always
 RestartSec=2
 
 # --- Hardening ---------------------------------------------------------------
--- a/deploy/server.env.example
+++ b/deploy/server.env.example
@@ -23,6 +23,13 @@ MEMPALACE_EMBEDDING_DEVICE=auto
 # MEMPALACE_MCP_TLS_CERT=/etc/mempalace/tls/cert.pem
 # MEMPALACE_MCP_TLS_KEY=/etc/mempalace/tls/key.pem
 
+# --- Server lifecycle -----------------------------------------------------------
+# The MCP server exits by itself after this many idle hours (default 8) to free
+# the ChromaDB/HNSW file handles that per-session stdio servers leave behind.
+# A dedicated always-on server does not need that, so 0 disables it. Note that
+# /healthz traffic does not count as activity; only MCP requests reset the timer.
+MEMPALACE_MCP_IDLE_HOURS=0
+
 # --- Palace location (systemd / bare-metal) -------------------------------------
 # In Docker the palace lives on the mempalace-data volume by default.
 # MEMPALACE_PALACE_PATH=/var/lib/mempalace/palace


igorls (approved):

PR #2204, opened 2026-08-09
docs: document MEMPALACE_MCP_IDLE_HOURS

### LINKED ISSUE #2202: docs: MEMPALACE_MCP_IDLE_HOURS is undocumented, including on the remote server guide
**What happened?**

`MEMPALACE_MCP_IDLE_HOURS` controls whether and when the MCP server terminates itself, and it is documented nowhere in the repository:

```
$ grep -rn "IDLE_HOURS" --include='*.md' .
(no matches)
```

It is read at `mempalace/mcp_server.py:698`, defaults to `8.0` hours (`:333`), and is armed on both transports, stdio at `:5472` and HTTP at `:5507`. When it fires the process calls `os._exit(0)` (`:5098`).

It is missing from both pages an operator would actually consult. `website/guide/remote-server.md` is the "Remote / Team Server" guide; its flag table and operating notes cover `--tls-cert`, `--read-only`, `--allow-insecure`, `MEMPALACE_MCP_HTTP_TOKEN`, `MEMPALACE_MCP_EXTRA_ALLOWED_HOSTS` and `MEMPALACE_HUB_FORWARD`, but not the one variable that decides whether the server is still running tomorrow. `website/guide/configuration.md` has an Environment Variables table listing four variables, and this is not one of them.

Comparing every `MEMPALACE_*` name read in `*.py` against those named in `*.md`, 30 are undocumented. Most are genuinely internal, but a few are operator-facing and cluster in this same remote-deployment story: `MEMPALACE_MCP_IDLE_HOURS`, `MEMPALACE_MCP_READ_ONLY` (which appears in no file outside `mempalace/mcp_server.py`), `MEMPALACE_MCP_TLS_CERT` and `_KEY` (present in `deploy/server.env.example` and `deploy/mempalace-server.service`, but in neither guide), and `MEMPALACE_MCP_WRITE_STALL_WARN_SECS` and `_EXIT_SECS`. I have scoped this issue to the idle variable, since that is the one with a failure mode attached, but the rest may be worth a sweep.

**What did you expect?**

That a variable which can silently stop a long-lived server is documented on the page explaining how to run a long-lived server.

**Why it matters**

Two properties make the undocumented default surprising rather than merely undocumented.

`/healthz` does not reset the idle timer. `_last_request_time` is assigned only in `handle_request` (`:4718`), and `do_GET` answers `/healthz` and returns before it (`:5305`). The operating notes recommend `/healthz` as a liveness probe, so the documented health-check pattern does not keep the server alive and gives no warning before it exits.

The exit status is `0`, so supervisors configured to restart only on failure treat it as an intentional stop. That is currently live in the project's own systemd template, which I have filed separately.

**Suggested fix**

Add the variable to the Environment Variables table in `configuration.md`, and add an idle shutdown entry to the operating notes in `remote-server.md` covering the default, that `0` disables it, that `/healthz` traffic does not count as activity, and that supervisors need to restart on a clean exit. I have this ready and will open a PR shortly.

This also partly answers #1043, which asks for remote MCP deployment and lifecycle documentation; the lifecycle question raised there is the one this variable governs.

rubicon:
## What does this PR do?

Documents `MEMPALACE_MCP_IDLE_HOURS`, which decides whether a long-lived server is still running tomorrow and appeared in no markdown in the repo.

Two changes. `website/guide/configuration.md` gains a row in the Environment Variables table. `website/guide/remote-server.md` gains an idle shutdown entry in the operating notes, covering the 8 hour default, that `0` disables it, and the part that is easiest to get wrong: the watchdog exits `0`, so a supervisor set to restart only on failure will leave the server down.

It also adds one clause to the existing health check entry. `/healthz` is recommended there as a load balancer or Kubernetes liveness probe, but probe traffic does not reset the idle timer, because `_last_request_time` is assigned only in `handle_request` and `do_GET` answers `/healthz` before reaching it. Without that clause an operator can reasonably read the current text as meaning a probed server stays alive.

Closes #2202.

## How to test

Prose only, so reading it is the test. `npm run docs:dev` in `website/` renders both pages; the new table row and the operating notes entry should read cleanly, and the `/guide/remote-server#operating-notes` link from the configuration table should resolve to the right anchor.

## Checklist
- [x] No hardcoded paths

On the first box, I did not run the suite. This changes only `website/guide/configuration.md` and `website/guide/remote-server.md`. The one test that reads the docs tree, `tests/test_readme_claims.py`, is scoped to `website/reference/mcp-tools.md` and `website/reference/modules.md`, so it does not cover either file. `ruff check .` passes on the branch. Glad to run the suite if you would prefer it recorded.

### DIFF
--- a/website/guide/configuration.md
+++ b/website/guide/configuration.md
@@ -172,3 +172,4 @@ python -m mempalace.mcp_server --palace /custom/palace
 | `MEMPAL_DIR` | Directory for auto-mining in hooks |
 | `MEMPALACE_MAX_BACKUPS` | Override `max_backups` retention count (`0` disables pruning) |
 | `MEMPALACE_BACKEND` | Select the storage backend (default `chroma`) — see [Storage backends](#storage-backends) for each backend's connection variables |
+| `MEMPALACE_MCP_IDLE_HOURS` | Hours with no MCP request before the server exits by itself (default `8`; `0` disables). See [Remote server](/guide/remote-server#operating-notes) |
--- a/website/guide/remote-server.md
+++ b/website/guide/remote-server.md
@@ -202,7 +202,15 @@ messages between machines.
   server state, `GET /statusz` returns JSON with version, uptime, request
   counters, SQLite integrity, writer mode, and recent observed MCP clients.
   `/statusz` follows the bearer-token policy because it exposes operational
-  metadata; it is not a public liveness probe.
+  metadata; it is not a public liveness probe. Probe traffic does **not**
+  count as activity for the idle watchdog below; only MCP requests do.
+- **Idle shutdown**: the server exits by itself once `MEMPALACE_MCP_IDLE_HOURS`
+  have passed with no MCP request (default `8`). That default is there to reap
+  the per-session stdio servers that would otherwise pile up holding ChromaDB
+  and HNSW file handles, which is not the case a dedicated always-on server is
+  in. Set `MEMPALACE_MCP_IDLE_HOURS=0` to disable it, and make sure your
+  supervisor restarts on a *clean* exit (`Restart=always` rather than
+  `Restart=on-failure`), because the watchdog exits `0`.
 - **Fronting proxies (Tailscale, nginx)**: the recommended personal-fleet
   setup is a loopback bind behind a tailnet-only proxy — nothing touches the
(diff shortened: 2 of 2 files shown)


igorls (approved):

PR #2208, opened 2026-08-10
fix(encoding): require local evidence before repairing the ambiguous [ÂÃ] window (#2193)

### LINKED ISSUE #2193: encoding_repair --apply destroys clean text: the [ÂÃ][continuation] window matches ordinary uppercase Portuguese/Vietnamese/Turkish prose

arnoldwender:
## What and Why

`repair_mojibake` rewrites text that was never damaged, and reports it as a successful repair. Anyone who has run `mempalace repair-encoding --apply` on a palace containing Portuguese, Vietnamese or Turkish prose has lost characters.

```
"«MAÇÃ»."           ->  "«MAÇû."          two characters collapsed into one
"“IRMÃ”"            ->  "“IRMÔ"
"İMÂ… edildi."      ->  "İM edildi." visible text replaced by an invisible control
```

## Root Cause

`encoding_repair.py` — the first alternative of `_HIGH_CONFIDENCE_RUN` is a **two-character** window:

```python
rf"[ÂÃ][{_CONTINUATION_CLASS}]"
```

Portuguese, Vietnamese and Turkish end all-caps words in Ã/Â — `IRMÃ`, `MAÇÃ`, `MANHÃ`, `AMANHÃ`, `BÃO`, `NHÃ`, `HÂLÂ`, `IMÂ`. Prose then puts a closing quote, guillemet, ellipsis or dash straight after, and **all 14 of those characters are inside `_CONTINUATION_CLASS`** (CP1252 bytes 0x80–0xBF). So clean text matches the same shape as mojibake, decodes as UTF-8, and two characters become one.

Three distinct failures share this root cause:

1. **Clean text destroyed** — the cases above.
2. **Visible text replaced by invisible controls.** `Â` + 0x80–0x9F decodes to exactly the C1 control block, so the repair swaps a character you can see for one you cannot.
3. **Not idempotent.** `repair_mojibake` runs up to `max_passes=3`. Genuinely mojibaked Portuguese is repaired *correctly* on pass 1 and then destroyed on pass 2:

```
damaged : â€œIRMÃƒâ€ Ã© o tÃ­tulo.
pass 1  : “IRMÃ” é o título.     <- correct
pass 2  : “IRMÔ é o título.      <- destroys its own correct output
```

The existing test corpus never caught this because every clean string in it puts `Ã` before a **letter** (`CORAÇÃO`, `São`) and never before punctuation.

## The Fix

This follows the precedent already in the file. `Ä`/`Å` were excluded as leads with the comment *"strings such as Å² can be legitimate scientific text"* — the identical argument, applied to the continuation side.

A **lone** `[ÂÃ]` window ending in typographic punctuation is no longer treated as high-confidence. It is repaired only when a lowercase letter runs directly into the lead — which is where mojibake actually occurs, *inside* a word:

| | |
|---|---|
| `coÃ»te`, `NoÃ«l` | lowercase before the lead → **repaired** |
| `IRMÃ”`, `MAÇÃ»`, `HÂLÂ»` | all-caps word boundary → **left alone** |

Unaffected: chained units (`aÃ§Ã£o`, `CORAÃ‡ÃƒO`) and the 3–4 character `â`/`ð`/`ï` windows, which no clean text can produce. Separately, a decode product containing a C0/C1 control is now always refused — that costs nothing, since natural text contains no C1 controls.

**Corroboration is deliberately local.** Four different approaches I tried inferred "this drawer is mojibake" from a run found elsewhere in the same text, and every one of them then destroyed clean prose in a **mixed** drawer. Drawers are mixed by construction — the miner concatenates several sources into one — so `cafÃ© e «MAÇÃ».` must repair the first half and preserve the second.

## Test Plan

- [x] 23 new assertions, verified **RED before the fix / GREEN after** (`git stash` on the module, tests kept)
- [x] `tests/test_encoding_repair.py`: 70 passed
- [x] Full suite: **3929 passed, 31 skipped**
- [x] `ruff format --check .` + `ruff check .` clean (212 files)
- [x] Exhaustive sweep: **14,040** generated clean strings — every `Ã`/`Â`-final word × 13 prose punctuation marks × 6 quote wrappers × 9 sentence frames — **0 false positives**
- [x] Idempotence checked on every corpus string: `repair(repair(x)) == repair(x)`

Tests added: clean uppercase-lead prose (16 cases across pt/vi/tr), mixed drawers (5), never-emit-control-characters, does-not-destroy-its-own-output, and a guard that evidence-backed repairs still work.

> One unrelated flake: `test_antigravity_hooks_shell.py::test_save_hook_rejects_non_jsonl_transcript_path` failed on one full-suite run (that run took 338s vs 115s — the machine was loaded and the test spawns subprocesses). It passes in isolation and in its own file on **both** this branch and clean `develop`, and the full suite is green on re-run. Not related to this change, but reporting it rather than hiding it.

## What this deliberately gives up

Measured against `develop` on a 30-string multilingual corpus: **21/30 repairs are unchanged, 6 are given up, and 0 clean strings that `develop` preserved are corrupted by this change.**

The 6 are words whose accented capital is **word-initial**, so there is no lowercase letter to corroborate:

```
Österreich · Übermäßig süß — Öl und Äpfel. · Smörgåsbord, Ångström and Øresund.
La canción «PERÚ» abre el disco. · L'Àngels diu: «això és català». · HÂLÂ bekliyoruz.
```

I chose this knowingly, because **the costs are asymmetric**: a missed repair leaves the drawer byte-identical and it can be repaired again later by a smarter rule, while a wrong repair silently destroys good text and is not recoverable from the result. For a tool whose own commit history says *"make repair conservative and reversible"*, refusing an ambiguous case is the correct default.

If you would rather recover some of those, the natural follow-up is a second evidence source (word-initial runs followed by lowercase). I prototyped it and **rejected it**: it reintroduced 180 false positives in the sweep above, which is exactly the trade this PR exists to refuse.

Closes #2193

### DIFF
--- a/mempalace/encoding_repair.py
+++ b/mempalace/encoding_repair.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import os
 import re
+import unicodedata
 from pathlib import Path
 from typing import Callable, Iterator, Optional, TextIO, Union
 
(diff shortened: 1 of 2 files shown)


igorls (approved):

Issue #2144, opened 2026-08-03
repair --dry-run performs the real rebuild in the default (legacy) mode


mvalentsev:
## What happened?

`mempalace repair --dry-run` with no `--mode` runs the real rebuild. It copies the whole palace over any existing `<palace>.backup`, re-files the drawers collection through a staged temp copy, rebuilds the FTS5 index and VACUUMs `chroma.sqlite3`.

#2095 and #2133 fixed this for `--mode from-sqlite`. The default path was not covered, and it is the one you get by typing `mempalace repair --dry-run`.

Measured on a throwaway 3-drawer palace, on `develop` at `511ead9a`:

```
sqlite sha before: 08106c652f1e39a0
sqlite sha after : 07c720458a3da426
```

`<palace>.backup/` appears and the run prints `Backup saved at ...`. Same on the released 3.6.0.

Two smaller things in the same place:

- The `--dry-run` help still reads `(--mode max-seq-id only)`. That stopped being true when #2133 taught `from-sqlite` to honour the flag.
- Reading the preflight, `maybe_autoheal_fts5_index` also runs under `--dry-run`, and it rebuilds the inverted index in place. I did not reproduce that one; it needs a corrupted FTS5 index.

## What did you expect?

`--dry-run` prints what the repair would do and leaves the palace alone, whichever mode is selected. #2133 put it plainly: "`--dry-run` should be honored on every repair path".

## How to reproduce:

1. `mempalace --palace /tmp/p mine <any-dir> --wing probe`
2. `sha256sum /tmp/p/chroma.sqlite3`
3. `mempalace --palace /tmp/p repair --dry-run --yes`
4. `sha256sum /tmp/p/chroma.sqlite3` now differs, and `/tmp/p.backup/` exists

## Environment:
- OS: Linux (WSL2, Ubuntu)
- Python version: 3.12.3
- MemPal version: 3.6.0 (uv tool install), and `develop` at `511ead9a` (reports 3.7.0)

PR #2209, opened 2026-08-10
feat: add JSONL export/import for cross-device palace sync

### LINKED ISSUE #452: feat: cross-device sync — export/import palace for multi-PC workflows
## Problem

Developers often work across multiple machines (home PC, office PC, laptop). Each machine has its own local palace with different conversation memories. Design decisions made during a session on one machine are invisible on another.

This defeats the core value proposition of mempalace — giving AI persistent memory across sessions.

## Recommended Solution: Private Git Repo

Export palace to JSONL (portable text format) and sync via a private GitHub repo:

```bash
# After a coding session — export and push
mempalace export --output ~/.mempalace/export/
cd ~/.mempalace/export && git add . && git commit -m "session $(date +%F)" && git push

# On another machine — pull and import
cd ~/.mempalace/export && git pull
mempalace import --from ~/.mempalace/export/
```

### Why git repo?

| Benefit | How |
|---------|-----|
| Backup | git history (every commit is a snapshot) |
| Sync | git push/pull between machines |
| Version history | git log shows when memories were added |
| Recovery | git checkout to any point in time |
| Cost | Free (GitHub private repo) |
| Size | ~15 MB for 6 months of daily AI use — well under GitHub's 1 GB recommendation |

### Export format (JSONL, git-friendly)

```
~/.mempalace/export/
├── rebuild/
│   ├── architecture.jsonl
│   ├── technical.jsonl
│   └── planning.jsonl
└── trello/
    └── general.jsonl
```

Each line is one drawer with content, metadata, and content-hash ID for conflict-free merging.

### Import behavior

- **Merge, not overwrite** — adds new drawers, skips existing (by content hash)
- **Idempotent** — safe to import the same export multiple times
- **Cross-version** — not tied to a specific ChromaDB version

## Alternative: Binary Backup for Fast Restore

For full palace restore without re-embedding:

```bash
mempalace backup --zip    # ~30-60 MB for 10k embeddings
# Upload zip to cloud storage
# On target: unzip and use directly
```

Not suitable for incremental sync (binary diffs are inefficient), but useful for disaster recovery.

## Why NOT cloud sync (OneDrive/Dropbox) directly?

ChromaDB stores interdependent binary files (SQLite + HNSW index). Cloud sync tools transfer files individually and non-atomically — partial sync corrupts the HNSW index silently. This was the root cause of #444.

## Related

- #444 — Global palace corruption
- #448 — Backup command (complementary feature)

0xdhx:
# feat: add JSONL export/import for cross-device palace sync

## What does this PR do?

Implements the recommended solution from #452: a portable, git-friendly export/import pair so a palace can be synced between machines through a private git repo.

- **`mempalace export`** (default `--format jsonl`) streams every drawer into a JSONL tree organized by wing/room (`<out>/<wing>/<room>.jsonl`, plus an `export-manifest.json` with a format version and counts). Output is deterministic — drawers sorted by id, JSON keys sorted, no timestamps — so re-exporting an unchanged palace produces a byte-identical tree and an empty git diff. `--output` defaults to `export/` inside the config directory, so it follows the XDG-aware config location from #148 rather than a fixed `~/.mempalace`. `--format markdown` exposes the existing `exporter.export_palace` (previously library-only) on the CLI. Both formats open an existing palace read-only and without create (`create=False, read_only=True`, the pairing the searcher uses). Through the CLI a path holding no palace is refused (`No palace found`, exit 1); called directly, the exporter keeps its long-standing behaviour there and exports empty. A refused output path (a symlinked target, an unwritable directory) is reported with exit 1 rather than a traceback.
  - **Fix to the previously pushed head:** JSONL export asked for `read_only=True` but left `create` at its default, and sqlite_exact refuses a read-only open that may create (`sqlite_exact read-only connections cannot create a palace`) — so exporting a sqlite_exact palace raised before writing anything. It exports now, and the read-only open is pinned by a test for both formats on both local backends. Chroma currently ignores the `read_only` option (its `get_collection` reads only `caller_vectors` from the options), so on Chroma the flag is a request the backend does not act on yet.
- **`mempalace import <dir>`** merges an export into a palace: adds drawers whose ids are absent, skips existing ones, and is idempotent on re-import. `--dry-run` is a parse-only true preview — it never opens (or creates) the palace at all, since opening a local backend is itself a write (the same principle as the #2144/#2133 dry-run fixes). Malformed input (bad JSON, pathological nesting, non-UTF-8 files) is counted and reported, never fatal; a concurrent insert of the same id degrades to a skip rather than an abort. Imports into a missing palace create it, so first import on a fresh machine works. A palace already being written by a mine or an MCP server refuses the import with the holder named and exit 1, as `mine` and `sync` do.
  - **Metadata round trip, stated:** import keeps flat scalar metadata values (`str`/`int`/`float`/`bool`) — the contract mempalace's own writers follow (`DrawerRecord` in `mempalace/sources/base.py`). The local backends can store more (Chroma accepts non-empty lists of one scalar type; sqlite_exact stores any JSON value, including null), and export writes whatever the palace holds, so a palace written outside that contract does not round-trip those values: they are dropped, counted in `metadata_dropped`, and reported in the import output — never silently.
- **`import` follows the CLI write-routing policy from #2033**, the same way `mine`, `sweep` and `sync` do: under `prefer`/`require` (or `--daemon`) it runs as a daemon `import` job instead of opening a direct writer, `--direct` forces the in-process path, and `--background` returns the job id. `--dry-run` travels in the job payload exactly as it does for `mine`, and a dry run never opens the palace wherever it runs. A held writer lease defers the job rather than failing it. Import dedups by drawer id, so a job the daemon re-queues after a crash cannot duplicate content. The default policy is still `direct`, so default behaviour is unchanged; `import` is added to the routed-command list in `docs/cli-write-routing.md`.
- Embeddings are deliberately **not** exported: they are large, binary, and embedder-specific. Import re-embeds locally and says so in its output — the runtime warning requested in the #453 review.

Scope is deliberately the narrow v1 from the issue discussion (idempotent hash-id dedup). Out of scope, per that thread: metadata-conflict detection for moved drawers, near-duplicate detection for divergent edits (#452 comments), binary backup (#448), and KG export (previously #435/#499). One known v1 caveat, stated rather than hidden: import accepts any `*.jsonl` tree without requiring `export-manifest.json` (tolerant by design; the manifest is informational in format v1). That tolerance is also why re-export prunes room files it did not write — otherwise a room whose last drawer was deleted survives as a stale file and the next device re-imports it, resurrecting the drawer. The prune is gated on an `export-manifest.json` that existed before the run, so exporting into an unrelated directory deletes nothing, and it is skipped on a zero-drawer palace (which warns instead, since zero drawers is also what the exporter reports, called directly, for a path holding no recognizable palace). The `--format` enum currently covers `jsonl` (round-trippable) and `markdown` (one-way, browsable); new formats extend the same seam.

## How to test

```bash
# Round trip: export from one palace, import into a fresh one (--palace is a global flag, before the subcommand)
mempalace export --output /tmp/palace-export
mempalace --palace /tmp/fresh-palace import /tmp/palace-export
mempalace --palace /tmp/fresh-palace import /tmp/palace-export   # idempotent: 0 imported

# Or the automated coverage
uv run pytest tests/test_export_import_jsonl.py tests/test_exporter.py tests/test_cli_package.py tests/test_cli_write_routing.py -v
```

New tests cover: export structure + byte-level determinism across two runs, full round-trip content fidelity, idempotent re-import, malformed-line tolerance, dry-run creating nothing, and non-directory input rejection. Later review rounds add coverage for the non-regular-file/FIFO guards, symlinked-directory containment, non-scalar metadata reporting, stale-export pruning with its first-run and zero-drawer gates, and the import writer lease. The CLI handlers live in their own `mempalace/cli/cmd_export.py` fragment, and one test drives `export` then `import` end to end through `mempalace.cli`, so a missing fragment registration, dispatch entry or package-relative import fails it. The routing is covered in `tests/test_cli_write_routing.py` alongside the #2033 tests it mirrors: daemon submission under `prefer` with an absolute source, the direct path under `direct`, a dry run routed with its flag, the parser flags, `run_import` (dry-run forwarding, a relative source refused, the lock refusal class), and `execute_job` dispatch. The read-only export open is pinned for both formats on both local backends (chroma, sqlite_exact), and export refusals through the CLI for both formats. Each of those fixes carries a test that fails when the fix is reverted. For this revision's changes (the routing, the default output directory, the lease and export refusals, the read-only open and its no-palace fallback) that was re-checked at the pushed head: 14 reversion controls, each reverting one change on its own, and each fails a named test.

## Verification

Run locally at `ee884c6`, rebased on `develop` at `38260df`, with each CI job's own command:

- `Tests / lint`: ruff 0.16.6 (the version `ci.yml` pins), `ruff check .` and `ruff format --check .` pass.
- `Tests / test-linux`: the job's pytest command, coverage gate included. Python 3.11 and 3.13: 5367 passed, 53 skipped each, coverage 82.54% (gate 80%). Python 3.9: 5366 passed, 54 skipped, coverage 82.48%.
- `Tests / test-windows`: **not green locally.** On a local Windows machine (Python 3.13, the job's pytest command), five full runs at this head each failed one or two tests (4989–4991 passed per run): six failures, six different tests. Three were multiprocessing lock tests whose child process did not take its lock in time, two were update-awareness background-check deadlines, and one was a Windows socket abort (`WinError 10053`) after a correctly rejected request. Three runs of `develop` at `38260df` on the same machine with the same command gave two green runs (4960 passed) and one with two failures, one of them the same socket abort. Apart from the one test later deselected (below), every test that failed at this head passes in isolation on that machine, 5 of 5 at `develop` and 5 of 5 at this head. None of them, and none of the lock, update-awareness or MCP HTTP code they exercise, is changed by this PR, and GitHub's `windows-latest` passes all of them at `38260df`. GitHub's own Windows run on this PR is the result that counts. Four tests are deselected in these local runs (`test_chroma_collection_lock.py::test_concurrent_writers_serialize`, `test_palace_locks.py::test_reentrant_same_thread_passes_through`, `test_palace_locks.py::test_lock_failure_message_names_holder`, and, from the fourth run on, `test_update_awareness.py::test_in_flight_check_cannot_restore_withdrawn_consent`): on this machine they fail the same way at `develop` without this PR (the last one intermittently, 2 of 5 isolated runs), while `windows-latest` passes all four there.
- `Tests / Test Rust Core & CLI (ubuntu-latest)`: stable Rust 1.98.1, `cargo fmt --all --check`, `cargo check --workspace --locked`, `cargo test -p mempalace-core -p mempalace-cli --locked` (9 passed) and the job's pytest step (129 passed).
- `Docker / smoke` and `Docker / build`: amd64 buildx. `scripts/docker-smoke.sh` reports all checks passed (MCP `tools/list` returns 45 tools); the build runs with provenance and SBOM and no push (local builder, no GHA cache).
- Not run locally: `test-macos` and the Rust job on `macos-latest` (no macOS machine), and the Rust job on `windows-latest` (no Rust toolchain on the Windows machine).
- Each of the 15 commits passes the full suite on its own (Python 3.13), and no test that runs at the head is skipped at any earlier commit.

First firings on Linux, stated rather than folded in: while several suites shared the machine, three timing-sensitive tests failed on a first run: `test_mcp_http_transport.py::TestPalaceReadsDoNotStarveTheHub::test_state_changes_wait_for_in_flight_search[mempalace_add_drawer]` (Python 3.9), `test_backend_header_probe_locks.py::test_probes_keep_chroma_writer_shared_lock` (two of the per-commit runs) and `test_chroma_collection_lock.py::test_concurrent_writers_serialize` (one of those two). Each passed on the serial re-runs reported above.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) on Linux (3.9, 3.11, 3.13); not green on a local Windows machine (see Verification); macOS not run locally
- [x] No hardcoded paths

Fixes #452


0xdhx:
Rebased onto `develop`. This was `CONFLICTING` against a base that had moved 122 commits since the branch point; the only conflict was `CHANGELOG.md`, with `cli.py` and `exporter.py` merging cleanly. Both sides of the changelog are kept, `### Features` before `### Bug Fixes` to match the ordering used in 3.7.0 / 3.6.0 / 3.5.0.

Reconciling against the new base turned up that the importer predated the non-regular-file work in #2221 / #2207 / #2244 and had exactly the defect those fixed, so there are two hardening commits on top of the rebase.

**`import_palace` could block forever on a non-regular file.** It opened every `*.jsonl` glob hit with a plain `open()`. That is fine for the cases the surrounding handler named — a directory, bad permissions, bad UTF-8 — because each raises. A FIFO does not: opening one for reading parks in the kernel until a writer appears, so a named pipe carrying a `.jsonl` name wedged the import, and the enclosing `except (OSError, UnicodeDecodeError)` could never see it. Now guarded the way `miner._read_text_no_follow` does it — `O_NONBLOCK` to make the `S_ISREG` check reachable, `O_NOFOLLOW`, and the `EAGAIN` retry that preserves the blocking read for a regular file whose write lease was broken.

**Import reads are contained to the import tree.** `O_NOFOLLOW` guards only the final path component, and `glob.glob(..., recursive=True)` traverses symlinked *directories* — so a regular file outside `input_dir`, reached through a symlinked subdirectory, was imported as if it belonged to the export. `_path_within_root` closes that, mirroring the containment test `_read_text_no_follow` pairs with its no-follow open. Stated plainly because it is a real limit rather than a complete fix: the check is resolve-then-open, so it stops a symlink *sitting* in the tree, not one swapped in between the check and the open. Closing that race properly needs per-component `openat` walking, which felt disproportionate for v1 — and an import source you do not control is outside the feature's threat model anyway. Happy to do it if you disagree.

**Non-scalar metadata is now reported rather than dropped silently.** The filter's premise was that the backend cannot store non-scalars. That holds for Chroma and not for the palace generally — `sqlite_exact` serializes metadata through an unrestricted `json.dumps`, so a palace on that backend can hold a list or a nested dict, the exporter writes it out raw, and the importer discarded it without a word. The values are still dropped (this import path stores scalars only) but they are counted in a new `metadata_dropped` stat and reported. A `metadata` that is not an object at all is discarded whole and counts as one drop; an absent `metadata` key is not a drop.

Each fix has a test that fails when the fix is reverted — four controls, all four fire. The FIFO test bounds the hang with `SIGALRM` and deliberately raises a non-`OSError`, because `TimeoutError` *is* an `OSError` subclass and the importer's own handler swallowed the first version, making the test pass against unfixed code.

Full suite on the rebased branch: **4303 passed, 6 failed, 31 skipped**. All 6 failures reproduce on a clean `develop` checkout at `06cb698` with this PR absent — 5 parametrisations of `test_init.py::test_init_filters_sys_path_from_leaked_pythonpath` and `test_mcp_server.py::TestStaleLibraryGate::test_startup_baseline_survives_module_reload` — so they look pre-existing rather than anything this branch introduces. Flagging them rather than touching them here.

Two things I did **not** change, both design calls rather than defects, and both yours to make:

1. The body already notes that re-exporting into an existing tree does not delete files for rooms that no longer exist. Worth sharpening: because import walks every `*.jsonl` without consulting `export-manifest.json`, that is not only a cosmetic git-diff wrinkle — deleting the last drawer in a room and re-exporting can leave a stale file that another device then re-imports, resurrecting the drawer. Either pruning on export or having import honour the manifest would close it.
2. `import_palace` opens the collection for writing without taking `mine_palace_lock`, which the adapter write path holds. The module docstring claims the same single-writer expectations as `mine`, so that is a doc/behaviour mismatch either way. Relatedly, `export_palace_jsonl` calls `get_collection()` with the default `create=True, read_only=False` despite being a pure read.

Happy to take any of these in this PR or a follow-up, whichever you prefer.

One process note: after the push the PR is `MERGEABLE` but sits at `UNSTABLE` with **zero** check runs reported. I read that as the workflow runs still awaiting approval for an outside contributor rather than anything failing — so the rebase clears the conflict but will not turn CI green on its own. Let me know if there is something I should do at my end instead.

0xdhx:
Went ahead and did both of the two design calls I flagged at the end of the last comment, one scoped commit each — so either can be dropped or split into a follow-up without touching the other, if you'd rather keep this PR narrow.

**Re-export now prunes stale room files** (`ab9e7f8`). This is the one I'd argue is a defect rather than a wrinkle: because import walks every `*.jsonl` without consulting `export-manifest.json`, deleting the last drawer in a room and re-exporting left a stale file that the next device re-imported, resurrecting the drawer. Two gates on the deletion, both deliberate:

- Pruning is gated on an `export-manifest.json` that existed **before** this run, captured before the manifest is rewritten. Exporting into an unrelated directory therefore deletes nothing on the first run — the gate is proof the tree is one of ours, not merely that it now contains our manifest.
- Only regular files are unlinked; symlinks are skipped rather than followed, matching the `_reject_symlink` posture already applied on the write side.

The zero-drawer path deliberately does **not** prune, and says so. An empty count is also what a palace that failed to open looks like, and silently deleting a good export on that reading is worse than leaving a stale one behind — so it warns and leaves the tree alone.

**`import_palace` now holds the writer lease for the whole walk** (`0e710c1`). The module docstring already claimed the same single-writer expectations as `mine`, but nothing enforced it, and the existing-id check and the add are not one atomic operation — a concurrent miner could file an id between them. The lease spans the whole walk rather than just the final flush, because batches are added as files are read; the walk moved into `_import_into` so that scope is explicit rather than an eighty-line reindent. A dry run still takes no lease and opens nothing, since acquiring one would create the palace directory the preview promises not to touch.

Same commit also opens the collection `read_only=True` in `export_palace_jsonl`, since export is a pure read. `create` is left at its default — flipping it changes what exporting a not-yet-existing palace does, which is a separate call from this one.

Four new tests (pruning, the first-export gate, the zero-drawer warning, and that a real import takes the lease while a dry run does not), each with a reversion control that fires against the unfixed code. Full suite on the branch: **4307 passed, 6 failed, 31 skipped** — the same 6 pre-existing failures I flagged last time, unchanged and unrelated.

PR body updated: the caveat about re-export leaving stale files is no longer true, so it has been replaced rather than left standing.

0xdhx:
Rebased onto current `develop`. The branch was `CONFLICTING` again; GitHub reports it mergeable as of this push.

`develop` had moved **167 commits** past the branch point since the last rebase. Of the five files this PR touches, `develop` had changed two, and only one of them conflicted:

- **`CHANGELOG.md` — conflicted.** 3.7.1, 3.8.0 and 3.9.0 have all released since this PR opened, so the entry no longer belonged where it was written. It now sits under `## [Unreleased]` → `### Features`, after the shared-brain rules entry already there. Resolved so that nothing of `develop`'s changelog is lost: the net diff against `develop` is `7 0 CHANGELOG.md` — the six-line bullet plus one blank line, zero deletions.
- **`mempalace/cli.py` — auto-merged cleanly**, despite `develop` having changed it by +918/−9 since the branch point. I checked the result rather than trusting the clean merge: both new parsers register in the current structure, and driving `main()` with each subcommand reaches both handlers.
- `mempalace/exporter.py`, `mempalace/importer.py` and `tests/test_export_import_jsonl.py` replayed without conflict.

**The patch itself is line-identical across the replay.** For all four code/test files, the set of lines this PR adds is byte-identical before and after the rebase, and the only lines it removes are the same two in `exporter.py` (a module-docstring replacement that has always been part of the branch). All eight commits replayed in order, none dropped or squashed. That is patch preservation rather than a claim of semantic equivalence against 167 intervening commits — which is what the test run below is for.

No fixes were applied in this round; it is a rebase only, so there is no finding-to-commit map to give you this time.

On the suite, at the rebased head: **4870 passed, 12 failed, 53 skipped**, with coverage at 81.90% against the 80% gate. Six of the twelve are the same pre-existing failures I flagged last time — the five `test_init_filters_sys_path_from_leaked_pythonpath` parametrisations and `TestStaleLibraryGate::test_startup_baseline_survives_module_reload` — unchanged. The other six are all in `tests/test_palace_locks.py`, and they are not this branch: I reproduced the identical six, with identical signatures, on a clean `develop` checkout with this PR absent, and they pass in a correctly installed environment. They look like an artifact of my local checkout layout rather than anything in the repo. `ruff check` and `ruff format --check` are clean at 0.16.1, the version CI pins.

CI is awaiting maintainer approval to run.

One thing worth saying plainly, since it may be why this has been sitting. RFC 004 step 0 landed `logsync` — pull-based anti-entropy for the logstream between mutually-reachable peers — with memory read replicas on top in step 1. What this PR does is narrower and, I think, complementary rather than overlapping: an offline, portable snapshot of the drawers themselves — plain sorted files in a git-friendly tree, no peer reachability and no hub token required, usable as a backup or as a one-way migration onto a fresh machine. I do not want to duplicate work or cut across the direction you are taking with the replicated palace. If you would rather this were folded into that direction, scoped down, or dropped, I am happy to do any of those — just say which.

PR #2212, opened 2026-08-10
fix(backups): stop a socket in the palace from aborting repair (#2207)

### LINKED ISSUE #2207: mempalace repair crashes with shutil.Error when backing up a live palace directory (copytree can't copy mcp.sock)
## Summary
`mempalace repair` (default/legacy mode) backs up the **entire palace directory** with `shutil.copytree(palace_path, backup_path)` before rebuilding (`cli.py`, `cmd_repair`, around line 1294). If the palace directory contains the daemon's live Unix domain socket (`mcp.sock`), `copytree` raises `shutil.Error` because sockets aren't a file type it knows how to copy — and the whole `repair` command aborts with an unhandled traceback, **before doing any rebuild work**.

This is easy to hit in the daemon topology from #1270/#1888: `mcp.sock` lives inside `~/.mempalace/` right next to `chroma.sqlite3`, so any repair run while the socket file is still present (even with the daemon process itself already stopped — the socket file can outlive the process) will crash at the backup step.

## Environment
- MemPalace **3.6.0** (installed via pipx)
- chromadb **1.5.9**
- Python **3.14.5**
- macOS **26.6** (25G72), arm64 (Apple Silicon)
- Palace: ~29,332 drawers, `chroma.sqlite3` ~202 MB
- Daemon topology: LaunchAgent `com.mempalace.daemon` (long-lived, per #1270/#1888)

## Repro
1. Run the daemon topology (LaunchAgent creates `~/.mempalace/mcp.sock`).
2. Stop the daemon (`launchctl bootout ...`) but **don't** remove the leftover `mcp.sock` file (it isn't cleaned up on stop).
3. Run `mempalace repair --yes`.
4. Backup step crashes:

```
  Backing up to /Users/marcoperez/.mempalace.backup...
  File "/Users/marcoperez/.local/bin/mempalace", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File ".../mempalace/cli.py", line 2283, in main
    dispatch[args.command](args)
  File ".../mempalace/cli.py", line 1294, in cmd_repair
    shutil.copytree(palace_path, backup_path)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../shutil.py", line 654, in copytree
    return _copytree(entries=entries, src=src, dst=dst, ...)
  File ".../shutil.py", line 608, in _copytree
    raise Error(errors)
shutil.Error: [('/Users/marcoperez/.mempalace/mcp.sock', '/Users/marcoperez/.mempalace.backup/mcp.sock', "[Errno 102] Operation not supported on socket: '/Users/marcoperez/.mempalace/mcp.sock'")]
```

## Impact
- `copytree` collects errors and only raises at the end, so **every other file does get copied successfully** first (confirmed: resulting `.backup/chroma.sqlite3` was byte-identical in size to the source) — the backup is *functionally* complete, but the command still dies with an unhandled exception instead of reporting success.
- Worse: the crash happens **before** `_rebuild_collection_via_temp` runs, so a user re-running `repair` immediately hits `shutil.rmtree(backup_path)` cleanup logic on a *partial-looking* backup dir, then repeats the same crash — it doesn't self-heal.
- No rebuild happens at all until the operator notices the socket is the problem, deletes it manually, and retries.

## Root cause
`shutil.copytree` has no built-in handling for non-regular files (sockets, FIFOs); it always raises on them rather than skipping. `cmd_repair`'s backup step copies the whole live directory tree instead of just the data files that matter.

Note: `mempalace/repair.py`'s module docstring (the "rebuild" implementation used by `repair rebuild-index` / `--mode from-sqlite`) explicitly says *"The rebuild backs up ONLY chroma.sqlite3 (the source of truth), not the full palace directory — so it works even when link_lists.bin is bloated."* That's the right approach and doesn't have this bug. The default/legacy `cmd_repair` path in `cli.py` does a full-directory `copytree` instead, which is both slower (backs up the multi-hundred-MB HNSW segment files unnecessarily) and hits this socket bug. The two backup strategies have drifted apart.

## Suggested fixes (any one would do)
1. Pass an `ignore=` callback to `copytree` that skips non-regular files (sockets, FIFOs), e.g. via `stat.S_ISSOCK`/`S_ISFIFO` checks, or ignore known daemon artifacts by name (`mcp.sock`, `.daemon.lock`).
2. Simplest: have `cmd_repair`'s legacy backup path only copy `chroma.sqlite3` (+ maybe `knowledge_graph.sqlite3`), matching what `repair.py`'s rebuild already does — smaller, faster, and sidesteps the socket entirely.
3. At minimum, wrap the `copytree` call and surface a clear error message ("stop the daemon / remove mcp.sock before running repair") instead of a raw traceback.

## Workaround
Stop the daemon and `rm ~/.mempalace/mcp.sock` before running `mempalace repair`.

mvalentsev:
## What does this PR do?

Fixes #2207.

`repair` in its default (legacy) mode and `migrate` each copy the whole palace directory before overwriting it: the bare `shutil.copytree` calls at `cli.py:1782` and `migrate.py:319` on `develop` @ `8516db7f`. `copytree` cannot duplicate a Unix domain socket or a named pipe. It copies everything else, then raises `shutil.Error`, so the command died at the backup step before any rebuild ran.

Reproduced on `develop` @ `8516db7f` through the console script, `mempalace --palace ... repair --yes`, on a real 3-drawer palace, one entry planted per run:

| planted in the palace dir | exit | |
| --- | --- | --- |
| Unix socket (`mcp.sock`) | 1 | `shutil.Error [Errno 6] No such device or address` |
| named pipe, top level or one directory down | 1 | `` `...` is a named pipe `` |
| symlink to a socket | 1 | `[Errno 6]`, both entries listed |
| nothing (control) | 0 | rebuild completed |

It does not recover on its own: three consecutive runs on one palace all exited 1, and the third run's `<palace>.backup/chroma.sqlite3` had the same sha256 as the live file. The data copy finishes every time; only the command dies. `--mode from-sqlite` was never affected, since it copies `chroma.sqlite3` alone (`repair.py:1211`), and is unchanged here.

### What changed

`mempalace/backups.py` gains `copy_palace_dir()`, which runs the same `copytree` with an `ignore=` callback. The predicate is an allowlist, matching how the rest of the package treats directory entries (`miner.py:68`, `convo_miner.py:112`, `repair.py:71`): only regular files and directories are copied, and anything else is named and left out. Device nodes are skipped for the opposite reason to sockets, since the copy does not fail on them but dereferences them: a link to `/dev/zero` wrote 493 MB in 8 seconds before it was stopped.

A failed `stat` is never a reason to skip. It says the entry cannot be resolved right now, not that it holds no data, and no errno separates the two: a symlink into a volume that is not mounted fails exactly like one whose target was deleted, and on Windows an unmapped drive letter and an unreachable network share both arrive as `ENOENT` too. Such an entry is left for the copy to fail on, exactly as on `develop`, because this backup is the safety net for the rebuild that overwrites the palace next. Five tests pin that boundary, each asserting `shutil.Error` still comes out.

Skipped entries are named on stdout, printed from a `finally` so the list survives a copy that fails afterwards:

```
  Backing up to /tmp/pal.backup...
  Backup: skipped 1 entry that cannot be copied:
    mcp.sock (socket)
  Building temporary collection: mempalace_drawers__repair_tmp
```

Checked afterwards through MemPalace rather than pytest: for every trigger in the table, in both `repair` modes, the command exits 0, the palace still holds its 3 drawers, and `search` still returns the drawer text.

The bug-class sweep over `mempalace/` found exactly two production sites that copy a whole palace directory; both are in this PR.

## How to test

```bash
uv sync --extra dev
uv run pytest tests/test_backups.py tests/test_cli.py tests/test_migrate.py -v
```

Tests needing a socket, a FIFO, a device node or a symlink skip themselves where the platform or the user cannot create one, and the unreadable-directory tests also skip for root.

End to end against a throwaway copy of a real palace, on Linux or macOS:

```bash
cp -a ~/.mempalace/palace /tmp/pal          # never test this on the live one
mempalace --palace /tmp/pal repair --yes    # baseline: this succeeds

# cd first: an absolute socket path can exceed the sun_path limit
(cd /tmp/pal && python3 -c "
import socket
s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM); s.bind('mcp.sock'); s.close()
")

rm -rf /tmp/pal.backup
mempalace --palace /tmp/pal repair --yes
# on develop: Traceback ... shutil.Error, exit 1, no rebuild, and it repeats
# on this branch: '  Backup: skipped 1 entry that cannot be copied:' then the
#                 rebuild runs to 'Repair complete'
```

## Checklist
- [x] Tests pass (`uv run pytest tests/ -v --ignore=tests/benchmarks`)
- [x] No hardcoded paths
- [x] Linter passes (`uv run ruff check .`, and `ruff format --check .` on the `0.16.1` pin CI uses)


tests/test_backups.py
```diff
@@ -155,3 +239,621 @@ def flaky_remove(path):
...
+    _symlink_or_skip(palace / "loop_b", palace / "loop_a")
+
+    with pytest.raises(shutil.Error):
+        copy_palace_dir(str(palace), str(tmp_path / "palace.backup"))
+
+
+def test_copy_palace_dir_still_raises_on_a_symlink_to_its_own_parent(tmp_path):
+    """The directory-cycle form, which bloats the copy rather than failing fast."""
+    palace = _make_palace(tmp_path)
+    _symlink_or_skip(palace / "self", palace)
```
igorls, line 525: **[suggestion]** `test_copy_palace_dir_still_raises_on_a_symlink_to_its_own_parent` pins pre-existing `shutil.copytree(symlinks=False)` cycle behavior rather than the new skip allowlist. With follow-mode copy, a `self -> .` link causes deep recursive `copytree` calls; on common platforms this eventually fails with `OSError`/`ENAMETOOLONG` folded into `shutil.Error`, but if path length limits are loose relative to `sys.getrecursionlimit()` the same case surfaces as `RecursionError` (which is not a subclass of `shutil.Error`). The test also performs O(PATH_MAX) nested copies of the palace tree, so it is comparatively expensive for CI relative to the value it adds for #2207.

**Suggestion:** Drop the test, or replace it with a cheaper assertion that does not depend on PATH_MAX vs recursion depth (for example, only keep the fast mutual-file-symlink cycle test above it). If kept, assert `pytest.raises((shutil.Error, RecursionError))` and document why the heavy case is worth the runtime.

tests/test_backups.py
```diff
@@ -155,3 +239,621 @@ def flaky_remove(path):
...
+        locked.chmod(0o700)
+
+    assert any("aaa.fifo (named pipe)" in line for line in logs)
+    assert (dest / "chroma.sqlite3").is_file()
+
+
+@needs_fifo
+@needs_unprivileged_posix
+def test_copy_palace_dir_report_never_replaces_the_copys_own_failure(tmp_path):
+    """A failing ``log`` must not become the error the caller diagnoses.
```
igorls, line 582: **[nit]** Several new tests rely on `chmod(0o000)` directories to force a non-classifiable copy failure (`needs_unprivileged_posix`). That pattern is new in this suite and can be environment-sensitive (root is already skipped; some sandboxed or ACL-heavy runners still let the owner traverse a 000 directory, which would make `pytest.raises(shutil.Error)` fail for reasons unrelated to `copy_palace_dir`).

**Suggestion:** Prefer a deterministic failure inject (monkeypatch `shutil.copy2` / `copystat` to raise `OSError`) for the “still raises on ordinary copy failure” and “report must not replace copy failure” cases, and reserve real permission bits only if you specifically need to integration-test OS permission errors.

igorls (approved):
Wave 1 for 3.7.0: LGTM.

PR #2214, opened 2026-08-11
fix: stream JSONL transcript normalization

RoshanDev:
## Review status / scope reset

Rebased onto current `develop` (`d9f0590`) and reduced back to the normalization-only boundary on 2026-09-03. This two-commit diff is ready for human review: current-head CI and Codex review are clean. #2408 owns current Codex v2 parsing and remains tracked separately; it does not block review of this normalization-only change.

Earlier review rounds temporarily combined this PR with current Codex v2 parsing and migration work from #2408. That made #2214 duplicate incremental re-mine, hook-performance, and migration-transaction responsibilities. Those experimental commits were removed; the final review diff is two commits and does not modify hook, migration, or palace-wide version behavior.

## Summary

- Stream the JSONL formats recognized by the current `develop` parsers instead of reading and splitting the complete source first.
- Reuse the current no-symlink, regular-file-only, FIFO-safe, UTF-8/BOM-tolerant descriptor path for streaming and non-streaming reads.
- Keep every format behind the existing 500 MiB input cap. Streaming removes duplicate full-source/split-line allocations below that cap; it does not claim that materialized normalized output/chunking is safe above it.
- Preserve parser behavior and unknown-JSONL plain-text fallback by reusing the same line-parser implementations.
- Do not write a registry sentinel when normalization raises: a failed source remains eligible for a later retry instead of being silently marked mined.
- Add a repeatable 12 MiB `tracemalloc` regression so returning to `f.read()` plus split-line copies fails mechanically.

## Root cause

Both public normalization entry points called `_read_transcript_file()` before format detection. For a recognized JSONL this retained the entire source string, then allocated split-line/parser structures and the normalized transcript. A valid near-limit transcript could therefore create several simultaneous source-sized allocations in a long-lived process.

The scanner also treated a caught normalization exception as successfully empty and wrote a registry sentinel. An over-limit fallback, permission failure, or future format error could consequently become a permanent skip with no drawers.

## Dependency on current Codex v2 support

Current Codex `event_msg/item_completed` parsing remains owned by #2408 and is intentionally not duplicated here. #2214 streams whichever parser recognizes the source after the dependency lands. Until then, #2214 alone does not claim to normalize present-day v2 rollouts.

A local integration branch combining #2214 with #2408 was used only as read-only validation against the reporting file:

- Source: 442,841,936 bytes (about 422 MiB)
- Elapsed: 1.38 seconds
- Maximum RSS: 45,684 KiB
- Swap: 0
- Normalized output: 863,559 characters

No Hub, miner, Chroma collection, or palace was opened or written. This evidence validates the combined path, not the standalone PR head.

## Verification

- Normalize/conversation-miner focused group: 275 passed.
- Local full suite: 4,827 passed, 31 skipped, 106 deselected.
- The sole local failure, `test_sqlite_integrity_errors_reports_a_path_python_cannot_encode`, reproduces on pristine `develop` under CPython 3.12.11 and is unrelated to this diff.
- Repo-wide Ruff check: passed.
- Repo-wide Ruff format check: passed.
- `git diff --check`: passed.
- GitHub Actions: 9/9 passed, including Linux 3.9/3.11/3.13, macOS, Windows, Docker build/smoke/GPU, and lint.
- Codex current-head review for `d3e0681`: completed with no finding.

## Explicit remaining limit

Files over 500 MiB remain rejected. Supporting them requires incremental normalized-output emission and incremental chunking, not only line-by-line input parsing.

## Upstream coordination

- #2408 is monitored for maintainer comments, reviews, commits, or merge activity before any further history/status change.
- #2419 remains a separate Draft for Hub concurrency/crash-safety work.
- If #2408 lands first, this PR will be rebased and the combined v2 path revalidated before merging.

Related: #396, #468, #998, #2403, #2408, #2419

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -28,7 +28,7 @@ from .ids import (
     make_convo_sentinel_id,
     make_exchange_drawer_id,
 )
-from .normalize import MAX_STREAMING_JSONL_FILE_SIZE, normalize_conversations
+from .normalize import normalize_conversations
 from .entities import entities_metadata
 from .palace import (
     NORMALIZE_VERSION,
@@ -166,15 +166,9 @@ _LINE_GROUP_SIZE = 25  # lines per fallback group when no paragraph breaks
 _LINE_FALLBACK_MIN_NEWLINES = 20  # trigger line-group fallback above this newline count
 DRAWER_UPSERT_BATCH_SIZE = 1000
 MAX_FILE_SIZE = 500 * 1024 * 1024  # 500 MB — skip files larger than this.
-# Matches miner.py for formats that still require whole-file reads. Known
-# JSONL formats use a bounded streaming normalizer and can safely use its
-# larger limit without admitting multi-gigabyte JSON bundles or plain text.
-
-
-def _source_file_size_limit(filepath: Path) -> int:
-    if filepath.suffix.lower() == ".jsonl":
-        return MAX_STREAMING_JSONL_FILE_SIZE
-    return MAX_FILE_SIZE
+# Matches miner.py. Streaming recognized JSONL avoids retaining the complete
+# source string, but normalized conversation text is still materialized before
+# chunking, so every format stays behind the same conservative input cap.
 
 
 def _path_within_root(path: Path, root: Path) -> bool:
@@ -203,7 +197,7 @@ def _is_regular_source_file(filepath: Path, root: Path) -> bool:
                 raise
             fd = os.open(filepath, flags & ~getattr(os, "O_NONBLOCK", 0))
         st = os.fstat(fd)
-        return stat.S_ISREG(st.st_mode) and st.st_size <= _source_file_size_limit(filepath)
+        return stat.S_ISREG(st.st_mode) and st.st_size <= MAX_FILE_SIZE
     except OSError:
         return False
     finally:
@@ -567,7 +561,7 @@ def scan_convos(convo_dir: str, include_subagents: bool = False) -> list:
                         )
                         continue
                     file_size = file_stat.st_size
-                    file_size_limit = _source_file_size_limit(filepath)
+                    file_size_limit = MAX_FILE_SIZE
                     if file_size > file_size_limit:
                         print(
                             f"  SKIP: {filepath.name} ({file_size / (1024 * 1024):.1f} MB)"
@@ -974,9 +968,12 @@ def _normalize_convo_conversations(
     """
     try:
         conversations = [c for c in normalize_conversations(str(filepath)) if c]
-    except (OSError, ValueError):
-        if not dry_run:
-            _register_file(collection, source_file, wing, agent, extract_mode)
+    except (OSError, ValueError) as exc:
+        # A normalization failure is not an empty, successfully processed
+        # transcript. Registering a sentinel here makes the unchanged source
+        # look permanently mined, so future runs never retry after format,
+        # permission, or size-limit failures.
+        print(f"  SKIP: {filepath.name} (normalize error: {exc})", file=sys.stderr)
         return None
 
     total_len = sum(len(c.strip()) for c in conversations)
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -28,7 +28,7 @@ from .ids import (
     make_convo_sentinel_id,
     make_exchange_drawer_id,
 )
-from .normalize import normalize_conversations
+from .normalize import MAX_STREAMING_JSONL_FILE_SIZE, normalize_conversations
 from .entities import entities_metadata
 from .palace import (
     NORMALIZE_VERSION,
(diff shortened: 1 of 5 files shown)


RoshanDev:
Rebased onto current develop and restored conflict-free streaming normalization in 201ad04. Added current Codex item_completed streaming support plus a bounded-memory regression in 2fd0994; the wire-format behavior is aligned with #2408 and isolated for clean removal if #2408 lands first. Local focused suites pass (171 and 280 tests); full suite is 4832 passed, 31 skipped, 106 deselected with only the independently reproduced develop baseline repair assertion failing on CPython 3.12.11. Real read-only 442,841,936-byte Codex v2 normalization completed in 1.57s at 45,652 KiB max RSS with zero swap and no palace access. @codex review

RoshanDev:
Addressed the P1 oversized/unknown-JSONL sentinel finding in 425f947. Normalization exceptions are no longer treated as successfully empty transcripts: the miner emits an explicit normalize-error skip, writes no registry sentinel, and leaves the unchanged source eligible for a later retry. The regression reproduces the exact path with an unknown JSONL whose descriptor reports just over the 500 MiB in-memory cap, then verifies no collection upsert occurs. Focused regression passes; broader normalize/conversation-miner suite: 281 passed; full Ruff/format/diff checks pass. @codex review

RoshanDev:
Addressed both second-round P1 findings in 6efe48a. All source formats now remain behind the existing 500 MiB input cap: streaming removes the duplicate full-source/split-line allocations below that cap, but the PR no longer claims multi-gigabyte safety while normalized output and chunking are still materialized. Codex v2 blocks now use strip() only as an emptiness predicate and append the original text unchanged, preserving leading/trailing whitespace. Regressions pin over-cap rejection, low-text-density bounded peak memory, and verbatim v2 whitespace. Related suite: 281 passed; Ruff/format/diff checks pass. The real 442,841,936-byte v2 file still normalizes in 1.38s at 45,684 KiB max RSS with zero swap. @codex review

RoshanDev:
Addressed the third-round P1 rebuild-gate finding in edfca40. NORMALIZE_VERSION is now 3, with the version history documenting that current Codex item_completed extraction replaces raw-JSON drawers. The schema-gate regression now constructs an explicit normalize_version=2 source and verifies it cannot short-circuit; the existing stale-drawer end-to-end test confirms old drawers are replaced and newly filed metadata carries the current version. Targeted version tests: 2 passed; broader normalize/miner suite: 409 passed; Ruff/format/diff checks pass. The service and automatic hooks remain disabled, so this version bump cannot trigger an unattended local rebuild. @codex review

RoshanDev:
Addressed the fourth-round P1 global-rebuild finding in c0469fe. The palace-wide NORMALIZE_VERSION remains 2. A bounded first-1000-record/cli-version probe identifies only current Codex v2 sources; bulk prefetch applies a per-source required version map, the lock-held file_already_mined recheck receives the same requirement, and only those Codex drawers/sentinels are stamped version 3. Ordinary project files, plain transcripts, and legacy Codex v1 remain at version 2. Regressions cover scoped detection, bulk and lock-held stale decisions, and an end-to-end v2 mine whose real metadata is version 3. Broader normalize/miner/non-regular-file suite: 454 passed; Ruff/format/diff checks pass. @codex review

RoshanDev:
Addressed both fifth-round findings in 6e213e7. Codex parsing tracks whether any v2 conversation item was accepted and calls _messages_to_transcript(..., spellcheck=False) for that mixed/current session, so optional spell correction cannot rewrite verbatim v2 user text. Registry replacement now uses a crash-safe pending marker: upsert pending=true, delete only prior same-source/mode IDs, then update the sentinel to pending=false as the commit point. file_already_mined, mined-mtime prefetch, and content-hash prefetch all ignore pending markers, so every crash/failure window remains retryable. Regressions cover a mutating spellchecker, pending-marker invisibility across all three paths, and end-to-end replacement of a short stale raw v2 drawer with one finalized sentinel. Broader suite: 456 passed; Ruff/format/diff checks pass. @codex review

RoshanDev:
Addressed both sixth-round findings by resetting the PR to its normalization-only boundary. The v2 parser, per-source migration, registry transaction, global/per-source version gates, and pre-skip format probes have all been removed from the review diff; those responsibilities remain with #2408 and its incremental re-mine foundation. Current head d3e0681 contains only streaming input on the parsers already present in develop, the shared 500 MiB cap, normalization-failure retry semantics, and a bounded-memory regression. Focused suite: 275 passed; Ruff/format/diff checks pass. The 422 MiB v2 field measurement is now explicitly documented as combined #2214 + #2408 integration evidence, not standalone-head behavior. @codex review

RoshanDev:
Final current-head snapshot for d3e0681: scope reset to 2 commits / 4 files (+289/-31); local focused suite 275 passed; local full suite 4827 passed, 31 skipped, 106 deselected with only the independently reproduced CPython 3.12 develop-baseline repair assertion failing; GitHub Actions 9/9 passed including Windows; Codex review completed with no finding. Keeping this PR Draft while #2408 ownership/landing order remains unresolved; the PR body distinguishes standalone behavior from combined #2214 + #2408 v2 field evidence.

Issue #2183, opened 2026-08-07
mine --mode convos: file-level already-filed tracking can silently skip individual exchanges from an already-mined transcript


fenrith:
# `mine --mode convos` / hook transcript ingest silently drops individual exchanges from an already-"filed" file

**Version:** mempalace 3.5.0 (pip)
**Platform:** Windows 11, Claude Code harness

## Summary

A specific tool-call exchange within a Claude Code session transcript (`.jsonl`)
was never ingested into the palace, even though:

- The Stop hook fired correctly and mined the same transcript file successfully
  multiple times over the course of the session (confirmed via `hook_state/hook.log`
  — `TRIGGERING SAVE at exchange 133`, `Drawers filed: 523`, etc.)
- Two subsequent manual `mempalace mine <dir> --mode convos --wing sessions` runs
  against the same transcript directory both reported `Files processed: 0`,
  `Files skipped (already filed): N` — i.e. the miner considers the file fully
  ingested and does not attempt to re-derive missing chunks.
- Targeted searches for content unique to the missing exchange all returned zero
  genuine matches, ruling out a search-ranking/embedding issue:
  - Author's distinctive first name (`Matshela`)
  - The paper's DOI (`10.2139/ssrn.5794522` / `5794522`)
  - A multi-word phrase from the paper title (`Cliff Intensity Index`,
    `Institutional Frictions Sociotechnical Inertia`)
- Neighboring exchanges in the *same transcript file*, both before and after the
  missing one chronologically, were successfully ingested and are findable via
  search (confirmed via `mempalace_reconnect` + `mempalace_search`).

So this isn't "the miner hasn't run yet" or "search can't find it" — it's a
genuine gap in what got chunked/embedded from one specific exchange, and the
file-level "already filed" tracking means re-running `mine` against the same
directory can never retry it.

## Reproduction context

The missed exchange was a `curl` tool call + response (a CrossRef API JSON
payload) followed by my own summary/citation text, inside a session that made
many other similar `curl`-based DOI lookups in immediate succession — several of
which *were* captured correctly. I don't have insight into what's structurally
different about the dropped one (response size? a `Bash` tool-result formatting
quirk? chunk-boundary placed mid-exchange?) since I haven't looked at the
extraction code — flagging behaviorally rather than with a root cause.

## Workaround used

Manually filed a backfill drawer via `mempalace_add_drawer` (MCP tool) with the
missing content — confirmed searchable afterward (scoped to
[...]

PR #2223, opened 2026-08-11
fix(ingest): never block on a non-regular file (#2221)

### LINKED ISSUE #2221: mempalace mine hangs forever on a named pipe in the mined directory (sweep, init, compress and split too)
**What happened?**

`mempalace mine <dir>` never returns when the directory holds a named pipe (FIFO) whose filename carries a readable extension, `notes.md` for example. No error and no exit: the process sits in the kernel until it is killed. On a terminal it prints the banner and the files it managed before the pipe, then stops dead mid-list; with stdout redirected it produces nothing at all, because the buffer is never flushed. `mine --mode convos`, `sweep`, `init`, `compress` and `split` wedge the same way on a pipe named after a file each of them reads.

Reproduced on `develop` @ `9a3afc9d` (3.7.0). Every run below was capped with `timeout 180`, and `exit=124` is the timeout killing it:

| command | the pipe | result |
|---|---|---|
| `mine CORPUS` | `notes.md` | printed `Files: 2`, filed `real.md`, then **hung** |
| `mine CORPUS` | `mempalace.yaml` | **hung** before the banner |
| `mine --mode convos CONVOS` | `piped.jsonl` | **hung** |
| `sweep CONVOS` | `piped.jsonl` | **hung** |
| `init --yes --no-llm DIR` | `notes.md` | **hung** |
| `init --yes --no-llm DIR` | `.gitignore` in a git repo | **hung** |
| `init --yes --no-llm DIR` | `settings.gradle` beside a real `build.gradle` | **hung** |
| `init --yes --no-llm DIR` | `mempalace.yaml`, the file `init` writes | **hung** |
| `compress` | `entities.json` in the working directory | **hung** |
| `split DIR` | `piped.txt` | **hung** |

I am not claiming this is common: a sweep of my own home and project trees found no FIFO at all outside `/run`. What makes it worth fixing is the shape rather than the frequency. `mkfifo` needs no privileges, the extension allowlist decides purely by name (54 suffixes, `.md` and `.jsonl` among them), and the same class already reached a user in #2207, where a daemon's leftover `mcp.sock` in a walked directory broke `repair` on a real machine. A socket makes the command crash; a pipe makes it hang with no way to tell what happened.

**What did you expect?**

The pipe holds no palace data, so it should be skipped and named, the way an oversized file or a symlink already is, and the rest of the corpus mined.

**Root cause**

Two shapes, both of which decide "can I read this?" without ever asking what the file *is*.

*Open first, check the type second.* `scan_project` accepts `notes.md` on its suffix (`miner.py:1683`) and clears the size gate at `miner.py:1711`, because `stat` never blocks on a pipe and reports `st_size == 0`. `process_file` then reaches `miner.py:66`:

```python
flags = os.O_RDONLY | getattr(os, "O_NOFOLLOW", 0)
fd = os.open(filepath, flags)
st = os.fstat(fd)
if not stat.S_ISREG(st.st_mode) or st.st_size > MAX_FILE_SIZE:
    return None
```

The refusal on the last two lines is written, and it is correct. Control never reaches it — opening a FIFO for reading blocks in the kernel until a writer appears, so `os.fstat` never runs. `convo_miner.py:191`, `normalize.py:128` and `repair.py:68` carry the identical dead guard. (`repair`'s is not reachable from the `repair` command today: `cli.py:1866` rejects the palace earlier via `contains_palace_database`, i.e. `os.path.isfile`, which is false for a pipe.)

*`exists()` is not a type check.* `Path.exists()` and `os.path.exists()` are true for a FIFO, and several fixed-name reads use one as their only guard before opening.

Every function below was probed on `9a3afc9d` with a real `mkfifo` under a 12 s cap, and every one was killed by the timeout:

| function | how the pipe gets there | pipe named |
|---|---|---|
| `miner.load_config` (`miner.py:489,521`) | `exists()` on a fixed name | `mempalace.yaml` |
| `sweeper.parse_claude_jsonl` (`sweeper.py:105`) | `sweep_directory`'s `rglob("*.jsonl")` | `session.jsonl` |
| `entity_detector.detect_entities` (`entity_detector.py:631`) | `scan_for_detection` | `notes.md` |
| `cli._gather_origin_samples` (`cli.py:134`) | `scan_for_detection` | `notes.md` |
| `cli._ensure_mempalace_files_gitignored` (`cli.py:268`) | `exists()` on a fixed name | `.gitignore` |
| `cli.cmd_compress` (`cli.py:2223,2227`) | `os.path.exists` on a fixed name | `entities.json` |
| `cli.cmd_init` (`cli.py:423`) | a write with no guard at all | `entities.json` |
| `project_scanner._collect_manifest_names` (`project_scanner.py:424`) | its own `os.walk` | `package.json` |
| `project_scanner._parse_gradle_root_project_name` (`project_scanner.py:192`) | a **sibling** path the parser builds at `:206` | `settings.gradle` |
| `room_detector_local.detect_rooms_local` (`room_detector_local.py:295`) | a write with no guard at all | `mempalace.yaml` |
| `llm_refine.collect_corpus_text` (`llm_refine.py:482`) | `_walk_prose`; the stat above it reads only `st_mtime` | `notes.md` |
| `split_mega_files.main` (`split_mega_files.py:278`) | `glob("*.txt")` | `x.txt` |
| `hook_shell.count_human_messages` (`hook_shell.py:77`) | no guard; `mempal_save_hook.sh` screens with `[ -f ]` | `transcript.jsonl` |

`mine --mode extract` is not affected — `format_miner.py` stats before it opens, and a pipe reports zero bytes, so it leaves at the existing empty-file gate.

**How to reproduce:**

```bash
mkdir -p /tmp/fifo-repro/corpus
cat > /tmp/fifo-repro/corpus/mempalace.yaml <<'YAML'
wing: fifo-repro
rooms:
  - name: general
    description: All project files
YAML
python3 -c "open('/tmp/fifo-repro/corpus/real.md','w').write('# Real note\n\n' + 'The quick brown fox jumps over the lazy dog. ' * 40)"
mkfifo /tmp/fifo-repro/corpus/notes.md

timeout 180 mempalace --palace /tmp/fifo-repro/palace mine /tmp/fifo-repro/corpus
echo "exit=$?"   # 124, killed by the timeout
```

**Environment:**
- OS: Linux 6.18 (WSL2). Measured on Linux only, but the blocking open is specified rather than Linux-specific — `fifo(7)`: "opening the FIFO blocks until the other end is opened also", and with `O_NONBLOCK` "opening for read-only succeeds even if no one has opened on the write side yet". Windows has no FIFO in the filesystem namespace, so no directory walk can reach one there.
- Python version: 3.12.3
- MemPalace version: 3.7.0 (`develop` @ `9a3afc9d`)

mvalentsev:
## What does this PR do?

Fixes #2221. A named pipe in a mined directory wedged `mempalace mine` forever; `mine --mode convos`, `sweep`, `init`, `compress` and `split` blocked the same way. All ten scenarios in the issue hang on `develop` under a 180 s cap and complete here.

Two shapes, neither of which asks what the file *is*.

**Open first, check the type second.** `scan_project` accepts `notes.md` on its suffix (`miner.py:1683`) and `stat` never blocks on a pipe, so `process_file` reaches `miner.py:66`:

```python
fd = os.open(filepath, flags)          # blocks forever on a FIFO
st = os.fstat(fd)
if not stat.S_ISREG(st.st_mode) or st.st_size > MAX_FILE_SIZE:
    return None                        # correct, and unreachable
```

`convo_miner.py:191`, `normalize.py:128` and `repair.py:68` carry the identical dead guard.

**`exists()` is not a type check.** It is true for a pipe, and `mempalace.yaml`, `.gitignore`, `entities.json` and a gradle `settings.gradle` all reach an `open` behind one.

## The change

`O_NONBLOCK` on those four opens, so the `S_ISREG` refusal already written there can run — the verdict is the file mode, not an errno, so a pipe with a live writer is refused too. Everywhere else a type gate: the walks drop non-regular entries with a `SKIP: <name> (not a regular file)` line beside the existing symlink and size lines, and fixed-name reads use `is_file()`. `scan_project` and `scan_convos` already stat for the size limit, so one `stat()` now answers both questions.

Two traps:

- **The flag is not inert under a write lease** — it fails `EAGAIN` where a blocking open waits out `lease-break-time` and succeeds, silently dropping a file the old code read. The retry re-checks the type first, and the kernel grants leases on regular files only (`F_SETLEASE` on a pipe fails `EINVAL`), so it is authorised by the type, not the errno.
- **A gate goes where the read was, and must not eat a real error.** `Path.is_file()` raises `PermissionError` on a directory without `x` (through 3.13), so it belongs inside whatever `try` the `open` sat in — including each `project_scanner` parser, which swallows its own `OSError`. And `sweep_directory` skips a non-regular *type* as a benign `SKIP` but still books a failed `stat`: a dangling symlink is a real error, and `sweep` still exits 2 on it.

## Audit

Every `os.open`, `open`, `read_text`, `read_bytes` and parser entry in `mempalace/` was traced to where its path comes from and probed with a real `mkfifo` on both commits. **Nineteen sites block on `develop`; none block here.** Two are hardening rather than live hangs, and I would rather say so than overclaim: `repair`'s helper sits behind `contains_palace_database` (`cli.py:1866`) and `repair.py:1692`, `hook_shell` behind `[ -f ]` in `mempal_save_hook.sh:228`. `detect_entities` is the chokepoint for all four callers of `scan_for_detection`. The one site no discovery gate could reach is a write: `split_file` synthesises its own output names, so a pipe at one of them blocked the write with the new `SKIP:` line for that pipe already on screen. `format_miner.extract_text` is left alone — immune by ordering rather than by type, which deserves its own change; the ordering is pinned by a test here. **Known limit:** the stat-then-open gates stay TOCTOU-racy, unlike `_read_text_no_follow`, which re-checks the type on the descriptor.

## How to test

```bash
mkdir -p /tmp/fifo-repro/corpus
printf 'wing: fifo-repro\nrooms:\n  - name: general\n    description: All project files\n' \
  > /tmp/fifo-repro/corpus/mempalace.yaml
python3 -c "open('/tmp/fifo-repro/corpus/real.md','w').write('# Real\n\n' + 'the quick brown fox. ' * 60)"
mkfifo /tmp/fifo-repro/corpus/notes.md

timeout 180 mempalace --palace /tmp/fifo-repro/palace mine /tmp/fifo-repro/corpus; echo "exit=$?"
```

`develop`: prints `Files: 2`, files `real.md`, then nothing — `exit=124`. Here: `SKIP: notes.md (not a regular file)`, `Files: 1`, `Drawers filed: 2`, `exit=0`, and `search` afterwards returns `real.md`.

The exit statuses are the easiest thing here to get wrong:

```bash
mkdir -p /tmp/sweep-repro && cd /tmp/sweep-repro
printf '{"type":"user","sessionId":"s","uuid":"u","timestamp":"2026-01-01T00:00:00Z","message":{"role":"user","content":"hi"}}\n' > real.jsonl
mkfifo piped.jsonl                                    # nothing to sweep
mempalace --palace ./palace sweep .; echo "exit=$?"   # 0, with a SKIP line
rm piped.jsonl; ln -s /tmp/sweep-repro/gone.jsonl dangling.jsonl   # a real error
mempalace --palace ./palace sweep .; echo "exit=$?"   # 2, as on develop
```

## Tests

`tests/test_non_regular_file_guards.py`, 42 tests, each under a `SIGALRM` deadline so a regression turns the file red instead of hanging pytest — the deadline raises a `BaseException` subclass, because `except OSError` guards every site here and `TimeoutError` *is* an `OSError`. Every production hunk was reverted on its own in an isolated copy; two are not covered by a unit test and I would rather name them than round the count up — the `except OSError` around `detect_rooms_local` in `cmd_init`, and the legacy `mempal.yaml` name in `load_config`. 36 of the 42 are POSIX-only (Windows has no FIFO in the filesystem namespace), and three more skip as root, matching `tests/test_backups.py`.

The suite's only failures here are the palace-lock family under full-suite load, and they are load artefacts rather than anything this branch does: the same commit produced one failure on a quiet machine and twelve on a busy one, and every one of them passes when the file is run on its own. CI is the arbiter. One unrelated fix rides along on a line this PR had to touch: `normalize._read_transcript_file` was wrapping its two refusals a second time, printing the path twice.

## Checklist
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`, and `ruff format --check .`, at the ruff 0.16.1 pinned in CI)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- **Ingest commands no longer hang on a named pipe.** A FIFO carrying a mined suffix (`notes.md`, `session.jsonl`) parked `mine`, `mine --mode convos`, `sweep`, `init`, `compress` and `split` in a blocking `open()` forever, because every reader decided by name and checked the file type only after opening. The four affected opens now pass `O_NONBLOCK` so the `S_ISREG` refusal already written there is reachable, and the discovery walks drop non-regular entries with a `SKIP: <name> (not a regular file)` line. (#2221)
+
 ---
 
 ## [3.7.0] — 2026-08-11
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -131,6 +131,13 @@ def _gather_origin_samples(project_dir) -> list:
         if total_chars >= _PASS_ZERO_TOTAL_CAP:
             break
         try:
+            # ``scan_for_detection`` picks candidates by extension, so a FIFO
+            # named ``notes.md`` reaches this loop; opening one for reading
+            # blocks until a writer appears. ``is_file()`` stats instead.
+            # It belongs inside the try: it raises PermissionError on an
+            # unreadable directory, which the open below used to absorb.
+            if not filepath.is_file():
+                continue
             with open(filepath, encoding="utf-8", errors="replace") as f:
                 content = f.read(_PASS_ZERO_PER_FILE_CAP)
         except OSError:
@@ -263,9 +270,17 @@ def _ensure_mempalace_files_gitignored(project_dir) -> bool:
     if not (project_path / ".git").exists():
         return False
     gitignore = project_path / ".gitignore"
+    # ``exists()`` is true for a FIFO, and both the read below and the append
+    # at the end of this function would block in the kernel on one. Decide by
+    # type instead: an absent file still yields "" as before, a regular one
+    # is read, and anything else is left untouched.
+    if gitignore.exists() and not gitignore.is_file():
+        return False
     # Force UTF-8: Windows defaults to GBK and chokes on non-ASCII .gitignore
     # comments, killing auto-init even though the file is valid UTF-8.
-    existing = gitignore.read_text(encoding="utf-8", errors="replace") if gitignore.exists() else ""
+    existing = (
+        gitignore.read_text(encoding="utf-8", errors="replace") if gitignore.is_file() else ""
+    )
     existing_lines = {line.strip() for line in existing.splitlines()}
     missing = [p for p in _MEMPALACE_PROJECT_FILES if p not in existing_lines]
     if not missing:
@@ -420,9 +435,19 @@ def cmd_init(args):
         if confirmed["people"] or confirmed["projects"] or confirmed.get("topics"):
             project_path = Path(args.dir).expanduser().resolve()
             entities_path = project_path / "entities.json"
-            with open(entities_path, "w", encoding="utf-8") as f:
-                json.dump(confirmed, f, indent=2, ensure_ascii=False)
-            print(f"  Entities saved: {entities_path}")
+            # Opening a pre-existing FIFO for writing blocks in the kernel
+            # until a reader appears. Only a regular file is a valid target
+            # for the per-project audit trail; the global registry merge
+            # below is unaffected either way.
+            if entities_path.exists() and not entities_path.is_file():
+                print(
+                    f"  ! Not writing entities: {entities_path} is not a regular file",
+                    file=sys.stderr,
+                )
+            else:
+                with open(entities_path, "w", encoding="utf-8") as f:
+                    json.dump(confirmed, f, indent=2, ensure_ascii=False)
+                print(f"  Entities saved: {entities_path}")
 
             from .config import normalize_wing_name
             from .miner import add_to_known_entities
@@ -438,7 +463,14 @@ def cmd_init(args):
         print("  No entities detected -- proceeding with directory-based rooms.")
 
     # Pass 2: detect rooms from folder structure
-    detect_rooms_local(project_dir=args.dir, yes=getattr(args, "yes", False))
+    try:
+        detect_rooms_local(project_dir=args.dir, yes=getattr(args, "yes", False))
+    except OSError as exc:
+        # Writing mempalace.yaml is the point of init; a target it cannot
+        # write (a pre-existing pipe, a full disk) is a hard failure, and a
+        # message beats the traceback this used to produce.
+        print(f"\n  ERROR: {exc}", file=sys.stderr)
+        sys.exit(1)
     cfg.init()
     backend = _backend_arg(args)
     if backend:
@@ -2219,12 +2251,15 @@ def cmd_compress(args):
     # Load dialect (with optional entity config)
     config_path = args.config
     if not config_path:
+        # ``isfile`` rather than ``exists``: the latter is true for a FIFO,
+        # and ``Dialect.from_config`` opens whatever it is handed, which
+        # blocks in the kernel on a pipe named entities.json in the cwd.
         for candidate in ["entities.json", os.path.join(palace_path, "entities.json")]:
-            if os.path.exists(candidate):
+            if os.path.isfile(candidate):
                 config_path = candidate
                 break
 
-    if config_path and os.path.exists(config_path):
+    if config_path and os.path.isfile(config_path):
         dialect = Dialect.from_config(config_path)
         print(f"  Loaded entity config: {config_path}")
     else:
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -8,6 +8,7 @@ Normalizes format, chunks by exchange pair (Q+A = one unit), files to palace.
 Same palace as project mining. Different ingest strategy.
 """
 
+import errno
 import os
 import sys
 import json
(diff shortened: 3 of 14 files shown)


mvalentsev:
Closing this: the #2221 work reached `develop` through #2228, which carried it as `db29959` and `0f3f0c6`.

The revision that landed predates this branch's last pass, so three guards and their tests did not come along with it:

- **`sweep_directory`** — the gate probes the file type with `f.stat()` inside a `try`, and its `except OSError` prints `SKIP` and continues. A dangling symlink named `*.jsonl` therefore leaves `sweep` reporting success on a transcript it could not read. On `9a3afc9d`, the merge base of this branch, the same input was booked in `failures` and exited 2; on `906b918a` it prints `SKIP: broken.jsonl (stat error: No such file or directory)` and exits 0.
- **`_parse_gradle_root_project_name`** — the `is_file()` gate needs to sit inside the `try` whose `except OSError` the parser already provides, and `_collect_manifest_names` needs to stat through `os.path.isfile`. Without that, a manifest under a directory with `r` but no `x` raises `PermissionError` out of a call that used to answer "no manifest name", so `init` tracebacks. (`Path.is_file()` propagates `EACCES` on 3.9 through 3.13, all checked.)
- **`split_file`** — it builds its output names itself, past the type gate in `main()`, so a pre-existing FIFO at one of those names still wedges `write_text`.

I opened #2244 with exactly that remainder — 6 files, +146/−10 on top of `906b918a`, CI green — rather than rebasing this branch. On the reason, correcting myself now that I have actually run the rebase instead of reasoning about it: it conflicts in four files (`miner.py`, `project_scanner.py`, `sweeper.py`, the test file), and every one of those conflicts is this branch's own earlier revision against its later one — `miner.py`'s single hunk is the `ENXIO`/`EINVAL` comment. None is a collision with someone else's work. Resolved hunk by hunk the rebase lands +144/−12 against `develop`, near-identical to #2244's +146/−10, with #2088 intact; #2183 and #2032 are untouched either way. Resolving file-wise is the trap — taking `miner.py` wholesale from this branch drops #2088's work, `chunk_total` going from 8 occurrences to 0. So a fresh branch avoided a footgun; it was not the only workable route.

#2221 is still open; #2244 closes it.

PR #2232, opened 2026-08-12
fix(mcp): answer every JSON-RPC request that carries an id (#2231)

### LINKED ISSUE #2231: MCP server never answers a malformed-envelope request, so the client hangs on its id
**What happened?**

The MCP server drops some requests without answering them at all, so a client that sent an `id` waits forever. The server stays alive and serves the next request, so it looks like one permanently pending call rather than a crash.

Measured on current `develop` against a real `mempalace-mcp`, every request carrying a valid `id`:

| request | stdio | HTTP |
|---|---|---|
| `"method": 123` | no response | socket closed, no reply |
| `"method": true` | no response | socket closed, no reply |
| `"params": "oops"` on `initialize` | no response | socket closed, no reply |
| `tools/call` with `"arguments": 5` | no response | socket closed, no reply |
| `tools/call` with `"name": {}` | no response | socket closed, no reply |
| malformed JSON line | no response | `-32700` (correct) |
| `"method": null` or `[]` | `-32601` | `-32601` |

The way in is `or` used as a type guard. In `handle_request`:

```python
method = request.get("method") or ""
params = request.get("params") or {}
```

`or` rescues only *falsy* values, so a truthy non-string reaches `.startswith()` and a truthy non-object reaches `.get()`, raising `AttributeError`. `method: null` and `[]` survive purely because they are falsy.

The same pattern repeats twice more, which is what the last two rows hit: `params.get("arguments") or {}` lets a truthy non-mapping through to `**` unpacking, and `tool_name` goes to `in TOOLS` with no type check, so an unhashable name raises `TypeError`. A merely *wrong* tool argument is fine and still returns `-32000` through `_internal_tool_error`.

Each transport then loses the exception differently. stdio catches everything and continues without writing a response:

```python
except Exception as e:
    logger.error(f"Server error: {e}")
    continue
```

HTTP has no `try` around `_http_dispatch`, so it escapes into `BaseHTTPRequestHandler` and the connection closes with no JSON-RPC body.

**What did you expect?**

Every request carrying an `id` gets a response. The pieces are already in the file: `_json_rpc_parse_error()` returns `-32700` but is wired only into HTTP, and `tools/call` guards its params with an explicit `isinstance` check while `initialize` does not.

**How to reproduce:**

1. Write three requests to `rpc.jsonl`:
   ```
   {"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"r","version":"1"}}}
   {"jsonrpc":"2.0","id":2,"method":123}
   {"jsonrpc":"2.0","id":3,"method":"ping"}
   ```
2. Run `python -m mempalace.mcp_server < rpc.jsonl`
3. Responses come back for ids 1 and 3 only. stderr shows `Server error: 'int' object has no attribute 'startswith'`, and nothing is ever written for id 2.

**Environment:**
- OS: Ubuntu 24.04.4 LTS (WSL2)
- Python version: 3.12.3
- MemPal version: 3.7.0 (current `develop`)

---

Thanks to @tvlgiao, who flagged the silent catch-all as a side note in #2229. That report's main symptom is a different path that was already fixed, so this is filed separately with its own repro.

mvalentsev:
## What does this PR do?

Fixes #2231: every JSON-RPC request carrying an `id` now gets a response.

`or` was being used as a type guard, and it rescues only *falsy* values. A truthy non-string `method` reached `.startswith()` and a truthy non-object `params` reached `.get()`, raising `AttributeError`. The stdio loop then caught everything and continued without writing a reply, so the client waited on that `id` forever; HTTP had no `try` around `_http_dispatch`, so the socket simply closed.

The pattern sits at four depths, so the fix follows it down:

* `_normalize_envelope()` gives `method` / `params` the same defaults their falsy counterparts always got.
* `_http_dispatch` reads `method` through it too. That dispatcher landed after this branch opened and repeats the same `or ""`, so a non-string method raised inside the handler thread before `handle_request` ever saw it.
* `tools/call` validates `name` and `arguments` (`arguments: 5` reached `**` unpacking; an unhashable `name` broke the `in TOOLS` test) and answers `-32602`.
* `_request_is_mutating` returns a verdict instead of raising. It decides whether a mid-flight call may be replayed locally, so an exception there defeated the guarantee it exists for.

Then neither transport loses what still escapes: stdio answers `-32700` on a parse failure (via `_json_rpc_parse_error`, until now wired only into HTTP) and `-32603` on a dispatch failure; HTTP answers `500` with `-32603` instead of dropping the connection. Neither new error path answers a notification: no `id`, no error response.

Nothing that used to be answered stops being answered: `params: []`, `0`, `""` and `false` were normalised to `{}` before this guard existed and still are.

**On the codes:** a non-string `method` returns `-32601`, not `-32600`, to match the `method: null` behaviour pinned by `test_malformed_method_none`. `-32603` covers dispatch-level failures while `_internal_tool_error` keeps `-32000` for tool-level ones. On a parse failure the id is unknowable, so the reply carries `id: null` per JSON-RPC 2.0 section 5.

Out of scope and unchanged: batch arrays get one `-32600` rather than one response per element, and a `notifications/*` method sent *with* an `id` still gets no reply.

## How to test

```bash
printf '%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"r","version":"1"}}}' \
  '{"jsonrpc":"2.0","id":2,"method":123}' \
  '{"jsonrpc":"2.0","id":3,"method":"ping"}' > rpc.jsonl
python -m mempalace.mcp_server < rpc.jsonl
```

On `develop` only ids 1 and 3 come back; with this branch all three do.

`python -m pytest tests/test_mcp_jsonrpc_envelope.py tests/test_mcp_http_transport.py -q` covers it: 41 cases in the new file and 3 added to the HTTP suite. Reverting each of the nine guards in turn reddens only its own tests (10 / 8 / 4 / 5 / 4 / 1 / 1 / 1 / 1), and both files are green again once it is restored.

## Checklist
- [x] No hardcoded paths

Full suite: 4948 passed, 53 skipped, 1 failed. The failure is `test_cli_import_under_startup_budget`, a startup-time budget that fails 3 runs out of 3 on unmodified `develop` too.

### DIFF
--- a/mempalace/mcp_server/http.py
+++ b/mempalace/mcp_server/http.py
@@ -365,7 +365,10 @@ def _http_dispatch(request):
     the lock; palace writes take it exclusively. Unclassified tools fail
     closed onto the exclusive side.
     """
-    method = request.get("method") or "" if isinstance(request, dict) else ""
+    # Same envelope trap as handle_request, on a dispatcher added after it: the
+    # `or ""` fallback only rescues falsy values, so a truthy non-string method
+    # reached .startswith() below and raised out of the handler thread.
+    method, _ = _normalize_envelope(request) if isinstance(request, dict) else ("", {})
     if method in _HTTP_PROTOCOL_METHODS or method.startswith("notifications/"):
         return handle_request(request)
     tool_name = None
@@ -1012,7 +1015,29 @@ def _build_http_server(host: str, port: int):
 
             # Locking policy lives in _http_dispatch: global lock for
             # Chroma-touching tools, lock-free for logstream tools.
-            response = _http_dispatch(request)
+            try:
+                response = _http_dispatch(request)
+            except Exception:
+                # Without this the exception escaped into BaseHTTPRequestHandler,
+                # which closes the connection with no reply at all -- the client
+                # sees a dropped socket instead of a JSON-RPC error. Log with the
+                # traceback: -32603 tells the client nothing diagnostic, so the
+                # stack is the only record of what actually failed.
+                logger.exception("HTTP JSON-RPC dispatch error")
+                req_id = request.get("id") if isinstance(request, dict) else None
+                if req_id is None:
+                    # A notification is owed no response body, failure included,
+                    # matching the 202 branch below and the stdio loop. The
+                    # status still reports the failure at the transport level.
+                    self._record_request(500)
+                    self.send_response(500)
+                    self.send_header("Content-Length", "0")
+                    self.send_header("Connection", "close")
+                    self.end_headers()
+                    self.close_connection = True
+                    return
+                self._send_json(500, _json_rpc_internal_error(req_id))
+                return
 
             if response is None:
                 # JSON-RPC notifications intentionally have no response body.
--- a/mempalace/mcp_server/protocol.py
+++ b/mempalace/mcp_server/protocol.py
@@ -663,6 +663,30 @@ def _decorate_mcp_tool_result(tool_name: str, result):
     return result
 
 
+def _normalize_envelope(request: dict) -> "tuple[str, dict]":
+    """Read `method` and `params` off a request without raising.
+
+    `or ""` / `or {}` only rescue falsy values, so a truthy non-string method
+    reached `.startswith()` and truthy non-object params reached `.get()`,
+    raising AttributeError out of `handle_request`. Over stdio that surfaced
+    as no response at all, leaving the client waiting on an id forever.
+
+    A malformed value falls back to the same default its falsy counterpart has
+    always been given, so nothing that used to be answered stops being
+    answered: `params: []` is legal by-position params this server does not
+    support, and it kept working as an empty mapping before this guard existed.
+    """
+    method = request.get("method")
+    if not isinstance(method, str):
+        method = ""
+
+    params = request.get("params")
+    if not isinstance(params, dict):
+        params = {}
+
+    return method, params
+
+
 def handle_request(request):
     global _last_request_time
     if not isinstance(request, dict):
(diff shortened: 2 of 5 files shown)


BuboTheWise:
@mvalentsev following up from closing our duplicate PR #2432 as superseded by this one: the matrix you raised over there included two cases I want to make sure this PR explicitly covers — (1) falsy params (params: [] / "" / 0 / false should still answer, not error), and (2) id: null (JSON-RPC allows a null id, a reply is required, and request.get("id") must not conflate it with a missing id). If this PR already handles both, no action needed — just confirming they are on the test matrix here rather than lost to the close.

mvalentsev:
Both are on the matrix here, and one neighbouring edge stays exactly as it is on `develop`.

**(1) falsy `params`.** Each of `[]`, `0`, `""`, `false` is covered twice: `test_odd_params_never_raise` sends them to `initialize`, the branch that used to raise, and `test_params_ignoring_methods_keep_working` sends them to `ping`, which never reads params. Both assert a `result` comes back, so the normalisation to `{}` these values always got is pinned rather than tightened.

**(2) `id: null`.** `test_null_id_is_a_request_and_gets_answered` asserts `handle_request({"id": null, "method": "ping", "params": []})` returns `{"jsonrpc": "2.0", "id": null, "result": {}}`.

What this PR deliberately does not change: `request.get("id")` still cannot separate `id: null` from an absent `id`, so an *unknown* method with `id: null` gets no reply, and a known method sent with no `id` at all gets one carrying `id: null`. Both behave that way on `develop` too (I re-measured on `e7102f97` through the stdio loop rather than by reading the code), and splitting the two cases needs a sentinel value, which is a different change from the one this PR makes.

PR #2233, opened 2026-08-12
fix(hooks): support current Codex item_completed events

### LINKED ISSUE #2230: Codex item_completed transcript events are not counted or mined by auto-save hooks
**What happened?**

  MemPalace 3.7.0 auto-save hooks do not recognize the current Codex transcript format.

  Current Codex sessions store canonical conversation events as:

  ```json
  {
    "type": "event_msg",
    "payload": {
      "type": "item_completed",
      "item": {
        "type": "UserMessage",
        "content": [
          {
            "type": "text",
            "text": "Hello"
          }
        ]
      }
    }
  }
  ```

  Assistant messages use the same structure with `"type": "AgentMessage"`.

  However:

  - `hooks_cli._count_human_messages()` only recognizes the legacy `event_msg → user_message` format.
  - `hooks_cli._extract_recent_messages()` also only recognizes the legacy format.
  - `normalize._try_codex_jsonl()` only recognizes `user_message` and `agent_message`.

  In a real transcript containing 20 `item_completed/UserMessage` records, `_count_human_messages()` returns `0`. Therefore the 15-message auto-save threshold is
  never reached and no checkpoint or transcript drawers are created.

  The hooks themselves are registered and executed successfully.

  **What did you expect?**

  MemPalace should:

  - Count `event_msg → item_completed → UserMessage` as a human message.
  - Extract text from `item.content[].text`.
  - Normalize `UserMessage` and `AgentMessage` as conversation turns.
  - Continue supporting the legacy `user_message`/`agent_message` format.
  - Avoid parsing `response_item` records because they may duplicate canonical events or contain synthetic context.

  **How to reproduce:**

  1. Create a JSONL file containing:

  ```json
  {"type":"session_meta","payload":{}}
  {"type":"event_msg","payload":{"type":"item_completed","item":{"type":"UserMessage","content":[{"type":"text","text":"Hello"}]}}}
  {"type":"event_msg","payload":{"type":"item_completed","item":{"type":"AgentMessage","content":[{"type":"Text","text":"Hi"}]}}}
  ```

  2. Run `_count_human_messages()` against this file.
  3. Observe that it returns `0` instead of `1`.
  4. Try normalizing or mining the transcript.
  5. Observe that the conversation is not recognized.

  **Suggested fix:**

  Add support for `item_completed` in:

  - `mempalace/hooks_cli.py::_count_human_messages`
  - `mempalace/hooks_cli.py::_extract_recent_messages`
  - `mempalace/normalize.py::_try_codex_jsonl`

  The role mapping is:

  ```python
  {
      "UserMessage": "user",
      "AgentMessage": "assistant",
  }
  ```

  Text can be extracted from string-valued `item.content[].text` fields.

  Please keep the existing legacy format support and add regression tests for both schemas.

  **Environment:**

  - OS: macOS 26.6.1
  - Python version: 3.14.4
  - MemPalace version: 3.7.0
  - Codex CLI version: 0.147.0
  - Hook setting: `silent_save=true`

fallenmi:
## What does this PR do?

Fixes #2230.

Current Codex rollouts persist canonical turns as `event_msg` / `item_completed` records, with role and content nested under `item.type` and `item.content`. MemPalace only handled the legacy `user_message` / `agent_message` payloads, so current user turns were not counted, the 15-message save threshold was never reached, and normalization discarded the conversation.

This PR:

- adds one strict Codex event-turn extractor shared by normalization, Python hooks, and the portable shell-hook helper;
- maps `UserMessage` and `AgentMessage` items to canonical user and assistant turns;
- accepts both bare-string spans and string-valued `text` fields, preserving order and mixed `text` / `Text` tags;
- counts current user turns and includes them in recent-message context;
- preserves the legacy counter contract, including string-valued empty and whitespace messages;
- keeps `response_item` records excluded to avoid duplicate turns and synthetic context;
- ignores malformed items, non-string fields, and blank current-schema turns safely.

The mixed casing is intentional rather than heuristic: Codex's upstream serde definitions keep turn-item variants as `UserMessage` / `AgentMessage`, while the envelope is snake_case and user/agent text-span tags serialize as `text` / `Text`. Extraction therefore keys on a string-valued `text` field instead of case-sensitive block tags.

The global `NORMALIZE_VERSION` is intentionally unchanged. Bumping it would force every source type in an existing palace to be re-read and re-embedded. New and active sessions are fixed by this patch. Users can backfill sessions that were not saved under the old parser with:

```bash
mempalace mine ~/.codex/sessions/ --mode convos
```

## How to test

```bash
python -m pytest -q \
  tests/test_hooks_cli.py \
  tests/test_normalize.py \
  tests/test_hook_shell.py \
  tests/test_hooks_shell.py \
  tests/test_hooks_bash_compat.py \
  tests/test_save_hook_mines.py \
  tests/test_legacy_shell_hooks.py \
  tests/test_save_hook_verbose.py

pre-commit run --all-files
```

Local focused result: 375 passed, 1 skipped. A broader repository run passed 4,303 tests with 31 skipped; only the unchanged environment-sensitive `test_init_filters_sys_path_from_leaked_pythonpath` subprocess assertion was excluded. Ruff check, Ruff format, and `git diff --check` pass.

Regression coverage includes current and legacy rollout schemas, bare and multi-block content, mixed tag casing, malformed and empty items, `response_item` exclusion, recent-message extraction, both shipped hook implementations, and the 15-message stop-hook save path.

## Checklist

- [x] Focused and adjacent tests pass
- [x] No hardcoded paths
- [x] Linter and formatter pass


mvalentsev:
I was working the same issue in parallel and had a branch ready when this went up. Yours landed first and the design is the same one I arrived at, a single shared extractor, so I am not opening a competing PR. Here is everything I found that might be useful to you instead.

Two things I think are defects, both run against the extractor from this diff rather than read off it.

**Legacy turns with empty text stop counting.** The extractor ends with `if not isinstance(text, str) or not text.strip(): return None`, which also gates the legacy branch. On develop these three records count as 3 exchanges:

```python
{"type": "event_msg", "payload": {"type": "user_message", "message": ""}}
{"type": "event_msg", "payload": {"type": "user_message"}}
{"type": "event_msg", "payload": {"type": "user_message", "message": "   "}}
```

`_count_human_messages` on develop returns 3 for that file, because the old branch checked only `isinstance(msg_text, str)` and a missing key defaulted to `""`. With this patch each returns `None`, so the count is 0 and the 15-exchange threshold arrives later than it used to for legacy Codex sessions. Counting empty messages is arguably wrong, but it is a behaviour change on a path the PR body says is preserved, so it is worth either keeping the legacy branch byte-identical or calling the change out.

**Bare-string spans are dropped.** `for block in content: if not isinstance(block, dict): continue` means a span that is a plain string contributes nothing:

```python
content=["hello"]                                  -> None          # whole turn lost
content=["bare", {"type": "text", "text": "block"}] -> ("user", "block")  # "bare" lost silently
```

The second shape is the one to worry about: the turn is still counted and mined, just short, so nothing signals the loss. `_extract_content` in the same module does accept bare strings inside a content list, which is the precedent I would follow.

Three things you may want that are not defects.

**There is a fourth counter this does not reach.** `hook_shell.count_human_messages` reads only the Claude Code shape and has never understood any Codex schema, legacy included. It backs `hooks/mempal_save_hook.sh:229`, and `hooks/README.md` documents that path for Codex CLI, so users on the shell hooks still get no auto-save. Pre-existing, not something this PR broke, and arguably out of scope since that module deliberately imports nothing beyond the stdlib, but a maintainer will ask.

**The casing is derivable, not just observed.** The user and agent sides disagree because of serde attributes upstream: `TurnItem` is `#[serde(tag = "type")]` with no `rename_all`, so items stay `UserMessage` / `AgentMessage`, while `UserInput` is `rename_all = "snake_case"` and tags text spans `"text"`, and `AgentMessageContent` has no such attribute and tags them `"Text"`. Same for the envelope: `RolloutItemWire` and `EventMsg` are both snake_case, which is why the line is `event_msg` / `item_completed`. Worth stating in the PR, because otherwise the mixed casing reads like a guess. Also relevant to the type-agnostic span read: no non-text variant of either span enum carries a `text` field today, but `UserInput` is `#[non_exhaustive]`.

**Sessions already lost stay lost.** The symptom is that nothing was saved at all, so anyone who ran Codex 0.147 against 3.7.0 has transcripts on disk and nothing in the palace. Merging this fixes new sessions; `mempalace mine ~/.codex/sessions/ --mode convos` picks up the old ones with the fixed parser. A line in the CHANGELOG would help the reporter more than anything else here.

fallenmi:
Thanks — this was an excellent review and the runnable examples caught two real compatibility gaps.

Updated in `a04799e`:

- restored the legacy counter behavior for missing, empty, and whitespace string messages, while downstream normalization still drops blank turns;
- accepted bare-string spans alongside object `text` fields, with mixed-content ordering pinned by tests;
- routed the documented portable shell hook through the same extractor, covering both legacy and current Codex schemas while still excluding commands and `response_item` duplicates;
- added the CHANGELOG recovery command for sessions already present on disk;
- documented why the upstream serde shapes legitimately mix `event_msg` / `item_completed`, `UserMessage` / `AgentMessage`, and `text` / `Text`.

Validation after rebasing onto current `develop`: 375 focused/adjacent tests passed (1 skipped), the broader run passed 4,303 tests (31 skipped, excluding one unchanged environment-sensitive subprocess-path assertion), and pre-commit is clean.

Timofa:
A synthetic recovery repro found one gap in the proposed backfill command.

Using a six-line fixture with only `session_meta`, canonical `event_msg/item_completed` user/assistant turns, and one synthetic `response_item`:

1. Mine it with current `develop`: one drawer plus the completion marker are written, and the drawer contains raw JSONL (including the `response_item`).
2. Mine the same unchanged source and palace with PR head `a04799e`: `Files skipped: 1`, `Drawers filed: 0`.
3. Mine it into a fresh palace with this PR: one canonical transcript drawer plus marker are written, and the `response_item` is absent.

The old raw-fallback mine already writes a completion marker at `NORMALIZE_VERSION=2`. Because this PR intentionally keeps the version at 2 and the source fingerprint is unchanged, `mempalace mine ~/.codex/sessions/ --mode convos` repairs sessions that were never filed, but not completed sessions already filed as raw JSON.

I agree that a global normalization-version bump is broader than this fix needs. Could the recovery note use a scoped Codex migration/force-reparse path (or documented exact-source delete + re-mine) for already-filed sources? The parser itself passes the fresh-palace control; this finding is only about recovery of existing raw-fallback markers.

Taylor-yuhb:
I independently reproduced a recovery gap in the CHANGELOG note.

With a six-line current-Codex fixture:

1. Current `develop` (`4bc0c43`) mines it as raw JSONL, including the synthetic `response_item`, while stamping `normalize_version=2` and the source mtime.
2. PR head `a04799e` against the same unchanged source and palace reports `Files skipped (already filed): 1` and `Drawers filed: 0`; the raw drawers remain.
3. PR head against a fresh palace produces the canonical transcript and excludes the `response_item`.

So the parser and hook fix look good, but `mempalace mine ~/.codex/sessions/ --mode convos` only backfills sessions that were never filed. It does not repair sessions that the old parser already filed as raw JSON.

Could the recovery note be narrowed, or could we add a scoped safe reparse/migration path for this Codex source class? A regression test covering old raw-fallback drawers → unchanged source → fixed parser would protect this behavior. I agree that a global normalization-version bump would be broader than necessary.

fallenmi:
Thanks @Timofa and @Taylor-yuhb — both reproductions are correct. Updated in 87c0f2a after rebasing onto current develop: the CHANGELOG now limits the mine command to sessions that were never filed and explicitly says unchanged raw-fallback sessions already stamped at normalization version 2 remain skipped and need a future scoped migration.

This PR deliberately does not add a global version bump or a delete-and-re-mine workaround: either would widen a parser/hook fix into a data migration with non-destructive-ingest implications. For the same reason, an old-raw-to-reparsed regression belongs with that future scoped migration rather than asserting unsupported behavior here.

Post-rebase validation: 375 focused/adjacent tests passed with 1 skipped; the full suite passed 4,501 tests with 32 skipped; pre-commit passed.

PR #2235, opened 2026-08-12
fix(config): handle non-object JSON documents

### LINKED ISSUE #2234: config: valid non-object config.json crashes all consumers
**What happened?**

A syntactically valid `config.json` whose top-level value is not an object crashes configuration consumers. JSON permits `null`, arrays, strings, numbers, and booleans at the top level, but `MempalaceConfig` stores any successful `json.load()` result and every setting property later assumes a mapping with `.get()`.

For example, a file containing only `null` makes `cfg.palace_path` raise:

```text
```

Equivalent crashes occur for `[]`, `"oops"`, `42`, and `true`. Because `MempalaceConfig` is shared by the CLI, MCP server, mining, and hooks, a single structurally invalid file can prevent all of those entry points from starting normally.

**What did you expect?**

Treat a non-object JSON document as an unusable configuration and fall back to defaults, matching the existing fail-soft behavior for malformed JSON, Unicode errors, and unreadable files. The invalid file should not be rewritten implicitly.

**How to reproduce:**

```python
import json
import tempfile
from pathlib import Path

from mempalace.config import MempalaceConfig

config_dir = Path(tempfile.mkdtemp())
(config_dir / "config.json").write_text("null\n", encoding="utf-8")
print(MempalaceConfig(config_dir=config_dir).palace_path)
```

Repeat with `[]`, `"oops"`, `42`, or `true`.

**Root cause**

`MempalaceConfig.__init__` assigns the result of `json.load()` directly to `_file_config` without validating that it is a dictionary. The first property access then calls `.get()` on the scalar or array.

**Environment:**

- OS: macOS 26 (arm64)
- Python version: 3.12.12
- MemPalace version: `develop` at `b210423`

fallenmi:
## What does this PR do?

- Validates that a successfully parsed `config.json` is a JSON object before storing it as MemPalace configuration.
- Falls back to the existing empty/default configuration for valid JSON scalars and arrays instead of crashing on the first `.get()` call.
- Adds regressions for `null`, array, string, number, and boolean top-level documents.

Closes #2234.

## Why?

JSON permits non-object values at the top level. `MempalaceConfig` previously assigned any successful `json.load()` result to `_file_config`, while all downstream properties assume a mapping. A file containing `null`, `[]`, a string, a number, or a boolean therefore raised `AttributeError` across CLI, MCP, mining, and hook consumers.

The fallback matches the class's existing fail-soft behavior for malformed JSON, Unicode decode errors, and unreadable files. It does not rewrite the user's file, emit output on the MCP stdio channel, or alter valid object configurations.

## How to test

```bash
python -m pytest -q tests/test_config.py tests/test_hooks_cli.py
pre-commit run --all-files
```

Local result after rebasing onto current `develop`: 286 passed, 1 skipped. Ruff check, Ruff format, and `git diff --check` pass.

## Checklist

- [x] Focused and adjacent tests pass
- [x] No hardcoded paths
- [x] Linter and formatter pass

### DIFF
--- a/mempalace/config.py
+++ b/mempalace/config.py
@@ -375,7 +375,8 @@ class MempalaceConfig:
         if self._config_file.exists():
             try:
                 with open(self._config_file, "r", encoding="utf-8") as f:
-                    self._file_config = json.load(f)
+                    loaded = json.load(f)
+                    self._file_config = loaded if isinstance(loaded, dict) else {}
             except (json.JSONDecodeError, UnicodeDecodeError, OSError):
                 self._file_config = {}
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -5,6 +5,7 @@ import tempfile
 
 import pytest
 from mempalace.config import (
+    DEFAULT_PALACE_PATH,
     MempalaceConfig,
     normalize_wing_name,
     sanitize_iso_date,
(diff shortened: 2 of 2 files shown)


fallenmi:
Merged PR #2357 supersedes this fix with a broader read/write path: non-object JSON is kept out of the mapping config, original bytes are preserved before a fresh write, and writes are crash-safe. Closing the now-conflicting duplicate rather than rebasing the narrower fallback. Codex assisted with this status check; the account owner authorized the closure.

PR #2243, opened 2026-08-13
fix: decode git subprocess output as UTF-8 regardless of locale

gurkansanli:
## What does this PR do?

Fixes a startup crash of `mempalace init` on Windows systems whose ANSI codepage is not UTF-8 (e.g. Turkish cp1254), triggered whenever the git identity or commit history contains non-ASCII characters.

Reported on a Turkish-locale Windows 11 setup with `git config user.name` = `Gürkan ŞANLI`:

```
Exception in thread Thread-1 (_readerthread):
...
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9e in position 9: character maps to <undefined>
...
File "mempalace/project_scanner.py", line 267, in _git_user_identity
  name = _run_git(repo, "config", "user.name", timeout=2).strip()
AttributeError: 'NoneType' object has no attribute 'strip'
```

**Root cause:** the three git `subprocess.run()` calls in `project_scanner.py` use `text=True` without an explicit encoding, so CPython decodes the pipe with `locale.getpreferredencoding()` — the ANSI codepage on Windows. git, however, emits UTF-8 on every platform. `Ş` encodes to `0xc5 0x9e`, and `0x9e` is undefined in cp1254, so the decode blows up inside subprocess's stdout reader thread; `subprocess.run()` then returns a `CompletedProcess` with `stdout=None` (returncode still 0), and `_run_git`'s `return r.stdout if r.returncode == 0 else ""` propagates that `None` into `.strip()`.

**Fix:**

- pass `encoding="utf-8", errors="replace"` to all three git subprocess calls (`_run_git` and both calls in `_global_git_identity`); `errors="replace"` keeps repos with legacy-encoded history from aborting a whole scan
- harden `_run_git` to return `""` if stdout is ever `None`, since every caller chains `.strip()`/`.splitlines()` on the result

## How to test

`uv run pytest tests/test_project_scanner.py -v` — 5 new regression tests:

- unit: `_run_git`/`_global_git_identity` must request UTF-8 decoding and must never propagate a `None` stdout
- integration (real git): a `Gürkan ŞANLI` identity survives both the `git config user.name` and `git log --format=%aN|%aE` paths through `scan()`; on a non-UTF-8 Windows codepage the unfixed code either crashed outright (cp1254) or mojibake'd the name (cp1252), so the equality assertions catch both failure modes

All five new tests fail on the unpatched code (verified on the affected cp1254 machine, reproducing the original traceback verbatim) and pass with the fix.

Manual repro: on Windows with a non-UTF-8 ANSI codepage (Turkish system locale → cp1254), set a non-ASCII `git config user.name` and run `mempalace init .` in any git repo — crashes before this patch, completes after.

Note on the full suite: on the affected Windows machine `pytest tests/ --ignore=tests/benchmarks` gives 3954 passed / 15 failed; all 15 are in `test_claude_plugin_hook_wrappers.py` and `test_mcp_http_transport.py` and reproduce identically on a clean `develop` checkout (local bash/socket environment), i.e. unrelated to this change.

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/project_scanner.py
+++ b/mempalace/project_scanner.py
@@ -260,10 +260,18 @@ def _run_git(cwd: Path, *args: str, timeout: int = GIT_TIMEOUT) -> str:
             ["git", "-C", str(cwd), *args],
             capture_output=True,
             text=True,
+            # git emits UTF-8 on every platform, but bare text=True decodes
+            # with the locale codepage (e.g. cp1254 on Turkish Windows), which
+            # crashes the stdout reader thread on non-ASCII names ("Ş" is
+            # 0xc5 0x9e in UTF-8; 0x9e is undefined in cp1254) and silently
+            # leaves r.stdout as None. Decode explicitly; errors="replace"
+            # keeps legacy-encoded history from aborting a whole scan.
+            encoding="utf-8",
+            errors="replace",
             timeout=timeout,
             check=False,
         )
-        return r.stdout if r.returncode == 0 else ""
+        return (r.stdout or "") if r.returncode == 0 else ""
     except (OSError, subprocess.SubprocessError):
         return ""
 
@@ -281,6 +289,9 @@ def _global_git_identity() -> tuple[str, str]:
             ["git", "config", "--global", "user.name"],
             capture_output=True,
             text=True,
+            # Same UTF-8 rationale as _run_git above.
+            encoding="utf-8",
+            errors="replace",
             timeout=2,
             check=False,
         ).stdout.strip()
@@ -288,6 +299,8 @@ def _global_git_identity() -> tuple[str, str]:
             ["git", "config", "--global", "user.email"],
             capture_output=True,
             text=True,
+            encoding="utf-8",
+            errors="replace",
             timeout=2,
             check=False,
         ).stdout.strip()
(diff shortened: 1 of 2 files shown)


moondrop-ro:
Reproduced on **3.10.0** (`uv tool install`, Windows 11, Python 3.13, locale encoding cp1252), and this PR fixes it.

The trigger is this repo's own history: a contributor name containing `ọ` (U+1ECD, UTF-8 `E1 BB 8D`; `0x8D` is undefined in cp1252). So on a cp1252 Windows machine, `mempalace init` crashes on any directory that contains a clone of mempalace:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 30648: character maps to <undefined>
...
  File "mempalace/project_scanner.py", line 309, in _git_authors
    for line in out.splitlines():
AttributeError: 'NoneType' object has no attribute 'splitlines'
```

Same root cause as described above, different call site (`_git_authors` rather than `_git_user_identity`); both go through `_run_git`. Still unfixed on `develop` as of `c4d3711`.

Tested against a clone at `22fd87f` (current `main`), locale left at cp1252, calling `project_scanner.scan()` directly:

| code | result |
|---|---|
| 3.10.0 as installed | the `AttributeError` above |
| this PR's `project_scanner.py` (`399f957`) | completes, 40 people detected |
| 3.10.0 with `PYTHONUTF8=1` | completes (workaround until this lands) |

PR #2248, opened 2026-08-13
feat(gossip): background daemon for multi-hop re-propagation

jjcav84:

## Summary
Adds a background `GossipDaemon` for multi-hop gossip propagation:
- Refactors `GossipProtocol.propagate()` into a `_propagate_message()` method that returns child messages for each target.
- `GossipDaemon` holds a `deque` queue and processes it on a background thread.
- `run_once()` propagates each pending message, then enqueues child messages with incremented `hops`.
- Drops expired messages and messages that have reached `max_hops`.
- `schedule()`, `start()`, `stop()` API for callers.

## Test plan
- [x] `tests/test_gossip.py` passes (19 tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ---
 
+## [Unreleased]
+
+### Bug Fixes
+
+- **`sweep` books a failed `stat` as a failure again.** The non-regular-file gate added in 3.7.1 reads `stat.S_ISREG(f.stat().st_mode)` inside a `try`, and its `except OSError` printed `SKIP` and moved on. A dangling symlink, a symlink loop and a file unlinked between `rglob` and the gate all raise there, and before the gate existed every one of them reached `sweep()` and was booked in `failures` — so `sweep` went from reporting a transcript it could not read to reporting success. A failed probe is now an error, not a benign file type: it is logged, printed as `WARNING`, and appended to `failures`, while a probe that succeeds and says "not regular" still skips silently. (#2221)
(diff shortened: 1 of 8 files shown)


jjcav84:

jjcav84:

fatkobra (changes requested):
Thanks for adding daemon-driven gossip processing. I found two blocking
queue-safety issues.

### 1. `max_requeue` does not actually cap the resulting queue

The generated children are filtered using a condition equivalent to:

`len(self._queue) < self.max_requeue`

However, `self._queue` is not extended until after that filtering finishes.
Every child therefore sees the same queue length.

For example, if the queue contains 9 entries and `max_requeue` is 10, every
one of 100 generated children passes the condition, and the queue can grow
to 109 entries.

Please calculate the available capacity once while holding the queue lock:

- account for entries already queued by concurrent callers;
- retain at most the available number of eligible children;
- append those children under the same lock.

Please add tests with:

1. more generated children than the total cap;
2. an existing queue that leaves only one available slot;
3. messages queued concurrently while `run_once()` is processing.

### 2. One processing exception can discard unrelated pending messages

`run_once()` copies the current queue and clears it before processing the
messages.

The loop has no per-message failure boundary. If message N raises:

- all later messages from the copied pending batch are lost;
- children already produced by earlier messages are not appended;
- the outer daemon loop logs the exception but cannot reconstruct the lost
  work.

Please isolate processing failures per message and define an explicit policy,
such as:

- retain or requeue messages that were not yet processed;
- move the failed message to a retry or dead-letter path;
- preserve children successfully produced before the failure;
- expose the failure through status or logs without discarding unrelated
  work.

A deterministic regression should queue three messages, make the second one
raise, and verify that the third message and the first message's generated
children remain available.

changes are resolved.

jjcav84:

jjcav84:

jjcav84:

jjcav84:

PR #2249, opened 2026-08-13
feat(gossip): meta-gossip analytics and MCP tool

jjcav84:

## Summary
Adds meta-gossip analytics to close the loop on the design spec:
- `KnowledgeGraph.query_gossip_triples()` — query all triples with `source_file LIKE 'gossip://%'`.
- `GossipAnalytics` with `trending_topics()`, `viral_facts()`, and `network_health()`.
- `GossipProtocol.analytics()` exposes the snapshot.
- New MCP tool `mempalace_gossip_analytics`.
- Updated `website/reference/mcp-tools.md`.

## Test plan
- [x] `tests/test_gossip.py` passes.
- [x] `tests/test_knowledge_graph.py` passes.
- [x] `tests/test_readme_claims.py` passes (no undocumented tools).
- [x] Full suite passed before doc update; `TestNoUnlistedTools` re-ran and passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ---
 
+## [Unreleased]
+
+### Bug Fixes
+
+- **`sweep` books a failed `stat` as a failure again.** The non-regular-file gate added in 3.7.1 reads `stat.S_ISREG(f.stat().st_mode)` inside a `try`, and its `except OSError` printed `SKIP` and moved on. A dangling symlink, a symlink loop and a file unlinked between `rglob` and the gate all raise there, and before the gate existed every one of them reached `sweep()` and was booked in `failures` — so `sweep` went from reporting a transcript it could not read to reporting success. A failed probe is now an error, not a benign file type: it is logged, printed as `WARNING`, and appended to `failures`, while a probe that succeeds and says "not regular" still skips silently. (#2221)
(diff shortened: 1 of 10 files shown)


jjcav84:

jjcav84:

fatkobra (changes requested):
I found two blocking analytics issues.

### 1. Viral-fact identity is fragmented by destination

`viral_facts` groups gossip records by the stored
`(subject, predicate, object)` triple.

Propagation currently replaces the original fact's predicate with
destination-specific predicates such as:

- `gossiped_in_<wing>`;
- `gossiped_in_room_<room>`.

The same original fact spreading through several wings or rooms therefore
appears as several unrelated facts rather than as one fact with broad
network spread.

The stored derived triple also does not retain enough stable identity to
recover the original predicate reliably.

Please preserve a stable original-fact identity or explicit provenance, and
calculate spread using distinct destinations or propagation events rather
than destination-specific predicates as the identity key.

A regression should propagate one original fact into multiple wings and
rooms and verify that analytics reports:

- one viral fact;
- its original identity;
- the correct number of distinct destinations;
- and the aggregate spread or reinforcement count.

### 2. Historical `as_of` metrics mix two different moments in time

The active metrics use the requested `as_of` snapshot, but the expired-fact
calculation performs a separate unfiltered query and compares records with
the current `datetime.now()`.

A historical analytics request can therefore combine:

- an active count from the requested past timestamp; and
- an expired count from today.

Please use one reference timestamp and one consistent snapshot for every
metric in the response.

A regression should create a fact that was active at a past `as_of` value
but is expired today, then verify that every returned metric describes the
past snapshot rather than mixing past and present state.

implementation is corrected.

jjcav84:

jjcav84:

jjcav84:

jjcav84:

PR #2250, opened 2026-08-13
feat(gossip): priority channels, noise tolerance, amplification, random walk

jjcav84:

## Summary
Implements the priority-channel model and several remaining chatter-behavior parameters from the gossip design spec:
- Maps message priority to one of four channels: `critical`, `high`, `medium`, `low`.
- Each channel carries `ttl_seconds`, `fanout`, `gossip_probability`, `max_hops`, plus `latency_ms`/`reliability`/`capacity` metadata.
- `select_chatter_nodes` enforces `noise_tolerance` (drops low-relevance nodes).
- `_should_forward` applies the channel's `gossip_probability`, `chatter_level` multiplier, and `amplification_factor` for critical/high messages.
- Random-walk gossip: occasional swap of a selected node for an off-radius node (up to 5 steps).
- `chatter_status` now returns `channels`, `noise_tolerance`, `amplification_factor`, `randomness_factor`, and `gossip_probability`.

## Test plan
- [x] `tests/test_gossip.py` passes (18 tests).
- [x] Full suite: 4315 passed, 31 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ---
 
+## [Unreleased]
+
+### Bug Fixes
+
+- **`sweep` books a failed `stat` as a failure again.** The non-regular-file gate added in 3.7.1 reads `stat.S_ISREG(f.stat().st_mode)` inside a `try`, and its `except OSError` printed `SKIP` and moved on. A dangling symlink, a symlink loop and a file unlinked between `rglob` and the gate all raise there, and before the gate existed every one of them reached `sweep()` and was booked in `failures` — so `sweep` went from reporting a transcript it could not read to reporting success. A failed probe is now an error, not a benign file type: it is logged, printed as `WARNING`, and appended to `failures`, while a probe that succeeds and says "not regular" still skips silently. (#2221)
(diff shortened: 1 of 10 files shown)


jjcav84:

jjcav84:

fatkobra (changes requested):
I found two blocking channel-semantics issues.

### 1. An explicit `max_hops` argument is overwritten by the channel default

The preliminary message correctly incorporates the caller's explicit
`max_hops` value.

When the final message is constructed, however, the implementation reads
`max_hops` from the channel parameters again. Since every built-in channel
defines a value, the caller's explicit value is replaced by the channel
default.

The precedence should be:

1. explicit caller argument;
2. channel default;
3. global fallback.

Please preserve the resolved value through the complete message-building
path.

Regression coverage should verify that:

- an explicit `max_hops=10` remains 10 even on a channel whose default is 2;
- omitting the argument uses that channel's value of 2;
- channels without a value use the global fallback.

### 2. The random walk does not ensure that the selected node is off-radius

`_random_walk_swap()` selects from any chatter node that was not already
selected by the normal ranking.

“Not selected” is not equivalent to “outside the source's gossip radius.”
A lower-ranked node can still be inside the same permitted radius.

Please either:

- calculate an actual off-radius candidate set and choose only from it; or
- change the feature contract and documentation so it no longer claims
  off-radius exploration.

The test should make the distinction observable by providing:

- one unselected node that is still inside the radius;
- one node that is genuinely outside it;
- deterministic random selection;

and verifying that only the genuine off-radius node is eligible.

changes are complete.

jjcav84:

jjcav84:

jjcav84:

jjcav84:

PR #2251, opened 2026-08-13
feat(gossip): echo-chamber attenuation with path vector

jjcav84:
Stacked on #2248 (feat/gossip-daemon).

## Summary
Adds echo-chamber attenuation to the gossip protocol:
- `GossipMessage` now carries a lightweight `path` vector (fixed-size hop history of chatter node ids).
- `child()` appends the forwarding `node_id` to the path.
- `_should_forward` checks prior visits. Each re-visit multiplies the forwarding probability by `echo_chamber_attenuation` (0.5). After `echo_chamber_reinforcement_count` (3) visits the node is suppressed entirely.
- New config keys: `echo_chamber_reinforcement_count`, `echo_chamber_attenuation`, `echo_chamber_similarity_threshold`.
- `chatter_status` returns the echo-chamber parameters.
- New test verifies daemon attenuation and path preservation.

## Test plan
- [x] `tests/test_gossip.py` passes (20 tests).
- [x] Full suite: 4317 passed, 31 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ---
 
+## [Unreleased]
+
+### Bug Fixes
+
+- **`sweep` books a failed `stat` as a failure again.** The non-regular-file gate added in 3.7.1 reads `stat.S_ISREG(f.stat().st_mode)` inside a `try`, and its `except OSError` printed `SKIP` and moved on. A dangling symlink, a symlink loop and a file unlinked between `rglob` and the gate all raise there, and before the gate existed every one of them reached `sweep()` and was booked in `failures` — so `sweep` went from reporting a transcript it could not read to reporting success. A failed probe is now an error, not a benign file type: it is logged, printed as `WARNING`, and appended to `failures`, while a probe that succeeds and says "not regular" still skips silently. (#2221)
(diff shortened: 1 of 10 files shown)


jjcav84:

jjcav84:

fatkobra (changes requested):
I found two child-specific blockers in addition to this branch's dependency
on #2248.

### 1. `echo_chamber_similarity_threshold` is currently a dead setting

The configuration adds `echo_chamber_similarity_threshold`, and the PR
describes similarity-based echo-chamber control.

The forwarding decision currently uses path node IDs, reinforcement count,
attenuation, and randomness, but it never reads or applies the similarity
threshold. `chatter_status` also omits the setting.

Please either:

- implement the documented similarity semantics and expose the active value
  in status; or
- remove the setting and the similarity claim until that behaviour exists.

If implemented, regression coverage should demonstrate messages immediately
below and above the threshold, with deterministic forwarding outcomes.

### 2. The attenuation test passes even when attenuation has no effect

The current test accepts `children_queued >= 0`, which is true for every
possible result.

It then inspects child paths only when children happened to be produced.
Producing zero children therefore makes the entire test pass without proving
attenuation, reinforcement suppression, or path handling.

Please make the forwarding tests deterministic by controlling
`random.random()` and directly exercise the boundary cases:

- no previous visits;
- one previous visit at the attenuation boundary;
- enough reinforcement to suppress forwarding;
- forwarding permitted below the threshold;
- forwarding rejected above the threshold.

Path preservation should be asserted separately from the forwarding
decision.

This branch is stacked on #2248. Final approval should wait until the queue
cap and exception-loss fixes from that parent have been rebased here.

jjcav84:

jjcav84:

jjcav84:

jjcav84:

Issue #2240, opened 2026-08-12
mempalace-issue-fts5-autoheal


darkrhodes7:
# FTS5 auto-heal is unreachable when the serving interpreter's SQLite cannot detect the fault

**Version:** 3.7.0 (also present in 3.5.0)
**Platform:** Windows 11, `uv tool install mempalace`
**Backend:** chroma, default `minilm` embedder
**Palace:** 322,848 drawers, `chroma.sqlite3` ~1.95 GB, 325,451 `embedding_fulltext_search_content` rows

## Summary

`repair.maybe_autoheal_fts5_index` only runs when `repair.sqlite_integrity_errors`
returns a non-empty list. That function asks the SQLite bundled with whichever
interpreter is running MemPalace. On this install that is **SQLite 3.47.1**, and
3.47.1 does not detect the FTS5 fault present in this palace, while **SQLite
3.50.4** reports it on the same file through the same `mode=ro` URI.

The result is that a palace can carry a malformed FTS5 inverted index
indefinitely while every MemPalace-run check — the startup integrity probe,
`repair`'s preflight, and the post-mine validation in
`palace._validate_palace_fts5_after_mine` — reports it clean. The auto-heal
added for exactly this condition can never fire.

Separately, and more surprising: **a `'rebuild'` issued by 3.47.1 does not clear
the fault**, while the same statement issued by 3.50.4 does. So even if
detection were fixed, the heal would still no-op on this interpreter.

## Evidence

Same file, same URI, both pragma forms, minutes apart:

| SQLite | `PRAGMA quick_check(1)` | `PRAGMA quick_check` |
|---|---|---|
| 3.50.4 (CPython 3.14.6) | malformed inverted index for FTS5 table `main.embedding_fulltext_search` | same |
| 3.47.1 (CPython 3.12.9 — MemPalace's uv venv) | ok | ok |

FTS5's own command, from 3.50.4:

```
INSERT INTO embedding_fulltext_search(embedding_fulltext_search) VALUES('integrity-check');
-- DatabaseError: database disk image is malformed
```

Table definition:

```sql
CREATE VIRTUAL TABLE embedding_fulltext_search USING fts5(string_value, tokenize='trigram')
-- embedding_fulltext_search_config: version = 4
```

Rebuild behaviour, on a byte copy of the live database so the result is
attributable to the rebuild alone:

| `'rebuild'` issued by | duration | 3.50.4 `quick_check` afterwards |
|---|---|---|
[...]

Issue #1931, opened 2026-07-05
status/sqlite_integrity hardcodes chroma.sqlite3, always reports ok:true on non-chroma backends


Zoz92:
## Description

`mcp_server.py`'s `_sqlite_integrity_payload()` and `repair.py`'s `sqlite_integrity_errors()` both hardcode the check path as:

```python
sqlite_path = os.path.join(palace_path, "chroma.sqlite3")
```

with no check of the configured `backend` (`config.json` → `"backend"`). On a palace running the `qdrant` (or presumably `pgvector`/`sqlite_exact`) backend where `chroma.sqlite3` doesn't exist, `sqlite_integrity_errors()` hits `if not os.path.exists(sqlite_path): return []` and silently returns no errors — so `mempalace_status` / MCP `status` reports:

```json
"sqlite_integrity": {
  "checked": true,
  "ok": true,
  "sqlite_path": "/path/palace/chroma.sqlite3",
  "error_count": 0,
  "errors": []
}
```

This isn't wrong per se (there's genuinely nothing to check), but it's misleading: it reads as "integrity verified OK" when in fact no check ran at all, and it references a file that no longer exists for the active backend.

## Repro

1. Migrate a palace from `chroma` to `qdrant` (backend registry already supports both — `backends/qdrant.py`, `set_backend()` in `config.py`).
2. Relocate/remove `chroma.sqlite3` from the palace dir (or just never have had one, e.g. a palace created directly with `--backend qdrant`).
3. Call `mempalace_status` / `mempalace status` — `sqlite_integrity.ok` is `true`, `checked` is `true`, despite no actual check having occurred.

## Suggested fix

Gate the check on the active backend (e.g. `if config.backend == "chroma"`), and for non-chroma backends either omit `sqlite_integrity` entirely or report something like `{"checked": false, "ok": null, "reason": "not applicable to backend 'qdrant'"}` so it's clear no verification ran.

## Environment

- mempalace 3.5.0
- Backend: qdrant (migrated from chroma)

mvalentsev:
Opened #1946 for this. The integrity check is already skipped on non-chroma backends, so the palace isn't flagged as corrupt. The only wrong part was status still reporting `checked`/`ok` true with a `chroma.sqlite3` path, like you said. The PR has that report as not-applicable (`checked: false`, `ok: null`, plus a `reason`) on non-chroma and leaves chroma alone.

PR #2254, opened 2026-08-14
fix(repair): anchor the FTS5 classifier, attribute integrity verdicts (#2240)

mvalentsev:
Related to #2240. Implements suggestion 1 from that report and does not close it: the asymmetry it describes did not reproduce here, so the issue stays open (see "What this does not do").

## What does this PR do?

`maybe_autoheal_fts5_index` rebuilds a corrupt FTS5 index in place when `PRAGMA quick_check`'s errors look like they came from that index alone. On `develop` it has six exits, two of them decide silently, nothing says which SQLite produced any verdict, and the classifier matches its phrasings anywhere inside a row.

- **The classifier is anchored to the start of the row.** `sqlite_integrity_errors` reports a failed probe as `PRAGMA quick_check failed: <error>`, a row whose whole meaning is that the state was never established. Unanchored it reads as an isolated FTS5 fault whenever the wrapped text contains a phrasing, and `develop` then takes the mine lock and writes to `chroma.sqlite3`. SQLite does produce such a row: dropping an FTS5 `%_config` shadow table fails the constructor with `vtable constructor failed: <table name>`, name interpolated. A chroma palace only ever has `embedding_fulltext_search`, so this is hardening, not a reachable fix.

- **The two silent exits now speak,** naming the build that decided. `--dry-run` prints the same explanation it withheld, marked as a preview, since unmarked it was byte-for-byte the run it predicts. The post-mine validator moves from `info` to `warning`: nothing configures logging on the `mempalace mine` path, so those records never reached a handler.

- **Every `quick_check` result these paths state names the build behind it,** which is suggestion 1 in #2240. `sqlite3.sqlite_version` now accompanies the abort banner, both dry-run branches, the heal's outcome line, the post-recovery check, all three shapes of the MCP integrity payload, the tool-refusal payload, the integrity log line, and `MineValidationError` — that last because the MCP `mine` tool and the daemon's job runner render the exception rather than the banner.

- **A skipped probe is no longer reported as a clean verdict.** Above `MEMPALACE_STARTUP_INTEGRITY_MAX_MB` (512 MB) no check runs, yet the payload answered `checked: true, ok: true`. It now returns the not-applicable shape #1931 introduced for the same reason, carried by the field #2291 added for a palace with no database at all, so the two absences are one payload shape with different reasons rather than two. The palace in #2240 is ~1.95 GB, so there the probe never runs at all. Write tools gate on recorded errors rather than `ok`, so nothing is newly refused, and `/statusz` has read an absent verdict as not-unhealthy since #2291, so naming this one does not turn a server red. The skip message also stopped rounding itself into "is 1 MB, over the 1 MB limit" for a 1.4 MB database.

### `fts5: checksum mismatch` stays unmatched on purpose

It is the largest wording the classifier does not match on 3.51.2 — 486 of 1,032 unmatched rows below. Widening to the whole `fts5:` prefix would be wrong: the message says index and content disagree without naming the damaged side, and a rebuild reads *from* content. Damaging `%_content` produces it while the index still holds the original token, and single-byte index damage produces it too, in 75 of the 115 flips that registered in a 120-trial run. Accepting it would rebuild from whichever side is broken and leave `quick_check` clean either way. A comment and a test record the counterexample.

### What this does not do

The asymmetry #2240 reports is not addressed here and the issue stays open. It did not reproduce: the same 2,975 of 3,200 damaged databases came back dirty on 3.45.1, 3.47.1 and 3.51.2, differing only in wording, with `quick_check` agreeing with FTS5's `integrity-check` 24 times out of 24 on each. Nor did the third observation — a `'rebuild'` left `quick_check` clean on every build. Caveat: I ran 3.45.1/3.47.1/3.51.2 rather than the reporter's 3.47.1/3.50.4 pair, and nowhere near their scale, so this is evidence the asymmetry is not general, not evidence against the report.

Two pre-existing defects surfaced here, and both have since been filed and fixed. Content-side damage produces the older wording on 3.45.1 and 3.47.1, which the classifier accepts, so the heal reindexed damaged text and reported success: that is #2278, fixed in #2279 and on `develop` since 3.8.0. And `sqlite_integrity_errors` returned `[]` when `chroma.sqlite3` is absent, so a palace with no database file reported `checked: true, ok: true`: that is #2290, fixed in #2291 and on `develop` since. The absence this branch reports, a palace nobody examined because it sits above the size limit, goes through the field that fix added, and it asks which backend is running before either reason, since both name a `chroma.sqlite3` a non-chroma server never opens.

Thanks to @darkrhodes7 for the report and its version matrix; the version reporting is their suggestion.

## How to test

```bash
uv run pytest tests/test_repair.py -q
uv run pytest tests/mcp/test_guards.py -q -k "integrity or rendered_mb"
uv run pytest tests/test_miner_fts5_validation.py -q
uv run pytest tests/test_mcp_http_transport.py -q -k statusz
```

Both later wordings are passed to the tests explicitly rather than produced by corrupting a database: 3.45.1 and 3.47.1 emitted neither across the sweep below nor across 115 single-byte index flips each.

## Verification

Against `develop` and this branch, on all three builds, with real palaces, the real CLI and a real MCP server rather than in-process stubs.

| claim | how |
|---|---|
| anchoring accepts and refuses exactly what the old pattern did | 3,200 damaged databases per build (10 modes × 4 tokenizers × 80), 2,975 dirty on each, no row gained or lost either way; sha256 of the dirty set is identical across builds, so it is the same databases and not just the same count |
| the wrapper row authorizes a write on `develop` | FTS5 table named after a phrasing, `%_config` dropped: `develop` calls it isolated and `chroma.sqlite3` changes on disk; this branch declines and the hash is unchanged |
| SQLite really interpolates a table name there | `vtable constructor failed: <name>` on all three builds; the FTS5 row itself arrives unprefixed, `aux.`-qualified from an ATTACHed database |
| a rebuild would hide `checksum mismatch` rather than fix it | damage either side, rebuild, re-probe: `quick_check` comes back clean on all three builds either way |
| `--dry-run` is no longer the run it predicts | same palace, both paths: outputs differ, only the preview says `DRY RUN`, neither writes |
| the skipped-probe payload | real MCP servers over stdio, same 3 MB palace against a 1 MB limit: `develop` answers `checked: true`, `ok: true` and states no build at all; this branch answers `checked: false`, `ok: null`, `sqlite_version`, and a `reason` naming the size and the limit |
| the tests pin all of this | 17 mutations applied one at a time in a copy of the tree — dropped anchor, classifier widened to the whole `fts5:` prefix, unmarked preview, whole-megabyte rounding, rounding that lets a value round away to nothing, reversed skip ordering, a skip that stops naming itself, an absent-verdict payload that stops existing, that payload asked before the backend rather than after it, and each removed `sqlite_version` — all 17 fail the suite |

`4943 passed, 53 skipped` on the CI selection (`--ignore=tests/benchmarks`), reported by CI's own `test-linux (3.11)` job on this branch, rebased onto current `develop`. `ruff check` and `ruff format --check` clean on the pin CI and pre-commit now share (0.16.6).

## Checklist
- [x] No hardcoded paths


darkrhodes7:
Hi [Mikhail Valentsev] & [Igor Lins e Silva],
Good to hear. I just updated from version 1.7.0 and encountered a few
issues during the process, which Claude identified.

I have included the report generated by Claude below for your review.

[# `repair --archive-existing` cannot archive on Windows: the repair
process holds the palace open while renaming it

**Version:** 3.10.0 (also reproduces on 3.7.0)
**Platform:** Windows 11, Python 3.12, chromadb 1.5.9, backend `chroma`
**Severity:** blocks `mempalace repair rebuild-index` entirely on Windows.
Fails closed — no data risk.

---

## Summary

`mempalace repair rebuild-index` archives the existing palace by **renaming
the
palace directory**. On Windows a directory rename fails if any handle
inside it
is open, and the repair process appears to hold the palace database open at
the
moment it attempts that rename. The command therefore refuses its own
archive
step and cannot proceed.

POSIX tolerates renaming a directory containing open files, which is likely
why
this has not surfaced elsewhere.

## Reproduction

```
> mempalace repair rebuild-index --yes

  MemPalace Repair -- Rebuild from SQLite

  Source: C:\Users\shale\.mempalace
  Dest:   C:\Users\shale\.mempalace
  Archiving C:\Users\shale\.mempalace ->
C:\Users\shale\.mempalace.pre-rebuild-20260916-145602

  Cannot archive the existing palace: a file inside it is held open by
another
  process (MCP server, running mine, another harness?).
  Close those processes and retry. The palace was left untouched.
  ([WinError 5] Access is denied:
   'C:\\Users\\shale\\.mempalace' ->
'C:\\Users\\shale\\.mempalace.pre-rebuild-20260916-145602')
```

## Why it is not the MCP servers the message suggests

The message attributes the lock to another process. It was tested:

* Killed all `mempalace-mcp.exe` / `mempalace-light-mcp.exe` processes,
confirmed
  zero remaining, then retried **six consecutive times** in a loop that
re-killed
  before each attempt. Attempts 2–6 each reported `killed 0 server
process(es)` —
  no MemPalace process was running at all — and every attempt failed
identically.
* **An external `Rename-Item` on the same directory, seconds later,
succeeded**
  and was renamed back. So the directory was not held by anything
persistent.
* The dry run (`--dry-run`) completes fine and reports the correct row
counts,
  which means the command opens and reads `chroma.sqlite3` before this
point.

Taken together: the handle belongs to the repair process itself.

## Workaround

The refusal for same-path rebuilds names the escape hatch:

```
Source and dest are the same path. Pass archive_existing_dest=True
(CLI: --archive-existing) to move the existing palace aside, or pass a
different source_palace= (CLI: --source).
```

So the archive can be done externally, by a process that is not holding the
palace open:

```powershell
Rename-Item ~\.mempalace  .mempalace.pre-rebuild-<stamp>
$env:MEMPALACE_CONFIG_DIR = "$HOME\.mempalace.pre-rebuild-<stamp>"
mempalace repair --mode from-sqlite --source
"$HOME\.mempalace.pre-rebuild-<stamp>" --yes
```

This completed successfully: 782 rows (742 drawers + 40 closets)
re-embedded,
counts identical to the archive, `integrity_check` clean.

### A trap worth documenting in that workaround

`config.json` lives **inside** the palace directory. Renaming the palace
away
means config resolution falls through to `$XDG_CONFIG_HOME/mempalace` →
`~/.config/mempalace`, finds nothing, and `embedding_model` silently
defaults to
`minilm`. The rebuild then re-embeds the whole palace **with the wrong
model**
and reports complete success.

That failure is invisible afterwards: a palace of internally-consistent but
wrong-model vectors searches perfectly well and simply returns worse answers
forever. Pinning `MEMPALACE_CONFIG_DIR` at the archive and asserting the
resolved `embedding_model` **before** rebuilding is what avoids it.

## Suggested fix

Close the palace handle before attempting the archive rename, or perform the
archive before any database open. A Windows-specific fallback
(copy-then-delete
rather than rename) would also work but is slower and less atomic.

## Praise where it is due

The command **fails closed and says so** — *"The palace was left
untouched."*
That sentence is why retrying was safe to attempt at all, and why eight
failed
attempts cost nothing. Whatever else changes, please keep it.

---

## Adjacent observation, possibly a separate issue

This palace had been silently unsearchable for roughly a month before the
rebuild. Both collections were created with `dimension: 768`, while both
3.7.0
and 3.10.0 define `_EMBEDDINGGEMMA_DIM = 384` (Matryoshka truncation), so
every
query failed with:

```
Search error: Collection expecting embedding with dimension of 768, got 384
```

Writes continued to succeed at 768 the entire time, and all metadata reads
(`status`, `list_drawers`, `get_drawer`) worked normally, so nothing
indicated a
problem unless a search was actually run and its error read.

3.10 surfaces the root cause itself:

```
EmbedderIdentityUnknownWarning: palace collection 'mempalace_drawers' has no
recorded embedder identity; assuming the current model 'embeddinggemma'.
Run `mempalace palace set-embedder --model <name>` to record it.
```

Recording the identity fixes it going forward. The suggestion: consider
making
that warning **fatal for queries** on a collection whose recorded identity
is
absent *and* whose stored dimension disagrees with the configured model. A
dimension mismatch is not recoverable by retrying, and failing loudly at the
first query would have turned a month-long silent outage into a one-line
error
on day one.]

Best regards,
Richard Williamson

PR #2275, opened 2026-08-15
fix(replica): make first identity mint race-safe

### LINKED ISSUE #2274: Concurrent first replica ID creation can fork logstream provenance
## Summary

`get_replica_id()` is not safe when two hub/CLI processes initialize the same fresh palace concurrently. Contenders share `replica.json.tmp` and each unconditionally replaces `replica.json`, so callers can fail or return an ID different from the durable identity.

This breaks the function's core invariant: one stable replica ID must stamp every op authored by that palace. A returned-but-not-persisted ID can fork logstream provenance.

## Reproduction

On current `develop` (`06cb6987f02610784fefbad4b2bd5d026d164ba6`):

1. Start multiple fresh processes against one empty palace.
2. Patch `_mint()` in each process to wait on a shared barrier after the missing-file check and return a distinct valid ID.
3. Call `get_replica_id()` in every process.

A four-process run reliably produces `FileNotFoundError` contenders at the shared `replica.json.tmp`. A deterministic two-contender run also demonstrated a successful caller returning one ID while `replica.json` held the other contender's ID.

The race is at:

```python
if path.exists():
    ...
tmp = path.with_suffix(".json.tmp")
tmp.write_text(...)
os.replace(tmp, path)
return replica_id
```

## Expected behavior

Exactly one first creator should publish the canonical identity without overwriting another creator. Every successful contender should return the ID stored in `replica.json`; the file must never be observable half-written.

## Proposed scope

- Write each candidate to a unique, same-directory temporary file.
- Flush it completely before publication.
- Publish with an atomic no-clobber operation.
- On a lost race, validate and return the winner.
- Preserve existing legacy-ID and corrupt-file fail-loud behavior.
- Add a real multi-process regression plus collision/error cleanup tests.

A hard-link-based no-clobber publication fails safely on filesystems that do not support hard links; a portable fallback would need equivalent serialization and crash safety rather than reverting to overwrite semantics.

fallenmi:
Closes #2274.

## Summary

- publish first-mint candidates from unique private sibling temp files
- use `os.link()` as an atomic no-clobber claim for `replica.json`
- make losing contenders validate and return the persisted winner
- flush and fsync the candidate before publication and best-effort fsync the directory
- fail closed on filesystems without hard-link support, with an actionable operator message and no identity write or unsafe fallback
- clean up temporary files on collisions and publication errors

## Why no direct `O_EXCL` write?

Creating `replica.json` itself with `O_CREAT | O_EXCL` would expose an empty or partially written canonical file to concurrent readers and could leave it corrupt after a crash.

The temp-plus-link path publishes only a complete file and never overwrites an existing identity. The sibling temp guarantees source and destination are on the same filesystem. A filesystem without hard-link support now fails safely and tells the operator to move the palace; it does not publish or rotate an identity.

## Compatibility

The change affects only a missing `replica.json`. Existing 12-hex and 32-hex IDs, valid-file reads, and fail-loud handling for corrupt or foreign-shaped files keep their existing behavior. No dependency or file-format changes are introduced.

Unsupported publication is explicitly recognized for POSIX `EOPNOTSUPP` / `ENOTSUP` and Windows `ERROR_INVALID_FUNCTION` / `ERROR_NOT_SUPPORTED` (1 / 50). Other errors still propagate unchanged, and a race loser still adopts the persisted winner.

## Validation

Rebased onto exact current `develop` `a9f345cc63254eb4dea7abad36963b85c9f8453a`; exact head is `8e359f9cfc26b092e8b5c57dfae7458cfdda6ae5`.

- the four unsupported-filesystem cases are RED on rebased pre-fix commit `586926d1243417a06ec78ceadd6a1c8eac0a53ae`
- all four are GREEN on the exact head, with no `replica.json`, no temporary file, and the actionable message verified
- `tests/test_logstream_sync.py`: 53 passed
- all five logstream test files: 207 passed
- full suite: 4,593 passed, 32 skipped
- Ruff check and format, pre-commit, and `git diff --check` passed

Linux/macOS/Windows repository CI remains the cross-platform merge gate.

## Provenance

Prepared with OpenAI Codex. The GitHub account owner authorized this public submission; no human line-by-line review is claimed.


igorls (changes requested):
The replica-identity race is real and worth fixing, but this publication mechanism needs current-base and cross-platform validation before we merge it.

The new first-mint path depends on `os.link()` for atomic no-clobber publication. That is appropriate on filesystems with hard-link support, but it changes first-use behavior on filesystems that reject hard links: replica creation now fails and logstream initialization cannot proceed. The current PR head has no GitHub checks, so Windows and the supported Python matrix have not exercised this path.

Please rebase onto current `develop`, get the normal Linux/macOS/Windows CI matrix green, and add an explicit regression for unsupported hard-link publication that verifies the failure is non-destructive and gives the operator an actionable explanation. A silent unsafe overwrite fallback would be wrong; the request is to make the compatibility boundary tested and clear. The current merge conflict is in `CHANGELOG.md`.

fallenmi:
The requested compatibility work is now on exact head `8e359f9cfc26b092e8b5c57dfae7458cfdda6ae5`, rebased onto current `develop` `a9f345cc63254eb4dea7abad36963b85c9f8453a`. POSIX `EOPNOTSUPP` / `ENOTSUP` and Windows errors 1 / 50 now fail closed with an actionable message; the tests prove no `replica.json`, no temp file, and no unsafe fallback. All four new cases are RED on the rebased pre-fix commit and GREEN on head; 207 logstream tests and the full 4,593-test suite pass, along with Ruff/format, pre-commit, and diff checks.

The PR body has the exact evidence. Cross-platform repository CI and maintainer re-review remain the final gates. Disclosure: this follow-up and verification were prepared with OpenAI Codex assistance and reviewed before publication.

PR #2277, opened 2026-08-15
fix(release): restore normalized ingestion and stdio shutdown

offendingcommit:
## Summary

MemPalace 3.7 can again ingest normalized transcript sidecars through the public mining path. The previous squash included the normalized modules and MCP callers but omitted the miner integration, so subject routing and generation-safe replacement were unreachable in normal use.

This also fixes an SDK stdio shutdown hang. A server now exits cleanly when its client closes the output pipe, even while the SDK stdin worker remains blocked.

## Design

- Preserve verbatim normalized transcripts and bind each staged generation to its source version, chunk size, normalization recipe, and optional subject-policy fingerprint.
- Verify a complete replacement generation before retiring older drawers, so an interrupted update keeps the previous complete source available.
- Keep generic transcript normalization, per-conversation deduplication, mtime skips, and bounded Chroma upserts intact.
- Isolate the SDK protocol stream from dependency output and terminate only after the protocol channel is already unusable.
- Restore the three operator MCP tools and subject-routing parameters in the reference documentation.

## Validation

- `uv run pytest tests/ -q --ignore=tests/benchmarks --durations=10` - 4326 passed, 31 skipped
- Focused normalized ingestion, routing, refile, and MCP tests - 77 passed
- Focused stdio protection, EOF, and hub tests - 35 passed
- `uv run ruff check mempalace/convo_miner.py mempalace/mcp_server.py`
- `uv run ruff format --check mempalace/convo_miner.py mempalace/mcp_server.py`
- Native CPU image build and full `scripts/docker-smoke.sh` flow
- Linux/amd64 CPU and GPU image builds

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -4,7 +4,7 @@ A Claude Code plugin that gives your AI a persistent memory system. Mine project
 
 ## Prerequisites
 
-- Python 3.9+
+- Python 3.10+
 
 ## Installation
 
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
       "name": "mempalace",
       "source": "./.claude-plugin",
       "description": "AI memory system — mine projects and conversations into a searchable palace. 36 MCP tools, auto-save hooks, guided setup.",
-      "version": "3.7.0",
+      "version": "3.7.0+oc.1",
       "author": {
         "name": "milla-jovovich"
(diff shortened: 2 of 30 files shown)


offendingcommit:
Closing because this release-fix branch belongs to the offendingcommit/mempalace release chain and was opened against the upstream repository by mistake.

Issue #2172, opened 2026-08-06
Live orphaned hook miner can hold writer lease indefinitely; MCP peers need bounded lifecycle recovery


allanbeke-sys:
Related: #1963 tracks the broader single-writer architecture, #1888 the daemon/peer-lock
topology, and #1908 interrupted-mine HNSW hangs. This report is narrower: a still-live
but wedged hook child can retain the current lease forever, while stdio peers accumulate.

## Environment and incident

Observed on 2026-08-06 with one shared Chroma palace, MemPalace 3.6.0 from PyPI,
the Claude plugin at 3.7.0, and multiple stdio MCP peers on Windows 11.

A Claude Stop hook detached python -m mempalace mine for a transcript. The hook parent
exited, but the miner survived as an orphan for more than 12 minutes with about 0.1 seconds
of accumulated CPU. While that live-but-stuck process held mine_palace_lock, every MCP peer
returned error -32001, "Peer MCP writer active; this server is read-only for mutating
tools". Reconnect attempts also encountered an HNSW segment-reader load error. Killing the
orphan released the kernel lock and peers recovered.

Separately, the client starts one mempalace-mcp chain per conversation. Twelve chains
accumulated over about five hours, and a fresh client created seven chains in about six
minutes. The current eight-hour idle default does not contain that pattern.

## Current behavior

mine_palace_lock has a sound cross-platform kernel lock and records a holder PID/command,
but it has no heartbeat, process-start identity, lease token, generation, bounded wait, or
safe live-stale recovery. MCP retries on each mutating call, so dead holders recover after
the kernel releases the handle. A live wedged holder never becomes reclaimable.

hooks_cli._spawn_mine deliberately detaches the raw miner. Its PID-slot timeout only permits
a later hook to reclaim the dedupe slot; it does not terminate the miner or release the
palace writer lock.

The MCP idle watchdog is long for per-conversation stdio processes and measures from request
start without tracking request IDs in flight, so simply lowering the value risks terminating
a long request.

## Requested core contract

Please introduce one Windows-safe WriterLease abstraction with:

- metadata containing holder PID, process-start identity, random lease token, command,
  acquired_at, heartbeat_at, and monotonically increasing generation;
- atomic heartbeat updates that never change the byte position protected by
  msvcrt.locking on Windows;
- bounded, jittered acquire retries for mutating calls and CLI mine;
- final refusal diagnostics naming the holder PID, heartbeat age, and elapsed timeout;
[...]

PR #2282, opened 2026-08-16
fix(mcp): accept a single conversation file as a convos mine source

### LINKED ISSUE #2281: Hub mine rejects a single conversation file, so every hook transcript ingest fails while a hub is running
### Summary

`mempalace_mine` on the MCP/hub server rejects a single conversation file, while the CLI's own contract explicitly accepts one. Whenever a hub is live, every hook transcript ingest fails, because the hooks pass a single `.jsonl` file.

`cli.py` documents the positional argument as:

```
"dir", help="Directory to mine, or one conversation file with --mode convos"
```

but `mcp_server.py` validates directory-only, regardless of mode:

```python
src = os.path.expanduser(source) if source else ""
if not src or not os.path.isdir(src):
    return {"success": False, "error": f"source directory not found: {source!r}"}
```

`cmd_mine` forwards to the hub whenever one is registered and healthy, so the CLI's documented single-file support is unreachable in exactly the configuration most users run.

### Reproduction

Same tool, same mode, against a live hub. Only the source differs:

```
FILE  + convos -> {"success": false, "error": "source directory not found: '.../<session>.jsonl'"}
DIR   + convos -> {"success": true, "mode": "convos", "dry_run": true, "output": "... -> room:technical (9 drawers) ..."}
```

The file is real and non-empty (21,460,814 bytes) and passes the hooks' own `path.is_file()` check immediately before it is submitted.

### Impact

`hooks_cli._ingest_transcript` mines one transcript file:

```python
_spawn_mine([..., "mine", str(path), "--mode", "convos", "--wing", "sessions"])
```

`path` is a file by construction — the function returns early unless `path.is_file()`. So with a hub running, transcript ingest fails on every Stop and every PreCompact. The failure is not surfaced to the user: `hook_precompact` returns `_output({})` on the success path and on this path alike, so a compaction that captured nothing is indistinguishable from one that captured everything. The only evidence is a stderr line in `hook_state/hook.log`.

On this machine the newest mined chunk for a session file predates the current hub by five days, across a period of continuous use.

### Suggested fix

Accept a file when the mode supports one, and keep the directory requirement for the modes that need it:

```python
if not src or not (os.path.isdir(src) or (mode == "convos" and os.path.isfile(src))):
    return {"success": False, "error": f"source not found: {source!r}"}
```

The error string should also stop asserting "directory" for a mode where a file is valid.

Related, but distinct and already tracked: #2024, #1924 and #2172 cover the palace write lock being held by a long-lived MCP server. That blocks the diary checkpoint on the same hook runs. This issue is only about the file-vs-directory precondition, which fails before any lock is taken.

### Environment

- MemPalace 3.7.0, chromadb 1.5.9, macOS, Python 3.12
- Hub: `mempalace.mcp_server --transport http --host 127.0.0.1 --port 8765`

rubicon:
Closes #2281

### What was wrong

`cli.py` documents the mine source as:

```
"dir", help="Directory to mine, or one conversation file with --mode convos"
```

`mcp_server.tool_mine` validated `os.path.isdir(src)` regardless of mode. Since `cmd_mine` forwards to the hub whenever one is registered and healthy, the documented single-file form was unreachable in the configuration most users run.

`hooks_cli._ingest_transcript` submits exactly one `.jsonl` — it returns early unless `path.is_file()` — so with a hub live, every Stop and every PreCompact transcript ingest failed with `source directory not found`.

### Why it went unnoticed

`hook_precompact` returns `_output({})` on the success path and on this path alike. A compaction that captured nothing is indistinguishable from one that captured everything; the only evidence is a stderr line in `hook_state/hook.log`. On the machine where this was found, the newest mined chunk for a session file predated the running hub by five days, across continuous use.

### The change

`convos` accepts a file or a directory. The tree-walking modes keep the directory requirement. The error string no longer asserts "directory" for a mode where a file is valid.

### Tests

Two added to `tests/test_mcp_mine.py`:

- `test_convos_mode_accepts_a_single_file` — mines one file and asserts drawers actually land, rather than asserting the call merely returned success.
- `test_projects_mode_still_rejects_a_file` — pins the other side, so the relaxation cannot widen into "any mode, any path".

Both were confirmed non-vacuous by mutation rather than by inspection:

| Mutation | Result |
|---|---|
| revert to `os.path.isdir(src)` | `test_convos_mode_accepts_a_single_file` fails |
| widen to `isdir(src) or isfile(src)` for every mode | `test_projects_mode_still_rejects_a_file` fails |

Each mutation is killed by the test whose name describes it, and neither test passes on the mutated code it exists to catch.

Full suite: 4302 passed, 31 skipped.

### Not in scope

The palace write lock held by a long-lived MCP server (#2024, #1924, #2172) blocks the diary checkpoint on the same hook runs. That is a separate failure with a separate cause; this precondition fails before any lock is taken.


igorls:
Thanks for this contribution! Enabling single conversation file ingest via MCP matches the CLI behavior and is very helpful for transcript hooks.

There are currently merge conflicts with the latest `develop` branch. Could you please rebase this PR against `develop`? Once resolved and CI passes, we'll merge it in.

rubicon:
Rebased onto develop as requested. The branch is now two commits on top of 98e6dfc with no merge commits.

The only conflict was the changelog entry. It moved into the [Unreleased] section, below the two entries already there. The code change itself is unchanged and merged cleanly.

Local run before pushing: 4525 passed, 32 skipped, with ruff check and ruff format both clean.

Issue #2224, opened 2026-08-11
mining with MEMPALACE_EMBEDDING_DEVICE=coreml doesnt work


daproject85:
**What happened?**

when trying to use coreml and leveraging the fp16 model, the performace is worst than CPU 

Test CoreML on a small controlled mine. FAILED — aborted per plan. Built an isolated 564KB/15-file test corpus in /tmp, mined into a throwaway _coreml_test wing with MEMPALACE_EMBEDDING_DEVICE=coreml. Daemon log: number of nodes in the graph: 1767, number of nodes supported by CoreML: 147 — only 8.3% node coverage, worse than the original quantized-model stall (20%). Job took 2:36 wall-clock at 540% CPU for a corpus that runs in seconds on CPU

**What did you expect?**

**How to reproduce:**

do a test mine with MEMPALACE_EMBEDDING_DEVICE=coreml , CPU STILL spikes

**Environment:**
- OS: mac os 26
- Python version: 3.14
- MemPal version: 3.6.0

Issue #1858, opened 2026-06-23
repair re-embed (minilm to embeddinggemma) runs indefinitely / never completes index build on Apple Silicon CoreML


KelTech-Services:
**Environment:** macOS (Apple Silicon, M1 Ultra), MemPalace 3.4.1, ChromaDB backend, embedding_device: coreml, ~726 drawers.

**What happened:** Switching an existing MiniLM palace to embeddinggemma per the documented path (set embedding_model: embeddinggemma, run mempalace repair) does not complete. Observed across multiple attempts with both repair --mode from-sqlite and plain repair, on a freshly-repaired healthy MiniLM source palace, with CPU otherwise idle:

**Failure mode A - indefinite runaway:**
- The 726 vectors ARE re-embedded and committed to the collection (verified: SELECT COUNT(*) FROM embeddings = 726).
- After the vectors commit, the process continues at ~1400-1750%% CPU indefinitely - in one run, 8h52m wall time / 9000+ min CPU before being killed - with no progress output and no completion message. du on the palace stays flat (~6MB) the entire time.
- Killing the process leaves search non-functional; restoring the pre-repair MiniLM palace recovers it.

**Failure mode B - silent empty rebuild:**
- On other attempts the re-embed exited quickly but wrote 0 drawers (empty ~188KB chroma.sqlite3), silently, with no error. This occurred both with and without other CPU load present, so CPU contention does not appear to be the cause.

**Expected:** repair completes the re-embed + index rebuild and exits, or surfaces an error.

**Workaround:** Recording the existing identity (mempalace palace set-embedder --model minilm) cleared a separate recurring HNSW .drift- quarantine issue and stabilized the MiniLM palace; staying on MiniLM.

Possibly related to the known HNSW-corruption failure class discussed in #1669.

pierrefrns:
[...]

PR #2283, opened 2026-08-16
fix(embedding): never let CoreML return NaN vectors for embeddinggemma

### LINKED ISSUE #2284: embeddinggemma on CoreML silently returns all-zero vectors (hidden state is all-NaN), and embedding_device=auto picks CoreML on Apple Silicon
**What happened?**

On Apple Silicon, `embedding_model=embeddinggemma` running on `CoreMLExecutionProvider` produces unusable vectors with no error raised:

- `last_hidden_state` comes back **entirely NaN**
- the pooled `sentence_embedding` that MemPalace actually stores comes back **all zeros** (L2 norm 0)

ONNX Runtime only warns on stderr — only ~20% of the graph is claimed by CoreML, over ~100 partitions:

```
CoreMLExecutionProvider::GetCapability, number of partitions supported by CoreML: 99
number of nodes in the graph: 1647, number of nodes supported by CoreML: 330
```

This matters because `embedding_device` defaults to `"auto"` and `_AUTO_ORDER` in `embedding.py` puts `CoreMLExecutionProvider` ahead of CPU. So **any Apple Silicon user who selects embeddinggemma and does not set `embedding_device` explicitly gets this silently.** The vectors are still 384-dim, so ChromaDB never complains — queries are compared against zero vectors and search results become meaningless.

The worst case is `repair rebuild-index`: it re-embeds the whole palace, so a single run rewrites every drawer with a zero vector. After the fact a zero vector is indistinguishable from a healthy one, and the only recovery is another full re-embed.

Note for anyone checking their own palace: **look for zero-norm vectors, not for NaN.** In every run I measured, the NaN stays in the hidden state and what reaches the collection is a zero vector.

`minilm` is not affected: chromadb's `ONNXMiniLM_L6_V2` removes `CoreMLExecutionProvider` from the provider list on its own ("it is not as well optimized as CPU"), so embeddinggemma is the only model that reaches CoreML.

**What did you expect?**

Either CoreML computes the model correctly, or MemPalace declines to use it and falls back to CPU with a warning. A silently degenerate vector is the one outcome that cannot be detected later.

**How to reproduce:**

1. On an Apple Silicon Mac with an `onnxruntime` build that exposes `CoreMLExecutionProvider`:

```bash
MEMPALACE_EMBEDDING_MODEL=embeddinggemma MEMPALACE_EMBEDDING_DEVICE=coreml python -c "
import math
from mempalace.embedding import get_embedding_function
v = get_embedding_function()(['texto de prueba'])[0]
print('dim', len(v), 'NaNs', sum(1 for x in v if x != x),
      'zeros', sum(1 for x in v if x == 0.0),
      'norm', math.sqrt(sum(x*x for x in v)))
"
# dim 384 | NaNs 0 | zeros 384 | norm 0.0
```

2. With `MEMPALACE_EMBEDDING_DEVICE=cpu` the same model is fine: norm 1.0, zero NaNs.

3. At the ONNX Runtime level, loading the same `model_quantized.onnx` twice and looking at both graph outputs:

| provider | `last_hidden_state` | `sentence_embedding` |
|---|---|---|
| CoreML | 9216 NaN, norm `nan` | 768 zeros, norm 0.0 |
| CPU | 0 NaN, norm 389.6 | norm 1.0 |

Verified identical under two ONNX Runtime versions and two different graph partitionings (99 partitions / 330 nodes, and 123 / 279), so it does not look partition-dependent.

**Environment:**
- OS: macOS 26.6.1 (Darwin 25.6.0), MacBook Pro M4 Max
- Python version: 3.14 (pipx install) and 3.12.12 (repo venv)
- MemPal version: 3.7.0 (pipx), also reproduced on `develop` @ c445d4f
- onnxruntime: 1.28.0 (pipx env) and 1.24.4 (repo venv) — both affected

---

**Also worth knowing:** even when CoreML does run, it is about half the speed of CPU for this model. Measured over 128 real documents on the M4 Max:

| config | docs/s |
|---|---|
| CPU, ORT default threads | 16.0 |
| CPU, 10 threads | 15.4 |
| CPU, 14 threads | 13.7 |
| CoreML | 7.9 |

**Related:** #2224 reports the same combination (embeddinggemma × CoreML) as a performance problem — 147 of 1767 nodes on the fp16 model — without mentioning the corrupt output; #1858 reports a minilm→embeddinggemma re-embed that never completes on Apple Silicon CoreML. Same underlying partial-support path, different symptoms.

I have opened #2283 with a fix: keep `auto` from selecting CoreML for this model, and make any non-CPU provider prove it computes (witness embedding, finite and non-zero norm) before it embeds anything, falling back to CPU with a warning otherwise.

marcoaperez:
Fixes #2284.

## The bug

On Apple Silicon, `embedding_model=embeddinggemma` with the default `embedding_device=auto` silently produces degenerate vectors. ONNX Runtime warns on stderr but does not fail:

```
CoreMLExecutionProvider::GetCapability, number of partitions supported by CoreML: 123
number of nodes in the graph: 1647, number of nodes supported by CoreML: 279
```

Same model, same input, two providers:

| provider | `last_hidden_state` | `sentence_embedding` |
|---|---|---|
| CoreML | 9216 NaN, norm `nan` | 768 zeros, norm 0.0 |
| CPU | 0 NaN, norm 389.6 | norm 1.0 |

The corruption is in the hidden state; the pooled vector is the consequence. What reaches the collection is a **zero** vector — verified identical under onnxruntime 1.24.4 and 1.28.0, and under two different graph partitionings (99 partitions / 330 nodes and 123 / 279), so it does not appear to be partition-dependent. The original report described NaN vectors, which is what the raw hidden state shows.

Either way the check has to require a **finite and non-zero** norm: an all-zero vector sails through any `isnan()` test and is exactly as unusable. It also means anyone auditing an existing palace should look for zero-norm vectors, not for NaN.

`embedding_device` defaults to `"auto"` and `_AUTO_ORDER` puts CoreML ahead of CPU, so every Apple Silicon user on this model was affected without opting into anything. If it happens during `repair rebuild-index`, the whole palace is rewritten with unusable vectors — and a NaN vector is indistinguishable from a healthy one once stored.

Reproduced on a MacBook Pro M4 Max, macOS Darwin 25.6.0, onnxruntime 1.28.0, mempalace 3.7.1.

## The fix — two layers

**Selection** — `_AUTO_PROVIDER_DENYLIST` keeps `auto` from ever picking CoreML for embeddinggemma. CUDA/DirectML are untouched, and so are other models: chromadb's `ONNXMiniLM_L6_V2` already strips CoreML on its own ("not as well optimized as CPU"), so minilm never reached it. CoreML is also ~2x slower here when it does run — measured over 128 real documents:

| config | docs/s |
|---|---|
| CPU, ORT default threads | 16.0 |
| CPU, 10 threads | 15.4 |
| CPU, 14 threads | 13.7 |
| CoreML | 7.9 |

**Validation** — a witness embedding at load time makes any non-CPU provider prove it computes (finite, non-zero norm) before anything real is embedded. An explicit `embedding_device=coreml` is still honoured, but falls back to CPU with a warning instead of returning garbage. If CPU is degenerate too, it raises rather than storing unsearchable vectors.

The probe lives in `EmbeddinggemmaONNX._lazy_load`, not right after the EF is constructed: validating in `get_embedding_function` would force the ~300 MB model load on every call site that only builds an EF without embedding (status, `describe_device`), defeating the lazy load. Inside `_lazy_load` the model is already resident and the cost is one short forward pass, once per process. CPU-only sessions skip the probe entirely — CPU is the fallback, so checking it could only add a forward pass to every cold start.

`describe_device` is model-aware for the same reason: the miner header would otherwise announce a provider we will not use.

## Verification

With the real model, all three devices now return norm 1.0 and zero NaN; `auto` resolves to CPU without even building a CoreML session, and an explicit `coreml` warns and falls back:

```
WARNING Embedding provider CoreMLExecutionProvider returned a degenerate vector (NaN or all-zero)
        for EmbeddingGemma — falling back to CPUExecutionProvider.
```

11 new tests, including a parametrized regression over auto/coreml/cpu against a runtime that advertises CoreML and poisons any accelerator session. Full suite on top of `develop` @ 639c69a: **4312 passed, 31 skipped**. `ruff check` and `ruff format --check` clean.

One unrelated test, `test_mcp_server.py::TestWriteTools::test_checkpoint_added_by_accepted_via_dispatch`, fails on my machine with `Peer MCP writer active` — it fails identically on the base commit **without** this change (a MemPalace MCP writer is live on that host), and passes both in isolation and in CI.

## Not included

- `describe_device("coreml")` still reports `coreml` even though the EF ends up on CPU. The label is computed without loading the model; the load-time warning tells the truth. Changing it would mean loading 300 MB to render a header.
- With minilm + auto the label says `coreml` while chromadb runs on CPU internally — a pre-existing cosmetic mismatch, untouched here.

## Related issues

Neither is claimed to be closed by this PR — both are the same combination (embeddinggemma × CoreML on Apple Silicon) reported through different symptoms:

- #2224 — `MEMPALACE_EMBEDDING_DEVICE=coreml` mining, 8.3% node coverage on the fp16 model and slower than CPU. Same partial-support path; that report does not mention NaN.
- #1858 — minilm→embeddinggemma re-embed that never completes on Apple Silicon CoreML.

I found no open issue reporting the NaN/zero vectors themselves.

### DIFF
--- a/mempalace/embedding.py
+++ b/mempalace/embedding.py
@@ -39,6 +39,10 @@ in ``~/.mempalace/config.json``):
 
 Requesting an unavailable accelerator emits a warning and falls back to CPU
 rather than hard-failing — mining must still work on a laptop without CUDA.
+The same applies to an accelerator that runs but computes the model wrongly:
+``embeddinggemma`` on CoreML returns NaN or all-zero vectors without raising,
+so ``auto`` never selects CoreML for it and an explicitly requested one is
+rejected by a witness embedding at load time.
 """
 
 from __future__ import annotations
@@ -72,6 +76,24 @@ _AUTO_ORDER = [
     ("DmlExecutionProvider", "dml"),
 ]
 
+# Providers that must never be picked *automatically* for a given model,
+# keyed by model name.
+#
+# embeddinggemma × CoreML: CoreML claims only ~280 of the quantized graph's
+# 1647 nodes, split across 100+ partitions, and the result is corrupt —
+# last_hidden_state comes back all-NaN, so the pooled sentence_embedding is
+# NaN or all-zero (which one depends on how that run partitioned) — with no
+# error raised. Since embedding_device defaults to "auto" and CoreML sits
+# ahead of CPU, every Apple Silicon user running this model would otherwise
+# embed degenerate vectors silently; a `repair rebuild-index` would write them
+# over the whole palace. CoreML is also ~2x slower here when it does run
+# (7.9 vs 16.0 docs/s on an M4 Max), so nothing is lost by skipping it.
+# An explicit embedding_device=coreml is still honoured — the witness probe in
+# EmbeddinggemmaONNX._lazy_load catches it and falls back to CPU.
+_AUTO_PROVIDER_DENYLIST = {
+    "embeddinggemma": {"CoreMLExecutionProvider"},
+}
+
 _EF_CACHE: dict = {}
 # Check-then-construct on the cache must be atomic: without it, two threads
 # resolving the same key each keep their own EF instance, and each instance
@@ -80,13 +102,18 @@ _EF_CACHE_LOCK = threading.Lock()
 _WARNED: set = set()
 
 
-def _resolve_providers(device: str) -> tuple[list, str]:
+def _resolve_providers(device: str, model: Optional[str] = None) -> tuple[list, str]:
     """Return ``(provider_list, effective_device)`` for ``device``.
 
     Falls back to CPU (with a one-shot warning) when the requested
     accelerator is not compiled into the installed ``onnxruntime``.
+
+    ``model`` gates ``_AUTO_PROVIDER_DENYLIST``: a provider known to produce
+    wrong results for that model is skipped under ``auto``. Explicit device
+    requests are left alone — a user who names an accelerator gets it.
     """
     device = (device or "auto").strip().lower()
+    denied = _AUTO_PROVIDER_DENYLIST.get((model or "").strip().lower(), frozenset())
 
     try:
         import onnxruntime as ort
@@ -97,7 +124,7 @@ def _resolve_providers(device: str) -> tuple[list, str]:
 
     if device == "auto":
         for provider, name in _AUTO_ORDER:
-            if provider in available:
+            if provider in available and provider not in denied:
                 return ([provider, "CPUExecutionProvider"], name)
         return (["CPUExecutionProvider"], "cpu")
 
@@ -235,6 +262,9 @@ _EMBEDDINGGEMMA_MAX_LEN = 2048
 # a sub-batch by size rather than by arrival order (#2104), because the run
 # is priced on that padded length and not on the document count.
 _EMBEDDINGGEMMA_BATCH_SIZE = 32
+# Short document embedded once per model load to prove the execution provider
+# actually computes (see _embeddinggemma_session_is_healthy).
+_EMBEDDINGGEMMA_WITNESS = "mempalace embedding provider health check"
 
 
 def _sanitize_embeddinggemma_input_ids(tokenizer, input_ids, np):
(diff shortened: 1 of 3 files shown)


igorls (approved):
Verified locally on Apple Silicon against current `develop` (merged in cleanly; full suite 4427 passed, 31 skipped, ruff clean).

Approving. Two things make this the right shape rather than just a working patch:

The guard requires a **finite and non-zero** norm rather than testing for NaN. An all-zero vector passes every `isnan()` check and is exactly as unusable, so a NaN-only test would have looked like a fix while leaving the case you actually observed unhandled. Worth repeating your other point too: anyone auditing an existing palace should be grepping for zero-norm vectors, not NaN.

And it is two layers rather than one — the denylist keeps `auto` from selecting CoreML for this model at all, with the norm check as the backstop if it is reached another way. Selection alone would have been silently bypassable; validation alone would have meant a fallback on every call.

The severity justifies including it in the next release: `embedding_device` defaults to `auto`, `_AUTO_ORDER` puts CoreML first, so every Apple Silicon user on embeddinggemma was affected without opting into anything — and a `repair rebuild-index` under those conditions rewrites the whole palace with unusable vectors that are indistinguishable from healthy ones once stored. That is silent data loss on a default path.

Thanks for the provider-by-provider table and for pinning it across two onnxruntime versions and two graph partitionings — that is what makes it diagnosable rather than anecdotal.

PR #2287, opened 2026-08-16
fix(mcp): default diary agent_name; report bad args as -32602 not -32000

alanFWFG:
## The bug

`mempalace_diary_write` and `mempalace_diary_read` failed with `MCP error -32000: Internal tool error` on **every** call, for months.

Both declared `agent_name` as `required`. The dispatcher calls `handler(**tool_args)`, so omitting it raised `TypeError` *before the handler body ran*, and the blanket `except Exception` flattened it into an opaque `-32000`. `logger.exception` writes the traceback to **stderr**, which MCP hosts discard — so the client saw a bare error code with no way to diagnose it.

Reproduced in-process against the module the server actually runs:

| call | result |
|---|---|
| `tool_diary_read(agent_name=...)` direct | success |
| dispatch **with** `agent_name` | success |
| dispatch **without** `agent_name` | `TypeError` → `-32000` |

This explains every symptom that had been recorded against it: payload size never mattered (190 B and 4 KB both failed — binding precedes the body), encoding never mattered, `diary_read` failed too (the only other tool whose sole required arg is `agent_name`), and `add_drawer` / `kg_add` / `search` / `status` never failed because none of them require it. Months of speculation about payload size and unicode encoding; neither was ever involved.

## Changes

1. **`agent_name` defaults** to `MEMPALACE_MODEL`, else `MEMPALACE_HARNESS`, else `"unknown"`. Only `adapters/openai.py` sets `MEMPALACE_MODEL`, so under other harnesses the harness name is the practical default. Both schemas drop `agent_name` from `required` so the declared contract matches the code.

2. **Legible argument errors, for every tool.** `handle_request` validates binding with `inspect.Signature.bind` before dispatch and returns `-32602` with the real message. Binding is checked up front rather than by wrapping the call in `except TypeError` — the latter would misreport a genuine `TypeError` raised *inside* a handler body as a caller argument error. A test pins that distinction.

3. **`diary_read` now derives its wing via `canonicalize_agent`**, matching `diary_write`. It previously did its own `lower().replace(' ', '_')`, so `"Claude Opus 5"` wrote to `wing_claude-opus-5` and read from `wing_claude_opus_5`. Latent while neither side had a default — but once both default to the same env string, read and write must agree, so this is required for (1) to be correct. (`normalize()` collapses only `[\s_]+`, so dotted slugs like `claude-opus-4.8` round-trip fine; only spaces and underscores diverged.)

## Verification

- Full suite: **985 passed, 1 skipped**.
- 6 new tests. **5 of 6 fail against the old source**; the 6th (`test_handler_internal_typeerror_still_reports_internal_error`) passes both ways by design — it is the guard that `-32602` did not widen to swallow real handler bugs.
- `ruff check` clean.
- `ruff format --check` still flags both files. This is **pre-existing on `main`** (verified by stashing) and deliberately not addressed here, to avoid burying a 53-line fix in an unrelated whitespace diff.

## Note for anyone running this

A live MCP server keeps the old module loaded — `-32000` persists until the server restarts. Until then, pass `agent_name` explicitly.

### DIFF
--- a/FWFG.md
+++ b/FWFG.md
@@ -1,0 +1,43 @@
+# FWFG fork — operating notes
+
+This is the FWFG-owned fork of the public `mempalace/mempalace`, carrying the
+multi-model provenance / wing-registry / bootstrap / adapter work. Upstream
+tracking is kept clean (see `docs`/the design spec); this file is FWFG-only and
+is **not** meant to go upstream.
+
+## Which MemPalace Is Which
+
+There are two related repos and (historically) more than one Python install. Keep
+them straight:
+
+| Thing | What it is |
+|---|---|
+| **`mempalace`** (this repo) | The **runtime package fork** — MCP server, provenance stamping, wing registry/canonicalization, `mempalace_bootstrap`, the OpenAI adapter, local migration/backfill. |
+| **`mempalace-bq`** | The **BigQuery sync/mirror** companion — DDL, the push/pull sync, forced-resync. It is *not* a MemPalace package; it never becomes the fork. |
+
+### Rules that prevent old/new confusion
+
+- **Never use bare `python -m mempalace`.** With multiple interpreters on a machine
+  (3.12, 3.14, uv, Store stub…), bare `python` is ambiguous and may hit a stale
+  global install or one with no `mempalace` at all.
+- **Every configured host points at the explicit fork interpreter:**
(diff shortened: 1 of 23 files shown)


alanFWFG:
Closing — opened against the wrong base repository by mistake. This is an internal fix intended for our own fork (findwhatfeelsgood/mempalace), not a contribution to upstream. Apologies for the noise.

PR #2288, opened 2026-08-16
fix(graph): report complete palace graph stats

### LINKED ISSUE #2260: graph_stats reports an empty palace graph: build_graph sees a strict subset of the collection, and explicit tunnels are never counted
# `graph_stats` reports an empty palace graph: missing wings, and explicit tunnels are never counted

**Version:** mempalace 3.5.0 · chromadb 1.5.9 · Python 3.12.7 · Ubuntu 22.04.5 LTS
**Affected:** `mempalace.palace_graph.graph_stats()` and the `mempalace_graph_stats` MCP tool

## Summary

On a palace with 5 wings and 9 rooms, `graph_stats()` reports 6 rooms, omits two wings
entirely, and returns `tunnel_rooms: 0, total_edges: 0` — while 44 explicit tunnels exist
in `tunnels.json` and the room `general` demonstrably appears in three different wings.

The practical effect is that a healthy, well-connected palace looks empty through this
API. We were evaluating `graph_stats` as the summary endpoint for a cross-project
dashboard and would have reported "no graph" to users.

This is **not** a stale cache — a cold call in a fresh Python process returns identical
numbers.

## Environment / palace shape

`mempalace status` (ground truth):

```
WING: desercion            ROOM: fact                        1 drawers
WING: desercion-pascual    ROOM: general                   706 drawers
WING: desertion            ROOM: general                  2646 drawers
                           ROOM: heatstgnn-model-selection    2 drawers
                           ROOM: diary                       1 drawers
WING: matlab-drive         ROOM: general                    38 drawers
WING: octopus              ROOM: documentation              87 drawers
                           ROOM: plans                      68 drawers
                           ROOM: controller                  3 drawers
```

5 wings, 9 rooms. Note `general` appears in **three** wings.

## Reproduction

```python
from mempalace.palace_graph import graph_stats, list_tunnels
print(graph_stats())
print("explicit tunnels:", len(list_tunnels()))
```

## Expected

- `rooms_per_wing` covering all 5 wings
- `total_rooms` = 9
- `tunnel_rooms` >= 1 (`general` spans 3 wings)
- the 44 explicit tunnels reflected somewhere in a function documented as
  *"Palace graph overview: total rooms, tunnel connections, edges between wings"*

## Actual

```python
{
  'total_rooms': 6,
  'tunnel_rooms': 0,
  'total_edges': 0,
  'rooms_per_wing': {'octopus': 3, 'desertion': 2, 'desercion': 1},
  'top_tunnels': []
}
explicit tunnels: 44
```

- Wings **`matlab-drive`** and **`desercion-pascual`** are absent entirely.
- `desertion` reports 2 rooms; it has 3.
- `tunnel_rooms`/`total_edges` are 0 despite `general` spanning three wings.

### 1. `build_graph` silently sees a strict subset of the collection

`graph_stats` derives everything from `build_graph()`, which paginates ChromaDB metadata:

```python
total = col.count()
offset = 0
while offset < total:
    batch = col.get(limit=1000, offset=offset, include=["metadatas"])
```

We ran that identical pagination by hand against **the same collection object**, in the
same process, immediately before calling `build_graph(col)`:

```
collection.count() = 3552
records actually paged: 3552      (metadata without a wing: 0)
RAW wings in metadata: {'desertion': 2649, 'desercion': 1, 'matlab-drive': 38,
                        'octopus': 158, 'desercion-pascual': 706}
RAW distinct (wing, room) pairs: 9        <- matches `mempalace status` exactly

build_graph nodes: 6   edges: 0
build_graph wings: ['desercion', 'desertion', 'octopus']
rooms spanning >=2 wings per build_graph: none
```

So the data is complete and every record carries wing metadata — the raw loop recovers all
5 wings and all 9 rooms. `build_graph`, given the same collection, returns 3 wings and
6 rooms. The wings it drops (`matlab-drive` 38 + `desercion-pascual` 706 = 744 records)
account exactly for the shortfall: 3552 − 744 = 2808, the number of records reflected in
its output.

This is a defect in `build_graph` itself, not in the store, not a cache (a cold call in a
fresh process is identical), and not concurrency (the raw read was complete under the same
conditions).

Because the wings containing the room `general` are among those dropped, the passive-tunnel
test (`len(node["wings"]) >= 2`) can never fire — which is why both tunnel metrics are 0
rather than merely undercounted.

### 2. Explicit tunnels are structurally invisible to `graph_stats`

`palace_graph.py` documents the distinction clearly:

> Passive tunnels are discovered from shared room names across wings.
> Explicit tunnels are created by agents when they notice a connection
> between two specific drawers or rooms in different wings/projects.

`graph_stats` only ever computes the passive kind. It never reads `tunnels.json`, so the
44 explicit tunnels created by the miner (`Entity tunnels: +N cross-wing entity link(s)`)
do not appear in the graph overview at all.

This may be intentional internally, but from the outside it is surprising: the tool
description promises a *palace graph overview*, and a caller has no way to know the number
excludes the tunnels the miner just reported creating.

## Suggested fix

1. Make `build_graph` cover every record `col.get()` will page through — the reproduction
   above shows a hand-rolled loop over the same collection recovering all 5 wings while
   `build_graph` recovers 3. At minimum, have it report skipped/unparsable records so
   silent omission is impossible.
2. Either include explicit tunnels in `graph_stats`, or add explicit counts
   (`explicit_tunnels`, `passive_tunnel_rooms`) so the two kinds are distinguishable and
   neither is silently dropped.
3. Consider a consistency check between `mempalace status` and `graph_stats` — they
   currently disagree about how many wings and rooms exist, which is what made this
   noticeable.

## Not affected (checked)

The knowledge-graph layer (`mempalace.knowledge_graph.KnowledgeGraph`,
`knowledge_graph.sqlite3`) is a separate store and behaved correctly in isolated testing,
including temporal validity:

```python
kg.add_triple("project:x", "canonical_root", "/old", valid_from="2026-07-01", valid_to="2026-08-14")
kg.add_triple("project:x", "canonical_root", "/new", valid_from="2026-08-14")
kg.query_entity("project:x", as_of="2026-07-15")  # -> /old   correct
kg.query_entity("project:x", as_of="2026-08-20")  # -> /new   correct
```

`list_tunnels()` also returns correct, complete data. The issue appears confined to
`build_graph`/`graph_stats`.

## Impact for us

We are building a machine-wide project graph on top of MemPalace and have routed around
this by using `list_tunnels()` and the KG API instead of `graph_stats`. Flagging it because
the failure mode is silent: a healthy palace reports an empty graph, with no error.

Happy to run any diagnostic against this palace — it reproduces reliably.

MichaelWave369:
Closes #2260.

`build_graph()` and the newer grouped-SQL graph fast paths were discarding every drawer/group whose room was named `general`. That makes wings represented only by `general` disappear from graph reporting and under-counts mixed wings.

This rework follows the maintainer guidance on the current read-path architecture and applies the same room-preservation rule in all three places:

- `mempalace/palace_graph.py` — client-paging `build_graph()` path;
- `mempalace/palace_graph.py` — grouped `_nodes_edges_from_grouped_rows()` path used by sqlite-backed graph reads;
- `mempalace/mcp_server.py` — grouped `_graph_stats_from_grouped_rows()` MCP fast path.

The reporting contract stays backwards-compatible while making placement counts explicit:

- include `general` as a real room node instead of discarding it;
- keep `total_rooms` as the historical count of unique room-name nodes;
- add `total_room_instances` for distinct `(wing, room)` placements;
- count grouped-SQL room instances with an actual distinct `(wing, room)` set, so the same placement appearing under multiple halls is not over-counted;
- keep `tunnel_rooms` and `total_edges` semantics intact;
- add `passive_tunnel_rooms` as the clearer alias;
- report `explicit_tunnels` from the existing tunnel sidecar;
- add `total_connections = total_edges + explicit_tunnels`.

The sqlite fast-path contract is preserved: existing tripwire coverage still fails if graph stats open the collection/HNSW path. Regression tests cover both the client path and grouped path, including a repeated `(wing, room)` across different halls.

Validation of the four feature/test files:

- `python -m pytest -q tests/test_palace_graph.py tests/test_palace_graph_tunnels.py` — **82 passed**
- `python -m pytest -q tests/test_mcp_server.py -k 'graph_stats or tunnel or 2288'` — **7 passed, 367 deselected**
- `ruff check mempalace/palace_graph.py mempalace/mcp_server.py tests/test_palace_graph.py tests/test_mcp_server.py` — **passed**
- `ruff format --check ...` — **4 files already formatted**
- `python -m compileall -q ...` — **passed**

Those tests ran on `develop` at `fe8e748`. While validation was running, `develop` advanced to `3e56979` for the 3.8.0 release. The intervening commits modify only release/version/workflow/docs files and do not touch any of the four files in this PR, so the validated blobs were transplanted unchanged onto `3e56979`.

Final branch shape:

- one clean commit on current `develop`;
- 1 ahead / 0 behind at rebase time;
- only four changed files:
  - `mempalace/palace_graph.py`
  - `mempalace/mcp_server.py`
  - `tests/test_palace_graph.py`
  - `tests/test_mcp_server.py`

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -94,6 +94,7 @@ from .palace_graph import (  # noqa: E402
     list_tunnels,
     delete_tunnel,
     follow_tunnels,
+    _load_tunnels as _load_graph_tunnels,
 )
 from .hallways import (  # noqa: E402
     list_hallways,
@@ -1980,7 +1981,7 @@ def _sqlite_graph_stats():
     (non-chroma backend, missing/unbootstrapped palace, sqlite error). The
     reconstruction mirrors ``palace_graph.build_graph`` /
     ``palace_graph.graph_stats`` exactly: a node is a room with a non-empty
-    wing and a usable room name (the catch-all ``"general"`` is excluded), and
+    wing and a usable room name (including the catch-all ``"general"``), and
     edges are the per-hall cross-wing crossings of multi-wing rooms.
     """
     rows = None
@@ -2006,25 +2007,30 @@ def _sqlite_graph_stats():
 
 
 def _graph_stats_from_grouped_rows(rows):
-    """Rebuild ``graph_stats`` from ``(room, wing, hall, n)`` grouped rows.
+    """Rebuild ``graph_stats`` from grouped sqlite metadata rows.
 
-    Backends may append a fifth ``last_date`` column for ``find_tunnels``;
-    stats do not use it, so extra columns are ignored rather than unpacked.
+    Rows are ``(room, wing, hall, n)`` with an optional fifth ``last_date``
+    column. Because grouping includes ``hall``, one room placement can occupy
+    multiple SQL rows; room instances therefore use a distinct ``(wing, room)`` set.
     """
     from collections import Counter, defaultdict
 
     room_data = defaultdict(lambda: {"wings": set(), "halls": set(), "count": 0})
+    room_instances = set()
     for row in rows:
         room, wing, hall, n = row[0], row[1], row[2], row[3]
-        if not room or room == "general" or not wing:
+        if not room or not wing:
             continue
-        node = room_data[room]
-        node["wings"].add(str(wing))
+        room_key = str(room)
+        wing_key = str(wing)
+        room_instances.add((wing_key, room_key))
+        node = room_data[room_key]
+        node["wings"].add(wing_key)
         if hall:
             node["halls"].add(str(hall))
         node["count"] += int(n)
 
-    tunnel_rooms = 0
+    passive_tunnel_rooms = 0
     total_edges = 0
     wing_counts = Counter()
     for data in room_data.values():
@@ -2032,20 +2038,25 @@ def _graph_stats_from_grouped_rows(rows):
         for wing in data["wings"]:
             wing_counts[wing] += 1
         if n_wings >= 2:
-            tunnel_rooms += 1
+            passive_tunnel_rooms += 1
             total_edges += (n_wings * (n_wings - 1) // 2) * len(data["halls"])
 
     top_tunnels = [
         {"room": room, "wings": sorted(data["wings"]), "count": data["count"]}
-        for room, data in sorted(room_data.items(), key=lambda kv: (-len(kv[1]["wings"]), kv[0]))[
-            :10
-        ]
+        for room, data in sorted(
+            room_data.items(), key=lambda item: (-len(item[1]["wings"]), item[0])
+        )[:10]
         if len(data["wings"]) >= 2
     ]
+    explicit_tunnel_count = len(_load_graph_tunnels(_config))
     return {
         "total_rooms": len(room_data),
-        "tunnel_rooms": tunnel_rooms,
+        "total_room_instances": len(room_instances),
+        "tunnel_rooms": passive_tunnel_rooms,
+        "passive_tunnel_rooms": passive_tunnel_rooms,
+        "explicit_tunnels": explicit_tunnel_count,
         "total_edges": total_edges,
+        "total_connections": total_edges + explicit_tunnel_count,
         "rooms_per_wing": dict(wing_counts.most_common()),
         "top_tunnels": top_tunnels,
     }
--- a/mempalace/palace_graph.py
+++ b/mempalace/palace_graph.py
@@ -135,7 +135,7 @@ def _nodes_edges_from_grouped_rows(rows):
     for row in rows:
         room, wing, hall, n = row[0], row[1], row[2], row[3]
         last_date = row[4] if len(row) > 4 else ""
-        if not room or room == "general" or not wing:
+        if not room or not wing:
             continue
         node = room_data[str(room)]
         node["wings"].add(str(wing))
@@ -252,7 +252,7 @@ def build_graph(col=None, config=None):
             wing = meta.get("wing", "")
             hall = meta.get("hall", "")
             date = meta.get("date", "")
-            if room and room != "general" and wing:
+            if room and wing:
                 room_data[room]["wings"].add(wing)
                 if hall:
                     room_data[room]["halls"].add(hall)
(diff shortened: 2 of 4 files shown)


igorls:
Heads-up: `develop` moved underneath this one since you validated against `639c69a`, and it changes what the patch needs to touch. That is on our side, not yours — the read-path work landed after you opened this.

The diagnosis is still right. `general` being discarded is exactly why whole wings vanish and rooms under-report, and separating `total_rooms` from `total_room_instances` rather than redefining the existing field is the right call for compatibility.

What changed is that the `general` filter is no longer in one place. It is now in three:

(item 1) `mempalace/palace_graph.py:255` — inside `build_graph()`'s client-paging loop, phrased `if room and room != "general" and wing:`. This is the one your patch fixes.

(item 2) `mempalace/palace_graph.py:138` — `_nodes_edges_from_grouped_rows()`, which builds nodes and edges from a grouped SQL read instead of paging the collection.

(item 3) `mempalace/mcp_server.py:2019` — `_graph_stats_from_grouped_rows()`, the same reconstruction for `graph_stats`.

The two new ones are fast paths added so `graph_stats`, `find_tunnels` and `traverse` stop cold-loading the vector index on large palaces, and **MCP now takes them by default** on both the chroma and sqlite_exact backends. The client-paging loop only runs when the sqlite read cannot serve. So as the branch stands, the fix would be correct but invisible to essentially every real caller — the tests you added exercise the client path, which is why they pass.

For this to take effect it needs the same treatment in all three, and the grouped-rows paths need a little more than deleting the condition: they receive `(room, wing, hall, n, last_date)` tuples rather than per-drawer metadata, so `total_room_instances` comes from counting distinct `(wing, room)` groups rather than from a per-drawer loop. The row shape is documented on `sqlite_room_wing_hall_counts` in both backends.

Worth knowing while you rework it: the sqlite paths are covered by tripwire tests that assert the collection is never opened, so if you accidentally make one of them fall back to client paging the suite will tell you rather than silently getting slower.

No rush on this, and it will not hold up the next release — I would rather it land correct than fast. If you would prefer to hand it back, say so and I will finish it with credit to you; the `total_room_instances` distinction is the part I would not have thought of.

MichaelWave369:
Thanks for the map — reworked on the current read-path architecture.

The fix now covers all three places you called out: client `build_graph()`, grouped `_nodes_edges_from_grouped_rows()`, and MCP `_graph_stats_from_grouped_rows()`. The grouped stats path uses a distinct `(wing, room)` set, so multiple hall rows for the same placement do not inflate `total_room_instances`.

I kept the sqlite/HNSW tripwire path intact. Focused validation is green:
- graph suites: 82 passed
- MCP graph/tunnel fast-path selection: 7 passed, 367 deselected
- Ruff check/format and compileall: passed

While that was running, `develop` advanced to `3e56979` for 3.8.0. That delta does not touch any of the four files in this PR, so I transplanted the validated blobs unchanged onto the new base. Final branch is one clean commit, 1 ahead / 0 behind, four files only.

Appreciate the precise guidance — especially the grouped-row/HNSW details.

PR #2305, opened 2026-08-19
mcp: release idle writer lease so peer sessions can write

jorgeccastro:
## Problem

The per-palace MCP writer lease (#1818) is acquired on the first mutating call and held for the life of the process. A stdio server whose client died without delivering EOF (client machine reboot; the SSH transport kept the connection half-open) keeps the lease indefinitely, so every peer session gets `-32001 Peer MCP writer active` for hours; the idle-exit watchdog (#1552) only reaps it after 8 h by default. Observed twice in one day on a multi-session deployment sharing one palace over SSH stdio.

## Change

- A new watchdog releases the writer lease after `MEMPALACE_MCP_WRITER_IDLE_MINUTES` (default 10, `0` disables) without a completed mutating call. `_acquire_mcp_writer_lock` already re-attempts the non-blocking flock on every mutating call, so the next write from any process (this one included) re-acquires transparently — no restart, no reconnect.
- An in-flight counter over chroma-touching tool calls (everything except the logstream set) guarantees the release never closes storage handles under a running request. Only completed mutating calls refresh the write-idle clock; lease promotion also refreshes it, so a lease cannot be dropped between preflight and dispatch.
- Acquire/release now run under one shared RLock, since release can now fire from the watchdog thread.
- The `-32001` refusal message now includes the holder identity (`data.reason` already carried it; most MCP clients only surface `message`).

## Testing

- 8 new unit tests (`tests/test_mcp_writer_idle_release.py`): release decision, in-flight guard, clock semantics, env parsing.
- Existing `test_palace_locks.py`, `test_vector_write_gate.py`, `test_mcp_http_transport.py` pass (84 total).
- Live integration on a scratch palace: two real stdio servers with a 3 s threshold — A writes and holds; B is refused while A holds (negative control); after A idles, B writes fine; A is refused while B holds; after B idles, A re-acquires. 5/5.

### DIFF
--- a/deploy/server.env.example
+++ b/deploy/server.env.example
@@ -30,6 +30,12 @@ MEMPALACE_EMBEDDING_DEVICE=auto
 # /healthz traffic does not count as activity; only MCP requests reset the timer.
 MEMPALACE_MCP_IDLE_HOURS=0
 
+# A separate watchdog releases an MCP writer lease after this many minutes
+# without a completed mutating call (default 10), while leaving the server
+# running. Set 0 only when this process is the intentional sole writer and all
+# writes are routed through it.
+# MEMPALACE_MCP_WRITER_IDLE_MINUTES=10
+
 # --- Palace location (systemd / bare-metal) -------------------------------------
 # In Docker the palace lives on the mempalace-data volume by default.
 # MEMPALACE_PALACE_PATH=/var/lib/mempalace/palace
--- a/docs/write-routing-policy.md
+++ b/docs/write-routing-policy.md
@@ -182,6 +182,18 @@ that topology and the client uses its local path, where the writer lease still
 applies. Configure those clients to use the HTTP server explicitly, or run the
 server and local forwarders under the same account.
 
+Hook diary checkpoints still write directly on a server machine. `mempalace
+mine` forwards a hook-spawned transcript ingest to a live server, but the
+session-end checkpoint writes ChromaDB in process through
+`_save_diary_direct`, and the hook entry points read no server registry. Every
+session end therefore opens the server's palace from a second process. That is
+the direct hook write the routing rules above tell you not to run beside a
+writable owner. The palace lock does not observe it, because the checkpoint
+takes no writer lease. The contention appears at the SQLite layer, where a
+checkpoint that wedges mid-write holds the write lock that the server's next
+request needs. Until the checkpoint forwards the way the mine already does,
+treat a session end on a server machine as a second writer.
+
 For a filesystem-level backup of a palace owned by an always-on service, stop
 the service, capture the complete palace, then restart the service and verify
 it. Do not make the backup conditional on `pgrep` finding no process under the
--- a/docs/write-routing-policy.md
+++ b/docs/write-routing-policy.md
@@ -117,11 +117,13 @@ enough because each long-lived process can retain SQLite/WAL, FTS, or vector
 index state between calls.
 
 - A writable daemon owns the palace writer lease for its full lifetime.
-- Writable MCP HTTP acquires that lease before binding, holds it through the
-  full serving lifetime, and releases it after active requests stop.
+- Writable MCP HTTP acquires that lease before binding. It releases the lease
+  after a configurable write-idle period and reacquires it on the next
+  mutating request.
 - MCP stdio opens `sqlite_exact` read-only until it acquires the writer lease.
   It may therefore coexist for reads; mutating tools refuse while another
-  process owns the lease and reopen writable storage after that owner exits.
+  process owns the lease. A server that owns the lease releases it after the
+  same write-idle period and reacquires it when it next needs to write.
 - Read-only MCP HTTP may coexist with the writer.
 - Read-only `sqlite_exact` clients use an immutable connection for a clean
   checkpointed database, or `mode=ro` when an active writer's complete WAL
@@ -144,6 +146,48 @@ Do not delete or unlink a live palace lock to recover ownership. Stop the
 owning process cleanly; the operating system releases its lock automatically.
 If corruption is suspected, back up the palace and run integrity/repair
 operations offline, with no writable service running.
+
+### Idle MCP writer lease release
+
+The stdio and HTTP MCP transports start a writer-idle watchdog. When no
+mutating tool has completed for 10 minutes, and no storage-touching request is
+in flight, the watchdog closes this process's storage handles and releases its
+writer lease. The next mutating request tries to acquire the lease again. It
+continues normally if the lease is free, or returns the peer-writer refusal if
+another process acquired it during the idle gap.
+
+Configure the threshold with `MEMPALACE_MCP_WRITER_IDLE_MINUTES`. The default
+is `10`; `0` disables idle release and restores hold-until-exit behavior. An
+unparseable value uses the 10-minute default, and a negative value is treated
+as `0`.
+
+This watchdog is independent of `MEMPALACE_MCP_IDLE_HOURS`. Setting the
+process idle timeout to `0` keeps an MCP server running, but does not disable
+writer-lease release. This separation lets an always-on server remain
+available without keeping local peers read-only while it is not writing.
+
+### Prefer one server where you control the deployment
+
+Where every client can use HTTP, prefer one `mempalace serve` process as the
+palace's writer. Concurrent client requests then serialize inside one process
+instead of moving the writer lease between processes. The systemd template is
+[`deploy/mempalace-server.service`](../deploy/mempalace-server.service).
+
+Local stdio clients and compatible CLI operations can discover that server
+through the per-palace registry in `~/.mempalace/server/<key>/`. The registry
+and its bearer token live under the current operating-system user's home. A
+server running under a dedicated service account therefore does not publish a
+registry that a human user's local clients can see. Discovery fails quietly in
+that topology and the client uses its local path, where the writer lease still
+applies. Configure those clients to use the HTTP server explicitly, or run the
+server and local forwarders under the same account.
+
+For a filesystem-level backup of a palace owned by an always-on service, stop
+the service, capture the complete palace, then restart the service and verify
+it. Do not make the backup conditional on `pgrep` finding no process under the
+service account: an always-on server makes that condition permanently false,
+so the backup would be skipped every time. Networked backends should use their
+backend-native snapshot procedure.
 
 ## Follow-up PRs
 
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -615,12 +615,13 @@ def _mine_args_forwardable(args, include_ignored) -> bool:
 def _forward_mine_to_hub(args, palace_path: str) -> bool:
     """Run this mine inside the palace's HTTP hub, if one is alive.
 
-    A long-lived hub (``mempalace serve``) holds the MCP writer lease for
-    its whole lifetime, so a direct mine from this process — including the
-    background save hooks, which spawn exactly this CLI — would be refused
-    and transcript capture would silently stop on the hub machine. The hub
-    itself may mine (the palace lock is process-re-entrant there, #1859),
-    so the fix is to hand it the job over HTTP.
+    A live hub (``mempalace serve``) is the intended writer for its palace. A
+    direct mine from this process — including the background save hooks, which
+    spawn exactly this CLI — would either be refused while the hub owns the
+    lease or take ownership after the hub releases an idle lease. Forwarding
+    keeps one process authoritative and avoids making the hub reacquire on its
+    next write. The hub itself may mine (the palace lock is process-re-entrant
+    there, #1859), so the job is handed to it over HTTP.
 
     Returns True when the hub handled the mine (this function has already
     printed the outcome and exited non-zero on failure). Returns False when
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -406,9 +406,11 @@ _STARTUP_INTEGRITY_MAX_MB_DEFAULT = 512.0
 #
 # The existing per-operation palace lock serializes individual writes, but it
 # cannot make another long-lived Chroma PersistentClient forget stale in-memory
-# HNSW/FTS state. Hold the same per-palace mine lock for this MCP process
-# lifetime. A peer MCP process can still serve read tools, but mutating tools
-# refuse before touching Chroma or the knowledge graph.
+# HNSW/FTS state. Acquire the same per-palace mine lock as a process-scoped
+# lease. The idle-release watchdog may drop it after a write-idle gap, and the
+# next mutating call reacquires it. While the lease is held, a peer MCP process
+# can still serve read tools, but mutating tools refuse before touching Chroma
+# or the knowledge graph.
 _MCP_WRITER_LOCK_CM = None
 _MCP_WRITER_READ_ONLY = False
 _MCP_WRITER_LOCK_FAILED = False
@@ -418,9 +420,9 @@ _MCP_ALLOW_PEER_WRITER_ENV = "MEMPALACE_MCP_ALLOW_PEER_WRITER"
 
 # Writer-lease idle release (fork patch 2026-08-19; upstream candidate).
 #
-# The lease above is otherwise held for the life of the process, so an
-# orphaned-but-alive stdio server (client machine rebooted; tailscale SSH
-# never delivered EOF) keeps every peer session read-only for hours
+# Without the watchdog, the lease above is held for the life of the process,
+# so an orphaned-but-alive stdio server (client machine rebooted; tailscale
+# SSH never delivered EOF) keeps every peer session read-only for hours
 # (observed twice on 2026-08-19). Acquisition is already self-healing per
 # mutating call, so the lease can be dropped whenever this process has not
 # written for a while: the next mutating call transparently re-acquires.
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -416,6 +416,22 @@ _MCP_WRITER_LOCK_ERROR = ""
 _MCP_WRITER_ATEXIT_REGISTERED = False
 _MCP_ALLOW_PEER_WRITER_ENV = "MEMPALACE_MCP_ALLOW_PEER_WRITER"
 
+# Writer-lease idle release (fork patch 2026-08-19; upstream candidate).
+#
+# The lease above is otherwise held for the life of the process, so an
+# orphaned-but-alive stdio server (client machine rebooted; tailscale SSH
+# never delivered EOF) keeps every peer session read-only for hours
+# (observed twice on 2026-08-19). Acquisition is already self-healing per
+# mutating call, so the lease can be dropped whenever this process has not
+# written for a while: the next mutating call transparently re-acquires.
+# The in-flight counter stops the watchdog from releasing (and closing
+# storage handles) under a running chroma-touching request.
+_MCP_WRITER_IDLE_MINUTES_ENV = "MEMPALACE_MCP_WRITER_IDLE_MINUTES"
+_MCP_WRITER_IDLE_MINUTES_DEFAULT = 10.0
+_MCP_WRITER_STATE_LOCK = threading.RLock()
+_MCP_WRITER_INFLIGHT = 0
+_last_mutating_time: float = time.monotonic()
+
 _MUTATING_TOOLS = frozenset(
     {
         "mempalace_kg_add",
@@ -690,6 +706,12 @@ def _discard_mcp_storage_handles() -> None:
 
 
 def _release_mcp_writer_lock() -> None:
+    """Thread-safe wrapper: release under _MCP_WRITER_STATE_LOCK."""
+    with _MCP_WRITER_STATE_LOCK:
+        _release_mcp_writer_lock_unlocked()
+
+
+def _release_mcp_writer_lock_unlocked() -> None:
     """Close writable handles and release this process's palace lease."""
 
     global _MCP_WRITER_LOCK_CM, _MCP_WRITER_READ_ONLY
@@ -709,6 +731,12 @@ def _release_mcp_writer_lock() -> None:
 
 
 def _acquire_mcp_writer_lock() -> tuple[bool, str]:
+    """Thread-safe wrapper: acquire under _MCP_WRITER_STATE_LOCK."""
+    with _MCP_WRITER_STATE_LOCK:
+        return _acquire_mcp_writer_lock_unlocked()
+
+
+def _acquire_mcp_writer_lock_unlocked() -> tuple[bool, str]:
     """Acquire this process's per-palace MCP writer lease.
 
     Returns (True, "") when this process may write. Returns (False, reason)
@@ -788,6 +816,8 @@ def _acquire_mcp_writer_lock() -> tuple[bool, str]:
     _MCP_WRITER_READ_ONLY = False
     _MCP_WRITER_LOCK_FAILED = False
     _MCP_WRITER_LOCK_ERROR = ""
+    global _last_mutating_time
+    _last_mutating_time = time.monotonic()
     return True, ""
 
 
@@ -804,7 +834,10 @@ def _mcp_peer_writer_refusal(req_id, tool_name: str):
         "id": req_id,
         "error": {
             "code": -32001,
-            "message": "Peer MCP writer active; this server is read-only for mutating tools",
+            "message": (
+                "Peer MCP writer active; this server is read-only for mutating tools"
+                + (f" [{reason}]" if reason else "")
+            ),
             "data": {
                 "tool": tool_name,
                 "palace": _config.palace_path,
@@ -6630,7 +6663,7 @@ def handle_request(request):
             if "entry" not in tool_args or tool_args["entry"] is None:
                 tool_args["entry"] = content_val
         try:
-            with _write_stall_watch(tool_name):
+            with _writer_inflight(tool_name), _write_stall_watch(tool_name):
                 result = _decorate_mcp_tool_result(
                     tool_name, TOOLS[tool_name]["handler"](**tool_args)
                 )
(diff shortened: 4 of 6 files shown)


jorgeccastro:
Positioning this against #2131, since both touch the writer lease and I would rather not have the two collide in review.

#2131 (thanks @Evgen197310) is demand-driven: a contender signals through the `.want` lock, the holder probes it, and ownership is handed over once no write is in flight and the minimum hold has elapsed, with a handoff window that parks writers arriving mid-release. For two live servers competing over the same palace that is the better mechanism, and it beats this PR on latency: seconds instead of minutes.

This PR is time-driven and asks nothing of the contender. The case that produced it, observed twice on 2026-08-19, is an orphaned but alive stdio server: the client machine went away, SSH never delivered EOF, and the process sat on the lease for hours without writing anything. There is no contender protocol to answer in that state, and by #2131's own compatibility note a holder that does not participate never answers the probe, so the contender still falls back to the read-only refusal. A holder that drops its own lease after an idle period covers that, and it also covers mixed-version setups where only one side runs the new code.

So I read them as complementary: #2131 as fast handoff under contention, this one as a floor guarantee that a lease is never held indefinitely by a process that is not writing. If you agree, the clean order is to land #2131 first and rebase this one on top: drop my in-flight counter in favour of #2131's write-frame depth tracking, and release only inside its handoff window. That would reduce this PR to the idle timer plus the clearer "who holds the lock" error message. I am happy to do that rebase.

One data point that may be useful either way: I have moved my own setup to `mempalace serve` over HTTP, a single server process owning the storage with several clients talking to it, and this class of problem disappears there, because there is no cross-process lease at all. Two diary writes fired concurrently from different sessions simply serialize inside the server. That is not an argument against either PR, since stdio stays the default path, but it may be worth documenting as an option for anyone running more than one client against one palace.

Would you prefer #2131 first and a rebase here, or should I close this and fold the idle timer into #2131?

Evgen197310:
Thanks for putting these side by side — and yes, I read them as complementary too. One premise I'd correct, and then what I think is a stronger justification for your PR than the one you're using.

**The orphaned-but-alive holder is already covered by #2131.** "Orphaned" is not the same as "not participating in the protocol". The handoff watchdog is a daemon thread started in `_run_stdio_loop`, and the dispatch lock is taken around `handle_request` only — never around `sys.stdin.readline()`. A server whose client vanished parks in readline holding no dispatch lock and no write frame, so its watchdog answers the demand probe on the next tick and the baton moves in poll + min-hold seconds. Your incident would still have happened on the code you were running (neither side had the protocol), but with #2131 on both sides that state resolves in seconds instead of waiting out the 8 h idle exit.

**The gap that does justify a holder-side floor is every writer that never signals demand.** The protocol itself lives in the shared layer (`mine_palace_lock(wait=…)`, `.want`, `palace_lock_wanted`), but only one caller opts in: `mcp_server.py` is the sole `wait > 0` site. Every other writer — `miner`, `convo_miner`, `sync`, `repair`, `closet_llm`, the backends, CLI `mine` — passes `wait=None`, gets `_lock_wait_default()` = 0, and never touches `.want` at all. A holder running #2131 therefore cannot see any of them and will never hand over. The default is deliberate (N hook-spawned mines must collapse rather than queue and then drive parallel HNSW inserts) and the only knob is one global env var, which is exactly why I could not make it the default: the wait that helps a peer server hurts a hook storm.

So the honest summary is that #2131 is universal in its API and MCP-only in practice, on both sides — contender and holder alike. Your timer is the half that needs nothing from the contender, which is why it covers what mine structurally cannot: pre-#2131 peers, non-MCP writers, and anything that opens the palace outside our code.

Worth noting the same asymmetry applies to yours: `_start_writer_idle_release_watchdog()` is called from `_run_stdio_loop` / `_run_http_loop` only, so a `daemon.py` holder runs no timer either. Neither PR is universal alone; together they cover both axes, minus the daemon — which is meant to be the sole owner anyway.

On mixed versions I'd narrow the claim slightly: an idle timer covers **new holder + old contender**. **Old holder + new contender** is covered by neither mechanism — the old holder answers no probe and runs no timer.

And neither PR is a superset of the other in one more place: a write wedged mid-call (#2024) blocks both designs — your in-flight counter and my write-frame depth both refuse, correctly.

**One question about the timer, not an objection.** The idle release fires whether or not a peer exists, so a single-server install — the common case — pays a Chroma/HNSW reopen on the first write after every idle period. My own `MIN_HOLD` knob exists because handing off means reopening storage, so I'm asking as someone who already conceded that cost: what is the reopen worth on a large palace, and should the default be more conservative than 10 minutes, or gated on the palace having been contended at least once in this process's life?

**On ordering — I'm bringing #2131 up to date with develop now.** It went `dirty` when 3.8.0 landed, and the update is not mechanical: `_HTTP_LOCK_FREE_TOOLS` arrived in develop, so the dispatch-lock barrier the watchdog leans on now has an exception. Correctness is fine by develop's own invariant — those tools write only to `logstream.sqlite3`, which is why they are already in `_PEER_WRITER_EXEMPT_TOOLS` — but it cuts the other way in stdio: wrapping them in the dispatch lock would let a five-minute `mempalace_event_wait` defer every handoff behind it. So stdio needs the same exemption HTTP has: the #1984 narrowing I flagged in the PR, arriving sooner than I expected. I'll merge develop in rather than rebase, so the branch history stays intact for anyone reading it mid-review — say the word if you'd rather have it linear and I'll rebase instead.

Your sequence is the one I'd pick: #2131 first, this one rebased on top as its own PR — the idle timer plus the clearer refusal message, dropping the in-flight counter in favour of the write-frame depth and releasing inside the handoff window. Two different guarantees deserve two reviews; folding a time-driven release into a 1.4k-line demand-driven diff mixes two models in one review. But the ordering is @igorls's call, not mine — if landing the timer first is the faster path to relief for your deployment, I have no objection, and I'll bring #2131 up to date either way.

There is also a cheaper way to close the hooks half of #1888 later — a fail-fast contender could leave a short-lived demand mark instead of queueing, so a holder learns about writers that cannot afford to wait — but that is a third design and this thread is about ordering, so I'll file it against #1888 once the sequence is settled.

+1 on documenting `mempalace serve` over HTTP as the recommended shape for more than one client on one palace. That belongs in `docs/writer-lease-handoff.md`; I'll add it in that update unless you'd rather own it.

jorgeccastro:
Thanks, that is a better justification than mine and I'll take it: the orphaned holder is covered once both sides run #2131; the structural gap is every writer that never signals demand, and the timer is the half that needs nothing from the contender. Agreed on the narrowed mixed-version claim (new holder + old contender only) and on the daemon gap; the daemon is meant to own the palace, so no timer there.

On the reopen cost, instead of guessing, here is the data from my palace (2.9k drawers, 34 MB on disk): cold `PersistentClient` 54 ms, collection count + first HNSW query 19 ms, so a full reopen is under 100 ms. It grows with the HNSW index on large palaces, but it is one reopen per idle gap of at least the threshold, so the write that pays it is already the first after 10 minutes of silence. I'd keep 10 minutes as default and keep the knob (`MEMPALACE_MCP_WRITER_IDLE_MINUTES`, 0 disables). Gating on "contended at least once" sounds right but the holder cannot observe contention from the writers this timer exists for: they fail fast on the flock and leave no mark. If your demand-mark idea for #1888 lands, that becomes the right gate and the timer can key on it; until then time is the only signal the holder has.

Merge commit is fine by me; readable review history beats linear. #2131 first, then I rebase this onto it as the timer plus the refusal message, using your write-frame depth and releasing inside the handoff window.

Please own the docs section; I'll review it and can add the `serve` specifics from my setup (systemd unit, token file, and one gotcha I hit: any backup or maintenance script that decides "palace in use" by `pgrep` on the service user will always skip once a server runs 24/7; mine now stops and restarts the unit around the backup window).

Evgen197310:
Thanks for measuring instead of arguing about it. Under 100 ms for a full reopen on a 34 MB palace settles my question: 10 minutes as the default and the knob to move it is the right shape. The cost I was worried about is real only at the top of the size range, and there it is one reopen paid by the first write after ten minutes of silence — which is exactly the write that is least likely to be latency-sensitive.

And you are right about the gate, in a way that kills my own suggestion: a holder cannot observe contention from the writers your timer exists for, because they fail fast on the flock and leave nothing behind. So "gate on ever contended" would be blind precisely where the timer earns its keep. Time really is the only signal available today.

I filed the demand mark as #2340. Short version: on the fail-fast path, stamp the demand file before raising `MineAlreadyRunning` instead of leaving no trace, and let the holder's existing probe read a fresh mark as demand. It is the one place where #2131's "signal by holding a lock, never by writing state" rule has to bend, and the bend is cheap — a stale mark costs one unnecessary handoff, which your measurement now prices for us. If it lands, your timer gets the gate you actually want, and the two mechanisms stop guessing separately about the same question. The open questions are in the issue; the Windows byte-locking one is the fiddly bit.

On the docs: I own the `serve` section, and before promising it to you I read both lease pages against the branch rather than against my memory of them. Five things did not match, all in the direction where a reader trusts the page over the source — the holder's steps were printed in the wrong order (the minimum-hold check runs before the demand probe, not after), the contender's back-off after a timed-out wait was undocumented, hub discovery was described as configuration when it is automatic through the per-palace registry (`MEMPALACE_HUB_FORWARD=off` disables it), the lock policy said nothing about the forward deliberately taking no lock, and `MIN_HOLD` was in code font in two places where no such symbol exists. `docs/write-routing-policy.md` also still told stdio readers that writable storage reopens "after that owner exits" — the one sentence someone would quote to conclude this branch does not exist. All fixed in `c35d324`.

The section itself is `## When one server is the better answer` in `docs/writer-lease-handoff.md`. It says what you said — one process owning the storage, no cross-process lease, writes from several sessions serializing inside the server, and the hub proxy making that the default for stdio clients too — and then says the handoff is for everywhere you do not control the deployment. Please review it, and send the `serve` specifics: the systemd unit and token file belong there, and the `pgrep` gotcha belongs there most of all. That one is not a footnote — a backup script that decides "palace in use" by looking for a process under the service user will skip its work *every night* once the server runs 24/7, and it will do so silently, which is the worst failure mode a backup can have. Your stop-restart-around-the-window fix is the kind of thing nobody derives from first principles; they discover it after checking a restore.

Merge commit stays, then. #2131 is up to date with develop and reads `mergeable: clean`; the branch is yours to rebase onto once @igorls lands it.

jorgeccastro:
Read `## When one server is the better answer` (c35d324) against the code and against the machine I run it on, rather than against my memory of either. It matches what I have, and the framing is right: the handoff is for the deployments you do not control. Three things back, one of them a caveat I would want in the section itself.

**A narrowing on the knobs.** "none of the knobs on this page apply" is true for the shape you describe, but `MEMPALACE_WRITER_HANDOFF_ENABLED` is on this page, and it is the one an operator of a single server is most likely to reach for, on the theory that more handoff is more safety. Worth one clause saying why not: with one process owning the storage there is no peer to hand to, so opting in only adds a watchdog and a release path to a process that is meant never to let go.

**The caveat: hub discovery is per-user, and failing to discover is silent.** The section says the hub proxy makes the single server the default experience for stdio clients too. That holds when the clients run as the same OS user as the server. It does not when they do not, and the project's own template puts them on different users: `deploy/mempalace-server.service` creates a dedicated `mempalace` system account, while hooks and `mempalace mine` run as the human.

`server_state_dir()` is `Path.home() / ".mempalace" / "server" / <key>`, with no environment override, so the reader computes the path in its own home and finds nothing the service user wrote. Worse, both lookups fail quietly by design: `read_live_serverinfo` catches `OSError` and returns `None`, which is the same answer as "no hub is running", and `load_server_token` returns `""` on `OSError`. So a cross-user deployment does not report a discovery problem. It reports no hub, silently falls back to the local path, and lands the operator on exactly the page they thought they had opted out of, with two direct writers contending over one palace.

That is not something this PR introduces, and I am not asking you to fix it here. But this section is where a reader decides to go the single-server route, so it is the right place for a sentence: the hub forward reaches local stdio clients only when they run as the same user as the server; a dedicated service account means those clients keep using the handoff. If you would rather it live in the remote-server guide, that works too, as long as this section points at it.

**On the unit and the token, I would link rather than paste.** `deploy/mempalace-server.service` is more hardened than what I run (`ProtectSystem=strict`, `RestrictAddressFamilies`, `ReadWritePaths`), and the `Restart=always` and idle-watchdog story is already settled in #2201 and documented in the remote-server operating notes. Pasting my unit into this doc would give readers a weaker template and a second place to keep in sync. Two deltas from my setup that might be worth a line in the remote-server guide, not here:

- The template binds `0.0.0.0` and tells you to front it with TLS. A third option, and the one I use, is to bind a private-network address, so the listener is never on the public interface at all. On a VPS the public NIC is the default route, and "bound to 0.0.0.0 behind a proxy I have not configured yet" is a real intermediate state.
- Token resolution is flag, then env, then auto-generated per-palace file at `~/.mempalace/server/<sha256 of the canonical palace path, first 24>/token`, 0600 in a 0700 directory. The template's `EnvironmentFile` path and the auto-generated path are both correct, but they put the token in different places, and the directory key comes from the canonical path (`realpath` + `normcase`), so a unit naming a path that resolves differently from the clients' gets a different directory.

## The backup gotcha, which I do think belongs somewhere

The remote-server guide hands backups to the storage backend (Milvus snapshots, Qdrant, Postgres). That is right for those deployments and leaves out the one where the palace is files on the server's own disk, which is the shape most people start with and the one where this bites.

My weekly backup decided whether the palace was in use like this:

```bash
for i in $(seq 1 5); do
    pgrep -u mempalace > /dev/null || break
    [ "$i" -eq 5 ] && { log "SKIP: mempalace processes still active"; exit 1; }
    sleep 60
done
```

Correct while every MCP server was spawned on demand and idle most of the time. The day the HTTP server moved to 24/7 under systemd, the predicate became permanently true, and the backup started exiting `SKIP` every week. Nothing crashed, nothing alerted: the job ran on schedule, logged a plausible reason, and did no work. You find out at restore time, which is the worst time.

The fix is to schedule the quiet instead of waiting for it, and to guarantee the restart on every exit path:

```bash
SERVE_UNIT=mempalace-server.service
SERVE_WAS_ACTIVE=no
if systemctl is-active --quiet "$SERVE_UNIT"; then
    SERVE_WAS_ACTIVE=yes
    systemctl stop "$SERVE_UNIT"
fi
restore_serve() {
    [ "$SERVE_WAS_ACTIVE" = yes ] && systemctl start "$SERVE_UNIT"
}
trap restore_serve EXIT
```

The `pgrep` wait stays after that block, where it means what it always meant: a `mine` or a CLI writer that happens to be running right now.

The general form, if the doc wants the lesson rather than my script: once one process holds the palace for the lifetime of the machine, every "is anyone using the palace" heuristic written against process presence inverts. Presence stops being evidence of activity, and any maintenance that used to wait for quiet has to create it. That applies to backups, to `repair`, and to anything else written when the palace was free most of the time.

Happy to turn any of this into a PR against your branch, or into a separate one against the remote-server guide, whichever keeps your diff clean.

rubicon:
Validated this against the reproduction I posted on #2024, on macOS with an always-on HTTP server under a LaunchAgent.

Setup: temp HOME and temp palace, `--transport http`, `MEMPALACE_MCP_IDLE_HOURS=0` so the idle-exit watchdog is out of the picture. That is my real configuration, and it is what made the lease permanent for me. Sequence: one `mempalace_add_drawer` so the server takes the lease, then a separate process attempts `mine_palace_lock`, which is what the Stop-hook diary checkpoint does.

On develop at 98e6dfc, with `MEMPALACE_MCP_WRITER_IDLE_MINUTES=0.1`, the hook probe is BLOCKED both while the lease is held and after the idle period. On this branch, same sequence and same knob, the probe is BLOCKED while held and TOOK after the idle period. The server re-acquired and wrote again on the next mutating call in both runs, so the handoff is not one-way.

Two details that matter for my case. The new watchdog starts independently of `_start_idle_exit_watchdog`, so `MEMPALACE_MCP_IDLE_HOURS=0` does not disable it. And with `MEMPALACE_MCP_WRITER_IDLE_MINUTES` unset, `_writer_idle_release_secs()` returns 600, so an existing deployment picks this up without a config change. Both are what I would want.

Results on the branch: `tests/test_mcp_writer_idle_release.py` passes 8/8, and the rest of the suite is 4300 passed, 31 skipped.

One thing worth flagging: the branch is 64 commits behind develop, and `tests/test_palace_graph_limits.py` no longer exists upstream. Four tests in that file fail here, but they also fail with your `mcp_server.py` change reverted, so they are inherited from the stale base rather than caused by this PR. A rebase should clear them.

jorgeccastro:
Thanks for running it on the configuration I do not have. An always-on HTTP server under a LaunchAgent with `MEMPALACE_MCP_IDLE_HOURS=0` is exactly the shape where the lease is permanent by construction, so a BLOCKED / TOOK pair on that setup is worth more than another Linux run from me.

Both details you flagged are deliberate, and I am glad they read that way from the outside. The new watchdog starts independently of `_start_idle_exit_watchdog` because the two answer different questions: idle exit is about file handles outliving a session, the lease timer is about peers being refused while this process is alive and simply not writing. Tying them together would mean anyone who set `MEMPALACE_MCP_IDLE_HOURS=0` to keep a server resident also opted out of the fix they need most. Same reasoning for the unset default: an existing deployment picks up 600 s without a config change, and `MEMPALACE_MCP_WRITER_IDLE_MINUTES=0` restores the old hold-until-exit behaviour for anyone who wants it.

## Rebased

Rebased onto develop at `4c1e6d0`. The commit applied clean, no conflicts, and the diff is unchanged at 212 insertions and 2 deletions across `mempalace/mcp_server.py` and `tests/test_mcp_writer_idle_release.py`. The branch was 64 commits behind when you measured and 67 by the time I rebased.

`tests/test_palace_graph_limits.py` no longer exists upstream, so the four failures you traced to the stale base are gone with the base, which matches your reading that the PR did not cause them.

Local run on the rebased branch, Linux, Python 3.13.15, chromadb 1.5.9:

- full suite: 4537 passed, 31 skipped, 106 deselected
- `tests/test_mcp_writer_idle_release.py`: 8/8
- `test_palace_locks.py`, `test_vector_write_gate.py`, `test_mcp_http_transport.py`: 89 passed together

## On the re-entrancy constraint you posted on #2172

You are right that this widens the window, and I want to be precise about which half of it the patch reaches.

The timer only ever releases. `_maybe_release_idle_writer` returns early unless the lease is held, `_MCP_WRITER_INFLIGHT` is zero and the write-idle clock is past the threshold, and it does all of that under the same `_MCP_WRITER_STATE_LOCK` the acquire path takes. It never calls `mine_palace_lock`, so it cannot turn inherited credit into a lock somebody else can steal. That failure mode belongs to the frame that releases the real flock, not to this thread.

What it can do, if `_MCP_WRITER_LOCK_CM` ever came from the passthrough branch, is drop the local credential and close storage handles while another frame still holds the real flock. `__exit__` on a passthrough CM releases no flock, so nothing is handed to an outsider, but `_discard_mcp_storage_handles()` runs either way. That is safe only while every chroma-touching path in the process goes through `handle_request`, which is what the in-flight counter covers. Today it does: HTTP and stdio both dispatch there, the logstream set is skipped because it never takes the lease, and the three background threads touch no storage (idle exit calls `os._exit`, the write-stall watchdog only reports, and this one takes the state lock). So I land where you did: incidental, not guaranteed.

If the WriterLease contract gives the acquire a generation, the timer wants one more condition rather than a redesign: refuse to release a lease whose generation is not the one this frame acquired. That is a line in `_maybe_release_idle_writer`, and it converts "safe because everything happens to funnel through one place" into "safe because the lease says so". I would rather it land that way than leave a time-driven release leaning on incidental serialization, so if the contract work in #2172 goes anywhere, treat this PR as a consumer of it.

## Ordering

Unchanged from what @Evgen197310 and I settled above: #2131 first, then I rebase this on top as the timer plus the clearer refusal message, dropping the in-flight counter in favour of the write-frame depth and releasing inside the handoff window. This rebase is only to take the stale base out of the review, not to jump the queue.

Evgen197310:
**Ordering is released — please don't sequence behind me.**

I'm stepping back from upstream work on the writer lease (details in #2131), so the plan we
settled on — #2131 first, this one rebased on top — should not hold your PR up any longer.
Your timer stands on its own: it needs nothing from #2131 to be correct, it covers the axis
#2131 structurally cannot (every writer that signals no demand), and @rubicon has now
validated it on a configuration neither of us runs. If it lands first and someone later
picks up #2131, the in-flight counter → write-frame depth swap is a small follow-up in the
other direction.

Two things I owe you before I go quiet.

**The docs section is yours.** `## When one server is the better answer` in
`docs/writer-lease-handoff.md` is on the #2131 branch (`c35d324`) — take it, in whatever
form suits you, along with your own three points on it: the
`MEMPALACE_WRITER_HANDOFF_ENABLED` clause, the cross-user hub-discovery caveat, and linking
`deploy/mempalace-server.service` rather than pasting a unit. The `pgrep` backup gotcha
belongs in there most of all, and you should write it, because you found it by checking a
restore.

**Field data for that section, since we now run the shape it recommends.** On 17 Aug we cut
over to a single `mempalace serve` process and every client talks to it over HTTP:

```
ExecStart=… -m mempalace.cli serve --host 127.0.0.1 --port 8787 \
    --backend qdrant --palace /…/palace350-qdrant
Restart=always
```

- The cross-process lease left our topology entirely, exactly as you predicted. Concurrent
  diary writes from several sessions serialize inside the server; nothing contends.
- Your cross-user caveat did not bite us, but only by accident worth naming: we bind
  loopback and all clients are HTTP, so hub discovery never runs. The trap is specific to
  *local stdio* clients under a different user than the server — worth saying that way in
  the section, because "we have a server, so we're fine" is the wrong reading.
- We kept `User=root` from the stdio era rather than the template's dedicated account:
  changing the owner and the transport in one step is two migrations, not one. Weaker than
  the template, so not an example for the docs — but probably the state most people cut over
  from.
- `MEMPALACE_MCP_IDLE_HOURS=0` with `Restart=always`, and we kept
  `MEMPALACE_MCP_WRITE_STALL_WARN_SECS` from #2158 — on a single-owner server a wedged write
  is the failure that has no second writer to route around it, so the report matters more
  there, not less.

Thanks for the reopen measurement, and for putting the two PRs side by side instead of
letting review discover the collision. That thread is the most useful review I've had here.

jorgeccastro:
Thanks, this closes the ordering question cleanly. I’ll keep #2305 standalone and carry the documentation points forward, including the cross-user discovery caveat and the backup `pgrep` failure mode. Your deployment notes are useful, and I appreciate you leaving #2131 in a takeable state. I won’t sequence #2305 behind it.

christianlappin:
A data point from a deployment this PR would have prevented, plus two things we changed when backporting the idea to 3.6.0.

Setup: one stdio mempalace-mcp per Claude Code session on a Mac mini, a second machine over SSH stdio, an hourly `mempalace mine` from launchd. Chroma, chromadb 1.5.9, ~150k drawers, legacy batch/sync thresholds 50000.

Cost of the lifetime lease: 2,818 mine calls refused by an MCP holder between May and mid-September; almost every hourly run over three days refused, by five session servers in turn.

What we run now (local patch on 3.6.0, same env var, default 2 min):
1. Discard storage handles on acquire as well as release (3.6.0 lacks #2026). Checked first that discarding writes nothing: close(), clear_system_cache() and exit leave HNSW files byte-identical with 50000 thresholds; a threshold-100 control flushes.
2. Count the call in flight before the peer-writer preflight, not only around dispatch, so a tick between the check and the handler cannot release a lease the call just passed.

Tests: two real stdio servers + the CLI miner; release 16-19 s after the last write on a 15 s window, exactly one winner of a simultaneous write, a 35 s mempalace_mine keeps the lock, quick_check / capacity probe / counts agree; reopen on a production clone about 2 s.

+1 to a holder-side floor independent of #2131.

rubicon:
Thanks. This is the deployment evidence the PR needed. The 2,818 refused mine calls establish the value of a holder-side idle floor independently of #2131.

One concurrency gap before merge. `_acquire_mcp_writer_lock_unlocked()` returns early when the lease is already held (mcp_server.py:762-763) without refreshing `_last_mutating_time`; only the fresh-acquire path does that (:822). So a held but write-idle lease clears `_mcp_peer_writer_refusal()` with `idle` still past the threshold, and `_writer_inflight` does not start until :6668. A watchdog tick in that window sees `_MCP_WRITER_INFLIGHT == 0`, discards the storage handles and drops the flock. All of that happens after the call was cleared to write, and a peer can then win the flock while the handler runs. The window is narrow, but it is not bounded by anything structural: the thread can be descheduled between those points, so this is rare rather than impossible, and it permits two live writers on one palace.

Suggested shape: begin counter-only in-flight protection immediately before the peer-writer ownership check and hold it through dispatch, refreshing `_last_mutating_time` only after dispatch completes. The earlier integrity, stale-library and diverged-index gates (:6491-6515) must stay outside that frame.

Reusing `_writer_inflight` around the whole preflight would not work: its `finally` refreshes the clock for any `_MUTATING_TOOLS` call including refused ones, so a server holding the lease while refusing at those earlier gates would never go idle, which is the starvation this PR fixes. Splitting the counter from the clock refresh avoids that. Note `_HTTP_LOCK_FREE_TOOLS` (:7147) is defined below `handle_request`, so keep that exempt-set lookup lazy.

Two deterministic regression tests would close it:

1. Hold the lease, set the clock past the threshold, pause after a successful peer-writer preflight, run the watchdog, and assert the lease survives dispatch.
2. Refuse a mutating call repeatedly at an earlier gate and assert those refusals do not refresh the write-idle clock.

The real stdio/CLI topology, post-reopen consistency checks and no-flush validation are good field evidence but belong in follow-ups, not this fix. The ~2s reopen on the production clone supports keeping the 10-minute default, and I agree this should stay independent of #2131.

jorgeccastro:
Thanks for the careful read. You were right about the window, and it's fixed in `db24f83`.

The ownership check now runs inside the in-flight frame, and the frame holds until dispatch returns. `handle_request` calls the preflight with `check_writer=False` and does the peer-writer check itself, inside `_writer_inflight`. `_mcp_tool_preflight_refusal` keeps its default, so the light server path is unchanged. The integrity, stale-library and diverged-index gates stay outside the frame. The counter no longer touches the clock. A separate `_writer_idle_clock` refreshes it after dispatch only, so a refused call at any gate never keeps an idle holder alive.

Both tests you described are in `tests/test_mcp_writer_idle_release.py`. The first pauses right after a successful ownership check, runs the watchdog, and asserts the lease survives dispatch. It fails on the previous head: the tick released the lease and the handler still wrote. The second refuses repeatedly at every gate, peer-writer included, and asserts the clock doesn't move. I also checked that it catches the naive fix: wrapping the whole preflight in the old `_writer_inflight` makes it fail.

The branch is rebased onto current develop and ported to the `mcp_server/` package split. Full suite: 5605 passed. Two tests fail the same way on pristine develop: `test_repair::test_sqlite_integrity_errors_reports_a_path_python_cannot_encode` and the light-server schema benchmark (69.98 vs a 70 threshold).

Two notes on scope. KG tools are in `_HTTP_LOCK_FREE_TOOLS`, so they don't count as in flight. That's safe as far as I can tell, since they never touch Chroma and have their own lock. And the light server and `daemon.py` still run no idle timer, same as before this PR.

CI needs a maintainer to approve the workflow runs for this fork.

PR #2306, opened 2026-08-19
Redact credential-shaped strings before mined chunks are persisted

ke-netizenmin:
## The problem

`mine_convos()` writes chunk content into the collection verbatim. `_file_chunks_locked()` does:

```python
batch_docs.append(chunk["content"])
```

Nothing on that path filters the text. Mined conversation transcripts routinely contain real credentials — a service-account JSON echoed to a shell, an `Authorization` header in a `curl` command, a token pasted while debugging — and once a chunk is embedded, the secret is durable: it persists in the vector store on disk, is returned by search, and is copied into every export and backup.

This is easy to miss because redaction *does* exist elsewhere. Deployments that wrap the MCP server can filter `add_drawer`/`update_drawer` calls, so manually-filed drawers are covered — but the periodic/cron miner doesn't go through that path. The result is an asymmetry where the interactive route is protected and the bulk-ingest route, which processes far more text, is not.

## The fix

Adds `mempalace/secret_redact.py` and calls it from `_file_chunks_locked()` before the document is appended to the upsert batch:

```python
content = redact_secrets(chunk["content"])
batch_docs.append(content)
```

The module is **regex-only and dependency-free**, so it runs inline during ingest with no network call or model load, and it short-circuits on a cheap substring check because the overwhelming majority of chunks contain no credential marker at all.

Patterns cover PEM private-key blocks, the JSON `private_key` field, Slack `xox*` / `xapp-` tokens, and `Authorization: Bearer` headers.

### Two details that are easy to get wrong

Both are covered by tests, because both are the difference between "looks like it redacts keys" and "actually does":

1. **The label between `BEGIN` and `PRIVATE KEY` is optional.** PKCS#8 blocks are plain `-----BEGIN PRIVATE KEY-----`, with no `RSA`/`EC`/`OPENSSH` word — that's the format Google service-account JSON uses, and so one of the most commonly leaked key types. A pattern written as `BEGIN [A-Z ]+ PRIVATE KEY` requires at least one character in that slot and therefore misses exactly that case while appearing to handle private keys.

2. **The `END` marker is optional.** Transcripts get chunked, so a key is frequently split mid-body and the terminator never arrives in the same chunk. A pattern anchored on `END` fails to match and the visible base64 fragment gets written out verbatim. This redacts the body it can see.

## Conservative by construction

Every pattern requires a structural marker — a PEM header, a known token prefix, an `Authorization` scheme — plus a minimum length. Prose *about* credentials and code that merely handles them is left untouched, and the tests assert that directly:

```python
"We discussed the xoxb- prefix and how tokens are shaped."
"The header is -----BEGIN PRIVATE KEY----- followed by base64."
"Set private_key in the config before deploying."
```

Non-secret sibling fields also survive, so a redacted drawer stays useful — a service-account blob keeps its `client_email` and loses only the key.

## Tests

22 tests in `tests/test_secret_redact.py`, covering all five PEM label variants, truncated blocks, escaped-`\n` PEM inside JSON, the `private_key` field without PEM markers, Slack and Bearer tokens, multiple secrets in one chunk, idempotence, and the negative cases above.

```
22 passed
```

`ruff check` and `ruff format --check` clean.

Run with `--noconftest` locally (`tests/conftest.py` imports `chromadb`, which isn't needed by this file). The `convo_miner` change is import- and lint-checked but I did not run the chromadb-backed suites.

One note on the Slack fixtures: they're assembled at import time from parts rather than written as literals. A literal in the live token shape trips GitHub push protection and downstream secret scanners — the strings hold no real credential, but a blocked push looks identical to a real leak to whoever hits it.

Targeting `develop` per CONTRIBUTING.md.

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -17,6 +17,7 @@ from datetime import datetime
 from collections import defaultdict
 
 from .normalize import normalize
+from .secret_redact import redact_secrets
 from .palace import (
     NORMALIZE_VERSION,
     SKIP_DIRS,
@@ -351,13 +352,20 @@ def _file_chunks_locked(collection, source_file, chunks, wing, room, agent, extr
                 if extract_mode == "general":
                     room_counts_delta[chunk_room] += 1
                 drawer_id = f"drawer_{wing}_{chunk_room}_{hashlib.sha256((source_file + str(chunk['chunk_index'])).encode()).hexdigest()[:24]}"
-                batch_docs.append(chunk["content"])
+                # Strip credentials before the chunk is embedded. Transcripts
+                # routinely contain real secrets (a service-account JSON echoed
+                # to a shell, an auth header in a curl command), and once a
+                # chunk lands in the vector store the secret is durable: it
+                # persists on disk, is returned by search, and is copied into
+                # every export and backup.
+                content = redact_secrets(chunk["content"])
+                batch_docs.append(content)
                 batch_ids.append(drawer_id)
                 batch_metas.append(
                     {
                         "wing": wing,
                         "room": chunk_room,
-                        "hall": _detect_hall_cached(chunk["content"]),
+                        "hall": _detect_hall_cached(content),
                         "source_file": source_file,
                         "chunk_index": chunk["chunk_index"],
                         "added_by": agent,
--- a/mempalace/secret_redact.py
+++ b/mempalace/secret_redact.py
@@ -1,0 +1,68 @@
+"""Strip credential-shaped strings out of text before it is persisted.
+
+Mined conversation transcripts routinely contain real credentials — a shell
+session that echoed a service-account JSON, a curl command with an auth header,
+a token pasted for debugging. Once a chunk is embedded into the vector store the
+secret is durable: it survives on disk, is returned by search, and gets copied
+into any export or backup.
+
+This module is deliberately regex-only and dependency-free so it can run inline
+on the ingest path without a network call or model load.
+"""
+
+import re
+
+_REDACTED = "[REDACTED_SECRET]"
+
+# PEM blocks. Two things matter here beyond the obvious:
+#   * The label between BEGIN and PRIVATE KEY is optional. PKCS#8 keys — the
+#     format Google service-account JSON uses — are plain
+#     "-----BEGIN PRIVATE KEY-----", with no RSA/EC/OPENSSH word. A pattern that
+#     requires a label silently misses the single most common leaked key type.
+#   * The END marker is optional. Transcripts get chunked mid-key, so a block
+#     may be truncated; we still redact the base64 body we can see rather than
+#     failing to match and writing the fragment out verbatim.
+_PEM_BEGIN = r"-----BEGIN (?:[A-Z0-9]+ )*PRIVATE KEY-----"
+_PATTERNS = [
+    # Complete PEM block.
+    (re.compile(_PEM_BEGIN + r"[\s\S]*?-----END (?:[A-Z0-9]+ )*PRIVATE KEY-----"), _REDACTED),
+    # Truncated PEM block: header plus a long base64 run, escaped \n included.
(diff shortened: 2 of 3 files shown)


ke-netizenmin:
Follow-up commit pushed: the truncated-block pattern only handled a single escaped newline.

A key embedded in one layer of JSON carries `\n` inside the base64 run, which the original alternation matched. But transcripts nest JSON — once that JSON is itself embedded in another JSON document the newline arrives as `\\n`. The alternation consumed the first backslash and stopped at the second, so the run fell short of the 40-character minimum and the key body was written out verbatim.

```diff
-    (re.compile(_PEM_BEGIN + r"(?:\\n|[A-Za-z0-9+/=\s]){40,}"), _REDACTED),
+    (re.compile(_PEM_BEGIN + r"[A-Za-z0-9+/=\s\\n]{40,}"), _REDACTED),
```

Folding the backslash into the character class lets the run cross any escape depth. The complete-block pattern is unaffected — it still runs first and consumes whole blocks including the `END` marker.

Test coverage: the escaped-newline case is now parametrized over both `\n` and `\\n`, plus a new truncated double-escaped block case. 24 tests pass, ruff clean.

Worth noting for reviewers that this is the difference between the pattern working on realistic input and not — in a corpus of mined transcripts, the doubly-escaped form was the more common of the two.

igorls:
Thanks for this — the problem is real and the engineering here is careful. The two details you called out are exactly the ones that separate working redaction from redaction-shaped code: PKCS#8 blocks having no label between `BEGIN` and `PRIVATE KEY` is precisely the Google service-account case, and handling a key split across chunks with no `END` marker is the kind of thing that only shows up against real transcripts. The conservative-by-construction tests, where prose *about* `xoxb-` survives untouched, are the right instinct too.

I can't take it at this layer, though, and I want to be specific about why rather than leave you guessing.

**Ingest-time redaction conflicts with the project's foundational promise.** From the design principles:

> **Verbatim always** — Never summarize, paraphrase, or lossy-compress user data. The system searches the index and returns the original words. If a user said it, we store exactly what they said. This is the foundational promise.

and the contribution guidance names "shortcuts that bypass verbatim storage" as out of scope. `redact_secrets()` in `_file_chunks_locked()` means the palace no longer contains what the user said. That is a one-way door: the original is not stored anywhere else, so a user who mines their own transcript and later needs that value has lost it, with no record of what was removed or why. A false positive is unrecoverable.

**The threat model also points somewhere else.** A credential sitting in a local palace has not crossed a trust boundary — the palace is local-first and by design never sends content anywhere. The moment content actually leaves the machine is recall time, when a BYOK external provider is configured and drawer text is about to be sent to it. That is the boundary worth filtering at, and it is the only one where redaction protects something.

So I think this lands as **recall-time redaction on the external-provider path**, not ingest:

- Storage stays verbatim. Nothing is destroyed, and the user can always get their own data back.
- Filtering applies when content is about to go to a cloud model, which is where the exposure actually is.
- It can be opt-in and reversible, and a false positive costs a redacted answer rather than deleted data.
- Local runtimes (Ollama and friends) are part of the user's machine, so they need no filtering — which also keeps the default path free of the regex work entirely.

Almost all of what you have written survives that move. `secret_redact.py` is the reusable part; the patterns, the minimum-length guards, the structural-marker requirement and the whole test corpus carry over unchanged. What changes is the call site, and the fact that it operates on outbound content rather than on what gets persisted.

Two smaller notes for whenever you pick it up:

(item 1) The branch is currently conflicting with develop, so it needs a rebase regardless — the read-path work has moved a lot underneath it.

(item 2) You noted the asymmetry where MCP wrappers can filter `add_drawer` but the miner bypasses it. That observation is worth keeping in the new framing: at recall time there is a single outbound path, so the asymmetry disappears rather than needing to be patched in two places.

If you would rather not carry the rework, say so and I will open an issue crediting the analysis so it does not get lost — the split-chunk and PKCS#8 findings are worth keeping either way.

PR #2310, opened 2026-08-19
fix: support Windows drive letters in transcript fallback regex

### LINKED ISSUE #2304: _wing_from_transcript_path fallback never matches Windows-encoded project folders (D--...)
**Observed on:** 3.3.5 (Windows 11, Claude Code); fallback code unchanged at current `develop`.

Claude Code encodes project directories into transcript paths by replacing path separators with `-`. On POSIX the encoded folder starts with a leading dash (`-home-user-project`), but on Windows it starts with the drive letter: `D--Code-BLM-blm-cadastral-returns-tools`.

The fallback branch of `_wing_from_transcript_path` matches `/\.claude/projects/-([^/]+)` — it requires the leading dash, so every Windows-encoded folder misses and diary checkpoints land in the generic `wing_sessions`. Per-project wings never form from the fallback path.

The v3.3.6 change making JSONL `cwd` the primary wing source (#1424 / #1410) covers most real sessions, but whenever `cwd` is absent from the transcript the fallback still silently mis-wings on Windows. Note the existing "Windows" test exercises a POSIX-encoded folder name (`-home-jp-Projects-myapp` style), so the suite currently encodes the POSIX assumption rather than guarding against this.

**Suggestion:** accept `[A-Za-z]--` drive-letter prefixes in the fallback regex (or relax it to "the last path component under `.claude/projects/`") and add a true Windows-encoded fixture to the test.

Effect at our site before upgrading: 100% of diary checkpoints in `wing_sessions`; project-filtered diary retrieval impossible. Searched open/closed issues for the drive-letter encoding and found nothing filed, hence this report.

faisalhussain-devs:
## What does this PR do?
Closes #2304

This PR updates the `_wing_from_transcript_path` fallback regex to gracefully handle Windows-encoded paths (e.g., `D--Code...`) generated by Claude Code. 

Previously, the regex strictly required a leading dash (`/\.claude/projects/-([^/]+)`), causing Windows drive letters to miss the match and silently fall back to `wing_sessions`. The regex has been updated to `(?:-|[a-zA-Z]--)` to support both POSIX hyphens and Windows drive letter prefixes.

## How to test
I have added a dedicated Windows test fixture in `tests/test_hooks_cli.py` using the exact failing path provided in the issue report. 

You can verify the fix by running:
`uv run pytest tests/test_hooks_cli.py::test_stop_hook_derives_wing_from_transcript_path_windows -v`

## Checklist
- [x] I have read the contributing guidelines.
- [x] Code is formatted via `pre-commit` (Ruff).
- [x] A test was added to prevent future regressions.
- [x] Local test suite passes (`uv run pytest tests/ -v`).


igorls (approved):
Thanks for the clear write-up and for including a real regression test — the issue described the fix and this matches it precisely.

CI is green across all nine jobs, including Windows. I also checked the regex against the path shapes that seemed most likely to break:

```
POSIX (regression)        -> wing_mempalace                              unchanged
Windows drive letter      -> wing_code_blm_blm_cadastral_returns_tools   fixed
lowercase drive           -> wing_lowercase_drive                        works
two letters (not a drive) -> wing_sessions                               unchanged
no prefix at all          -> wing_sessions                               unchanged
```

The alternation ordering is what makes this safe: `-` is tried first, so every POSIX-encoded folder takes exactly the path it took before and the new branch is only reachable for a single leading letter followed by `--`. Nothing else can start matching by accident. Merged against current `develop` the full suite passes here as well.

Two small things before this goes in:

1. **Please add a `CHANGELOG.md` entry** under `## [Unreleased]` -> `### Bug Fixes`, referencing the issue number. The house style is a bolded one-line summary followed by the mechanism — the surrounding entries are the pattern to copy.
2. **The branch is 4 commits behind `develop`**, which has moved a few times today. A merge or rebase from `develop` will be needed before this can go in.

One observation rather than a change request: the fallback only triggers when `cwd` is missing from the transcript JSONL, so as the issue notes this is the second line of defence rather than the common path. That is the right place to fix it — just worth knowing the blast radius is smaller than the "100% of checkpoints in `wing_sessions`" symptom suggests for sessions that do carry `cwd`.

Nice catch on the test that was encoding the POSIX assumption while being named for Windows.

PR #2315, opened 2026-08-20
feat(logstream): background watcher agents can be woken by, plus the monitoring protocol

igorls:
Coordination friction on the shared brain is usually a *listening* failure, not a protocol failure: a task sits `open` because the agent it was addressed to was never watching, and the requester cannot tell "working on it" from "nobody is home".

Two commits — the protocol, then the tool that makes it practical.

## `logstream watch`

`logstream wait` is a primitive, not a watcher. It caps at `MAX_WAIT_TIMEOUT_MS` and reports a timeout, so every caller writes the same re-arm loop and each one has to remember to carry the cursor forward. Most did not.

Two filters a watcher needs cannot be expressed by `list_events`, whose SQL is single-valued and positive-only:

- "wake me for `task.request` **or** `patch.ready`"
- "everything addressed to me **except** my own events"

The second is not a nicety. `to_agent=<me>` deliberately also matches `*` broadcasts, and an agent's own broadcasts are broadcasts — so a hand-rolled watcher wakes itself every time it posts a status. This was found by running one: it woke on its own fleet announcement within seconds.

```bash
mempalace logstream watch \
  --agent mac-claude \
  --type task.request --type patch.ready \
  --state-file ~/.mempalace/watch/mac-claude.json --json
```

- `--agent ID` is shorthand for `--to-agent ID --exclude-from-agent ID`
- repeat any filter to mean "or" — that is how you get "or nothing", so routine status traffic stops waking you
- `--state-file` resumes exactly; the cursor advances past events examined and rejected, not only matches
- exit `0` on a match, `2` on `--idle-exit-ms` — same convention as `wait`, so a harness can background the process and treat its exit as "you have mail"
- `--follow` stays alive past the first match, for daemons

Single-valued filters push down to SQL to keep the query selective; the multi-valued and negative parts are re-checked client-side. Pushdown is an optimization, never a correctness dependency.

## The cursor rule

Events are ordered by append order (`ORDER BY rowid`), not wall clock. Across replicas those diverge: a peer's event is appended whenever it syncs, so it can already be older than a timestamp high-water mark. **Resuming with `since_created_at` therefore drops late-arriving cross-replica events permanently.**

Measured one such inversion in a single 50-event window: a windows-origin event created `09:10:48Z` was ingested *after* a mac-origin event created `09:13:21Z`.

`list_events`' docstring already said `since_event_id` is "the precise cursor … regardless of timestamp ties" — that just never reached an agent. It now appears in the tool descriptions at the point of use, and both `since_created_at` params are marked "NOT a resume cursor".

## Docs

- `coordination-protocol.md` — new **Monitoring the stream** section: cursor rule, the modes, the announce-your-watch convention, and declaring when you are *not* watching. The **system-prompt snippet** is updated, since that block is copied verbatim into every agent's instructions.
- `website/concepts/agent-logstream.md` and `website/reference/cli.md` — matching sections.

The announce-your-watch shape is generalized from an existing fleet announcement that named its filter, its cursor, and the work not to duplicate.

## Testing

20 new tests: self-exclusion at both matcher and CLI level, multi-valued ORs, cursor advance-past-rejected, resume-without-replay, `--since-event-id` overriding the state file, corrupt state files degrading rather than raising, and the exit-code contract.

Full suite **4390 passed, 31 skipped**; ruff clean. Also exercised end-to-end against a live 165k-drawer palace and a real multi-machine fleet — the watcher is currently running as this repo's own coordination watcher.

### DIFF
--- a/integrations/shared/coordination-protocol.md
+++ b/integrations/shared/coordination-protocol.md
@@ -120,7 +120,10 @@ mempalace logstream watch \
   routine status traffic stops waking you.
 - **`--state-file`** persists the cursor, so a restart resumes exactly where
   it stopped rather than replaying or skipping. It advances past events that
-  were examined and rejected, not only matches.
+  were examined and rejected, not only matches. When the cursor cannot be
+  read, or the watcher first started against an empty log, it replays rather
+  than jumping to the tip — a restart may cost you a duplicate, never a
+  missed delegation.
 - **Exit codes** are the wake signal: `0` when it printed a match, `2` when
   `--idle-exit-ms` expired having seen nothing, `130` when interrupted. Only
   `0` means "you have mail" — an interrupted watcher must never claim it.
--- a/integrations/shared/coordination-protocol.md
+++ b/integrations/shared/coordination-protocol.md
@@ -59,6 +59,100 @@ the event trail is only auditable if identities are stable.
    `type=task.reply` with `status=blocked` or `failed` and verbatim
    notes. Silence is the only unrecoverable failure.
 
+## Monitoring the stream
+
+Most coordination friction is not a protocol failure — it is a *listening*
+failure. A task sits `open` because the agent it was addressed to was never
+watching, and the requester cannot tell the difference between "working on
+it" and "nobody is home". Pick a monitoring mode deliberately and make it
+visible.
+
+### The cursor rule
+
+**Resume with `since_event_id`. Never resume with `since_created_at`.**
+
+Events are ordered by *append* order (rowid), not by wall clock. Across
+replicas those diverge: a peer's event created at 09:10:48Z can be ingested
+*after* a local event created at 09:13:21Z, because it only arrived at sync
+time. A cursor based on `since_created_at` silently skips such an event —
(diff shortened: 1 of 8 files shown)


igorls:
Windows-side check of the Codex P2s, now on the branch as `0d3f216`:

- **Idle vs poll:** confirmed. `--idle-exit-ms 400 --poll-timeout-ms 2500` from the log tip waited 2687ms. After the fix, 572ms. `watch_events` accepts a callable timeout so each poll is capped to remaining idle.
- **Checkpoint before stdout:** agreed. Matched batches now persist the cursor after a successful print; unmatched advances still checkpoint immediately. Regression: `json.dumps` raising `OSError` leaves no state file.
- **preview + since_event_id of the matched event:** agreed, docs bug. Tool descriptions now say to repeat the original filters with `preview=false` — the cursor is strictly after that id.

`pytest tests/test_cli_logstream.py tests/test_logstream.py`: 89 passed on Windows. `--palace` remains a global flag (before `logstream`). First watch without a cursor still replays historical `*` broadcasts, which is spec-correct but a first-run footgun on a long fleet log.

Issue #2320, opened 2026-08-21
bug: sync --apply deletes drawers whose source file it could not reach


mvalentsev:
**What happened?**

`mempalace sync --apply` deletes drawers whose source file it could not reach, not only drawers whose source file was removed.

`_classify_drawer` draws that line with `Path.exists()` (`sync.py:123-124`). `missing` is a removal bucket: `sync.py:276-277` appends those drawer ids to `removable_ids`, `sync.py:293` hands the list to `_delete_in_batches`, `sync.py:192` calls `col.delete(ids=chunk)`, and the matching closets go at `sync.py:311`. Nothing looks at the file again in between.

Measured on `develop` @ 3e56979f, one drawer per state, through the repo's own `_classify_drawer`:

| source file state | `os.stat` | bucket | deleted by `--apply` |
|---|---|---|---|
| present | ok | `kept` | no |
| deleted from the project | `ENOENT` | `missing` | yes, correctly |
| a path component is a regular file | `ENOTDIR` | `missing` | **yes** |
| an empty directory where a volume is not mounted | `ENOENT` | `missing` | **yes** |
| parent is a dangling symlink into the project | `ENOENT` | `missing` | **yes** |
| parent is a dangling symlink out of the project | `ENOENT` | `out_of_scope` | no |
| a symlink loop | `ELOOP` | `resolve()` raises | no |
| a parent the process may not enter | `EACCES` | `exists()` raises | no |

The mount-point row is the one that costs data. A project on an external disk or a network share, mined while it was mounted, leaves an empty mount point when it is not, and one `--apply` removes every drawer filed from it. The operator is shown `Missing: N  (removed)` (`cli.py:1140`), which does not separate a deleted file from an unreachable one.

The last two rows are version-dependent, and they do not turn at the same version. `resolve(strict=False)` raises `RuntimeError` on a symlink loop up to 3.12 and stops raising in 3.13, so the loop row becomes `missing` and is deleted from 3.13 on. `Path.exists()` raises `PermissionError` on `EACCES` up to 3.13 and answers False in 3.14, so the unreadable-parent row is deleted from 3.14 on. Measured on 3.9, 3.10, 3.11, 3.12, 3.13 and 3.14 against this repository's own `_classify_drawer`; `requires-python` is `>=3.9`. On 3.14 six of the eight rows are deleted, five of them wrongly.

[...]

PR #2321, opened 2026-08-21
fix(sync): keep drawers whose source path cannot be resolved instead of pruning

Hughhhhcoder:
## Problem

`sync --apply` used `Path.exists()` to detect deleted source files. `exists()` conflates every stat failure with "gone", so a drawer whose source was merely *unreachable* was pruned as if deleted — and two of those unreachable states crashed the whole sync run instead of being classified. This violates MemPalace's "Incremental only — never destroy existing data" principle.

## Fix

Replace the lossy `exists()` check with a direct `stat()`, classify each failure precisely, and guard path resolution itself:

- `FileNotFoundError` (ENOENT): treated as a true deletion → `missing` (pruned), as before.
- `NotADirectoryError` (ENOTDIR): a path component is a regular file → `unreachable` (kept).
- `PermissionError` (EACCES/EPERM): cannot prove the file is gone → `unreachable` (kept).
- `RuntimeError` from `resolve(strict=False)`: symlink loop (raised up to Python 3.12) → `unreachable` (kept). Previously ended the whole run.
- `ValueError`: unencodable `source_file` (e.g. embedded NUL) → `unreachable` (kept). Previously ended the whole run.
- other `OSError` (ELOOP/EIO/ENAMETOOLONG/...): unverifiable → `unreachable` (kept).

Only `gitignored` and `missing` are pruned on apply. `unreachable` drawers are kept and surfaced in the report output (`cli.py`/`service.py`).

Refs #2320.

## Not fixed here (important)

An unmounted volume also surfaces as `ENOENT` — indistinguishable from a real deletion by any filesystem call, because `os.path.ismount()` returns `False` on an empty mount point. That case (the heart of #2320) is not solved by per-file errno classification and is addressed by the corroboration approach in #2322.

## Test

- `test_unreachable_non_directory_component_kept` (ENOTDIR)
- `test_unreachable_symlink_loop_kept` (ELOOP → `RuntimeError`)
- `test_unreachable_unencodable_path_kept` (embedded NUL → `ValueError`)
- Full `tests/test_sync.py` suite passes (43 tests).

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1138,6 +1138,7 @@ def cmd_sync(args):
     print(f"  Kept:           {report['kept']}")
     print(f"  Gitignored:     {report['gitignored']}  {removed_suffix}")
     print(f"  Missing:        {report['missing']}  {removed_suffix}")
+    print(f"  Unreachable:    {report['unreachable']}  (kept)")
     print(f"  No source:      {report['no_source']}  (kept)")
     print(f"  Out of scope:   {report['out_of_scope']}  (kept)")
 
--- a/mempalace/service.py
+++ b/mempalace/service.py
@@ -364,6 +364,7 @@ def run_sync(payload: dict[str, Any]) -> dict[str, Any]:
     print(f"  Kept:           {report['kept']}")
     print(f"  Gitignored:     {report['gitignored']}  {removed_suffix}")
     print(f"  Missing:        {report['missing']}  {removed_suffix}")
+    print(f"  Unreachable:    {report['unreachable']}  (kept)")
     print(f"  No source:      {report['no_source']}  (kept)")
     print(f"  Out of scope:   {report['out_of_scope']}  (kept)")
 
--- a/mempalace/sync.py
+++ b/mempalace/sync.py
@@ -12,7 +12,6 @@ Usage:
 """
 
 import logging
-import os
 from collections import defaultdict
 from pathlib import Path
 from typing import Callable, Optional, TypedDict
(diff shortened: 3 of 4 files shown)


mvalentsev:
I opened #2320 and had a branch in progress on it, which is #2322. Saying so here rather than letting it be found on review.

This one does not fix the reproduction in #2320. Run against this head with nothing else changed, the five steps from the issue body end where `develop` ends them: `Missing: 20 (removed)`, `Unreachable: 0 (kept)`, `Removed 20 drawers, 2 closets`, `embeddings` down to 0, `No results found` once the files come back.

`_looks_unmounted` walks up to the first ancestor that exists and asks `os.path.ismount` about it, but a mount point with nothing mounted on it is an ordinary directory, so `ismount` answers False there. The predicate fires in the opposite case from the one it is named for: True when the volume **is** mounted and the path below it is gone. Driven through `sync_palace` with a real `mount` and `umount`, an unmounted volume is still pruned whether its mount point is empty or holds a `.gitkeep`, and on a mounted one a file or a subdirectory that really was deleted is now kept instead.

Two of the states the description lists as handled never reach the new `except` clauses. `resolve(strict=False)` raises `RuntimeError` on a symlink loop up to Python 3.12, before the `try` around `stat`, so ELOOP still ends the whole run; an unencodable `source_file` raises `ValueError`, which is not an `OSError`, so that ends it too. ENOTDIR and EACCES are genuinely fixed, and `test_unreachable_non_directory_component_kept` pins the first.

The trap underneath is that no filesystem call separates "nothing is here" from "this cannot be reached right now" for a path, and `ismount` is one more call of that kind.

Hughhhhcoder:
Thanks for the thorough review — every point is correct, and I've pushed a fix for the parts of it this PR can address.

**What I changed:**

1. **Removed `_looks_unmounted`.** You're right that it was inverted: `os.path.ismount()` returns `False` on an empty mount point, so it never fired for an unmounted volume, and instead kept genuinely-deleted files under a live mount. The heuristic can't be salvaged — an unmounted mount point is indistinguishable from any ordinary directory.

2. **`resolve(strict=False)` → `RuntimeError` (ELOOP ≤3.12).** Now guarded; a symlink loop classifies as `unreachable` (kept) instead of ending the whole run.

3. **`ValueError` on an unencodable `source_file`.** Now caught (in both `resolve` and `stat`), classified `unreachable` instead of ending the run.

Added `test_unreachable_symlink_loop_kept` and `test_unreachable_unencodable_path_kept` to pin those two; full `tests/test_sync.py` is 43 passing.

**What I did not fix, and agree with you on:**

The mount-point case is the heart of #2320, and you're right that no filesystem call separates "nothing is here" from "this cannot be reached right now" — `ENOENT` on an unmounted volume is the same errno as a deletion. Per-file errno classification cannot solve it. I've retitled this PR to *Refs #2320* and narrowed its claim: it now only stops sync from pruning (or crashing on) sources whose path *cannot be resolved* (ENOTDIR/EACCES/ELOOP/unencodable), and leaves the `ENOENT` unmounted-volume case to the corroboration approach in #2322, which is the right tool for it.

Happy to adjust further — and if the maintainers would rather take #2322 whole and drop this one, that's fine by me.

Hughhhhcoder:
The latest `Tests` and `Docker` pull-request workflow runs are completed with `action_required`. Could a maintainer approve the workflows when convenient so the checks can run? The addressed review fixes are already pushed and covered by the local test suite.

igorls:

PR #2341, opened 2026-08-23
feat(search): automatic cross-wing search expansion

jjcav84:
feat(search): automatic cross-wing search expansion

When no wing/room/source_file filter is given, search_memories,
mempalace search, and the mempalace_search MCP tool now automatically
expand the query to the most relevant wings using existing palace structure:
- Tunnel rooms (rooms that appear in multiple wings)
- Hallways (entity co-occurrence within wings)
- Room names within wings

This fixes the problem where an LLM searching one wing misses context in
related wings, without writing synthetic triples to the KG or adding a new
storage layer. The wing scoring is query-time only and uses existing
tunnels.json, hallways.json, and room metadata.

New files:
- mempalace/wing_affinity.py — wing relevance scoring
- tests/test_wing_affinity.py — unit tests

Changed:
- mempalace/searcher.py — _maybe_expand_wings, build_where_filter list,
  search_memories + search expansion
- mempalace/mcp_server.py — tool_search accepts expand_wings (default true)
- mempalace/cli.py — --expand-wings / --no-expand-wings
- tests/test_cli.py — updated search call assertions
- website/reference/mcp-tools.md — updated mempalace_search docs

Verification:
- Full test suite: 4478 passed, 31 skipped
- ruff clean

Co-Authored-By: Devin <158243242+devin-ai-integration[bot]@users.noreply.github.com>

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -1575,6 +1575,7 @@ def cmd_search(args):
             n_results=args.results,
             since=args.since,
             before=args.before,
+            expand_wings=getattr(args, "expand_wings", True),
         )
     except SearchError:
         sys.exit(1)
@@ -3407,6 +3408,12 @@ def main():
     p_search.add_argument("--wing", default=None, help="Limit to one project")
     p_search.add_argument("--room", default=None, help="Limit to one room")
     p_search.add_argument("--results", type=int, default=5, help="Number of results")
+    p_search.add_argument(
+        "--expand-wings",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help="When no --wing is given, automatically search the most relevant wings (default: true)",
+    )
     p_search.add_argument(
         "--since",
         default=None,
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -2477,6 +2477,7 @@ def tool_search(
     context: str = None,
     candidate_strategy: str = "vector",
     cli_compatible: bool = False,
+    expand_wings: bool = True,
 ):
     limit = max(1, min(limit, _MAX_RESULTS))
     try:
@@ -2580,6 +2581,7 @@ def tool_search(
         vector_disabled=_vector_disabled,
         candidate_strategy=candidate_strategy,
         collection_name=_config.collection_name,
+        expand_wings=expand_wings,
     )
     if _is_transient_index_error(result):
         # Post-bulk-write HNSW flush window (#1315): drop caches, give
@@ -2601,6 +2603,7 @@ def tool_search(
             vector_disabled=_vector_disabled,
             candidate_strategy=candidate_strategy,
             collection_name=_config.collection_name,
+            expand_wings=expand_wings,
         )
         if not _is_transient_index_error(result):
             result["index_recovered"] = True
@@ -5368,12 +5371,11 @@ TOOLS = {
         "handler": tool_follow_tunnels,
     },
     "mempalace_search": {
-        "description": "Semantic search. Returns verbatim drawer content with similarity scores. IMPORTANT: 'query' must contain ONLY search keywords. Use 'context' for background. Results with cosine distance > max_distance are filtered out.",
+        "description": "Semantic search. When no wing is given, automatically expands to the most relevant wings via graph tunnels and hallways (set expand_wings=false to disable). Returns verbatim drawer content with similarity scores. IMPORTANT: 'query' must contain ONLY search keywords. Use 'context' for background. Results with cosine distance > max_distance are filtered out.",
         "input_schema": {
             "type": "object",
             "properties": {
                 "query": {
-                    "type": "string",
                     "description": "Short search query ONLY — keywords or a question. Max 250 chars.",
                     "maxLength": 250,
                 },
@@ -5426,6 +5428,10 @@ TOOLS = {
                 "context": {
                     "type": "string",
                     "description": "Background context for the search (optional). NOT used for embedding — only for future re-ranking.",
+                },
+                "expand_wings": {
+                    "type": "boolean",
+                    "description": "When true (default) and no wing is specified, automatically search the most relevant wings based on palace graph structure.",
                 },
             },
             "required": ["query"],
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -395,7 +395,7 @@ def _maybe_expand_wings(
     wing: str = None,
     room: str = None,
     source_file: str = None,
-) -> list[str] | None:
+) -> Optional[list[str]]:
     """Compute which wings to search when no explicit filter is given.
 
     Returns ``None`` if expansion is disabled, the caller specified a
@@ -1693,7 +1693,7 @@ def _search_result_envelope(
     candidates_fetched: int,
     pool_size: int,
     date_window_active: bool,
-    expanded_wings: list[str] | None = None,
+    expanded_wings: Optional[list[str]] = None,
 ) -> dict:
     """Assemble the ``search_memories`` response dict.
 
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -388,15 +388,56 @@ def _hybrid_rank(
     return results
 
 
-def build_where_filter(wing: str = None, room: str = None, source_file: str = None) -> dict:
+def _maybe_expand_wings(
+    query: str,
+    palace_path: str,
+    expand_wings: bool,
+    wing: str | None,
+    room: str | None,
+    source_file: str | None,
+) -> list[str] | None:
+    """Compute which wings to search when no explicit filter is given.
+
+    Returns ``None`` if expansion is disabled, the caller specified a
+    filter, or the affinity scorer could not find relevant wings. The
+    caller falls back to unfiltered search in all three cases.
+    """
+    if not expand_wings:
+        return None
+    if wing or room or source_file:
+        return None
+    try:
+        from .wing_affinity import expand_wings as _expand_wings
+
+        return _expand_wings(query, config=MempalaceConfig(palace_path=palace_path))
+    except Exception:
+        logger.debug("wing expansion failed; falling back to unfiltered", exc_info=True)
+        return None
+
+
+def build_where_filter(
+    wing: str = None,
+    room: str = None,
+    source_file: str = None,
+    wings: list[str] = None,
+) -> dict:
     """Build a ChromaDB where filter from optional wing/room/source_file.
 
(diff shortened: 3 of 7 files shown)


jjcav84:
Rebased onto latest `develop`. Resolved conflicts in `searcher.py` (upstream `collection=` parameter) and `mcp_server.py` (upstream `candidate_strategy`/`cli_compatible` parameters). All 110 search and wing-affinity tests pass. Ruff clean. Ready for review.

jjcav84:
Hi @fatkobra @igorls @milla-jovovich — gentle ping for review. This PR adds automatic cross-wing search expansion to the searcher — when a query returns few results, the search automatically expands to adjacent wings via palace graph tunnels. Rebased on latest `develop`, all 110 search and wing-affinity tests pass, ruff clean. All CI green.

fatkobra (changes requested):
Thanks for the ping. I reviewed the current head, `2e57fab`. The structural-affinity idea is useful, but I am requesting changes because the current implementation can reduce search recall rather than expand it.

1. This is a pre-query hard filter, not thin-result expansion.

`_maybe_expand_wings()` runs before the first query, selects at most three wings, and passes them into the storage `where` filter. It does not inspect the number or quality of the normal search results, and supplying an explicit wing disables it entirely.

A weak room/entity-name match can therefore exclude the wing containing the best semantic result that the previous palace-wide search would have returned. This is especially risky because CLI and MCP enable the feature by default.

Please either:

- preserve the baseline search and add results from structurally related wings only when a defined thin/low-quality condition is met, then merge, deduplicate, and rerank; or
- describe this as affinity-based pre-query routing, make it opt-in, and demonstrate that its recall trade-off is acceptable.

Please add a regression where wings A/B/C score structurally but the best semantic drawer is in wing D. Enabling “expansion” must not remove D from the results.

2. Affinity is not computed for the exact search target.

`search_memories()` can open an explicit `collection_name`, but `_maybe_expand_wings()` receives only `palace_path` and reconstructs configuration independently. `build_graph()` also has a warm process-global cache that ignores its collection/configuration arguments.

A custom collection—or a second palace searched in the same process—can therefore be filtered using wings from a different target.

Please derive affinity from the already-opened collection, or pass and cache the complete palace/backend/collection identity. Add regressions for two collections with disjoint wings and two sequential palace searches in one process.

3. Tunnel and retrieval semantics are inconsistent.

`wing_affinity.py` says it reads `tunnels.json`, but `find_tunnels()` returns passive same-room connections from collection metadata. Explicit and topic tunnels stored in `tunnels.json` are not used.

Also, `candidate_strategy="union"` adds lexical candidates without the expanded-wing filter, while the vector-disabled fallback bypasses expansion entirely. The returned `wing_expansion` envelope can therefore describe a scope that was not applied consistently.

Please either implement explicit-tunnel support and propagate the same scope through every candidate path, or narrow the feature documentation accordingly.

Finally, the current `search_memories` regression exits with the mocked “no palace” error before expansion is reached, so it does not test the integration. Please add end-to-end result assertions covering recall preservation, target isolation, explicit tunnels, union candidates, and fallback behavior.

igorls:

jjcav84:
Reworked per @fatkobra's review — the pre-query wing filter is gone. Expansion is now additive and post-baseline:

**Design change**
- The unfiltered palace-wide baseline always runs first. When it is thin (fewer hits than `n_results`, or top `effective_distance` > 1.0), `wing_affinity.expand_wings` scores wings and each scored wing gets its own scoped `search_memories` query; hits merge, dedupe (drawer_id + text), rerank, and cut to `n_results`. Baseline hits are never removed.
- `wing_expansion` appears in the envelope only when expansion fired: `{applied, wings, added}`.
- `expand_wings=True` default on `mempalace_search`, `search_memories`, and CLI `search`; `--no-expand-wings` CLI flag; hub forwarding propagates the flag.

**Reviewer points**
- *Recall*: `test_baseline_hit_never_removed_by_expansion` — wings alpha/beta score structurally while the best semantic drawer sits in wing delta; the delta hit survives and every baseline hit remains (`baseline ⊆ merged`).
- *Target isolation*: `MempalaceConfig` gained a `collection_name` ctor kwarg; affinity binds to the opened collection + bound config. `build_graph`'s warm cache is now keyed on `(palace_path, collection_name)` and caller-supplied collections bypass it entirely (fixes the path where a col-built graph could be cached under a stale key).
- *Tunnel semantics*: narrowed — expansion uses passive same-room connections only; `tunnels.json` records are not consulted (`test_explicit_tunnels_json_not_read` pins this).
- *Path parity*: union candidates and the vector-disabled BM25 fallback both expand under the same scope (each scoped query runs the full pipeline with `expand_wings=False`).

**Tests** — `tests/test_wing_affinity.py`: 15 tests covering scoring signals, recall preservation, opt-out, filtered-search skip, healthy-baseline skip, union, BM25 fallback, collection isolation, sequential-palace isolation, and error fallback. `test_palace_graph.py` cache test updated to the new identity-keyed contract; hub-forward test updated for the forwarded flag.

**Validation** — ruff + ruff-format clean; `tests/test_wing_affinity.py` 15/15; full suite 5396 passed. Two benchmark failures reproduce on clean `develop` and are unrelated: `test_schema_reduction_benchmark` (68.6% < 70% on develop; this branch reads 68.8%) and `test_cli_import_under_startup_budget` (140ms on develop, 107ms here — host is loaded).

igorls (changes requested):
Thanks for the rework. Moving expansion after the baseline and binding affinity to the opened collection is the right direction, and narrowing the tunnels scope is fine. I'm still requesting changes, because the core promise doesn't hold yet.

1. **Baseline hits are dropped.** `_expand_result_dict` re-sorts the merged pool by `effective_distance` and cuts to `n_results` (query.py:201-202), but the baseline is ordered by `_hybrid_rank`.
   - Repro: a drawer that is the only literal match for the query (dist 0.786, bm25 1.397) is baseline #1. With expansion it disappears at n=1 and n=2, and drops to #3 at n=3.
   - Even with `added: 0`, the re-sort still reorders the baseline.
   - The CLI path is worse. It sorts by raw `distance` (cli_search.py:310), which undoes the hybrid fix for far-vector exact matches.

   Baseline hits must keep their slots and their order; expansion hits should only fill empty slots after them. Please also fix `added` to count only hits that survive the cut, and add a regression where an added hit scores *better* than a baseline hit.
2. **Union crashes.** When `max_distance=0`, BM25-only union candidates have `distance=None`. `_baseline_is_thin` then evaluates `top.get("distance", 9.9) > 1.0` and raises `TypeError`, and the merge sort key has the same problem. Reproduced with `candidate_strategy="union"`.
3. **Cost.** Passing `drawers_col` makes `build_graph` skip both the sqlite fast path and the cache, and `score_wings` calls it twice. That's two full metadata scans per thin search, with expansion on by default in MCP, CLI and Hermes prefetch. The CLI does it twice more, because it scores wings itself and then calls `search_memories`, which scores again. This change also removes caching for `traverse`/`find_tunnels` on non-sqlite backends. Please use the config-keyed sqlite path, score once, and include numbers on a large palace.
4. **Scores aren't comparable across scopes.** Closet boosts are based on rank within the scope, and BM25 IDF is computed per pool, so the merged sort compares different scales. Also, a scoped query runs the same query over a subset of the same corpus, so it rarely finds anything the baseline didn't already rank. Where does expansion actually add recall (date-window pool truncation?)? Please show a case, or widen the pool instead.
5. **The CLI ignores the caller's collection.** CLI expansion doesn't pass the caller's `collection` (cli_search.py:14, :274).

Given the above, please make expansion opt-in (default `False`) until a benchmark shows a recall gain with no regressions. Please also drop the "baseline hits are never removed" wording from the schema, the CLI help and the docs until it's true.

jjcav84:
@igorls @fatkobra — reworked at `b35adfc`. Expansion is now strictly additive slot-filling and opt-in everywhere.

**1. Baseline hits keep their slots and order — literally.** `_expand_result_dict` no longer re-sorts the merged pool (that re-sort is exactly what evicted your literal-match drawer at n=1/n=2 and demoted it at n=3, and it's also why the merged pool compared incompatible score scales — point 4 resolved by not comparing them). Expansion hits dedupe on `drawer_id`/text and append after the baseline up to `n_results`. `added` counts only hits that survive the cut; `applied = added > 0`. Regressions added: a better-scoring added hit still appends after the baseline; `added` counts survivors only; a full baseline gains nothing.

The CLI path is fixed the same way — no more raw-`distance` re-sort undoing `_hybrid_rank`; expansion appends into empty slots only.

**2. Union crash gone.** `distance=None` can no longer reach a comparison — the thin gate is now count-only (see below), and there is no merge sort key anymore. The `union` + `max_distance=0` case is covered by an end-to-end regression.

**3. Cost.** `query.py` no longer passes `col` to the affinity scorer — `build_graph` takes the config-keyed path: sqlite grouped-counts fast path + the `(palace_path, collection_name)`-keyed warm cache, so `find_tunnels` and `build_graph` inside `score_wings` share one read. The CLI no longer scores wings itself; it delegates to `search_memories`, so scoring happens exactly once. The `col`-bypass in `build_graph` remains for isolation-critical callers, and non-sqlite backends still populate the warm cache from the metadata scan.

**4. Where expansion adds recall — honestly answered.** Under strict slot-filling it can only add hits when the baseline is count-thin: post-filtered pools (date-window/distance truncation) or corpora smaller than `n_results`. On a full baseline it appends nothing — so the quality-based "weak top hit" trigger was removed entirely (it burned wing-scoring + 3 scoped queries for zero possible gain). The recall case is the date-window pool truncation you hypothesized: a wing-scoped query's candidate pool isn't truncated the way the global pool was, so in-window drawers the global pool cut can surface. `expand_wings` is now **opt-in (default False)** in `search_memories`, the MCP tool schema, `--expand-wings` (replaces `--no-expand-wings`), and the hub forwarder — matching your ask pending a real recall benchmark. Schema/CLI/docs wording no longer overpromises.

**5. CLI collection fixed** — `search()` passes the caller's collection name through to `search_memories`; the BM25-only fallback's `_expand_wing_list` binds `collection_name` too.

Tests: 200 passing across `test_wing_affinity.py`, `test_cli.py`, `test_hub_forward.py`, including the fatkobra regression (structural wings A/B/C score, best semantic drawer in D, D is never dropped) and the collection/sequential-palace isolation tests. `ruff check` + `ruff format` clean.

One pre-existing failure unrelated to this PR: `tests/benchmarks/test_mcp_light_bench.py::test_schema_reduction_benchmark` fails on the base commit too (68.8% vs the 70% assert) — the light-vs-legacy schema ratio drifted before this branch. Happy to address separately.

PR #2351, opened 2026-08-24
fix(miner): remine files when content changes

masalu0105-gif:
## Summary
- stop treating a previously seen path as permanently mined
- compare mtime and SHA-256, then re-file edited or growing sources
- backfill legacy drawer metadata without re-embedding unchanged files
- retain UTF-8 config loading and silence broken Chroma telemetry noise

## Verification
- new file-state unit suite: 6 passed
- remaining non-Chroma integration suite: 94 passed
- Ruff: passed
- the two existing Chroma integration tests execute successfully but fail only while deleting open database files on Windows (`WinError 32`); this PR does not change that existing cleanup behavior

### DIFF
--- a/mempalace/__init__.py
+++ b/mempalace/__init__.py
@@ -14,7 +14,7 @@ _os.environ.setdefault("ANONYMIZED_TELEMETRY", "False")
 _logging.getLogger("chromadb.telemetry").setLevel(_logging.CRITICAL)
 # ---------------------------------------------------------------------------
 
-from .cli import main
-from .version import __version__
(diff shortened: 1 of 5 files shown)


masalu0105-gif:
Closing as superseded after refreshing against current upstream. `main` now has a substantially newer mutable-source design: mtime-aware mining, conversation content hashes, safe re-mine handling, incomplete-batch detection, and extensive regressions. Rebasing this old 3.0-era implementation would duplicate weaker logic and conflict with the current architecture.

PR #2353, opened 2026-08-24
fix(config): read and write project YAML as UTF-8

masalu0105-gif:
## Summary

- Read mempalace.yaml with explicit UTF-8 in the miner.
- Write generated project configuration with explicit UTF-8.
- Add focused Windows/CJK regression tests for load and round-trip behavior.

## Validation

- .venv/Scripts/python.exe -m pytest --noconftest tests/test_config_encoding.py -q
- 3 passed
- git diff --check passed

## Note

The repository-wide local fixture currently depends on a newer Chroma API than the checked-out virtual environment. The encoding tests therefore use pytest built-ins and were run without the unrelated repository conftest.

### DIFF
--- a/mempalace/miner.py
+++ b/mempalace/miner.py
@@ -549,7 +549,7 @@ def load_config(project_dir: str) -> dict:
                     }
                 ],
             }
-    with open(config_path) as f:
+    with open(config_path, encoding="utf-8") as f:
         return yaml.safe_load(f)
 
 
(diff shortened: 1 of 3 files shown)


igorls:
Thanks for the fix! Setting explicit UTF-8 encoding for project YAML config load/save is great for cross-platform compatibility on Windows.

Per our contributing guidelines, feature and bugfix PRs should target the `develop` branch instead of `main` (`main` is reserved for tagged releases). Could you please update the base branch of this PR to `develop`?

PR #2357, opened 2026-08-24
fix(config): stop writing over a config.json that was never read (#2356)

### LINKED ISSUE #2356: mempalace init --backend overwrites a config.json it could not read, losing every setting in it
**What happened?**

Editing `~/.mempalace/config.json` by hand and getting one character wrong costs every setting in it, at the next command that changes a setting.

`MempalaceConfig.__init__` reads the file behind `Path.exists()` and folds every failure into an empty dict (`config.py:375-380` on `develop`):

```python
if self._config_file.exists():
    try:
        with open(self._config_file, "r", encoding="utf-8") as f:
            self._file_config = json.load(f)
    except (json.JSONDecodeError, UnicodeDecodeError, OSError):
        self._file_config = {}
```

All four setters then serialize that dict back over the whole file, opening it with mode `"w"`: `set_entity_languages` (`:719`), `set_embedding_model` (`:812`), `set_backend` (`:830`), `set_hook_setting` (`:1103`). `mempalace init --backend` reaches one of them and `--lang` a second, on the global config, whatever directory the command is run in.

Measured through the CLI on a 219-byte config holding `palace_path`, `collection_name`, `backend`, `embedding_model`, a three-entry `people_map` and `max_backups`. One `mempalace init <another-project> --yes --no-llm --backend chroma` afterwards:

| what was wrong with the config | what was left |
|---|---|
| one trailing comma before the closing brace | 25 bytes, `{"backend": "chroma"}` |
| a UTF-8 byte-order mark, which an editor can add on its own | 25 bytes, `{"backend": "chroma"}` |

Nothing in the output mentioned it, and no copy was left in the directory.

What that costs: `people_map` is only ever written by hand, since nothing in the package calls `save_people_map`, so those mappings are gone. `palace_path` falls back to the default, so the palace the user configured is not the one the tools read any more. `embedding_model` falls back to `minilm`, and the docstring at `config.py:746` says switching models on an existing palace needs re-embedding because ChromaDB rejects reads when the persisted EF name does not match, so a palace mined with `embeddinggemma` stops answering until `mempalace repair rebuild-index` is run.

Hand-editing is how the file is meant to be changed: `website/guide/configuration.md` prints its JSON and a table of its keys, and `people_map` has no other way in.

The setters' own writes are the second source of files that do not parse. On a 3,617,966-byte config one `set_hook_setting` held the file open for 357 ms, and a reader in another thread got a file that would not parse twice against three complete reads. `SIGKILL` inside that window left a truncated config in 10 runs out of 10, from 508,435 to 1,770,851 bytes. A process started on one of those answers the default `palace_path` and 0 `people_map` entries, and one `set_backend("chroma")` in it leaves the same 25 bytes.

A one-key config is not by itself a sign of this: a first `mempalace init` on a machine with no config leaves exactly that file, for its own reason.

A second shape loses nothing and says nothing either. With the config present and at mode `000` the read fails, `set_backend("qdrant")` returns without raising, and the file is unchanged: the `except OSError: pass` in each setter swallows the write failure, so the command reports success for a setting it did not save.

**What did you expect?**

`FileNotFoundError` is the one outcome that establishes there is no config here, because nothing resolved under that name. Everything else says this process did not read it, which is a reason to keep it: rename it aside and say where it went, or leave it alone and say so, but do not let defaults become the file's new contents.

And a write that replaces a file wholesale belongs behind a temporary file and a rename. The package does that in eight other places, and `hallways.py:125` states the reason in a comment.

**How to reproduce:**

1. `mempalace init <project> --backend chroma`, then edit `~/.mempalace/config.json`: point `palace_path` somewhere of your own and add a `people_map` entry.
2. Leave one trailing comma before the closing brace, or save the file with a UTF-8 byte-order mark.
3. `mempalace init <another-project> --yes --no-llm --backend chroma`.
4. Read the file. It is 25 bytes holding the one key that run set, and `mempalace status` now points at the default palace path.

For the second shape, `chmod 000 ~/.mempalace/config.json` before step 3 instead: the command exits 0 and the backend is not recorded.

**Environment:**
- OS: Linux (WSL2)
- Python version: 3.12.3
- MemPal version: `develop` @ 4bc0c431

**Notes**

`mempalace init` reaches two of these setters, `set_entity_languages` under `--lang` (`cli.py:316`) and `set_backend` under `--backend` (`:477`); onboarding reaches a third (`onboarding.py:407`, `:518`), and the MCP hook-settings tool the fourth (`mcp_server.py:4239`, `:4242`). `cfg.init()` itself writes only when the file does not exist. `save_people_map` (`:1146`) writes its own file the same truncating way.

#2234 reports the crash a non-object `config.json` raises, addressed by #2235; this report is about a file that does not parse at all and about what the setters then write over it.

This is the overwrite half of the shape #2320 reports for deletions, and that report names this file as one of the two places outside its scope. The other, `miner.add_to_known_entities`, is filed separately.

Four more places fold a read that did not conclude into an empty default and write that default back over the same file: `palace_graph._load_tunnels` (`:490-499`), `hallways._load_hallways` (`:98-104`), `EntityRegistry.load` (`entity_registry.py:308-316`) and `write_embedder_sidecar` (`backends/_sidecar.py:53-68`), whose docstring says it preserves the other collections' entries. The first two are reached by a plain `mempalace mine` and both write atomically, so what they drop is dropped durably. Each is its own report, none filed yet.

`CLAUDE.md` states the rule this breaks: "Never destroy existing data to rebuild. A crash mid-operation must leave the existing palace untouched."

mvalentsev:
Fixes #2356.

## What does this PR do?

`MempalaceConfig` read `config.json` behind `Path.exists()` and folded every failure into an empty dict, and each setter then serialized that dict back over the whole file with `open(..., "w")`. A config hand-edited into something that does not parse, one trailing comma or a byte-order mark an editor added, was enough to lose every setting in it at the next `mempalace init --backend`, which reaches one of those setters on the global config whatever directory it runs in. The setters' own writes are the other source of such files: on a 3,617,966-byte config one `set_hook_setting` is a window in which the file on disk is not the config, and ten of ten runs killed inside it left truncated remains.

Reading now distinguishes four states rather than two. `FileNotFoundError` is the only one that establishes there is no config here, since nothing resolved under that name. A file that reads but is not a JSON object is renamed aside before a fresh config is written, its new name printed, so its bytes stay recoverable. A file that exists and cannot be read at all is not touched, and the setter says why it declined, since writing this session's defaults there is how a permission bit becomes a lost config. A UTF-8 byte-order mark stopped counting as a parse failure, since `json.loads` accepts one on bytes.

Writing goes through a temporary file, an `fsync`, a rename, and an `fsync` of the directory, which is what makes the rename survive a crash and what `EntityRegistry.save` already does (`entity_registry.py:350-363`). The temporary file carries this process's pid and is opened `O_NOFOLLOW`. A name the open itself refuses, a directory dropped at it or an orphan this user cannot write, sends the write to a name the directory picks: what the directory says about a name it chose is about the directory, what it says about this one is not.

Two of `develop`'s behaviours are kept on purpose. A config reached through a symlink is written through it, since renaming over the link would leave the real file, in a dotfiles checkout, holding what it held. And a directory that will not take a name it chose itself gets the write in place, with a message saying the crash-safety is missing: losing the setting outright is the worse half of that trade.

A write that fails now prints why instead of being swallowed by `except OSError: pass`; the exit status is unchanged, since `cli.py` does not read a setter's result. `save_people_map` writes its own file the same way, so `people_map.json` holds non-ASCII names as UTF-8 rather than `\uXXXX` escapes.

#2234 reports the `AttributeError` a non-object `config.json` raises, and #2235 fixes it with an `isinstance` check in these same lines. The read replaced here reaches it from the other side: `null`, `[]`, `"oops"`, `42` and `true` answer defaults rather than raising, and the file is kept aside under a printed name. The two change the same hunk, so whichever lands first the other needs a rebase, and I will take that rebase.

### What this does not do

It does not recover a config that was already truncated: the bytes are kept as `config.json.unreadable-<stamp>-<suffix>` with the permissions the original had, and nothing prunes them. A hard link to the config is lost, since the rename gives the name a new inode. Two processes that both read and then both write still lose the earlier one's change, as on `develop`.

Interrupted or refused, the write leaves a file behind: `config.json.tmp-<pid>`, or `.config.json.<random>.tmp` when something was already at the pid name, and nothing here removes either. A directory that refuses the rename leaves one on every call rather than on an interrupted one, since removing it needs that same directory; the message names it, and it holds what the call was asked to save. A directory that takes neither gets the write in place, which truncates first, so an interrupted one leaves a config that does not parse where `develop` wrote over the damaged file instead.

It touches nothing outside `config.json` and `people_map.json`. The same shape sits in four more places the issue names, each its own report.

## How to test

```bash
uv run pytest tests/test_config_unreadable_rewrite.py -q
uv run pytest tests/test_config.py tests/test_config_extra.py tests/test_config_palace_path.py -q
uv run pytest tests/test_onboarding.py tests/test_write_routing.py tests/test_encoding_hardening.py -q
```

## Verification

Measured against `develop` and this branch, with real `MempalaceConfig` instances and real child processes, on Python 3.12.3. Counts are one run each.

| claim | how |
|---|---|
| a config that does not parse costs every setting in it, and the bytes are kept here | 219-byte hand-edited config, then `mempalace init <another-project> --yes --no-llm --backend chroma`: one trailing comma leaves 25 bytes on `develop`, and here 25 bytes alongside a 219-byte `config.json.unreadable-<stamp>` whose name is printed; a UTF-8 byte-order mark leaves 25 bytes on `develop`, while here every setting survives, the file rewritten to 233 bytes with all six keys |
| the setter's write is itself a window, and what it leaves is permanent | 3,617,966-byte config, a reader thread through one `set_hook_setting`: `develop` 2 reads that would not parse against 3 that would, this branch 0 against 4. `SIGKILL` inside the write, 10 runs: `develop` left a truncated config every time, 508,435 to 1,770,851 bytes, this branch left the previous config byte for byte every time |
| a config that cannot be read is left alone, and a failed write is no longer silent | config mode `000`: `develop` answers `set_backend` without raising and without writing, so the setting is silently lost, while this branch leaves the file alone and prints why. Directory mode `500`, config `400`: the `OSError` is printed, the config unchanged. Directory `500`, config writable: the setting is saved in place, with the message naming what it gave up |
| an orphan at the temporary name does not cost the rename | the pid-named file left at mode `400`: the write goes through a name the directory picks, the config is updated and the orphan is untouched |
| the tests pin it | 36 mutations one at a time on a copy of the tree, all 36 failing the suite and none by a syntax error: among them an unreadable file counted as absent, the quarantine skipped, the write going straight into the file, the temporary file published without an `fsync`, and the errno of the pid-named file deciding for the directory |

Thirty-four new tests in `tests/test_config_unreadable_rewrite.py`; twenty-four fail against `develop`'s `config.py`, and the ten that do not cover cases `develop` cannot reach, since it writes no temporary file at all. `4526 passed, 31 skipped` on the full suite with `--ignore=tests/benchmarks`, as CI runs it, and nothing failing. `ruff check` and `ruff format --check` clean, `mypy mempalace/config.py` reports the same zero errors it does on `develop`.

## Checklist
- [x] No hardcoded paths


igorls (commented):
### Review: LGTM ✅

Excellent fix for #2356.

**Key Highlights:**
- Prevents configuration clobbering by tracking read/decode error state on `config.json` rather than falling back to an empty dictionary and blindly rewriting it on setter calls.
- Safe atomic file replacement with proper umask and permissions preservation.
- Full test coverage in `tests/test_config_unreadable_rewrite.py` passing across all platforms in CI.

Looks great and ready to be marked as **Ready for review** and merged.

PR #2359, opened 2026-08-24
fix(miner): keep a known_entities.json the merge could not read (#2358)

### LINKED ISSUE #2358: mempalace init replaces a known_entities.json it could not read, and ends in a traceback when it is not UTF-8
**What happened?**

`mempalace init` replaces a `known_entities.json` it could not read with the entities of that one run and reports it as an update. When the file is not valid UTF-8 it does not get that far: the run ends in a traceback.

`add_to_known_entities` reads the registry behind `Path.exists()` and falls back to an empty dict (`miner.py:917-924` on `develop`), then writes the merged dict over the whole file at `:980`:

```python
existing: dict = {}
if registry_path.exists():
    try:
        loaded = _json.loads(registry_path.read_text(encoding="utf-8"))
        if isinstance(loaded, dict):
            existing = loaded
    except (_json.JSONDecodeError, OSError):
        existing = {}
```

Nothing between the fallback and the write asks whether the registry was read. `cli.py:460` is the only caller, and it prints `Registry updated: <path>` afterwards.

Measured through the CLI on a project with a git history and a `pyproject.toml`, so `init` has entities to record. For scale: a clean `init` on a shallow clone of this repository, 126 commit authors, writes a 366-byte registry holding `people` (15 names, the cap) and `projects` (1).

| registry before | what one `mempalace init <project> --yes --no-llm` did |
|---|---|
| a JSON array, 28 bytes | replaced by this run's entities, `Registry updated` printed |
| a truncated object, 172 bytes | replaced by this run's entities, `Registry updated` printed |
| valid JSON saved as cp1251, 118 bytes | `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcc in position 21`, and the run stops there, leaving no `config.json` behind either |

The first two lose every name and wing the registry held, for every project it was built from, and say the opposite. The third is not a data loss but a command that does not finish: `UnicodeDecodeError` is neither `JSONDecodeError` nor `OSError`, so it passes the `except` and out through `cli.py`. `config.py:379` catches that exception explicitly, in the same package, for the same reason.

None of the three needs an interrupted write. The registry is not documented anywhere, but it sits next to the `config.json` that is, and it holds names a scan chose rather than names anyone typed: 15 of those 126 authors, bots and duplicates included. Editing it by hand is how they get cleaned up, and one editor saving it in the machine's legacy encoding is the cp1251 row. The write is not atomic either, since `write_text` builds the document in memory and then truncates the file, so an interrupted run or a full disk leaves the same kinds of file.

Widening that `except` to catch `UnicodeDecodeError` is not the fix it looks like. Measured on a 94-byte registry, it stops the traceback and turns the run into a silent replacement: the file becomes 38 bytes holding that run's one name, and nothing keeps what was there. The crash at least leaves the bytes alone.

**What did you expect?**

`FileNotFoundError` is the one outcome that establishes there is no registry here. Everything else says this process did not read the one that is there, which is a reason to keep it rather than to write over it: rename it aside and print where it went, or leave it alone and say so. A file that is not valid UTF-8 belongs in the same branch rather than in a traceback.

And this file already has an atomic writer in this package. `migrate._apply_topics_by_wing_renames` writes `known_entities.json` through `tempfile.mkstemp` and `os.replace` (`migrate.py:529-534`), and seven other state files are published by a rename as well: `hallways.py:146`, `entity_registry.py:339`, `palace_graph.py:546`, `repair.py:114`, `replica.py:63`, `server_registry.py:135` and `logstream.py:398`. The registry's own merge path is the one that truncates.

Note that two existing tests pin the current behaviour rather than an accident: `test_malformed_existing_registry_starts_fresh` and `test_non_dict_existing_registry_starts_fresh` (`tests/test_known_entities_registry.py:127` and `:134`) assert that the file is replaced. Keeping the old bytes beside the new registry satisfies both as written, since what the new registry holds does not change.

**How to reproduce:**

1. `mempalace init <project-a>` and confirm some entities, so `~/.mempalace/known_entities.json` holds them.
2. Make it unreadable the way an editor would: `printf '["a","b"]' > ~/.mempalace/known_entities.json`, or re-save it in a legacy encoding, or truncate it: `python -c "import os; p=os.path.expanduser('~/.mempalace/known_entities.json'); os.truncate(p, os.path.getsize(p)//2)"`.
3. `mempalace init <project-b> --yes --no-llm` and confirm its entities.
4. Read the registry. For the array and the truncated file it holds project B's categories only, project A's entities and every `topics_by_wing` entry are gone, and step 3 printed `Registry updated`. For the legacy encoding, step 3 ended in a `UnicodeDecodeError` traceback and wrote no config.

**Environment:**
- OS: Linux (WSL2)
- Python version: 3.12.3
- MemPal version: `develop` @ 4bc0c431

**Notes**

`topics_by_wing` is the signal `palace_graph` reads for topic tunnels at mine time, and `_load_known_entities` (`miner.py:1061`) is what tags drawer metadata for entity-filtered search, so a lost registry is not only a lost file: the next mine files drawers without the entities and tunnels the old one would have produced, and nothing about that mine looks unusual. `mempalace mine` itself does not fail on any of the three files.

This is the overwrite half of the shape #2320 reports for deletions, and that report names this function as one of the two places outside its scope. The other, `MempalaceConfig`, is #2356.

Four more places fold a read that did not conclude into an empty default and write that default back over the same file: `palace_graph._load_tunnels` (`:490-499`), `hallways._load_hallways` (`:98-104`), `EntityRegistry.load` (`entity_registry.py:308-316`) and `write_embedder_sidecar` (`backends/_sidecar.py:53-68`), whose docstring says it preserves the other collections' entries. The first two are reached by a plain `mempalace mine` and both write atomically, so what they drop is dropped durably. Each is its own report, none filed yet.

`CLAUDE.md` states the rule this breaks: "Never destroy existing data to rebuild. A crash mid-operation must leave the existing palace untouched."

mvalentsev:
Fixes #2358.

## What does this PR do?

`add_to_known_entities` read `~/.mempalace/known_entities.json` behind `Path.exists()`, fell back to an empty dict whenever that read did not conclude, and rewrote the file whole. One read that did not conclude was therefore enough to lose the registry of every project it was ever built from, while the command printed `Registry updated`. A byte that is not UTF-8 did not even get that far: `UnicodeDecodeError` is neither `JSONDecodeError` nor `OSError`, so it left the call and ended `mempalace init` in a traceback.

Reading now treats only `FileNotFoundError` as an absent registry, since nothing else establishes that there is no registry here. One that does not parse, whether from a truncated write, a byte that is not UTF-8, or a hand edit that lost a brace, is renamed aside and its new name printed before a fresh registry is written. One that exists but cannot be read at all is left alone with a message rather than overwritten, and the call answers `None` when it wrote nothing, so `mempalace init` no longer prints `Registry updated` over a registry it left.

The write goes through a temporary file, an `fsync`, a rename, and an `fsync` of the directory, which is what `EntityRegistry.save` already does (`entity_registry.py:350-363`); `migrate._apply_topics_by_wing_renames` writes this same file through a temporary file and a rename, and `config.py`, `hallways.py`, `logstream.py`, `palace_graph.py`, `repair.py`, `replica.py` and `server_registry.py` publish their own state the same way. The temporary file carries this process's pid, so two concurrent `init` runs never share one, and it is opened `O_NOFOLLOW`. A name the open itself refuses, a directory dropped at it or an orphan this user cannot write, sends the write to a name the directory picks: what the directory says about a name it chose is about the directory, what it says about this one is not.

Two of `develop`'s behaviours are kept on purpose. A registry reached through a symlink is written through it, since renaming over the link would leave the real file, in a dotfiles checkout, holding what it held. And a directory that will not take a name it chose itself gets the merge in place, with a message saying the crash-safety is missing: losing the merge outright is the worse half of that trade.

The narrow fix does not hold. Adding `UnicodeDecodeError` to that `except` stops the traceback, and measured on a 94-byte registry it turns the run into a silent loss instead: the call returns, the file becomes 38 bytes holding this run's one name, and nothing keeps what was there. Today's crash at least leaves the bytes. That is why the read is replaced rather than widened.

### What this does not do

It does not merge anything out of a registry it could not parse. The bytes are kept beside the new one as `known_entities.json.unreadable-<stamp>-<suffix>`, and nothing prunes them, where `repair` prunes its own backups through `backups.prune_backups`. A registry reached through a hard link loses it, since the rename gives the name a new inode. Two `init` runs that both read the registry and then both write it still lose the earlier one's entities: nothing here holds a lock across the read and the write, and what this removes is the case where the loser is a registry nobody read. `test_malformed_existing_registry_starts_fresh` and `test_non_dict_existing_registry_starts_fresh` pass unchanged, since what the new registry holds is the same as before and the old file is now beside it rather than gone.

Interrupted or refused, the write leaves a file behind: `known_entities.json.tmp-<pid>`, or `.known_entities.json.<random>.tmp` when something was already at the pid name, and nothing here removes either. A directory that refuses the rename leaves one on every call rather than on an interrupted one, since removing it needs that same directory; the message names it. A hard link at the temporary name is left alone rather than written through, since the link count is checked before anything is truncated. A quarantine of a symlinked registry lands in the target's directory, where the registry really is, so a dotfiles checkout gains an untracked file; the message prints the full path. A `known_entities.json` that is a FIFO still blocks the merge until something opens the other end, as on `develop`.

It touches nothing outside `known_entities.json`. The same shape sits in four more places the issue names, two of them inside the same `mempalace mine` that reaches this function; each is its own report.

## How to test

```bash
uv run pytest tests/test_known_entities_registry.py tests/test_cli.py -q
uv run pytest tests/test_miner.py tests/test_corpus_origin_integration.py tests/test_project_scanner.py -q
```

## Verification

Measured against `develop` and this branch, through the real CLI on a project with a git history and a manifest, and through `add_to_known_entities` directly, on Python 3.12.3. Counts are one run each.

| claim | how |
|---|---|
| a registry that does not parse is replaced there and kept here | one `mempalace init <project> --yes --no-llm` over a 28-byte JSON array and over a 172-byte truncated object: `develop` replaced both with this run's entities and printed `Registry updated`; here both went to `known_entities.json.unreadable-<stamp>` with the name printed, and the new registry holds the same entities either way |
| a byte that is not UTF-8 ends the command there | the registry re-saved in cp1251, three names: `develop` ended `mempalace init` in `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcc in position 21` and wrote no `config.json` either; here it is quarantined and the run finishes |
| a registry that cannot be read at all is left alone | registry at mode `000`: `develop` raises `PermissionError` out of the call, here the call prints why, answers `None`, and the file is byte for byte what it was |
| what is at stake is a real file | one `mempalace init <clone> --yes --no-llm` on a clone of this repository, 139 distinct commit authors in its history, writes a 370-byte registry: `people` at its cap of 15 names, `projects` 1 |
| the tests pin it | 30 mutations one at a time on a copy of the tree, all 30 failing the suite and none by a syntax error: among them the fail-closed exit removed, the caller printing an update anyway, the quarantine skipped, the write going straight into the file, the temporary file published without an `fsync`, and the errno of the pid-named file deciding for the directory |

Thirty-two new tests in `tests/test_known_entities_registry.py` and one in `tests/test_cli.py`; twenty-three of the thirty-two fail against `develop`'s `miner.py`. Reading the registry moved into a helper, which takes `add_to_known_entities` from complexity 22 on `develop` to 19 against the project's C901 limit of 25. `4620 passed, 31 skipped` on the full suite with `--ignore=tests/benchmarks`, as CI runs it, and nothing failing. `ruff check` and `ruff format --check` clean. `mypy mempalace/miner.py` reports the same 25 errors it does on `develop`.

## Checklist
- [x] No hardcoded paths


igorls (commented):
### Review: LGTM ✅

Thorough and well-reasoned fix for #2358.

**Key Highlights:**
- Properly handles corrupted, unreadable, and non-UTF-8 registry files by quarantining them rather than silently clobbering with partial entities.
- Implements atomic writes with `fsync` and safe temporary naming.
- Comprehensive test coverage in `tests/test_known_entities_registry.py` (all tests passing and verified against failure modes).

This looks solid and ready to be marked as **Ready for review** and merged.

PR #2368, opened 2026-08-26
fix(sync): record which directory a source was mined from (#2367)

### LINKED ISSUE #2367: sync --apply still deletes drawers for files hidden under a mount, in the shapes #2322 left open
**What happened?**

`sync --apply` still removes drawers whose source file it could not reach, in three mount shapes. They are one gap: the neighbour that corroborates the removal is not in the directory the missing file was mined from, and nothing in the palace records which directory that was.

Measured on `develop` @ 4c1e6d0c, which carries #2322, with real `mount` and `umount` inside `unshare --mount --map-root-user`, through the real CLI. One run each, with prose short enough that each file files a single drawer, since these counts are drawers. In every shape `mempalace search` stops naming a file that is on disk the whole time.

| shape | `sync --apply` runs | reports | what was removed |
|---|---|---|---|
| **B**, a volume mounted over a directory the palace knows | while it is mounted | `Missing: 1`, `Unresolved: 0` | the drawer of `x.md`, which is underneath the mount |
| **A**, a mined file in the mount point's lower layer | after the unmount | `Missing: 2`, `Unresolved: 0` | the drawers of the volume's two files, corroborated by the `x.md` underneath that outlived the unmount. Remount and both are back without their drawers |
| **C**, a bind mount over a directory the palace knows | while it is in place | `Missing: 1`, `Unresolved: 0` | the drawer of `keep.md`, which is underneath the bind mount |

Shape C is what `docker -v` does: measured, a `-v` mount is a bind mount of that host subtree in the container's `mountinfo`, and the directory answers with the host's own inode. The README's own recipes mine a source that exists only as such a mount (`README.md:101`, `:123`).

**What did you expect?**

A removal to require that the corroborating neighbour is in the directory the missing file was mined from, decided without writing into the source tree, since `README.md:99` promises mining never does.

The identifiers a filesystem hands out for itself do not separate these shapes, and the directory's own inode does. Measured, one run each: a bind mount leaves `st_dev` the same `2096` on both sides, while the directory's inode answered `1669634`, then `1` under a tmpfs mounted over it, then `1669634` again after the unmount, and the sibling's own number under a bind mount of that sibling.

**How to reproduce:**

Inside `unshare --mount --map-root-user`, with `<p>` a scratch palace and `prose` a paragraph short enough to file one drawer and no more. A and B share one setup and differ only in when the sync runs.

1. `mkdir -p proj vol && prose > proj/x.md`
2. `mempalace --palace <p> mine proj --wing w` reports `Drawers filed: 1`
3. `mount --bind vol proj`
4. `prose > proj/y.md && mempalace --palace <p> mine proj --wing w`
5. Shape B: `mempalace --palace <p> sync proj --wing w --apply` reports `Missing: 1 (removed)`. Then `umount proj`: `x.md` is still there, its drawer is not.
6. Shape A: on a fresh palace, repeat 1 to 4 with a second file mined from the volume, then `umount proj` before the sync. It reports `Missing: 2 (removed)`, and remounting brings both files back without their drawers.

Shape C needs no volume: `mkdir -p proj/sub other && prose > proj/sub/keep.md`, mine, `mount --bind other proj/sub`, add `proj/sub/o.md` and mine again, and the sync reports `Missing: 1 (removed)` for a `keep.md` that never left the disk.

**Environment:**
- OS: Linux (WSL2)
- Python version: 3.12.3
- MemPal version: `develop` @ 4c1e6d0c

**Notes**

A re-mine brings a removed drawer back once the file is reachable again, but not anything the miner did not write: measured, a drawer filed by `mempalace_add_drawer` against that source went `1`, then `0` after the sync, then `0` after the re-mine.

#2320 is the report these were carved out of. #2322 already names shape A as staying open ("separating those needs the filesystem identity of each source at mine time, which nothing records"); B and C are not named there and remove drawers on `develop` today as well.

mvalentsev:
Fixes #2367.

## What does this PR do?

#2322 made `sync --apply` ask for corroboration before removing a drawer whose source file is not at its path: a file gone while a neighbour in the same directory is still there has been deleted. Three mount shapes defeat that, because the neighbour proves the directory is reachable, not that it is the one the file was mined from:

* a mount point whose **lower layer** holds a mined file of its own, which #2322 names as staying open;
* a volume mounted **over** a directory the palace already knows a file in;
* a **bind mount** of another directory over one the palace knows, which is what `docker -v` does. Both container recipes in `README.md` mine a source that exists only as such a mount.

What separates them is the inode of the directory itself, because a path is only a name: mount something at it and the name resolves to the root of what was mounted, a different inode, and unmounting brings the original back. `st_dev` cannot, not for being unstable but because a bind mount puts both sides on one filesystem where it is the same number. **Nothing is written to the source tree for this**, which `README.md:126` promises and the read-only container mounts rely on.

**`mempalace/source_identity.py`** holds the mechanism: a function that stats a directory and answers its inode, `None` on any failure and on a zero, and two that derive it for a file path. Ten call sites record it, and `update_drawer` carries it forward when it refiles a drawer as chunks. What makes a recorded identity mean anything is that it came from the directory of the very path the drawer then names in `source_file`, which is why that step is one function rather than an idiom repeated at each site. Two kinds of row need none and get none: a convo registry sentinel, which `_is_registry_row` keeps before the classification runs, and a row with no `source_file`, which `sync` classifies `no_source`.

**`sync`** reads the directory on both sides of the corroboration, so a mount arriving or leaving while that source's witnesses are stat'ed settles nothing and the drawers are kept. Both values are read as the number an inode is, since a backend that round-trips metadata through JSON hands back what the miner wrote as a string. The comparison is **per drawer**, since one source's drawers need not agree: a sweep records none where a mine records one. `removed_closets` now counts sources a pass emptied rather than sources it removed a drawer from, since purging by source would strand a surviving drawer's own index lines. That is one bounded read per batch of sources; a batch the backend refuses is halved until it answers, and one whose answer fills the bound, like a single source that still cannot be batched, is read one source at a time instead. The bound is what keeps a source still holding rows in a wing the pass did not scan from setting the size of the read.

The bounds live in the docstrings and the changelog entry. A drawer is protected only once it carries the identity, so an existing palace gains it as it is re-mined; backfilling is not offered on purpose, since reading the directory now is reading whatever is mounted now. A filesystem reporting no inode of its own gains nothing, and the `gitignored` route is unchanged. A directory deleted and recreated may answer differently and then keeps the drawers of files that really went, reported as `unresolved`.

## How to test

```bash
uv run pytest tests/test_source_identity.py tests/test_sync.py -q
```

The mount test drives real `mount` and `umount` inside `unshare --mount --map-root-user`, and its gate performs a mount rather than only entering a namespace, so a kernel that allows one and refuses the other skips instead of failing. A runner that withholds namespaces skips it, so a second test asks the same question of a mount somebody else made: `/proc` answers `1`, where the root filesystem answers `2`.

## Verification

Against `develop` @ a9f345cc and this branch, through the real CLI, with real `mount` and `umount` in a mount namespace. One run each unless stated.

| claim | how |
|---|---|
| all three shapes remove drawers on `develop`, and are kept here | `develop` reports `Missing: 1`, `2`, `1` and `mempalace search` stops naming files that are on disk the whole time; here `Missing: 0`, the same drawers as `unresolved`, and the search still answers |
| removal itself is not weakened | a file deleted beside a surviving neighbour, a whole directory deleted, and a gitignored file report the same counts on both sides, and a drawer carrying no identity is removed here exactly as on `develop` |
| the field reaches no caller | against a palace whose rows all carry it, it is in the output of none of twelve MCP tools, `search_memories`, the CLI or the exporter; it does reach the local write-ahead log, beside the absolute path already there |
| the closet purge costs nothing measurable | 2,000 emptied sources, eight runs a side, alternating on an idle machine: 1.88 to 2.40 seconds on `develop` against 1.81 to 2.26 here, this branch's whole range inside `develop`'s and neither side ahead in every run, both removing the same 2,000 drawers and 2,000 closets. Asked per source rather than per batch, the same verdict costs 12.0 to 14.8 seconds against 0.04 over a palace of 80,000 rows, since a probe is a scan |
| the bound on that read earns its place | one source still holding 50,000 rows in a wing the pass did not scan, same verdict either way: unbounded, the backend refuses the read and the halving that follows takes 38 to 40 seconds and a peak RSS of 1.8 GiB; bounded, 3.1 to 3.5 seconds and 236 MiB, beside 2.6 to 3.3 seconds and 162 MiB for reading that batch one source at a time |
| the tests pin it | 46 mutations one at a time on a copy of the tree, all 46 failing the suite, none by a syntax error and none on a stale anchor |

Sixty-five new tests across nine files, of which 45 fail against `develop`'s code with the new module dropped in, and nothing it already had fails there: its own copies of those files report 541 passed, 2 skipped. `ruff check` and `ruff format --check` clean on the version CI pins, `mypy` gives every file this changes the same reports it gives on `develop`, line numbers aside, plus one more `name-defined` in the `mcp_server` fragments — the 84th of a kind `develop` already has 83 of, since a fragment resolves its imports through the package `__init__` — with the new module itself clean, and `sync_palace` comes back at the complexity it has there, 20 against a limit of 25. `5472 passed, 54 skipped` on the full suite as CI runs it.

A second commit was needed for that last number: the hook tests spawned a real background mine and never waited for it, and the child held the per-palace writer lease into the next test file. `develop` fails the same way on the pairs `test_hooks_cli.py tests/mcp/test_write_tools.py` and `test_hooks_cli.py tests/test_mcp_http_transport.py`, one failure each, where either file on its own is green there and both pairs are green here.

## Checklist
- [x] No hardcoded paths


igorls:

mvalentsev:
@igorls Rebased onto `a855bc07` — the three tests moved into `tests/mcp/test_read_tools.py` and `tests/mcp/test_write_tools.py`, CI green.

PR #2380, opened 2026-08-28
fix(search): reuse filtered-query fallback in CLI

### LINKED ISSUE #2373: CLI `search --wing/--room` hard-crashes on "Error finding id" while MCP search degrades gracefully (CLI never uses the existing fallback)
**What happened?**

On a palace that's in the "Error finding id" state, a scoped CLI search hard-crashes, while the exact same scoped search over MCP degrades gracefully and returns results. Unscoped CLI search and `mempalace status` both work fine — only the wing/room-filtered `mempalace search` command errors out:

```
$ mempalace search "email preview error code" --wing lindris_monorepo

  Search error: Error executing plan: Internal error: Error finding id
```

The recovery code for this already exists in the repo — the MCP/programmatic path (`search_memories()`) catches this failure and falls back to an unfiltered query plus a Python-side re-filter. The CLI `search()` function just never got wired through it, so it's the one search path that still calls ChromaDB raw and surfaces the hard error. Still present in 3.8.0.

**What did you expect?**

The CLI `search --wing/--room` to degrade the same way the MCP tool does (fall back and return what it can), instead of erroring out. On a healthy palace the two paths should be indistinguishable; on a diverged one, neither should crash.

Not a data-loss bug, and not the underlying index divergence itself (that's tracked in #1035 / #1665). This is narrowly that one CLI code path is missing the fallback the programmatic path already has — a small, self-contained gap.

**How to reproduce:**

1. On a palace where the HNSW index has drifted from SQLite (e.g. after heavy mining or a bulk delete — any state where filtered `query(where=...)` throws `Error finding id`):
2. `mempalace search "<any query>"` → works (unscoped).
3. `mempalace search "<any query>" --wing <wing>` → `Search error: Error executing plan: Internal error: Error finding id`.
4. The same wing-scoped search via the `mempalace_search` MCP tool → returns results (no crash).

**Environment:**
- OS: macOS 15 (Darwin 25.5.0), Apple Silicon
- Python version: 3.14
- MemPal version: 3.6.0 (confirmed still present in 3.8.0 source)

<details>
<summary>Investigation — root cause, the code asymmetry, and a fix</summary>

**Why it happens.** `searcher.py` has a recovery helper, `_query_drawers_with_filter_fallback(...)`, added in #1245 for exactly this ChromaDB planner failure: it tries `Collection.query(where=...)`, and on failure over-fetches unfiltered and re-applies the wing/room/source_file filter in Python.

The programmatic path uses it. `search_memories()` routes its drawer query through the helper (two call sites). But the CLI path — `cmd_search()` → `search()` — builds the same `kwargs` (including `where`) and calls:

```python
# searcher.py, def search(...)  — the CLI path
results = col.query(**kwargs)          # raw; no fallback
```

So when `query(where=...)` throws `Error finding id`, `search()` has nothing to catch it and prints `Search error: ...`. Meanwhile `search_memories()` (MCP) sails through via the helper. That's the whole asymmetry: **the fallback exists, the CLI just doesn't call it.**

**Fix** — route the CLI query through the existing helper, same as `search_memories()`:

```python
# def search(...)
        if where:
            kwargs["where"] = where

-       results = col.query(**kwargs)
+       results = _query_drawers_with_filter_fallback(
+           col, kwargs, query, n_results, wing, room
+       )
```

Verified locally: with the raw call, `--wing`/`--room` throws `Error finding id`; after routing through the helper the command no longer crashes and returns whatever the over-fetch pool contains for that wing (i.e. it now behaves like the MCP path). One-line-ish change; no new logic, just reuse.

**Relationship to other issues.**
- #1245 (merged) — added `_query_drawers_with_filter_fallback` and wired it into `search_memories()`. This issue is that the CLI `search()` was left out of that same wiring.
- #951 (open PR) — adds graceful degradation to `search_memories()`; does not touch the CLI `search()` path.
- #1035 / #1665 — the underlying HNSW↔SQLite divergence that *causes* `Error finding id`. `mempalace repair` is the fix for the corruption; this issue is only about the CLI not degrading gracefully when that corruption is present.

**Secondary observation (separate, lower priority).** On the palace where I hit this, `mempalace repair-status` reports a 56k-drawer gap (SQLite 345,303 / HNSW 289,018) as `status: OK — within flush-lag tolerance`. Whole wings were unreachable by vector search in that state, so the "flush-lag tolerance" band looks far too permissive at that magnitude — it labels a real divergence as healthy. Happy to file that separately if useful.

</details>

MichaelWave369:
Closes #2373.

Routes the CLI `search()` drawer query through the existing `_query_drawers_with_filter_fallback()` recovery path already used by `search_memories()`.

When a scoped Chroma query (`--wing` / `--room`) hits the known `Error finding id` planner failure, the CLI now matches the MCP behavior: it retries unfiltered with the existing over-fetch and reapplies the requested scope in Python instead of hard-failing.

No new recovery semantics are introduced; this only removes the CLI/MCP asymmetry by reusing the existing helper. Unscoped query failures still raise as before because the helper only falls back when a `where` filter was present.

Regression coverage drives the CLI path through:
- a failing wing+room filtered query;
- the unfiltered fallback;
- Python-side scope filtering;
- output verification that only the in-scope drawer is rendered.

Validation against current `develop` (`a9f345c`):
- `python -m pytest -q tests/test_search_drawer_id.py` — **7 passed**
- `ruff check mempalace/searcher.py tests/test_search_drawer_id.py` — **passed**
- `ruff format --check mempalace/searcher.py tests/test_search_drawer_id.py` — **passed**
- `python -m compileall -q mempalace/searcher.py tests/test_search_drawer_id.py` — **passed**

Final branch shape: one commit, 1 ahead / 0 behind, two changed files only.

### DIFF
--- a/mempalace/searcher.py
+++ b/mempalace/searcher.py
@@ -740,7 +740,7 @@ def search(
         if where:
             kwargs["where"] = where
 
-        results = col.query(**kwargs)
+        results = _query_drawers_with_filter_fallback(col, kwargs, query, n_results, wing, room)
 
     except Exception as e:
         print(f"\n  Search error: {e}")
--- a/tests/test_search_drawer_id.py
+++ b/tests/test_search_drawer_id.py
@@ -7,6 +7,7 @@ from mempalace.searcher import (
     _aligned_query_ids,
     _finalize_candidate_hits,
     _query_drawers_with_filter_fallback,
+    search,
     search_memories,
 )
 
@@ -328,3 +329,58 @@ def test_union_results_use_parent_or_lexical_hit_id():
         and "_parent_drawer_id" not in hit
         for hit in hits
     )
+
+
+def test_cli_scoped_search_uses_existing_filter_fallback(capsys):
+    drawers_col = MagicMock()
+    drawers_col.distance_metric = "cosine"
+    drawers_col.metadata = {"hnsw:space": "cosine"}
+    drawers_col.query.side_effect = [
+        RuntimeError("Error finding id"),
+        {
+            "ids": [["drop-id", "keep-id"]],
+            "documents": [["drop document", "keep document"]],
+            "metadatas": [
+                [
+                    {
+                        "wing": "keep",
+                        "room": "other",
+                        "source_file": "/palace/drop.md",
+                    },
+                    {
+                        "wing": "keep",
+                        "room": "notes",
+                        "source_file": "/palace/keep.md",
+                    },
+                ]
+            ],
+            "distances": [[0.2, 0.1]],
+        },
+    ]
+
+    with patch(
+        "mempalace.searcher.resolve_backend_name",
+        return_value="sqlite_exact",
+    ):
+        with patch(
+            "mempalace.searcher._open_collection_or_explain",
+            return_value=drawers_col,
+        ):
+            search(
+                "needle",
+                "/unused",
+                wing="keep",
+                room="notes",
+                n_results=2,
+            )
+
+    output = capsys.readouterr().out
+    assert drawers_col.query.call_count == 2
+
+    filtered_call, fallback_call = drawers_col.query.call_args_list
+    assert filtered_call.kwargs["where"] == {"$and": [{"wing": "keep"}, {"room": "notes"}]}
+    assert "where" not in fallback_call.kwargs
+    assert fallback_call.kwargs["n_results"] == 30
+
+    assert "keep document" in output
+    assert "drop document" not in output


BuboTheWise:
Heads up: https://github.com/MemPalace/mempalace/pull/2425 was our independent fix for the same #2373 (identical one-line change). It has been closed in favor of this PR — all 9 CI checks green here. No conflicts or objections from our side. One small suggestion for completeness: consider adding the one-line rationale comment from #2425 (CLI has no `source_file` dimension, so the helper re-applies wing/room in Python) so future readers see why `source_file` is intentionally left `None` on this path.

Issue #1693, opened 2026-06-05
Hook diary checkpoints are never visible to diary_read — wing derivation mismatch


YC-AIUSER:
## Environment
- OS: Windows 11 Pro
- MemPalace: 3.3.6 (plugin install)
- Claude Code: 2.1.161

## Problem

Stop hook diary checkpoints are saved to a **project-derived wing** (e.g. `wing_.claude`, `wing_portable_claude_paseo_kit`), but `mempalace_diary_read(agent_name="claude")` reads from `wing_claude`. These two naming conventions never intersect, so **hook-saved diary checkpoints are invisible to diary_read in every project**.

This is not edge-case — it's structural. The hook and diary_read use fundamentally different wing derivation logic:

| Operation | Wing derivation | Example wing |
|---|---|---|
| Hook checkpoint (`hook_stop` → `_diary_checkpoint`) | Project directory name from transcript path | `wing_.claude`, `wing_portable_claude_paseo_kit` |
| `diary_read(agent_name=X)` | Agent name parameter | `wing_claude`, `wing_X` |

## Reproduction

1. Install MemPalace plugin, configure hooks
2. Run a Claude Code session with 3+ exchanges (SAVE_INTERVAL=3)
3. Hook fires, saves checkpoint: `diary_wing_.claude_20260605_...`
4. New session: call `mempalace_diary_read(agent_name="claude")` → returns only agent-written entries, never hook checkpoints
5. Call `mempalace_list_drawers(wing="wing_.claude", room="diary")` → hook checkpoints are there

## Evidence from hook.log

```
[13:30:36] TRIGGERING SAVE at exchange 16
[13:30:37] Diary checkpoint saved: diary_wing_.claude_20260605_133037150537_1e71cc58b856
```

But `diary_read(agent_name="claude")` returns 0 entries after 2026-06-03 (only manual `diary_write` entries are visible).

## Expected behavior

Hook-saved diary checkpoints should be discoverable by `diary_read`, either by:
- Saving hook checkpoints to the same wing that `diary_read` queries (agent-name-based), or
- Having `diary_read` also search project-derived wings, or
- Providing a `diary_read` parameter to query by wing (current `wing` param seems to not work for this)

## Workaround

Agent manually calls `diary_write` every N turns (CLAUDE.md rule), bypassing hooks entirely. This defeats the purpose of auto-save hooks.

YC-AIUSER:
## Root cause update

[...]

PR #2382, opened 2026-08-28
feat: accept hermes as a hook harness

PostProtoroman:
## Problem

`SUPPORTED_HARNESSES` is not only an input-validation set — `_diary_agent_for_harness` derives the diary identity from the same string:

```python
return "claude" if harness == "claude-code" else harness
```

So a harness that is *not* in the set has no way to run hooks under its own name. It has to borrow an existing one, and every checkpoint it files then lands in that harness's diary. The session's own `mempalace_diary_read(agent_name=...)` returns nothing — which is exactly the invisibility failure #1693 fixed for the hardcoded `"session-hook"` identity, reappearing through a different door.

The concrete case: [Hermes Agent](https://github.com/NousResearch/hermes-agent) stores sessions in SQLite rather than JSONL transcripts, so bridging them into MemPalace means exporting to a transcript `normalize` already recognises and calling `mempalace hook run`. That call currently has to pass `--harness claude-code`, and every Hermes diary checkpoint is filed under agent `claude`, mixed into the Claude Code diary. Drawers are unaffected (they key on wing, not agent), so only continuity between the bridged agent's own sessions breaks — silently, since the hook still exits 0.

## Why this is a small change

Nothing downstream of the name is harness-specific:

- `_parse_harness_input` reads `session_id`, `transcript_path` and `stop_hook_active` and nothing else — it validates the harness and is otherwise format-agnostic.
- `_count_human_messages` distinguishes the Claude Code and Codex line shapes by inspecting each line, not by the harness argument.

Adding a name to the set is therefore all a new harness needs, once it can produce a transcript `normalize` accepts. The `_diary_agent_for_harness` docstring already states this as the intent:

> returning the harness name keeps a newly supported harness discoverable instead of silently invisible again

## Changes

- `mempalace/hooks_cli.py` — add `hermes` to `SUPPORTED_HARNESSES`; document that the set drives diary identity, so an omission is a correctness bug rather than a missing convenience.
- `mempalace/cli.py` — lift the `--harness` choices into `HOOK_HARNESS_CHOICES`. The list stays duplicated rather than imported from `hooks_cli`: the parser is rebuilt on every CLI invocation and `hooks_cli` is a deliberate lazy import (see `run_hook`), so deriving it there would put the hook module on the critical path of every unrelated command and eat the startup budget.
- `tests/test_hooks_cli.py`
  - `_diary_agent_for_harness("hermes") == "hermes"`
  - `test_every_supported_harness_has_its_own_diary_identity` — no supported harness may collide with another's diary name, so the next addition cannot silently reintroduce the bug
  - `test_cli_harness_choices_match_supported_harnesses` — guards the deliberate duplication against drift
  - `test_stop_hook_files_checkpoint_under_harness_agent` extended to `hermes`

## Verification

```
149 passed, 1 skipped
ruff check .            All checks passed!
ruff format --check .   237 files already formatted
```

No change to storage format, hook wire protocol, or existing harness behaviour.

### DIFF
--- a/mempalace/cli.py
+++ b/mempalace/cli.py
@@ -2828,6 +2828,15 @@ def _reconfigure_stdio_utf8_on_windows():
     reconfigure_stdio_utf8_on_windows(stdout_errors="replace", stderr_errors="replace")
 
 
+# Harness names accepted by ``hook run --harness``. Duplicated from
+# ``hooks_cli.SUPPORTED_HARNESSES`` on purpose: the parser is rebuilt on every
+# CLI invocation and ``hooks_cli`` is a deliberate lazy import (see ``run_hook``)
+# to hold the startup budget, so importing it here to derive the list would put
+# the hook module on the critical path of every unrelated command.
+# ``test_cli_harness_choices_match_supported_harnesses`` asserts the two stay equal.
+HOOK_HARNESS_CHOICES = ("claude-code", "codex", "hermes")
+
+
 def main():
     """CLI entry point for the ``mempalace`` console script.
 
@@ -3196,7 +3205,7 @@ def main():
     p_hook_run.add_argument(
         "--harness",
         required=True,
-        choices=["claude-code", "codex"],
+        choices=list(HOOK_HARNESS_CHOICES),
         help="Harness type (determines stdin JSON format)",
     )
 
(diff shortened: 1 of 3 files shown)


igorls:

PR #2154, opened 2026-08-04
fix(status): wire CLI status() to get_all_metadata() fast path (#2153)

javapacr:
Fixes #2153.

`miner.status()` now delegates to `get_all_metadata()` when the collection exposes it (same pattern as `mcp_server._fetch_all_metadata()`, #1796), before falling back to the existing offset loop.

**Before:** 1m46s (O(n²) — 32 full-collection re-fetches on Qdrant)
**After:** 3.4s (single `_scroll_all()` pass)
**Palace:** 159,601 drawers, Qdrant backend

Issue #104, opened 2026-04-07
Architectural similarities to Sara Brain (prior art — public since March 16, 2026)


LunarFawn:
Hi there,

  I'm Jennifer Pearl, the author of Sara Brain — a path-of-thought knowledge representation
  system published as a Zenodo preprint (DOI: 10.5281/zenodo.19441821).

  I'm reaching out because I've noticed significant architectural similarities between
  MemPalace and Sara Brain's publicly documented design. I want to establish the timeline and
   ask about attribution.

  Sara Brain's public record:

  - March 16, 2026 — Sara Brain web app deployed to GitHub Pages (LunarFawn/SaraBrain), fully
   public, with live interactive demo
  - March 19-23, 2026 — Design philosophy documents (v005-v009) published on the web app
  detailing: path-of-thought architecture, "LLM as senses not the brain," SQLite-based
  knowledge graph, no-forgetting/no-decay policy, local-only operation, zero cloud dependency
  - March 23, 2026 — Ollama provider committed: "LLM as sensory interface, not the brain"
  - April 6, 2026 — Zenodo preprint with DOI

  All of the above is verified by immutable git history and GitHub Pages deployment
  timestamps.

  MemPalace timeline:

  - April 4, 2026 — Initial commit (full architecture in a single commit)
  - April 5, 2026 — Repository created on GitHub

  Specific architectural similarities:

  1. Local-only, no cloud, privacy-first — both systems store all knowledge locally with no
  data leaving the machine
  2. LLM as interface, not brain — both explicitly position the LLM as a sensory/interface
  layer, not the intelligence
  3. SQLite knowledge graph — both use SQLite to store directed relationships (Sara Brain:
  neurons/segments, MemPalace: entities/triples)
  4. No forgetting / no decay — both reject the mainstream approach of summarizing or
  discarding memories
  5. Graph traversal for retrieval — both use BFS-style traversal to find knowledge
  connections
  6. Zero/minimal external dependencies — both make this a design principle

  These are the core architectural pillars of Sara Brain, all of which were publicly
  documented on GitHub Pages weeks before MemPalace's first commit.

  I'm not making accusations — convergent thinking happens. But given the timeline and the
  specificity of the overlap, I'd appreciate knowing whether you were aware of Sara Brain's
  public documentation, and whether you'd be willing to add an acknowledgment or citation.

  The paper: Path-of-Thought: A Neuron-Chain Knowledge Representation System with Parallel 
  Wavefront Recognition
[...]

PR #2385, opened 2026-08-29
fix(palace): wire prefetch_content_hashes to get_all_metadata() fast path (#1796)

jhsmith409:
## What does this PR do?

`prefetch_content_hashes()` is the last unmigrated caller of the #1796 O(n²) bulk-metadata pattern. It paged the collection with `get(limit=1000, offset=N)`, which assumes `limit`/`offset` reach the store — true for Chroma's SQL cursor, false for Qdrant, where `get()` goes through `_rows() → _scroll_all()` to materialize every matching row and slices the result in Python afterwards. Each page re-walked the entire collection to keep a thousandth of it.

This is the same fix already applied to `mcp_server._fetch_all_metadata()` in #1832 and proposed for `miner.status()` in #2154, at the convo miner's own call site.

## Impact

Found by profiling an idle-looking box: `mempalace mine --mode convos` sat at 44–70% of a core for 17+ minutes with zero forward progress. `py-spy` put every sample in the same stack:

```
prefetch_content_hashes   (mempalace/palace.py:1651)
  → get                   (backends/embedding_wrapper.py:143)
  → get                   (backends/qdrant.py:1030)
  → _rows / _scroll_all   (backends/qdrant.py:804/779)
  → json.loads            (backends/qdrant.py:404)
```

On a **260,415-drawer** Qdrant palace: 261 outer pages × 260,415 rows ≈ **68M payloads** fetched and JSON-decoded, 800 MB resident — to build a map then consulted for one transcript. Since the save hook fires a mine every few exchanges, one core stayed pegged for the length of a working session.

| | before | after |
|---|---|---|
| `prefetch_content_hashes` | 17+ min | **4.1 s** |
| full `mine --mode convos --dry-run` | never observed to finish | **4.5 s** |

## Why this is safe

`BaseCollection.get_all_metadata()`'s default implementation *is* the same `limit`/`offset` loop this replaces, so backends with a real server-side cursor are behaviorally unchanged. Only backends that override it — Qdrant and pgvector — change, and they change to one native cursor pass. Verified on the live palace that `get_all_metadata()` returns exactly `collection.count()` rows (260,415), so nothing is dropped relative to the paginated read.

## Relationship to #2384

#2384 bounds `_scroll_all()` by `offset + limit` at the backend layer, which helps this call site but does not fix it: the loop would still walk `offset + 1000` rows per page, i.e. ~n²/2 total. The two changes are complementary — #2384 makes small-page `get()` calls cheap in general, this one stops asking for pages when the caller wants the whole collection. They don't conflict textually (different files).

## Tests

New `tests/test_prefetch_content_hashes_fast_path.py`:

1. The fast path is taken — one `get_all_metadata()` call, `get()`/`count()` never touched.
2. Filtering/aggregation semantics unchanged: extract_mode scope (including the #104 `ingest_mode="sweep"` exclusion), `normalize_version` floor, comma-joined bundle hashes, first-source-wins, wing-keyed.
3. A failing backend still returns the documented partial result rather than propagating.

All four fail against the unpatched function and pass with it. `tests/test_convo_miner.py`, `tests/test_miner.py`, `tests/test_closet_llm.py` and `tests/test_qdrant_bulk_metadata_scroll.py` pass unchanged (206 tests). `ruff check` / `ruff format` clean.

https://claude.ai/code/session_01JQTBsH7RpYegFEAqnMQ6Yt


jhsmith409:
Rebased onto current `develop` (`d9f0590`) — the branch had gone `CONFLICTING`. Now `832bb83`, `MERGEABLE` again, one commit.

**Conflict resolution (one file, `CHANGELOG.md`).** The only conflict was positional: this branch's entry was written under `## [Unreleased]`, and 3.9.0 shipped on 2026-08-31 (`434f00b`), which turned that section into `## [3.9.0] — 2026-08-31`. A naive resolution would have filed an unmerged change under a released version. The entry now sits under a fresh `## [Unreleased]` / `### Bug Fixes`, and the 3.9.0 section is byte-identical to `develop`. `palace.py` and the new test file applied unchanged.

**One correction to the title.** It reads "(#1796)", but #1796 was closed on 2026-06-22 — this doesn't reopen it. `prefetch_content_hashes()` is the last call site still using the O(n²) `get(limit, offset)` pattern that issue was about; the migrations for `mcp_server._fetch_all_metadata()` (#1832) and `miner.status()` (#2154) left this one behind. Happy to retitle if referencing a closed issue is confusing.

**Verification** (clean venv, same machine, this branch vs. `develop` at `d9f0590`):

| | `832bb83` | `develop` |
|---|---|---|
| `pytest tests/ --ignore=tests/benchmarks` | 4825 passed, 31 skipped | 4821 passed, 31 skipped |
| `ruff check .` | clean | clean |
| `ruff format --check .` | clean | clean |

The delta is the four new tests in `tests/test_prefetch_content_hashes_fast_path.py`; all four fail against the unpatched function.

**Could a maintainer approve the workflow run?** No check has ever run on this PR — the runs queued on the previous head sat at `action_required` from 2026-08-29 and the push above queued fresh ones (`33825821653`, `33825821642`), also waiting. For what that state does when it's left alone: the equivalent runs on my other open PR (#2043) were queued on 2026-08-05, never approved, and 30 days later GitHub flipped them to `failure` with zero jobs ever created — a red X that reports the approval window expiring rather than anything about the code.

jhsmith409:
Developers - if you are not planning on picking this up and merging into Mempalace, please let me know so I stop maintaining it current with the base branch.

PR #2391, opened 2026-08-30
feat: name the stdio MCP writer in peer-writer refusal

xxxn3m3s1sxxx:
## Summary

When multiple stdio \mempalace.mcp_server\ replicas share one palace on the same host, exactly one holds the per-palace single-writer lease (via \lock\) and every other process is correctly refused palace writes. But the refusal message is opaque: `Peer MCP writer active; this server is read-only for mutating tools` with no way to tell *which* replica holds the lease. In a live multi-session setup (dev / suckz / orchestrator all connected to one palace) that reads like a fault even though the lease is working as designed.

This PR makes the refusal diagnostic by publishing a per-palace marker naming the current stdio writer, and including it in the refusal message.

## What changed

- \server_registry.py\: add \stdio_writer_path\ / \write_stdio_writer\ / \clear_stdio_writer\ / \
ead_stdio_writer\. A small, hub-independent record (\{pid, role, started_at, palace_path}\) stored next to the existing \serverinfo.json\, guarded by the same \_pid_alive\ liveness rule so a crashed writer's marker is never reported as live.
- \mcp_server.py\:
  - \_acquire_mcp_writer_lock\ publishes the marker when the flock lease is won (best-effort; failure never fails the acquire — the OS flock stays the real authority).
  - \_release_mcp_writer_lock\ clears it (only if it still names this pid, so a slow exit can't delete a newer writer's record).
  - \_mcp_peer_writer_refusal\ reads the marker and appends `(writer: pid <pid> since <started_at>)` to the message plus a \writer\ data field.

## Why not a hard singleton guard

Rejecting a second replica outright would break the legitimate case where several sessions read one palace concurrently with a single writer. The lease already serializes writes correctly; the gap was only diagnosability. This stays purely informational and changes no write/deny behavior.

## Tests

\	ests/test_stdio_writer_marker.py\ (7 tests): write/read/clear roundtrip, clear only removes own pid, dead-pid marker is ignored, refusal names the writer, refusal without a marker still works.

Full run: \pytest tests/test_stdio_writer_marker.py tests/test_hub_forward.py tests/test_palace_locks.py tests/test_mcp_logstream.py\ \u2192 76 passed. \pytest tests/test_mcp_server.py -k 'writer or peer or refusal or lock'\ \u2192 20 passed. \
uff check\ and \
uff format --check\ clean.

## Scope

No new dependencies, no breaking changes, no upload/telemetry. Purely additive + a clearer error payload.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -817,9 +817,7 @@ def _mcp_peer_writer_refusal(req_id, tool_name: str):
     except Exception:
         logger.debug("could not read stdio writer marker", exc_info=True)
 
-    message = (
-        "Peer MCP writer active; this server is read-only for mutating tools"
-    )
+    message = "Peer MCP writer active; this server is read-only for mutating tools"
     if holder:
         message += f" (writer: pid {holder.get('pid')} since {holder.get('started_at')})"
 
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -697,6 +697,14 @@ def _release_mcp_writer_lock() -> None:
         # repeatedly without exiting the same context manager twice.
         _MCP_WRITER_LOCK_CM = None
         _MCP_WRITER_READ_ONLY = False
+        # Best-effort: drop the stdio writer marker only if it still names us,
+        # so a stale clear cannot remove a newer writer's record.
+        try:
+            from .server_registry import clear_stdio_writer
+
+            clear_stdio_writer(_config.palace_path)
+        except Exception:
+            logger.debug("could not clear stdio writer marker", exc_info=True)
         lock_cm.__exit__(None, None, None)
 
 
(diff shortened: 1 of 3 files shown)


xxxn3m3s1sxxx:
## Independent verification (reliability-verifier, clean checkout)

Since this PR originates from a fork and no CI checks are reported on the branch, here is the independent verification evidence from a clean-checkout re-run (not the author's own report).

**Branch state at verification:** HEAD \2880986\, parent \90b8d64\, base \87e6f38\ (\aseRefOid\ confirmed).

| Check | Result |
|---|---|
| \pytest tests/test_stdio_writer_marker.py\ | 7 passed |
| \pytest tests/test_mcp_server.py -k 'writer or peer or refusal or lock'\ | 20 passed |
| \
uff check\ (new files) | All checks passed |
| \
uff format --check .\ (CI-equivalent) | 230 files already formatted |
| \git diff 87e6f38..HEAD --name-only\ | exactly the 3 intended files |
| \git status --porcelain\ | clean |

Full-suite: 17 failures were confirmed **pre-existing** (identical node set reproduced against base \87e6f38\, environment/packaging artifacts, none touch patch code). mypy: identical error count to base (no new errors; mypy is not a repo gate).

**Scope:** purely additive — marker publish on acquire (best-effort, flock remains the authority), clear on release (own-pid guarded), refusal message appends \(writer: pid <pid> since <started_at>)\ plus a \writer\ data field. No write/deny behavior changed, no new dependencies, no telemetry.

igorls:

Issue #2403, opened 2026-08-31
Auto-save Stop hook re-mines the whole transcript in a loop, starving the hub; interrupted mine leaves a partial index


PostProtoroman:
**What happened?**

A long Claude Code session (75 MB transcript, ~5,200 human messages) with the plugin auto-save enabled drove the palace into a self-sustaining denial of service: every Stop event re-mined the entire transcript from scratch, the mine held the writer lease for minutes at a time, the hub answered nothing else, and MCP clients saw timeouts and "Server unavailable". Killing the stuck mine mid-run then left the palace with a **partial index** — 5,765 of 8,022 previously-indexed drawers — until a later full re-mine restored it.

Root-causing this surfaced five distinct defects. All were verified against `develop` @ e809834 (3.9.0); line numbers below refer to that commit.

**1. Re-mining a changed transcript is purge-and-refile, not incremental** (`mempalace/convo_miner.py` ~L674)

When a source file's mtime changes, all of its drawers are deleted and the whole file is refiled. An *active session transcript changes every turn*, so with auto-save every N exchanges the same session is fully purged and fully re-mined over and over — O(n²) work over the life of a session, and a >5-minute full re-mine for a 75 MB transcript. CLAUDE.md calls "Incremental only — append-only ingest after initial build" a non-negotiable design principle; the convo path doesn't have an incremental mode at all.

**2. The purge-before-refile window loses previously-safe data on interruption** (`mempalace/convo_miner.py` L674–L780)

Because the purge precedes the refile, a crash or kill mid-refile leaves the palace missing data it already had. Observed directly: SIGTERM during a re-mine of an already-fully-indexed transcript left 5,765 of 8,022 drawers. The `chunk_total` guard from #2183 makes the *next* mine detect and repair the incomplete set — but between the interruption and that next mine, search silently returns less than it did before the operation started. That violates "a crash mid-operation must leave the existing palace untouched" (CLAUDE.md, same principle). The `except Exception` cleanup at L755+ only covers in-process errors, not SIGKILL/SIGTERM/power loss.

**3. `hooks.mine_transcript: false` is silently ignored — transcript mining cannot be disabled without losing diary checkpoints**

`git grep mine_transcript` on `develop` returns nothing: the config key is read nowhere. The only way to stop the runaway transcript mining is `hooks.auto_save: false`, which also disables the (cheap, valuable) diary checkpoints. A config key that validates as fine and does nothing is worse than an error.

[...]

PR #2408, opened 2026-09-01
fix: incremental convo re-mine, crash-safe refile, mine_transcript gate, stop-hook loop (#2403 defects 1–4)

PostProtoroman:
Fixes the first four defects of #2403 — the Stop-hook auto-save self-DoS: an active session's transcript was fully purged and re-mined on every save point, an interrupted mine shrank the palace, `hooks.mine_transcript` was silently ignored, and a failing diary write re-armed the expensive mine on every subsequent Stop.

## Commits (in order)

**`fix(hooks): forward diary checkpoints to a live hub; add hooks.mine_transcript gate`** (defects 3, 4a)
A long-lived hub holds the writer lease for its whole lifetime, so the hook's in-process diary write was refused on exactly the machines the shared-brain guide describes — the expensive transcript mine kept working while the cheap continuity layer silently vanished. `_save_diary_direct` now discovers a live, writable hub the same way the CLI forwarder does (serverinfo → `/healthz` → `tools/call`), with no local retry after a hub-side error (it may have filed the entry). New `hooks.mine_transcript` (env `MEMPALACE_HOOKS_MINE_TRANSCRIPT`, default `true`) gates the verbatim transcript ingest separately from the diary checkpoint, honored by all three hooks via `_ingest_transcript`; with the mine off, PreCompact files a checkpoint instead.

**`fix(miner): incremental convo re-mine — never purge before the new set is written`** (defects 1, 2)
The convo miner's rebuild contract was purge-before-refile: any changed source had ALL of its drawers deleted, then the whole file re-embedded — O(n²) over the life of an active session (>5 min per pass on a 75 MB transcript), and a crash inside the refile left the palace holding less than before the mine started (observed live: 5,765 of 8,022 drawers after one SIGTERM), which "Incremental only" forbids. Drawer ids were already deterministic over `(source_file, extract_mode, chunk_index)`, so the pass now snapshots existing drawers, re-embeds only chunks whose stored `chunk_hash` differs (pre-hash drawers count as changed once and self-migrate), refreshes unchanged drawers with a metadata-only touch so the `chunk_total` completion check (#2183) still sees one full current-mtime group, and deletes orphaned ids last. An interruption anywhere leaves every pre-existing drawer present and the group incomplete, so the next mine repairs instead of skipping; the worst crash window leaves transient duplicates, never loss. Chroma's metadata-only `update()` keeping documents/embeddings untouched was verified against a scratch palace on chromadb 1.5.7.

**`fix(hooks): a failed diary no longer re-fires the transcript mine every Stop`** (defect 4)
The save marker advanced only on diary success while `_ingest_transcript` ran unconditionally before it — with a hub refusing the in-process diary write, EVERY Stop re-triggered a full mine of the transcript, which occupied the palace write path, which kept the diary failing (observed as `TRIGGERING SAVE at exchange 5193` repeating turn after turn, with the hub's starvation warning logged 48×). The marker now advances when the diary landed OR the ingest is in flight (`_spawn_mine` reports "already running" as in-flight for the same reason), and stays put only when nothing succeeded — preserving retry-on-total-failure.

**`fix: support current Codex rollout messages`**
Codex CLI ≥ 0.149 emits `event_msg`/`item_completed` with `payload.item` conversation items instead of the v1 `user_message`/`agent_message` shape; both wire formats now parse (they coexist in real corpora). Without this, mining a current rollout registered the file as a permanently-empty sentinel.

## Testing

- `uv run pytest tests/ --ignore=tests/benchmarks`: **4756 passed, 32 skipped** on this branch. The 2 failures in `tests/test_mcp_http_transport.py` under the full suite fail identically on pristine `develop` @ e809834 (verified in a separate worktree) and pass in isolation on both — pre-existing full-suite flakes on this machine, not introduced here.
- New tests: incremental re-mine of a grown file upserts only the delta and touches the rest; an interrupted pass deletes nothing and still reads as incomplete to `file_already_mined`; the failed-diary/dispatched-ingest marker matrix; `_spawn_mine` "already running" semantics; `hooks.mine_transcript` gate + hub-forwarding coverage. Both key regression tests were mutation-checked (fix reverted → test goes red).
- `ruff check .` and `ruff format --check .` clean repo-wide.
- Running in production on the reporting machine's hub since 2026-09-01: the incident session's transcript now re-mines as a cheap delta, and a 278-file historical import ran through the same code path with zero failures.

Defect 5 of #2403 (chromadb writes on the hub's exclusive request path have no timeout, so any mine holds every peer read-only) is deliberately **not** addressed here — it needs an upstream design decision (daemon routing vs. threading a write-gate through the miners), per CONTRIBUTING's issue-first rule for changes of that size. #2403 carries the discussion.


RoshanDev:
I am maintaining the stacked Draft #2419, which currently carries equivalent cherry-picks of this PR's four commits so that defect 5 and the crash-safety follow-ups could be validated end-to-end. #2419 now has a clean current-head Codex review and 10/10 CI, but I am intentionally keeping it Draft to avoid presenting maintainers with a duplicated review surface.

Could you advise which landing path you prefer?

1. Merge #2408 first; I will then rebuild #2419 on current `develop` and drop the four duplicate commits.
2. Let #2419 supersede #2408; I will consolidate its title, body, and commit stack before marking it Ready.

Either way, the separate large-JSONL streaming-memory work remains isolated in #2214.

PostProtoroman:
Thanks for validating the stack end-to-end — option 1 works best for us: land #2408 first, then rebase #2419 on develop and drop the duplicate commits.

That keeps the review surface small (the four commits here are the reviewed unit for defects 1–4) and gives your defect-5 work a clean base. Happy to address review feedback on this PR quickly if anything comes up — and glad the write-gate design from #2403 held up under a full validation run.

RoshanDev:
> Thanks for validating the stack end-to-end — option 1 works best for us: land #2408 first, then rebase #2419 on develop and drop the duplicate commits.
> 
> That keeps the review surface small (the four commits here are the reviewed unit for defects 1–4) and gives your defect-5 work a clean base. Happy to address review feedback on this PR quickly if anything comes up — and glad the write-gate design from #2403 held up under a full validation run.

Thanks — option 1 works for me too. I’ll keep #2419 Draft and wait for #2408 to actually land in develop. Once it does, I’ll rebuild #2419 on the new base, drop the four duplicate commits, rerun the full CI matrix, and request a fresh current-head review.

mempalace/convo_miner.py
```diff
@@ -700,73 +708,121 @@ def _file_chunks_locked(
...
+        # mine retries (#2183 / #2403).
+        for batch_start in range(0, len(to_upsert), DRAWER_UPSERT_BATCH_SIZE):
+            batch = to_upsert[batch_start : batch_start + DRAWER_UPSERT_BATCH_SIZE]
+            batch_ids = [drawer_id for drawer_id, _, _ in batch]
+            batch_docs = [content for _, content, _ in batch]
+            batch_metas = [meta for _, _, meta in batch]
+            assert_no_collisions(list(zip(batch_ids, batch_metas)), collection)
+            try:
+                collection.upsert(
+                    documents=batch_docs,
```
igorls, line 792: [P1] Preserve previous content before overwriting positional IDs

These upserts replace existing rows before the entire pass succeeds. If batch one rewrites an old chunk and batch two fails, the original text from batch one is already gone, even though its drawer ID remains. I reproduced that partial state with a controlled collection. It contradicts the comment above claiming everything previously held is still there and `CLAUDE.md`'s crash-preservation requirement. Stage a new generation or otherwise retain/restore old content, and test a failure after a successful replacement batch. Distinct version identities are also needed for the additive-history requirement in #1593.

igorls (changes requested):
Requesting changes because the crash-preservation fix remains incomplete.

The exact changed function, exercised with a controlled collection and a second-batch failure, leaves `replacement0/original1` after starting with `original0/original1`: the first successful upsert has already destroyed `original0`. Delaying deletion protects row existence, but not the original verbatim content. This is an incomplete fix of the existing purge/rewrite problem, not a claim that the patch makes the old implementation worse.

Please preserve the previous generation across failed rewrites and add a regression that checks old content, not only drawer IDs/counts. Successful rewrites also need to respect #1593's additive-history requirement: in-place overwrites and deletion of chunks absent from the new transcript discard prior conversational history.

Focused Windows validation of the convo-miner, hooks, normalization and config modules: 535 passed, 4 skipped. The current branch also has merge conflicts; the eventual resolution needs fresh validation.

PR #2416, opened 2026-09-02
feat(dynamics): add read-time effective_strength, and correct the module docstring

scotmcc:
## What does this PR do?

Adds `effective_strength(connection, now)` — the same Ebbinghaus curve as
`apply_decay`, computed when the strength is read instead of written back.
It is pure, so a read cannot change a ranking, and the answer depends only on
elapsed time rather than on how many times a maintenance pass ran.

That last property is the reason for it. `apply_decay` multiplies the stored
`strength` while only `potentiate` resets `last_activated`, so applying it
repeatedly compounds:

```
strength 1.0, stability 5, last activated 10 days ago
applied once at day 10        -> 0.135335
applied at day 5 and day 10   -> 0.050000   (STRENGTH_FLOOR)
```

Two installations with identical usage would rank memories differently
depending on maintenance cadence, and the twice-decayed connection sits at
the floor where it cannot be told apart from one nobody has touched in a
year. `apply_decay` is unchanged — its docstring now records the hazard, and
`effective_strength` avoids it with no maintenance pass at all.

Also corrects the module docstring, which says `hallways.py` and
`palace_graph.py` invoke these functions. Both import only
`initialize_dynamics_fields`, and `potentiate`/`apply_decay` have no callers
anywhere in the package.

This deliberately does **not** wire up the missing calls — which access counts
as a co-access seems like a design decision that is yours to make. Happy to
follow up with that half if it would help.

## How to test

```
uv run pytest tests/test_dynamics.py -v
```

Five new tests in `TestEffectiveStrength`: agreement with `apply_decay` for a
single application, that a read leaves the connection untouched, that an
intermediate read does not change a later answer, and the
unparseable-timestamp and floor cases.

The third one is the property `apply_decay` lacks, and it fails if
`effective_strength` is changed to write its result back.

Note for anyone extending these tests: `DEFAULT_STABILITY` is 1.0 day, so a
`_fresh_connection()` decays below `STRENGTH_FLOOR` within a few days and the
floor hides the curve. The elapsed-time test sets `stability` explicitly for
that reason.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — 4828 passed, 32 skipped
- [x] No hardcoded paths

Refs #2417.


scotmcc:
Thanks — both were real.

- `effective_strength` raised `TypeError` on a naive `now`, since `_parse_iso` forces awareness on the stored timestamp but nothing did the same for the argument. It now reads a naive value as UTC, matching what the record gets, with a test that fails against the previous behaviour.
- Docstring fixed: it showed a positional call for a keyword-only parameter.

Full suite still passes (4828 passed, 32 skipped) and `ruff check .` is clean.

PR #2419, opened 2026-09-02
fix(mcp): keep Hub reads responsive during conversation mines

RoshanDev:
## Draft / stacked integration PR

This PR remains Draft while the overlapping four-commit foundation in #2408 is unresolved. It retains the overlapping foundation until #2408 lands; after that, the duplicate changes will be removed before marking this PR Ready.

### Dependency structure

- Includes equivalent cherry-picks of the four fixes from #2408 so the local branch could validate the complete crash-safe re-mine contract.
- Adds defect 5 from #2403 and the follow-up correctness fixes found during repeated current-head Codex reviews.
- After #2408 merges, this branch will be rebuilt on current `develop` and the four duplicate commits will be dropped.
- #2214 remains the separate streaming-normalization fix for very large JSONL memory use.

### Hub read responsiveness and single-writer correctness

- Conversation mines retain one mutation owner for the complete operation while parsing/chunking no longer hold the request-wide exclusive gate.
- Chroma reads use the shared Hub gate; collection open/upsert/update/delete operations use bounded write bursts with queued-reader handoff.
- Other mutations remain serialized, and project/extract modes preserve their existing request-wide locking contract.
- Conversation-mine output is thread-local so concurrent search output cannot capture mine progress or errors.
- Once a live matching Hub is discovered, CLI search fails closed instead of loading a second local HNSW copy after a transient Hub error.

### Crash-safe hook delivery

- Diary checkpoints carry stable idempotency keys through direct, Hub, and daemon paths.
- Exact pending checkpoint payloads are persisted atomically and are deleted only after the durable marker advances.
- SessionEnd drains frozen payloads without discarding blocked or failed captures.
- Transcript follow-ups survive active-job overlap, handoff races, spawn failure, and watcher retry; retry backoff prevents process-spin loops.

### Crash-safe conversation generations

- Changed chunks are staged under content-addressed physical IDs while preserving stable logical drawer IDs.
- A per-source/per-mode commit marker publishes a complete generation atomically.
- Staged and retired generations are excluded before vector, lexical, SQLite/BM25, closet, list, and get limits are applied.
- Cleanup-pending markers force a retry, including no-stale-ID and A→B→A generation cases.

## Historical field evidence (before the develop migration)

- Incident input: one real 423,368,328-byte Codex JSONL; palace contained 937,016 drawers and 19,305 closets.
- A concurrent exact Hub search completed in 21.8 seconds while the same source was mining; CLI peak RSS was 79,496 KiB and output contained no mine progress/error contamination.
- The real mine was deliberately stopped after 11m43s when Windows free memory crossed the safety gate. Post-stop drawer/closet counts were unchanged and HNSW/SQLite divergence remained zero.
- This PR does **not** claim to solve the large-input normalization memory spike. That remains #2214; the reported run did not validate that streaming fix.

## Verification

- Current head: `72f881d`; base includes `develop` at `8dd30cb`.
- Migrated changes to the MCP, CLI, searcher, and palace packages, with MCP regressions in `tests/mcp/`.
- Kept the four #2408 foundation commits separate pending the agreed landing order.
- Latest MCP/search/hooks/publication/miner/Hub tests: **746 passed**. Ruff and format checks pass.
- Later marker-page failures disable dedup. Generation filters group source checks instead of adding a nested guard for every source; SQLite FTS scans bounded windows and refills until enough visible results are found.
- Closet memoization assertions distinguish mine and diary marker reads from source-content fetches, retaining the one-fetch-per-source/group check. Diary marker parsing accepts dict or namespace results so closet mocks cannot crash search.
- Incomplete marker snapshots, logical-ID queries, generation resolution, and lexical hydration propagate explicit request errors. A successful empty lookup and an absent optional SQLite index retain normal not-found/fallback behavior.
- The BM25 CLI renders generation errors and exits nonzero; SQLite commit-state errors use the same API error envelope as other generation reads.
- The initial tokenless-to-tokened transition copies unchanged rows with stored embeddings before publishing; later appends retain existing active row IDs and vectors.
- Editing a generated drawer into a parent chunk group preserves every chunk for get/search/update/delete, including legacy groups that inherited a shared generation identity.
- Interrupted legacy-group shrink reads merge parent-linked siblings so the rewritten prefix remains reachable and later update/delete can clean up the residual tail.
- Singleton-to-group cleanup failures retain parent-linked chunks for get/update/delete. Partial mined-set scans are discarded, including later-page compatibility errors, so cleanup remains retryable.
- Parent-group recovery accepts only genuinely parent-linked rows and deduplicates physical IDs, preserving singleton update behavior and empty-metadata compatibility.
- Search refills matching parent-linked leftovers via get() when HNSW omits a leftover tail after a failed shrink delete. Refill runs after the scored-pool cut, only for mixed leftover tails, and reapplies wing/room/source_file so a moved prefix cannot leak the old-wing tail.
- SessionEnd keeps the final pending checkpoint until the epoch file is written, and uses `session-end:{epoch}:{boundary}` so a crash cannot mint a second diary entry.
- Idempotent diary replacements publish through a commit marker on new physical IDs, so a failed stale-chunk delete cannot keep leftover text in get/read/search.
- Independent full suite on `fbc4940` (Linux / CPython 3.13.5): 5471 passed, 53 skipped; 2 closet-enrichment tests failed because diary commit-marker `get()` was counted as a source hydration and mock namespaces lacked `.get`. Fixed in `72f881d` (those two tests plus the diary leftover case: 7 passed). Coverage 83.02% is the local venv fail-under 85% gate, not the product gate.
- Regression coverage includes publication failures, repeated append without cloning old active rows, retry after transcript changes, retired tokenless rows, marker-read failures, and filtering before result limits.

## Remaining gates before Ready

- Wait for #2408 to land, following the agreed landing order.
- Drop the duplicate foundation changes after that merge.
- Complete CI and a current-head Codex review after the final history change.


RoshanDev:
@codex review

Please review the current stacked draft head with particular attention to lock ordering between _HTTP_MUTATION_LOCK and _HTTP_REQUEST_LOCK, reader progress between conversation-mine write bursts, mutation serialization across the whole mine, thread-local output isolation, watchdog scope, and the live-Hub fail-closed guarantee that must never import the local searcher.

RoshanDev:
Real-palace E2E update on current head 70d6037:

- Started one HTTP convos mine against a stable 423,368,328-byte Codex JSONL in a 937,016-drawer palace.
- A concurrent exact search completed through the same Hub in 21.8s; CLI max RSS was 79,496 KiB and the result contained no mine progress/error text.
- No request-wide mempalace_mine write-stall warning fired during normalize/snapshot work.
- The run was intentionally interrupted after 11m43s when Windows host free RAM crossed the operational gate. Post-stop repair-status remained exactly drawers 937,016/937,016 and closets 19,305/19,305, divergence 0.
- This verifies reader progress, thin-client reuse, output isolation, and interruption safety. It also confirms that #2214 streaming normalization remains a separate required performance improvement for very large JSONL; this PR does not duplicate it.

The Hub is offline after the memory-gate stop. Draft remains not ready.

RoshanDev:
Addressed all four review findings in commit 3c36221:

- FTS5 validation now runs under the mine access write gate; hallway computation remains under the read gate.
- HTTP shutdown now acquires the mutation lock before the request write lock before releasing the Hub writer lease.
- Interleaved write bursts explicitly hand progress to an already-queued reader before reacquiring the next write burst.
- Hook diary forwarding now uses urlopen_with_server_tokens, including its safe pre-dispatch 401 credential rotation.

Regression evidence: 268 focused tests passed (1 skipped); 578 wider Hub/MCP/FTS tests passed; full suite 4871 passed, 31 skipped, with only the previously reproduced clean-develop failure in test_sqlite_integrity_errors_reports_a_path_python_cannot_encode. Ruff and format checks pass for the full repository.

@codex review

RoshanDev:
Addressed all three second-round findings in commit 1c7ff8f:

- Codex v1 and v2 conversation events now go through one shared parser used by normalization, Stop-hook message counting, and checkpoint extraction; current item_completed-only sessions trigger saves correctly.
- _ingest_transcript now returns a distinct disabled status, so hooks.mine_transcript=false cannot acknowledge a failed diary checkpoint or advance its marker.
- Shrink/rewrite passes now withhold the current source-mtime completion marker until orphan deletion succeeds. A transient delete failure stays retryable even with an unchanged source, and the successful retry finalizes without re-embedding unchanged chunks.

Added focused failure/retry and Codex-v2 hook regressions. Validation: 398 parser/hook/convo tests passed (1 skipped); 640 wider Hub/MCP/hook tests passed; full suite 4874 passed, 31 skipped, with only the already reproduced clean-develop path-encoding assertion. Full-repository Ruff and format checks pass.

@codex review

RoshanDev:
Addressed the third-round daemon submission finding in commit 8093e27. A daemon submit exception now returns pending/failed instead of confirmed-dispatched: the hook still avoids an unsafe immediate direct fallback, but it no longer advances the checkpoint marker when the diary also failed. The next Stop retries the dedupe-keyed transcript job.

Added a two-fire regression that simulates the daemon disappearing before enqueue and verifies both attempts retry while the marker remains absent. Current-head focused validation: 419 passed, 1 skipped; full-repository Ruff and format checks pass. The immediately preceding 1c7ff8f head also completed the full suite with 4874 passed, 31 skipped and only the clean-develop path-encoding assertion.

@codex review

RoshanDev:
Addressed all three fourth-round findings in commit e819644:

- Changed conversation chunks now stage under content-addressed physical generation IDs linked to stable logical positions. Existing physical drawers and their verbatim documents remain untouched through every new batch; superseded generations are deleted only after all replacement batches succeed, and crash-left generations are reused on retry.
- Hook-to-Hub diary forwarding now has one 350 ms deadline (50 ms health slice, remaining budget split across token attempts), below the repository 500 ms hook budget instead of the previous 60-second POST timeout.
- Hook checkpoints carry a stable session/content idempotency key through daemon, Hub, and direct routes. mempalace_diary_write maps that key to a deterministic entry and upserts retries, preventing response-loss duplicates while preserving legacy append behavior for calls without a key.

Evidence: old-document equality is asserted after an injected later-batch upsert failure; repeated keyed diary writes produce one readable entry; timeout kwargs are bounded. Validation: 653 core tests passed (1 skipped), 270 routing/HTTP/daemon tests passed, full suite 4877 passed and 31 skipped with only the repeatedly reproduced clean-develop path-encoding assertion. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed all three fifth-round findings in commit d021038:

- Hub response-body reading now runs behind the same absolute monotonic deadline in a daemon reader thread. On deadline expiry the response is closed and forwarding returns pending; a drip-feed can no longer extend the hook one socket receive at a time.
- Keyed diary replacement snapshots every prior direct/chunked physical row, writes the new complete representation first, then deletes prior IDs not used by the replacement. Long-to-short, chunk-size changes, and chunked/un-chunked transitions no longer leave readable suffix rows, while a failed new write still preserves the old representation.
- Codex item_completed text blocks now append their original text instead of per-block stripped text; leading/trailing whitespace and indentation are retained.

Evidence: a deliberately blocking response is forcibly closed under a 50 ms test budget; a chunked keyed diary replaced by a short entry reads as exactly one row; block whitespace equality is exact. Hooks/MCP/Normalizer groups: 739 passed, 1 skipped. Full-repository Ruff/format checks pass. The preceding e819644 head completed the full suite with 4877 passed, 31 skipped and only the clean-develop path-encoding assertion.

@codex review

RoshanDev:
Addressed the sixth-round deadline finding in commit bf523a8. The complete POST response lifecycle—urlopen, response read, and context-manager close—now runs inside one daemon worker. The hook thread owns only the monotonic deadline/event wait and returns pending at expiry without touching the HTTPResponse object, so HTTPResponse.close cannot block it on the reader lock.

The blocking-response regression now explicitly asserts that the hook returned under budget while the response was not synchronously closed by the hook thread. Hooks suite: 163 passed, 1 skipped; full-repository Ruff/format checks pass. The earlier e819644 full-suite evidence remains 4877 passed, 31 skipped with only the clean-develop path-encoding assertion, and d021038 added 739 passing Hooks/MCP/Normalizer tests.

@codex review

RoshanDev:
Addressed the seventh-round health-probe finding in commit 6c61550 by removing the redundant synchronous /healthz request entirely. A live registry match now proceeds directly to the one authenticated, idempotent diary POST; its connection, header parsing, body read, and close all remain inside the deadline-controlled daemon worker, so no urllib response lifecycle can block the hook thread.

Hooks suite remains 163 passed, 1 skipped; full-repository Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the eighth-round daemon replay finding in commit 8d2c0b6. service.run_diary_write now forwards idempotency_key into tool_diary_write, so a daemon recovery replay reaches the deterministic/upsert path instead of creating a fresh timestamp ID. The service forwarding regression asserts the exact key at the final call boundary; daemon/Hook/MCP focused validation: 58 passed. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the ninth-round pending-identity finding in commit 93599ec. Stop-hook diary saves now derive checkpoint_id from the persisted last-save boundary (for example stop:15) and include that boundary—not the moving recent-message window—in the idempotency key. While a write remains unacknowledged the marker does not advance, so later prompts cannot mint a new identity; retries upsert the same logical diary entry.

Regressions append a new user prompt between failed attempts and assert both Stop calls use stop:15; a direct save test also verifies the internal key remains identical while the extracted message count changes. Hooks suite: 163 passed, 1 skipped; full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed both tenth-round findings in commit cf16153:

- Changed physical generations are written with mine_staged=true and excluded from vector results, filtered-query fallback, backend lexical union, sqlite BM25 fallback, CLI rendering, and neighbor hydration. Palace completeness checks also ignore staged rows. Final metadata publication plus superseded-ID cleanup hold one HTTP write gate, so queued readers see the old committed set or the complete new set; injected publication/delete failures restore prior metadata and re-hide new generations for retry.
- Stop checkpoint IDs now include a durable per-session epoch. SessionEnd advances that epoch before clearing the throttle marker, so resuming the same session cannot reuse an earlier stop:...:15 identity or overwrite prior diary history; pending retries within one epoch still share the same boundary/key.

Evidence: real Chroma search excludes a staged sentinel; later-batch failure leaves new generations staged and old verbatim documents unchanged; SessionEnd writes a non-initial epoch and resume identity differs. Searcher/Convo/Hooks groups: 375 passed, 1 skipped. Full suite: 4882 passed, 31 skipped, with only the repeatedly reproduced clean-develop path-encoding assertion. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed all three eleventh-round findings in commit 0a62d9b:

- Drawer vector queries and backend lexical-union queries now add mine_staged != true to the backend where clause before top-K. Legacy rows with no flag remain eligible (verified against Chroma). SQLite BM25 adds a schema-compatible NOT EXISTS predicate before LIMIT; post-filter defenses remain for fallback paths.
- Publication update, superseded-ID delete, and exception rollback now stay inside the same _HTTPMineAccessGate write context. A queued reader cannot enter between a failure and metadata restoration.
- prefetch_content_hashes now ignores staged metadata, preventing an invisible generation from suppressing a committed copy under another path.

Regressions: eight staged exact matches cannot displace one committed top-1 result; a failing delete records exactly one write-gate enter/exit including rollback; staged hashes are absent while committed hashes remain. Searcher/Convo/Palace groups: 309 passed; full Ruff/format checks pass. The preceding cf16153 head completed the full suite with 4882 passed, 31 skipped and only the clean-develop path-encoding assertion.

@codex review

RoshanDev:
Addressed both twelfth-round findings in commit 2902c37:

- Filtered-query fallback now progressively increases its unfiltered candidate request up to collection.count() until enough committed rows survive post-filtering or the collection is exhausted; staged rows can no longer fill the old 500-row cap and hide committed matches.
- Conversation publication now uses a stable per-source commit marker with a generation token. All replacement rows are hidden while embedding batches run; one marker upsert atomically makes the complete token eligible. If the process dies afterward, the marker remains cleanup-pending, so the next mine resumes metadata/stale-ID cleanup; searches can still see the complete new token, never a partially published batch. Normal completion clears pending under the same write gate.

Marker tokens are honored before top-K in vector/lexical where clauses and in sqlite BM25 SQL. Regressions cover pre/post-marker visibility, progressive fallback refill, pending cleanup retry, and schema compatibility. Searcher/Convo/Palace groups: 310 passed; full Ruff/format checks pass. The preceding cf16153 full suite remains 4882 passed, 31 skipped with only the clean-develop path-encoding assertion.

@codex review

RoshanDev:
Addressed both thirteenth-round findings in commit e3768dd:

- An existing mine_cleanup_pending marker now routes through generation finalization even when stale_ids is already empty. If stale deletion succeeded but the final marker update was interrupted, the next unchanged mine clears pending instead of reparsing forever.
- SQLite BM25 commit-token visibility now joins marker metadata through its embedding, segment and collection and requires marker_collection.name == the selected collection. A same-token marker in another Chroma collection cannot publish target rows.

Regressions cover pending-only retry to completed and a two-collection SQLite database where a foreign marker stays hidden until a target-collection marker is added. Searcher/Convo groups: 312 passed; full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the fourteenth-round mode-scoping finding in commit b893767. prefetch_mined_set now applies _metadata_matches_extract_mode to cleanup-pending commit markers before excluding a source. A pending general-mode generation no longer defeats the complete exchange-mode bulk skip (and vice versa).

The two-mode regression asserts exchange returns its stored mtime while general remains pending. Convo/Palace groups: 94 passed; full Ruff/format checks pass.

@codex review

RoshanDev:
@codex review

Retrying the same current head after the prior reviewer fetch failed with Provided git ref ... does not exist. The PR head is now visible as b89376787ee0e67fd3faff6680bb8fb5b7cbbb84.

RoshanDev:
Addressed all four fifteenth-round findings in commit a6bd347:

- Stop checkpoints now atomically persist the exact pending entry, message window, themes and idempotency key; retries load that frozen payload until acknowledgement, so later prompts cannot mutate an ambiguously committed checkpoint. Successful diary writes (or a separately successful transcript dispatch that advances the marker) remove the pending file.
- A live same-target mine now creates/updates one durable pending command and one detached watcher. The watcher waits for the active snapshot to exit, atomically claims its slot, and starts a latest-state follow-up; repeated hooks coalesce rather than stack concurrent mines.
- list_drawers excludes commit markers and uncommitted generations before totals/pagination while retaining token-committed generations.
- A transcript that vanishes, rotates, becomes unreadable, or falls below the ingest gate reports undispatched instead of acknowledging a failed interval.

Regressions cover frozen payload equality after appending a prompt, pending watcher persistence/uniqueness, list visibility/totals, and vanished-source status. Hooks/MCP/Daemon groups: 650 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed all three sixteenth-round findings in commit 78cfcd4:

- Creating a checkpoint boundary now aborts before any diary dispatch unless its frozen pending payload was durably written and atomically renamed. A full/read-only hook_state cannot create an unretryable ambiguous write.
- The queued-mine watcher now treats MEMPALACE_MINE_TIMEOUT_HOURS=0 as truly unbounded: no monotonic deadline is created, and it waits until the active PID exits before launching the latest-state follow-up.
- list_drawers obtains generation commit tokens independently of wing/room filters. Chroma uses a dedicated indexed sqlite token query, so room-scoped listings retain committed generations without adding a palace-wide metadata/document scan.

Regressions cover freeze-write failure/no dispatch, timeout-zero watcher behavior, room-scoped generation listing, and the original SQL filter-pushdown/performance contract. Hooks/MCP/Daemon groups: 652 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed both seventeenth-round findings in commit 9497063:

- Frozen checkpoint payloads are no longer deleted inside _save_diary_direct after storage acknowledgement. hook_stop removes them only in the successful else branch of the durable last_save write; if that marker write fails or the process exits first, the exact payload remains for the same-boundary retry.
- logical_drawer_id now round-trips through search results, list collapsing/previews, and MCP get/update/delete resolution. Physical old/new generations collapse to the newest visible logical row; pre-re-mine deterministic references continue resolving after the physical ID changes.

Regressions cover last-save write failure retaining pending state, exact payload retry, logical search IDs, room-scoped list/get with physical preview hydration, and pending old/new dedup. Hooks/MCP/Searcher groups: 757 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed all three eighteenth-round findings in commit 233d073:

- list_drawers now applies marker/staged visibility to physical rows before logical collapse, so an unpublished newer row cannot displace the still-valid old row and then disappear.
- Search candidates carry internal physical/logical generation IDs and collapse to the newest visible physical row per logical ID before ranking/limiting; union and sqlite BM25 paths use the same collapse and strip internal fields before returning.
- MCP logical generation resolution returns newest content/metadata but retains every visible physical ID. delete removes all old/new rows; non-chunked update writes the newest and deletes superseded siblings, preventing logical drawers from reappearing.

Regressions cover pre-marker old-row listing, post-marker one-hit search, logical get with physical preview hydration, and delete of both old/new physical rows. Hooks/MCP/Searcher groups: 757 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the nineteenth-round pre-top-k generation finding in commit 50efa3d. Each candidate page now batch-resolves every encountered logical ID to its newest visible physical generation. Non-current rows are filtered before ranking; if that leaves fewer candidates than requested and the backend page was full, the vector query expands geometrically until enough current logical rows are present or the collection is exhausted. CLI and programmatic searches share the same query/refill helper; union and sqlite BM25 retain logical collapse before their final limits.

Regression: an older physical row that ranks first causes limits [1, 2], then only the lower-ranked current generation is returned. Searcher/MCP groups: 513 passed. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the twentieth-round lexical top-k finding in commit 2419761. Before lexical-union merge, the returned logical IDs are batch-resolved to newest visible physical generations. If the backend only returned an old row, the current row is hydrated directly and its current document is locally BM25-scored; no longer-matching current text is dropped, while matching current text replaces the stale hit. Distance gating and later logical collapse operate on the resolved hits.

Regression: an old high-score lexical hit absent from the current backend page resolves to the lower-ranked current physical row and current verbatim document. Searcher suite: 107 passed. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the twenty-first-round active-generation finding in commit 4262e27. Finalized rows retain mine_generation_token, and current-generation resolution now orders by membership in the per-source committed token before considering filed_at/physical ID. This handles A→B→A reversions even when stale B cleanup fails: A is authoritative despite its older ingestion timestamp. Vector refill, lexical replacement, sqlite BM25 collapse, list collapse, and MCP logical get/update/delete use the same active-token priority.

Regressions deliberately give the active A row an older filed_at than stale B and verify search/list/get still choose A. Searcher/MCP/Convo groups: 608 passed. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed both twenty-second-round fallback findings in commit 5c1dc97:

- SQLite BM25 candidates with logical IDs are resolved through a collection-scoped marker/token SQL lookup before local BM25 ranking. A stale FTS hit is replaced by the active physical row and rescored from its current document; if the current document no longer matches, the stale candidate is dropped.
- Closet neighbor hydration now retains physical IDs, applies visibility, and collapses each logical generation by active commit token before chunk ordering/query-term passage selection. Obsolete siblings cannot replace an active hit text.

Regressions cover a stale FTS-only B hit whose active A no longer matches (zero results) and closet source rows where older-filed active A beats newer stale B. Searcher suite: 108 passed. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed both twenty-third-round removed-generation findings in commit 7ba9423:

- Token-bearing logical rows whose generation token is no longer present in any current commit marker are now excluded from current-generation resolution and physical-row collapse. Legacy logical rows with no generation token remain compatible.
- SQLite BM25 distinguishes legacy unversioned candidates from retired tokened rows: if a logical ID has no active marker-selected row, a retired candidate is dropped rather than preserved. Closet source hydration applies the same global active-token rule, so orphaned removed chunks cannot enter passage selection/concatenation.

Regressions cover an FTS-only removed logical chunk (zero results) and a closet source with an orphaned third logical row (only active A remains). Searcher suite: 108 passed. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed both twenty-fourth-round retired-generation findings in commit 75cc516:

- Lexical-union resolution no longer returns early on an empty current map. Logical hits without a current physical row are retained only when they are legacy/unversioned; token-bearing retired hits are dropped.
- MCP generation lookup and list visibility now reject finalized token-bearing rows whose token is absent from current commit markers, while preserving rows with no generation token for backward compatibility.

Regressions cover an orphan lexical hit with no current row (empty) and room-scoped list/get containing a retired finalized logical row (hidden/error). MCP/Searcher groups: 515 passed. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed all three twenty-fifth-round findings in commit 66c5e36:

- Pending checkpoint files now record session identity. SessionEnd discards them and advances epoch/clears last_save only after a final diary or transcript capture is confirmed; an unconfirmed final flush leaves pending payload, epoch and marker intact for resume.
- SQLite BM25 candidate SQL now rejects any token-bearing row unless a same-collection commit marker currently names that token, before FTS LIMIT. Retired rows cannot occupy the bounded pool.
- Backend lexical-union fetch now resolves generations on each page and grows n_results geometrically until the resolved pool reaches 3K, the backend returns a short page, or the collection is exhausted.

Regressions cover failed SessionEnd preservation, max_candidates=1 retired-vs-ordinary SQLite selection, and lexical limits [3, 6] refilling after three retired hits are dropped. Hooks/Searcher groups: 279 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the twenty-sixth-round SessionEnd finding in commit a2cb7f7. Pending files record session/checkpoint identity and SessionEnd now replays each frozen payload by its original checkpoint_id before creating the final rolling snapshot. Each pending file is deleted only after that exact write is confirmed; any unconfirmed payload remains, preventing epoch advance and last_save cleanup in finally. A successful unrelated final diary or transcript ingest can no longer acknowledge it.

Regression freezes messages 1–15, appends through 35, and verifies SessionEnd pending flush writes the byte-identical frozen entry before deleting it; the failure regression retains pending/epoch/marker. Hooks suite: 171 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the twenty-seventh-round final-snapshot finding in commit ea9eb45. SessionEnd final rolling snapshots now use their own stable session-end:<epoch> checkpoint_id and are frozen through the same pending-payload mechanism. A successful exact final write removes that pending file; failure leaves it, which prevents epoch advancement and last_save cleanup in finally. Earlier Stop payloads are still flushed individually first and cannot be acknowledged by the unrelated final snapshot.

Regression: a 35-message SessionEnd final diary failure leaves last_save intact, epoch=initial, and one pending payload whose checkpoint_id is session-end:initial. Hooks suite: 172 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the twenty-eighth-round final-identity/cleanup findings in commit 9d91066. Every SessionEnd rolling snapshot now uses a distinct frozen ID session-end:<epoch>:<exchange_count>:<uuid>, so flushing an older final payload can never be overwritten by the next window under the same key. Session cleanup is governed by explicit session_end_complete: once a valid transcript capture begins, cleanup requires all prior pending payloads plus either the exact final diary or transcript ingest to succeed; inability to persist the final pending payload leaves last_save/epoch intact. No-transcript/invalid-input no-op paths retain their historical cleanup behavior.

Regressions cover a failed 35-message final snapshot remaining pending and a capture exception preserving the marker, while no-transcript/parse-failure paths still clean up. Hooks suite: 172 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the twenty-ninth-round blocked-routing finding in commit a722789. As soon as SessionEnd validates a real transcript, session_end_complete is set false before resolving/entering the write-routing context. A require-unavailable or invalid-routing return therefore preserves epoch and last_save; no capture attempt is falsely finalized. No-transcript and invalid-input no-op paths retain historical cleanup.

Routing/Hook groups: 192 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed both thirtieth-round findings in commit 3c05e0a:

- content-hash prefetch now collects commit markers first, then indexes only legacy rows or rows whose generation token is currently committed. Finalized retired hashes cannot suppress a visible copy under another path.
- SessionEnd pending flush returns the exact confirmed checkpoint IDs. The final snapshot uses session-end:<epoch>:<count>:<uuid>; if the same epoch/count final was just replayed, no second write is created. A later boundary gets a distinct identity. Cleanup still requires all pending payloads plus final diary/transcript confirmation.

Regressions cover staged/retired/active/legacy hash selection and replay of a pending final at an unchanged 15-message boundary producing exactly one write. Hooks/Convo groups: 267 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed all three thirty-first-round findings in commit 5105402:

- The follow-up watcher atomically renames pending.json to a per-watcher claimed file before reading/spawning. A Stop arriving during handoff writes a new pending.json that is never unlinked with the consumed request; the same watcher loops behind the spawned mine, with an ownership-aware watcher-file handoff closing the final exit race.
- _file_chunks_locked reads cleanup-pending before the file_already_mined fast path. Pending sources bypass the complete-mtime shortcut and retry stale cleanup/marker completion; marker read failures safely skip writes.
- Pending replay records checkpoint_id in flushed_ids immediately after storage confirmation, independently of local unlink success. SessionEnd will not duplicate the same final boundary even when hook-state cleanup fails, while the surviving file still blocks epoch cleanup.

Hooks/Convo groups: 267 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the thirty-second-round claimed-followup finding in commit 643d8ff. The watcher now atomically renames pending to a claimed file, and every read/slot/log/Popen failure restores it to pending unless a newer pending request already exists (the newer request subsumes the old snapshot). The ownership-aware finally block removes only its own watcher slot and automatically starts a replacement watcher whenever pending remains, so recovery does not depend on another Hook fire.

Regression injects a missing-executable Popen failure and verifies pending is restored plus a replacement watcher is spawned immediately. Hooks suite: 174 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the thirty-third-round watcher-spin finding in commit 34acdae. Spawn/read/slot/log failures restore claimed work with watcher_attempt incremented. Replacement watcher launch waits exponential backoff 1, 2, 4... seconds capped at 60; replacement preserves the attempt count, while a genuinely new Hook request overwrites pending with attempt=0. The pending request remains durable throughout, but persistent failures cannot create a high-rate process storm.

Regression injects a missing executable, verifies claimed→pending restoration, attempt=1, one-second backoff, and replacement watcher launch. Hooks suite: 174 passed, 1 skipped. Full Ruff/format checks pass.

@codex review

RoshanDev:
Addressed the current-head full-suite regressions in the latest commit: generation commit markers are now read once per search request and passed through filtered-query fallback plus closet enrichment, avoiding redundant collection reads while preserving crash-safe visibility. Test doubles now implement the commit-marker lookup contract. Validation: focused search/miner suite 249 passed; full suite 4903 passed, 31 skipped, 106 deselected, with only tests/test_repair.py::test_sqlite_integrity_errors_reports_a_path_python_cannot_encode failing (independently reproduced on pristine origin/develop under CPython 3.12.11). Full Ruff and format checks pass. @codex review

RoshanDev:
Final validation snapshot for 37e9f89: Codex review completed with no new findings, and all 10 current-head CI checks pass, including Windows (the previous Windows failures were exactly the three regressions fixed in 37e9f89). Local full suite: 4903 passed, 31 skipped, 106 deselected; the sole CPython 3.12 repair-test failure is independently reproducible on pristine origin/develop. Keeping this PR Draft while stacked dependency #2408 remains open.

igorls:

RoshanDev:
Thanks for the heads-up. Rebuilt this branch on `develop` at `df330d8`; the new head is `10c2e64`.

The changes now live in the corresponding MCP and CLI package modules, with MCP tests in `tests/mcp/test_write_tools.py` and `tests/mcp/test_diary.py`. I also migrated the searcher and palace changes to their new packages and updated the scoped-search fallback regression to check that staged content stays hidden.

On a clean worktree with CPython 3.13.5: **5,361 tests passed, 53 skipped; coverage 82.69%**. Ruff 0.16.6, format, and diff checks pass. CI is running for the new head.

I have kept this Draft and retained the four #2408 foundation commits separately, following the agreed order of landing #2408 first and then dropping the overlap here.

@codex review

Please review the current head, especially the package migration, Hub lock ordering and reader progress, generation visibility across search paths, and hook checkpoint delivery.

RoshanDev:
Fixed the reused-row publication window from https://github.com/MemPalace/mempalace/pull/2419#discussion_r3998799486 in `6fb2cd5`.

Reused chunks now stage under generation-specific physical IDs with their existing embeddings before the commit marker changes. This preserves the old readable generation before publication and the complete new one afterward. Fault-injection regressions cover failures before and after every publication write, retry, and A→B→A; unchanged chunks are not re-embedded.

Clean-worktree validation on CPython 3.13.5: **5,364 passed, 53 skipped; coverage 82.73%**. Ruff and format checks pass. The preceding migration head passed all 13 CI jobs; CI is now running again for this fix.

@codex review

Please recheck the generation-publication fix on the current head.

RoshanDev:
Addressed both follow-ups in `82ea526`:

- Appended rows now carry a generation token, so a later shrink with failed cleanup cannot expose retired appended content as legacy data. Regression coverage exercises initial ingest → growth → shrink, cleanup failure, and retry.
- Publication keeps mutation serialization while releasing the reader access gate between bounded metadata updates and stale-row deletion batches. The concurrency regression checks reader progress between batches.

Clean-worktree CPython 3.13.5 validation: **5,367 passed, 53 skipped; coverage 82.75%**. Ruff/format checks pass. CI is running on the new head.

@codex review

Please recheck these two fixes on the current head.

RoshanDev:
Fixed the append-growth performance finding in `c7f7277`.

Append-only growth now reuses the active generation token. Existing physical rows and embeddings stay in place; only the new tail is inserted, with nonlegacy generation membership. Rewrites and shrinks retain copy-on-write publication. The new repeated-append regressions check that old active rows are neither upserted nor deleted, alongside the existing failure/retry and reader-progress tests.

Clean-worktree CPython 3.13.5 validation: **5,373 passed, 53 skipped; coverage 82.76%**. Ruff/format checks pass. CI is running on the new head.

@codex review

Please recheck the append-only growth fix on the current head.

RoshanDev:
Fixed the retired-row append predicate in `eef2911`.

Token reuse now requires the matching row to belong to the committed active generation. Retired tokenless matches are rejected. The regression covers legacy A → failed A→B cleanup → failed B→A+append cleanup and checks that the intended text is readable immediately and after retry. The repeated-append efficiency regressions still pass.

Independent focused validation: **88 passed** on CPython 3.13.5; Ruff/format checks pass. The full suite and CI are running for this head.

@codex review

Please recheck the active-row token reuse fix on the current head.

RoshanDev:
Fixed the active-token lookup in `96b77f5`.

The planner now uses the commit marker as the authority for the active generation. Residual T1 rows after a committed T2 rewrite no longer force whole-transcript copies on subsequent appends or same-set cleanup retries. Retired rows remain excluded from token reuse.

Regressions cover repeated appends while cleanup fails, stable IDs and embeddings for unchanged active rows, correct visible content, and eventual cleanup. Independent focused validation: **90 passed**; Ruff/format checks pass. Full-suite and CI validation are running for this head.

@codex review

Please recheck the commit-marker token lookup on the current head.

RoshanDev:
Fixed the remaining residual-row subset check in `6974cf1`.

Both the append-only subset check and candidate membership now use the same authoritative active-generation filter. Retired logical IDs left by failed shrink cleanup are excluded. The new regressions cover repeated appends after such a shrink, unchanged active row IDs/vectors, current visible content, and eventual cleanup.

Independent focused validation: **92 passed** on CPython 3.13.5; Ruff/format checks pass. Full-suite and CI validation are running for this head.

@codex review

Please recheck the append predicate’s active-generation filtering on the current head.

RoshanDev:
Fixed the staged-append visibility issue in `b598868`. New tail rows now use an unpublished token until every batch is staged; publication exposes the complete tail while preserving existing row IDs and vectors. Regressions cover reads between batches, interrupted staging and publication, retry, and repeated appends.

Also merged `develop` at `8dd30cb`, retaining the caller-relative daemon source fix and resolving the changelog conflict. Focused validation: **98 passed**; Hub-related validation: **236 passed**; Ruff and format checks pass. The full suite and CI are running for this head.

@codex review

Please recheck append publication and recovery on the current head.

RoshanDev:
Fixed interrupted-append recovery in `8dae65e`. Reused pending rows are now restaged under the current unpublished token before publication. Rows already visible through an earlier tail marker are copied with their stored embeddings so readers retain a complete view during the transition.

Regressions cover growth between retries, changed/shrunk pending tails, failures around publication, and reader handoff. Validation: **31 publication regressions** and **106 Hub-related tests passed**; Ruff and format checks pass. The full suite and CI are running for this head.

@codex review

Please recheck recovery when a transcript changes between interrupted append attempts.

RoshanDev:
Current-head full suite on Linux / CPython 3.13.5: **5,394 passed, 53 skipped**, coverage **82.83%** (`8dae65e`).

The last review job reported Failed without findings. Retrying the review on the same head.

@codex review

RoshanDev:
Fixed both findings in `899991e`.

Content-hash prefetch now fetches commit markers first and processes drawer metadata in 1,000-row pages, releasing each page instead of retaining the whole palace. Regression coverage checks bounded metadata retention and generation filtering.

Rewrite/shrink publication now neutralizes a leftover tail marker in the same upsert as the primary marker switch. This preserves the old complete view before publication and prevents removed tail rows from remaining visible if cleanup fails.

Validation: **113 publication/miner tests** and **201 Hub/search tests passed**; Ruff and format checks pass. The full suite and CI are running on this head.

@codex review

RoshanDev:
Fixed both follow-ups in `62b59ab`. Content-hash prefetch now excludes tokenless predecessors once their source/mode has a committed generation, preventing a retired row from suppressing ingestion in another file. Paging remains bounded.

Non-tail publication now neutralizes the deterministic tail marker in the primary publication upsert without depending on a marker read. Regressions cover the combined marker-read and stale-delete failures.

Validation: **117 publication/miner tests** and **222 Hub/search tests passed**; Ruff and format checks pass. Full-suite and CI validation are running on this head.

@codex review

RoshanDev:
Fixed the three findings in `54865de`.

- Incomplete marker pagination disables cross-file hash dedup for that pass.
- Source/mode marker filtering now excludes tokenless predecessors across search, backend, and MCP read paths.
- An unreadable tail marker causes conservative copying of potentially visible staged rows before restaging, preserving the prior view.

Regressions cover marker-page failures, initial-generation shrink with failed cleanup, and unreadable tail-marker recovery. Independent Ruff, format, and diff checks pass. The full suite and CI are running for this head.

@codex review

RoshanDev:
Fixed the two review findings and the CI failures in `19c0236`.

Later-page `TypeError` now marks a marker scan incomplete. Source filtering uses grouped membership checks instead of one nested query guard per source; SQLite FTS uses bounded windows and refills after filtering so retired rows do not consume the result limit.

The closet tests now distinguish marker metadata reads from source-content fetches and still verify that each source/group is fetched once. Validation: **21 targeted regressions** and **526 Hub/search tests passed**; Ruff and format checks pass. Full-suite and CI validation are running on this head.

@codex review

RoshanDev:
Fixed both error-propagation findings in `a55beac`. Incomplete marker snapshots and failed generation-resolution reads now produce explicit request errors instead of silently dropping conversation hits. Regression coverage exercises these failures through search and logical drawer reads.

The Chroma regression retains all 80 sources and its visibility assertions. Its marker documents are now distinct, avoiding the identical-vector HNSW fixture issue that caused the CI failure.

Validation: **120 searcher tests** and **535 Hub/search tests passed**; Ruff and format checks pass. Independent full-suite validation on Python 3.13.5 and CI are running for this head.

@codex review

RoshanDev:
Fixed the logical-ID and SQLite BM25 error paths in `0d2c44a`. Failed logical-ID queries now produce explicit get/update/delete errors; failed SQLite generation resolution also produces a search error. Successful empty lookups and an absent optional SQLite index retain their normal behavior.

The related lexical-hydration failure path now propagates an error as well. Validation: **16 fault/Chroma regressions** and **597 search/Hub/MCP/HNSW tests passed**; Ruff and format checks pass. Full-suite and CI validation are running for this head.

@codex review

RoshanDev:
Fixed both CLI/API propagation findings in `662b345`. The BM25 CLI now prints the search error and exits nonzero, while SQLite commit-state read failures return the same explicit API error envelope as other generation-read failures.

Validation: **9 CLI/marker/Chroma regressions** and **602 search/Hub/MCP/HNSW tests passed**; Ruff and format checks pass. Full-suite and CI validation are running for this head.

@codex review

RoshanDev:
Fixed the initial tokenless-to-tokened transition in `0a3c625`. Unchanged rows are copied with their stored embeddings before the first generation marker is published, preserving complete views across publication failures. Later appends retain active row IDs and vectors.

Regressions cover append, partial rewrite, shrink, interrupted retries, and marker/update failure boundaries. Validation: **261 publication/miner/search/MCP tests passed**; Ruff and format checks pass. Full-suite and CI validation are running for this head.

@codex review

RoshanDev:
Fixed long edits of generated drawers in `f492ac8`. Parent chunks no longer share a competing generation identity. Get reassembles their full verbatim content in chunk order, and search retains each chunk. Existing groups that inherited the old identity are handled as groups too.

Regressions cover long update, complete readback, per-chunk search, subsequent update/delete, and legacy groups. Validation: **445 MCP/search/diary/publication/Hub tests** and **81 miner/chunk/generation tests passed**; Ruff and format checks pass. Full-suite and CI validation are running for this head.

@codex review

RoshanDev:
Fixed interrupted legacy-group shrink in `f309671`. Logical reads now merge parent-linked siblings before returning legacy generation matches, keeping the rewritten prefix reachable if stale-tail deletion fails. Subsequent update/delete sees the full physical ID set and can clean up the residual tail. Successful shrink still returns only the new content.

Validation: **528 MCP/search/diary/publication/miner/Hub tests passed**; Ruff and format checks pass. Full-suite and CI validation are running for this head.

@codex review

RoshanDev:
Fixed both interruption paths in `14e397c`. Generation lookup now probes parent-linked siblings even when the surviving match is a singleton, so failed singleton deletion does not hide the new chunks or strand them during later cleanup.

An incomplete mined-set scan is now discarded, including later-page `TypeError`, so an unread cleanup marker cannot cause the source to be skipped. Regressions cover failed pages and cleanup markers on the final page.

Validation: **578 MCP/search/miner/publication/Hub tests passed**; Ruff and format checks pass. Full-suite and CI validation are running for this head.

@codex review

RoshanDev:
The full suite caught singleton compatibility regressions after the sibling-probe change. Fixed them in `fe4c54c`: only genuinely parent-linked rows enter a chunk group, and physical IDs are deduplicated. Empty metadata, last-modified stamping, and ordinary singleton updates retain their prior behavior.

Validation: **616 MCP/search/transport/publication/miner/Hub tests passed**; Ruff and format checks pass. The transport timing test passed 16 isolated repetitions without loosening its assertions. Full-suite and CI validation are running for this head.

The previous head received a clean Codex review; requesting a recheck of this compatibility fix.

@codex review

RoshanDev:
Linux CI on `fe4c54c` failed `test_interrupted_legacy_shrink_readback_keeps_prefix_and_stale_tail`: get() already recovered the leftover tail, but default MCP search missed `shrinkGamma`. Chroma HNSW can omit that leftover after the prefix upsert. macOS/Windows were green, and the same head had a clean Codex review.

Fixed in `eda9374`. After generation collapse, search now `get()`s parent-linked siblings and adds leftover chunks whose verbatim text matches the query. A failed sibling lookup leaves the existing hits unchanged.

Validation: **567 MCP/search/transport/publication/miner/Hub tests passed**; the shrink search test passed 20 isolated repetitions; new HNSW-omit and failed-get unit tests pass. Ruff and format checks pass. Full-suite and CI validation are running for this head.

@codex review

RoshanDev:
Addressed both leftover-refill findings in `7993b8e`:

- Scoped searches reapply wing/room/source_file to refilled siblings, so a failed shrink-plus-move cannot leak the old-wing tail into the new wing.
- Refill now runs after the pre-enrichment cut and only for mixed leftover tails (prefix stripped of `logical_drawer_id`, unread tail still carrying it). Ordinary parent groups stay out of the scored pool.

Validation: **572 MCP/search/transport/publication/miner/Hub tests passed**. Ruff and format checks pass. Full-suite and CI validation are running for this head.

@codex review

RoshanDev:
Addressed both leftover-refill follow-ups from `7993b8e` in `fbc4940`:

- SessionEnd now uses `session-end:{epoch}:{boundary}` and keeps that pending file until the epoch is written. A crash before the epoch advance retries the same checkpoint instead of minting a new UUID.
- Idempotent diary replacements write new physical IDs first, then publish a commit marker. Get/read/search hide leftover tails if the old-chunk delete fails.

Validation: **746 MCP/search/hooks/publication/miner/Hub tests passed**; the new SessionEnd and diary leftover tests pass. Ruff and format checks pass. Full-suite and CI validation are running for this head.

RoshanDev:
Pushed `72f881d` after the independent full suite on `fbc4940`.

5471 passed, 53 skipped. The two failures were `test_closet_enrichment_memoises_by_source_and_parent_group` and `test_search_promotes_distinct_results_from_wider_pool_and_fetches_source_once`: diary commit-marker `get()` was treated as a source hydration, and the closet mock’s `SimpleNamespace` has no `.get`. Marker reads now parse dict or namespace inside the same failure handler; the memoization helper ignores `diary_commit` gets the same way it already ignores mine markers.

Draft stays Draft until #2408.

RoshanDev:
@igorls Status after the package-split rebase you flagged.

This branch is on current `develop` (includes `8dd30cb`). The unique work is in the split MCP/CLI/searcher/palace packages; MCP tests are under `tests/mcp/`. Head is `72f881d`.

I am keeping this Draft until #2408 lands. After that I will rebuild on the new `develop` and drop the four overlapping foundation commits.

Linux/macOS CI is green. Independent Linux suite: 5471 passed, 53 skipped, then two closet-enrichment failures from diary-marker `get()` were fixed in `72f881d`. Windows still fails `test_hybrid_search_enrichment_populates_drawer_index_and_total` with an empty result list.

#2214 stays the separate large-JSONL streaming fix.

Issue #2028, opened 2026-07-15
ChromaBackend._client() inode/mtime reopen rebuilds without a chromadb system-cache reset (parallel to #2002)


colorpanda82:
Follow-up from the review of #2026 (fix for #2002). Non-blocking on that PR; filing separately per the reviewer's suggestion rather than expanding it.

## Context

#2026 fixes #2002 on the MCP path: `mcp_server._get_client()` now calls `_force_chroma_cache_reset()` **before** `ChromaBackend.make_client()` whenever an inode/mtime change is detected. That drops chromadb's path-keyed `SharedSystemClient` cache (and the live HNSW segment) so the client is rebuilt from the on-disk state, instead of persisting a stale segment over a peer writer's changes and driving the persisted count backwards.

## The gap

`ChromaBackend._client()` (`mempalace/backends/chroma.py`, ~L2000-2064 on `develop`) has its **own** inode/mtime freshness check and reopen path, independent of the MCP `_get_client`:

```python
# chroma.py:2044
if cached is None or inode_changed or mtime_changed or mtime_appeared:
    ...
    ChromaBackend._prepare_palace_for_open(palace_path)
    cached = chromadb.PersistentClient(path=palace_path)   # L2058
    self._clients[palace_path] = cached
```

On an inode/mtime change it discards the quarantine gate, runs `_prepare_palace_for_open`, and reconstructs `chromadb.PersistentClient(path=...)` — but it does **not** call `_force_chroma_cache_reset()` first. `_force_chroma_cache_reset` does not appear anywhere in `chroma.py`. Because chromadb caches its `System` keyed by path, the reconstructed client can be handed back the same stale System/HNSW segment #2002 describes, just reached through `_client()` instead of `_get_client`.

## Who hits it

The MCP server goes through `_get_client`, so the shipped fix covers it. Any **long-lived non-MCP caller** that holds a `ChromaBackend` instance across a palace rebuild or peer in-place write (inode swap or mtime bump) and reads through `_client()` would be exposed:

- direct library users of `ChromaBackend`
- CLI / job paths that construct a backend once and reuse it
- the sidecar / daemon path (worth confirming)

## Suggested direction (not scoped yet)

[...]

PR #2421, opened 2026-09-02
fix(chroma): release systems before cache reset

### LINKED ISSUE #2375: _client() clears chromadb's System cache before closing the displaced client, so close() silently no-ops and each rebuild orphans an HNSW segment set
### Summary

In `ChromaBackend._client()` (`mempalace/backends/chroma.py`, v3.8.0), `_clear_chroma_system_cache()` runs on line 2620, **before** `_close_client(self._clients.pop(palace_path, None))` on line 2632.

`SharedSystemClient.clear_system_cache()` empties `_identifier_to_system` and `_identifier_to_refcount`. The subsequent `close()` therefore cannot find its identifier, `_decrement_refcount` reports zero, the `pop` returns `None`, and **`System.stop()` is never called**. `close()` returns cleanly and sets `_closed = True`, so nothing looks wrong — but the native HNSW segment memory the comment on line 2622 is explicitly trying to release is not released.

Every external-change rebuild therefore orphans one HNSW segment set. The comment directly above the call states the intent:

> Each live `PersistentClient` pins its own copy of every HNSW segment it has opened (…) and neither dict eviction nor `_clear_chroma_system_cache()` returns that native memory. Dropping it here keeps a long-lived server flat across rebuilds instead of accumulating one orphaned index set per external change.

That is exactly the right goal. The ordering silently defeats it.

### Environment

- mempalace **3.8.0** (latest tag), chromadb **1.5.8**, Python 3.12.3, Linux x86-64
- Backend: `chroma`, rollback-journal mode

### Reproduction

Pure chromadb, random vectors, no mempalace and no external data. The only difference between the two runs is the order of the two calls.

```python
"""repro-order.py — clearing chromadb's System cache before closing the displaced
PersistentClient makes close() a silent no-op.

Usage:  python repro-order.py clear_then_close     # the order shipped in 3.8.0
        python repro-order.py close_then_clear
"""
import gc, os, shutil, sys, tempfile
import numpy as np
import chromadb
from chromadb.api.client import SharedSystemClient
from chromadb.config import System

N, DIM, REBUILDS = 100_000, 384, 5
ORDER = sys.argv[1] if len(sys.argv) > 1 else "clear_then_close"

def rss_mb():
    for line in open(f"/proc/{os.getpid()}/status"):
        if line.startswith("VmRSS:"):
            return int(line.split()[1]) / 1024
    return 0.0

STOPS = {"n": 0}
_orig_stop = System.stop
def counting_stop(self, *a, **k):
    STOPS["n"] += 1
    return _orig_stop(self, *a, **k)
System.stop = counting_stop

path = tempfile.mkdtemp(prefix="chroma-repro-")
try:
    c = chromadb.PersistentClient(path=path)
    col = c.create_collection("vectors")
    rng = np.random.default_rng(0)
    for i in range(0, N, 5_000):
        col.add(ids=[str(i + j) for j in range(5_000)],
                embeddings=rng.random((5_000, DIM), dtype=np.float32).tolist())
    del col, c
    SharedSystemClient.clear_system_cache(); gc.collect()
    print(f"order={ORDER}  N={N} dim={DIM}  chromadb={chromadb.__version__}")
    print(f"{'rebuild':>8} {'RSS MB':>9} {'stops':>6}")

    client = None
    for i in range(REBUILDS + 1):
        if client is not None:
            if ORDER == "clear_then_close":          # 3.8.0 order
                SharedSystemClient.clear_system_cache()
                client.close()
            else:
                client.close()
                SharedSystemClient.clear_system_cache()
            client = None
        client = chromadb.PersistentClient(path=path)
        q = client.get_collection("vectors")
        q.query(query_embeddings=[[0.01] * DIM], n_results=3)   # materialise the segment
        del q
        print(f"{i:>8} {rss_mb():>9.0f} {STOPS['n']:>6}")
finally:
    shutil.rmtree(path, ignore_errors=True)
```

### Measured

```
order=clear_then_close  N=100000 dim=384  chromadb=1.5.8
 rebuild    RSS MB  stops
       0       436      0
       1       627      0
       2       817      0
       3      1008      0
       4      1205      0
       5      1395      0

order=close_then_clear  N=100000 dim=384  chromadb=1.5.8
 rebuild    RSS MB  stops
       0       443      0
       1       438      1
       2       451      2
       3       465      3
       4       467      4
       5       467      5
```

Shipped order: **+959 MB across 5 rebuilds (~192 MB each), linear, `System.stop()` never called.**
Swapped order: **+24 MB total, `stop()` called on every rebuild.**

At 384 dimensions each orphaned segment is ~190 MB; on a wider embedding it scales with `max_elements * size_data_per_element`.

### Impact in practice

The rebuild fires on any external inode/mtime change, which is the normal case whenever a second process writes to the palace (a sweep, a mine, a CLI call) while a long-lived MCP server is up. On a live server we measured each such rebuild costing ~6.5–7.3 s and several hundred MB of anonymous RSS, against 0.9 s and ~0 MB for a search with no intervening write.

The memory is not permanently lost — it is unreachable only until a gen-2 collection; an explicit `gc.collect()` followed by `malloc_trim(0)` recovers most of it. But nothing on the hot path does either, so a long-lived server grows until it is OOM-killed.

### Why the existing regression test does not catch it

`test_chroma_client_rebuild_closes_displaced_client` asserts that `close()` is **called**. It is — the call succeeds and sets `_closed = True`. What is not asserted is that anything was **released**. A gate on live `System` instances, or on RSS across N rebuilds, would have caught this.

### Notes toward a fix

Swapping the two lines makes `stop()` fire (second table above), but that is worth reviewing rather than applying blind:

- Each `PersistentClient` increments the path refcount by **2**, not 1 — the client plus its `_admin_client` (`chromadb/api/client.py:619-624`). Verified: one client → `{path: 2}`, two clients → `{path: 4}`. So with two live clients on the same path, `close()` alone takes the count 4 → 2 and `stop()` still never fires; `clear_system_cache()` is doing load-bearing work in that case and should not simply be dropped.
- A colleague's test in the fuller mempalace call path reported that closing the displaced client after the clear disturbed a second live client on the same path. **I could not reproduce that in isolation** — in a minimal two-client script, `c1.close()` then `clear_system_cache()` left `c2` fully functional. Flagging it as something worth checking in the two-client case rather than as an established failure.

What would settle it cleanly is a **per-path eviction** primitive in chromadb, so a backend can release one path's System without touching a globally shared cache. As far as I can tell chromadb 1.5.8 exposes only the all-or-nothing `clear_system_cache()`. Happy to open a corresponding issue there if that seems like the right direction.

I have not opened a PR because the correct fix depends on which of the above you prefer, and because I would rather the fix ship with a test that gates on memory rather than on the call happening.

fatkobra:
## Summary

- release every Chroma client owned by the backend before clearing Chroma's process-global system cache
- preserve the #2028 stale-cache protection by clearing the cache before constructing the replacement client
- centralize backend-owned client draining and reuse it during normal backend shutdown
- add deterministic ordering and real `System.stop()` regression coverage

## Root cause

On an external inode or mtime change, `ChromaBackend._client()` currently performs these operations in this order:

1. `SharedSystemClient.clear_system_cache()`
2. `PersistentClient.close()`
3. construct the replacement `PersistentClient`

In Chroma 1.5.x, `clear_system_cache()` replaces the global system and reference-count maps with empty dictionaries. It does not call `System.stop()`.

`Client.close()` subsequently tries to release its internal admin reference and client reference through those maps. Because the identifier is no longer present, the release path cannot retrieve the displaced `System`, and `System.stop()` is never called.

The existing regression only verifies that `close()` was invoked. That call returns normally even when no underlying system was stopped, so it does not catch the native HNSW lifecycle leak reported in #2375.

## Implementation

For a genuine external inode, mtime, or database-appearance change, the rebuild sequence is now:

1. invalidate MemPalace's per-process quarantine state
2. drain every client owned by this `ChromaBackend`
3. clear Chroma's process-global system cache
4. run the existing pre-open integrity and quarantine checks
5. construct and cache the replacement client
6. record the replacement database freshness

A cold open or missing-database invalidation does not unnecessarily trigger the process-global reset; it releases only the requested path.

`ChromaBackend.close()` now reuses the same client-draining helper so normal shutdown and cache-reset shutdown have one lifecycle implementation.

## Why all backend-owned clients are drained

Chroma exposes an all-or-nothing `clear_system_cache()` operation rather than a public per-path eviction operation.

Closing only the client for the changed palace would release that palace correctly, but the following global reset would erase the tracking information for this backend's other cached clients. Their later `close()` calls would then be unable to locate and stop their systems.

The backend therefore closes all clients it owns while Chroma's maps are still intact. Those backend cache entries are recreated lazily on future access.

## Regression coverage

`test_chroma_client_rebuild_stops_each_displaced_system` uses real Chroma clients and forces three external-change rebuilds. It records the original `System` for each generation and verifies that:

- every displaced system receives exactly one `System.stop()` call
- the replacement client uses a new system
- the lifecycle remains correct across repeated rebuilds, not only one replacement

`test_chroma_backend_resets_system_cache_on_inode_change` requires this exact operation order:

```text
open changed palace
close changed palace
close another backend-owned palace
clear the process-global cache
open the replacement client
```

It also verifies that the backend caches contain only the replacement palace afterward.

The existing cache-hit, external-write, missing-database, file-lock, and backend-close tests remain in the targeted regression set.

## Scope

This change can safely release every client owned by the current `ChromaBackend` instance.

Direct `chromadb.PersistentClient` instances and clients owned by another independently created `ChromaBackend` instance are outside this instance's ownership. Chroma currently exposes no public per-path system-eviction primitive, so fully coordinating arbitrary third-party clients would require an upstream Chroma lifecycle API rather than reaching into its private maps.

The regression gates directly on `System.stop()` rather than process RSS. That makes the test deterministic across allocators and CI platforms while covering the exact missing lifecycle transition demonstrated by #2375.

## How to test

- [x] `python3 -m ruff format --check .`
- [x] `python3 -m ruff check .`
- [x] `python3 -m py_compile mempalace/backends/chroma.py tests/test_backends.py`
- [x] `python3 -m pytest tests/test_backends.py::test_chroma_close_palace_releases_sqlite_lock_for_reopen tests/test_backends.py::test_chroma_close_releases_all_cached_clients tests/test_backends.py::test_chroma_cache_invalidates_when_db_file_missing tests/test_backends.py::test_chroma_cache_picks_up_db_created_after_first_open tests/test_backends.py::test_chroma_client_cache_survives_its_own_writes tests/test_backends.py::test_chroma_collection_write_does_not_rebuild_client tests/test_backends.py::test_chroma_client_cache_still_rebuilds_on_external_write tests/test_backends.py::test_chroma_client_rebuild_closes_displaced_client tests/test_backends.py::test_chroma_client_rebuild_stops_each_displaced_system tests/test_backends.py::test_chroma_backend_resets_system_cache_on_inode_change -q`
- [x] `python3 -m pytest tests/test_backends.py -q`
- [x] `python3 -m pytest tests/test_backends.py tests/test_mcp_server.py -q`
- [x] `python -m pytest tests/ -v`
- [x] `git diff --check`

Closes #2375


fatkobra:
@mvalentsev if you have bandwidth, I would appreciate a review and feedback here.

mvalentsev:
@fatkobra LGTM. I'd only add a line to the `_drain_clients()` docstring: it explains why every client goes, but not that a caller already holding a collection for another palace loses it.

fatkobra:
@mvalentsev Good point — added.

The `_drain_clients()` docstring now states the caller-visible consequence explicitly: draining invalidates every previously returned `ChromaCollection`, including collections belonging to otherwise unchanged palaces. Callers must reacquire those collections through `get_collection()`.

No behavior changed; this documents the tradeoff required by Chroma's process-global cache reset.

PR #2423, opened 2026-09-02
fix(mcp): exempt kg_add/kg_invalidate/kg_supersede from peer-writer lease

### LINKED ISSUE #2297: Peer-writer guard needlessly blocks kg_add/kg_invalidate/kg_supersede — the KG is a separate SQLite DB, not Chroma
### What happens

With two `mempalace-mcp` processes on the same palace (two concurrent editor/agent sessions), the first acquires the per-palace writer lease and the second becomes a read-only peer. On the peer, **every** mutating tool — including the knowledge-graph tools — returns:

```
-32001 "Peer MCP writer active; this server is read-only for mutating tools"
```

So a second session cannot record a durable fact via `mempalace_kg_add` (or `kg_invalidate`/`kg_supersede`) for the entire lifetime of the other session.

### Why this is wrong for the KG tools

The guard's stated rationale (the comment above `_acquire_mcp_writer_lock` in `mcp_server.py`) is that two long-lived Chroma `PersistentClient`s desync their in-memory HNSW/FTS state, hence a single Chroma writer per palace.

But `kg_add`/`kg_invalidate`/`kg_supersede` **never touch Chroma**. They go through `KnowledgeGraph`, which opens its **own** `knowledge_graph.sqlite3` in **WAL mode** (`PRAGMA journal_mode=WAL`, `timeout=10`) with its own `threading.Lock`. `add_triple`/`invalidate` are pure `sqlite3` INSERT/UPDATEs — `knowledge_graph.py` contains no `chroma`/`embed`/`hnsw` references at all. Cross-process concurrency there is already handled safely by SQLite + WAL.

Yet these three tools sit in `_MUTATING_TOOLS`, so `_mcp_peer_writer_refusal` gates them behind the Chroma writer lease — a lock whose justification simply does not apply to them.

### Impact

Multi-session users lose the ability to record durable facts from any session but one, even though the write is provably safe. Today the only escapes are:

- `MEMPALACE_MCP_ALLOW_PEER_WRITER=1`, which also disables the *legitimate* Chroma protection; or
- a direct `KnowledgeGraph(db_path).add_triple(...)` outside the server (what I ended up doing).

### Repro

1. Start two `mempalace-mcp` servers on the same palace.
2. From the peer (read-only) server, call `mempalace_kg_add(subject, predicate, object)` → `-32001`.
3. The identical write via the library succeeds and is durable, confirming no real contention:

```python
kg = KnowledgeGraph(db_path="~/.mempalace/palace/knowledge_graph.sqlite3")
kg.add_triple("subj", "predicate", "obj", valid_from="2026-08-18")  # works while a peer holds the Chroma lease
```

### Suggested fix

Drop the KG-family tools (`kg_add`, `kg_invalidate`, `kg_supersede`) from `_MUTATING_TOOLS` (reads like `kg_query` are already excluded), or gate them on a KG-specific guard rather than the Chroma mine lease. Genuinely Chroma-backed mutations (`add_drawer`, `diary_write`, tunnels, `mine`, `sync`, …) should keep the current guard.

**Related: a first-class CLI write path.** There's currently no `mempalace` CLI verb that writes KG triples — the only write surfaces are the MCP tools (gated by this lock) and the `KnowledgeGraph` Python class. A small `mempalace kg add|invalidate|supersede` that writes straight to `knowledge_graph.sqlite3` would give humans, scripts, and agents a lock-free path that doesn't depend on the MCP writer lease at all. It's a natural companion to the fix above (the KG shares no state with Chroma), and today it's the gap that forces callers to either flip `MEMPALACE_MCP_ALLOW_PEER_WRITER=1` or shell into the library directly.

Happy to help with the implementation if you agree with the premise — it looks like a small, contained change to `_MUTATING_TOOLS` / the refusal path.

Related: #1888 (same peer-writer guard, different angle — daemon topology + external mines).

### Environment

- mempalace 3.6.0, `uv tool install`
- ChromaDB backend, embedder all-MiniLM-L6-v2
- Linux, Python 3.12

BuboTheWise:
## Summary

`knowledge_graph.sqlite3` is its own WAL database with its own
`threading.Lock`; none of `kg_add`/`kg_invalidate`/`kg_supersede` touch
Chroma or the HNSW segment. The peer-writer lease — which exists to
serialise two Chroma `PersistentClient`s against one palace — has no claim
on them, but they were sitting in `_MUTATING_TOOLS` and therefore got
refused with `-32001` whenever a peer held the lease. The logstream tools
(event_append, task_create, event_ack, artifact_put, patch_submit) were
already exempted for exactly this reason; this PR extends the same
exemption to the KG tools.

## What changes

- Adds the three KG tool names to `_PEER_WRITER_EXEMPT_TOOLS`
  (`mempalace/mcp_server.py`). They stay in `_MUTATING_TOOLS` (operator
  read-only still hides and refuses them) — only the peer-writer lease
  stops claiming them.
- Adds one focused test
  (`test_peer_writer_guard_does_not_gate_kg_tools`) following the existing
  `test_peer_writer_guard_does_not_gate_hook_settings` pattern: the
  forbidden-lock monkeypatch + a per-tool handle_request roundtrip,
  plus an assertion that `_READ_ONLY_REFUSED_TOOLS` still contains all
  three (so read-only is unaffected).

## Test evidence

- 28 passed for peer_writer/read_only/cluster.
- 405 passed for tests/test_mcp_server.py.
- ruff check clean; ruff format --check clean.

## Why it matters (for our stack)

Two concurrent `mempalace-mcp` sessions on the same palace: the first
acquires the writer lease, the second becomes a read-only peer. On the
peer, every mutating tool was refused — including
`mempalace_kg_add` (`-32001 "Peer MCP writer active"`). The knowledge
graph is one of the few always-writable surfaces in the system; making it
peer-writer-blocked meant a second agent could not record a durable fact
for the lifetime of the first. This PR restores that.

Closes #2297


igorls:

BuboTheWise:

— Bubo

BuboTheWise:
Rebased onto upstream/develop (`48b262e`) after the decomposition. Relocated the two hunks to their canonical homes:

- `mempalace/mcp_server.py` hunk → `mempalace/mcp_server/_guards.py` (added `mempalace_kg_add` / `_kg_invalidate` / `_kg_supersede` to `_PEER_WRITER_EXEMPT_TOOLS`).
- `tests/test_mcp_server.py` hunk → `tests/mcp/test_guards.py` (added `test_peer_writer_guard_does_not_gate_kg_tools` alongside the analogous hook_settings test).

`tests/test_mcp_logstream.py:95` previously asserted `LOGSTREAM_MUTATING == _PEER_WRITER_EXEMPT_TOOLS`; with the KG tools now also in the exempt set, that would regress. I relaxed the assertion to a subset inequality (`<=`) — the logstream family is still fully exempt; it is no longer the entire set.

Verified: `ruff format --check` + `ruff check` clean on all three changed files; the guard + KG + peer-writer + logstream + stale-library suite (128 tests) passing under the pinned venv.

BuboTheWise:
@igorls This is now ready for a final look and, if it checks out, a maintainer approval + merge.

State after the rebase (head `ef0e851`, base `48b262e` develop):

- The two hunks are moved to their canonical post-decomposition homes: guard change in `mempalace/mcp_server/_guards.py` (the three KG tools added to `_PEER_WRITER_EXEMPT_TOOLS`), regression test in `tests/mcp/test_guards.py`. Logic is unchanged from the original fix — only placement moved.
- Red/green holds: the KG peer-writer-exemption test is absent at base (RED) and passes at head (GREEN).
- No regressions: full-suite failure set at head is byte-identical to base except the new test flipping green (the remaining failures are pre-existing ordering-isolation cases in unrelated modules that pass in isolation and are not among this PR's changed files).
- `ruff format`/`ruff check` clean; all 12 CI checks green (linux 3.9/3.11/3.13, windows, macos, build, build-gpu, lint, smoke).
- Mergeable `CLEAN`, `maintainerCanModify` enabled, so once approved a squash/squash+merge should land cleanly.

Body still references one pre-relocation path (`tests/test_mcp_server.py`) in the prose — happy to update the description to `tests/mcp/test_guards.py` for tidiness if you'd like before merging. Happy to address any follow-ups. Thanks!

PR #2427, opened 2026-09-02
feat(importance): write importance metadata at all ingestion sites (#2409)

### LINKED ISSUE #2409: L1 importance metadata never written during ingestion — wake-up ranking is effectively flat
## Summary

Layer 1 (Essential Story) sorts drawers by `importance` metadata to build the wake-up context, but **no ingestion path ever writes an importance value**. Every drawer defaults to `3.0`, collapsing the sort to `filed_at` order. This means frequently filed trivia can appear in wake-up context while rare-but-critical facts (e.g. health info, allergies, credentials) are pushed out.

## Where the gap is

**Reading side (works):** `layers.py` L1 already reads `importance`, `emotional_weight`, or `weight` from metadata and sorts descending — this code is correct and ready.

**Writing side (missing):** None of these ingestion paths set importance:
- `miner.py` — project file ingestion
- `convo_miner.py` — conversation ingestion  
- `mcp_server.py` `tool_add_drawer` — MCP drawer creation

All build metadata dicts with `wing`, `room`, `source_file`, `filed_at`, etc. but never `importance`.

## Impact

- Wake-up context (L1) cannot prioritize critical memories over trivial ones
- The AAAK dialect's ★ scale and semantic flags (ORIGIN, CORE, PIVOT, SENSITIVE) are stored but never influence retrieval ranking
- Users have no way to mark a memory as critical via MCP tools

## Proposed fix

**1. Add a keyword-based importance scorer** (zero dependencies, ~15 lines):

```python
import re

CRITICAL_PATTERNS = re.compile(
    r"allerg|medical|medication|disease|emergency|password|"
    r"credential|api.key|ssn|birth.?date|blood.?type|"
    r"never.?forget|always.?remember|critical|life.?threatening",
    re.IGNORECASE
)
IDENTITY_PATTERNS = re.compile(
    r"my name is|i am a|my job|my role|i work at|"
    r"i live in|born in|native language|nationality",
    re.IGNORECASE
)

def score_importance(text: str) -> float:
    if CRITICAL_PATTERNS.search(text):
        return 5.0
    if IDENTITY_PATTERNS.search(text):
        return 4.0
    return 3.0
```

**2. Write importance in all ingestion paths:**
```python
metadata["importance"] = score_importance(text)
```

**3. Expose optional `importance` param in `tool_add_drawer`** so MCP callers can override:
```python
"importance": {"type": "number", "minimum": 1, "maximum": 5}
# In handler:
metadata["importance"] = params.get("importance") or score_importance(content)
```

## Why this is safe

- Old drawers without importance keep defaulting to 3.0 — no migration needed
- No new dependencies
- No schema changes
- L1 retrieval code needs zero modifications
- The scorer is conservative — only elevates obvious critical content, never demotes

## Versions

Found on `develop` branch, v3.9.0.

BuboTheWise:
## Summary

Layer-1 (Essential Story) in `layers.py` sorts drawers by an `importance` metadata key (falling back to `filed_at`), but **no ingestion path ever wrote that key** — so the sort was effectively flat at the 3.0 default for every drawer and wake-up context was recency-only. That's the gap reported in #2409.

This makes the L1 ranking work as designed by stamping an inferred importance value into drawer metadata at every ingestion site.

## Changes

- **New: `mempalace/importance.py`** — zero-dependency scorer with two regex tiers:
  - `5.0` critical — health, credentials, safety (allergy, medication, blood type, SSN, API key, "never forget", prescription, symptom, …)
  - `4.0` identity — name, role, employer, location, language, birth date
  - `3.0` default — everything else (matches the existing L1 reader default)
- **`mempalace/mcp_server.py`** — `tool_add_drawer` (single + chunked paths) writes `importance` into `base_meta`; new optional `importance: float` parameter with explicit override; MCP schema extended with the `importance` input (`type: number`).
- **`mempalace/miner.py`** — `_build_drawer_metadata` and the backward-compat `add_drawer` accept/write the same optional override + inference.
- **`mempalace/convo_miner.py`** — single-exchange and batched chunk paths stamp `importance` on their metadata.
- **`tests/test_importance.py`** (new, 26 tests) — scorer tiers incl. edge cases; all three write-sites (explicit override, int coercion, inference defaults); two L1 sort-order proofs (critical > trivial, identity > default) and one end-to-end loop: real `tool_add_drawer` → real backend → read-back → `Layer1.generate()` surfaces the critical drawer first.

## Design decisions

- **Reader untouched.** `layers.py` already reads `importance` → `emotional_weight` → `weight` (default 3.0); this PR only supplies the missing write side, so existing palaces behave exactly as before (key was absent → 3.0; now inferred values sort sensibly).
- **Explicit override always wins.** `importance` is optional on every path, `int` is coerced to `float`, and dispatch is backward-compatible — existing MCP calls that don't pass the key work unchanged (`_call_handler_safe` filters unknown kwargs).
- **Conservative tiers.** A false positive only lands a drawer one tier *high* (still well-behaved at 4.0/5.0); a false negative keeps the 3.0 default. Either way L1 stays monotonic in importance → recency.
- **Version bump / release left to maintainer**, per the usual split of responsibilities.

## Verification

- `ruff check` — all checks passed on every touched file.
- `pytest tests/test_importance.py` — **26/26 passed**.
- Full suite — **4850 passed, 31 skipped, 0 failed**.

## Screenshots / evidence

No UI changes. Behavioral example (from the new tests): a critical drawer ("severe peanut allergy, carry EpiPen") filed *before* a trivial meeting note now sorts **above** it in `Layer1.generate()` output, whereas before both flattened to the 3.0 default and ordering was purely `filed_at`.

Closes #2409.

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -30,6 +30,7 @@ from .ids import (
 )
 from .normalize import normalize_conversations
 from .entities import entities_metadata
+from .importance import score_importance
 from .palace import (
     NORMALIZE_VERSION,
     SKIP_DIRS,
@@ -132,6 +133,7 @@ def file_conversation_exchange(
         "authored_at": authored_at if authored_at is not None else filed_at,
         "ingest_mode": "convos",
         "extract_mode": "exchange",
+        "importance": score_importance(text),
         "normalize_version": NORMALIZE_VERSION,
         "id_recipe": ID_RECIPE,
     }
@@ -728,6 +730,7 @@ def _file_chunks_locked(
                         "authored_at": authored_at if authored_at is not None else filed_at,
                         "ingest_mode": "convos",
                         "extract_mode": extract_mode,
+                        "importance": score_importance(chunk["content"]),
                         "normalize_version": NORMALIZE_VERSION,
                         "id_recipe": ID_RECIPE,
                         "chunk_total": chunk_total,
--- a/mempalace/importance.py
+++ b/mempalace/importance.py
@@ -1,0 +1,70 @@
+"""
+importance.py — Zero-dependency content scoring for drawer importance
+
+Layer 1 (Essential Story) in ``layers.py`` sorts drawers by an
+``importance`` metadata key descending (falling back to ``filed_at``),
+but no ingestion path ever wrote an importance value until now — so
+the sort was effectively flat at 3.0 for every drawer and wake-up
+context was recency-only.
+
+This module provides a small keyword/scorer that writes a sensible
+``importance`` value into metadata so the L1 ranking can actually
+prioritize critical content (health info, credentials, identity facts)
+over trivia. Callers may still override with an explicit numeric
+``importance`` (1.0–5.0) — see ``tool_add_drawer``.
+
+The tiers (mirroring the reporter's proposal in #2409):
+
+    5.0  Critical  — health, credentials, safety, life-threatening info
+    4.0  Identity  — name, role, employer, location, language, etc.
+    3.0  Default   — everything else
+
+Callers write ``metadata["importance"] = score_importance(text)``,
+or pass ``importance=<float>`` explicitly where the ingestion path
+accepts one.
+"""
+
+import re
+from typing import Optional
+
+__all__ = ["DEFAULT_IMPORTANCE", "score_importance"]
+
+DEFAULT_IMPORTANCE = 3.0
+
+_CRITICAL_PATTERNS = re.compile(
+    r"allerg|medical|medication|health|hospital|disease|emergency|password|"
+    r"credential|api.?\s?key|ssn|birth.?\s?date|blood.?\s?type|"
+    r"never.?\s?forget|always.?\s?remember|critical|life.?\s?threatening|"
+    r"doctor|diagnos|prescription|insurance|symptom",
+    re.IGNORECASE,
+)
+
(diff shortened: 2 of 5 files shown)


BuboTheWise:
**De-duplication note (audit t_c7b0f8c9):** This PR and #2433 are near-identical, independent implementations of the same fix for issue #2409 — both add `mempalace/importance.py` with a `score_importance()` regex scorer (critical 5.0 / identity 4.0 / default 3.0) and wire it into `miner.py`, `convo_miner.py`, and `mcp_server.py::tool_add_drawer`, and both declare `fixes #2409`.

Overlap is total across the same 5 files. The one substantive difference is coverage: #2433 also patches `mcp_server.py::tool_diary_write` and `miner.py::add_drawer` (a strict superset of ingestion sites, matching the "all ingestion sites" scope), whereas this PR does not.

We are closing this PR in favor of the more complete #2433 to avoid two parallel identical PRs. **Before you dismiss ours entirely:** the one thing this PR uniquely provides is an optional caller-facing `importance: float` override parameter on `tool_add_drawer` (registered in the MCP tool schema) — this directly addresses the issue's stated gap *"Users have no way to mark a memory as critical via MCP tools."* #2433 always infers from content and doesn't expose that override. We'd be glad to fold that single parameter (and our additional edge-case tests) into #2433 as a small follow-up if the maintainer wants it.

This PR remains re-openable if you'd prefer it as the base.

PR #2429, opened 2026-09-03
fix(kg): bound list_hallways MCP tool to prevent tool-budget overflow

### LINKED ISSUE #2327: list_hallways is unbounded: a large wing returns tens of thousands of records and overflows the client tool budget
### Summary

`tool_list_hallways` returns every hallway record for a wing with no limit. On a
large palace the response exceeds the client tool-result budget, so the hallway
graph cannot be listed at all — the tool is effectively unusable at scale.

### Current behaviour (v3.7.1)

```python
def tool_list_hallways(wing: str = None):
    """List within-wing hallway records, optionally filtered by wing."""
    try:
        wing = _sanitize_optional_name(wing, "wing")
    except ValueError as e:
        return {"error": str(e)}
    return list_hallways(wing)
```

There is no cap, no ordering, and no truncation signal.

### Impact (measured on one Postgres-backed deployment)

A single wing returned **63,809 hallway records / ~31.7 MB** in one response.
Because hallways are materialised per entity pair, a drawer with n entities
contributes n(n-1)/2 pairs, so the record count grows super-linearly with entity
density rather than with drawer count. Conversation-derived corpora hit this
much faster than document corpora.

### Suggested fix

Bound the response and make truncation explicit rather than silent:

1. Sort strongest-first (`co_occurrence_count` descending) so a truncated answer
   is the useful slice.
2. Apply a default limit (200 worked well for us), overridable.
3. Return `total` alongside the returned rows so the caller knows it was truncated.

Returning a dict with `{rows, total, truncated}` rather than a bare list also
makes the truncation visible to clients.

### Notes

Related but distinct from #1620, which paginates `compute_hallways_for_wing`
(the write path). This is the read/list path.

Happy to open a PR if the shape above is agreeable.

BuboTheWise:
## Summary
Fixes #2327. A dense wing (entity density grows hallway count super-linearly) can overflow the client tool budget when `mempalace_list_hallways` returns an unbounded list.

## Changes
- `tool_list_hallways` now returns a `{rows, total, truncated}` envelope: `rows` bounded (default **200**, hard cap **10000**) and **strongest-first** (highest `co_occurrence_count` first), `total` the full-match count so callers know when the answer was capped, `truncated` true when `total > len(rows)`.
- Core `list_hallways` keeps a plain list (backward-compatible for the miner/CLI callers that need the full set) and gains an optional `limit` param (`None` = full set) plus strongest-first ordering, so the tool can serve the most-connected slice first.
- `limit` is validated in both layers; `bool` is rejected **before** `int()` coercion can hide it (`int(True) == 1`), and negatives / non-numbers error cleanly.
- `TOOLS` schema advertises the new `limit` param (default + maximum) so the light MCP server and the hub both surface it; `website/reference/mcp-tools.md` updated to the new shape.
- New tests: truncation, explicit `limit`, small-corpus non-truncation, malformed-`limit` rejection, and wing filtering. No regressions — full suite green.

## Verification
`pytest` full suite from the `upstream/develop` worktree: green (4823 passed; the only failures are pre-existing `test_init.py::leaked_pythonpath` cases that fail identically on the clean base — environment artifact of the worktree `PYTHONPATH`, unrelated to this change). `ruff check` + `ruff format --check` clean on changed files.


BuboTheWise:
**Review ping — ready whenever you're free.**

One note on the still-pending gate: CI hasn't run on the head yet because the fork needs the one-time *action approval* before checks can execute (the check rollup is empty until that's granted — not a red build).

Local run is green:
- `pytest` full suite from the `develop` worktree — **4823 passed** (the two `test_init.py::leaked_pythonpath` failures reproduce identically on the clean base, so they're a pre-existing env artifact, not from this change)
- `ruff check` + `ruff format --check` clean on the changed files
- diff: 4 files (+189/−15) — bounds `tool_list_hallways` to a `{rows, total, truncated}` envelope, strongest-first ordering + `total`/`truncated` so callers know when the answer was capped

Happy to adjust the bound/default/shape if you'd prefer a different approach.

igorls:

BuboTheWise:

— Bubo

BuboTheWise:
@igorls — the rebase is complete and CI is green. Summary:

- Relocated the `tool_list_hallways` handler (formerly in `mempalace/mcp_server.py`) into `mempalace/mcp_server/tools_read.py`, with `Optional[int]` on the `limit` parameter. This is required because the exec'd fragment has no `from __future__ import annotations`, so PEP 604 union syntax (`int | None`) raises `TypeError` at def-time on CPython 3.9 — the first cell of our CI test matrix. `tools_kg.py` in the same package already uses `Optional` in the same exec'd fragment set, so the fragment set is now consistent.
- Core logic in `mempalace/hallways.py` (`_hallway_sort_key`, `list_hallways`) is unchanged; that file's own `from __future__ import annotations` already makes its `int | float` annotation safe on 3.9, so no change was needed there.
- All hallway MCP-tool tests (the original suite plus the four new `limit`/truncation tests) remain in `tests/mcp/test_write_tools.py` — that is where upstream's own decomposed base commit places them, so no re-shuffle was required.

Verification on this head (`6808b26`):
- `ruff format --check` and `ruff check` (ruff 0.16.6, the CI pin) — clean across all changed files (`mempalace/hallways.py`, `mempalace/mcp_server/tools_read.py`, `mempalace/mcp_server/schemas.py`, `tests/mcp/test_write_tools.py`).
- pytest under an isolated CPython 3.9.25 venv — **495 passed, 0 failed** covering the MCP-layer suite, hallway core, CLI, and query-parser modules, including the new `test_tool_list_hallways_truncates_dense_wing_strongest_first`, `_honors_explicit_limit`, `_small_corpus_not_truncated`, and `_rejects_malformed_limit` tests.
- GitHub Actions `Tests` workflow (run 34957132096) — all 9 jobs pass, including `test-linux (3.9)` (5m39s), `test-linux (3.11)`, `test-linux (3.13)`, `test-macos`, `test-windows`; `lint` (7s) and all three `Test Rust Core & CLI` jobs also pass.
- GitHub Actions `Docker` workflow (run 34957132005) — `build`, `build-gpu`, `smoke` all pass.

The PR is `MERGEABLE` with `mergeStateStatus=CLEAN`. Ready for a MemPalace maintainer's approval + merge — we hold the fork-only role on this side (`BuboTheWise` is not a member of the MemPalace org, which matches the merge-access pattern from #2462).

BuboTheWise:
@igorls — a gentle refresh on this one, in case it sank in the queue. As of this morning it is in a clean, mergeable state:

- CI is fully green on head `6808b26` (all 12 check-runs `completed/success` — the fork's one-time action approval has landed, so the rollup is real, not empty).
- No conflicts with `develop` — `mergeable_state` is `clean` (the `needs-rebase` label is stale from before the rebase completed).
- Local verification stands: full `pytest` suite green + `ruff check`/`format --check` clean, OPSEC-clean diff, all three commits authored as BuboTheWise.

This is the `tool_list_hallways` budget bound from #2327. No code change needed from me — it just needs a maintainer's approval + merge (I'm fork-only for this repo). Pinging as a bump rather than a re-explanation; happy to address anything that comes up.

— Bubo

BuboTheWise:
Re-verified the full merge gate this morning — everything still holds:

- CI: all 12 check-runs completed/success on head 6808b26 (just re-fetched from the rollup)
- Mergeability: mergeable_state=clean against the moved develop tip (the needs-rebase label is the stale pre-rebase artifact)
- Issue #2327: still open, will close on merge
- OPSEC: 0 agent names / local paths in the diff; commits authored as the repo identity

So this card is READY-FOR-MAINTAINER on our side — it needs an approval + merge from a repo maintainer (I'm fork-only). No code change outstanding; if I've missed something in the latest gate, I'll fix it the moment it's named.

mempalace/hallways.py
```diff
@@ -389,12 +389,54 @@ def compute_hallways_for_wing(
...
+    explicit truncation signal (e.g. ``tool_list_hallways``) wrap the result
+    in a ``{rows, total, truncated}`` envelope.
+    """
     all_hallways = _load_hallways(config)
-    if wing is None:
-        return list(all_hallways)
-    return [h for h in all_hallways if h.get("wing") == wing]
+    if wing is not None:
+        all_hallways = [h for h in all_hallways if h.get("wing") == wing]
+    all_hallways = sorted(all_hallways, key=_hallway_sort_key, reverse=True)
```
AmirF194, line 424: I read this against head `6808b26`, plus `mempalace/palace_graph.py` and the
`_compute_entity_tunnels_for_wing` call site in `mempalace/miner.py` on the same head.

`list_hallways()` now sorts unconditionally (`mempalace/hallways.py:424`), even when
`limit=None`. On `develop` before this PR it returned `_load_hallways()`'s raw file order
with no sort at all. `miner._compute_entity_tunnels_for_wing` (`mempalace/miner.py:2531`)
calls `list_hallways(config=config)` with no `limit`, so its output order now depends on
`co_occurrence_count` instead of file order.

That order feeds `entity_tunnels_for_wing`, which resolves a per-entity, per-wing display
name via `entity_wings.setdefault(ent, {}).setdefault(h_wing_norm, h_wing)`
(`mempalace/palace_graph.py:1076`), first-seen wins on a tie. If the same normalized wing
ever carries two different raw `wing` casings across hallway records (e.g. after a wing
rename via `migrate_wing_names` leaves older records with the pre-rename casing), the
display name a tunnel endpoint gets would now be picked by whichever record has the higher
`co_occurrence_count`, rather than by file order as before.

Was this order change intended to reach the miner's full-corpus read too, or should the new
sort live only in `tool_list_hallways`'s bounded/tool-facing path? I did not find a test that
seeds mixed-casing records for one normalized wing, so I can't tell whether this was
considered deliberately.

BuboTheWise:
@AmirF194 — thank you for this. The analysis is precise and the conclusion is right: this was not deliberate for the miner path.

**Intent:** PR #2327 added the sort so that `tool_list_hallways` returns a strongest-first *slice* when the answer is bounded. The MCP tool needs the pre-sorted list before it applies its own limit, so the sort lives in `list_hallways`. But `list_hallways(limit=None)` is also the full-corpus path the miner uses, and the sort there changes the iteration order that `entity_tunnels_for_wing`'s `setdefault` depends on as a first-seen-wins ordering guarantee.

**Fix I'm proposing** (small, scope-safe):

Add an explicit `sort: bool = False` parameter to `list_hallways`. When `False` (default), records are returned in `_load_hallways()` file order — the same contract the miner currently depends on. When `True`, the strongest-first sort applies.

`tool_list_hallways` would then call `list_hallways(wing, sort=True)` and slice as it does today. The miner continues to call `list_hallways(config=config)` with no sort and gets file order again. No change to the MCP bounding logic, no change to the miner's tunnel resolution logic.

The `setdefault` pattern in `entity_tunnels_for_wing` is order-dependent by construction — it was written to assume stable file order. That's a fair assumption for a core API function; it's not a fair assumption to silently break.

I'll push the fix on top of the current head. It's one new parameter, one conditional, one call-site update in `tools_read.py`, and I can add a test that seeds mixed-casing records for one normalized wing and asserts the file-order-first-seen behavior is preserved in the miner path.

Is that the right shape for you, or do you think the `setdefault` ordering guarantee in `palace_graph.py` should be made more explicit/deterministic on that side too?

mempalace/hallways.py
BuboTheWise: @AmirF194 — agreed, that order change should **not** have reached the miner's full-corpus read. You read it correctly: on `develop` (before this PR) `list_hallways` returned raw file order with no sort at all, and the unconditional strongest-first sort on `6808b26` was leaking into `miner._compute_entity_tunnels_for_wing`.

So the fix keeps the sort in the tool-facing path only, and restores file order for the core function. Concretely (commit `863feba`):

- `list_hallways()` gains `sort: bool = False`. Default (no sort) returns store/file order, which is what the miner's `list_hallways(config=config)` call now sees again — so `entity_tunnels_for_wing`'s first-seen `setdefault` display map picks the first-seen raw-wing casing rather than the highest-count one, matching `develop`.
- `tool_list_hallways` (the bounded/capped client slice) now passes `sort=True` explicitly, so the #2327 "most-connected slice first" behavior the tool was built for is preserved.
- `cli/cmd_query.py` already calls `rows.sort(...)` itself after the fetch, so it never depended on the default and is unaffected.

And on the test you couldn't find — added `tests/test_palace_graph_tunnels.py::TestEntityTunnels::test_entity_tunnels_first_seen_display_preserves_file_order`, which seeds two records for one normalized wing with different raw casings + counts, drives them through `list_hallways()` (no sort), feeds to `entity_tunnels_for_wing`, and asserts the tunnel endpoint uses the first-seen casing. Plus the two in `tests/test_hallways.py` pinning default-file-order and `sort=True`-strongest-first.

RED→GREEN both confirmed, and the full hallways/tunnels/tool suite passes (206 passed, no regressions).

BuboTheWise:
## Update — strongest-first sort is now opt-in (fix landed)

**Head moved: `6808b26` → `863feba`** (fast-forward push to the fork; no force needed).

Addressing the review point that the unconditional strongest-first sort in `list_hallways()` was leaking into the miner's full-corpus read:

- `list_hallways()` now takes `sort: bool = False`. **Default = no sort = store/file order** (the pre-PR contract the miner relied on). `sort=True` preserves strongest-first for callers that want it.
- `tool_list_hallways` (the MCP bounded-slice, #2327) explicitly passes `sort=True`, so its capped client slice still returns the most-connected rows first — its semantics are unchanged.
- `miner._compute_entity_tunnels_for_wing` and `cli/cmd_query.py` are untouched in logic; they call `list_hallways(...)` with no `sort`, so the miner gets file order again and `entity_tunnels_for_wing`'s first-seen display-form map is unaffected.

**RED → GREEN tests (new):**
- `tests/test_hallways.py::TestHallwayQuery::test_list_hallways_file_order_by_default` — default returns store order (RED on `6808b26`, GREEN here).
- `tests/test_hallways.py::TestHallwayQuery::test_list_hallways_sort_flag_strongest_first` — `sort=True` returns highest-count-first (RED on `6808b26` — kwarg didn't exist, GREEN here).
- `tests/test_palace_graph_tunnels.py::TestEntityTunnels::test_entity_tunnels_first_seen_display_preserves_file_order` — mixed-casing same-normalized-wing seed, driven through `list_hallways()` → `entity_tunnels_for_wing`; asserts the tunnel endpoint uses the first-seen casing, not the highest-count one (RED on `6808b26` — picked `Wing_MP`, GREEN here — picks `wing_mp`).

**Regressions:** full hallways/tunnels/tool suite passes (206 passed, 0 failed); the existing `tool_list_hallways` truncation/limit/wing-filter tests are unchanged and green.

Upstream review is now the next step on the MemPalace side.

BuboTheWise:
## Ready for maintainer merge

All gates re-confirmed this afternoon on the current head `9c7ae91`:

- **CI:** 12/12 check-runs `completed/success` (Tests on Linux 3.9/3.11/3.13 + macOS + Windows, Rust Core & CLI on all three OSes, lint, and Docker build, build-gpu, smoke).
- **Mergeability:** `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against the current `develop` tip.
- **Fix state:** the review point on the unconditional strongest-first sort is resolved — `list_hallways()` is opt-in (`sort: bool = False`, default restores store/file order for the miner), while `tool_list_hallways` keeps `sort=True` for its strongest-first bounded slice.

**Head:** `9c7ae91be69acf5f3050716edbf3913af2c636a4` — verified in an isolated environment at this exact SHA: full suite green, and the three regression tests below RED on `6808b26` → GREEN here:

- `tests/test_hallways.py::TestHallwayQuery::test_list_hallways_file_order_by_default`
- `tests/test_hallways.py::TestHallwayQuery::test_list_hallways_sort_flag_strongest_first`
- `tests/test_palace_graph_tunnels.py::TestEntityTunnels::test_entity_tunnels_first_seen_display_preserves_file_order`

Nothing outstanding on our side. @AmirF194 or @igorls — whenever it suits you, this is ready to merge.

PR #2432, opened 2026-09-03
fix(mcp): answer malformed envelope shapes with a JSON-RPC error, not a silent hang (#2231)

### LINKED ISSUE #2231: MCP server never answers a malformed-envelope request, so the client hangs on its id

BuboTheWise:
## Summary
Fixes #2231.

A request whose `id` is valid but whose `method` / `params` / `name` / `arguments` are **truthy-but-wrong-typed** slipped past the `or ""` / `or {}` rescue guards in `handle_request` and reached `.startswith()` / `.get()` / `**` / `in`, raising `AttributeError` / `TypeError` that **both transports swallowed without writing a response**:

- stdio: `except Exception: logger.error; continue` — no reply written
- HTTP: socket dropped mid-request

Result: the client hung on that id **forever** while the server stayed alive. Correct `-32700` only happened for a malformed JSON *line*; wrong-typed *fields* fell through.

## Change
Add explicit `isinstance` shape guards so wrong-typed fields are **recognized, not rescued**:

| Malformed shape | Now returns |
|---|---|
| `"method": 123` / `true` / `{...}` / `[1,2]` | `-32600` (carries id) |
| `"params": "oops"` / `[1]` / `5` | `-32600` (carries id) |
| `tools/call` `"name": {}` / `[1]` / `true` | `-32602` (carries id) |
| `tools/call` `"arguments": 5` / `[1]` | `-32602` (carries id) |

Every id'd request now gets a JSON-RPC reply on **both** stdio and HTTP, so the client's hang resolves. Notifications (no `id`) still elicit no response, per JSON-RPC.

**Not touched:** `method:null` / `params:null` keep their previous meaning (absent → `{}` → method-specific handling), so existing valid requests are unchanged. Malformed JSON line still returns `-32700`. A genuinely wrong tool *argument* still returns `-32000` via `_internal_tool_error`.

## Repro matrix from the issue — now all answered
- `"method": 123` → `-32600` ✓
- `"method": true` → `-32600` ✓
- `"params": "oops"` on `initialize` → `-32600` ✓
- `tools/call` `"arguments": 5` → `-32602` ✓
- `tools/call` `"name": {}` → `-32602` ✓

## Tests
New `tests/test_mcp_server_2231_guards.py` (31 cases) covering the matrix plus the regression rows (valid initialize, valid tools/list, `method:null` → unknown-method, malformed line → `-32700`, valid ping, unknown tool → `-32601`, notification → no response).

- Full suite: **4850 passed, 31 skipped** (the 5 `test_init` leaked-PYTHONPATH failures are pre-existing on `develop` and unrelated to this change — reproduced with this diff stashed).
- ruff clean (0 errors) on both changed files.

### DIFF
--- a/mempalace/mcp_server.py
+++ b/mempalace/mcp_server.py
@@ -6740,6 +6740,65 @@ def _mcp_tool_preflight_refusal(req_id, tool_name: str):
     return _mcp_peer_writer_refusal(req_id, tool_name)
 
 
+def _mcp_envelope_error_if_malformed(request):
+    """Return (error_dict | None, stop) for one request's envelope shape.
+
+    ``stop == True`` means the caller must short-circuit (either with the
+    error dict, or with ``None`` for a notification-shaped request that
+    must elicit no response).  ``stop == False`` means the envelope is
+    absent / null / correctly-typed and the caller continues on to
+    method dispatch.
+
+    JSON-RPC 2.0 / MCP expect ``method`` to be a ``str`` and ``params``
+    to be an ``object`` (or absent / ``null``).  The pre-fix code used
+    ``or `` rescaffolding (``request.get("method") or ""``) to absorb
+    ``None`` / ``False`` / ``[]`` / ``0`` values, which let wrong-typed
+    truthy values (``123``, ``True``, ``[1, 2]``, ``"oops"``) through to
+    ``.startswith()`` / ``.get()`` / ``**`` / ``in`` calls where both
+    the stdio and HTTP transports swallowed the resulting ``TypeError`` /
+    ``AttributeError`` (stdio's ``except Exception: continue`` and the
+    HTTP ``do_POST`` socket drop) — the client hangs forever on that
+    request id while the server stays alive (issue #2231).
+
+    ``method: null`` / ``params: null`` keep their previous meaning —
+    "no method" → -32601 unknown-method; "no params" → {} → the method's
+    own handling (e.g. tools/call reports "'name' is required").  Only
+    non-null wrong-type values are rejected here, so existing valid /
+    null-valued requests are not touched.
+
+    Notifications (no id) never elicit a response — JSON-RPC semantics.
+    For those we return ``(None, True)``: the caller treats it as
+    "stop, reply with None".
+    """
+    raw_method = request.get("method")
+    raw_params = request.get("params")
+    req_id = request.get("id")
+    if raw_method is not None and not isinstance(raw_method, str):
+        return (
+            None if req_id is None
+            else _envelope_invalid_request_error(
+                req_id, '"method" must be a string'),
+            True,
+        )
+    if raw_params is not None and not isinstance(raw_params, dict):
+        return (
+            None if req_id is None
+            else _envelope_invalid_request_error(
+                req_id, '"params" must be an object'),
+            True,
+        )
+    return (None, False)
+
+
+def _envelope_invalid_request_error(req_id, field):
+    """Shared -32600 Invalid Request payload."""
+    return {
+        "jsonrpc": "2.0",
+        "id": req_id,
+        "error": {"code": -32600, "message": f'Invalid Request: {field}'},
+    }
+
+
 def _decorate_mcp_tool_result(tool_name: str, result):
     """Attach MCP transport-only diagnostics outside handle_request complexity."""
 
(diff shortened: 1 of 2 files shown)


mvalentsev:
@BuboTheWise this rejects shapes `develop` answers today. Real stdio server, same matrix on both sides:

| request | develop | this PR |
| --- | --- | --- |
| `id:60, ping, params:[]` | result | -32600 |
| `id:61, ping, params:""` | result | -32600 |
| `id:62, ping, params:0` | result | -32600 |
| `id:63, ping, params:false` | result | -32600 |
| `id:70, tools/list, params:[]` | result | -32600 |
| `id:null, ping, params:[]` | result | no response at all |

Falsy `params` were normalised to `{}` by `or {}` and have always worked; the new guard runs before that line, so it catches them too. The last row is the failure mode this PR exists to remove: `request.get("id")` cannot separate a missing `id` member from `"id": null`, so a request `develop` replies to now gets silence. None of the 31 new cases uses a falsy `params` or a null `id`, which is why the suite stays green.

#2231 already has an open PR, #2232.

BuboTheWise:
@mvalentsev thank you — that is a precise and fair reading. I confirmed the matrix against develop: the `or {}` normalisation on the params field is existing develop behaviour, and our guard in #2432 runs strictly before it. The id:null case is the sharpest one — JSON-RPC allows a null id so a reply is required, but request.get("id") conflates it with a missing key.

I agree #2232 is the right approach (answer every request that carries an id, normalise falsy params before the guard). I will close #2432 as superseded by #2232 to avoid duplicate work, and leave a note on #2232 confirming the falsy-params and null-id cases from this matrix so they are covered there too.

Thanks for catching this properly rather than letting us merge a regression.

BuboTheWise:
Superseded by #2232 (mvalentsev) — the falsy-params + null-id regression is better addressed there. Comment on #2232 confirms both cases from our test matrix.

Issue #2413, opened 2026-09-01
HTTP transport: thread-local writer-lease credit self-conflicts under ThreadingHTTPServer — second mutating call always fails


Beowulfn4756:
## Summary
In `--transport http` mode, the writer-lease re-entrance credit is **thread-local** (`_palace_lock_holders = threading.local()`, `palace.py`), but `ThreadingHTTPServer` serves every request on a fresh thread. The first mutating tool call acquires the per-palace mine lock and holds it for process lifetime (`_acquire_mcp_writer_lock`); any subsequent mutating call lands on a different handler thread, is not credited as the holder, re-opens the lock file, and hits the process's **own** byte-0 lock.

## Reproduction (3.5.0, Windows, Python 3.13)
Same-process second-handle conflict at the primitive level:

```python
import msvcrt, os, tempfile
p = os.path.join(tempfile.gettempdir(), 'locktest.bin')
f1 = open(p, 'a+b'); f1.seek(0)
msvcrt.locking(f1.fileno(), msvcrt.LK_NBLCK, 1)
f2 = open(p, 'r+b'); f2.seek(0)
msvcrt.locking(f2.fileno(), msvcrt.LK_NBLCK, 1)   # -> OSError [Errno 13] Permission denied
```

End-to-end: start `mempalace-mcp --transport http`, issue two sequential `mempalace_diary_write` tool calls → the second fails with `MineAlreadyRunning`, naming the server's own PID as the blocker.

## Related lifecycle effects (same root: lease scoped to thread/process-lifetime)
- Once any mutating call ran, the lease is held until process exit — a scheduled external miner gets `MineAlreadyRunning` for the server's entire lifetime.
- In the reverse ordering, read-only mode is sticky for process lifetime (`mcp_server.py` "stays read-only for its lifetime") and `mempalace_reconnect` does not reset it.

## Suggested direction
Scope the lease credit to the process (e.g., record owner PID + a process-wide flag) rather than the thread, so any handler thread of the holding process passes re-entrance; or serialize mutating tool execution onto a single dedicated writer thread in HTTP mode.

Context: we run 3.5.0 as a long-lived loopback HTTP service (Windows/NSSM) and currently route all mutating tools through short-lived CLI invocations as a workaround. Related: #2411 (startup quick_check gate). Happy to test a patch.

Beowulfn4756:
[...]

PR #2436, opened 2026-09-03
test(mcp): e2e regression guard for #2413 — two sequential writes over HTTP both succeed

BuboTheWise:
Branch: `fix/2413-process-scoped-writer-lease`
Into:   `MemPalace/develop`
SHA:    65e731baf8d088b482d874d493f84d08c1356567
Diff:   1 file changed, 118 insertions(+) — `tests/test_mcp_http_transport.py`

## Problem (#2413)

Reported on v3.5.0: with the HTTP transport in writable mode, the FIRST mutating
`tools/call` acquires the per-palace writer lock and holds it for the process
lifetime; the SECOND call — landing on a different ThreadingHTTPServer worker
thread — got no re-entrance credit (credit was thread-local,
`_palace_lock_holders = threading.local()`), so it re-opened the lock file and hit
the server's OWN flock, returning `MineAlreadyRunning: another mine is in progress
… held by PID <server's own PID>`. The server was colliding with itself.

## Finding

**Both root causes are already fixed and landed in upstream/develop (3.9.0):**

- `faa8643` (PR #1859) — `fix(palace): process-wide mine_palace_lock re-entrancy`.
  Replaces the thread-local holder set with a process-wide credit
  (`_palace_lock_guard` + `_palace_lock_keys` keyed on this process's PID); any
  thread in the holding process short-circuits re-acquisition. This is Site B.
- `9e871b4` — `fix(mcp): self-heal the writer lease instead of latching`. The MCP
  writer lease is no longer sticky-read-only; a failed acquire is re-attempted
  each call instead of latching (self-heal). This is the Site A self-heal.

I confirmed both commits are ancestors of `upstream/develop` via
`git merge-base --is-ancestor`. No open PR currently references #2413.

## What this PR adds (the missing deliverable)

An **e2e regression test** that no other test covered: two sequential mutating
`tools/call` POSTs against the *production* `_build_http_server()` server (writable
mode), asserting BOTH succeed and both writes land. It drives the full dispatch
chain that used to break:

```
_http_dispatch → (exclusive lock) → handle_request
  → _mcp_peer_writer_refusal → _acquire_mcp_writer_lock
        call 1: lease acquired (mine_palace_lock → flock), _MCP_WRITER_LOCK_CM set
        call 2: _MCP_WRITER_LOCK_CM present → short-circuit (no re-acquire)
  → tool_diary_write → _get_collection(create=True) → col.add(...) → mine_palace_lock  ← Site B
```

The mock collection's `add()` re-enters the **real** `mine_palace_lock(path)`
(mirroring `ChromaCollection._write_lock`, backends/chroma.py:1901) on the worker
thread, so Site B is genuinely exercised — not bypassed.

## Mutation proof (the test is a genuine guard, not an accident)

I patched `_palace._held_by_this_process` so the credit is invisible to non-main
threads (simulating the old thread-local credit). Result:

```
under MUTATION: call1 success=False  call2 success=False
  call2 error='another mine is in progress: palace … is held by PID 866221 (…)'
  -> 'both succeed' assertion holds = False   (want False)
MUTATION-PROOF PASS: the e2e test FAILS under the old behavior => genuine regression guard.
```

Under the fix: both calls `success=true`, 2 writes recorded, lease held process-wide.

## Verification

- `pytest tests/test_mcp_http_transport.py::test_two_sequential_writable_mutations_over_http_both_succeed -v` → **1 passed**
- Repeated 5× on fresh ephemeral ports → **1 passed×5** (no locking flakiness)
- `pytest tests/test_mcp_http_transport.py tests/test_palace_locks.py -q` → **59 passed** (up from 58; new test coexists with the existing writable-lease tests)
- `ruff check tests/test_mcp_http_transport.py` → **All checks passed**
- OPSEC sweep of diff → clean (no credential / agent-name / absolute-path leaks)

## Review notes for Bubo

- Test-only change; no production code touched (the fix is already in upstream/develop).
- If preferred, this can be merged directly to `MemPalace/develop` as a regression guard
  and issue #2413 closed with a pointer to `faa8643` + `9e871b4` + this PR.
- Suggested issue close line: "Fixed in 3.9.0 (PR #1859 + self-heal). Regression guard: PR <N>."

### DIFF
--- a/tests/test_mcp_http_transport.py
+++ b/tests/test_mcp_http_transport.py
@@ -627,6 +627,124 @@ def test_writable_http_releases_writer_lease_after_bind_failure(monkeypatch):
     assert mcp._MCP_WRITER_LOCK_CM is None
 
 
+def test_two_sequential_writable_mutations_over_http_both_succeed(http_server, monkeypatch, tmp_path):
+    """Regression for #2413: two sequential mutating tools/call over HTTP.
+
+    The original bug (v3.5.0): the writer-lease re-entrance credit was
+    thread-local (``_palace_lock_holders = threading.local()``). ThreadingHTTPServer
+    serves each request on a fresh thread, so the FIRST mutating call acquired the
+    per-palace lock and held it for process lifetime; the SECOND call — landing on
+    a different handler thread — was not credited as holder, re-opened the lock
+    file, and hit the server's OWN flock → ``MineAlreadyRunning`` naming the
+    server's own PID.
+
+    The fix (PR #1859, landed in 3.9.0): the re-entrance credit is
+    PROCESS-scoped (module-global ``_MCP_WRITER_LOCK_CM`` checked at the top of
+    ``_acquire_mcp_writer_lock``), so handler thread identity is irrelevant.
+
+    This test drives two real ``tools/call`` POSTs to the production
+    ``_build_http_server`` server in writable mode and asserts BOTH succeed.
+    It exercises the full dispatch path:
+        _http_dispatch → (exclusive lock) → handle_request
+          → _mcp_peer_writer_refusal → _acquire_mcp_writer_lock
+              1st call: lease acquired, _MCP_WRITER_LOCK_CM set (process-scoped)
+              2nd call: _MCP_WRITER_LOCK_CM present → (True, "") short-circuit
+          → tool_diary_write → _get_collection → col.add(...)
+    """
+    port, _ = http_server
+
+    # --- writable mode, no peer conflict ---
+    monkeypatch.setattr(mcp, "_READ_ONLY", False)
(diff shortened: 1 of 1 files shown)


BuboTheWise:
Linter format fix pushed — b11fe97 — ruff format --check now clean; re-ran on the new head. Confirmed all checks green: lint ✅ · build ✅ · build-gpu ✅ · smoke ✅ · test-linux (3.9/3.11/3.13) ✅ · test-macos ✅ · test-windows ✅ (run 35164218507). Format-only change on tests/test_mcp_http_transport.py (test_mcp_http_transport.py: def line-wrap + 3 monkeypatch collapses, 10 ins/6 del, zero logic); #2413 regression monkeypatch guard preserved.

Issue #2122, opened 2026-08-01
A failed upsert sub-batch marks the file fully mined, so the retry never happens


joshuafontany:
`miner.py` stamps `source_mtime` into every batch's metadata before the upsert loop (:1565). A failed batch increments `batches_failed` (:1588) and continues, so later successful batches still carry the current stamp. `file_already_mined()` returns True when any stored group reads current (:1256-1261), so the next run skips the file and the logged "re-run to retry the missing spans" never fires.

Two adjacent effects:

- `drawer_ids` builds from every chunk (:1616-1618) while `all_metas` holds only successful batches (:1600) — the closet gets pointers to drawers nothing filed.
- `collection.delete` purges the file's stale drawers *before* the loop, so a mid-run failure leaves the file holding less than it started with.

`batches_failed` is consulted only for a log line (:1602).

Minimal fix: skip stamping `source_mtime` when `batches_failed > 0`, so the next run re-mines. Deterministic drawer ids already make re-ingest idempotent. Finer resume (`partial=true` / `failed_spans`) reads as an escalation rather than a first step.

I can send the minimal fix if that suits.

zrh805:
Implemented in #2151. The fix clears drawers and the closet index after any batch-upsert error, so a partial successful batch cannot make the next run incorrectly skip the retry. It preserves the original error and includes a regression test for a second-batch failure.

PR #2439, opened 2026-09-04
fix(convo_miner): incremental re-mine — skip unchanged drawers, delete only orphans (#2336)

### LINKED ISSUE #2336: convo miner re-mines whole transcripts: measured 95.1% of each re-mine re-embeds byte-identical content
### Summary

`convo_miner._file_chunks_locked` deletes every drawer derived from a transcript
and then re-chunks, re-embeds and re-upserts the whole file. For an append-only
transcript — which is exactly what a live Claude Code session is — that means
~95% of each re-mine is recomputing vectors for content that is byte-identical
to what is already stored.

The design already contains everything needed to avoid this. This issue reports
the measurement and a minimal patch.

### The measurement

On a live 86.4 MB transcript with 5,842 stored drawers, re-deriving the chunks
with the current pipeline and comparing each against the stored document for the
same drawer id:

```
stored drawers            : 5,842
chunks after re-derivation: 6,141
  byte-identical to stored: 5,842   (95.1%)
  content differs         :     0
  genuinely new           :   299
```

**Zero chunks changed.** The session simply grew a 299-chunk tail. A re-mine
deletes and re-embeds 5,842 drawers to capture 299. On an active install this
repeats roughly every 30 minutes per live session.

### Why it happens

The only skip is `_is_unchanged_since_last_mine`, a whole-file exact-mtime
equality test. A live session appends constantly, so it always returns False,
and the response to "changed" is an unconditional full rebuild:

```python
delete_ids = _source_file_delete_ids(collection, source_file, extract_mode)
if delete_ids:
    collection.delete(ids=delete_ids)
```

The one content-level dedupe (`_split_new_and_duplicate_conversations`) hashes
the whole normalized conversation, which for a single session is one hash —
appending one line invalidates it. There is no offset tracking, no size delta and
no per-chunk comparison.

### Why an incremental path is straightforward here

Two properties already hold. Neither needs to be introduced.

**1. Drawer ids are positional and content-free.** `make_convo_drawer_id` hashes
`source_file|extract_mode|chunk_index`. Chunk N of a transcript always resolves
to the same id, so an upsert overwrites in place — the purge was never needed for
id correctness. (Verified: recomputed ids matched stored rows exactly.)

**2. Chunking walks conversation turns forward.** `_chunk_by_exchange` is "one
user turn (`>`) plus the AI response that follows", scanning forward, so
appending cannot reshuffle earlier chunks. The 0-differing result above is the
empirical confirmation.

What the purge genuinely protects against is **orphans**: a re-mine producing
fewer chunks, or routing a chunk to a different room (room is part of the id, and
`chunk_room` varies per chunk when `extract_mode == "general"`). That protection
should be kept — but it only requires deleting the stored ids this pass will not
rewrite, not all of them.

### Suggested change

1. Prefetch `id -> document` for the source file (mirroring
   `_source_file_delete_ids`, adding `documents` to `include`).
2. Precompute this pass's ids; delete only stored ids absent from that set.
3. In the batch loop, skip any chunk whose stored document is byte-identical.
4. Refresh `source_mtime` and `chunk_total` on the skipped ids with a
   metadata-only `collection.update()`. This matters: `prefetch_mined_set` groups
   by `(source_file, source_mtime)` and requires a group to reach `chunk_total`
   before treating the file as mined, so leaving skipped rows on the old mtime
   would split that bookkeeping.

Step 4 is metadata-only — no documents, no embeddings — so no vector is read or
written, and a backend that merges metadata leaves entities, novelty tags and
everything else intact.

It should fail safe: if the prefetch fails, fall back to the existing full purge.
If the metadata refresh fails, the drawers simply keep their old mtime and the
file is re-mined next pass — wasteful, never lossy.

### Verification

End-to-end, mining a transcript into an isolated database, then appending to it
and re-mining, counting the texts actually sent to the embedding endpoint:

```
                      drawers        embedded
before  run 1          71              71
        append
        run 2          71 -> 232      232      <- 71 redundant
after   run 1          71              71
        append
        run 2          71 -> 232      161      <- exactly the new ones
```

Third case, `touch` the file so mtime changes but content does not:
**0 texts embedded, 232 drawers unchanged.**

Post-change integrity on the resulting 232 rows: 0 null embeddings, 0 empty
documents, a single distinct `chunk_total`, a single distinct `source_mtime`, all
rows retaining their `entities`, and 232 distinct `chunk_index` values with no
gaps or duplicates.

### Scope note

This does not make the *normalize + chunk* step incremental — the file is still
read and re-chunked in full each pass. Eliminating that needs offset tracking,
which is a larger change and interacts with `/compact` rewriting a transcript in
place. The change above removes the embedding and row-rewrite cost, which is the
dominant term.

### Related

#2183 (file-level already-filed tracking skips exchanges) · #2213 (dedupe by
filename rather than content) · #2115 · #930 · #2222 / #2225 (RFC 002 source-item
currentness) — all open; none provides an incremental path today.

### Environment

`mempalace` tracking `develop` (`3.8.0` unreleased), backend `postgres`,
`--mode convos --extract exchange`, Python 3.14, Linux.

BuboTheWise:
Fixes #2336

## Problem
A previously-mined conversation source that is re-mined (Claude Code /compact
re-chunking a session file, a file that shrank, a normalized re-run) used to
purge *every* drawer, destroy all its embeddings, and re-embed everything,
regardless of how many exchanges were actually byte-identical (#2336).

## Change
The incremental path keeps byte-identical drawers in place (no re-embed, no
re-file), upserts only the changed / new drawers, and deletes *only* the
stored ids this pass will not rewrite (the orphans a rewrite / compaction left
behind). A metadata-only `update()` re-stamps `source_mtime` / `chunk_total`
on the skipped rows so `file_already_mined` / `prefetch_mined_set` still see
one complete group at the file's current state.

Any case where the incremental path is not provably safe — first mine, no
stored rows, a normalize-schema bump, an interrupted prior pass (several
mtime-groups), or any prefetch backend error — falls back to the legacy full
rebuild exactly as before.

The partial-cleanup-on-upsert-failure contract from #2122 / #2183 / #2403 is
retained in both paths.

## Verification
- `ruff check .` — exit 0
- `ruff format --check .` — exit 0
- `pytest tests/test_convo_miner_unit.py` — 67 passed

### DIFF
--- a/mempalace/convo_miner.py
+++ b/mempalace/convo_miner.py
@@ -290,6 +290,100 @@ def _source_file_delete_ids(collection, source_file: str, extract_mode: str) -> 
     return ids
 
 
+def _source_file_stored_drawers(collection, source_file: str, extract_mode: str) -> dict:
+    """Fetch ``{drawer_id: (document, metadata)}`` for one source file + mode.
+
+    Same row scoping as ``_source_file_delete_ids`` (which decides which
+    drawers belong to this convo-miner pass), but also returns the documents
+    so an incremental re-mine can byte-compare each stored chunk against the
+    re-derived one and skip the identical ones (#2336). Raises on backend
+    errors so the caller can fall back to the legacy full-purge path.
+    """
+    stored: dict = {}
+    offset = 0
+    while True:
+        batch = collection.get(
+            where={"source_file": source_file},
+            limit=1000,
+            offset=offset,
+            include=["metadatas", "documents"],
+        )
+        batch_ids = batch.get("ids") or []
+        metadatas = list(batch.get("metadatas") or [])
+        documents = list(batch.get("documents") or [])
+        for idx, drawer_id in enumerate(batch_ids):
+            meta = metadatas[idx] if idx < len(metadatas) else None
+            if not _metadata_matches_extract_mode(meta or {}, extract_mode):
+                continue
+            doc = documents[idx] if idx < len(documents) else None
+            stored[drawer_id] = (doc, meta or {})
+        if not batch_ids:
+            break
+        offset += len(batch_ids)
+    return stored
+
+
+def _plan_pass_drawers(
+    chunks,
+    wing,
+    room,
+    source_file,
+    agent,
+    extract_mode,
+    authored_at,
+    filed_at,
+    source_mtime,
+    chunk_total,
+    content_hash,
+):
+    """Build this pass's ``(docs, ids, metas)`` plus per-room counters.
+
+    Pure construction — the collection is never touched — so the legacy
+    full-purge path and the incremental path derive identical metadata from
+    the same code. The incremental path additionally byte-compares these
+    documents against the stored rows before upserting (#2336).
+    """
+    room_counts_delta: dict = defaultdict(int)
+    docs: list = []
+    ids: list = []
+    metas: list = []
+    for chunk in chunks:
+        chunk_room = chunk.get("memory_type", room) if extract_mode == "general" else room
+        if extract_mode == "general":
+            room_counts_delta[chunk_room] += 1
+        drawer_id = make_convo_drawer_id(
+            wing, chunk_room, source_file, extract_mode, chunk["chunk_index"]
+        )
+        docs.append(chunk["content"])
+        ids.append(drawer_id)
+        meta = {
+            "wing": wing,
+            "room": chunk_room,
+            "hall": _detect_hall_cached(chunk["content"]),
+            "source_file": source_file,
+            "chunk_index": chunk["chunk_index"],
+            "added_by": agent,
+            "filed_at": filed_at,
+            "entities": entities_metadata(chunk["content"]),
+            "authored_at": authored_at if authored_at is not None else filed_at,
+            "ingest_mode": "convos",
+            "extract_mode": extract_mode,
+            "normalize_version": NORMALIZE_VERSION,
+            "id_recipe": ID_RECIPE,
+            "chunk_total": chunk_total,
+        }
+        if source_mtime is not None:
+            meta["source_mtime"] = source_mtime
+        # Stamp content_hash only on chunk 0 so multi-conversation
+        # privacy-export hashes are not O(N²)-duplicated across every
+        # chunk row. ``prefetch_content_hashes`` still finds them —
+        # it scans all drawers and splits comma-joined hash fields.
+        if content_hash is not None and chunk.get("chunk_index", 0) == 0:
+            meta["content_hash"] = content_hash
+        metas.append(meta)
+    return docs, ids, metas, room_counts_delta
+
+
 # =============================================================================
 # CHUNKING — exchange pairs for conversations
 # =============================================================================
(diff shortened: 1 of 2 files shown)


BuboTheWise:
Audit 2026-09-06 (Cthugha, t_b71fd8b9)

Verdict: NO-ACTION-WAIT-ON-MAINTAINER

Gates:
1. #2336 still OPEN; no merged PR supersedes it. (#2408 covers adjacent #2403 Stop-hook/self-DoS defects, still OPEN, different scope.)
2. Base d9f0590 == upstream/develop head d9f0590. mergeable=MERGEABLE.
3. CI on head 638cbc5: Tests + Docker both completed/action_required — first-time-fork Actions approval pending on the BuboTheWise fork. Author-side gate, not an upstream-maintainer gate.
4. OPSEC commit + diff clean (no local home paths, no t_<8hex>, no ~/.hermes, no agent names in added lines).
5. Diff: 2 files, +659/-99 (mempalace/convo_miner.py + tests/test_convo_miner_unit.py) — scoped to the incremental re-mine path with the legacy rebuild preserved as the unsafe-case fallback.

Action pending (BuboTheWise): approve the fork Actions run for head 638cbc5 so Tests + Docker actually execute on the upstream CI badge.

mempalace/convo_miner.py
```diff
@@ -625,6 +719,148 @@ def _extract_authored_at(filepath):
...
+                except Exception as e:
+                    if "already exists" not in str(e).lower():
+                        raise
+        drawers_added = len(upsert_ids)
+    except Exception:
+        # Upsert failed after the orphan purge already ran:
+        # this file now has fewer drawers than it started
+        # with, so do the partial-cleanup delete too — the
+        # next mine re-files the full set from scratch.
+        cleanup_partial()
```
igorls, line 823: [P1] Do not delete unchanged prior drawers when a new append fails

This handler calls source-wide `cleanup_partial()` even when `orphan_ids` was empty and the existing drawers are still valid. With one unchanged stored chunk and one new chunk whose embedding/upsert fails, the real cleanup path deletes the old chunk and leaves the collection empty. The outer exception handler can repeat cleanup. This fails the documented requirement that a crash leave the existing palace untouched. Keep the previous complete set and track newly created partial rows separately; add a failing-append regression that asserts the old drawer and its exact content survive.

igorls (changes requested):
Requesting changes on the incremental failure path and the preservation contract.

I reproduced the cleanup problem through the actual `_file_chunks_locked` -> `_apply_incremental_pass` -> cleanup helpers, with a controlled collection containing one unchanged prior drawer and one newly appended chunk. Making the new chunk's upsert raise an embedding error deletes the unchanged old drawer and leaves the collection empty. No orphan deletion is needed to trigger this. The supplied unit module passes: 64 passed, 3 skipped.

Please preserve the previous complete set on failure instead of performing source-wide cleanup, and add a regression for an unchanged old drawer plus a failed new append.

Separately, the orphan deletion and in-place positional upserts still discard previous conversational content when a transcript shrinks or changes. That conflicts with the additive-history requirement in #1593 and `CLAUDE.md`. Please resolve this through preserved version identities/currentness rather than treating an absent current chunk as authorization to delete its history.

BuboTheWise:
@igorls Thanks for reproducing this — the scenario is exactly the one the incremental path got wrong, and I agree with the contract.

**[P1] Unchanged prior drawer deleted when the new append fails.** Confirmed in code: in `_apply_incremental_pass`, the orphan purge runs first (`orphan_ids` empty when the file only grew) and then the upsert `except` path calls `_cleanup_partial()` unconditionally — which is `_source_file_delete_ids()` over the *whole* source. So one unchanged prior drawer + one new chunk whose upsert throws = the old drawer is deleted and the collection is empty, with no orphan deletion involved. That breaks the documented "a crash leaves the existing palace untouched" requirement. Fix I'm queueing: on an upsert failure in the incremental path, clean only the new/changed rows this pass already wrote (the `upsert_ids` set) rather than the source-wide `cleanup_partial()` — the byte-identical survivors stay intact, and the next re-mine re-files the changed rows. That's the "track newly created partial rows and clean only those" form of the same #2122/#2183/#2403 contract, scoped to the incremental path. I'll add the exact regression you described — one unchanged prior drawer plus one new append whose embedding/upsert raises — asserting the old drawer and its exact content are still present and the new row is absent, and I'll make it fail on the current head and pass on the fixed one.

**Positional upserts discarding history on shrink/re-chunk (#1593 additive-history).** I want to take this as a design pass rather than paper over it: the real requirement is that an *absent current chunk* is not authorization to delete prior conversational content, and the correct mechanism is version-identity/currentness (which chunk-generation a drawer belongs to, and which generation is currently valid) rather than positional id equality. I'll scope a small design note + the identity-tracking shape for that next, and bring it back before treating this PR as fully resolved — I don't want to claim the P1 and the additive-history contract are the same fix when they're not.

On process: the self-review "PASS" was mine — re-calling it as a checkpoint, agreed it's not a second pair of eyes. And the fork first-time Actions approval is still pending on my side (author gate, not upstream), so I'll approve it and confirm CI runs on the new head. I'll update this thread once the P1 fix is green and re-base.

BuboTheWise:
P1 fix pushed — `88900cd` (live head `526506f` after a ruff-format pass) — the incremental upsert-failure path now rolls back only the new/changed rows this pass already wrote (`upserted_ok_ids`) instead of the source-wide `cleanup_partial()`, so byte-identical prior drawers stay intact when a later append fails to upsert. The outer handler is guarded by a rolled-back flag, so cleanup does not repeat.

Regression test in place: `TestFileChunksLockedIncremental::test_upsert_failure_preserves_unchanged_drawers` — it fails on the prior head (`638cbc5`, unchanged drawer wiped, `rows={}`) and passes on the fixed head (drawer preserved verbatim, new row absent).

CI is green on the new head (Tests + Docker both passing).

One note for merge-gate awareness: the local full suite carries a single pre-existing failure in `test_repair.py` (a Python-version-gated path-encoding test) that fails identically on the base head and lives in a module this PR does not touch — it is not a regression of this fix; tracked separately.

Re-ran on the new head (`526506f`); reviewer re-review requested.

BuboTheWise:
Rebased onto `develop` (head `36ec72f`) to resolve the drift — the branch was DIRTY/stale against `develop`. New head is `568ed08`, and the previous `CHANGES_REQUESTED` review was on the pre-rebase head, so it will need a re-review to clear.

**What the rebase touched** (base-drift reconciliation only, no behavior change to the incremental re-mine):
- `develop` landed #2320 (source-directory identity) and the Codex unparsed-transcript early-return between our merge-base and `develop`. Both touched the same metadata-building region my refactor moved into `_plan_pass_drawers`, so I re-ported develop's `source_dir_ino` stamping into that helper and kept develop's `UnparsedCodexTranscriptError` handling — the incremental "skip unchanged drawers, delete only orphans" logic is unchanged.
- `tests/test_convo_miner_unit.py` re-merged cleanly (develop's new source-directory tests are present and passing alongside ours).

**Verification on the new head:**
- CI: 12/12 checks SUCCESS (Linux 3.9/3.11/3.13, Windows, macOS, GPU build, lint, Rust core/CLI, smoke).
- Local isolated-venv run of `tests/test_convo_miner_unit.py`: 71/71 passed.
- ruff check + format clean on both changed files.

Ready for a re-review whenever convenient.

BuboTheWise:
@igorls — bumping for re-review on the refreshed head `568ed08` (rebased onto `develop`). The incremental re-mine behavior is unchanged; the only content change is re-porting develop's new `source_dir_ino` stamping into `\_plan_pass_drawers`. CI 12/12 green + conve\_miner 71/71 local. Your earlier P1 is addressed; just needs a second pass on the rebase. Thanks!

BuboTheWise:
Ready for final look — rebase to 568ed08 (base develop d9f0590) is behavior-identical to the incremental re-mine; only develop source_dir_ino stamping was re-ported into _plan_pass_drawers. P1 rollback-scope fix is in, CI 12/12 green, conve_miner 71/71 local. Awaiting a second pass on the refreshed head. Thanks, @igorls!

PR #2447, opened 2026-09-05
docs(rfc006): Rooms — open discussion between agents and people

igorls:
## RFC 006 v2 — Rooms: open discussion between agents and people

**Complete rewrite** in response to @milla-jovovich's review. The v1 draft and its `examples/multi_agent_room/` simulation are gone; this PR is now **docs-only** (the RFC plus a Rooms section in `integrations/shared/coordination-protocol.md`) and retargeted at `develop`.

### What the review said, and what changed

> Complicated and doesn't account for agents that don't argue and when you just need to have one person at a time speak. Doesn't work with the agent that needs to speak on the desktop app.

All three points were right, and they had one root cause: v1 designed the *agents* (scripted personas in Python threads, a four-factor gating heuristic, jitter backoff, pre-flight cancellation) instead of the *room*. A real agent session — least of all a desktop chat app — could not join it.

v2 defines a room as something any participant joins **from the session it is already running in**, with the three operations every agent already has: catch up, speak or stay quiet, and find out when it is your turn *in the way your harness can*.

- **A room is a `room` on the project stream** — `stream=project/mempalace, room=search-brainstorm`. Reuses RFC 003's existing field; no new namespace, no new matcher, no schema change.
- **Six event types**: `room.open / join / floor / message / pass / close`. All valid today.
- **Moderated mode is the default: one speaker at a time.** The moderator hands the floor with `room.floor to_agent=X`; X posts one `room.message` (or `room.pass`); floor returns. Round-robin / standup / panel are moderator *policies*, not modes. Agents that don't argue simply answer when asked and are otherwise silent. **Open mode** is opt-in.
- **Wake models are declared on join** — `self` (has a background watcher; woken by events) or `turn-based` (desktop app; woken only by a pasted floor line). A turn-based agent participates within a single turn: catch up from its cursor, post once, report the cursor. Nothing spawned, no second window, full context because the context *is the room*. Same shape as `mempalace task create`'s "Ready to paste" line.
- **Anti-chatter is four lines of instructions, not code**: read everything since your cursor; post only if you add a fact, constraint, proposal, objection, or an answer to something addressed to you; never agree/acknowledge/restate; at most one message per wake. The per-wake bound is what makes an open room converge.
- **Closing files the transcript verbatim as drawers** plus the decision and KG facts — the same "file the outcome" rule delegations already follow.

### Why collisions and backoff are gone

The logstream is append-only and totally ordered (HLC). Two agents posting "at the same time" produce two events in a definite order; nothing can be corrupted. Backoff and cancellation were solving a race that wasn't one.

### Phases

- **Phase 1 (this PR)** — convention only. §7 of the RFC walks a full moderated session (one self-waking agent, one turn-based agent, one human) using nothing but existing `mempalace logstream` commands.
- **Phase 2** — `mempalace room open|join|floor|say|pass|catchup|close` sugar and a `mempalace-room` skill, after dogfood.
- **Phase 3** — a room view in PalaceMind with a "hand floor to…" button that copies the paste line.

### Related

- #2457 (@jjcav84) extended the v1 direction with participation modes and an async worker. The modes idea overlaps with moderated/open here; the async/threaded split addresses GUI event-loop starvation, which was not the desktop-app problem — the problem is that a turn-based agent cannot be woken by an event at all, only by a human. Happy to converge on one PR.

### Verification

- `uv run ruff check .` clean
- `uv run pytest tests/test_cli_rules.py tests/test_cursor_plugin_manifest.py` — 61 passed (the coordination-protocol snippet fence is untouched; the Rooms section sits outside it)


examples/multi_agent_room/room_prototype.py
milla-jovovich, line 1: It's really complicated for everyone codex Astra and it doesn't work at all for my got agent personally. It connects to the app and then spawns a brand new window so the instance there has no clue what he's doing there.

milla-jovovich (changes requested):
Complicated and doesn't account for agents that don't argue and when you just need to have one person's at a time speak. Doesn't work with the agent that needs to speak on the got desktop app.

jjcav84:
Hi @milla-jovovich — the CHANGES_REQUESTED here is on commit `9853c844` (the v1 with the simulation). igorls rewrote it as docs-only v2 in commit `db539739` addressing your three points (complicated, agents that don't argue, desktop app). Could you re-review the v2?

I've closed #2457 to converge on this PR. Happy to pick up Phase 2 (the CLI sugar: `room open|join|floor|say|pass|catchup|close`) as a follow-up once this lands — the mode-dispatch work from #2457 maps onto moderator policies there.

PR #2448, opened 2026-09-05
fix(chroma): close displaced client before clearing shared System cache (#2375)

BuboTheWise:
## What does this PR do?

Fixes a silent HNSW native-memory leak in the chroma rebuild path (`ChromaBackend._client`).

On the external-change rebuild branch, the displaced `PersistentClient` was released
**after** `_clear_chroma_system_cache()` had already emptied chromadb's shared registry.
`PersistentClient.close()` frees its native index memory through
`SharedSystemClient._release_system(self._identifier)`, but that call needs the client's
identifier to still be registered in `_identifier_to_system` / `_identifier_to_refcount`.
Because the clear ran first, those dicts were already `{}` — so `close()` became a silent
no-op (`_decrement_refcount` returned 0, `pop` returned `None`, `System.stop()` never
fired), and every external-change rebuild orphaned one HNSW segment set. This is the
native-index leak in the same stale-data family as #2002/#2028, and it compounds over a
long-lived server.

The fix reorders the rebuild branch so the displaced client is closed **first** (which
lets `System.stop()` fire and release the native HNSW memory), and only then does it drop
the shared System cache so the freshly constructed `PersistentClient` starts clean. The
gating on *genuine external change* (not first open) is preserved, so cold opens still pay
no global-evict cost. The DB-missing early path and the `close_palace`/`close` paths
already closed-before-cleared and are unchanged.

Two regression tests are added (`tests/test_backends.py`):

1. `test_chroma_backend_closes_displaced_client_before_clearing_cache` — pins the
   relative close-then-clear ordering on the rebuild branch. Under the old ordering it
   emits `[..., "clear", "close", "open"]` and **fails**; with the fix it emits
   `[..., "close", "clear", "open"]` and passes.
2. `test_chroma_displaced_client_close_stops_system_before_cache_clear` — proves the
   load-bearing mechanism against the shared registry directly: close-before-clear
   triggers `System.stop()`, clear-before-close does not (the bug).

Scope: `mempalace/backends/chroma.py` (reorder in `_client`, rebuild branch only) and one
new test block in `tests/test_backends.py`. No API, config, or dependency changes.

## How to test

```bash
pip install -e ".[dev]"

# Regression tests (both must pass; the ordering test fails on the pre-fix code):
python -m pytest tests/test_backends.py::test_chroma_backend_closes_displaced_client_before_clearing_cache \
                  tests/test_backends.py::test_chroma_displaced_client_close_stops_system_before_cache_clear -v

# Full suite + coverage gate (CI command):
python -m pytest tests/ -v --ignore=tests/benchmarks --cov=mempalace --cov-report=term-missing --cov-fail-under=80

# Lint / format (pinned ruff 0.16.1):
ruff check .
ruff format --check .
```

Verified locally (chromadb 1.5.7, ruff 0.16.1, Python 3.12): `ruff check .` clean,
`ruff format --check .` clean, full suite green (one pre-existing, unrelated
locale/path-encoding failure in `tests/test_repair.py` reproduces identically on the
`upstream/develop` baseline with this change stashed), and coverage 81.91% (≥ 80% gate).

## Checklist
- [x] No hardcoded paths

### DIFF
--- a/mempalace/backends/chroma.py
+++ b/mempalace/backends/chroma.py
@@ -2604,20 +2604,6 @@ class ChromaBackend(BaseBackend):
             # path); an mtime/appearance change means an external in-place
             # write (closet_llm, mine, compress) that may have drifted the
             # HNSW index while this process was running.
-            if (
-                inode_changed
-                or mtime_changed
-                or (mtime_appeared and palace_path in self._freshness)
-            ):
-                ChromaBackend._quarantined_paths.discard(palace_path)
-                # #2028: the same external change means chromadb's path-keyed
-                # System cache is now stale. Reconstructing PersistentClient
-                # below would reuse the cached System (and its in-memory HNSW
-                # segment), so drop the shared cache first -- otherwise the
-                # rebuilt client persists an outdated index over the on-disk
-                # change. Gated on genuine external change (not first open) so
-                # cold opens never pay the global-evict cost.
-                _clear_chroma_system_cache()
             # Release the client we are about to displace. Each live
             # PersistentClient pins its own copy of every HNSW segment it has
             # opened (``max_elements * size_data_per_element`` bytes -- ~440 MB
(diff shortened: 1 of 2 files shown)


BuboTheWise:
Closing as superseded — thank you for raising this earlier than we caught it (both our PR and #2421 traced the same root cause: `close()` running after chroma's shared-registry clear, so `System.stop()` silently no-ops and each rebuild orphans one HNSW segment set).

#2421 (fatkobra) landed the same close-before-clear semantics on 2026-09-12, with a slightly broader drain — `self._drain_clients()` closes **every** client owned by this backend before `_clear_chroma_system_cache()`, which covers the same #2375 failure mode plus the sibling-palaces case this PR did not reach. The regression test `test_chroma_client_rebuild_stops_each_displaced_system` (tests/test_backends.py:673) pins the fix at the exact line we identified.

Nothing in our branch is unique to keep. Closing.

BuboTheWise:
Superseded by #2421 (merged 2026-09-12): the close-before-clear ordering this PR added is now upstream via  before , with the #2375 regression test in place. No reviewer thread was open on this PR.

Issue #2438, opened 2026-09-03
bug(windows): harden migration temporary allocation and prevent partial wing-name updates


fatkobra:
## Summary

`mempalace/migrate.py` contains two temporary-allocation paths that remain outside #2420:

- a temporary directory used to construct a replacement palace
- an adjacent temporary file used to publish the rewritten `known_entities.json` during wing-name migration

These paths need bounded failure behavior and migration-specific atomicity tests rather than being folded into the config fix.

### Full-palace migration

`migrate()` creates a staging palace through `tempfile.mkdtemp()` before importing drawers and swapping the replacement palace into position.

This path should be audited separately from `mkstemp()`, because the exact standard-library exception handling may differ, but it must still fail promptly and leave the original palace untouched when staging allocation cannot succeed.

### Wing-name migration

`migrate_wing_names()` currently applies:

1. drawer metadata updates
2. closet metadata updates
3. the `topics_by_wing` rewrite in `known_entities.json`

The registry rewrite allocates an adjacent temporary file through `tempfile.mkstemp()` and publishes it with `os.replace()`.

If that allocation or publication fails after the collection updates, the migration can leave drawer/closet metadata normalized while `topics_by_wing` remains under the old keys.

## Impact

Potential outcomes include:

- excessive temporary-name retries on affected Windows/Python behavior
- an apparently hung migration
- a failed full-palace migration before staging begins
- a partially applied wing-name migration across drawers, closets, and `known_entities.json`
- mismatched search metadata and topic-registry keys

### Shared allocation behavior

- use a bounded secure temporary allocator
- retain exclusive creation
- distinguish genuine collisions from access failures
- clean up all temporary paths after failure
- avoid duplicating the private helper introduced in #2420

### Full-palace staging

- fail before changing the original palace when staging allocation is unavailable
- leave the original palace and its backup state intact
- emit an actionable error naming the failed staging location
- verify cleanup when construction fails after staging succeeds

### Wing-name migration

The operation should not leave collections and the registry in different naming states.

A complete solution should either:

- prepare and validate the registry replacement before mutating collection metadata, then publish it as part of a rollback-capable sequence; or
- snapshot the affected collection metadata and restore it if registry publication fails.

Simply replacing `tempfile.mkstemp()` without addressing the partial-update order would be insufficient.

## Regression coverage

Add tests proving that:

- denied staging allocation leaves the original palace untouched
[...]

Issue #2437, opened 2026-09-03
bug(windows): bound hallway sidecar temporary-file allocation


fatkobra:
## Summary

`hallways._save_hallways()` publishes the palace-scoped `hallways.json` through an adjacent `tempfile.mkstemp()` file followed by `os.replace()`.

PR #2420 demonstrated that, on affected Python/Windows behavior, repeated `PermissionError` results during temporary-name allocation can be interpreted as candidate collisions and cause an extremely large retry loop before eventually surfacing an unrelated-looking `EEXIST`.

The hallway writer retains the same unbounded standard-library allocation shape.

## Current path

`_save_hallways()` currently:

1. creates the destination directory
2. calls `tempfile.mkstemp()` beside `hallways.json`
3. writes the complete payload
4. applies restrictive permissions
5. publishes it with `os.replace()`

Temporary-file allocation occurs before the existing cleanup `try` block.

## Impact

When the destination directory refuses every adjacent temporary name, hallway persistence can hang for an impractical amount of time or fail only after exhausting the standard-library retry budget.

This can affect hallway recomputation and hallway deletion.

The previous `hallways.json` remains valuable state and must not be truncated merely because an adjacent temporary file cannot be created.

## Desired behavior

- place a deterministic upper bound on temporary-name allocation
- retain exclusive creation and cryptographically strong random names
- preserve the atomic `os.replace()` publication model
- preserve restrictive file permissions
- propagate a prompt, actionable filesystem error when the directory cannot accept a temporary file
- do not fall back to truncating the existing `hallways.json` in place
- do not leave a temporary file behind after write, chmod, or replace failures
- avoid copying a config-private helper into this module; use a shared internal atomic-file utility where practical

## Regression coverage

Add tests proving that:

- repeated `EPERM` or `EACCES` reaches the caller after a bounded number of opens
- the previous `hallways.json` remains byte-for-byte unchanged
- no temporary file remains after failure
- a genuine first-name collision is retried successfully
- a successful save still publishes complete JSON atomically
- restrictive permissions remain applied where supported

## Relationship

Follow-up to #2420 and the review observation identifying the remaining `hallways.py` allocation path.

This is intentionally separate from #2392 because config persistence has an explicit in-place fallback, while hallway persistence should preserve the existing sidecar rather than abandon atomic publication.

PR #2455, opened 2026-09-07
test(config): skip EPERM rewrite test on Windows

### LINKED ISSUE #2392: `test_eperm_on_both_names_falls_back_too` is missing `@needs_unprivileged_posix`, and on Windows it does not fail — it spins for ~68 hours

BuboTheWise:
## What does this PR do?

`test_eperm_on_both_names_falls_back_too` in `tests/test_config_unreadable_rewrite.py` is missing the `@needs_unprivileged_posix` marker that its ten sibling permission-injection tests all carry. On Windows, its `os.open` monkeypatch routes every `tempfile.mkstemp` attempt into CPython's Windows-only `PermissionError -> continue` branch; with `tempfile.TMP_MAX` at 2,147,483,647, the test spins for hours without converging (reported ~13,527 s of CPU) instead of being skipped.

This adds the single missing marker, matching the EROFS sibling `test_erofs_on_both_names_falls_back_rather_than_raising`. The test now skips on Windows (`os.name == "nt"`) and as root, and is unchanged on POSIX/non-root.

Closes #2392.

## How to test

- `ruff check tests/test_config_unreadable_rewrite.py` + `ruff format --check ...` — clean (ruff 0.16.1, the CI pin).
- `pytest tests/test_config_unreadable_rewrite.py` — 34 passed on Linux/non-root (target test + EROFS sibling both PASS, not skipped).
- Diff: 1 file, 1 insertion, 0 deletions — the `@needs_unprivileged_posix` line only.

## Checklist
- [x] No hardcoded paths


mvalentsev:
This closes #2392 while leaving the defect that issue describes in place.

#2420, opened five days earlier, replaces both `tempfile.mkstemp` call sites in `config.py` with a bounded allocator and rewrites this same test to assert the bound. Measured on merges onto develop `5a2abffe`, with CPython's Windows `PermissionError -> continue` branch forced on under Linux and every `config.json.*` open refused:

| merge | temp-file open attempts | outcome |
| --- | --- | --- |
| this PR | 20000, still going | safety cap hit |
| #2420 | 2 | setting saved, user told the write went in place |

After this PR, `config.py:415` and `config.py:522` still call `tempfile.mkstemp`, so the spin the issue reports is untouched.

The two PRs do not conflict, so both can land. If they do, the marker added here skips, on Windows, the very test #2420 rewrote to prove the bound holds. The skip reason reads "directory permission bits gate neither root nor Windows", and this test reads no permission bits: it monkeypatches `os.open`. The one thing that made it unrunnable on Windows was the unbounded loop, which #2420 removes.

The remaining allocation sites are already filed as #2437 and #2438.

Please check whether the work an issue needs has already been done before opening a PR against it, or have your agent check.

PR #2457, opened 2026-09-07
fix(rfc006): add participation modes and async/desktop compatibility per reviewer feedback

jjcav84:


jjcav84:
The `test-windows` failure is a pre-existing flaky daemon timing test, not caused by this PR. The latest `develop` (#2465 merge, run [34270414218](https://github.com/MemPalace/mempalace/actions/runs/34270414218)) has the same `test-windows` failure with a *different* daemon test timing out (`test_shutdown_while_a_job_is_cooling_leaves_it_queued`).

On this PR's run, `test_repeated_lease_refusal_never_exhausts_attempts` timed out after 4390 tests passed. On develop, that same test *passed* and a different one timed out. This is a Windows CI timing issue with daemon lease tests, not an RFC 006 regression.

All other checks pass: smoke, test-linux (3.9/3.11/3.13), test-macos, build, build-gpu, lint.

Ready for review.

jjcav84:
Rebased onto latest `develop` (was 22 commits behind, previously targeting `main` — now correctly targeting `develop` per the branching model). Full test suite passes (4869 passed, 53 skipped). Ruff clean. Ready for review.

igorls:
Thanks for picking this up so quickly, @jjcav84 — the modes split (round-robin / sequential / broadcast) is the right instinct, and the "agents that don't argue" point was the one v1 missed most.

Heads-up so we don't fork the work: I've just **rewritten RFC 006 from scratch** on #2447 (now docs-only, retargeted at `develop`). The rewrite goes in a different direction than this PR on two points, and I want to be upfront about why:

1. **The desktop-app problem isn't event-loop starvation.** Milla's agent lives in a chat app: it only acts when a human gives it a turn, so *no* worker — threaded or async — can wake it from a logstream event. Swapping `time.sleep` for `asyncio.sleep` keeps the same "the room is a Python process" model that excludes those agents. v2 instead has each participant declare a **wake model** on `room.join` (`self` = has a background watcher; `turn-based` = woken only by a pasted floor line) and defines participation as three operations any already-running session can do with its existing tools: catch up from your cursor, post once, report the cursor. Nothing is spawned.

2. **The prototype grew instead of shrinking.** The review's first word was "complicated"; v2 answers that by removing the simulation entirely and making Phase 1 pure convention on top of existing `mempalace logstream` commands (§7 of the RFC is a full moderated session, zero new code). The modes collapse to two — **moderated** (one speaker at a time, moderator hands the floor; round-robin/standup are moderator policies) and **open** — with a four-line anti-chatter rule in the participant's instructions rather than a scoring function.

Full write-up in the PR description and in the RFC's "What the first draft got wrong" section. If the direction in #2447 lands, the CLI sugar (`mempalace room open|join|floor|say|pass|catchup|close`) is Phase 2 and would be a great place for a follow-up if you're interested — that's where the mode-dispatch work you did could map naturally onto moderator policies. Happy to converge on one PR either way.

jjcav84:
Thanks @igorls for the heads-up. I read the v2 rewrite on #2447 — the wake model declaration (`self` vs `turn-based`) and the moderated/open mode collapse are the right call. The v1 simulation was over-engineered.

I'm happy to close this PR in favor of #2447 if that's the direction the maintainers want. The mode-dispatch work (round-robin / sequential / broadcast) I did here could map naturally onto moderator policies in Phase 2 as you suggested — I'd be glad to pick that up as a follow-up once #2447 lands.

Let me know if you'd prefer I close this or keep it open as an alternative. Either way, no fork in the road — I'll align with whatever @milla-jovovich and the team decide.

jjcav84:
Closing this to converge on #2447 — igorls's v2 rewrite addresses @milla-jovovich's review feedback (complicated, doesn't account for agents that don't argue, desktop app) directly by removing the simulation and collapsing to moderated/open modes with wake model declarations. No point maintaining two RFC006 implementations.

The mode-dispatch work here (round-robin / sequential / broadcast) maps naturally onto Phase 2 moderator policies — I'd be glad to pick up the CLI sugar (`room open|join|floor|say|pass|catchup|close`) as a follow-up once #2447 lands.

PR #2458, opened 2026-09-07
fix(kg): return source_file from query_entity

realburhanhusain:
`query_entity()` builds its result dicts with `source_closet` but not `source_file`, so `kg_query` drops the provenance that `kg_add` stored. The `triples` table holds both columns and the row is already selected with `SELECT t.*`, so the value is fetched and then discarded.

`source_closet` is null for any fact added directly rather than mined from a closet, which makes `source_file` the only provenance a manually-added fact can carry. Callers reading `kg_query` output have no way to tell where such a fact came from.

This adds `source_file` to both the outgoing and incoming result dicts, matching the field already accepted by `add_triple` and exposed by the `kg_add` MCP tool.

### Tests

Three regression tests in `tests/test_knowledge_graph_metadata.py` cover the outgoing edge, the incoming edge, and the not-supplied-is-None case. All three fail with `KeyError: 'source_file'` without the two-line change and pass with it. The wider knowledge-graph suite (70 tests) passes unchanged.

### DIFF
--- a/mempalace/knowledge_graph.py
+++ b/mempalace/knowledge_graph.py
@@ -525,6 +525,7 @@ class KnowledgeGraph:
                             "valid_to": row["valid_to"],
                             "confidence": row["confidence"],
                             "source_closet": row["source_closet"],
+                            "source_file": row["source_file"],
                             "current": row["valid_to"] is None,
                         }
                     )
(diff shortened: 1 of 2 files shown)


igorls:
Logic and tests look good — verified `source_file` comes back on both the outgoing and incoming branches, and `tests/test_knowledge_graph_metadata.py` + `tests/test_knowledge_graph.py` pass locally (53 passed).

One blocker before merge: `ruff format --check .` fails on the new test file. The new class needs two blank lines after the previous method:

```diff
         assert stats["relationship_types"] == ["loves", "works_at"]
+
+
 class TestQueryEntityProvenance:
```

`uv run ruff format tests/test_knowledge_graph_metadata.py` will fix it. Push that and this is ready.

PR #2467, opened 2026-09-09
fix(mine): resolve relative daemon source against the caller's cwd

### LINKED ISSUE #2441: bug(daemon): mine submitted with a relative source is mined from the daemon's cwd
**What happened?**

`mempalace mine . --wing X --daemon --background` (the form a Stop hook runs from the repo directory) puts the literal `.` in the job payload (`cli.py`: `"source": args.dir`). `service.run_mine` then mines whatever directory the daemon process was started from. The daemon inherits the cwd of the process that autostarted it, so every hook-driven mine since the daemon appeared indexed that one directory into the current session's wing. On this machine: 618 foreign drawers in wing `dev_inserm` and 345 in `BIGBARBER`, all from the repo whose hook had started the daemon. The direct path is fine; only the daemon path loses the caller's cwd.

**What did you expect?**

The CLI to resolve the source to an absolute path before submitting, as `_forward_mine_to_hub` already does (`os.path.abspath(os.path.expanduser(args.dir))`).

**How to reproduce:**

1. `cd /repo-a && mempalace mine . --wing a --daemon` (autostarts the daemon with cwd = repo-a)
2. `cd /repo-b && mempalace mine . --wing b --daemon --background`
3. `mempalace search "<text that only exists in repo-a>" --wing b` returns repo-a files

**Environment:**
- OS: Windows 11
- Python version: 3.13
- MemPal version: 3.9.0 (uv tool), chromadb 1.5.9

AmirF194:
`cmd_mine`'s `--daemon` path put the raw `args.dir` straight into the job payload's `source` field. The daemon is a long-lived process with its own cwd, set once at whatever moment it happened to autostart, so a relative source like `.` resolved against that stale cwd instead of the directory the CLI call actually ran from. Every hook-driven mine after the daemon appeared indexed one fixed directory into whatever wing the caller thought it was mining into.

`_forward_mine_to_hub` already resolves the same field with `os.path.abspath(os.path.expanduser(args.dir))` before handing a mine to the palace hub; the daemon path just never got the same treatment. This applies that one line to the daemon payload too.

Added a test that chdirs into a tmp dir, submits `mine . --daemon --background`, and asserts the payload's `source` is the resolved absolute path: fails on `develop` (asserts `.` against the tmp dir), passes with the fix.

`cmd_sync`'s daemon payload has the identical unresolved `args.dir` field. Didn't touch it here since issue #2441 is specifically about `mine`, but it's the same shape if you want it as a follow-up.

Fixes #2441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ### Bug Fixes
 
 - **The transcript-path fallback no longer gives every git worktree its own wing.** `_wing_from_transcript_path`'s primary path (reading `cwd` from the JSONL) already collapsed a `<project>/.claude/worktrees/<wt>` segment before deriving the wing; the fallback path, used whenever `cwd` is absent, had no equivalent strip, so the flattened `--claude-worktrees-<wt>` segment survived into the wing name. Applied the same collapse there. (#2388)
+- **`mempalace mine --daemon` no longer mines the daemon's own working directory instead of the caller's.** The daemon is a long-lived background process that keeps whatever cwd it happened to start with, so a relative source (`mempalace mine .`) submitted to it resolved against that stale cwd rather than the directory the CLI call actually ran from, silently mining the wrong project into the wrong wing on every subsequent hook-driven call. `cmd_mine` now resolves the source to an absolute path before it enters the daemon job payload, matching the resolution `_forward_mine_to_hub` already does for the hub-forwarding path. (#2441)
 
 ---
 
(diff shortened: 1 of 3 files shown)


igorls:

AmirF194:
Rebased onto develop and moved the fix into cmd_mine.py in the new cli/ package. Tests and ruff pass locally, back to you.

igorls (approved):
Reviewed against current develop (df330d8). The daemon payload now uses the same os.path.abspath(os.path.expanduser(args.dir)) resolution _forward_mine_to_hub already applies, so a relative source (mine . --daemon) is bound to the caller cwd before it enters the long-lived daemon process. The chdir regression in tests/test_cli.py fails on develop and passes with this change. cmd_sync still has the same unresolved args.dir shape; leaving that as a follow-up is fine for #2441. Squash-merging.

Issue #2471, opened 2026-09-09
`detect()` / `_database_signature()` header sniff drops the writer's POSIX locks on the palace db; an external clean `close()` then checkpoints and unlinks the live `-wal`/`-shm` (silent write loss)


felixmurphybox-dot:
**Labels suggested:** bug, storage, area/mcp

## Environment

- macOS (Darwin 25.6), APFS
- Python 3.14.3 / SQLite 3.50.4 (python.org framework build), mempalace 3.9.0 installed via pip
- Reproduced on `develop` @ f9297a228586ebbfc5ac37763793d6c89ef69104 with uv env Python 3.12.13 / SQLite 3.53.3 (see note on SQLite version below)
- Backend: `sqlite_exact` (chroma has the same `detect()` code path)

## Summary

`SQLiteExactBackend.detect()` and `ChromaBackend.detect()` verify the palace by doing
`open(db_path, "rb").read(16)` on the database file. They run through
`palace.resolve_backend_name()` -> `registry.detect_backends_for_path()` on nearly every
MCP tool call, i.e. *after* the process already holds its long-lived WAL writer connection
on that same file. On `develop`, `SQLiteExactBackend._database_signature()` does the same
plain `open()` (100 bytes) on the read-only query path.

POSIX `fcntl` advisory locks belong to the (process, inode) pair, not to a descriptor.
Closing that plain descriptor drops every lock the process holds on the inode, including
the SHARED lock a WAL-mode connection keeps on the main database for its whole life.
SQLite works around this for its *own* descriptors (`unixInodeInfo` / `pUnused` in
os_unix.c) but cannot see descriptors opened outside it.

Once the SHARED lock is gone, any other process that opens the database read-write and
closes cleanly obtains EXCLUSIVE in `sqlite3WalClose()`, checkpoints, and (with SQLite
builds that delete the WAL on close, e.g. 3.50.4) unlinks `-wal` and `-shm`. The MemPalace
process keeps writing into the unlinked inodes, every write reports success, and no other
connection can see the rows.

## Root cause (file:line on `develop` @ f9297a2)

- `mempalace/backends/sqlite_exact.py:1880` — `SQLiteExactBackend.detect()`: `with open(db_path, "rb") as f: return f.read(16) == b"SQLite format 3\x00"`
- `mempalace/backends/chroma.py:2866` — `ChromaBackend.detect()`: same
[...]

Issue #2302, opened 2026-08-19
A freshly spawned `mempalace.mcp_server` truncates `chroma.sqlite3-shm` while the daemon holds a live connection → recurring FTS5/B-tree corruption and daemon SIGBUS


ebolamerican:
## Summary

Context: MemPalace is the shared memory for every agent on this machine (Claude Code sessions, a Hermes agent, and scheduled tasks all use one palace) — great tool, I rely on it daily. The same palace has corrupted 7 times in 6 weeks; what follows is the result of instrumenting the failure until it was caught in the act.

*Authorship note: this report was written by Claude Fable 5 (Anthropic's model, running in Claude Code) from its own investigation logs on my machine over the past six weeks; I reviewed it before filing but am not the author of the technical analysis and can't vouch for every detail at the code level. Related open issues that look like the same family: #1599 (recurring FTS5 corruption under live writes), #1092 (concurrent writers corrupt the palace on chromadb 1.5.8), #2240 / #2278 (FTS5 auto-heal). This one adds a syscall-level capture of the mechanism.*

On a palace with the daemon (`mempalace.daemon serve`) holding a long-lived connection, a **newly started `mempalace.mcp_server` process can win SQLite's WAL-index "first opener" check and truncate/recreate `chroma.sqlite3-shm` out from under the daemon's live mmap**. Depending on timing this produces either:

- **daemon SIGBUS** (`EXC_BAD_ACCESS`, `FS pagein error: 22`, faulting thread `sqlx-sqlite-worker-N`, faulting region = a 32 KB `-shm` mapping unit) — a hard kill that lands mid-write in ChromaDB's non-transactional FTS/HNSW flush, or
- **silent wal-index desync** — the daemon keeps checkpointing against a stale `-shm`, writing wrong pages; the next `PRAGMA quick_check` reports `malformed inverted index for FTS5 table main.embedding_fulltext_search`, and in worse cases page-level B-tree damage (`2nd reference to page`, hundreds of never-used pages).

We have hit resulting corruption **7 times between 2026-07-03 and 2026-08-17** on one machine, and have now captured one full occurrence end-to-end with `fs_usage`, an inode/size watcher, `lsof` censuses, and the client's own reconnect log. Every repair has been recoverable so far (FTS rebuild or `.recover`), but the recurrence rate is roughly weekly and each incident risks real data loss (one `.recover` cost 32 embedding rows that existed nowhere else).

[...]

Issue #2045, opened 2026-07-19
sqlite_exact backend: "database disk image is malformed" from co-writer race (MCP HTTP + daemon serve + mine, WAL sidecars unlinked under live writers)


C-LaForest:
> Identifiers scrubbed: private container/host/user names replaced with generic placeholders (`<container>`, `<host>`, `<batch-timer>`, `<agent>`). Instance-specific unit-name suffixes (e.g., `-<suffix>` in service unit names) are the local instance's naming — you can treat them as `mempalace-*.service` / `mempalace-*.timer` in your reading.

---

## Summary

On MemPalace **3.6.0** with `MEMPALACE_BACKEND=sqlite_exact`, `sqlite_exact.sqlite3` hit `sqlite3.DatabaseError: database disk image is malformed` **twice in 48h** (2026-07-17 and 2026-07-19). At the second event I caught the failure state live: **both long-lived writer processes hold `(deleted)` FDs to `-wal` and `-shm`**, which no longer exist on disk. The trigger correlates tightly with a long-running 4th writer (a weekly closet-regen sweep, 1130 upserts, ~87 min). This looks like the exact multi-writer race class that v3.5.0's serialized-writer daemon and v3.6.0's writer-lease recovery are supposed to eliminate — the docs state "MCP server refuses a second writer for the same palace." That contract does not appear to extend to the `daemon serve` path in this config.

## Environment

- MemPalace: `3.6.0` (`pip show mempalace` + `mempalace --version`).
- Python: 3.12.
- Backend: `sqlite_exact` (`MEMPALACE_BACKEND=sqlite_exact`).
- Deployment: rootful Podman quadlet on RHEL-family Linux (kernel 6.12), palace bind-mounted from host XFS. Volume mount pattern: `Volume=/srv/<data-dir>:/data:Z`. Container args `--memory=1g --cpus=1`.
- Palace layout inside container at `/data/palace/`: `sqlite_exact.sqlite3` (~134 MB) + `knowledge_graph.sqlite3` in the same directory.
- Writer processes inside the container:
  - PID 1 — `mempalace-mcp --transport http --host 0.0.0.0 --port 8765 --palace /data/palace`
  - PID 5 — `python3 -m mempalace.daemon serve --palace /data/palace`
- Host writer path: a systemd timer runs `podman exec <container> mempalace mine /projects --wing memory --agent <agent>` every 15 min.
[...]

Issue #1893, opened 2026-06-28
bug: ChromaBackend.detect() returns True for 0-byte chroma.sqlite3, causes BackendMismatchError on multi-artifact palaces


messelink:
## Summary

`ChromaBackend.detect()` returns `True` for a 0-byte `chroma.sqlite3` file. On palaces that have any non-chroma backend marker alongside an empty `chroma.sqlite3`, `resolve_backend_name` then raises `BackendMismatchError: palace at '<path>' contains multiple backend artifacts: chroma, <other>` and the palace becomes unopenable until the user manually `rm`s the 0-byte file.

The 0-byte file appears as a side effect of `sqlite3.connect()` on a missing path — Python's sqlite3 driver creates the file immediately but only writes the SQLite header on the first actual statement. So any code path that touches the chroma.sqlite3 path with bare `sqlite3.connect()` (including chromadb's own `PersistentClient` lazy-init — see the comment at `backends/chroma.py:2052`) can leave a 0-byte artifact behind. On a non-chroma palace this artifact then "looks like" a chroma palace to detection.

## Environment

- mempalace 3.5.0 (PyPI), pgvector backend
- Surfaced during a chroma → pgvector migration on a two-machine remote pgvector deployment — the empty file was created during the transition window where both backends co-existed, and persisted as the pgvector palace took over.

## Repro

```python
from pathlib import Path
from mempalace.backends.chroma import ChromaBackend

palace = Path("/tmp/repro-palace")
palace.mkdir()
(palace / "chroma.sqlite3").write_bytes(b"")  # 0-byte file
(palace / "pgvector_backend.json").write_text("{}")  # any other marker

from mempalace.backends.registry import detect_backends_for_path
print(detect_backends_for_path(str(palace)))
# → ['chroma', 'pgvector']

from mempalace.palace import resolve_backend_name
resolve_backend_name(str(palace))
# → BackendMismatchError: palace at '/tmp/repro-palace' contains multiple
#   backend artifacts: chroma, pgvector
```

## Root cause

`ChromaBackend.detect` at `mempalace/backends/chroma.py:407`:

```python
@classmethod
def detect(cls, path: str) -> bool:
    return os.path.isfile(os.path.join(path, "chroma.sqlite3"))
```

`os.path.isfile()` returns True for any regular file including 0-byte ones. A 0-byte file isn't a SQLite database — no header, no schema, no data. Treating it as "chroma palace detected" mis-routes the entire backend resolution.

[...]

PR #2472, opened 2026-09-09
fix(storage): stop header sniff from dropping the writer's POSIX locks

felixmurphybox-dot:
**Base:** `develop` · **Head:** `fix/sqlite-header-probe-drops-posix-locks` · **Closes:** #2471 · **Related:** #2302, #2045, #1893

## What does this PR do?

`SQLiteExactBackend.detect()`, `ChromaBackend.detect()` and
`SQLiteExactBackend._database_signature()` opened the palace database with a plain
`open(db_path, "rb")` to sniff the header. `detect()` runs on nearly every MCP tool call,
after the process already holds its long-lived WAL writer on the same file. POSIX fcntl
locks are per (process, inode): closing that plain descriptor drops every lock the process
holds on the file, including the SHARED lock the WAL writer keeps for its lifetime. From
then on any external connection's clean `close()` can take EXCLUSIVE in `sqlite3WalClose()`,
checkpoint, and (on SQLite builds that delete the WAL on close, e.g. 3.50.4) unlink
`-wal`/`-shm` underneath the writer, which keeps writing into the unlinked inodes with no
error.

Fix: all reads of the database file go through SQLite. New `mempalace/backends/_magic.py`:

- `read_header_fields()` reads page 1 through a throwaway `mode=ro&immutable=1`
  connection (no locks, no sidecars; SQLite's unix VFS parks the descriptor on close while
  a sibling connection holds locks on the inode).
- `has_sqlite_magic()` = `os.stat` + that probe, positive results cached per
  `(st_dev, st_ino)`, negatives never cached (#1893). Files SQLite refuses fall back to the
  old 16-byte prefix read, which is safe because no SQLite connection in this process can
  hold a file SQLite cannot open; keeps magic-only marker files detecting as before.
- Both `detect()`s reduce to one call. `_database_signature()` keeps the stat tuple and
  replaces the raw 100-byte header with `(schema_version, page_count, freelist_count,
  user_version, application_id)` from the probe — same purpose (page-1 changes a checkpoint
  can land without changing the size), no plain descriptor.

No new dependencies. No behaviour change for detection except that the probe no longer
touches lock state.

## How to test

```bash
uv run pytest tests/test_backend_header_probe_locks.py -v
uv run pytest tests/ -v
```

`tests/test_backend_header_probe_locks.py`:

- `test_probes_open_no_plain_descriptor_on_live_databases` — monkeypatches
  `builtins.open` / `io.open` / `os.open`; after a writer exists, `detect()` (both
  backends), `resolve_backend_name()` and `_database_signature()` open nothing on the db
  path. Fails against the pre-fix `detect()` (see Results for how that was checked).
- `test_probes_keep_writer_shared_lock` (POSIX) — subprocess tries a conflicting write
  lock on SQLite's SHARED range; the writer's lock survives three rounds of
  `detect()` + `resolve_backend_name()` + `_database_signature()` and is released by
  `backend.close()`. Fails against the pre-fix `detect()` (lock gone after the first
  probe). Independent of SQLite version — this is the build-independent guard.
- `test_external_close_cannot_orphan_wal_after_probes` (POSIX) — three tool-level
  operations, then an external `sqlite3.connect` + read + `close()` from a subprocess;
  asserts `-wal`/`-shm` inodes unchanged and the next write is visible from a fresh
  process. Fails against the pre-fix `detect()` when the attacking build deletes the WAL on close
  (SQLite 3.50.4 and Ubuntu's 3.45; not 3.53.3 — documented in the test).
- `test_detect_rejects_empty_then_accepts_written_header` — #1893 regression: 0-byte ->
  False, header written -> True, cache hit, replaced inode re-probed, magic-only marker
  still True, chroma analogue.
- `test_database_signature_tracks_changes`, `test_read_header_fields_reads_through_sqlite_only`.

Results on macOS (uv env, Python 3.12.13 / SQLite 3.53.3):

- unpatched develop @ f9297a2: full suite 4868 passed, 54 skipped, 0 failed (161 s).
  The new test module does NOT run there: it imports `mempalace.backends._magic` at
  module level, so pytest stops at collection (`ImportError: cannot import name '_magic'`,
  "1 error in 0.14s"); with that import stubbed out, all 6 tests error at fixture setup
  for the same reason. The "fails against the pre-fix `detect()`" statements above
  therefore describe the assertions run by hand against the old `detect()` body during
  development, not a run of this file on develop. The regression guard that does not
  depend on the SQLite build is `test_probes_keep_writer_shared_lock`: it checks the
  fcntl lock directly, so it catches any future probe that reintroduces a plain
  descriptor regardless of whether the attacking build deletes the WAL on close.
- patched: full suite 4874 passed, 54 skipped, 0 failed (4868 baseline + 6 new); new file 6/6 passed; `ruff check .` and
  `ruff format --check .` clean (ruff 0.16.1).

Not run: Windows (the two POSIX tests skip there; the unit test and #1893 test run).

## Checklist

- [x] No hardcoded paths
- [x] Conventional commit, targets `develop`

## Provenance

Report, patch and tests were produced with Claude Fable 5.1 (Claude Code) under Felix
Murphy's direction and reviewed by him before submission. The bug was first hit in
production on mempalace 3.9.0 and reproduced there before being ported to `develop`.


mempalace/backends/_magic.py
```diff
@@ -0,0 +1,150 @@
...
+    try:
+        conn = sqlite3.connect(uri, uri=True)
+    except sqlite3.Error:
+        return None
+    try:
+        return tuple(int(conn.execute(f"PRAGMA {name}").fetchone()[0]) for name in _HEADER_PRAGMAS)
+    except sqlite3.Error:
+        return None
+    finally:
+        conn.close()
```
igorls, line 98: [P1] Preserve the Rust Chroma writer's lock when closing the probe

The descriptor parking described here belongs to Python sqlite3's SQLite instance. It does not protect locks held by Chroma's separately linked Rust SQLite instance. On a live WAL-mode Chroma palace, this `conn.close()` still drops the process's SHARED lock: reproduced with Chroma 1.5.7 using a subprocess lock probe, held before `ChromaBackend.detect` and absent afterward. Please use a probing mechanism that cannot close a foreign descriptor in the writer process and add the real-Chroma regression; the current sqlite_exact-only lock test misses this path.

igorls (changes requested):
Requesting changes because the Chroma path still loses the writer's POSIX lock, although the supplied tests pass.

Reproduced on Ubuntu/WSL with Python 3.12.3, Python SQLite 3.45.1 and Chroma 1.5.7: create a WAL-mode database before opening the persistent Chroma client, add a drawer with explicit embeddings, clear the probe cache, and call `ChromaBackend.detect`. A separate process attempts an exclusive lock on SQLite's SHARED range (`0x40000002`, length 510): it is blocked before detection and succeeds afterward. Detection itself returns true. This establishes lost lock protection; I did not measure lost production writes.

All six supplied POSIX regression tests pass, but they use a live Python sqlite_exact writer and only a minimal database fixture for Chroma. Please cover a live WAL-mode Chroma writer and preserve its lock across detection before merging.

igorls (approved):
Rebased onto latest develop, resolved the Chroma POSIX lock dropping issue by retaining the read-only probe connection on POSIX, added 	est_probes_keep_chroma_writer_shared_lock covering the exact reproduction, and verified all tests pass across Ubuntu/WSL and Windows.

PR #2474, opened 2026-09-09
perf(qdrant): walk the cursor once for list_drawers and graph_stats

### LINKED ISSUE #2452: qdrant: list_drawers and graph_stats are still O(n^2) after #2384 (caller-side offset loops were never bounded)
## Summary

On a qdrant-backed palace of ~126,000 drawers, `list_drawers` takes **274.7 s** and `graph_stats` takes **292.3 s** on 3.9.0.

#2384 (which closed #2363) is a real, measurable improvement: bounding `QdrantCollection.get()`'s scroll to `offset+limit` roughly halved the cost of `list_drawers` on this palace. But it bounded the *callee*, not the callers. `mcp_server._fetch_drawer_rows` and `palace_graph.build_graph` still walk the entire matching set as `col.get(limit=1000, offset=k*1000)`, and on qdrant each of those pages re-scrolls from row 0. The total work is still quadratic in collection size: #2384 cut the constant in half, not the complexity.

The PR says as much itself:

> It does not change `list_drawers`'s own `_fetch_drawer_rows` loop in `mcp_server.py`, which still requests pages up to exhaustion of the matching set. [...] bounding it further looked like a separate, larger change

This issue is that separate change, plus a second victim of the same chroma-only fast-path fork that #2363 did not cover: **`graph_stats`**, which unlike `list_drawers` is exposed on the MCP tool surface, where 292 s is not a slow command but a client-side timeout. In fairness: neither call is on our own hot path (our workload is semantic search plus targeted KG lookups, both sub-second), so both numbers come from targeted benchmarking rather than from a production incident. We are reporting the shape of the cost, not a broken deployment.

## Environment

- `mempalace` **3.9.0** (pip, current release), plus **3.7.1** for the before/after comparison
- Backend: **qdrant** (local), Python 3.14
- Palace: **~126,000 logical drawers**, **128,297 physical chunk rows**; largest single wing 122,893 rows
- Line numbers below are from the installed 3.9.0 wheel

## Measurement 1: cost of a single page

The unit of work both loops are built from: one `col.get(include=["metadatas"], limit=1000, offset=K)` against the same live palace, same script, identical arguments on both versions.

| offset | 3.7.1 | 3.9.0 |
|---------|--------|--------|
| 0 | 4.54 s | 0.04 s |
| 30,000 | 3.98 s | 1.22 s |
| 60,000 | 4.25 s | 1.85 s |
| 90,000 | 4.08 s | 2.83 s |
| 120,000 | 4.20 s | 3.77 s |

The shape is the whole story:

- **3.7.1 is flat** at ~4.1 s: every page re-scrolls all 128k rows, regardless of which page was asked for.
- **3.9.0 is linear in `offset`**: page *k* now scrolls `(k+1)*1000` rows, exactly as #2384 intended.

A page is still never O(page_size). Summed over a full walk, linear-per-page is still quadratic overall.

The same run with a pushdown-able `where` on the largest wing gives the same picture (3.7.1: 3.75 to 4.27 s flat; 3.9.0: 0.03 s rising to 3.90 s). A small wing (199 drawers, one page, `offset` never leaves 0) is 0.01 s on both versions, which is the only reason wing-filtered listing feels fine.

Implied scroll throughput is ~32,000 rows/s and is consistent across both versions and all offsets, which is what makes the row counts below comparable.

## Measurement 2: the full call

| call | 3.7.1 | 3.9.0 |
|------|-------|-------|
| `list_drawers(limit=20)` | approx. 530 s (derived) | **274.74 s** (measured) |
| `list_drawers(limit=20)` on the largest wing, 122,893 rows | approx. 490 s (derived) | **259.56 s** (measured) |
| `list_drawers(limit=20)` on a small wing, 199 drawers | 0.02 s | 0.02 s |

`limit=20` changes nothing: the page is sliced in Python at the very end.

Rows actually scrolled to produce 20 drawers, where 128,297 would suffice:

- 3.7.1: 129 pages x 128,297 = **16.6M rows**
- 3.9.0: 1000 x (1+2+...+129) = **8.4M rows**

Sanity check on the 3.9.0 model: 8.385M rows at ~32,000 rows/s predicts ~262 s against 274.74 s measured, and the remainder is multi-chunk collapse plus preview hydration for the displayed page. The model matches, so the quadratic term is unambiguously the cost driver.

## Where the time goes

`tool_list_drawers` (`mcp_server.py`):

- `:4076`: the fast path is gated on `_is_chroma_backend()`, so `sqlite_list_id_metadata` is chroma-only.
- `:4090`: every other backend, qdrant included, falls into `_fetch_drawer_rows(col, where=where, include=["metadatas"])`.
- `:3190`: `_fetch_drawer_rows` loops `col.get(include=..., limit=1000, offset=offset)` until a short page returns, accumulating the whole matching set.
- `:4094`: `since`/`before` are still applied client-side after the full fetch, so they never narrow the scan.
- `:4100`: only now is the requested page sliced out, `drawers[offset:offset+limit]`.

`backends/qdrant.py`:

- `:1036`: `get()` has no server-side integer offset; it materializes rows via `_rows()` / `_scroll_all()` and slices in Python.
- `:1047`: the #2384 bound, `max_rows = (offset or 0) + limit ...`. This is what turned the flat 4.1 s into a linear ramp. It is correct and it helps; it just cannot help a caller that asks for every page.

`graph_stats`, same fork, different function:

- `mcp_server.py:2932`: `tool_graph_stats` tries `_sqlite_graph_stats()` first.
- `mcp_server.py:2276`: that fast path returns `None` for qdrant (it is chroma / sqlite_exact only), and its own docstring already predicts the outcome: "the client path builds the whole graph by paging every metadata row (`build_graph` -> `col.get(limit, offset)`) ... which times out on six-figure palaces" (#1379).
- `palace_graph.py:239`: the fallback loop, `batch = col.get(limit=1000, offset=offset, include=["metadatas"])`, same quadratic walk.

Measured 292.29 s on this palace. This is the documented six-figure timeout case, reached rather than merely predicted.

Neighbouring tools on the same palace and version are fine (`memories_filed_away` 0.00 s, `list_rooms` 0.04 s, `list_wings` 0.08 s, `status` 0.10 s, `list_hallways` 2.80 s), so the problem is specific to the code paths that offset-paginate, not to the backend or the palace size.

## Suggested fix

The primitive is already in the codebase, added for #1796 in the same file: `QdrantCollection.get_all_metadata` (`backends/qdrant.py:1069`) walks the native cursor exactly once, and its docstring names this exact trap. `BaseCollection.get_all_metadata` (`backends/base.py:474`) keeps the paginated default for backends with a real server-side cursor, and even spells out the contract: backends that Python-slice a materialized set "MUST override this method ... Calling the default implementation on such a backend is O(n^2)". Two call sites are still calling the equivalent of that default by hand.

**1. `graph_stats` is a drop-in.** `build_graph`'s loop consumes only `batch["metadatas"]` and uses `batch["ids"]` purely as a loop guard, so `col.get_all_metadata()` replaces the whole `while offset < total` block with one cursor pass. No new seam, no behaviour change, and every backend keeps working through the base default. This is the one worth doing first: it is the cheaper of the two, and it is the one exposed as an MCP tool, so on a six-figure palace it is unusable rather than merely slow.

**2. `list_drawers` needs ids, so `get_all_metadata` is not directly reusable.** `_collapse_drawer_rows` needs the row ids to build `drawer_id` and `chunk_ids`. Note that the ids are already in hand and then dropped:

```python
# backends/qdrant.py:1087
rows = self._rows(where=where)
return [row["metadata"] for row in rows]   # row["id"] discarded
```

So the minimal version is an ids-plus-metadata sibling of `get_all_metadata` on the same seam (qdrant overriding it with the same single `_rows()` pass, base class keeping the paginated default), with `_fetch_drawer_rows` routed through it. The multi-chunk collapse and the client-side `since`/`before` filter then run over the same complete set they run over today, so the correctness constraints #2384 cited are untouched and the set is simply gathered in one pass instead of 129 overlapping ones. Expected effect on this palace: 274 s down to a few seconds. The more general form of this is exactly the `scan()` primitive proposed in #1797.

Wider point, for whoever picks this up: there are seven `_is_chroma_backend()` forks in `mcp_server.py` (`:1104`, `:1580`, `:1645`, `:2240`, `:2295`, `:4076`, `:5095`) where a sqlite fast path exists only for chroma and everything else falls back to a generic `col.get(limit, offset)` walk. We measured two of them and both were pathological on qdrant. The remaining five are unexamined.

## Methodology and caveats

Stated plainly, since one column above is derived rather than measured:

- The per-page table is **direct measurement**: one call per cell, same script on both versions, identical `include`/`limit`/`offset` arguments against the same live palace and the same Qdrant instance. The 3.7.1 vs 3.9.0 difference in that table is attributable to the `max_rows` bound alone, as no other variable changed between the runs.
- 3.9.0 full-call numbers are **wall-clock**.
- 3.7.1 full-call numbers are **derived** (page count x measured flat per-page cost), not wall-clock, because a full 3.7.1 run did not fit the time budget. They are marked as derived in the table for that reason.
- 3.7.1 was loaded from a package snapshot via `PYTHONPATH` against the same palace and Qdrant instance rather than installed into a separate environment. `QdrantCollection.get()` in that snapshot differs from 3.9.0 by exactly the `max_rows` line.
- The palace is live and grew ~2% between the two version runs. That affects neither the ~2x gap nor the flat-vs-linear shape.
- Single sample per cell, no repeats. Everything read-only, no `repair` / `rebuild` / `sync`.
- `graph_stats` was measured on 3.9.0 only, so whether its 292 s changed between versions is unknown.

## Related

- #2363: same root cause for `list_drawers`, closed by #2384. Filing new rather than asking for a reopen, since #2384 explicitly scoped the caller-side loop out and this adds `graph_stats`.
- #2384: the qdrant `get()` bound. Helped, measurably, but is not the whole fix.
- #1796: established `get_all_metadata` and the single-cursor-pass rule that both call sites here violate.
- #1797: the systemic version of this, a `scan()` seam primitive. Field measurements from this report posted there as well.
- #1379: predicted the `graph_stats` timeout on six-figure palaces in the `_sqlite_graph_stats` docstring.

L4XB:
## Summary

Fixes #2452

`list_drawers` and `graph_stats` still walked the whole matching set with `col.get(limit=1000, offset=offset)` loops (`mcp_server._fetch_drawer_rows` and `palace_graph.build_graph`). On qdrant `get()` has no server-side offset: it materialises `offset + limit` rows from a fresh scroll and slices in Python, so #2384 turned a flat per-page cost into a linear one, but a caller that asks for every page still pays a quadratic total — 275 s for `list_drawers(limit=20)` and 292 s for `graph_stats` on a 126k-drawer palace, i.e. an MCP client timeout.

## Change

- `BaseCollection.get_all_rows(where=None, include=None)`: the ids-carrying sibling of `get_all_metadata()` (#1796). Returns a `GetResult` with `ids` plus the `include`d fields (`["metadatas"]` by default) in one logical pass. The default implementation pages through `get()` exactly like the old loop, so Chroma / pgvector / sqlite backends are unchanged.
- `QdrantCollection.get_all_rows()`: one `_rows()` pass — the same single scroll `get_all_metadata()` uses, keeping the ids it used to discard.
- `mcp_server._fetch_drawer_rows()`: backend collections (`BaseCollection` instances) go through `get_all_rows()`; the result is padded exactly like the loop did. Plain collection objects keep the offset loop.
- `palace_graph.build_graph()`: reads metadata through `get_all_metadata()` for backend collections (`_iter_all_metadata`), falling back to the offset loop otherwise. The loop body is unchanged.

Both call sites use `isinstance(col, BaseCollection)` rather than `getattr` probing so that `MagicMock` collections in the existing tests keep hitting the legacy path.

## Tests

- `tests/test_qdrant_bulk_metadata_scroll.py`: default `get_all_rows()` paging (ids aligned with metadata, documents only when included, `where` passthrough); qdrant `get_all_rows()` walks the cursor once (2 scroll calls for 2 pages), never calls `get()`, honours `include`, and applies local `$or` filters.
- `tests/test_mcp_server.py`: `_fetch_drawer_rows` delegates to `get_all_rows()` on a backend collection (and never pages `get()`), and keeps the offset loop for a plain object.
- `tests/test_palace_graph.py`: `build_graph` reads a backend collection through `get_all_metadata()` without touching `get()`.

```
pytest tests/test_palace_graph.py tests/test_palace_graph_tunnels.py tests/test_qdrant_bulk_metadata_scroll.py tests/test_qdrant_backend.py tests/test_mcp_server.py tests/test_backends.py tests/test_backend_conformance.py tests/test_pgvector_backend.py tests/test_cli.py
786 passed, 2 skipped
ruff check / ruff format --check: clean
```

The other `_is_chroma_backend()` forks the issue lists (taxonomy, hallways, …) are left for follow-ups; this covers the two measured ones.


igorls:

L4XB:
Rebased onto `develop` and ported to the new layout: `_fetch_drawer_rows` moved into `mempalace/mcp_server/tools_write.py` with the package split, so the delegation lives there now, and the two tests moved to `tests/mcp/test_read_tools.py`, next to the existing `_fetch_drawer_rows` tests, after `test_mcp_server.py` was split up. The `backends/base.py`, `backends/qdrant.py` and `palace_graph.py` parts are unchanged.

`tests/mcp/test_read_tools.py`, `tests/test_palace_graph.py` and `tests/test_qdrant_bulk_metadata_scroll.py` are at 108 passed, `ruff check` and `ruff format --check` clean.

L4XB:
CI caught a real gap and it is fixed at `b91ff6e`: `test_wrapper_forwards_all_concrete_basecollection_methods` requires `EmbeddingCollection` to forward every concrete `BaseCollection` method, and `get_all_rows` was new. Without the forwarder MRO resolves the call on the wrapper and runs the base default, which pages through `get()` again, so a wrapped backend would have kept exactly the offset walk this PR removes. Same shape as the `get_all_metadata` and `facet_counts` forwarders next to it.

`tests/test_embedding_wrapper.py`, `tests/mcp/test_read_tools.py`, `tests/test_palace_graph.py` and `tests/test_qdrant_bulk_metadata_scroll.py` are at 118 passed locally.

PR #2476, opened 2026-09-09
fix(daemon): log lifecycle events in foreground mode

### LINKED ISSUE #2475: Foreground daemon emits no lifecycle logging (empty launchd StandardOutPath)
# Foreground daemon emits no lifecycle logging (empty launchd StandardOutPath)

## Environment
- MemPalace 3.8.0 (`mempalace --version` -> `MemPalace 3.8.0`; package metadata
  and `mempalace.version.__version__` agree)
- STILL PRESENT ON LATEST: v3.9.0 (2026-08-31) is the current release, and
  `mempalace/daemon.py` is byte-identical between v3.8.0 and v3.9.0 — the
  `log_message` no-op and the `serve_forever` path are unchanged there.
- Platform: macOS 26.6.2 (Sequoia-class), python3.9 (uv tool install)
- Launch form: launchd agent `ai.mempalace.daemon`
  (`/Users/hermes/Library/LaunchAgents/ai.mempalace.daemon.plist`) running
  `/Users/hermes/.local/bin/mempalace daemon start --foreground`, with
  `StandardOutPath=/Users/hermes/.mempalace/logs/daemon.log` and
  `StandardErrorPath=/Users/hermes/.mempalace/logs/daemon.error.log`.

## Observed
- `StandardOutPath` stays empty indefinitely while the daemon runs:
  - `/Users/hermes/.mempalace/logs/daemon.log` -> 0 bytes (mtime 2026-08-11)
  - `/Users/hermes/.mempalace/logs/daemon.error.log` -> 1848 bytes, 8 lines, all
    of them prior *startup errors* (e.g. "writable daemon startup refused:
    another writer owns local backend 'chroma'"), none of them lifecycle output
    from a healthy run. So stderr proves the redirect works; the healthy
    foreground path simply never writes to it.
- `mempalace daemon status` reports the daemon is running (PID alive, endpoint reachable),
  so the process is healthy — it simply emits no output.

## Expected
- In `--foreground` mode the server should log at least startup and lifecycle
  events to stderr (or stdout), so a supervisor that captures via
  `StandardOutPath`/`StandardErrorPath` actually receives log lines.

## Cause
- `run_server` (`mempalace/daemon.py:872`) emits no intentional startup or HTTP
  access logging in
  foreground mode. Its only output hook is the HTTP handler's
  `log_message`, which is a deliberate no-op:
  ```python
  # mempalace/daemon.py:920-921
  def log_message(self, fmt, *args):  # pragma: no cover - stdlib access logging noise
      return
  ```
  No other code path in `run_server` prints, so with stdout redirected to the
  launchd log path that file stays empty for the life of a healthy run.
- Contrast: non-foreground mode spawns the child with stdout/stderr bound to a
  per-daemon file via `_detached_kwargs` (`mempalace/daemon.py:1257-1260`) and
  `kwargs = _detached_kwargs(sd / "daemon.log")` (`mempalace/daemon.py:1332`).
  That path works; only foreground is silent.
- The daemon does NOT import `mcp_server`, so the MCP stdio guard there is
  unrelated and is not the cause.

## Proposed fix
Make the foreground path emit timestamped, level-gated lines to stderr:
```diff
--- a/mempalace/daemon.py
+++ b/mempalace/daemon.py
@@ run_server (foreground), after `runtime = DaemonRuntime(...)`:
+        import sys, logging
+        logging.basicConfig(stream=sys.stderr, level=logging.INFO,
+                            format="%(asctime)s %(levelname)s %(message)s")
+        log = logging.getLogger("mempalace.daemon")
+        log.info("daemon listening on %s", runtime.endpoint)
```
Optionally replace the `_Handler.log_message` no-op with
`logging.getLogger("mempalace.daemon.http").info(fmt, *args)` so HTTP access
logging also flows to stderr when level permits. Keep non-foreground behavior
unchanged (it already has its own log file).

## Workaround
- Liveness/health is observable via `mempalace daemon status` (no log needed).
- For actual on-disk logs, run without `--foreground`: the non-foreground path
  writes to `/Users/hermes/.mempalace/daemon/<sha>/daemon.log`.

L4XB:
Fixes #2475

## Problem

`mempalace daemon start --foreground` under launchd (or any supervisor that captures the process's stdio) leaves `StandardOutPath` empty for the life of a healthy run. `run_server` emits no startup or lifecycle output, and `_Handler.log_message` is a deliberate no-op, so the only lines that ever reach the redirected files are startup errors. The background path is fine because its detached child has its stdio bound to the per-daemon `daemon.log`.

## Change

- New module logger `mempalace.daemon`. `run_server` logs `daemon listening on http://<host>:<port> (pid ..., backend ..., palace ...)` once the server is bound and the worker started, and `daemon stopping` / `daemon stopped` around the drain in `finally`.
- `_Handler.log_message` routes the stdlib access lines to that logger at DEBUG instead of dropping them, so they stay quiet by default but can be enabled.
- `configure_foreground_logging()` attaches a timestamped stderr handler (INFO) to the logger and is called from `start_daemon(..., foreground=True)` only, so the detached child's log file and library users of `run_server` are not affected. Idempotent.

## Tests

`tests/test_daemon.py`:
- `test_run_server_logs_lifecycle_events`: through the existing `_start_server` helper, the listening line (with palace path and pid) is logged at startup and the stopping/stopped lines on shutdown.
- `test_configure_foreground_logging_is_idempotent`: one stderr handler, INFO level, no duplicates on repeat calls.

`pytest tests/test_daemon.py` passes; `ruff check` / `ruff format` on both files.


L4XB:
The red `test-windows` here is `tests/test_cli_logstream.py::TestLogstreamWatch::test_idle_deadline_caps_the_poll`, failing with `idle 200ms waited 3.71s (poll was 5s)`. This branch touches only `mempalace/daemon.py` and `tests/test_daemon.py`, and that test measures wall-clock latency of the logstream poll, so it looks like the Windows runner being slow rather than anything here. Happy to rebase if you would rather see a clean run.

L4XB:
The red `test-windows` job here is not this change. It failed on `tests/test_cli_logstream.py::TestLogstreamWatch::test_idle_deadline_caps_the_poll`, which measures wall-clock time and requires a 200ms idle deadline to finish inside 1.5s:

```
AssertionError: idle 200ms waited 3.71s (poll was 5s)
assert 3.7050543999999945 < 1.5
```

That file and the watch path are untouched here (this PR changes `mempalace/daemon.py` lifecycle logging plus `tests/test_daemon.py`), the same case passes on my other two open PRs' Windows jobs, and locally it runs in 0.3s on this branch across repeated runs. The remaining 4568 tests passed and coverage held at 81.4%.

I rewrote that assertion in #2491 so it records the timeouts the watcher hands to the long poll instead of timing the machine. Once that lands, or after a re-run of this job, this PR should go green without a change here.

L4XB:
Rebased onto `develop` (`99b1c97`). The Windows red was `tests/test_cli_logstream.py::TestLogstreamWatch::test_idle_deadline_caps_the_poll` — `idle 200ms waited 3.71s` — an unrelated wall-clock assertion on a slow runner, since fixed by #2491 (`a30c451`). Picking that up is all the rebase does; this PR's own files are unchanged. `tests/test_daemon.py` + `tests/test_cli_logstream.py`: 125 passed locally.

Issue #2388, opened 2026-08-29
_wing_from_transcript_path fallback does not collapse --claude-worktrees- — one wing per worktree when cwd is absent


trek-e:
**Observed on:** 3.8.0 (macOS, Claude Code plugin). Fallback code unchanged at current `develop`.

`_wing_from_transcript_path` has two ways to derive a project wing, and only one of them collapses git worktrees.

**Primary path (correct).** `_wing_from_jsonl_cwd` reads `cwd` from the JSONL and explicitly strips the worktree segment — `hooks_cli.py`:

```python
# A cwd inside "<project>/.claude/worktrees/<wt>" (a git
# worktree) belongs to <project>, not the ephemeral worktree
# directory -- otherwise every worktree spawns its own wing.
_wt_marker = "/.claude/worktrees/"
if _wt_marker in cwd_norm:
    cwd_norm = cwd_norm.split(_wt_marker, 1)[0]
```

**Fallback path (bug).** The branch that decodes the `.claude/projects/-encoded-dir` name has no equivalent collapse. Claude Code flattens `/Users/<user>/projects/gsd-core/.claude/worktrees/hardcore-wilbur-48691a` to `-Users-<user>-projects-gsd-core--claude-worktrees-hardcore-wilbur-48691a`, and the `--claude-worktrees-` segment survives into the wing name.

**Repro** (3.8.0, no transcript needed — the fallback is reached whenever `cwd` is absent):

```python
from mempalace.hooks_cli import _wing_from_transcript_path as f
f("/Users/u/.claude/projects/-Users-u-projects-gsd-core--claude-worktrees-hardcore-wilbur-48691a/x.jsonl")
```

| | result |
|---|---|
| actual | `wing_gsd_core__claude_worktrees_hardcore_wilbur_48691a` |
| expected | `wing_gsd_core` |

Non-worktree transcripts under the same fallback resolve correctly (`-Users-u-projects-gsd-core/x.jsonl` -> `wing_gsd_core`), and a real worktree transcript that *does* carry `cwd` also resolves correctly (verified: a `chronophon` worktree session -> `wing_chronophon`). So the defect is isolated to the fallback branch.

[...]

PR #2478, opened 2026-09-10
feat(hooks): add generic harness adapters for Grok and Copilot

vijayyadav1002:
## What does this PR do?

`mempalace hook run` currently treats `--harness` as a closed `{claude-code, codex}` enum, and `_parse_harness_input` only reads Claude's snake_case keys (`session_id`, `transcript_path`). Plugin wrappers also hardcode `--harness claude-code`. That means the same save policy cannot be reused by other coding-agent harnesses: Grok sessions live under `~/.grok/sessions/<urlencoded-cwd>/<id>/chat_history.jsonl`, and Copilot CLI sessions live under `~/.copilot/session-state/<id>/events.jsonl`. Pointing the existing hook at those transcripts counts 0 user turns and never saves.

This PR keeps **one shared save policy** (every 15 user turns, silent diary checkpoint, SessionEnd flush, PreCompact mine) and makes the harness-specific part a small adapter:

- payload aliases (`sessionId` / `session_id`, `transcriptPath` / `transcript_path`, `cwd` / `workspaceRoot`, Copilot `stopReason`)
- transcript location when the payload omits a path
- user-turn counting for that transcript format

`--harness` now accepts `claude-code`, `codex`, `grok`, `copilot`, `auto`, or any other token. Unknown tokens parse with the generic aliases instead of `sys.exit(1)`, so a new harness can start from `mempalace hook run --harness <name>` without a core change.

`--harness auto` detects:

| Signal | Harness |
|---|---|
| `GROK_SESSION_ID` / `GROK_HOOK_EVENT` / `GROK_WORKSPACE_ROOT` | grok |
| `CURSOR_TRANSCRIPT_PATH` or `conversation_id` + `workspace_roots` | cursor |
| `COPILOT_HOME`, `stopReason`, or a `session-state` / `events.jsonl` path | copilot |
| otherwise | claude-code |

Grok real turns are records with `type=user` **and** `prompt_index` (skipping `user_info` and synthetics). Copilot real turns are `type=user.message` with `data.content`. Stop fires with a `subagentType` / `agentType`, or a reason other than `end_turn`, are skipped so teardown/subagent stops do not double-save.

Plugin wrappers default to `--harness "${MEMPALACE_HOOK_HARNESS:-auto}"`. A Copilot CLI example lives at `examples/copilot/hooks.json`.

Claude Code and Codex paths are unchanged: existing snake_case payloads, transcript_path, and diary agent mapping (`claude-code` → `claude`) still hold.

## How to test

```bash
uv sync --extra dev
uv run ruff check mempalace/hooks_cli.py mempalace/cli.py tests/test_hooks_harness_adapters.py
uv run pytest tests/test_hooks_harness_adapters.py tests/test_hooks_cli.py tests/test_hooks_shell.py tests/test_claude_plugin_hook_wrappers.py tests/test_hooks_bash_compat.py -q
```

Smoke the CLI:

```bash
mempalace hook run --help
# --harness is a free token; grok/copilot/auto are documented values

printf '%s' '{"sessionId":"s","cwd":"/tmp/proj","reason":"end_turn"}' \
  | mempalace hook run --hook stop --harness grok
# below-interval Stop → {}

printf '%s' '{"sessionId":"s","cwd":"/tmp/proj","stopReason":"end_turn"}' \
  | mempalace hook run --hook stop --harness copilot
# below-interval Stop → {}
```

To wire Copilot CLI, copy `examples/copilot/hooks.json` to `~/.copilot/hooks/mempalace.json`.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — scoped hook/CLI suite: 323 passed, 1 skipped
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`) — ruff check/format clean on the touched Python files


.claude-plugin/hooks/mempal-stop-hook.sh
```diff
@@ -21,4 +21,4 @@ run_mempalace_hook() {
   return 1
 }
 
-run_mempalace_hook --hook stop --harness claude-code
+run_mempalace_hook --hook stop --harness "${MEMPALACE_HOOK_HARNESS:-auto}"
```
igorls, line 24: **[bug]** Defaulting the Claude plugin Stop wrapper to `auto` is an unstated behavior change for Claude Code, and it is also the only Grok install path in this PR.

Under Grok, `GROK_SESSION_ID` / `GROK_HOOK_EVENT` make `auto` resolve to `grok` and locate `chat_history.jsonl`. Under Claude Code with no Grok env, this still falls through to `claude-code`. Keep this wrapper on `claude-code` unless you document Grok-via-Claude-plugin as the supported install, and ship a Grok-native `examples/grok/` hook file that passes `--harness grok` with an explicit timeout.

mempalace/hooks_cli.py
```diff
@@ -1144,21 +1213,208 @@ def _diary_agent_for_harness(harness: str) -> str:
...
+            sessions_root=sessions_root,
+        )
+        return str(found) if found else ""
+    if resolved == "copilot":
+        found = _find_copilot_events(
+            str(parsed.get("session_id") or ""),
+            sessions_root=sessions_root,
+        )
+        return str(found) if found else ""
+    return explicit
```
igorls, line 1407: **[bug]** Non-grok/copilot harnesses return the empty explicit path and never search `~/.grok/sessions`.

`hooks/mempal_save_hook.sh` (unchanged) still calls `--harness claude-code`. A Grok user who installs that script gets `transcript_path=""`, this `return explicit`, turn count 0, and no saves. Either teach this locator to honor `GROK_SESSION_ID` even when `--harness` is `claude-code`, or update the bash save hook / README so Grok cannot hit this path.

mempalace/hooks_cli.py
```diff
@@ -1310,7 +1566,11 @@ def hook_stop(data: dict, harness: str):
     parsed = _parse_harness_input(data, harness)
     session_id = parsed["session_id"]
     stop_hook_active = parsed["stop_hook_active"]
-    transcript_path = parsed["transcript_path"]
+    harness = parsed.get("harness") or harness
+    if _should_skip_stop(parsed):
```
igorls, line 1570: **[suggestion]** Skipping any Stop whose `reason` is set and not `end_turn` is correct for Grok (session-end Stop fires with `channel_closed` / `shutdown`).

This now applies to every harness, including Claude Code. Claude Stop payloads usually omit `reason`, so this is likely a no-op there — please add a Claude regression that a snake_case Stop with no `reason` still saves at interval, so a future Claude `reason` field cannot silently disable auto-save.

igorls (changes requested):
Requesting changes: the Grok adapter is the right shape, but Grok users still cannot install this as-is.

I checked the Grok side against a live Windows session (`chat_history.jsonl` under `~/.grok/sessions/<urlencoded-cwd>/<id>/`). `_is_grok_user_turn` matches that file: real prompts carry `prompt_index`, `user_info` does not, synthetics have `synthetic_reason`. `quote(cwd, safe="")` matches the Windows session directory. Auto-detect from `GROK_SESSION_ID` / `GROK_HOOK_EVENT` / `GROK_WORKSPACE_ROOT`, camelCase aliases, skipping `subagentType`, and skipping Stop `reason != end_turn` all match Grok's hook docs.

Please address the following before merge:

1. **Grok has no install path.** Copilot got `examples/copilot/hooks.json`. Grok got nothing. `hooks/mempal_save_hook.sh` still hardcodes `--harness claude-code`, so a bash-script install never locates `chat_history.jsonl` (empty `transcript_path` returns as-is). Add a Grok-native example under `examples/grok/` using Grok events (`Stop`, `SessionEnd`, `PreCompact`) with explicit timeouts. Grok `SessionEnd` defaults to 5s and will kill a mine.

2. **Claude plugin wrappers now default to `auto`.** That is the only path that would work if Grok loads the Claude plugin (`GROK_*` env present). It is also an unstated behavior change on every Claude Stop/PreCompact/SessionEnd wrapper. Either document that as the Grok install path, or keep Claude wrappers on `claude-code` and ship a Grok example that passes `--harness grok`.

3. **Silent-save output is Claude's `systemMessage`.** Grok Stop only documents `decision` / `reason` / `continue` / `stopReason` / `additionalContext`. The diary write still happens; the confirmation likely never shows in the TUI. Map the success notice onto Grok's vocabulary (or emit no JSON and exit 0).

4. **CI has not run on this head.** Do not merge on the author's local 323-pass note.

The locator, turn counting, and teardown skip can stay. They are the correct Grok contract.

igorls:
Re-checked the new head (`73c616a`, force-pushed after my review). None of the four items from the change request are addressed yet:

1. **Grok install path** — still no `examples/grok/`; the PR ships `examples/copilot/hooks.json` only.
2. **Plugin wrappers default to `auto`** — `.claude-plugin/hooks/mempal-{stop,precompact,session-end}-hook.sh` still use `--harness "${MEMPALACE_HOOK_HARNESS:-auto}"`, and the CHANGELOG entry documents that as the new default for every Claude wrapper.
3. **`systemMessage` on Grok Stop** — the silent-save output still emits `systemMessage` (tests still assert on it); nothing maps it onto Grok's `additionalContext` / `reason` vocabulary.
4. **CI** — this head is `mergeStateStatus=BLOCKED`; CI has not completed on it.

The review stays at changes-requested until those land. The locator, turn counting and teardown skip are fine as they are — no need to touch them.

mempalace/hooks_cli.py
vijayyadav1002: Went with “document the Grok path,” not “search ~/.grok/sessions under claude-code.” Locator stays harness-gated. Grok users copy examples/grok/hooks.json; the bash save hook is Claude/Codex only.

mempalace/hooks_cli.py
vijayyadav1002: Added the Claude no-reason interval save test. Also gated the reason skip to grok/copilot so a future Claude reason field cannot turn auto-save off. subagent_type skip is still for every harness.

mempalace/hooks_cli.py
```diff
@@ -1144,21 +1231,326 @@ def _diary_agent_for_harness(harness: str) -> str:
...
 
+def _wing_from_cwd(cwd: str) -> str:
+    """Derive ``wing_<slug>`` from a workspace path leaf."""
+    name = Path(str(cwd).rstrip("/\\")).name if cwd else "sessions"
+    return f"wing_{_safe_wing_slug(name)}"
+
+
+def _project_wing(parsed: dict, transcript_path: str) -> str:
+    cwd = str(parsed.get("cwd") or "")
+    if cwd:
```
igorls, line 1358: **[bug]** `_project_wing` prefers payload `cwd` via `_wing_from_cwd`, which uses only the path leaf and does not collapse `<project>/.claude/worktrees/<wt>`. Claude Code Stop/SessionEnd stdin includes `cwd`, so this is now the default wing path for Claude, not just Grok/Copilot. `_wing_from_jsonl_cwd` and the encoded-folder fallback (fixed in #2388) map that cwd to the parent project; this change files worktree checkpoints under `wing_<worktree>` and splits one project's diary.

**Suggestion:** Collapse `/.claude/worktrees/` (and the encoded `-claude-worktrees-` form) inside `_wing_from_cwd`, or only prefer payload cwd for grok/copilot and keep `_wing_from_transcript_path` for claude-code/codex. Add a Stop-hook test whose payload cwd is a worktree path and whose expected wing is the project leaf.

mempalace/hooks_cli.py
```diff
@@ -1369,9 +1773,18 @@ def hook_stop(data: dict, harness: str):
...
                 toast = False
 
-            project_wing = _wing_from_transcript_path(transcript_path)
+            project_wing = _project_wing(parsed, transcript_path)
 
             if silent:
+                if harness == "grok" and (
+                    not transcript_path or _count_human_messages(transcript_path) == 0
+                ):
+                    if _spawn_deferred_grok_save(data):
```
igorls, line 1782: **[bug]** Live Grok Stop is documented as running while `chat_history.jsonl` is still empty, so the interval path defers and returns without writing `{session_id}_last_save`. `since_last` therefore stays `exchange_count - 0`. The next user-turn Stop (16, 17, …) also satisfies `since_last >= 15` and spawns another deferred child. Until a child successfully extracts messages and advances the marker, every post-interval Stop forks a 5s waiter and can write duplicate diary checkpoints.

**Suggestion:** Record a per-session defer-in-flight marker (or advance `last_save` to `exchange_count` when the spawn succeeds) so later Stops do not re-defer. If the child fails, clear the marker so the next interval can retry. Test two Grok Stops at counts 15 and 16 with unflushed history and assert a single spawn.

examples/copilot/hooks.json
```diff
@@ -0,0 +1,29 @@
+{
+  "version": 1,
+  "hooks": {
+    "agentStop": [
+      {
+        "type": "command",
+        "bash": "mempalace hook run --hook stop --harness copilot",
```
igorls, line 7: **[bug]** Copilot CLI hook commands are `bash` (POSIX) and `powershell` (Windows). This example sets `bash` and a non-schema `command` key, with no `powershell` entry. On Windows the install path in the PR body (`~/.copilot/hooks/mempalace.json`) will not run `mempalace hook run`.

**Suggestion:** Add a `powershell` field with the same `mempalace hook run ...` invocation (and drop or keep `command` only if Copilot actually honors it). Mirror that in any Copilot README/install notes.

mempalace/hooks_cli.py
```diff
@@ -1494,7 +1904,8 @@ def hook_session_end(data: dict, harness: str):
     try:
         parsed = _parse_harness_input(data, harness)
         session_id = parsed["session_id"]
-        transcript_path = parsed["transcript_path"]
+        harness = parsed.get("harness") or harness
+        transcript_path = _locate_transcript(harness, parsed)
```
igorls, line 1908: **[suggestion]** Grok SessionEnd locates `chat_history.jsonl` immediately and does not wait or defer. If the user exits while history is still unflushed — the same condition that makes live Stop defer — `_save_diary_direct` sees no user turns and the session flush is skipped. Short sessions that never got a completed deferred Stop can leave nothing in the palace.

**Suggestion:** Reuse `_wait_for_grok_chat_history` (SessionEnd already has a 60s example timeout) or spawn the same deferred save from SessionEnd when history is empty. Cover this with a SessionEnd test whose session dir has events but no history yet.

mempalace/hooks_cli.py
```diff
@@ -1144,21 +1231,326 @@ def _diary_agent_for_harness(harness: str) -> str:
...
+    explicit = str(env.get("MEMPALACE_HOOK_HARNESS") or "").strip()
+    if explicit and explicit != "auto":
+        return explicit
+    if env.get("GROK_SESSION_ID") or env.get("GROK_HOOK_EVENT") or env.get("GROK_WORKSPACE_ROOT"):
+        return "grok"
+    if env.get("CURSOR_TRANSCRIPT_PATH") or (
+        data.get("conversation_id") and data.get("workspace_roots")
+    ):
+        return "cursor"
+    if env.get("COPILOT_HOME"):
```
igorls, line 1274: **[suggestion]** `--harness auto` treats any `COPILOT_HOME` env value as Copilot, even when stdin is a Claude/Codex envelope. Plugin wrappers default to `claude-code`, so the Claude plugin is safe, but documented `auto` detection is wrong for machines that export `COPILOT_HOME` while running another harness. `transcriptPath` ending in `events.jsonl` has the same false-positive shape.

**Suggestion:** Detect Copilot from payload shape (`stopReason` + session-state path) or from `COPILOT_HOME` only when the payload lacks Claude/Codex keys (`transcript_path` + `session_id` snake_case). Document that `auto` is best-effort and that install JSON should pass `--harness grok` / `--harness copilot` explicitly (as the Grok example already does).

igorls (changes requested):
Nice direction. Opening `--harness` past the claude-code/codex enum and splitting the harness-specific bits into a small adapter (aliases, transcript location, turn counting) is the right shape. Claude's snake_case path and the 15-turn silent save still look intact, and the new tests cover the happy paths well.

Three things I'd want fixed before merge:

1. `_project_wing` now prefers payload `cwd`, which skips the worktree collapse from #2388. Claude Stop/SessionEnd always send `cwd`, so worktree sessions will start filing under `wing_<worktree>` and split one project's diary.
2. Grok's deferred save never advances `last_save`, so every Stop after turn 15 can spawn another waiter and write duplicate checkpoints.
3. `examples/copilot/hooks.json` has `bash` but no `powershell`, so the Windows install path in the PR body won't actually run.

Two smaller notes on the inline comments: SessionEnd should wait (or defer) if Grok history hasn't flushed yet, and `--harness auto` treating any `COPILOT_HOME` as Copilot is a bit too eager.

Happy to re-review once those land.

PR #2479, opened 2026-09-10
ci: pin GitHub Actions to immutable commit SHAs for CI/CD best practices

vibemasshq-dev:
Hi team! 👋

I was looking through the CI/CD pipelines and noticed that several GitHub Actions are currently using mutable version tags (e.g., `@v4`, `@v2`). 

Relying on mutable tags for CI dependencies reduces our control over what executes in our build environment. To improve auditability and adhere to strict CI/CD best practices, this PR pins all third-party actions directly to their immutable commit SHAs.

*(For readability, I kept the original version tags as comments next to the SHAs so dependabot can still read them if enabled).*

---

### 🔍 Architecture Heads-Up (Optional)
Some of your core modules (specifically `format_miner.py`, `entity_detector.py`, and `project_scanner.py`) have grown into massive "God-files", carrying a high risk of technical debt and making them harder to maintain as the project scales.

An operational AI risk in `.claude-plugin/skills/mempalace/SKILL.md` where the agent relies on ephemeral context for `since_event_id` instead of persistent storage, which could lead to dropped logs. 

Let me know if you want me to adjust anything in the CI fixes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,8 @@ jobs:
       matrix:
         python-version: ["3.9", "3.11", "3.13"]
     steps:
-      - uses: actions/checkout@v7
-      - uses: actions/setup-python@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
@@ -24,8 +24,8 @@ jobs:
   test-windows:
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@v7
-      - uses: actions/setup-python@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
         with:
           python-version: "3.13"
           cache: 'pip'
@@ -46,8 +46,8 @@ jobs:
   test-macos:
     runs-on: macos-latest
     steps:
-      - uses: actions/checkout@v7
-      - uses: actions/setup-python@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
         with:
           python-version: "3.13"
           cache: 'pip'
@@ -56,8 +56,8 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v7
-      - uses: actions/setup-python@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
         with:
           python-version: "3.11"
           cache: 'pip'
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -80,9 +80,9 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
-      - uses: actions/checkout@v7
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: actions/setup-python@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: dtolnay/rust-toolchain@6bed0761d98439e5a578e2877258200ad565ba87 # stable
```
igorls: **[bug]** `dtolnay/rust-toolchain@6bed0761d98439e5a578e2877258200ad565ba87 # stable` pins the current tip of the `stable` *branch*, not a commit in `master` history. That SHA is the one-off `toolchain: stable` commit whose parent is `master` (`d1031067263f94b142dd6c0ce24c5eb9d02d52a0`). Upstream's `scripts/update-revs.sh` deletes and force-pushes `stable` on every rev update. The action README requires SHA pins to be "within the history of the master branch" and states that any other commit "will eventually get garbage-collected and your workflows will fail." After the next upstream force-push, this pin becomes unreachable and the `test-rust` job cannot resolve the action.

**Suggestion:** Pin a `master` commit and pass the toolchain explicitly:

```yaml
      - uses: dtolnay/rust-toolchain@d1031067263f94b142dd6c0ce24c5eb9d02d52a0 # master
        with:
          toolchain: stable
```

Re-pin from `master` by hand when the action changes; do not pin `stable`/`v1`. Leaving `uses: dtolnay/rust-toolchain@stable` unpinned is the other supported option, but then this workflow is the odd one out.

.github/workflows/release-native.yml
```diff
@@ -71,8 +71,8 @@ jobs:
             artifact_name: mempalace-native-macos
             binary_path: target/release/mempalace-native
     steps:
-      - uses: actions/checkout@v7
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: dtolnay/rust-toolchain@6bed0761d98439e5a578e2877258200ad565ba87 # stable
```
igorls: **[bug]** Same `dtolnay/rust-toolchain` pin as in `ci.yml`. When GitHub GCs `6bed0761…` after the next `stable` force-push, the `binaries` job (`cargo build --release --locked --bin mempalace-native`) will fail to start, blocking native release assets.

**Suggestion:** Apply the same `master` SHA + `with: toolchain: stable` form as in `ci.yml`.

.github/workflows/publish.yml
```diff
@@ -121,4 +121,4 @@ jobs:
           python -m build
 
       - name: Publish to PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # release/v1
```
igorls: **[suggestion]** The SHA is a real release commit (tagged `v1.14.2` and currently the `release/v1` tip), so this is not a GC/rewrite bug. The trailing comment is `# release/v1` — a moving branch name, not a version. This repo already has `package-ecosystem: github-actions` in `.github/dependabot.yml`. Dependabot parses the comment as the current version; a branch name is not SemVer, so the PyPI publish action will not get automated SHA+comment bumps.

**Suggestion:** Use the release tag in the comment:

```yaml
        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
```

.github/workflows/ci.yml
```diff
@@ -13,8 +13,8 @@ jobs:
       matrix:
         python-version: ["3.9", "3.11", "3.13"]
     steps:
-      - uses: actions/checkout@v7
-      - uses: actions/setup-python@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
```
igorls: **[suggestion]** First-party and third-party pins throughout this PR comment only the major (or major.minor) tag that was originally used (`# v7`, `# v6`, `# v5`, `# v4`, `# v2`, `# v1`), not the exact release the SHA belongs to (e.g. this checkout SHA is `v7.0.1`). Dependabot's GitHub Actions updater bumps at the version depth it already sees, so a `# v7` comment typically stays on the v7 major line and does not produce patch/minor SHA updates. GitHub's hardening examples and pinact/zizmor expect `# v7.0.1`. Applies to every `actions/*`, `docker/*`, `oven-sh/setup-bun`, `PyO3/maturin-action`, `actions/upload-artifact`, and `actions/download-artifact` pin in this PR.

**Suggestion:** Put the full tag that the SHA resolves to in each comment (e.g. `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1`).

PR #2491, opened 2026-09-12
test(logstream): assert the idle deadline caps the poll without a clock

L4XB:
## What

`TestLogstreamWatch::test_idle_deadline_caps_the_poll` asserted wall-clock time: a 200ms idle deadline had to finish inside 1.5s. That pins the machine, not the behaviour. The Windows job took 3.71s while capping every poll correctly and failed with

```
AssertionError: idle 200ms waited 3.71s (poll was 5s)
assert 3.7050543999999945 < 1.5
```

(run 34681721829, `test-windows`, on an unrelated PR). Nothing in the watcher was wrong; the budget was.

## How

The contract is the timeout the watcher hands to the long-poll primitive, documented on `Logstream.watch_events` as "an idle of 400ms must not wait out a 300s long-poll". The test now records those timeouts through `Logstream.wait_events` and asserts that none exceeds the 200ms idle deadline while the 5000ms poll timeout is in force, and that the watcher still exits 2. The stub sleeps the timeout it was granted so the deadline elapses the way it would with a real poll instead of spinning the loop.

No production code changes.

## Evidence

- `pytest tests/test_cli_logstream.py` — 65 passed in 4.3s. The rewritten case itself runs in 0.4s with no timing budget.
- Negative control: replacing `return min(args.poll_timeout_ms, remaining_ms)` with `return args.poll_timeout_ms` in `mempalace/cli/cmd_coord.py` fails the new assertion (`poll asked for 5000ms on a 200ms idle deadline`), so the test still guards the cap it was written for.
- `ruff check` and `ruff format --check` (0.16.6, the pinned version) clean on the file.


L4XB:
The point of this PR is confirmed by its own run: `test_idle_deadline_caps_the_poll` **passed on test-windows and on every test-linux matrix entry** in run 34690814848, where the wall-clock version failed on Windows at 3.71s in the run this PR cites.

The two red jobs are different tests, neither of them in the file this PR touches:

| job | test | failure |
|---|---|---|
| test-linux (3.13) | `tests/test_mcp_writer_diagnostics.py::test_unavailable_backend_refusal_and_recovery` | `AssertionError: assert not True` |
| test-windows | `tests/test_closets.py::TestDrawerGrepExpansion::test_hybrid_search_enrichment_populates_drawer_index_and_total` | `assert []` |

Each job reports `1 failed` out of 4566 and 4916 respectively, and the two are disjoint, which is the shape of two independent flakes rather than one regression. This branch changes only `tests/test_cli_logstream.py`, and both of those tests pass locally on this branch (`2 passed in 0.35s`). Neither failure matched the `--only-rerun` filter the workflow uses for the known hnsw flakes, so no rerun was attempted: `test_closets` is a hybrid-search enrichment assertion and `test_mcp_writer_diagnostics` is a backend-availability one, both in the same area that filter exists for.

Happy to leave this until a re-run is convenient, or to widen that `--only-rerun` pattern in a separate PR if you would rather have the filter cover these two as well.

L4XB:
CI is red, and neither failure is from this PR. This branch changes one file, `tests/test_cli_logstream.py` (+25 −5), and no production code at all.

Run [34690814848](https://github.com/MemPalace/mempalace/actions/runs/34690814848), one failure per job, different tests:

- **test-linux (3.13)** — `tests/test_mcp_writer_diagnostics.py::test_unavailable_backend_refusal_and_recovery`, at `assert not (isolated_writer / "palace").exists()`. This is leaked process-global MCP writer state, already diagnosed and fixed in flight by #2422 ("fix(ci): isolate MCP writer-lock state between tests", `tests/conftest.py` only) — its root-cause section names `_MCP_WRITER_LOCK_CM` and friends.
- **test-windows** — `tests/test_closets.py::TestDrawerGrepExpansion::test_hybrid_search_enrichment_populates_drawer_index_and_total`, at `assert []`: the vector side returned nothing. The sibling `..._isolates_chunks_across_drawers_sharing_source_file` passed in the same run. This looks like the Windows hnsw-segment flake the job already reruns for, under a message the `--only-rerun` filter does not match.

Both pass locally on this branch: `pytest tests/test_closets.py::TestDrawerGrepExpansion::test_hybrid_search_enrichment_populates_drawer_index_and_total tests/test_mcp_writer_diagnostics.py::test_unavailable_backend_refusal_and_recovery tests/test_cli_logstream.py` → 67 passed.

Worth noting for this PR's premise: run [34681721829](https://github.com/MemPalace/mempalace/actions/runs/34681721829) on `fix/daemon-foreground-lifecycle-logging` — an unrelated branch — failed test-windows on `tests/test_cli_logstream.py::TestLogstreamWatch::test_idle_deadline_caps_the_poll`, at the wall-clock `elapsed = time.monotonic() - t0` assertion this PR removes. That is the flake this change is meant to eliminate, observed on someone else's branch.

Happy to rebase once #2422 lands if you want a green run on this one first.

igorls (approved):
Verified locally on the PR merged with `develop @ 863462c` (tonight's merge round, after #2422 landed):

- `ruff check` / `ruff format --check` on `tests/test_cli_logstream.py`: clean.
- `pytest tests/test_cli_logstream.py` → `65 passed in 6.84s`. The rewritten case ran 5×, `0.23s call` each, no variance.
- Negative control reproduced: with `return min(args.poll_timeout_ms, remaining_ms)` → `return args.poll_timeout_ms` in `cmd_coord.py`, the test fails with `AssertionError: poll asked for 5000ms on a 200ms idle deadline`.

On the seam: `watch_events` yields at once when the callable returns `<= 0` and never reaches `wait_events`, so the stub only ever records the positive, already-capped values and the `if timeout_ms:` guard can't let the loop spin. Good.

Agree with the read on the two red jobs. Both are the same flakes that hit #2453 and #2474 tonight on heads that already included #2422's conftest fix, so #2422 did not fully close the writer-diagnostics one — tracked in #2493. I've updated the branch to current develop for a fresh run; merging once it's green.

PR #2495, opened 2026-09-13
fix(mcp): stop peer sync thread and bind logstream path to avoid test flake

### LINKED ISSUE #2493: flaky: test_unavailable_backend_refusal_and_recovery asserts palace dir absent, fails intermittently in full-suite CI
`tests/test_mcp_writer_diagnostics.py::test_unavailable_backend_refusal_and_recovery` failed on two unrelated PR heads today, on different platforms, while passing on `develop` pushes and passing in isolation (3/3 locally on Windows at one of the failing heads).

**Occurrences (2026-09-12)**

| PR | Head | Job | Run |
|---|---|---|---|
| #2474 | `8d3f397` (develop @ `3b57f25` merged in) | test-windows | https://github.com/MemPalace/mempalace/actions/runs/34718955868/job/103621133749 |
| #2453 | `b8e9449` (develop @ `d27e515` merged in) | test-linux (3.11) | https://github.com/MemPalace/mempalace/actions/runs/34725627139/job/103639162447 |

Neither PR touches `mempalace/mcp_server/_guards.py`, `mempalace/palace.py`, or this test file. Re-running the failed job passed on #2474.

**Failure**

```
tests/test_mcp_writer_diagnostics.py:38: AssertionError
E       AssertionError: assert not True
ERROR    mempalace_mcp:_guards.py:553 could not acquire MCP peer-writer lock for '.../test_unavailable_backend_refus0/palace': BackendUnavailableError("unknown backend 'unregistered_test_backend'; ...")
```

Line 38 is `assert not (isolated_writer / "palace").exists()`. The refusal itself is produced correctly (the preceding assertions on the error payload pass); what fails is the expectation that no palace directory is created when the backend cannot be resolved. In the full suite that directory sometimes exists, so either the refusal path creates it in some ordering, or state left by an earlier test (a stray lock/handle or a lingering background writer) does.

Only seen in full-suite runs, so likely order- or timing-dependent. Filing so it can be pinned down rather than re-run around.

igorls:
Summary:
- Bind peer sync logstream database path to the specific palace_path at thread startup instead of dynamically resolving global _config.palace_path in the loop.
- Track `_peer_sync_thread` and `_peer_sync_stop_event`, implement `_stop_peer_sync_thread()`, and terminate the loop immediately via `stop.wait(interval)`.
- Clean up the peer sync thread in `_serve_http()` finally block, `TestPeerSyncThreadStartup` teardown fixture, and global `tests/conftest.py` `_reset_mcp_cache`.
- In `tests/conftest.py` `_reset_mcp_cache`, close and clear cached `Logstream` instances in `_logstream_by_path`.
- Add regression test in `tests/test_mcp_writer_diagnostics.py` ensuring peer sync thread does not mutate or create unrelated palace directories.

Fixes #2493

Verification:
- Deterministic reproduction script confirmed failure before fix and clean pass after fix.
- 10 consecutive iterations of targeted test suites ran with zero flakes.
- 504 tests across `tests/mcp/` and logstream suites: 495 passed, 9 skipped, 0 failures.
- Ruff lint (`uv run ruff check`) and formatting (`uv run ruff format --check`) passed cleanly.

### DIFF
--- a/mempalace/mcp_server/http.py
+++ b/mempalace/mcp_server/http.py
@@ -708,6 +708,10 @@ def _mesh_peers_payload() -> dict:
     }
 
 
+_peer_sync_thread: threading.Thread | None = None
+_peer_sync_stop_event: threading.Event | None = None
+
+
 def _peer_sync_interval_s() -> float:
     try:
         return float(os.environ.get("MEMPALACE_SYNC_INTERVAL", "") or 15)
@@ -715,7 +719,20 @@ def _peer_sync_interval_s() -> float:
         return 15.0
 
 
-def _start_peer_sync_thread() -> None:
+def _stop_peer_sync_thread(timeout: float = 2.0) -> None:
+    """Signal shutdown to the peer sync thread and wait for it to exit."""
+    global _peer_sync_thread, _peer_sync_stop_event
+    if _peer_sync_stop_event is not None:
+        _peer_sync_stop_event.set()
+    if _peer_sync_thread is not None and _peer_sync_thread.is_alive():
+        _peer_sync_thread.join(timeout=timeout)
+    _peer_sync_thread = None
+    _peer_sync_stop_event = None
+
+
+def _start_peer_sync_thread(
+    stop_event: threading.Event | None = None,
+) -> threading.Thread | None:
     """Background anti-entropy loop for the logstream (RFC 004 step 0).
 
     Runs in the serving process so a hub with configured peers converges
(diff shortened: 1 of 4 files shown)


igorls:
Addressed review comments in commit 6089f7b:
- **Avoid dynamic path fallback**: Removed \	ry...except TypeError\ wrapper in the sync loop and bound directly to \_get_logstream(canonical_ls_path)\.
- **Worker lifecycle**: Check \stop.is_set()\ between sync iterations; return boolean from \_stop_peer_sync_thread()\ and retain worker tracking if join times out rather than dropping globals; refuse to start duplicate workers if a previous thread remains alive.
- **Deterministic test synchronization**: Replaced fixed sleep in \	est_peer_sync_thread_does_not_pollute_unrelated_palace\ with an explicit \	hreading.Event\ signaled by \sync_all\.

Issue #2500, opened 2026-09-14
Writable HTTP startup exits 2 while a peer holds the writer lease, so the Restart=always from #2203 becomes an unbounded restart loop


empaa:
**What happened?**

A writable HTTP server refuses to start whenever any other process holds the palace writer lease, and it signals that refusal with `SystemExit(2)`:

```python
# mempalace/mcp_server/runtime.py:278-282  (_run_http_loop, develop @ 38260df)
if not _READ_ONLY:
    writer_ok, writer_reason = _acquire_mcp_writer_lock()
    if not writer_ok:
        logger.error("Writable MCP HTTP startup refused: %s", writer_reason)
        raise SystemExit(2)
```

Since #2203, `deploy/mempalace-server.service:30-31` sets `Restart=always` with `RestartSec=2`. The two combine badly: a contended start is not a transient condition the process can recover from, because it exits before binding and re-acquires nothing on the way back up. The unit therefore restarts every few seconds indefinitely — it never starts, never gives up, and never reaches `failed`, so `systemctl is-active` reports `activating` forever and nothing in the unit's state says why.

systemd's own start limiter does not catch this, because the cycle is too slow to trip it. Each attempt lives roughly a second (interpreter start, imports, lease probe, exit), so with the template's `RestartSec=2` that is about 3.3 starts per 10 s against the default `StartLimitBurst=5` / `StartLimitIntervalSec=10s`. On my unit (`RestartSec=5`, ~6.5 s per cycle) the restart counter climbed monotonically past 27 before I intervened:

```
sep 14 05:21:51 systemd[1166]: mempalace-hub.service: Scheduled restart job, restart counter is at 18.
sep 14 05:21:52 mempalace[780941]: Writable MCP HTTP startup refused: another mempalace writer already
    holds the palace lock for '/home/emil/.mempalace/palace': palace ... is held by PID 88779
    (/home/emil/.local/bin/mempalace-mcp); wait for it to finish or stop the holder before retrying
sep 14 05:21:57 systemd[1166]: mempalace-hub.service: Scheduled restart job, restart counter is at 19.
sep 14 05:22:03 systemd[1166]: mempalace-hub.service: Scheduled restart job, restart counter is at 20.
```

Each cycle pays a full interpreter and import startup (~1.6 s CPU here) to fail the same way.

**Why this state is reachable in normal operation**

This is not an exotic race. It is the successor state to #2201, and the fix for #2201 is what makes it visible.

[...]

PR #2501, opened 2026-09-14
fix(mcp): run the registered cleanup when the idle watchdog exits

L4XB:
Refs #2500 — the two halves that do not need a design decision. The "wait for the lease" option is deliberately **not** here; see **Scope**.

## 1. The watchdog exit — and why the issue's own suggestion is wrong

`os._exit(0)` skips every `atexit` handler, so a watchdog exit leaves `serverinfo.json` advertising a dead PID. Two handlers are registered by then: `server_registry.clear_serverinfo` and `_release_mcp_writer_lock`.

The issue proposes `sys.exit(0)`. **That would break the watchdog.** It runs in a daemon thread, where `SystemExit` unwinds that thread and nothing else. Measured, three modes side by side on this interpreter:

| | process exits? | `atexit` runs? |
| --- | --- | --- |
| `os._exit(0)` — today | yes | **no** ← the reported bug |
| `sys.exit(0)` — the suggestion | **no**; the thread dies and the watchdog, having returned out of its own `while True`, never fires again | — |
| `atexit._run_exitfuncs()` + `os._exit(0)` | yes | **yes** |

So that one-liner trades a stale `serverinfo.json` for the file-handle accumulation the watchdog exists to prevent (#1552) — the same class of failure as the 900 MB–1 GB RSS sessions in your own field data.

`atexit._run_exitfuncs` is private, and chosen on purpose: the public surface has no "run the handlers now", and calling the two known handlers directly from the watchdog silently drops any third one a later change registers. A handler that raises must not keep the process alive either, so the exit is in a `finally`.

## 2. The restart loop

`deploy/mempalace-server.service` gains `RestartPreventExitStatus=2`, with the reasoning inline. Status 2 is the writable-startup refusal, and it is not transient *from this process's side* — it exits before binding and re-acquires nothing on the way back up — so the `Restart=always` from #2203 turns it into a loop that never starts, never gives up, and never reaches `failed`. systemd's start limiter does not catch it, for exactly the arithmetic in your report.

The unit now lands in `failed` with the refusal reason in the journal: option 2 of your "expected", and a state `systemctl status` can show.

## Tests

`tests/test_idle_watchdog_exit.py`, four cells, including a control that measures the daemon-thread `SystemExit` semantics directly — so if that ever stops holding, the helper's docstring goes red rather than quietly becoming wrong.

| mutation | red |
| --- | --- |
| watchdog calls `os._exit(0)` directly (i.e. `develop` today) | the wiring cell |
| exit before the cleanup | 2 cells |
| drop the `try`/`finally`, so a raising handler blocks the exit | the raising-handler cell |

The wiring cell's stand-ins **raise** rather than return, and that is load-bearing: the watchdog loop ignores a return, so a returning stand-in leaves a live daemon thread that calls the real `os._exit` the moment monkeypatch restores it at teardown — killing the pytest process and the rest of the suite. I hit that while writing this.

```
pytest <the ten files naming the watchdog, serverinfo or the writer lock>   265 passed
```

## Scope

Not done: **option 1**, the writable server waiting for the lease instead of exiting. That is the better end state and your error message already prescribes it, but it is a behaviour change to the startup contract — how long to wait, what to log while waiting, whether a read-only fallback is preferable to a wait — and I did not want to pick those on your behalf inside a fix for the loop. Say the word and I will send it as its own PR.

The two changes here are independent of that decision: the watchdog cleanup is right regardless, and `RestartPreventExitStatus=2` stops the loop whether or not the refusal later becomes a wait.


igorls:
Reviewed against `develop` @ `e59e368` (merges cleanly, CI 12/12). Approach is right: `sys.exit(0)` from the watchdog daemon thread would only unwind that thread, and running `atexit` handlers before `os._exit(0)` in a `finally` keeps both the serverinfo cleanup and the exit. `RestartPreventExitStatus=2` also composes with the follow-up we are adding for #2500 (a bounded wait for the writer lease at startup): on timeout the server still exits 2 and the unit lands in `failed`.

One note, not a blocker: a registered `atexit` handler that blocks (rather than raises) would stall the idle exit, since the watchdog now waits for the handlers. The two handlers registered today (`clear_serverinfo`, `_release_mcp_writer_lock`) do not block. Merging.

Issue #2498, opened 2026-09-14
A named pipe at chroma.sqlite3 stops the MCP server from answering


mvalentsev:
**What happened?**

If `chroma.sqlite3` in a palace is a named pipe, or a symlink to one, the integrity probe opens it for reading and blocks until a writer shows up, while holding `_sqlite_integrity_refresh_lock`. Gated tool calls, `mempalace_status`, `mempalace_reconnect` and `GET /statusz` all run that probe or wait behind it.

On `develop` @ 38260df5, with a palace whose only entry is that `chroma.sqlite3`:

- stdio: the first `tools/call` for `mempalace_list_wings`, `mempalace_status` or `mempalace_reconnect` gets no answer in 30 s, and a `ping` sent after it gets none in 5 s. The session stays silent until the idle watchdog exits the process, 8 h by default.
- HTTP: the same three tools and `GET /statusz` get no answer in 15 s each, while `ping` and `GET /healthz` still answer.
- `python -m mempalace.repair rebuild` hangs in the same probe.

With an empty regular file in its place, those MCP requests answer within 1 s. `faulthandler` shows the probe blocked in `open(db_path, "rb")` in `config._is_wal_without_sidecars`, reached from `repair._quick_check_errors`. 3.9.0 hangs too, in `sqlite3.connect` in that function.

**What did you expect?**

The probe to refuse the pipe by type before opening it, as `sqlite_drawer_count` does for `repair-status` since #2393, and to report an integrity error. Gated tool calls would then get the `-32002` refusal, `mempalace_status` and `/statusz` would report `ok: false`, and the session would keep answering.

**How to reproduce:**

1. `mkdir /tmp/fifo-palace && mkfifo /tmp/fifo-palace/chroma.sqlite3`
2. Start `python -m mempalace.mcp_server --palace /tmp/fifo-palace` and send `initialize`, then `tools/call` for `mempalace_list_wings`, then `ping`, one JSON-RPC message per line.
3. `initialize` is answered. The tool call and the `ping` are not.

**Environment:**
- OS: Linux (WSL2)
- Python version: 3.12.3 (SQLite 3.45.1)
- MemPal version: `develop` @ 38260df5, and 3.9.0

Issue #1599, opened 2026-05-24
Bug: recurring FTS5 corruption under live writes after transient vector failure mitigation


trek-e:
**What happened?**

A recurring FTS5 corruption path still appears under live writes, even after the transient vector/HNSW hard-failure path was mitigated.

Observed behavior:
- `PRAGMA quick_check` intermittently reports malformed FTS state (`database disk image is malformed`).
- `embedding_fulltext_search` can be temporarily repaired via FTS rebuild, but corruption reappears later during active write periods.
- Wing-scoped search no longer hard-fails after #1586-class fixes (retry + BM25 fallback), but this appears to mask a separate ongoing FTS integrity problem.

This appears operationally related (same live-write pressure period) but mechanistically separate from the HNSW false-quarantine issue.

**What did you expect?**

FTS index integrity should remain stable under concurrent/live writes, with `PRAGMA quick_check` remaining healthy and without recurring malformed FTS state.

**How to reproduce:**

1. Start with a palace receiving active writes (CLI mine / hooks / MCP writes).
2. Run periodic integrity/status checks during write activity:
   - `sqlite3 ~/.mempalace/palace/chroma.sqlite3 'PRAGMA quick_check;'`
   - `mempalace repair-status --palace ~/.mempalace/palace`
3. If malformed appears, run an FTS rebuild (e.g. `INSERT INTO embedding_fulltext_search(embedding_fulltext_search) VALUES('rebuild');`) and continue live writes.
4. Observe that FTS can return to healthy state briefly, then malformed status recurs under continued live writes.

**Environment:**
- OS: macOS (Darwin)
- Python version: 3.12.13
- MemPal version: commit `d9a2e67` (and reproduced before that during #1586 diagnosis)

Related:
- PR #1598 addresses transient search hard-failure + rebuild lock race + HNSW fresh-segment false quarantine.
- This issue tracks remaining recurring FTS corruption under live writes.

trek-e:
Diagnosis + resolution is now in PR #1598 (head: ad2631e).\n\nWhat was implemented:\n- transient vector retry/fallback parity for CLI + API search paths\n- lock-aware FTS self-heal for isolated  malformed-index quick_check failures\n- explicit  path\n-  lock-holder diagnostics for live contention triage\n- regression tests for legacy auto-heal and explicit FTS rebuild modes\n\nPR body now includes , so this issue will auto-close on merge.

trek-e:
[...]

PR #2506, opened 2026-09-15
fix(storage): keep Python sqlite3 readers from breaking Chroma's SQLite in-process (#2302)

igorls:
**Base:** `develop` · **Head:** `fix/2302-inprocess-sqlite-locks` · **Refs:** #2302 (fixes the Linux and in-process parts; macOS cross-process part stays open, see below) · **Related:** #2472, #2471, #2240, #1599

## What does this PR do?

ChromaDB opens `chroma.sqlite3` through the SQLite bundled in `chromadb_rust_bindings`. mempalace's fast paths (`status`, taxonomy, `list_drawers`, BM25, the MCP integrity gate, repair) open the same file through Python's `sqlite3` **in the same process**: two independent SQLite libraries on one file. #2472 removed the plain-descriptor header sniff; the SQLite connections themselves had the same two problems.

Measured on Linux (WSL2, Python 3.12 / SQLite 3.45.1, chromadb 1.5.7), unpatched `develop`:

| Check | Unpatched |
|---|---|
| SHARED lock on the db and DMS lock on `-shm` after one `_sqlite_wing_room_counts()` | **both gone**, and still gone after further Chroma writes and reads |
| `sqlite_integrity_status()` looping while a thread writes through `ChromaBackend`, rollback journal, 10 s | **35 / 66 probes failed** (`database disk image is malformed`) |
| same, WAL, 10 s | **9 / 58 probes failed** |
| fresh `PRAGMA integrity_check` afterwards | `ok` (the failures are torn reads, not on-disk damage) |

1. POSIX fcntl locks belong to (process, inode). Closing a Python connection dropped every lock the process held on the file, Chroma's included, and Chroma never re-took them. The palace stayed lock-naked to every other process: the precondition for the `-shm` truncation / WAL unlink in the issue.
2. fcntl locks never conflict within a process, so neither library sees the other's locks and a Python read races a Chroma commit.

### The fix: `mempalace/backends/_inproc_sqlite.py`

- **`palace_db_lock(db_path)`**: process-local re-entrant lock per database file. Every Python connection opened through this module holds it from open to close; `ChromaBackend` holds it around `add`/`upsert`/`update`/`delete` (inside `mine_palace_lock`), client open/replace (`_client`, `make_client`), collection open/create/delete and `close_palace`. Fixes (2) on every platform.
- **`open_reader()` / `open_writer()`**: `sqlite3.Connection` proxies whose `close()` also releases the lock, so each call site changes by one line (`connect_sqlite_read(...)` → `open_palace_reader(...)`, `sqlite3.connect(...)` → `open_palace_writer(...)`). Converted: 9 readers + 2 migrations in `chroma.py`, 3 readers + FTS heal + VACUUM + 3 max-seq-id writers in `repair.py`, `sqlite_bm25`, and the MCP `status` sqlite fallback.
- **Linux: an anchor per database inode.** Before a reader opens, one idle Python connection is kept on the file for the life of the process. It only reads the schema cookie, which in WAL mode leaves it holding SHARED and the wal-index mapping, so Python's SQLite *parks* the descriptor of every per-call connection that closes afterwards instead of closing it. Chroma keeps its locks. The anchor also takes **OFD read locks** on SQLite's SHARED range and the `-shm` DMS byte (`F_OFD_SETLK`): per-descriptor, released by no other close, and in conflict with this process's own fcntl locks.

Two things I measured while building this, which is why it is shaped the way it is:

| Variant (Linux, WAL) | Chroma close → reopen → 2 more writes |
|---|---|
| no anchor | sidecars recreated on reopen; fresh connection sees 3/3 rows |
| anchor **without** OFD guards | sidecars recreated under the anchor; **a fresh Python connection sees 1/3 rows** (attached to the stale wal-index) |
| anchor **with** OFD guards (this PR) | sidecars untouched; anchor-side and fresh reads see 3/3 |

- A plain long-lived Python connection is worse than none, hence the guards.
- The reads cannot go through the long-lived connection either: after Chroma's commits it reported **`malformed inverted index for FTS5 table main.embedding_fulltext_search`** (both journal modes, even after Chroma closed) while fresh connections and a later `integrity_check` said `ok`. That would have made the integrity gate refuse tools and triggered needless FTS heals. So every read uses a fresh connection; the anchor never serves queries.

**Windows**: locks are per handle, so a close never drops another handle's locks; only the lock applies (per-call connections, no anchor). **macOS / other POSIX**: CPython does not expose `F_OFD_SETLK`, so no anchor; the in-process serialization applies, but a Python close can still drop Chroma's locks. I have not verified OFD behaviour on macOS, so this stays open on #2302 rather than being guessed at.

### After (same machine, same scripts)

| Check | Patched |
|---|---|
| SHARED + DMS after every fast-path read, Chroma rewrite, integrity probe | held throughout |
| integrity probe vs writer thread, rollback journal, 10 s | **0 / 124 failed** |
| same, WAL, 10 s | **0 / 137 failed** |

### Costs and limits worth reviewing

- Python readers of a palace now wait while this process's Chroma write, client open or collection open is running (including Chroma's embedding computation inside `add`). A long `quick_check` likewise delays this process's Chroma writes to that palace. Cross-process behaviour is unchanged.
- `repair.extract_via_sqlite` is a generator and holds the source palace's lock across its yields; its consumer writes into a different palace.
- Rollback-journal mode without an anchor lock between reads: a per-call close can still drop the SHARED lock of a Chroma *read* in flight on another thread (writes are serialized). WAL palaces are covered by the anchor.
- `migrate.py` still opens `chroma.sqlite3` directly; it runs offline before a client exists, and importing `backends` there would pull chromadb into a module that imports it lazily on purpose.
- `_is_wal_without_sidecars` still does a plain `open()`; it only runs when both sidecars are absent, which no live WAL connection allows.

## How to test

```bash
uv run pytest tests/test_inproc_sqlite_locks.py -v
```

`tests/test_inproc_sqlite_locks.py`:

- all platforms: Chroma writes hold the palace lock; a reader holds it until `close()` (idempotent); a failed open releases it; nested reads and `row_factory` don't leak; `open_writer` as a context manager commits and releases; a replaced database is read afresh.
- POSIX: `test_quick_check_never_reads_a_chroma_commit_in_progress[rollback-journal|wal]`: writer thread vs integrity loop for 2 s, no errors.
- Linux (OFD): `test_fast_path_reads_keep_the_palace_locked`: SHARED + DMS survive `_sqlite_embedding_count`, `_sqlite_wing_room_counts`, `sqlite_room_wing_hall_counts`, `sqlite_integrity_status` (which must also be clean); an external clean close leaves the sidecars alone and a later write is visible from another process. `test_chroma_reopen_keeps_python_reads_current`: the 1-of-3-rows case above.

Test adjustments: `tests/conftest.py` closes anchors after each test (otherwise thousands of temp palaces exhaust descriptors); `test_sqlite_integrity_errors_uses_bounded_contention_timeout` pins the non-anchored open, since on Linux the anchor's connect precedes the one it asserts on.

Results:

- Linux (WSL2 Ubuntu, Python 3.12, SQLite 3.45.1, chromadb 1.5.7): patched 5139 passed, 215 failed, 53 skipped; unpatched base `527d0810` in the same environment 5129 passed, 215 failed, 53 skipped. The two failure sets are identical (compared by test id) and all 215 are shell-hook tests (`test_cursor_hooks_shell`, `test_antigravity_hooks_shell`, `test_hooks_bash_compat`, ...) failing because the copy was made from a Windows checkout with CRLF hook scripts; the +10 passes are the new tests. CI's Linux job is the clean reference.
- Windows 11 (local uv env, Python 3.12): full suite on the rebased tree: 4925 passed, 10 failed, 431 skipped. All 10 failures are `test_claude_plugin_hook_wrappers.py` (bash wrapper exits 127, runner not found in this shell environment); the same 10 test ids fail on the unpatched base in the Linux run above. Targeted suites earlier: 496 passed, 12 skipped (the POSIX-only lock tests). Locally the run has to skip `tests/test_known_entities_registry.py`: `test_eperm_on_both_names_falls_back_too` patches `os.open` to raise EPERM, and `tempfile.mkstemp` on Windows retries PermissionError as a name collision, so on Python 3.12 it spins indefinitely (py-spy: `_mkstemp_inner` ← `miner._publish_registry`). Unrelated to this branch (neither file is touched); CI's Windows job runs 3.13.
- after rebase onto `develop` @ `30cb69ab` (includes #2498 and #2504), Linux: `test_inproc_sqlite_locks`, `test_backend_header_probe_locks`, `test_non_regular_file_guards`, `test_config`, `test_backends`, `test_repair`, `mcp/test_read_tools`: 563 passed.
- `ruff check .` and `ruff format --check .` clean.

Not run: macOS.

## Checklist
- [x] No hardcoded paths


mempalace/backends/_inproc_sqlite.py
```diff
@@ -0,0 +1,386 @@
...
+
+# Whether readers keep an anchor on the database. Module-level so tests can pin
+# either path.
+_ANCHORED = _F_OFD_SETLK is not None
+
+# A guard that conflicts is somebody's short EXCLUSIVE or DMS recovery. Retry
+# briefly, then serve the read and try again on the next one.
+_GUARD_ATTEMPTS = 3
+_GUARD_RETRY_SECONDS = 0.002
+
```
igorls: **[bug]** `_ANCHORED` is true only when `fcntl.F_OFD_SETLK` exists and `sys.platform` starts with `linux`. Python's `fcntl` module does not expose `F_OFD_SETLK` on macOS, so every macOS (and other non-OFD POSIX) reader still opens and closes a per-call connection with no anchor. Closing that descriptor drops every POSIX fcntl lock this process holds on `chroma.sqlite3` and on `-shm`, including the DMS lock Chroma's separate SQLite keeps for the life of the client. Chroma does not re-take those locks. The next process, including a freshly spawned `mempalace.mcp_server`, can pass SQLite's first-opener check and truncate `-shm` under the live daemon. That is the failure this PR is aimed at, and it was captured on macOS. `mode=ro` does not prevent it. The module docstring states this residual exposure on purpose: an anchor without OFD locks lets Chroma's in-process `close` take EXCLUSIVE, recreate `-wal`/`-shm`, and leave later Python connections on the stale wal-index (measured as 1 of 3 committed rows). Leaving the anchor off avoids the stale read and leaves the truncation in place. `has_sqlite_magic`'s retained `immutable=1` probe does not close this hole: it never maps `-shm`, and a retained fd does not stop a later Python `close` from dropping process locks. The new lock tests skip the lock-holding cases unless `_ANCHORED`.

**Suggestion:** Hold the DMS and SHARED locks from something whose `close` cannot drop Chroma's locks on macOS (a short-lived helper process, or an equivalent per-fd lock that actually conflicts with SQLite's fcntl locks). Do not enable the in-process anchor on macOS until a Chroma close/reopen can no longer leave Python on a replaced wal-index. Add a regression that is not Linux-only: a live holder must still own DMS after a same-process fast-path read, and a second process must fail to truncate `-shm`.

mempalace/backends/_inproc_sqlite.py
```diff
@@ -0,0 +1,386 @@
...
+        try:
+            conn.execute("PRAGMA schema_version").fetchone()
+        except sqlite3.Error:
+            conn.close()
+            return
+        anchor = _anchors[key] = _Anchor(conn, ident)
+
+    try:
+        _guard(anchor, db_path)
+    except sqlite3.Error:
```
igorls: **[bug]** On the Linux path, a failed anchor setup closes the connection it just opened. `connect_sqlite_read(..., timeout=0.0)` plus `PRAGMA schema_version` returns `SQLITE_BUSY` immediately if another process holds a checkpoint or recovery lock. That open has already attached to the database and, once the WAL is touched, to `-shm`. `conn.close()` is the POSIX `close` that drops every fcntl lock this process holds on those inodes, including Chroma's, and the anchor is not stored, so `_guard` never runs. The per-call reader then opens with no anchor and drops the locks again when the caller finishes. Chroma does not re-acquire them. One busy probe under load disarms the live client until a later read both anchors and takes the OFD guards. `anchor.discard()` has the same `close` when a wal-index inode changes; that path at least runs only after a replacement. This one runs while the existing `-shm` is still the live one.

**Suggestion:** Do not `close()` a connection that has touched this database while Chroma may be open. Retain the fd (park the failed connection as the anchor, or otherwise keep the descriptor open) and let a later read retry `schema_version` and `_guard`. A busy anchor must not be worse than no anchor.

mempalace/backends/_inproc_sqlite.py
```diff
@@ -0,0 +1,386 @@
+"""Python ``sqlite3`` access to a database another SQLite library in this process holds.
+
+ChromaDB 1.5+ opens ``chroma.sqlite3`` through the SQLite statically linked
+into ``chromadb_rust_bindings``. mempalace's fast paths (status, taxonomy,
+``list_drawers``, BM25, the integrity gate, repair) open the same file through
+Python's ``sqlite3``: a second, independent SQLite library in the same
+process. Two consequences, both measured on Linux (#2302):
```
igorls: **[suggestion]** New comments and test docstrings cite a GitHub issue number, including this module docstring and comments in `chroma.py`, `repair.py`, `tests/conftest.py`, `tests/test_inproc_sqlite_locks.py`, and `tests/test_repair.py`. The pre-existing mention in `_magic.py` is unchanged. A changelog line could carry the number; these are code comments and test docstrings.

**Suggestion:** Describe the POSIX lock and wal-index failure in words and drop the issue number from comments and test docstrings.

PR #2512, opened 2026-09-15
test(storage): wait for the writer lock to release after close

igorls:
## Summary

`tests/test_backend_header_probe_locks.py::test_probes_keep_chroma_writer_shared_lock` fails intermittently on Linux CI, turning `develop` red with no code change involved:

- `527d081` (#2497) — `test-linux (3.13)`
- `752965d` (#2505) — `test-linux (3.11)`

Both failures are the final assertion, `closing the writer releases the lock`, right after `backend.close()`.

## Cause

Closing Chroma's last `PersistentClient` stops its `System`, and for the Rust backend that is `RustBindingsAPI.stop()` → `del self.bindings`. That only drops the Python reference: the native object that owns the SQLite connection (and the SHARED lock) is freed when it is actually collected, which is not guaranteed by the time `close()` returns — the test also still held `col`. Probing the lock on the very next line raced that teardown.

## Change

Test-only. In both writer tests (`test_probes_keep_writer_shared_lock`, `test_probes_keep_chroma_writer_shared_lock`), after `backend.close()` the test drops `col` and asserts through `_lock_released_within(db_path, timeout=5.0)`, which runs `gc.collect()` and polls the out-of-process lock probe every 100 ms. A lock that is never released still fails after five seconds, so the test keeps guarding against a leaked writer lock.

## Evidence (WSL2 Ubuntu, Python 3.12.3, chromadb 1.5.7, SQLite 3.45.1)

The single test, run repeatedly: 100 sequential runs, then 100 rounds of 4 concurrent runs (400) to mimic CI load.

| Ref | Sequential | 4 in parallel | Total |
|---|---|---|---|
| `develop` @ `48b262e` | 2 / 100 failed | 4 / 400 failed | **6 / 500** |
| this branch | 0 / 100 failed | 0 / 400 failed | **0 / 500** |

A failure from an earlier 125-run baseline on `develop` was inspected and is the same `closing the writer releases the lock` assertion (line 206) seen in CI; the per-run logs of the 500-run baseline were lost when the WSL instance recycled `/tmp`, so those six are counted, not individually inspected. At develop's rate (~1.2%), 500 clean runs would occur by chance about 0.2% of the time.

The whole module plus `tests/test_backends.py` also passes on this branch under WSL: 112 passed. (The module is POSIX-only, so it skips on Windows.)


igorls:
Force-pushed 38e3987 to replace c3c7ffb: the original commit rewrote 322 files with CRLF line endings (Windows checkout, core.autocrlf=false plus git commit -a), which broke the Docker smoke test. The new commit is the same one-file test change with LF endings.

Issue #1398, opened 2026-05-07
MCP mempalace_search returns 'Error finding id' after bulk-add until WAL flushes; CLI works


docrandy:
### Summary

`mcp__mempalace__mempalace_search` returns `Error finding id <drawer_id>` for every call where `wing` is set, immediately after a bulk-add transaction (e.g. an on-commit hook that backfills hundreds of drawers into one wing). The CLI command `mempalace search --wing <name> "..."` works against the same palace at the same moment.

The error self-heals once chromadb's WAL flushes and HNSW segment compaction catches up — typically a few minutes to a few hours later. While the divergence is active, the MCP audit path is blind: an audit of a 108-drawer wing surfaced only 9 results.

### Environment

- mempalace 3.3.4 (pip)
- chromadb 1.5.8
- Windows 11 / Python 3.13
- Palace built via `mempalace mine` then bulk-extended via a custom on-commit hook that adds ~600 drawers in one transaction

### Reproducer

1. `mempalace init ./testpalace`
2. Add a small initial set: 5–10 drawers in wing `wing_a`
3. In one transaction (one `col.add(...)` call with batched ids/docs/metas), add 500+ drawers tagged `wing_a`
4. **Immediately** call `mempalace_search` via MCP with `wing="wing_a"`
5. Observe: `{"error": "Search error: Error finding id <some-drawer-id>"}`
6. Run `mempalace search --wing wing_a "..."` from CLI — works fine
7. Wait 10–30 min, retry MCP — works fine

### Why MCP fails when CLI doesn't

CLI search path (`searcher.py:284 search()`) makes one `col.query(...)` call.

MCP search path (`searcher.py:548 search_memories()`) makes that call **plus** a second `closets_col.query(where=...)` and, for closet-boosted hits, a third `drawers_col.get(where={"source_file": ...})`. Each extra call multiplies the chance of hitting a stale-ID lookup in the divergent segment.

### Root cause (educated guess — confirm with chroma maintainers)

The chroma 1.5.x Rust core (`chromadb_rust_bindings`) raises `Error finding id` when its internal where-filter → ID-lookup walks up an ID that exists in the metadata layer but is missing from the HNSW segment file. This is a WAL-vs-segment divergence, not a logic bug in mempalace's query construction — the where filter `{"wing": "wing_a"}` is well-formed.

I have filed a companion issue against chroma-core/chroma asking for the Rust core to return an empty result + warning rather than raise.

### Suggested fixes (in mempalace)

[...]

Issue #1254, opened 2026-04-28
`repair --mode max-seq-id` crashes with ValueError on BLOB-typed `embeddings.seq_id` rows


NickShtefan:
**What happened?**

`mempalace repair --mode max-seq-id` (added in #1135 to recover palaces poisoned by the old `_fix_blob_seq_ids` shim) crashes during the dry-run inspection on palaces where `embeddings.seq_id` rows are still in chromadb 1.5.x's native BLOB format (8-byte big-endian uint64).

`_compute_heuristic_seq_id` (`mempalace/repair.py:552`) computes the recovery value via:

```sql
SELECT MAX(e.seq_id)
FROM embeddings e
JOIN segments s ON e.segment_id = s.id
WHERE s.collection = (SELECT collection FROM segments WHERE id = ?)
```

then calls `int(row[0])` at line 576. When `MAX(e.seq_id)` returns a BLOB (which it does on palaces where chromadb 1.5.x has been writing seq_ids natively), `int(bytes)` raises `ValueError`.

**What did you expect?**

The recovery tool should successfully detect and report the heuristic value for poisoned segments without crashing on BLOB-typed `embeddings.seq_id`. After dry-run, the user should see what would be repaired and have the option to apply.

Specifically:
- `_compute_heuristic_seq_id` should accept both INTEGER and BLOB return types from `MAX(e.seq_id)`
- BLOB should be decoded as 8-byte big-endian uint64 (matching chromadb's storage format)
- Dry-run should print detected poisoned rows and the proposed clean values, exit 0

**How to reproduce:**

1. Have a palace where `embeddings.seq_id` has at least some BLOB-typed rows. In my case all 11684 rows had been migrated by mempalace's narrowed `_fix_blob_seq_ids` shim, but at least one raw row remained (chromadb 1.5.x writes new seq_ids in native BLOB format).

2. Run dry-run repair:

   ```
   $ mempalace repair --mode max-seq-id --dry-run

   Traceback (most recent call last):
     File ".../mempalace/__main__.py", line 5, in <module>
       main()
     File ".../mempalace/cli.py", line 1484, in main
       dispatch[args.command](args)
     File ".../mempalace/cli.py", line 813, in cmd_repair
       repair_max_seq_id(...)
     File ".../mempalace/repair.py", line 677, in repair_max_seq_id
       new_val = _compute_heuristic_seq_id(cur, seg_id)
[...]

PR #2513, opened 2026-09-15
fix(chroma): clear a quarantined segment's max_seq_id watermark

### LINKED ISSUE #2428: quarantine_stale_hnsw / quarantine_invalid_hnsw_metadata leave max_seq_id behind, so the rebuild they force is permanently incomplete (silent partial index; only filtered queries error)
## Summary

Both quarantine paths in `mempalace/backends/chroma.py` rename a suspect HNSW segment directory but leave that segment's `max_seq_id` row in `chroma.sqlite3`. Chroma then treats the brand-new **empty** index as already synced through the old watermark and never replays anything below it.

The quarantine exists to force a rebuild. Instead it deterministically produces a **permanently incomplete index** — and `collection.count()` keeps reporting the full total, because that reads the metadata segment.

Still present on `develop` today:

- `quarantine_stale_hnsw` — rename at L541, no `max_seq_id` handling
- `quarantine_invalid_hnsw_metadata` — rename at L1655, same

## Observed

Palace on mempalace 3.3.5 / chromadb 1.5.8, macOS. Three numbers that should agree:

| | value |
|---|---|
| `embeddings` rows, drawers metadata segment | **981** |
| `max_seq_id` for the VECTOR segment | **798** |
| vectors actually in the HNSW (`length.bin` 400 B ÷ 4) | **100** |

Triggered by a `.drift-20260831-094810` quarantine three days before I noticed. The palace holds **62** `.drift-*` directories spanning 2026-05-18 → 2026-08-31, the first dated **8 minutes** after install — so this fires constantly, and each firing is a chance to truncate the index.

## Why it stayed hidden — the part I think matters most

**It fails asymmetrically.** Unfiltered queries kept working and kept returning *plausible, relevant* results, because an unfiltered NN walk just traverses whatever nodes exist. Only `where={"wing": ...}` filtered queries raised:

```
Error executing plan: Internal error: Error finding id
```

because those resolve specific ids and look each one up in the index.

So for three days search silently answered from ~10% of the corpus while looking healthy. The first filtered query is what surfaced it. That is strictly worse than a hard failure, and I think it explains why this cluster keeps getting re-reported from the symptom end rather than the cause end (#1665, #1035, #1082, #1398, #2373 are all the filtered-query symptom).

## Reproduction

1. Build a collection past `hnsw:sync_threshold` so a `max_seq_id` row is persisted.
2. `mv <palace>/<vector-seg-uuid> <palace>/<vector-seg-uuid>.drift-TEST` (exactly what quarantine does).
3. Reopen and run a **filtered** query.

Note step 1 — below the sync threshold no watermark row exists, so the bug does not reproduce. I could not reproduce the hard error synthetically at 1200 docs on 1.5.8; my evidence for the failure mode is the live palace above plus code inspection. Flagging that honestly rather than overclaiming.

## Fix

Delete the row for the quarantined segment after a successful rename. The directory name **is** the segment UUID:

```python
def _clear_segment_watermark(db_path: str, segment_id: str) -> None:
    try:
        with contextlib.closing(sqlite3.connect(db_path)) as conn:
            conn.execute("DELETE FROM max_seq_id WHERE segment_id = ?", (segment_id,))
            conn.commit()
    except sqlite3.Error:
        logger.exception(
            "Could not clear max_seq_id for quarantined segment %s — "
            "the index will rebuild INCOMPLETE.", segment_id,
        )
```

called after `os.rename(...)` in both quarantine functions. With no watermark, Chroma replays every record from `embeddings` and rebuilds a complete index.

### This does not conflict with the `_fix_blob_seq_ids` decision

`_fix_blob_seq_ids` deliberately leaves `max_seq_id` alone because chromadb 1.5.x writes its own BLOB format there and misreading it via `int.from_bytes(..., 'big')` yields ~1.23e18 and suppresses writes (#1134, #1254).

**`DELETE` never interprets the value.** The concern that motivated "chromadb owns this column" is about *rewriting* it; removing a row for a segment whose index was just discarded is the one operation that is format-agnostic and provably safe.

### `repair --mode max-seq-id` cannot catch this class

That detector looks for implausibly **large** watermarks left by the old shim (`_MAX_PLAUSIBLE_SEQ_ID`). `798` is entirely plausible — it is only wrong *relative to an index that was just emptied*. There is no absolute threshold that separates these; the check would have to compare the watermark against the index's actual element count.

## Recovery for an already-broken palace

Clearing the watermark alone fixed **unfiltered** search but **not** filtered. Both steps were needed:

```sql
DELETE FROM max_seq_id WHERE segment_id IN ('<vector-seg-uuids>');  -- leave METADATA rows
```
```
python3 -m mempalace.repair rebuild --palace ~/.mempalace/palace --confirm
```

Verified afterwards in a **fresh process** (the MCP server caches segment handles and will mislead you in both directions): drawers unfiltered, `wing=` ×2, `room=` ×2, and closets — all OK across 981 + 28.

## Minor, same area

Quarantine skips names containing `.drift-` or starting with `.`, so anything else with a dash and a `data_level0.bin` looks live. A backup I parked as `<seg>.stale-watermark-<stamp>` **inside** the palace got quarantined on the next scan. Widening the skip pattern, or documenting "never keep backups inside the palace directory", would help. Possibly related to the `.drift-` churn in #1715.

## Related

#1710 (closest — quarantine destroying flushed vectors), #1526 (quarantine cascade + silent empty index), #1665 / #1035 / #1082 / #1398 / #2373 (the filtered-query symptom), #1963 (epic), #1715 (`.drift-` churn), #1134 / #1254 (the `max_seq_id` format hazard this proposal avoids).

## Environment

mempalace 3.3.5 (dist-info 2026-05-18), chromadb 1.5.8, Python 3.9, macOS. Bug verified present on `develop` at time of filing.

L4XB:
## What does this PR do?

Closes #2428.

Both quarantine paths in `mempalace/backends/chroma.py` rename a suspect HNSW segment directory so Chroma rebuilds into a fresh, empty index, and both leave that segment's `max_seq_id` row behind in `chroma.sqlite3`. Chroma reads the surviving row as "already synced through seq_id N" and replays nothing below it, so the rebuild the quarantine exists to force deterministically produces a **permanently short index**.

The reporter's three numbers, which should agree:

| | value |
|---|---|
| `embeddings` rows, metadata segment | 981 |
| `max_seq_id` for the VECTOR segment | 798 |
| vectors actually in the HNSW | 100 |

What makes it worth fixing rather than documenting is how quietly it fails. `collection.count()` keeps reporting the full total, because it reads the metadata segment. An unfiltered nearest-neighbour walk keeps returning plausible, relevant results from whatever nodes exist. Only a filtered query, which resolves specific ids and looks each one up, raises — which is why this cluster keeps arriving from the symptom end (#1665, #1035, #1082, #1398, #2373) rather than the cause end.

## The change

`_clear_segment_watermark(db_path, segment_id)` deletes the row, called from both paths **after** the rename succeeds.

Three properties, each with a cell:

- **Targeted.** The directory name is the segment UUID, so it deletes that one VECTOR row. The sibling METADATA row and every other collection are untouched, which is what keeps this away from the `max_seq_id` format hazards in #1134 / #1254.
- **After the rename, never before.** Clearing the watermark of a segment that is still live would make Chroma replay the whole queue into an index that already holds it, which is the HNSW bloat of #1046. A failed rename now leaves the watermark exactly where it was.
- **Best effort.** The rename is the safety action; the watermark is bookkeeping. A `chroma.sqlite3` that will not open logs and returns `False` rather than raising out of the quarantine. That is not defensive padding: the existing tests in this file write a text placeholder for `chroma.sqlite3`, and seeding a mutation that lets `sqlite3.Error` escape reddens `test_quarantine_catches_link_bloat_without_mtime_drift`, which predates this PR.

This stops new palaces reaching the state. It does not repair one already in it: as #2428 records, that needs the row cleared **and** `repair rebuild`, because the index is already short. The "minor, same area" note about the `.drift-` skip pattern catching a backup parked inside the palace is not addressed here.

## How to test

```
python -m pytest tests/test_hnsw_payload_health.py -v
```

Four new cells. The main one fails on `develop` with the reported shape — the quarantined segment's watermark survives next to the metadata one:

```
E  AssertionError: assert {'11111111-22...5555555': 798, '99999999-88...5555555': 981}
                      == {'99999999-88...5555555': 981}
```

A second covers `quarantine_invalid_hnsw_metadata`, since both paths rename and both had the bug. A third pins that a segment left **in place** keeps its watermark, so the fix cannot start clearing live segments. The fourth pins the unreadable-sqlite path.

Mutation-checked: four seeded defects — dropping the call from either path, deleting the whole table instead of one row, and letting an unreadable sqlite abort the quarantine — each caught by a named cell, none surviving.

## Checklist
- [x] Tests pass (`python -m pytest tests/ -v`) — 5387 passed, 54 skipped
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`) — and `ruff format --check` with the pinned 0.16.6


L4XB:
Pushed a correction to my own change, found while reading #2510.

The first commit opened the connection as `with sqlite3.connect(db_path) as conn:`. That context manager commits the transaction; it does **not** close the connection. This function runs during quarantine, which happens before `PersistentClient` opens the palace, and `_fix_blob_seq_ids` documents exactly what that costs:

> Opening a Python sqlite3 connection against a ChromaDB 1.5.x WAL-mode database leaves state that segfaults the next `PersistentClient` call.

So the version I first sent could have turned a quarantine into a segfault on 1.5.x. Both other pre-open readers in this file already guard against it — `_fix_blob_seq_ids` with `contextlib.closing`, the collection-type migration with an explicit `finally: conn.close()` — and this now matches them, with an explicit `commit()` since `closing` does not commit.

`test_the_watermark_connection_is_closed` pins it by counting `close()` through a tracking connection, so reverting to the bare `with` reddens it. The battery is five seeded defects now, all caught, none surviving.

Sorry for the round trip; better here than in a palace.

mvalentsev:
Clearing the watermark doesn't bring the vectors back.

I ran this branch's `quarantine_stale_hnsw` on a 2,500-record collection with mempalace's `hnsw:sync_threshold` (1000) and `hnsw:batch_size` (100), set off by an impossible `header.bin` count, then queried it from a fresh process. chromadb 1.5.7 (the `uv.lock` pin) and 1.5.9 give the same numbers:

| | VECTOR `max_seq_id` | records in the index | `where={"wing": "w1"}` |
|---|---|---|---|
| no quarantine | 2000 | 2500 | 5 results |
| develop | 2000 | 501 | `Error finding id` |
| this PR | deleted | 501 | `Error finding id` |

Those 501 are exactly the rows still in `embeddings_queue` (seq 2000-2500). Everything flushed before that was purged from the queue, so there's nothing left to replay it from, watermark or not: `d7` doesn't come back even for a query with its own embedding.

None of the new tests queries the collection after the quarantine. A filtered query there would catch this.

L4XB:
@mvalentsev you're right, and it goes beyond the purged case: I can't find any case where deleting the row changes a result.

I reran your setup on develop and on this branch: 2,500 records, `hnsw:sync_threshold` 1000, `hnsw:batch_size` 100, an impossible `header.bin` count, and every read in a fresh process. I also added a palace whose queue is never purged (`automatically_purge: false`). chromadb 1.5.7:

| | VECTOR `max_seq_id` after quarantine | in index | `where={"wing": "w1"}` | `d7` by its own embedding |
|---|---|---|---|---|
| purged queue, develop | 2000 | 501 | `Error finding id` | top-1 `d2452` (0.338) |
| purged queue, this PR | deleted | 501 | `Error finding id` | top-1 `d2452` (0.338) |
| unpurged queue, develop | 2000 | 2500 | 5 results | `d7` (0.0) |
| unpurged queue, this PR | deleted | 2500 | 5 results | `d7` (0.0) |

"In index" is the number of hits for an unfiltered query with `n_results` above the count. 1.5.4 and 1.5.9 give the same numbers. I also wrote 300 and then 1,200 more records after the quarantine. The index held 801 / 1,701 with a purged queue and 2,800 / 3,700 without, and the filtered results matched on both branches.

The unpurged rows explain it: develop rebuilds everything even though the row still says 2000.
- In 1.5.x a missing segment directory makes `LocalHnswSegmentReader::from_segment` return `UninitializedSegment` (`rust/segment/src/local_hnsw.rs:120-122` at the 1.5.7 tag).
- The backfill maps that to `hnsw_max_seq_id = 0` (`rust/log/src/local_compaction_manager.rs:175-177`) without reading `max_seq_id`, so it replays whatever is still in the queue.
- The row is only read once an `index_metadata.pickle` with ids exists (`local_hnsw.rs:149`), and the new index overwrites it on its first persist.

So, as you said, the purge decides the outcome. "This stops new palaces reaching the state" is wrong. The new cells pass only because they check the `max_seq_id` table, and a filtered-query cell can't separate this branch from develop, because no case behaves differently.

On getting the vectors back: after the purge they aren't anywhere in `chroma.sqlite3`. `embeddings` has no vector column, and `d7`'s queue row is gone. Only the documents and metadata survive (all 2,500). Re-embedding from those works on the quarantined palace, on both branches:
- `mempalace repair --yes` and `mempalace repair rebuild-index --yes` each rebuilt 2,500/2,500.
- Afterwards `where={"wing": "w1"}` returns 5 and `d7` is its own top hit again. (For this check I used a deterministic embedding function that reproduces the original vectors.)
- `python -m mempalace.repair rebuild` refuses the same palace at its divergence check and points to `--mode from-sqlite --archive-existing`.

I think this PR should be closed. #2515 still does something: on the purged palace it reports 1999 embeddings the queue can no longer replay (2,500 - 501), and on the unpurged one it says nothing. I'd move it onto develop without these two commits, add a cell that runs a filtered query after the quarantine, and fix its message: repair re-embeds from SQLite and doesn't use the vectors in the `.drift-*` directory. If you'd rather not have that warning either, I'll close #2515 too.

PR #2515, opened 2026-09-15
fix(chroma): say when quarantine leaves the WAL unable to rebuild

### LINKED ISSUE #2510: quarantine_stale_hnsw destroys flushed vectors when the WAL has been purged (sibling path to #1710)
**What happened?**

A 347k-drawer palace silently lost 47,099 embeddings — its entire first month,
contiguous — with no error. Search kept working over the surviving tail, so the loss
stayed invisible until a wing that should have had hundreds of hits returned almost none.

`quarantine_stale_hnsw` renamed a segment aside and let Chroma rebuild from the WAL, but
chromadb 1.5.x purges `embeddings_queue` by default, so the rebuild replayed only from the
purge watermark (seq 51,396 here). Everything below it was gone, and the rename counted as
success. From `hook.log`:

```
2026-07-08 15:39:36  Quarantined corrupt HNSW segment … (sqlite 694s newer than HNSW and integrity check failed); renamed to …c33345ec….drift-20260708-153936
```

**No SQLite metadata was lost** (0 orphans), so `repair --yes` recovered all 347k drawers
in ~3.5h. Nothing is permanently gone if the `.drift-*` dir survives. The defect is the
silence: meanwhile filtered queries errored with `Error finding id` while unfiltered KNN
returned confident results over two thirds of the corpus.

Same failure class as #1710, different guard. #1710 closed for
`quarantine_invalid_hnsw_metadata`; `quarantine_stale_hnsw`
(`mempalace/backends/chroma.py:455`) still quarantines against a WAL it never checks.

**What did you expect?**

Quarantine shouldn't be able to lose flushed vectors. Before renaming, compare the
`embeddings_queue` floor against the collection's embedding count; if the WAL can't
rebuild the collection, fail loudly and require an explicit `repair` instead of renaming.

**How to reproduce:**

1. Grow a palace until chromadb purges `embeddings_queue` behind the current segment.
2. Get `data_level0.bin`'s mtime >300s behind `chroma.sqlite3` and fail the
   `index_metadata.pickle` sniff test (continuous mining plus an interrupted shutdown).
3. Open the palace. The segment is renamed `*.drift-<stamp>` and the rebuilt index starts
   at the purge watermark. No error; `mempalace status` still shows the full SQLite count.

**Environment:**
- OS: macOS 26 (Darwin 25.6.0), Apple Silicon
- Python version: 3.14
- MemPal version: MemPalace 3.9.0, chromadb 1.5.9

<details>
<summary>Investigation</summary>

**Why the loss is contiguous.** After the rename, the rebuild replays `embeddings_queue`
from its current floor, and everything below had already been purged — so the gap is one
solid block at the bottom rather than scattered holes. mempalace never sets
`automatically_purge` itself (no occurrences in the 3.9.0 tree), which is the point: the
guard depends on a WAL whose retention it neither controls nor checks.

**3.9.0 narrowed the trigger but not the damage.** The current guard has a stage-2 gate
the version that ate these vectors lacked:

```python
if not payload_corrupt and _segment_appears_healthy(seg_dir):
    logger.info(... "flush-lag, not corruption. Leaving in place.")
    continue
```

Real improvement, but `os.rename(seg_dir, target)` at line 541 still proceeds with no check
that the WAL can reconstruct what it replaces. A segment that genuinely fails
`_segment_appears_healthy` on a purged-queue palace still loses everything below the
watermark.

**Where to start.** `_vector_segment_id` already reads `chroma.sqlite3` directly to avoid
opening a segfault-prone segment, so the access pattern for the pre-rename check exists in
this module. The `>=` relaxation from #1723 is present in both 3.8.0 and 3.9.0
(`chroma.py:1557`, identical); the remaining #1710 fix hasn't shipped (3.9.0 published
2026-08-31, two weeks before the close).

Full timeline, both `hook.log` firings and the 0-orphan verdict: my 2026-08-31 comment on
 #1710.

</details>

L4XB:
## What does this PR do?

Closes #2510.

This is no longer stacked on #2513, which is closed (clearing the watermark turned out to change no query result). The branch now carries only this change on top of `develop`.

Quarantine renames a segment so Chroma rebuilds it from `embeddings_queue`, and chromadb 1.5.x purges that queue. Everything written below the purge watermark is therefore absent from the rebuilt index, while its `embedding_metadata` row survives. So `collection.count()` keeps reporting the full total and an unfiltered nearest-neighbour walk keeps returning confident, relevant results over whatever remains. The reporter's palace lost its entire first month — 47,099 embeddings, contiguous — and only noticed when a wing that should have had hundreds of hits returned almost none.

Nothing is permanently lost while the drawers are in SQLite: `mempalace repair` re-embeds them, which is how the reporter recovered all 347k. The old index stays in the `.drift-*` directory, but nothing reads it back. What was missing is any signal that the shortfall happened, which is why this adds one rather than refusing the rename.

## The change

After a successful rename, count the collection's embeddings that sit below the queue's floor, and log an error naming the count and the command that re-embeds them.

The rename itself is untouched, deliberately. It is what stops Chroma opening a segment that can SIGSEGV before Python fallback code runs, and #2510 asks for a refusal instead — but since the data survives the rename and only the silence is harmful, refusing would trade a recoverable shortfall for a crash. If maintainers prefer the refusal, it is a one-line change at the same site and I will send it.

Three properties, each with a cell:

- **It measures rather than assumes.** The query is written against a real chromadb 1.5.7 palace, not against a guessed schema: `embeddings_queue(seq_id, topic, ...)` where `topic` is `persistent://default/default/<collection_id>`.
- **Unmeasurable is not "nothing lost".** A palace whose sqlite will not open returns `None`, and the caller emits nothing — the failure never reads as an all-clear.
- **The wiring is graded, not just the helper.** A cell drives `quarantine_stale_hnsw` against a real purged palace and asserts the error names the right count.

One thing that did not survive review by mutation: I first guarded the query with `typeof(e.seq_id) = 'integer'` to skip 0.6.x BLOB rows. A seeded mutant that removed it never reddened a cell, and the reason is that the guard cannot matter — SQLite's storage-class order puts every blob after every integer, so a BLOB `seq_id` can never satisfy `< <floor>`. The guard is gone and the comment now says why the comparison is sufficient on its own.

## How to test

```
python -m pytest tests/test_hnsw_payload_health.py -v
```

Five new cells build a real `PersistentClient` palace, purge `embeddings_queue` the way 1.5.x does, and assert the count (`4` of `8` below a floor of `5`), the intact case (`0`), the unmeasurable case (`None`), that a BLOB row is not counted, and that the quarantine itself reports the remainder.

Mutation-checked: three seeded defects — dropping the call, letting an unmeasurable palace read as zero, and inverting the floor comparison — each caught by a named cell, none surviving.

## Checklist
- [x] Tests pass: `tests/test_hnsw_payload_health.py` 17 passed; the `-k "chroma or hnsw or quarantine or repair"` selection 557 passed and 5 failed, the same five lock tests that fail on `develop` locally
- [x] No hardcoded paths
- [x] Linter passes (`ruff check .`) — and `ruff format --check` with the pinned 0.16.6


L4XB:
The Windows job was red on my own new test, and the cause was in the fixture rather than the code under test:

```
PermissionError: [WinError 5] Access is denied:
  ...\palace\767c5fb4-...  ->  ...\palace\767c5fb4-....drift-20260915-201957
  File "mempalace\backends\chroma.py", line 901, in quarantine_stale_hnsw
    os.rename(seg_dir, target)
```

`_palace_with_purged_queue` returned while chromadb still held the segment directory open. On POSIX that is invisible, since an open file does not stop its directory from being renamed, so the test passed on all four Linux/macOS jobs and only Windows saw it. Quarantine caught the OSError, logged it, and returned an empty list, so the assertion read `0 == 1` with no reason attached.

Fixed by releasing the palace in the fixture the way the repair path does: `_close_client(client)` (which drops the last System reference, so `system._running` goes False and the segment files are closed), then `_clear_chroma_system_cache()` and `gc.collect()`. The fixture's sqlite connection is now closed as well, since `with sqlite3.connect(...)` commits but does not close.

I also moved the captured log records into the `moved` assertion, so if that rename ever fails again the failure names the reason rather than just the count.

L4XB:
Reworked as proposed on #2513, which is now closed. Instead of a rebase, `develop` is merged in, and the last commit removes the watermark helper, its two call sites and its cells, so the diff against `develop` is only the #2510 warning (`chroma.py` +63, tests +129). The warning text used to say the vectors were preserved in the renamed directory, which was wrong: nothing reads that directory back. It now says to run `mempalace repair`, which re-embeds the drawers from SQLite. `test_hnsw_payload_health.py`: 17 passed; ruff 0.16.6 check and format are clean.

L4XB:
`test-windows` on 4e44bcd is red on one cell — `tests/test_dsh_plugin.py::test_dsh_plugin_node_suite` — and inside it on one of the suite's 44 Node subtests:

```
not ok 40 - a palace slower than the budget delays only the first assembly, not every step
  location: '.dsh-plugin/test/recall.test.mjs:235:1'
  error: The input did not match the regular expression /I am Atlas/. Input: ''
  duration_ms: 2255.2049
```

That cell is out of this PR's reach. The diff is two Python files (`mempalace/backends/chroma.py`, `tests/test_hnsw_payload_health.py`); `git diff --name-only 25203ed 4e44bcd -- .dsh-plugin tests/test_dsh_plugin.py` is empty, and the Node suite runs no Python at all — `test/host.mjs` fakes `ctx.subprocess.spawn`, so nothing in it imports or executes `mempalace`. On that same Windows job all 17 cells of `tests/test_hnsw_payload_health.py` passed and coverage came in at 82%.

The subtest is a wall-clock race with about 300 ms of slack: the faked `wake-up` is delayed 1500 ms, the first assembly spends `firstAssemblyBudgetMs` (300 ms), and the test then sleeps exactly `tick(1500)` before requiring the memory to have landed — while the read only spawns after `projectWing()` and `executableFor()`. Replaying that subtest against the merge base (25203ed, none of this PR applied) and observing the faked child's exit 300 ms late reproduces the CI message verbatim:

```
lag=   0ms  third assembly at t+1806ms  -> PASS
lag= 250ms  third assembly at t+1807ms  -> PASS
lag= 300ms  third assembly at t+1805ms  -> FAIL  The input did not match the regular expression /I am Atlas/. Input: ''
lag= 500ms  third assembly at t+1803ms  -> FAIL
```

The runner was that slow: the identical Node suite took 12.50 s on this job against 6.63 s on the merge base's own green `test-windows` run (35089512865), and the whole suite 1008 s against 604 s. The Windows step only retries the two Chroma HNSW messages, so a timing flake here fails the job on its first try; the same wrapper was Windows-only red once before, on the branch that introduced the plugin (run 34933606262).

A re-run should clear it. If it is worth hardening, I can send a separate PR that makes that subtest wait for the memory to land instead of sleeping a fixed 1500 ms.

PR #2517, opened 2026-09-15
chore(release): 3.10.0

### LINKED ISSUE #2500: Writable HTTP startup exits 2 while a peer holds the writer lease, so the Restart=always from #2203 becomes an unbounded restart loop

igorls:
## Summary

Release prep for **3.10.0**, following the shape of `chore(release): 3.9.0` (`434f00b5`).

**Version bump to 3.10.0** in every source:
- the seven that `version-guard.yml` checks: `mempalace/version.py`, `pyproject.toml`, `.claude-plugin/marketplace.json`, `.claude-plugin/plugin.json`, `.codex-plugin/plugin.json`, `.dsh-plugin/package.json`, `integrations/openclaw/SKILL.md`
- `Cargo.toml` (workspace package) and the three workspace crates in `Cargo.lock`, so the native wheels built by `release-native.yml` are no longer stamped `3.9.0`
- `uv.lock` (`mempalace` editable entry) and the README version badge

**Changelog.** `[Unreleased]` is filed under `## [3.10.0] — 2026-09-15` with a one-line intro and an **Upgrade notes** section for the breaking changes (`mempalace rules --agent` removed, `get_collection()` name check, newest-first logstream listings without a cursor, XDG paths for new installs). Compare links updated.

**Rust benchmark figures dropped** from the changelog entry, the README table and `crates/README.md`. They came from an early Windows run that predates the correctness hardening and were never rerun; the docs now point to `mempalace-native bench` for measuring on your own data.

## Merge order

1. #2516 (writer-lease wait for `mempalace serve`, closes #2500): its changelog entry lands under `[Unreleased]` and needs to be moved into 3.10.0 here, or this PR rebased after it.
2. This PR.
3. `develop` → `main`, then tag `v3.10.0` on `main` and publish the GitHub release (PyPI upload waits for approval on the `pypi` environment).

## Verification

- All ten version sources read `3.10.0`; no `3.9.0` remains outside changelog history, tests and a measurement note in `.dsh-plugin/README.md`.
- `pytest tests/test_version_consistency.py tests/test_readme_claims.py tests/test_cli_update.py tests/test_update_awareness.py tests/test_mcp_proxy.py`: see the comment below.
- The `release-native.yml` dry run on `develop` (run 34998158738) built all 5 wheels and 3 CLIs and skipped `publish-assets`; the Windows wheel installed in a fresh venv and `tests/test_rust_exact_backend.py` passed 27/27 against it.


igorls:
Local results on head `7b7a782` (Windows, Python 3.12 venv):

- `pytest tests/test_version_consistency.py tests/test_readme_claims.py tests/test_cli_update.py tests/test_update_awareness.py tests/test_mcp_proxy.py` → **84 passed**.
- Version sources: `mempalace/version.py`, `pyproject.toml`, `.claude-plugin/marketplace.json`, `.claude-plugin/plugin.json`, `.codex-plugin/plugin.json`, `.dsh-plugin/package.json`, `integrations/openclaw/SKILL.md`, `Cargo.toml`, `Cargo.lock` (3/3 crates), `uv.lock` and the README badge all read `3.10.0`.
- The commit touches exactly these 13 files, with LF line endings (checked on the committed blobs).

igorls:
Superseded by #2518. This branch (`release/3.10.0-prep`) is covered by the "Protect release branches" ruleset (`refs/heads/release/*`: changes only through a pull request), so the Copilot review fixes could not be pushed here. #2518 carries both commits: the 3.10.0 prep and the review fixes. The `release/3.10.0-prep` branch cannot be deleted without ruleset bypass; it can be removed by an admin, or left, since it points at a commit already contained in #2518.

PR #2533, opened 2026-09-16
fix(convo): preserve source state while transcripts grow

ariannamethod:
## What does this PR do?

Fixes two ways conversation mining can skip newly appended messages:

- The miner reads a transcript, then samples its mtime later when filing. If
  the source grows between those operations, old content receives the newer
  timestamp and the next pass skips the appended text.
- The existing 1 ms mtime tolerance can consider a changed transcript current,
  even when the append happens between completed mining passes.

The reader compares `(st_dev, st_ino, st_size, st_mtime_ns)` from the open file
descriptor before and after reading. A changed fingerprint defers the source
until a later pass, preserving existing records. The verified state is carried
through drawer and registry writes and compared exactly by both conversation
currentness checks. This is a stat fingerprint, not a content hash.

`source_mtime` remains present for compatibility. Existing conversation records
without a fingerprint are reprocessed once, so the first pass after upgrading
may reindex existing transcripts. Other callers retain the default mtime-based
behavior.

The branch also preserves `source_dir_ino` from #2368. That directory identity
protects sync's deletion checks; the file fingerprint here protects mining
freshness. Conversation drawers carry both independently.

This addresses the source-consistency issue raised by @fatkobra in
[#1959](https://github.com/MemPalace/mempalace/pull/1959#pullrequestreview-4720416239).
An audit independently reproduced the failures on `develop` at `25203ed6`;
the stable fingerprint approach was already proposed in that review. This PR
implements that correctness fix separately from the broader feature work in #1959.

## How to test

```sh
python -m pytest tests/test_convo_source_snapshot.py -q
python -m pytest tests/ -q --ignore=tests/benchmarks \
  --cov=mempalace --cov-report=term-missing --cov-fail-under=80
ruff check .
ruff format --check .
```

Five original regression cases failed against unchanged upstream: append after
reading for ordinary, short, and empty transcripts; a 100,000 ns mtime change
between passes; and modification during reading. Previously, appended text was
missing after the next pass. With the fix, the next stable pass imports it.

The final seven cases also cover atomic replacement with unchanged size/mtime
and a one-time upgrade of legacy metadata. The atomic-replacement case also
checks that the file fingerprint changes while the recorded directory identity
remains present and unchanged. The during-read case checks that
existing IDs, documents, and metadata remain unchanged on the deferred pass.

Latest local results on macOS arm64 / Python 3.14.4, after merging `develop`
at `36ec72f`:

- Regression tests: **7 passed**.
- Full suite with the command above: **5502 passed, 56 skipped**.
- Coverage: **82.80%**, above the 80% gate.
- Ruff lint and format checks: passed.

## Checklist

- [x] Tests pass (commands and local results above)
- [x] No hardcoded paths

MemPalace's approach to memory resonates with our work on Arianna Method.


ariannamethod:
Updated the branch with `develop` at `36ec72f` in 5ef9053. The conflict with #2368 is resolved; GitHub now reports `MERGEABLE` (no merge conflicts).

The two changes protect different things: `source_dir_ino` records the parent directory identity for sync's deletion checks, while `source_fingerprint` records the verified file snapshot for conversation-mining freshness. The resolution retains both fields and keeps `source_mtime` tied to the read snapshot.

Extended the atomic-replacement regression to check that the file fingerprint changes while the directory identity stays present and unchanged.

Validation on macOS arm64 / Python 3.14.4:
- Snapshot, conversation-miner unit, source-identity and sync tests: **194 passed, 2 skipped**.
- Full suite: **5502 passed, 56 skipped**, coverage **82.80%** (80% gate).
- `ruff check .` and `ruff format --check .`: passed.

The PR description now includes the updated local test results.

Issue #2347, opened 2026-08-23
mempalace_drawers is the real collection name; any other name silently writes data that can't be searched


JyslaFancy:
**Title:** `mempalace_drawers` is the real collection name; any other name silently writes data that can't be searched

**Current behavior:** Programmatic access must use collection name `mempalace_drawers`. Passing `drawers` or `mempalace` succeeds but writes to a collection MemPalace's own search/CLI never reads, so data is silently invisible. Related wrong-collection bug #1244 (`_compressed` vs `_closets`).

**Proposed change:** Alias or validate the collection name on `get_collection`, or document it prominently in the Python API docs so mismatched names fail loudly instead of storing ephemeral data.

JyslaFancy:
This appears fixed on `main`. #2446 (`c0c9615`, "fix(palace): fail loudly on a
mismatched get_collection name") added `CollectionNameMismatchError` in
`mempalace/palace/collection.py`, raised from the `get_collection` wrapper when the
requested name is neither the configured drawers collection nor
`mempalace_closets`. The docstring cites this issue directly.

So `get_collection(palace, "drawers")` now raises instead of silently creating an
orphan collection, which is the "fail loudly" behaviour the issue asked for. No
alias was added, but validation covers the reported failure mode.

One residual note, not a request to reopen: the check lives in the wrapper only. A
caller who goes straight to a backend object (`backend.get_collection(...)`) still
gets no validation. That seems reasonable if the wrapper is the supported surface,
it may be worth one line to that effect in the Python API docs.

Closing this from my side unless there is a case the check misses.

---

igorls:
Yes — this is solved on `develop`.

[#2446](https://github.com/MemPalace/mempalace/pull/2446) merged 2026-09-12 (`c0c9615`). `get_collection` now raises `CollectionNameMismatchError` before touching the backend when the name is neither the configured drawers collection nor `mempalace_closets`. `get_collection(..., \"drawers\")` / `\"mempalace\"` no longer create an orphan store.

GitHub did not auto-close this: the squash merge left `closingIssuesReferences` empty even though the PR title said `fixes #2347`. Closing now.

The original reporter already confirmed the same in https://github.com/MemPalace/mempalace/issues/2347#issuecomment-5655736839. Residual note from that comment still stands: a caller that goes straight to `backend.get_collection(...)` is outside the wrapper and is not validated.

PR #2534, opened 2026-09-16
fix(mcp): parse the server's flags in main(), not when mcp_server is imported (#2528)

### LINKED ISSUE #2528: mcp_server parses sys.argv at import time: mempalace-light-mcp --help shows the full server's options, and any importer's argv is consumed
**What happened?**

`mempalace/mcp_server/_guards.py` calls `parse_known_args()` on `sys.argv` at **module scope** (line 73, `_args = _parse_args()`), and then acts on the result — setting `os.environ["MEMPALACE_PALACE_PATH"]`, selecting a backend, and so on. Because `mempalace.mcp_server` is an ordinary importable module, this runs on *import*, not on entry, so it reads the argv of whatever process imports it.

Two visible consequences:

**1. `mempalace-light-mcp --help` documents options the light server rejects.**

```
$ mempalace-light-mcp --help
usage: mempalace-light-mcp [-h] [--palace PATH] [--backend NAME]
                           [--transport {stdio,http}] [--host HOST]
                           [--port PORT] [--tls-cert PATH] [--tls-key PATH]
                           [--read-only]

MemPalace MCP Server
```

That is the **full** server's parser (note `description="MemPalace MCP Server"` and `metavar="PATH"`). The light server's real parser is in `mcp_light_server.main()` and accepts only four options:

```
$ mempalace-light-mcp --transport http --host 127.0.0.1 --port 8775
usage: mempalace-light-mcp [-h] [--palace PALACE] [--collection COLLECTION]
                           [--backend BACKEND] [--read-only]
mempalace-light-mcp: error: unrecognized arguments: --transport http --host 127.0.0.1 --port 8775
```

`mcp_light_server.py:38` does `from . import mcp_server`, which executes `_guards.py` at import; argparse's `-h` action prints that parser's help and `sys.exit(0)`s before `main()` ever builds its own parser. So `--help` can never reach the light server's parser.

The practical cost: the help text advertises an HTTP transport the light server does not have. Following it is the natural way to deploy the light server beside an existing HTTP one, and it fails only at runtime.

**2. Importing the library consumes the importing program's argv.** This is the part I think matters more.

```python
# import_test.py
import sys, os
print('MEMPALACE_PALACE_PATH before:', os.environ.get('MEMPALACE_PALACE_PATH'))
from mempalace import mcp_server   # library import, not a CLI entry point
print('MEMPALACE_PALACE_PATH after :', os.environ.get('MEMPALACE_PALACE_PATH'))
```

```
$ python import_test.py --palace /tmp/not-my-palace
MEMPALACE_PALACE_PATH before: None
MEMPALACE_PALACE_PATH after : /tmp/not-my-palace
```

An unrelated host program that happens to have its own `--palace` flag silently redirects MemPalace to a different palace, just by importing it. And `--help` is hijacked outright:

```
$ python import_test.py --help
usage: import_test.py [-h] [--palace PATH] [--backend NAME]
                      [--transport {stdio,http}] [--host HOST] [--port PORT]
                      [--tls-cert PATH] [--tls-key PATH] [--read-only]
```

The host program prints MemPalace's options under its own name and exits 0 — its own `--help` never runs.

Since `parse_known_args()` tolerates unknown flags, this is silent in every case except `--help`: there is no error, and nothing indicates that argv was read.

**What did you expect?**

Argument parsing and its side effects belong behind an entry point, not at module scope. Concretely: move `_parse_args()` and the `os.environ` / backend mutations that follow it into the `main()` of the full server, and have `mcp_light_server.main()` parse its own argv only. Importing `mempalace.mcp_server` should not read `sys.argv`, should not mutate `os.environ`, and should not be able to terminate the importing process.

`mempalace-light-mcp --help` should then print the four options the light server actually accepts — or, if an HTTP transport is intended for the light server too, it could gain the `--transport/--host/--port` trio its help currently promises. From the outside the latter looks useful: the light server can only be spawned as a local stdio child today, which rules it out for any client that reaches the palace over the network rather than through the filesystem.

**How to reproduce:**

1. `mempalace-light-mcp --help` — observe the full server's option list, including `--transport`.
2. `mempalace-light-mcp --transport http --port 8775` — observe `unrecognized arguments`.
3. Save the `import_test.py` above and run it with `--palace /tmp/not-my-palace`, then with `--help`.

**Environment:**
- MemPalace 3.10.0, chromadb 1.5.9, backend `chroma`
- Python 3.13.5
- Debian 13 (trixie), Linux 6.18.34 aarch64 (Raspberry Pi 4B)

<details>
<summary>Where it comes from</summary>

`mempalace/mcp_server/_guards.py`:

```python
    args, unknown = parser.parse_known_args()      # ~line 67
    if unknown:
        logger.debug("Ignoring unknown args: %s", unknown)
    return args


_args = _parse_args()                              # line 73 — module scope

if _args.palace:
    os.environ["MEMPALACE_PALACE_PATH"] = os.path.abspath(_args.palace)
if _args.backend:
    backend_name = str(_args.backend).strip().lower()
    ...
```

`mempalace/mcp_light_server.py:38`: `from . import mcp_server`, which triggers the above. The light server's own parser lives at `mcp_light_server.py:1007-1016` and is never reached when `-h/--help` is present.

I found this while trying to deploy `mempalace-light-mcp` alongside three existing `mempalace-mcp --transport http` systemd units, following its `--help`. The 3-tool surface itself works well — 9.9 KB of schema against ~41.6 KB, and it started cleanly against a palace whose HTTP server was live, taking no writer lease.

</details>

mvalentsev:
## What does this PR do?

Fixes #2528.

`mcp_server/_guards.py` parsed `sys.argv` at import, so any program that imports the package had its command line parsed as server flags. On develop, `mempalace-light-mcp --help` prints the full server's options, a host program exits 2 on its own `--port abc`, and `mempalace-light-mcp --palace <dir>` fails with `AttributeError`.

The import now only builds the defaults, and `_apply_server_flags()` applies the flags. `main()` calls it, and so do three paths that relied on the import and never run `main()`: the stdio proxy's local fallback, `mempalace-light-mcp`, and the daemon's `mcp_tool` jobs. A spawned daemon kept its knowledge graph beside the palace only because the import saw its `serve --palace`.

One behavior change: `mempalace daemon start --foreground` without `--palace` now also writes `mcp_tool` knowledge-graph facts beside its palace, not to `~/.mempalace/knowledge_graph.sqlite3`.

Not fixed: `mempalace-light-mcp --collection` still raises `AttributeError`. `get_collection()` refuses every name except the drawers and closets collections (#2347).

## How to test

```
uv run pytest tests/test_mcp_server_package.py tests/test_mcp_proxy.py tests/test_mcp_light_server.py tests/test_daemon.py tests/test_service_palace_env.py -v
```

All new tests but one fail on develop. That one checks that flags survive a module reload.

## Checklist
- [x] No hardcoded paths


audrum:
Thanks for picking this up, can confirm this matches the root cause found while filing #2528 (`_guards.py` parsing `sys.argv` at module scope via `_parse_args()` at line 73, executed on import rather than gated behind `main()`).

Moving that into `_apply_server_flags()` called from `main()` (plus the three call sites that relied on the import side-effect: stdio proxy fallback, `mempalace-light-mcp`, daemon `mcp_tool` jobs) looks like the right fix, and it should also close the "importing the library hijacks the importer's argv/`--help`" half of the report, not just the `mempalace-light-mcp --help` symptom.

One thing worth calling out for anyone tracking this against #2528: this PR fixes the parsing bug but doesn't add an HTTP transport to `mempalace-light-mcp`, it's still stdio-only after this merges. Not a gap in the PR (out of scope for #2528 as filed), just flagging it so it's not read as "light server now supports `--transport http`."

igorls (approved):
Verified locally: on develop, `mempalace-light-mcp --help` prints the full server's options, an importer's `--palace` leaks into `MEMPALACE_PALACE_PATH`, `--port abc` exits the host with 2, and `mempalace-light-mcp --palace` raises `AttributeError`. All four are fixed at this head. 16 of the new tests fail on develop's source and pass here; the PR tests plus `tests/mcp` gave 578 passed, and ruff is clean.

I checked that deferring the flags is safe: `_config.palace_path` and `_config.backend` are lazy, the KG cache resolves its path per call, and every other entry point (`python -m`, `serve`, hooks, Hermes, the daemon's `run_server`) either goes through `main()` or gets its palace from the environment. The foreground-daemon KG change matches what spawned daemons already did.

Non-blocking follow-up: in the proxy fallback, a refused `--backend` makes `load()` raise on every request, so the client gets no JSON-RPC response. Develop fails the same way at import, but a proper error response would be nicer. Thanks!

PR #2537, opened 2026-09-17
fix(mcp): stop reloading the HNSW index after the server's own writes (#2535)

### LINKED ISSUE #2535: mcp_server._get_client has no own-write re-stamp: every add_drawer and check_duplicate reloads the HNSW index (#2307 covered only ChromaBackend)
## What happens

#2307 stopped `ChromaBackend._client` invalidating its cache on the backend's own writes,
by re-stamping the freshness stat once an operation finishes. `mcp_server._get_client` keys
on the same `chroma.sqlite3` mtime but never re-stamps, and `_get_collection` calls it on
every tool call (v3.9.0 `mempalace/mcp_server.py:1604`). So after the server's own write,
the next MCP tool that goes through `_get_collection` sees its own footprint as an external
change, runs `_force_chroma_cache_reset()` and reopens, and the next vector query reloads
the index from disk.

Search is unaffected: it goes through the backend path that #2307 fixed.

## Measured

mempalace 3.9.0 (pip, clean venv), chromadb 1.5.9, Python 3.11, macOS 26 arm64, chroma
backend, a scratch palace of 40 mined notes. One in-process server driven through
`handle_request`, counting chromadb `System` creations per call (a new System means the HNSW
index is read from disk again):

| Call | Reloads on 3.9.0 |
|---|---|
| `status` | 0 |
| `search` (first) | 1 (cold open, expected) |
| `search` again | 0 |
| `check_duplicate` | **1** |
| `search` | 0 |
| `add_drawer` | **1** |
| `search`, `search` | 0 |
| `add_drawer` | **1** |
| `check_duplicate` | **1** |
| `list_drawers` | 0 |
| `search` | 0 |
| *(another process mines)* | — |
| `search` | 1 (correct: it picks up the new drawers) |

**6 reloads in 14 calls.** Four of them are the server reacting to its own writes.

Cost per reload is palace-sized. On our 151k-drawer palace a reload takes about 2.5 s
(2.9 s for a search that reopens, against 0.2 s warm). That timing is from our own 3.6.0
install; we did not repeat it against 3.9.0 on a palace that size.

## Reproduction

```python
# after: pip install mempalace==3.9.0, and a palace with some drawers in wing "seed"
import mempalace.mcp_server as m
from chromadb.api.shared_system_client import SharedSystemClient

created = []
orig = SharedSystemClient._create_system_if_not_exists.__func__
def counting(cls, identifier, settings):
    if identifier not in cls._identifier_to_system:
        created.append(identifier)
    return orig(cls, identifier, settings)
SharedSystemClient._create_system_if_not_exists = classmethod(counting)

def call(tool, args):
    before = len(created)
    m.handle_request({"jsonrpc": "2.0", "id": 1, "method": "tools/call",
                      "params": {"name": tool, "arguments": args}})
    print(f"{tool:<26} reloads={len(created) - before}")

q = {"query": "whatever is in your palace", "wing": "seed", "limit": 3}
call("mempalace_search", q)          # 1, cold open
call("mempalace_search", q)          # 0
call("mempalace_add_drawer", {"wing": "probe", "room": "own", "content": "x", "added_by": "p"})  # 1
call("mempalace_check_duplicate", {"content": "x"})  # 1
```

## Suggested direction

The same shape as #2307, on the MCP path: after the server's own write completes, and after
`_get_collection` has opened a collection, re-stat `chroma.sqlite3` into `_palace_db_inode` /
`_palace_db_mtime`, so a later difference is genuinely another process's write. The same
trade-off #2307 documents applies: an external write landing inside one of the server's own
operations is absorbed until the next change.

If you would rather not rely on mtime at all on this path, the signal we ended up using
downstream is the write-ahead log high-water mark (`SELECT max(seq_id) FROM
embeddings_queue`): reads never move it, and writes from any process always do. That needs
another connection to `chroma.sqlite3`, so it interacts with #2302 / #2506; on a palace in
WAL mode it should go through whatever `_inproc_sqlite` ends up being.

## What this report does not claim

- Searches on 3.9.0 are fine, and do see other processes' writes.
- Not tested through the hub/proxy path (`mempalace serve`), only a stdio server.
- The palace here is small; the reload count is the finding, not the timing.

---

*Found while keeping a 3.6.0 install alive on a shared palace; the measurements above are
from a clean 3.9.0 venv with no local patches. Investigation and write-up by Claude Opus 5
running in Claude Code on my machine.*

mvalentsev:
## What does this PR do?

Fixes #2535.

The MCP server kept its own ChromaDB client next to the one `ChromaBackend` caches for search and mining, and never re-stamped it after its own writes. So `mempalace_add_drawer`, `mempalace_check_duplicate` and `mempalace_diary_write` reloaded the HNSW index from disk. Reopening that client also dropped the backend's stat record. A search after another process's `mempalace mine` could then attach to the server's older in-memory index and miss the new drawers until the next reopen, the symptom of #608.

`_get_client()` now returns the backend's client, and the server's collection opens and writes re-stamp its record, as the backend's own have since #2307. Sharing the client took three more changes. The retry in `_get_collection` closes the client in the backend's cache. The backend re-arms the stale-HNSW quarantine (#1573) whenever it drops a stat record, in `close_palace` or a drain, since the next open has nothing to compare. `mempalace_update_drawer` purges closets after the drawer write, because the purge's open can rebuild the shared client.

Thanks @christianlappin for the report. Re-stamping only the server's stamp, as the issue suggests, cut the reloads in six add, check and search cycles from 12 to 5. A search writes the file when it opens its collections, so the server's next call still reloaded.

## How to test

```bash
uv run pytest tests/mcp/test_protocol.py tests/mcp/test_write_tools.py tests/test_backends.py
```

Six of the new test cases fail on develop. A `mempalace-mcp` stdio server:

| | develop | this branch |
|---|---|---|
| index reloads in 30 add, check and search cycles, 60k drawers | 60 | 0 |
| median `mempalace_add_drawer` in those cycles | 303-351 ms | 83-84 ms |
| new drawers from another process's `mempalace mine` that search finds after a check | 0 of 5 | 5 of 5 |

## Checklist
- [x] No hardcoded paths


mempalace/backends/chroma.py
```diff
@@ -3325,6 +3331,9 @@ def close_palace(self, palace) -> None:
             return
         _close_client(self._clients.pop(path, None))
         self._freshness.pop(path, None)
+        # Without its stat record the next open cannot see a change on disk, so
+        # re-arm the HNSW pre-checks for it (#1573).
```
igorls: **[suggestion]** New and rewritten comments cite GitHub issue numbers. That includes this note, `_adopt_opened_collection`'s note in `mempalace/mcp_server/_session.py` (around the own-write re-baseline comment), the rewritten probe comment in the same file, and the new test docstrings in `tests/mcp/test_protocol.py` and `tests/test_backends.py`. The changelog entry matches the existing unreleased style and is fine.

**Suggestion:** Drop the issue numbers from those comments and docstrings and name the behavior (own-write re-stamp, quarantine re-arm, capacity probe before reopen) instead.

igorls (approved):

mvalentsev:
Rebased onto develop and kept your commit. The conflict with #2506 was in `close_palace`: the quarantine re-arm now runs inside `palace_db_lock`, with the close.

The rebase brought one more commit. #2572's `mempalace_delete_drawers` purges each id's closets through the backend, and with the shared client a rebuild there closed the handle the loop kept, so the ids after it failed with `'RustBindingsAPI' object has no attribute 'bindings'` and kept their drawers and closets. It now takes the collection again when the client its handle came from has been replaced. `sync_palace` had the same shape: it asked which sources were empty after opening the closets, and now asks before. Tests for both fail without the change.

Left as is: a search whose closets open rebuilds the client skips the neighbour fetch for its closet hits in that one call. I can send a fix for that separately.

PR #2541, opened 2026-09-17
fix(cli): resolve the hooks write-routing scope for shell-hook-triggered mine

AmirF194:
This is the same shell-hook write-routing gap you found, but the mechanism has moved since 3.8.0: #2033 landed general policy consultation for CLI writes after you filed this, so `cli.py` does call into the resolver now (your grep against `cli.py` predates that; the call lives in `cli_write_routing.py`). What's left is that it always resolves the `cli` scope, never `hooks`, so your exact repro config (`{"default": "direct", "hooks": "require"}`) still falls through to `write_routing.default` for every shell-hook-triggered mine.

Added `MEMPALACE_CLI_ROUTING_SCOPE`, which the five shell hooks now set to `hooks` before shelling out to `mempalace mine`. `resolve_cli_write_routing` reads it instead of the hardcoded `"cli"` literal, falling back to `cli` when it's unset or unrecognized, so an interactive `mempalace mine` from a terminal is unaffected.

Didn't reproduce the lease-contention numbers from your report (needs a live daemon plus concurrent sessions), but confirmed the routing decision directly against a real `MempalaceConfig` with your exact config: before this change, an unscoped mine resolves `use_daemon=False`; with the env var set the way the hooks now set it, it resolves `use_daemon=True, reason=daemon-auto-start`.

Tests: unit tests on the scope resolution (env unset, set, case-insensitive, an unrecognized value falls back to `cli`, an explicit `--daemon`/`--direct` bypasses scope entirely), subprocess tests driving the save and precompact hooks end to end against a stub `mempalace`, and a source-level check that every `mempalace mine` call site across all five hook scripts carries the env var, so a new call site or hook variant can't silently regress it back to `cli`.


mempalace/cli_write_routing.py
```diff
@@ -112,7 +132,7 @@ def resolve_cli_write_routing(
         )
         explicit = True
     else:
-        resolved = MempalaceConfig().resolve_write_routing("cli")
+        resolved = MempalaceConfig().resolve_write_routing(_cli_routing_scope())
         explicit = False
 
     decision = choose_write_route(
```
igorls: **[bug]** Switching the config scope does not apply hook caller rules. The route is still chosen with `daemon_can_start=True` and `daemon_available=False`. `choose_write_route` then returns `target=daemon`, `auto_start_daemon=True`, `reason=daemon-auto-start` for both `prefer` and `require`, and the blocked branch stays unreachable. Hook writes are required to pass `daemon_can_start=False`: `require` must skip the write when no daemon is already up and must not cold-start, and `prefer` must fall back to direct. PreCompact mines are synchronous, so this auto-start runs on the compaction path. A daemon that is already up is reused by `ensure_client`, which covers the lease-contention repro, but nothing here fail-closes when the hook is not allowed to start one. The new tests only assert the scope string under `WriteRoutingPolicy.DIRECT`.

**Suggestion:** When the resolved scope is `hooks`, probe an already-running daemon and call `choose_write_route` with `daemon_can_start=False`. A blocked decision must exit non-zero with no direct mine and no `start_daemon`. Add tests that hooks `require` with no daemon is blocked and hooks `prefer` with no daemon stays direct.

mempalace/cli_write_routing.py
```diff
@@ -22,6 +23,25 @@
...
+    Defaults to ``cli``. The portable shell hooks under ``hooks/`` shell out
+    to ``mempalace mine`` instead of writing in-process through
+    :mod:`mempalace.hooks_cli`, so they set
+    ``MEMPALACE_CLI_ROUTING_SCOPE=hooks`` before invoking it: without this,
+    a ``write_routing.hooks`` policy configured to keep hook-triggered
+    writes on the daemon is silently ignored for every shell-hook mine,
+    which instead reads ``write_routing.cli`` (default direct).
+    """
+
+    value = os.environ.get(_CLI_ROUTING_SCOPE_ENV, "cli").strip().lower()
```
igorls: **[bug]** `MEMPALACE_CLI_ROUTING_SCOPE` is read for every `resolve_cli_write_routing` caller, not only shell-hook mines. Any normal invocation can set it to `hooks` and resolve `write_routing.hooks` (including `MEMPALACE_HOOK_WRITE_ROUTING` and legacy `hooks.daemon`) instead of `write_routing.cli`. With the documented `write_routing.cli=require` and hooks left unset, the hooks scope falls through to `direct`, so `MEMPALACE_CLI_ROUTING_SCOPE=hooks mempalace mine` is a direct writer beside the daemon. An unrecognized value, including a typo of `hooks`, silently selects `cli`, which is the fail-open the policy parser refuses for a misspelled `require`. Nothing checks that the caller is a hook. `--direct` is a separate, explicit override and does not make this scope switch safe.

**Suggestion:** Do not honor a user-set environment variable as the hooks scope on the CLI entrypoint. Route shell-hook mines through the in-process hooks path (as `hooks/mempal_session_end_hook.sh` does with `mempalace hook run`), or otherwise make `hooks` unreachable from an ordinary CLI argv and environment. If a scope override remains, an unrecognized value must fail closed rather than select `cli`.

tests/test_cli_write_routing.py
```diff
@@ -211,6 +212,81 @@ def test_parser_flags_are_mutually_exclusive():
...
+        self.requested_scopes: list[str] = []
+        self.palace_path = "/tmp/palace"
+
+    def resolve_write_routing(self, scope: str) -> ResolvedWriteRoutingPolicy:
+        self.requested_scopes.append(scope)
+        return ResolvedWriteRoutingPolicy(policy=self._policy, source="test")
+
+
+class TestCliRoutingScopeSelection:
+    """Issue #2326: the portable shell hooks shell out to ``mempalace mine``
```
igorls: **[suggestion]** The new class docstring cites an issue number. `tests/test_shell_hooks_cli_routing_scope.py` repeats it in the module docstring and in the failure strings. Those are ephemeral issue numbers in source comments and test text.

**Suggestion:** State the routing contract without the issue number.

igorls (changes requested):

AmirF194:
Thanks for pushing the daemon fix and the docstring cleanup directly, both land exactly right: fail-closed on an unrecognized scope, no cold start under `require`, fallback to direct under `prefer`.

The env var point stands on its own though. Trusting `MEMPALACE_CLI_ROUTING_SCOPE` from any caller's environment is the wrong place to put this decision, and moving hook-triggered mines onto the in-process path `session-end` already uses is a real redesign of five shell scripts plus whatever `hook run` needs to accept in place of `mine`'s mode and wing flags, not something to patch blind in this PR.

Land this with the daemon fix as is, and I will open a follow-up issue for the env var and take the `hook run` migration there, unless you would rather fold it in here yourself.

PR #2546, opened 2026-09-18
feat: bulk drawer operations. Mine room override, move_drawers, and delete_drawers

edwardmakesthings:
## What does this PR do?

Two related additions, both about targeting drawers more precisely.

**Call-level `room` override for `mine`.** A mine run can now target a specific room without editing the configured default. Threaded through `cli/parser.py`, `cli/cmd_mine.py`, `service.py` and `miner.py`.

**Bulk `move_drawers` and `delete_drawers` tools.** Both accept either explicit `drawer_ids` or a `wing`/`room` scope, and both refuse a call with neither. If unscoped, they would match every drawer in the palace, which is never what a caller means.

The two shapes differ because the backend APIs do:
- A **delete** pushes down as a single `where` clause, so neither the SQLite variable limit nor the 10k `get()` truncation applies.
- A **move** needs explicit ids, so it enumerates and then batches — one `col.update` per `_BULK_DRAWER_BATCH` (500) rows, never one call per drawer. The fan-out form can wedge a palace. While holding the mine lock, chromadb's Rust upsert blocks when it has with no timeout of its own, so N concurrent round trips can turn a slow move into a palace-wide outage.

Both report **logical drawer counts**, not physical rows, so the returned `match_count` agrees with `list_drawers` for the same scope on a palace with chunked drawers.

## How to test

```
python -m pytest tests/mcp/test_write_tools.py tests/test_mcp_mine.py tests/test_miner.py tests/test_cli_write_routing.py tests/test_daemon.py -v
```

Manual:
1. `mempalace_mine(..., room="some-room")` — confirm drawers land in that room, not the configured default.
2. `mempalace_move_drawers(drawer_ids=[...], wing="dest")`, then `mempalace_delete_drawers(wing="W", room="R")` — confirm the returned count matches `list_drawers` for the same scope.
3. Call either with no `drawer_ids` and no `wing`/`room` — confirm it is refused rather than matching the whole palace.

## Checklist
- [x] No hardcoded paths


edwardmakesthings:
Converted to draft because this actually relies in part on changes from #2545 to land first.

PR #2555, opened 2026-09-18
fix(mcp): answer the requests the stdio proxy cannot serve (#2554)

### LINKED ISSUE #2554: mempalace-mcp leaves requests unanswered when it cannot serve them
**What happened?**

With a hub running, `mempalace-mcp` forwards requests to it, and once the hub is gone it serves the session itself. On develop @ 36ec72f9 these requests never get an answer, so the client waits on each one:

- The local server cannot start: `--backend` names no backend, or the storage stack fails to import. Each request raises inside `_run_proxy_loop` and is dropped. This came up in the review of #2534, and 25203ed6, from before that PR, does the same. A failed import also leaves fd 1 on stderr, so the hub's later answers go there too.
- The hub's answer breaks off (`IncompleteRead`, `BadStatusLine`), which `_handle` does not catch. With a real hub whose answer to a diary write was cut in transit, the write was committed and the client heard nothing for 120 s.
- A line that is not a JSON-RPC object: invalid JSON, or `[1, 2]`, `7`, `null`. The full server answers these with `-32700` and `-32600`.
- `--palace` or `--backend` with no value. Once the hub is gone, the proxy's own parser exits with status 2. Without a hub, the full server refuses it at startup.

**What did you expect?**

An answer to every request that carries an id.

**How to reproduce:**

1. `mempalace-mcp --transport http --port 8765 --palace /tmp/p`
2. `mempalace-mcp --backend bogus --palace /tmp/p`, then send `{"jsonrpc": "2.0", "id": 1, "method": "initialize", "params": {}}` on its stdin. The hub answers it.
3. Stop the hub and send `{"jsonrpc": "2.0", "id": 2, "method": "ping"}`. Nothing comes back, and stderr shows `Server error: "unknown backend 'bogus'; ...`.

**Environment:**
- OS: Linux (WSL2)
- Python version: 3.12.3

mvalentsev:
## What does this PR do?

Fixes #2554.

- When the local server cannot start, the request gets a `-32000` error naming why, e.g. `unknown backend 'bogus'`. After a failed hub call it gets the hub's `palace hub proxy failed` error, with that reason in `data.local_server`. A failed import of the server hands fd 1 and `sys.stdout` back.
- A hub answer that breaks off counts as a failed hub call: a call that changes state is not replayed, and a read is served locally. The full server's own forwarding answered `-32603` there and now does the same. Which calls change state now comes from read-only mode's set, so `mempalace_memories_filed_away`, which consumes the hook checkpoint, is no longer rerun locally and answered `"status": "quiet"`.
- A line that does not parse gets `-32700`, JSON that is not an object `-32600`, and any other failure `-32603`. `params` that are not an object no longer lose the hub's answer.
- `--palace` or `--backend` with no value goes to the full server, which refuses it at startup, as it does without a hub.

Not fixed here: the full and light servers' own stdio loops still exit on a line that `json.loads` rejects with something other than `JSONDecodeError`.

## How to test

```
uv run pytest tests/test_mcp_proxy.py tests/test_hub_forward.py -v
```

The new tests fail on develop, except two that pin what develop already does. One checks that a notification whose handling raises gets no answer, the other that `--collection` with no value stays on the proxy.

## Checklist
- [x] No hardcoded paths


mempalace/mcp_proxy.py
```diff
@@ -188,7 +225,7 @@ def load(self):
                 "MemPalace hub unavailable; serving this session locally. "
                 "Loading the local storage stack (this process will grow)."
             )
-            from . import mcp_server
+            mcp_server = _import_server()
```
igorls: **[suggestion]** A failed `mcp_server` import still leaks one stdout fd per request. `__init__` does `os.dup(1)` and `os.dup2(2, 1)` before the import that can fail (chromadb). On failure importlib drops the half-initialized module, so that dup is never closed and `_restore_stdout` is gone with it. `_import_server` correctly `dup2`s its own saved fd back onto 1 and closes only that saved fd (`os.close(saved_fd)`). `_LocalServer.load` leaves `_module` unset when import raises, so the next request imports again, dups again, and leaks again. Before this change the client hung on the first request, so the leak happened once; answering and continuing makes it one fd per call for the life of the session. Unknown-backend startup does not do this: that import succeeds and is cached, and the later `BackendUnavailableError` does not dup. Answers still go to the real stdout (the new subprocess test). The leak is only the orphaned server dup.

**Suggestion:** Remember the load failure on `_LocalServer` and re-raise it without calling `_import_server` again. That caps the leak at one fd and skips a repeated failed init. Closing even that fd means the server's own import has to restore fd 1 in a `finally` if init fails; the proxy cannot see the fd number after the module is discarded.

PR #2557, opened 2026-09-18
fix(mcp): keep the stdio servers up on a line json.loads rejects (#2556)

### LINKED ISSUE #2556: The stdio MCP servers exit on a line json.loads rejects with ValueError or RecursionError
**What happened?**

`json.loads` rejects an integer longer than Python's digit limit (4300 by default) with `ValueError`, and nesting too deep to parse with `RecursionError`. The stdio loops of the full server (`mempalace-mcp` with no hub running) and of `mempalace-light-mcp` do not catch these, so one such line ends the process with status 1.

The full server's catch was narrowed to `(JSONDecodeError, UnicodeDecodeError)` in #2232. Before that, on 50fb09d2, it kept serving after both lines. The hub's HTTP transport answers them with `-32700`.

**What did you expect?**

The server keeps serving and answers the line with `-32700`, as it does for invalid JSON.

**How to reproduce:**

1. With no hub running: `python3 -c 'print("{\"jsonrpc\": \"2.0\", \"id\": 1, \"method\": \"ping\", \"x\": " + "1" * 5000 + "}"); print("{\"jsonrpc\": \"2.0\", \"id\": 2, \"method\": \"ping\"}")' | mempalace-mcp --palace /tmp/p`
2. The process exits with status 1 on `ValueError: Exceeds the limit (4300 digits) for integer string conversion`, and request 2 gets no answer.
3. `mempalace-light-mcp --palace /tmp/p` does the same.

**Environment:**
- OS: Linux (WSL2)
- Python version: 3.12.3

mvalentsev:
## What does this PR do?

Fixes #2556.

`json.loads` rejects an integer past Python's digit limit with `ValueError` and nesting too deep to parse with `RecursionError`. Neither the full server's stdio loop nor the light server's caught them, so one such line ended the process.

- The full server answers any line `json.loads` rejects with `-32700`, as the hub's HTTP transport does. #2232 kept this catch narrow because a `RecursionError` is no parse error, but here it is one, and the narrow catch let it end the server.
- The light server skips and logs such a line, as it skips invalid JSON.

The proxy loop that `mempalace-mcp` runs while a hub is up is covered by #2555.

## How to test

```
uv run pytest tests/test_mcp_jsonrpc_envelope.py tests/test_mcp_light_server.py -v
```

The four new cases fail on develop. The digit-limit case is skipped on an interpreter without that limit.

## Checklist
- [x] No hardcoded paths


mempalace/mcp_server/runtime.py
```diff
@@ -221,15 +221,19 @@ def _run_stdio_loop() -> None:
...
-            # Narrow on purpose: reporting a MemoryError or RecursionError as
-            # "Parse error" would be a lie. The id is unknowable here, so it is
-            # null per JSON-RPC 2.0 section 5 -- the "never answer a
-            # notification" rule cannot bind when the notification is exactly
-            # what could not be parsed. Staying silent left the client waiting
-            # on a request it had already sent, while the HTTP transport has
-            # answered -32700 all along.
-            logger.error("Server error: %s", exc)
+        except Exception as exc:
+            # Whatever json.loads raises, the line gave no request: invalid JSON,
```
igorls: **[suggestion]** The replacement comment narrates the bug ("catching only the decode errors let the last two end the server") and restates what `-32700` means, on top of the protocol note that was already there. A reader needs the non-obvious part (id is null because the line never became a request; log the type because `MemoryError` has an empty message), not the history of the narrower catch.

**Suggestion:** Keep two or three lines: these `ValueError` / `RecursionError` cases are still parse failures, so answer `-32700` with a null id the way the HTTP transport does, and include the exception type in the log.

PR #2559, opened 2026-09-19
fix(hooks): treat null transcript paths as absent

wstczyw:
## What does this PR do?

Fix the SessionEnd edge case where a harness sends JSON `null` for `transcript_path`.

`_parse_harness_input()` previously applied `str(...)` unconditionally, turning JSON `null` into the literal path `"None"`. The validator then rejected that path and logged a misleading warning, preventing the normal no-transcript SessionEnd path from running.

This change normalizes JSON `null` to the existing absent-path representation (`""`). Non-empty paths still pass through the existing validator unchanged.

## Tests

- `pytest tests/test_hooks_cli.py::test_parse_harness_input_normalizes_null_transcript_path tests/test_hooks_cli.py::test_parse_harness_input_valid -q` — 2 passed
- `ruff check mempalace/hooks_cli.py tests/test_hooks_cli.py`
- `ruff format --check mempalace/hooks_cli.py tests/test_hooks_cli.py`
- `git diff --check`

This is intentionally independent of the Codex `item_completed` work in #2233 and contains no local Hub-specific changes.


mempalace/hooks_cli.py
```diff
@@ -1227,10 +1227,12 @@ def _parse_harness_input(data: dict, harness: str) -> dict:
     if harness not in SUPPORTED_HARNESSES:
         print(f"Unknown harness: {harness}", file=sys.stderr)
         sys.exit(1)
+    raw_transcript_path = data.get("transcript_path", "")
+    transcript_path = "" if raw_transcript_path is None else str(raw_transcript_path)
     return {
         "session_id": _sanitize_session_id(str(data.get("session_id", "unknown"))),
```
AmirF194, line 1233: I ran `_parse_harness_input` from this branch (head `933c06f`) in a clean container against the same shape of payload this PR's own test uses, with `session_id` also null.

`session_id` has the identical bug this PR fixes for `transcript_path`, two lines below: `str(data.get("session_id", "unknown"))` only supplies `"unknown"` when the key is missing. When a harness sends `"session_id": null` (a `.get` with a default never fires on an explicit `None`), this becomes `str(None)`, and `_sanitize_session_id` does not catch it either: its `[^a-zA-Z0-9_-]` filter passes every character of the literal string `"None"` through unchanged, so the fallback in `_sanitize_session_id`'s own `sanitized or "unknown"` never triggers.

```
$ docker run --rm -v $(pwd):/repo:ro -w /repo python:3.11-slim bash -c "
pip install -q -e . && python3 -c \"
from mempalace.hooks_cli import _parse_harness_input
print(_parse_harness_input({'session_id': None, 'transcript_path': None}, 'claude-code'))
\""
{'session_id': 'None', 'stop_hook_active': False, 'transcript_path': ''}
```

`session_id` comes back as the string `'None'` instead of `'unknown'`, so every hook that receives a null session id gets grouped under the same literal-`'None'` identity, which is a different bucket than the intended `'unknown'` fallback that a missing key would produce. Same fix as line 1231, applied to line 1233:

```python
raw_session_id = data.get("session_id", "unknown")
session_id = "unknown" if raw_session_id is None else str(raw_session_id)
```

PR #2565, opened 2026-09-21
feat(config): let operators override mempalace_status protocol text (#2451)

### LINKED ISSUE #2451: Allow an operator to append to the protocol text returned by mempalace_status
**Version:** mempalace 3.9.0 (PyPI), served via `mempalace serve`, backend `chroma`

## Summary

The protocol string returned by `mempalace_status` is hardcoded and cannot be
extended. It instructs the reading agent to query the palace before answering
and never to guess. An operator who *knows* their corpus carries a systematic
bias has no way to tell the agent that is reading it.

## Where

`mcp_server.py:2278`. No `config.json` key and no environment variable appends
to or overrides it — I checked for `MEMPALACE_PROTOCOL`, `protocol_override`,
`custom_protocol`, and environment reads around the string.

Worth noting the boundary falls in a slightly surprising place: `topic_wings`
and `hall_keywords` *are* operator-configurable via `config.json`, so the palace
already accepts operator input about how it classifies — just not about how its
output should be read.

## Why it matters: two independent, measured biases

**1. Extraction keeps rules and drops exceptions.**

Scoped honestly: one model, at one size, under four schema shapes. Not a claim
about extraction in general.

A passage whose entire thesis was a caution had its positive claim extracted on
every run (`bm25 | is | a FOREIGN-VOCABULARY detector`) while the contrast that
carried the warning — *not* a confabulation detector — was dropped in all four.

Attempts to require the missing content made it worse rather than better:

- a free-text `caveats` array returned commentary *about* the note
  ("The note does not specify...") rather than the warning the note issued;
- renaming it `warnings` with an explicit description — *imperatives the NOTE
  ITSELF issues, quote the note, NOT observations about what it omits* —
  produced exactly one item, and it was the forbidden kind;
- tightening to a `{is, is_not}` object produced tautologies:
  `is: "below a confabulation at 0.473"`, `is_not: "above a confabulation at 0.473"`.

`minItems: 1` forces presence, not correctness. Notably, the negative predicates
in this palace that *are* trustworthy — `do_not_trust`, `must_not_be_used`,
`should_not_be_used_in`, around 38 distinct ones — all appear because the model
volunteered them, not because a schema demanded them.

**2. Retrieval is dominated by base rate in a skewed corpus.**

One wing holds roughly 91% of ~3.4k drawers. `total_before_filter` reads 20
unscoped against 10 when scoped to a small wing, so the candidate pool is drawn
palace-wide and capped. A generic-vocabulary query whose best answer lived in a
10-drawer wing returned 5 of 5 results from the large wing. The relevant drawer
scored 0.441 against a top result of 0.486 — marginal on relevance, but rarely
in the candidate set to be ranked at all.

These are different mechanisms with different fixes. Both are facts the operator
knows and the reading agent cannot.

## Why a local patch is not a workaround

The image here is `FROM python:3.12-slim` + `pip install --no-cache-dir
mempalace==3.9.0`, so editing `mcp_server.py` in `site-packages` is discarded by
the next image build, silently and with no warning.

More importantly, and this is the part that would rule it out even if it were
easy: **a protocol that sometimes carries the caveat and sometimes does not is
worse than one that never does**, because a reader cannot distinguish it from
its own absence. The workaround would introduce the exact failure mode the
caveat exists to prevent.

## Request

A config key (e.g. `protocol_append` in `config.json`) or an environment
variable whose contents are appended to the protocol payload returned by
`mempalace_status`.

Append rather than replace, so the shipped guidance stays authoritative and the
operator's note is strictly additive.

## Current mitigation, and its limit

Per-machine context injected at session start — a `CLAUDE.md` on one machine,
memory notes on another. It is durable and it works, but it is per-machine: a
machine added later silently has no copy, and nothing reports the gap. That is
why the caveat wants to live with the corpus rather than with the reader.

BuboTheWise:
## Issue
Resolves #2451

## Problem
`mempalace_status` returns a hardcoded protocol string that an operator cannot customize. `topic_wings` and `hall_keywords` are already operator-configurable — the read-protocol is not.

## Solution
Adds a `status_protocol` config property and `MEMPALACE_STATUS_PROTOCOL` env var, mirroring the existing `topic_wings`/`hall_keywords` pattern:

- **Precedence:** env var > config file > built-in default
- **Fallback:** `None` / blank → built-in `PALACE_PROTOCOL` constant (unchanged)
- **Scope:** 3 status-dict sites in `tools_read.py` changed to read through a new `_status_protocol()` helper. No search/query hot-path change.

## Usage
```json
{ "status_protocol": "YOUR CUSTOM INSTRUCTIONS" }
```
or
```
MEMPALACE_STATUS_PROTOCOL="YOUR CUSTOM INSTRUCTIONS"
```

## Tests
- 4 new config-layer tests (default, file, env-overrides-file, blank-as-unset)
- 4 new MCP server tests: resolver + end-to-end handler test on the fast sqlite path
- Full affected test matrix: **454 passed**
- 534 total pass across the 7 affected test files
- 6 pre-existing failures in `test_init.py` (site-packages path env-dependent) and 1 benchmark — identical on clean upstream/develop; **zero regressions**

## Diff
4 files, +141 −3


BuboTheWise:
All upstream gates are green and this PR is merge-ready.

- CI: 12/12 passing
- State: mergeable, clean (no conflicts)
- Base: develop
- No blocking review comments outstanding

I'm a fork contributor (not a MemPalace org member) so I can't merge directly — the merge is one click away for a maintainer when it suits the team. If anything in the diff needs a tweak I'm standing by and will act immediately; otherwise it stays ready to go.

Addresses #2451.

igorls:
Thanks for this. The config plumbing and the three status return paths are in the right place, and the blank-as-unset fallback is the right default. Please change the override from a replacement to an append before we merge.

Issue #2451 asks for a note added after the built-in `mempalace_status` protocol, so the shipped guidance stays in every response. `_status_protocol()` currently returns the operator string in place of `PALACE_PROTOCOL`. A set value removes "query the palace, never guess" and substitutes whatever was configured. That is the failure the issue wants to avoid: a reader cannot tell a replaced protocol from the real one.

Please return `PALACE_PROTOCOL` plus the operator text when `status_protocol` / `MEMPALACE_STATUS_PROTOCOL` is set, and the built-in text alone when the setting is missing or blank.

Two smaller notes for the same revision:

- A whitespace-only `MEMPALACE_STATUS_PROTOCOL` currently hides a real `config.json` value instead of falling through to it. Treat a blank env var as unset and keep reading the file.
- The new comments cite the issue number. Describe the behavior and leave the number out of the source.

BuboTheWise:
Thanks, @igorls — all three points are right, and I agree with the underlying reasoning on the append one in particular: a reader has to be able to tell the shipped guidance apart from anything an operator tacked on.

**1. Append, not replace.** I'll change `_status_protocol()` (`mempalace/mcp_server/tools_read.py`) so that when `status_protocol` / `MEMPALACE_STATUS_PROTOCOL` is set it returns `PALACE_PROTOCOL` + the operator text (separated by a blank line), and `PALACE_PROTOCOL` alone when the setting is missing or blank. So the built-in "query the palace, never guess" text stays in every response, as #2451 intends. The `status_protocol` config semantics (env > file, blank = unset) are unchanged — only the consumer composes the final string.

**2. Whitespace-only env hiding the file value.** Correct bug. The `status_protocol` property in `config.py` returns `None` the instant the env var is present-but-blank, so it never falls through to `config.json`. I'll change it so a blank env is treated as *unset* and the file value is read — only a non-blank env short-circuits.

**3. Issue number out of source.** I'll strip the `#2451` citation from the new comments in `config.py` and `tools_read.py` and keep only the behavior description.

I'll also update the two status tests to assert the *appended* form (built-in text present **and** operator text present), and add the blank-env-falls-through-to-file case as a new test. Will push the fix, confirm CI, and ping here when it's up. Thanks for the careful read.

BuboTheWise:
@igorls — all three points from your review are addressed in the post-fix head `8a8347d`:

1. `_status_protocol()` now **appends** the operator text after `PALACE_PROTOCOL` (blank-line separator) rather than replacing it. The built-in "query the palace, never guess" guidance is always present in every status response.
2. A whitespace-only `MEMPALACE_STATUS_PROTOCOL` is treated as **unset**, so the `config.json` value is read as a fallback. Only a non-blank env wins.
3. `#2451` has been stripped from the new source comments in `config.py` and `tools_read.py`; behaviour is described without the number.

Gates: CI 12/12 green · mergeable · state CLEAN · OPSEC clean (verified by REVIEW t_ccbc644c).

Ready for merge to `develop` whenever you have a moment.

PR #2567, opened 2026-09-21
fix(mine): scope prefetch_mined_set to candidate files below a threshold

### LINKED ISSUE #2561: convo miner: prefetch_mined_set and prefetch_content_hashes scan the whole collection even for a one-file mine (79% of a 23 min mine at 660k drawers)
mempalace 3.10.0, chromadb 1.5.9 (Rust bindings), Python 3.14, Linux.

Profiled one mine of a 742 KB Claude Code transcript (95 drawers) against a 660528-drawer palace and against an empty palace, same process, cProfile plus `/usr/bin/time`.

- Full palace: 1411 s wall. Empty palace: 6.85 s. Embedding 3.9 s in both. Chroma upsert incl. HNSW insert: 1.06 s vs 0.13 s.
- `prefetch_mined_set` (`palace/mined.py:129`): 595.7 s, 661 calls to `collection.get(limit=1000, offset=..., include=["metadatas"])`.
- `prefetch_content_hashes` (`palace/mined.py:215`): 521.5 s, same 661 pages over the same collection.
- `compute_hallways_for_wing`: 180.4 s. FTS5 `quick_check`: 102.1 s.
- `Bindings.get` mean 0.867 s per page; the scan is linear in drawer count.

Both prefetches are called unconditionally at `convo_miner.py:1065` and `:1074`, guarded only on `collection is not None`. Their docstrings justify the full scan by the 2000-file sweep case, where it replaces per-file `where` queries. For a one-file mine (the shape every hook-triggered mine has) it is a pure 18 min tax.

## Proposed fix

Pass the candidate `source_file` list (and wing for the content-hash map) into both functions. Below a threshold (say 50 files) use `collection.get(where={"source_file": {"$in": files}}, include=["metadatas"])` instead of paging the collection; above it keep the full scan.

`prefetch_content_hashes` cannot be scoped by `source_file` alone since its purpose is finding the same conversation under a different path, so scope it by wing plus the candidate content hashes with a `where` on `content_hash` where the backend supports it, or skip it entirely when the caller passes a single already-known path (the Claude Code hook case).

Happy to test a patch on the 660k palace.

AmirF194:
## What does this PR do?

`prefetch_mined_set` and `prefetch_content_hashes` each page through the whole
drawer collection to decide which of the files a mine is about to touch are
already filed, even when the mine covers a single file (the shape every
hook triggered mine has). On the reporter's 660k drawer palace that full
scan was 79 percent of a 23 minute one file mine.

The mtime skip check only ever needs the exact `source_file` paths the
caller already has, so below a small candidate count `prefetch_mined_set`
now asks ChromaDB for just those rows via
`where={"source_file": {"$in": ...}}` instead of paging the whole
collection. Above that threshold, or when no candidate list is passed, the
full scan runs exactly as before, since a bulk sweep still benefits more
from one scan than from many filtered queries.

`prefetch_content_hashes` is not scoped the same way. Its `content_hash`
field can hold several comma joined hashes per drawer (one privacy export
bundle covers several conversations), so a `where` filtered query can only
match a document whose whole stored value equals one candidate hash and
would silently miss any bundle row. I tried scoping and then skipping it
below the threshold; both broke `test_mine_convos_skips_same_content_under_new_filename`
and `test_mine_convos_skips_same_conversation_within_re_exported_bundle`,
so it stays a full scan and only the mtime check got faster. This still
removes the larger of the two costs the issue measured.

Fixes #2561

## How to test

```
docker run --rm python:3.11 bash -c "pip install -e '.[dev]' && pytest tests/test_convo_miner.py tests/test_miner.py -v"
```

The new tests seed a collection directly and assert the scoped query
returns the same entry an unscoped scan would, omits files outside the
candidate list, and falls back to a full scan once the candidate count
exceeds the threshold. `test_mine_convos_scopes_mined_set_prefetch_to_candidate_files`
checks the convo miner actually passes the file list to `prefetch_mined_set`
and that `prefetch_content_hashes` is still called unscoped.

## Verification

- `tests/test_convo_miner.py` (46 tests, all passing) and `tests/test_miner.py`
  (131 tests, all passing), including the two cross-path duplicate tests
  named above, which are the ones that would catch a regression in the
  untouched `prefetch_content_hashes` path.
- `ruff check .` and `ruff format --check .` clean.
- Not run: the full suite across the whole repo and the 3.9/3.13 legs of
  the CI matrix, only 3.11 locally. The change is confined to
  `mempalace/palace/mined.py` and its one call site, both fully exercised
  above.

## Checklist
- [x] No hardcoded paths


mempalace/palace/mined.py
```diff
@@ -160,41 +166,61 @@ def prefetch_mined_set(
     `collection.get(where={"source_file": X})` costs ~2s on a 150k-drawer
     palace, making a 2000-file sweep take >1h of pure skip-checking. This
     helper drops that to a single paginated scan plus O(1) lookups.
+
+    When `source_files` is given and holds at most `_PREFETCH_SCOPE_THRESHOLD`
+    paths, the caller already knows the only source_file values that could
+    possibly match, so a single `where={"source_file": {"$in": ...}}` get()
+    replaces the full-collection scan (#2561: for a one-file hook-triggered
+    mine against a 660k-drawer palace, the full scan was 79% of total mine
```
igorls: **[suggestion]** The one-file mine in the report spent 595.7s in `prefetch_mined_set` and 521.5s in `prefetch_content_hashes` (together about 79% of 1411s). This change only scopes the first call. The new docstring says that full scan "was 79% of total mine time", which attributes both costs to `prefetch_mined_set` alone (595.7/1411 is about 42%). Leaving `prefetch_content_hashes` unscoped is the right correctness call: a stored `content_hash` can be several comma-joined hashes, and a `where` on one candidate hash, or skipping the scan, looks like "no duplicate" and re-files bundle rows. The bash save and precompact hooks still mine the transcript's parent directory, so the candidate list is every session file in that directory, not one path; once that directory exceeds 50 files the new branch does not run either, and the hash scan still does.

**Suggestion:** Keep the hash prefetch unscoped. Correct the 79% comment so it names both calls. If the hook case is the goal, skip `prefetch_content_hashes` only when every candidate is already an unchanged `mined_mtimes` hit (nothing will be re-read, so a missing bundle row cannot cause a re-file), or raise the threshold so a project directory of session files still takes the scoped mtime query.

mempalace/palace/mined.py
```diff
@@ -160,41 +166,61 @@ def prefetch_mined_set(
...
-                version = meta.get("normalize_version", 1)
-                if version < NORMALIZE_VERSION:
-                    continue
-                stored_mtime = meta.get("source_mtime")
-                mtime_key = float(stored_mtime) if stored_mtime is not None else None
-                entry = groups.setdefault(src, {}).setdefault(
-                    mtime_key, {"count": 0, "chunk_total": None}
+        if source_files is not None and len(source_files) <= _PREFETCH_SCOPE_THRESHOLD:
+            if source_files:
+                scoped = collection.get(
```
igorls: **[suggestion]** The `$in` list itself is capped at 50, so the filter clause is not the SQLite variable problem. The get has no `limit`. Chroma 1.5.9 treats a missing limit as `u32::MAX` and then issues a follow-up `embedding_metadata_array.id IN (...)` over every matched row. This repo has measured that follow-up raising `too many SQL variables` past 32766 matched drawers. `MAX_FILE_SIZE` is 500MB, so one candidate transcript, or up to 50 files together, can exceed that. The `except` logs and returns whatever was absorbed, which on a failed get is `{}`, and does not fall back to the paginated scan. That is a false "not mined", not a false skip: `_file_chunks_locked` calls `file_already_mined` (pages of 1000) before `collection.delete`, so an unchanged complete file is not purged. The cost is that every candidate is normalized and chunked as if new, and a later change that removed the recheck would delete on this error.

**Suggestion:** Page the scoped get (`limit`/`offset`, stop on a short page) the same way the fallback scan does, and on a scoped-get failure fall back to that scan instead of returning an empty map.

mempalace/palace/mined.py
```diff
@@ -126,8 +126,14 @@ def file_already_mined(
         return False
 
 
+# Above this many candidate files, a single `where`-filtered get() binds
+# more query variables than a full paginated scan costs in round trips, so
+# the scoped path below stops paying off (#2561).
```
igorls: **[nit]** New comments in `mined.py` and `convo_miner.py` embed an issue number. This repo does not put issue numbers in code comments.

**Suggestion:** Drop the issue number and keep the behavioral reason (threshold, comma-joined hashes).

PR #2568, opened 2026-09-21
fix(mcp): keep stdio loops alive on a huge JSON integer

### LINKED ISSUE #2556: The stdio MCP servers exit on a line json.loads rejects with ValueError or RecursionError

sxh313:
## What does this PR do?

Keeps both stdio MCP servers alive when a line contains a JSON integer longer than `sys.get_int_max_str_digits()`.

`json.loads` raises a **plain `ValueError`** for that input, not a `json.JSONDecodeError`. The stdio loops caught the narrower classes, so the exception escaped the loop and the process died with exit status 1 on that one line - and every request after it went unanswered:

- `mempalace/mcp_server/runtime.py` caught `(json.JSONDecodeError, UnicodeDecodeError)`;
- `mempalace/mcp_light_server.py` caught `json.JSONDecodeError`.

Both are now `except ValueError`, which is a strict superset of what they caught before (`JSONDecodeError` and `UnicodeDecodeError` are both `ValueError` subclasses), so nothing that used to be handled stops being handled. The full server answers the line with `-32700` as it already does for other unparsable input; the light server keeps dropping the line it cannot parse, but serves the next one.

`MemoryError` and `RecursionError` stay outside the guard on purpose - that narrowness is the existing documented decision in the comment above the handler ("reporting a MemoryError or RecursionError as 'Parse error' would be a lie"), and this change does not touch it. #2556 lists the `RecursionError` case too; it needs its own answer code or a recursion-limit decision, which is a separate call.

Fixes #2556.

## How to test

Two regression tests, one per loop, both reproduced red on `develop` before the change with the identical cause:

```text
$ python -m pytest tests/test_mcp_jsonrpc_envelope.py -k digit_limit -q
E           ValueError: Exceeds the limit (4300 digits) for integer string conversion: value has 5000 digits; use sys.set_int_max_str_digits() to increase the limit
C:\...\Lib\json\decoder.py:354: ValueError
FAILED tests/test_mcp_jsonrpc_envelope.py::TestStdioLoopAlwaysResponds::test_integer_over_the_digit_limit_is_answered_not_fatal
1 failed, 41 deselected in 31.52s

$ python -m pytest tests/test_mcp_light_server.py::test_light_stdio_loop_survives_an_unparsable_huge_integer -q
E           ValueError: Exceeds the limit (4300 digits) for integer string conversion: value has 5000 digits; use sys.set_int_max_str_digits() to increase the limit
1 failed in 7.26s
```

After the change:

```text
$ python -m pytest tests/test_mcp_jsonrpc_envelope.py tests/test_mcp_light_server.py -q
86 passed in 38.57s
```

```text
# every test file that drives the MCP servers, minus tests/mcp/test_kg.py (run separately below)
$ python -m pytest tests/mcp tests/test_mcp_jsonrpc_envelope.py tests/test_mcp_light_server.py -q --ignore=tests/mcp/test_kg.py
470 passed, 9 skipped, 2 warnings in 432.27s (0:07:12)

$ python -m pytest tests/mcp/test_kg.py -q
30 passed in 11.98s

$ python -m ruff check .
All checks passed!
$ python -m ruff format --check mempalace/mcp_server/runtime.py mempalace/mcp_light_server.py tests/test_mcp_jsonrpc_envelope.py tests/test_mcp_light_server.py
4 files already formatted
$ pre-commit run --files <the four changed files>
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
```

The tests reuse the harnesses already in those files (`_run_loop` with the startup side effects stubbed out, and `_keep_server_command_line_state` for the light server's `main()`), so they need no palace, no hub and no network. The payload is the reporter's own: `{"jsonrpc": "2.0", "id": 1, "method": "ping", "x": <5000 ones>}` followed by a valid `ping`.

Environment: Windows 11, Python 3.12.13, `develop` at `e038af9`. The hub-proxy loop is not part of this: `mempalace/mcp_proxy.py:301` already catches `Exception` around its `json.loads`, so it survives this line, and its separate problem (the request is dropped without an answer) is #2554, which #2555 addresses.

Also related, so a reviewer does not have to diff twice: #2555 edits the *same file* (`mempalace/mcp_server/runtime.py`) but a different hunk - the hub-forward `except (urllib.error.URLError, OSError, TimeoutError, ValueError)` in the proxy path - and its tests exercise `RecursionError` on a deeply nested document there, which is the other half of #2556 that this PR deliberately leaves alone in the two stdio loops. Expect no conflict, and whichever merges second should just rebase.

## Checklist
- [ ] Tests pass (`python -m pytest tests/ -v`) -- see the caveat below; I did not obtain a full-suite result on this machine
- [x] No hardcoded paths

Caveat on the first box, with numbers rather than a shrug: I could not get a whole-tree result on Windows, so I ran the tree in halves and here is exactly what each one did. **Half 1 (88 test files + `tests/mcp`) completed: 2727 passed, 317 skipped, 3 failed in 12m45s.** All three failures are reproducible-without-my-diff and diagnosed: two are `tests/test_cli_rules.py::TestRulesCli::test_cli_renders_*`, where the test captures a CLI subprocess in text mode and the child prints a U+2014, so on a cp936 console `result.stdout` arrives as `None` and the assertion dies with `TypeError: argument of type 'NoneType' is not iterable' (the raw error under the hood is `UnicodeDecodeError: 'gbk' codec can't decode byte 0x94`); one is `tests/test_hub_forward.py::TestStdioProxy::test_retries_process_token_after_stale_palace_token_401`, which passes 83/83 when that file is run alone in 29.08 s. **Half 2 did not finish**: it stalled at 43% inside `tests/test_idle_watchdog_exit.py`, and an earlier whole-tree attempt stalled at ~73% with the process near 2.7 GB RSS; both look like wait-for-process-exit tests that do not return on Windows, which is a test-portability question for the maintainers, not something this PR changes. The complete MCP surface (470 + 30 above) is green, and none of the above can involve this diff: it adds no import and no module-level code, and changes only the exception class of two `except` clauses inside stdio loop functions plus comments.


mvalentsev:
@sxh313 #2557 has been open for #2556 since Sep 18, three days before this one, and it's linked on the issue. It handles both inputs the issue names. This PR handles one.

The other one still kills both servers on `d772d556`:

```
$ python3 -c 'print("[" * 100000); print("{\"jsonrpc\": \"2.0\", \"id\": 2, \"method\": \"ping\"}")' | mempalace-mcp --palace /tmp/p
MemPalace MCP Server starting...
...
RecursionError: maximum recursion depth exceeded while decoding a JSON array from a unicode string
$ echo $?
1
```

`mempalace-light-mcp` exits the same way, and request 2 gets no answer from either. `RecursionError` isn't a `ValueError`, and the new tests only feed the 5000-digit integer, so nothing here catches it. With `Fixes #2556` this would close the issue with the crash still there. Please close this in favor of #2557.

sxh313:
Closing this as a duplicate. While re-verifying my claims before the deadline I found #2557, opened by @mvalentsev - who also filed #2556 - and it is the same fix, arrived at independently: widen the stdio parse guard so a `ValueError` from an integer past the digit limit no longer ends the process, answer -32700 in the full server and skip the line in the light server. Their PR is broader than mine: it also handles the `RecursionError` case I deliberately left out of scope (I declined to touch it because the existing comment in `mempalace/mcp_server/runtime.py` calls reporting it as a parse error a lie, and their changelog entry shows the maintainers did want it covered), and it updates CHANGELOG.md and `tests/_mcp_server_helpers.py`. Mine adds two regression tests and a measured caveat about this repo's full suite not completing on Windows, which is still in #2569. Nothing in mine needs review; #2557 is the one to merge. My original timeline check missed the linked PR and I am sorry for the duplicate notification.

PR #2569, opened 2026-09-21
fix(cli): hold the palace lock across the legacy repair

### LINKED ISSUE #2562: repair (legacy mode) extracts and backs up for minutes before taking mine_palace_lock, so a hook miner started in that window makes it fail at the rebuild step
mempalace 3.10.0, Python 3.14, Linux, 660k-drawer palace (chroma.sqlite3 4.1 GB).

`mempalace repair --yes` on the legacy path (`cli/cmd_repair.py`) runs `_extract_drawers`, `check_extraction_safety` and `copy_palace_dir` to `<palace>.backup` with no `mine_palace_lock` held. On this palace that is about 10 minutes of work. It only contends for the lock once `_rebuild_collection_via_temp` starts upserting (per-batch acquire in `backends/chroma.py`). Any `mempalace mine` launched by a Claude Code hook inside that window takes the lock first, and repair then fails at the rebuild step with `MineAlreadyRunning`, discarding the extract and the backup. Three consecutive attempts lost this way on 2026-09-19 (3.9.0; the code path is unchanged in 3.10.0). It only succeeded with a wrapper that killed every `mempalace mine` process every 2 s for the 8.5 h run.

The other two repair entry points already hold the lease across the whole cycle: `repair.py` `rebuild_index` takes `mine_palace_lock` before the snapshot ("A writer landing after the snapshot but before the rebuilt collection becomes authoritative would otherwise be lost"), and `rebuild_from_sqlite` takes it before the archive/rename so contention fails clean. The legacy path is the odd one out.

Proposed fix: in `cmd_repair`, wrap everything from `backend.get_collection` through `_post_rebuild_cleanup` in `with mine_palace_lock(palace_path):`, catching `MineAlreadyRunning` up front and exiting non-zero with the holder identity before any extraction starts. The per-batch acquires inside the rebuild pass through re-entrantly on the same thread, as they already do for `rebuild_from_sqlite`.

Observation, unverified: after the successful run, `locks/mine_palace_<hash>.lock` still named repair's pid and the next hook miner appeared refused until the file was removed. Reading `palace_lock.py`, the file is never unlinked and the flock is released at exit, so this should be inert; noting it in case there is a code path I missed.

sxh313:
## What does this PR do?

Makes the legacy `mempalace repair` path hold the palace's mine lease for the whole pass, so contention is reported in the first second instead of ten minutes in - after the extraction and the backup copy have already been thrown away. Fixes #2562.

`cmd_repair`'s legacy path ran `backend.get_collection` -> `_extract_drawers` -> `check_extraction_safety` -> `copy_palace_dir` -> `_rebuild_collection_via_temp` -> `_post_rebuild_cleanup` with no lock held, and only the per-batch acquires inside the rebuild contended. A `mempalace mine` started by a hook during that window wins, and the rebuild step then fails with `MineAlreadyRunning`.

The other two repair entry points already take the lease up front, and say why: `rebuild_index` holds it before the snapshot ("A writer landing after the snapshot but before the rebuilt collection becomes authoritative would otherwise be lost") and `rebuild_from_sqlite` takes it before the archive/rename so contention fails clean. This makes the third entry point match them:

```python
    try:
        with mine_palace_lock(palace_path):
            ... existing legacy pass ...
    except MineAlreadyRunning as exc:
        print(f"  Another writer already holds this palace: {exc}")
        print("  Repair stopped before reading or copying anything; nothing changed.")
        raise SystemExit(2)
```

Exit code 2 matches how this command already refuses to start (`_maintenance_requires_chroma` above it), and the message carries the holder identity that `mine_palace_lock` puts in the exception (PID plus what holds it).

Re-entrancy: `mine_palace_lock` passes through when the same process already holds the lease for that palace, which is what lets the per-batch acquires in `ChromaCollection` compose with the outer one - the same reason `rebuild_from_sqlite` can do this today. `tests/test_chroma_collection_lock.py` covers that behavior and passes unchanged (see below).

## How to test

Two regression tests in `tests/test_cli.py`, next to the existing `cmd_repair` tests, driving the real function with the same mock-backend harness those tests use.

Before the change, the denial test proves the whole pass ran anyway - the repair printed the extraction, the backup, the rebuild and `Repair complete` while the lock was raising `MineAlreadyRunning`:

```text
$ python -m pytest tests/test_cli.py -k "palace_lock or lock_is_taken" -q      # before the fix
  Extracting drawers...
  Extracted 2 drawers
  Backing up to C:\Users\...\pytest-83\test_cmd_repair_refuses_to_ext0\palace.backup...
  Building temporary collection: mempalace_drawers__repair_tmp
  Staged 2/2 drawers...
  Rebuilding live collection...
  Re-filed 2/2 drawers...
  Repair complete. 2 drawers rebuilt.
FAILED tests/test_cli.py::test_cmd_repair_holds_the_palace_lock_before_extracting
FAILED tests/test_cli.py::test_cmd_repair_refuses_to_extract_when_the_lock_is_taken
2 failed, 97 deselected in 4.36s

$ python -m pytest tests/test_cli.py -k "holds_the_palace_lock or lock_is_taken" -q   # after
2 passed, 97 deselected in 4.35s
```

Regressions:

```text
$ python -m pytest tests/test_cli.py tests/test_repair.py -q
255 passed, 9 skipped in 55.80s
$ python -m pytest tests/test_chroma_collection_lock.py tests/test_cli_source_adapters.py -q
19 passed, 2 skipped in 21.03s
$ python -m ruff check mempalace/cli/cmd_repair.py tests/test_cli.py
All checks passed!
$ pre-commit run --files mempalace/cli/cmd_repair.py tests/test_cli.py
ruff (legacy alias)...Passed    ruff format...Passed
```

End-to-end, with the **real** lock held by a **separate process** (only the Chroma backend is mocked, so this needs no multi-GB palace):

```text
holder pid 2732 holds the lock; running repair now

 MemPalace Repair
  Palace: C:\Users\...\repair-lock-proof-iwoe66kb\palace
  Another writer already holds this palace: palace C:\Users\...\palace is held by PID 21196 (-c ...); wait for it to finish or stop the holder before retrying
  Repair stopped before reading or copying anything; nothing changed.

repair exit code: 2
backup dir created: False
files under palace : ['palace']
```

Environment: Windows 11, Python 3.12.13, `develop` at `e038af9`.

Diff size is misleading by design: `mempalace/cli/cmd_repair.py` is +105/-88 because the existing pass moves inside the `with`, so 88 of those lines are pure re-indentation - reviewing with `?w=1` shows the new logic as 17 lines. `tests/test_cli.py` is additive (+97/-0).

Not verified here, and unchanged by this PR: the reporter's 4.1 GB palace and its ~10-minute window, and the side observation that `locks/mine_palace_<hash>.lock` still named repair's PID after a successful run. `palace_lock.py` never unlinks the file and releases the flock at exit, so the stale name looks inert, but that is a separate question from this one and I did not touch it.

## Checklist
- [ ] Tests pass (`python -m pytest tests/ -v`) - measured on this branch rather than borrowed from #2568: half the tree ran to completion here (88 top-level files + `tests/mcp`) -> **2730 passed, 317 skipped, 2 failed in 8m00s**. Both failures are `tests/test_cli_rules.py::TestRulesCli::test_cli_renders_rules` and `..._test_cli_renders_light_mcp`, which capture a CLI subprocess in text mode; the child prints a U+2014 and on a cp936 console that raises `UnicodeDecodeError: 'gbk' codec can't decode byte 0x94`, leaving `result.stdout` as `None` and failing the assertion with `TypeError: argument of type 'NoneType' is not iterable'. Note the second one renders the *light MCP* help text, which this branch does not touch at all (that is #2568's file) - so it is a Windows console-encoding artifact, reproducible independently of either PR. The rest of the tree (`tests/test_idle_watchdog_exit.py` onward) does not finish on this machine and stalls the runner; see the same finding written up on #2568. Everything that does run here for the touched surface is green: 255 passed / 9 skipped across `test_cli.py` + `test_repair.py`, 19 passed / 2 skipped across the lock-neighbor files, and the ruff + pre-commit gates pass.
- [x] No hardcoded paths


mempalace/cli/cmd_repair.py
```diff
@@ -241,102 +242,118 @@ def cmd_repair(args):
...
+            except Exception as e:
+                print(f"  Error reading palace: {e}")
+                print(index_read_recovery_guidance())
+                return
+
+            if total == 0:
+                print("  Nothing to repair.")
+                return
+
+            if not confirm_destructive_action(
```
igorls: **[bug]** `confirm_destructive_action` calls `input("  Continue? [y/N]: ")` with no timeout when `--yes` is absent. That call now sits inside `with mine_palace_lock(palace_path)` (line 254), after `get_collection` has already opened the palace. `input()` blocks until the operator answers, stdin hits EOF, or the process is killed, and the exclusive mine lease stays held the whole time. A hook `mempalace mine` then fails immediately with `MineAlreadyRunning` even though repair has not extracted or copied anything and may still be declined. An abandoned prompt is an unbounded hold, wider than the extract/backup/rebuild window. The from-sqlite branch of this same function confirms first and only then calls `rebuild_from_sqlite`, which takes the lease inside the library call. `--yes` skips the prompt; the reported runs used `--yes`.

**Suggestion:** Drop the lease before the prompt. A short hold to read the drawer count is fine, but close the Chroma client before releasing it — a live client without the lease is the race this lock exists to close. Call `confirm_destructive_action` with the lease free, then acquire again only after the user accepts, and let the existing `MineAlreadyRunning` handler exit before extraction if a miner took the lease during the prompt.

